PR #327, opened 2019-03-05
[WIP] Testing with webdriverio

### LINKED ISSUE #333: Enable frontend-tests (maybe using Webdriver IO?)
It would be very helpful to have functionality tests for each component. Something that's as close as possible to a real user clicking on buttons, entering text, selecting options, etc etc.

adrianschmidt:
## DO NOT MERGE

fix #333

Resources:
https://webdriver.io/docs/gettingstarted.html
https://github.com/blueimp/wdio

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@ www/
 src/components.d.ts
 src/**/readme.md
 docsDist/
+docsProdWww/
+reports/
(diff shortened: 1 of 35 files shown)


adrianschmidt:
This is way too old by now.

PR #401, opened 2019-08-29
fix(slider): fix background color for disabled slider

### LINKED ISSUE #304: limel-slider: doesn't look nice when disabled and value on min/ max
## Current behavior
- go to the disabled example in the docs
- enable the slider
- set the slider to max or min value
- diable it
image

## Expected behavior
the pin with the current value should have a background, so the `valuemin`/`valuemax` are covered

## Environment

- **lime-elements** version: 16.0.0

adrianschmidt:
Add a background color to the slider pin when the slider is disabled.

fix #304

### DIFF
--- a/src/components/slider/slider.scss
+++ b/src/components/slider/slider.scss
@@ -3,7 +3,15 @@
 @import '@limetech/mdc-slider/mdc-slider';
 @import "@limetech/mdc-floating-label/mdc-floating-label";
 
+/**
+ * @prop --slider-bg-color-behind-component: Color of "transparent" parts of disabled slider. Defaults to #ffffff.
+ */
+
 $colorGrey: #999999;
+
+:host {
+    --mdc-slider-bg-color-behind-component: var(--slider-bg-color-behind-component, #ffffff);
+}
 
 .mdc-slider {
     @include mdc-slider-highlight-color(var(--lime-primary-color, $mdc-theme-primary));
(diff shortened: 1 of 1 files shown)


src/components/slider/slider.scss
```diff
@@ -76,6 +84,11 @@ $colorGrey: #999999;
     top: pxToRem(-16);
     transform: none;
     width: pxToRem(40);
+
+    .mdc-slider.mdc-slider--disabled & {
+        background: var(--mdc-slider-bg-color-behind-component);
+        background: linear-gradient(90deg, rgba(255,255,255,0) 0%, var(--mdc-slider-bg-color-behind-component) 15%, var(--mdc-slider-bg-color-behind-component) 85%, rgba(255,255,255,0) 100%);
```
adrianschmidt: It's inevitable that there will be positions where the slider thumb will partially cover one of the min- and max-labels, so I added a small gradient to make it a little nicer looking when that happens (only applies when the slider is disabled).

Hot or Not?

BregenzerK: in my opinion it's rather overkill. And maybe I need to let my eyesight checked, but I really not see much of a difference especially with a white background. I'm more bothered by the fact, that the pins background is not just the pin but also the dot on the line. You can't really tell when it's white, but it looks kinda weird with a color.
image
anyhow I think I've been at that point and you can't set the background just to the pin :roll_eyes:

adrianschmidt: Yeah, but that's because you're not supposed to set the background colour to anything but the background colour. If you change the background colour that the component rests on as well, you'll see why there's a ring around the dot:

master:

image

pr:

image

adrianschmidt: The reason for the gradient is different.

Without gradient:

image image

With gradient:

image image

BregenzerK: okay, you convinced me, it's better with the gradient. But maybe we shouldn't have the `--slider-bg-color-behind-component` variable then and just set it to white?

adrianschmidt: It defaults to white, but the variable is there in case anyone ever puts it on a different background colour. They probably shouldn't in the webclient, but maybe in another application. Go wants to use lime-elements for example.

PR #408, opened 2019-09-05
refactor(tsconfig): add @limetech/lime-elements module

jgroth:
### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/interface.d.ts
+++ b/src/interface.d.ts
@@ -8,3 +8,4 @@ export * from './components/spinner/spinner.types';
 export * from './components/flex-container/flex-container.types';
 export * from './components/dialog/dialog.types';
 export * from './components/chip-set/chip.types';
+export * from './components/file/file.types';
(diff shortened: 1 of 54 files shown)


src/interface.d.ts
```diff
@@ -7,3 +7,7 @@ export * from './components/icon/icon.types';
 export * from './components/spinner/spinner.types';
 export * from './components/flex-container/flex-container.types';
 export * from './components/dialog/dialog.types';
+export * from './components/chip-set/chip.types';
+export * from './components/file/file.types';
+export * from './components/date-picker/date.types';
+export * from './components/list/list.types';
```
adrianschmidt: Had we forgotten to export these? Or are they actually supposed to be internal (in which case we shouldn't export them).

jgroth: I don't think they are supposed to be internal, I think we just forgot to export them :) All of them are used in the public API of the components

adrianschmidt: 👍

PR #412, opened 2019-09-05
feat(dialog): indicate scrollable content via a shadow

### LINKED ISSUE #312: limel-dialog: indicate scrollable content via a shadow
## New feature motivation

Because there aren't any visible scroll bars on iOS, it's pretty hard to recognise scrollable content in the dialog right now

## New feature description

add a shadow on the bottom of the dialog, that indicates invisible content

## New feature implementation

this might help: https://stijndewitt.com/2018/06/12/pure-css-drop-shadow-on-scroll/

adrianschmidt:
fix #312

shadow scroller

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dialog/dialog.scss
+++ b/src/components/dialog/dialog.scss
@@ -1,3 +1,4 @@
+@import '../../style/functions';
 @import '../../style/variables';
 
 $mdc-dialog-max-width: false;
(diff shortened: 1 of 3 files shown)


jgroth:
In Material design shadows are used to express different degrees of elevation between different surfaces and should not be used for styling purposes. If we add a shadow here, it indicates that the content has been pushed down below the header and button bar, which might look odd. I think material adds a line at the top and bottom of the content area inside dialogs to indicate that it can be scrolled.

I'm not saying we should or should not add the shadow, just raising some concerns :slightly_smiling_face:

adrianschmidt:
Good point @jgroth. Though I wonder if a line is enough to make the average user understand that something can be scrolled. (That said, I guess the same is a valid question for the shadow…)

It should be easy to replace the shadow with a 1px solid line, using the same method as now, so we can definitely try it.

Thoughts @magnusfagerlund?

adrianschmidt:
Btw, I notice there is a line in the dialog in the gif I made. That might have been added in MDC since this issue was opened. It doesn't disappear when the dialog is scrolled to the bottom though, but maybe it shouldn't according to the Material spec?

jgroth:
Yeah I think it's always supposed to be there when the content is scrollable. There's one added to the top as well if there is a header present

magnusfagerlund:
Is there any Material Design example on how they solve this?

adrianschmidt:
Looking at the Material guidelines for elevation, it actually looks as though there ought to be a shadow there, since the content can scroll in underneath it. Things that have the same elevation should collide, while things that don't shouldn't.

https://material.io/design/environment/elevation.html

adrianschmidt:
There are already lines between header and body, and between body and footer, when the content is scrollable:

image

So, if that's not enough of an indication to users that the dialog can be scrolled (which I completely understand, unless the content happens to be obviously cut, like in this screenshot, I think the line is a much too subtle clue), we'll need something more. Adding a shadow seems to be in line with the Material guidelines for elevation, even though it's not there in their specification for dialogs.

adrianschmidt:
Kia ok:d this 👍

adrianschmidt:
@BregenzerK Maybe you would like to do the review, since you opened the issue? 🙂

BregenzerK:
> @BregenzerK Maybe you would like to do the review, since you opened the issue?

I can't remember doing that but sure :laughing:

PR #413, opened 2019-09-06
fix(icon): move icons to private repo and package

adrianschmidt:
Remove the icon set from lime-elements, and move it to a private repo and private package.

BREAKING CHANGE: The icon set previously delivered with lime-elements has been removed. The icons will be provided by the Lime products which use lime-elements. External consumers of lime-elements need to supply their own icons.

**Note to reviewer:** Use `git diff --diff-filter=d HEAD^` to see all changes excluding deleted files.

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [x] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


adrianschmidt:
I'm not quite sure why this isn't building on Jenkins. The problem is that it can't reach Verdaccio, but why on earth it can't reach Verdaccio in this build when it can reach it just fine in the lime-webclient build, I have no idea.

BregenzerK:
- update link to icons8 on `src/components/icon/icon.mdx`

PR #425, opened 2019-09-16
chore(package): update @stencil/core to v1.4.0

adrianschmidt:
The `dispatch-resize-events` util has been updated to create the UIEvent in the modern way, since the way it was doing it is deprecated and started causing a compiler error with this update of Stencil. So that should be tested.

There's a slider in one of the dialog examples in the docs. If the resize-event happens as it should, that slider will work as expected. If the resize event doesn't happen, the slider should, at least intermittently, have an incorrect scaling for the thumb-placement.

### Browsers tested:

Windows:
- [x] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [x] Firefox
- [x] Safari

Mobile:
- [ ] Chrome on Android
- [x] iOS

### DIFF
--- a/src/util/dispatch-resize-event.ts
+++ b/src/util/dispatch-resize-event.ts
@@ -1,11 +1,4 @@
 export const dispatchResizeEvent = () => {
-    /*
-     * The shorthand (`window.dispatchEvent(new Event('resize'));`)
-     * causes compiler errors, so we go the long way around.
-     * See https://stackoverflow.com/a/1818513/280972
-     * /Ads
-     */
-    const resizeEvent = window.document.createEvent('UIEvents');
-    resizeEvent.initUIEvent('resize', true, false, window, 0);
+    const resizeEvent = new UIEvent('resize', { view: window, detail: 0 });
     window.dispatchEvent(resizeEvent);
 };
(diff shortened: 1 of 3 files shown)


src/util/dispatch-resize-event.ts
```diff
@@ -1,11 +1,4 @@
 export const dispatchResizeEvent = () => {
-    /*
-     * The shorthand (`window.dispatchEvent(new Event('resize'));`)
-     * causes compiler errors, so we go the long way around.
-     * See https://stackoverflow.com/a/1818513/280972
-     * /Ads
-     */
-    const resizeEvent = window.document.createEvent('UIEvents');
-    resizeEvent.initUIEvent('resize', true, false, window, 0);
+    const resizeEvent = new UIEvent('resize', { view: window, detail: 0 });
```
adrianschmidt: 😞

This isn't working as it should in Firefox on Windows:

image

(Don't mind the colour, that's just to make sure the browser has actually loaded from the branch, and isn't using a cached copy.)

adrianschmidt:
Replaced by #447

PR #443, opened 2019-09-25
Improve keyboard navigation for input chip-set, list, and picker

### LINKED ISSUE #428: limel-picker: doesn't blur after focus when already full
## Current behavior

1. Use the first example in the limel-picker docs.
2. Select a value.
3. Click on the picker (but not on the chip), so that the picker is focused.
4. Click outside the picker (or tab away).

- The picker still looks like it's focused (green line underneath).

## Expected behavior

- The picker should not look like it's focused.

## Environment

- **lime-elements** version: v23.0.2
- Framework used: Replicable in the docz-app.
- Logs:

adrianschmidt:
Chips can't be reached by keyboard at the moment.

Fix #428 

***NB!*** We really should test this on mobile.

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [x] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -66,6 +66,12 @@ export class ChipSet {
     public searchLabel: string;
 
     /**
+     * Whether the input field should be emptied when the chip-set loses focus.
+     */
+    @Prop({ reflectToAttr: true })
+    public emptyInputOnBlur: boolean = true;
+
+    /**
      * Dispatched when a chip is interacted with
      */
     @Event()
@@ -140,6 +146,17 @@ export class ChipSet {
     public async setFocus() {
         this.editMode = true;
         this.host.shadowRoot.querySelector('input').focus();
+    }
+
+    /**
+     * Used to empty the input field. Used in conjunction with `emptyInputOnBlur` to let the
+     * consumer control when the input is emptied.
+     *
+     * @returns {Promise<void>} does not return anything, but methods have to be async
+     */
+    @Method()
+    public async emptyInput() {
+        this.textValue = '';
     }
 
     public componentDidLoad() {
@@ -309,8 +326,10 @@ export class ChipSet {
      * @returns {void}
      */
     private handleInputBlur() {
+        if (this.emptyInputOnBlur) {
+            this.textValue = '';
+        }
         this.editMode = false;
-        this.textValue = '';
         this.blurred = true;
 
         // This timeout is needed in order to let a new element receive focus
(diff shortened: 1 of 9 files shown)


src/components/chip-set/chip-set.tsx
```diff
@@ -241,30 +242,16 @@ export class ChipSet {
...
                     'mdc-text-field': true,
                     'mdc-text-field--invalid': this.isInvalid(),
                 }}
-                onFocus={this.handleTextFieldFocus}
-                tabindex="0"
+                onClick={this.handleTextFieldFocus}
             >
                 <div class="mdc-chip-set mdc-chip-set--input">
-                    {this.value.map(this.renderChip)}
+                    {this.value.map(this.renderInputChip)}
```
adrianschmidt: We're in the `renderInputChips` function, so we already know that we want to render input-chips.

src/components/chip-set/chip-set.tsx
```diff
@@ -273,38 +260,33 @@ export class ChipSet {
...
                 </div>
                 <label
                     class={{
                         'mdc-floating-label': true,
-                        'mdc-floating-label--float-above': !!(
-                            textValue || this.editMode
-                        ),
                         'mdc-text-field--disabled': this.disabled,
                         'mdc-text-field--required': this.required,
+                        'force-float': !!(this.value.length || this.editMode),
```
adrianschmidt: This works because `MDCTextField` doesn't know about this class, and doesn't try to add or remove it.

src/components/chip-set/chip-set.tsx
```diff
@@ -273,38 +260,33 @@ export class ChipSet {
...
-                >
-                    {searchLabel}
-                </label>
                 <div class="mdc-line-ripple" />
-            </div>
-        );
+            </div>,
+            /*<div class="mdc-text-field-helper-line">
+                <div class="mdc-text-field-helper-text">{this.searchLabel}</div>
+            </div>*/
```
adrianschmidt: I found this alternative to using a placeholder. It shows the text underneath the input field when the field is focused. The benefit is that the helper-text is visible even when the user has already started writing something. That's unless we cover it up with a menu of course, which we would in the case of the picker. Still an interesting approach which I think we should consider, which is why I temporarily left this here.

src/components/chip-set/chip-set.tsx
```diff
@@ -384,7 +366,6 @@ export class ChipSet {
             case 'filter':
                 return this.renderFilterChip(chip);
 
-            case 'input':
```
adrianschmidt: If the type is `input`, we no longer get here in the first place.

adrianschmidt:
> bug when chip-set is required:
> 
> * select a value, but state stays invalid
> 
> great progress besides that!

⚡️

PR #453, opened 2019-10-02
fix(chip-set): fix incorrect look and behavior when input chip-set is disabled

### LINKED ISSUE #429: limel-picker: looks bad when readonly (disabled)
## Current behavior

1. Use the first example in the limel-picker docs.
2. Select a value.
3. Disable the picker.

- The picker is very faded, and the chip still has a remove-button. The label has a background (and in lime-webclient, the whole picker get's a background, though I don't quite know why).
- The picker doesn't fire `interact` events when clicking on the chip.
- You can't focus the chips in the picker.

image

(Screenshot from a branch)

## Expected behavior

- The chip should not have a remove-button. It also probably shouldn't be faded.
- It should be possible to click on it and have the `interact` event fire.
- It should be possible to focus the chips.

If, for some reason we want to keep the possibility to disable clicks on the chip, we might want to have `disabled` fade the chip like now, but use a new `readonly` attribute for the above behaviour.

image

## Environment

- **lime-elements** version: v23.0.2
- Framework used: Replicable in the docz-app.
- Logs:

adrianschmidt:
fix #429

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -18,6 +18,7 @@ $mdc-text-field-fullwidth-bottom-line-color: $lime-text-field-bottom-line-color;
  * @prop --icon-background-color: Background color of the icon. Defaults to transparent.
  * @prop --icon-color: Color of the icon. Defaults to 54% black.
  * @prop --background-color: Background color of the field when type is set to input.
+ * @prop --background-color-disabled: Background color of the field when type is set to input and the component is disabled or readonly.
  * @prop --input-chip-set-selected-chip-color: Color of the highlight around selected chips in input chip-sets.
  */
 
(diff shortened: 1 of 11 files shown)


src/components/chip-set/chip-set.scss
```diff
@@ -21,14 +21,6 @@ $mdc-text-field-fullwidth-bottom-line-color: $lime-text-field-bottom-line-color;
  * @prop --input-chip-set-selected-chip-color: Color of the highlight around selected chips in input chip-sets.
  */
 
-:host([disabled]) {
-    pointer-events: none;
-
-    .mdc-chip-set {
-        opacity: 0.5;
-    }
-}
```
adrianschmidt: Oops… removing this means it's no longer possible to disable chip-sets which are not of type `input`.

I'll fix, but don't merge for now!

adrianschmidt: ⚡️ (see new lines 31-35 below)

PR #454, opened 2019-10-02
Fix hiding of input

### LINKED ISSUE #455: limel-picker: Hide the input-element when focus leaves component, to avoid empty space below the chips when the picker is narrow (like on the card in the webclient).
We already know the difference between focus leaving the input for the dropdown, and focus leaving the component completely. So we should be able to hide the input when focus leaves the component completely.

adrianschmidt:
Based on #453. I would suggest reviewing and merging that one first, and then pointing this PR to master, before reviewing this. /Ads

fix #455 

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -126,6 +126,8 @@ export class ChipSet {
     @State()
     private inputChipIndexSelected: number = null;
 
+    private inputHidden: boolean = false;
+
     private mdcChipSet: MDCChipSet;
     private mdcTextField: MDCTextField;
     private handleKeyDown = handleKeyboardEvent;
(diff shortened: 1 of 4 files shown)


src/components/chip-set/chip-set.tsx
```diff
@@ -347,6 +353,16 @@ export class ChipSet {
         }, 0);
     }
 
+    private syncEmptyInput() {
+        this.textValue = '';
+
+        if (this.value && this.value.length) {
+            // If there are chips in the picker, hide the input to avoid the input
+            // being placed on a new line and adding ugly space beneath the chips.
+            this.inputHidden = true;
```
adrianschmidt: It seems like the input is sometimes hidden when it shouldn't be. I'll need look further into this.

adrianschmidt: ⚡️

PR #466, opened 2019-10-09
[WIP] Dividers in list and icon size in picker dropdown

adrianschmidt:
### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/chip-set/chip-set.tsx
```diff
@@ -63,6 +63,12 @@ export class ChipSet {
     @Prop({ reflectToAttr: true })
     public readonly: boolean = false;
 
+    /**
+     * Set to `true` to prevent addition of new chips in an input chip-set.
+     */
+    @Prop({ reflectToAttr: true })
+    public full: boolean = false;
```
jgroth: What's the difference between this and `readonly`?

adrianschmidt: In readonly you can't edit the content at all, neither removing nor adding chips. When the chip-set is full, you can't add more chips, but you can still remove the ones that are already there (well, the *one*, since for the time being the only limit that can be placed is 1).

jgroth: Ah ok!

I think I have some issues with this. Is this something we should handle inside the component? It seems we are adding a prop when we don't have to. Since the consumer always have to set this from the outside, the consumer should be aware of when another chip can be added and when it is not allowed. The same behavior can be achieved with the `change` event and not giving the component any new chips in the `value` prop.

I'm also not sure I like the word `full`, and this seems kind of related to validation, so if we want to have this kind of functionality we could rename it to `maxItems`. Adding validation seems like a much bigger thing to add though and if we want to do that we should maybe do it in another PR? 

Just some thoughts :slightly_smiling_face:

adrianschmidt: Hmm… yeah, adding a `maxItems` property instead is a good idea. I don't know why I didn't think of that 😄

I was just thinking "limel-picker has the concept of a picker being full, but limel-chip-set doesn't", and then… well… 😂

src/components/chip-set/chip-set.tsx
```diff
@@ -63,6 +63,12 @@ export class ChipSet {
     @Prop({ reflectToAttr: true })
     public readonly: boolean = false;
 
+    /**
+     * Set to `true` to prevent addition of new chips in an input chip-set.
```
adrianschmidt: We need to update this comment.

adrianschmidt:
> Looks good! We need to fix the documentation that Adrian mentioned, but I'll approve this anyway and assume it get's fixed before it is merged 😄

Great! I'll just have a quick look through the changes too before we merge.

PR #474, opened 2019-10-17
feat(input-field): add support for a helper text next to the input field

jgroth:
### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -43,6 +43,12 @@ export class InputField {
     public label: string;
 
     /**
+     * Optional helper text to display below the input field when it has focus
+     */
+    @Prop({ reflectToAttr: true })
+    public helperText: string;
+
+    /**
      * Set to `true` to indicate that the field is required.
      * Defaults to `false`.
      */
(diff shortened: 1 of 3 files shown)


src/components/input-field/input-field.tsx
```diff
@@ -167,6 +173,9 @@ export class InputField {
                 {this.renderTrailingIcon()}
                 <div class="mdc-line-ripple" />
             </label>,
+            <div class="mdc-text-field-helper-line">
+                <div class="mdc-text-field-helper-text">{this.helperText}</div>
+            </div>,
```
adrianschmidt: I haven't tried it yet, so maybe MDC takes care of it, but I suspect this should only be output if there is actually any text in `this.helperText`?

adrianschmidt: ⚡️

src/components/input-field/input-field.tsx
```diff
@@ -112,6 +112,7 @@ export class InputField {
     private action: EventEmitter<void>;
 
     constructor() {
+        this.renderHelperText = this.renderHelperText.bind(this);
```
jgroth: There is no need to bind this function, since it is called with `this.`

adrianschmidt: ⚡️

PR #505, opened 2019-11-07
Update grid config

adrianschmidt:
Some breaking changes, but I think it's fairly unlikely that anyone is actually affected by them. Still, please see the individual commit messages for more info.

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/grid/grid.scss
+++ b/src/components/grid/grid.scss
@@ -4,7 +4,6 @@
  * @prop --lime-grid-area: Grid layout
  * @prop --lime-grid-columns: Number of columns in the grid, defaults to 4
  * @prop --lime-grid-cell-height: Height of grid cells, defaults to 150 pixels
- * @prop --lime-grid-cell-width: Minimum width of grid cells, defaults to 150 pixels
  */
 
 slot {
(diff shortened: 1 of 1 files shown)


src/components/grid/grid.scss
```diff
@@ -2,21 +2,31 @@
...
-        var(--lime-grid-columns, 8),
-        minmax(var(--lime-grid-cell-width, pxToRem(150)), 1fr)
+        var(--lime-grid-columns, 4),
+        minmax(0, 1fr)
     );
-    grid-gap: pxToRem(15);
-    grid-auto-flow: dense;
+    grid-gap: var(--lime-grid-gutter, pxToRem(16));
+    grid-auto-flow: row dense;
     grid-auto-rows: var(--lime-grid-cell-height, pxToRem(150));
```
adrianschmidt: Considering the columns are now rather narrow, we might want the minimum size of the rows to be less than 150px.

Also, we need to make sure this actually sets the row height properly. I think it probably just sets the minimum height. If we need to change it, we should definitely change it before merging this, since that would be another breaking change.

PR #524, opened 2019-11-26
feat(grid): set default row height to 2.5rem (40px)

adrianschmidt:
BREAKING CHANGE: The default row height of the limel-grid component has been set to 2.5rem (40px). The row height of any given instance of limel-grid can be set using the `--lime-grid-cell-height` CSS variable.

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/grid/grid.scss
+++ b/src/components/grid/grid.scss
@@ -3,7 +3,7 @@
 /**
  * @prop --lime-grid-area: Grid layout
  * @prop --lime-grid-columns: Number of columns in the grid, defaults to 4
- * @prop --lime-grid-cell-height: Height of grid cells, defaults to `4rem` (64px)
+ * @prop --lime-grid-cell-height: Height of grid cells, defaults to `2.5rem` (40px)
  * @prop --lime-grid-gutter: width of the gutter between columns and rows, defaults to `1rem` (16px) - see https://material.io/design/layout/responsive-layout-grid.html#breakpoints for guidelines
  */
 
@@ -25,7 +25,7 @@ slot {
     );
     grid-gap: var(--lime-grid-gutter, pxToRem(16));
     grid-auto-flow: row dense;
-    grid-auto-rows: var(--lime-grid-cell-height, pxToRem(64));
+    grid-auto-rows: var(--lime-grid-cell-height, pxToRem(40));
     margin: 0;
     height: 100%;
     width: 100%;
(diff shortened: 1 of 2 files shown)


calleboketoft:
Activity Matrix uses Flex container, addon-acd has no grid, and Infotiles explicitly sets cell height in the grid so those are safe.

adrianschmidt:
> Are there any known current implementations of plugins where this could lead to issues?

Since infotiles is setting the height explicitly, I think there probably aren't any that would be affected.

The breaking-ness of this change should be propagated to lime-webclient in the commit that ups the dependency. From there on, I don't know how such changes may or may not be propagated to the list of changes that are communicated for a new release. A breaking change like this should definitely be communicated clearly (even though it's probably fine in this specific case).

@magnusfagerlund Ping on ☝️

adrianschmidt:
I would really like to show this to either @Shaddla or Kia before we merge, because it's a trade-off between flexibility and structure, and we really need to get it right, because changing it later will require manually updating *every single* customer configuration (aka, Horrible Mess™).

calleboketoft:
I agree, it's an important thing UX wise so the UX team must be involved.

Shaddla:
Thank you Adrian! Please run this by Kia.

Kiarokh:
> I would really like to show this to either @Shaddla or Kia before we merge, because it's a trade-off between flexibility and structure, and we really need to get it right, because changing it later will require manually updating _every single_ customer configuration (aka, Horrible Mess™).

Thanks Adrian. Let's talk; anytime! I'm always available. I don't really understand the issue when reading the comments here. Would be much easier if you show me. :) Cheers

adrianschmidt:
> Thanks Adrian. Let's talk; anytime! I'm always available. I don't really understand the issue when reading the comments here. Would be much easier if you show me. :) Cheers

Hey, there you are! I couldn't find you when trying to @-mention you yesterday.

Of course I'll show it to you! 😄 I just wrote the note to have the current status available on the PR. I was going to show you yesterday, but got stuck on some issues with my dev-environment while trying to set it up. I'll come by today though!

adrianschmidt:
Kia was ok with this, so I'm merging.

PR #527, opened 2019-12-02
fix(dialog): prevent header from shrinking due to large body

FredrikWallstrom:
prevent content to hide the header on scroll bounce

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [x] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dialog/dialog.scss
+++ b/src/components/dialog/dialog.scss
@@ -52,6 +52,7 @@ slot[name="header"] {
         display: flex;
         align-items: center;
         margin: pxToRem(30) pxToRem(24) pxToRem(5);
+        flex-shrink: 0;
 
         limel-icon {
             margin-right: pxToRem(15);


src/components/dialog/dialog.scss
```diff
@@ -52,6 +52,7 @@ slot[name="header"] {
         display: flex;
         align-items: center;
         margin: pxToRem(30) pxToRem(24) pxToRem(5);
+        min-height: pxToRem(46);
```
adrianschmidt: Using `flex-shrink: 0;` fixes the issue with parts of the badge being cut off, and we should have that regardless.

I'm testing this in Safari, and while I get the list to bounce, I'm not seeing the fixed element bouncing, with or without the `min-height` or the `flex-shrink`. So I'm wondering how you implemented that element in your plugin?

adrianschmidt: For reference: Fredrik used `position:sticky` while I used `position:absolute`. `sticky` is very nice in that it just works, but has the small downside that when the list is scrolled to the top in for example Safari, the sticky element bounces together with the list. When using `absolute`, this doesn't happen. On the other hand, when using `absolute`, the position and width of the element has to be set, and you also need to add whitespace before the next element in the dialog body, to match the height of the absolutely positioned element.

PR #550, opened 2020-01-09
chore(github actions): switch from forked workflows to parents

adrianschmidt:
### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -11,7 +11,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v1
-      - uses: Lundalogik/commitlint-github-action@master
+      - uses: wagoid/commitlint-github-action@v1
 
   autosquash:
     name: Block Autosquash Commits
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-checks.yml
```diff
@@ -11,15 +11,15 @@ jobs:
...
   autosquash:
     name: Block Autosquash Commits
     runs-on: ubuntu-latest
 
     steps:
       - name: Block Autosquash Commits
-        uses: Lundalogik/block-autosquash-commits-action@master
+        uses: xt0rted/block-autosquash-commits-action@v2.0.0
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
```
adrianschmidt: @kickapoo As far as I understood, you said it should be safe to use third party actions, even though we give them the `secrets.GITHUB_TOKEN`, right?

PR #555, opened 2020-01-17
feat(form): first basic version of form component (unreleased)

jgroth:
Connected to Lundalogik/crm-feature#947

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,9 @@
     "awesome-debounce-promise": "^2.1.0",
     "chart.js": "^2.8.0",
     "jsx-dom": "^6.4.0",
-    "moment": "^2.24.0"
+    "moment": "^2.24.0",
+    "react-jsonschema-form": "^1.8.1",
+    "react-shadow-dom-retarget-events": "^1.0.11"
   },
   "devDependencies": {
     "@commitlint/config-conventional": "8.3.4",
--- a/src/components/form/form.mdx
+++ b/src/components/form/form.mdx
@@ -1,0 +1,19 @@
+---
+name: Form
+route: /form
+menu: Components
+---
+
+# Form
+
+**<span style={{color:"red"}}>The form component is still work in progress and has not yet been released.</span>**
+
+<limel-props name="limel-form" />
+
+## Example
+
+<limel-example name="limel-example-form" />
+
+### Dynamic schema
+
+<limel-example name="limel-example-dynamic-form" path="form"/>
--- a/src/components/form/form.tsx
+++ b/src/components/form/form.tsx
@@ -1,0 +1,93 @@
+import {
+    Component,
+    Element,
+    Event,
+    EventEmitter,
+    h,
+    Prop,
+} from '@stencil/core';
+import React from 'react';
+import { render, unmountComponentAtNode } from 'react-dom';
+import JSONSchemaForm from 'react-jsonschema-form';
+import retargetEvents from 'react-shadow-dom-retarget-events';
+import {
+    ArrayFieldTemplate,
+    FieldTemplate,
+    ObjectFieldTemplate,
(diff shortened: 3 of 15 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "postinstall": "node postinstall.js",
     "build": "cross-env-shell NODE_ENV=prod SASS_PATH=node_modules \"stencil build --config stencil.config.dist.ts\"",
     "cm": "git-cz",
-    "dev": "cross-env-shell SASS_PATH=node_modules \"stencil build --dev --docs\"",
+    "dev": "cross-env-shell SASS_PATH=node_modules \"stencil build --dev --docs --watch\"",
     "watch": "cross-env-shell SASS_PATH=node_modules \"stencil build --dev --watch --docs\"",
```
adrianschmidt: There's already `watch` for watching. But if we'd rather the `dev` command also watched, we can remove `watch`.

adrianschmidt: There's already `watch` for watching. But if we'd rather the `dev` command also watched, we can remove `watch`.

jgroth: Yeah that was not supposed to be there :smile:

stencil.config.ts
```diff
@@ -29,6 +29,14 @@ export const config: Config = {
...
     namespace: 'lime-elements',
     outputTargets: [targetDist, targetWww],
+    commonjs: {
+        namedExports: {
+            'node_modules/react-dom/index.js': [
+                'render',
+                'unmountComponentAtNode',
+            ],
+        },
+    },
```
adrianschmidt: What does this do?

src/components/form/form.tsx
```diff
@@ -0,0 +1,93 @@
...
+                    ObjectFieldTemplate: ObjectFieldTemplate,
+                },
+                []
+            ),
+            rootElement
+        );
+    }
+
+    private handleChange(event: any) {
+        this.change.emit(event.formData);
```
adrianschmidt: This should stop propagation of the original event, or we'll have two change-events.

src/components/form/form.mdx
```diff
@@ -0,0 +1,17 @@
+---
+name: Form
+route: /form
+menu: Components
+---
+
+# Form
+
```
adrianschmidt: I assume these docs will be visible once we create a new release, even if the component itself isn't bundled into the release, so we should write something about the component not being available yet.

stencil.config.ts
jgroth: I think it's needed in order to convert the CommonJS modules exported by React to ES6 modules. Rollup was not able to find them otherwise.

src/components/form/form.tsx
jgroth: I will test it, but I _think_ we are fine. The original event is from React and I'm not sure it's a proper DOM event

src/components/form/form.mdx
jgroth: I was hoping it would not be visible, but you might be right! The `excludeSrc` property will probably not work on the docs files. Maybe there is some other way of hiding it :thinking:

src/components/form/form.tsx
jgroth: `event` here is just a plain object, so there is no way to stop the event propagation. It will not propagate anyway since it's not a real event :slightly_smiling_face:

adrianschmidt: Ah, ok 👍

PR #559, opened 2020-01-21
limel-select for limel-form

jgroth:
Connected to Lundalogik/crm-feature#947

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/select/select.template.tsx
+++ b/src/components/select/select.template.tsx
@@ -8,6 +8,7 @@ interface SelectTemplateProps {
     options: Option[];
     value: Option | Option[];
     label?: string;
+    helperText?: string;
     multiple?: boolean;
     native: boolean;
 
@@ -42,11 +43,13 @@ export const SelectTemplate: FunctionalComponent<SelectTemplateProps> = props =>
         'limel-select--required': props.required,
         'limel-select--invalid': !isValid,
         'limel-select--empty': !hasValue,
+        'limel-select--with-helper-text': typeof props.helperText === 'string',
     };
 
     return (
         <div class={classList}>
             <SelectValue {...props} hasValue={hasValue} />
+            <HelperText text={props.helperText} />
             <SelectDropdown {...props} />
         </div>
     );
@@ -83,6 +86,21 @@ const SelectValue: FunctionalComponent<SelectTemplateProps & {
             <span class={labelClassList}>{props.label}</span>
             <div class={lineClassList} />
         </div>
+    );
+};
+
+const HelperText: FunctionalComponent<{ text: string }> = props => {
+    if (typeof props.text !== 'string') {
+        return;
+    }
+
+    return (
+        <p
+            class="mdc-select-helper-text mdc-select-helper-text--persistent"
+            aria-hidden="true"
+        >
+            {props.text.trim()}
+        </p>
     );
 };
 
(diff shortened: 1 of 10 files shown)


tslint.js
```diff
@@ -89,6 +89,7 @@ module.exports = {
         'ter-no-script-url': true,
         'ter-no-self-compare': true,
         'triple-equals': true,
+        'prefer-conditional-expression': false,
```
adrianschmidt: 👍

Yeah, I agree. I ran the linter yesterday, and changed once instance to a conditional. I'm not sure if it was in this branch. Perhaps I never pushed the change? But anyway, looking at the result, I did feel that it made it less readable. I don't know why I didn't just decide to change the rule then, so thanks for taking action! 😄

jgroth: Yeah I had the same issue in another branch and didn't bother to change this rule there as well :stuck_out_tongue: , hoping that this would be merged before the other branch. I think the ternary operator always makes things less readable, unless the expression is really short on a single line (which is almost never the case)

adrianschmidt: I might be able to find the other instance and change it back. If I remember 🙂

src/components/select/select.template.tsx
```diff
@@ -42,11 +43,13 @@ export const SelectTemplate: FunctionalComponent<SelectTemplateProps> = props =>
         'limel-select--required': props.required,
         'limel-select--invalid': !isValid,
         'limel-select--empty': !hasValue,
+        'limel-select--with-helper-text': typeof props.helperText === 'string',
     };
 
     return (
         <div class={classList}>
             <SelectValue {...props} hasValue={hasValue} />
+            <HelperText text={props.helperText} />
```
adrianschmidt: This is a little annoying:

image

But that's down to MDC of course. I wonder if we could clearfix the helper-text or something… or maybe we could clearfix the bottom inside of the form frame… 🤔

adrianschmidt: I can experiment a bit with that later.

jgroth: Hmm yeah maybe it should be a bit closer to the line of the select as well? I'm trying to compare it to the input field which seems to be a bit closer. Also, I'm not sure we want to have the line around the form, but that's related to the field template. However, I guess this will still be an issue if there are other fields below this one, they will probably flow into each other

jgroth: Pushed some changes to the positioning. The issue seems to be that the height of the field is hard coded by material and doesn't take the helper text into account

adrianschmidt: Ah, yeah, there are definitely some differences. The text is further away here, and not indented like for input-field.

I saw you just pushed some changes for this, so I'll take a look at it with those.

jgroth: I just remembered though that the field is not styled by material since we are not using their select component haha :upside_down_face: Anyway, it looks a bit better now... :smile:

adrianschmidt: Haha, yeah, I had also forgotten that, but I was about to comment that it doesn't seem like it's styled by material, at least not completely 😄

It looks much better now! And this should be a step on the way towards adding a text for when a select is required but has been left empty, as well 👍

adrianschmidt: Well isn't that beautiful! 🥳

image

PR #563, opened 2020-01-22
fix(select): make select work on mobile devices again

jgroth:
### Browsers tested:

Windows:
- [x] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/select/select.scss
+++ b/src/components/select/select.scss
@@ -4,11 +4,9 @@ $mdc-select-fill-color: $mdc-theme-background;
 $mdc-select-bottom-line-idle-color: $lime-text-field-bottom-line-color;
 $mdc-select-bottom-line-hover-color: $lime-text-field-bottom-line-color;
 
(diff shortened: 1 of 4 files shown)


src/components/select/select.scss
```diff
@@ -135,3 +116,13 @@ $mdc-select-bottom-line-hover-color: $lime-text-field-bottom-line-color;
         }
     }
 }
+
+select.mdc-select__native-control {
```
adrianschmidt: Huh? Still something that's both native and mdc?

jgroth: No MDC select should not be used at all, I just reused the same class-name we already had. I can change it if you want

adrianschmidt: It's probably less likely to cause confusion if renamed, yes 😄

PR #566, opened 2020-01-23
fix(form): remove double titles and descriptions from form fields

jgroth:
### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/form/templates/field.ts
+++ b/src/components/form/templates/field.ts
@@ -1,1 +1,12 @@
-export const FieldTemplate = undefined; // TODO implement field template
+import React from 'react';
+
+export const FieldTemplate = props => {
+    const { classNames, children } = props;
+    return React.createElement(
+        'div',
+        {
+            className: classNames,
+        },
+        children
+    );
+};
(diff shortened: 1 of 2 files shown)


src/components/form/form.scss
```diff
@@ -0,0 +1,6 @@
+@import '../../style/variables';
+
+limel-select {
+    display: block;
+    margin-bottom: pxToRem(38);
```
adrianschmidt: image

I think we need to only apply this margin if there's a helper-text. It creates a huge gap otherwise:

image

adrianschmidt: And it probably should go in the limel-select css, unless there's a particular reason it's only needed when used in a form?

src/components/form/templates/field.ts
```diff
@@ -1 +1,12 @@
...
+export const FieldTemplate = props => {
+    const { classNames, children } = props;
+    return React.createElement(
+        'div',
+        {
+            className: classNames,
+        },
+        children
+    );
+};
```
adrianschmidt: That was a lot simpler than I thought it would be! 👌

PR #567, opened 2020-01-23
Limel form validation

jgroth:
### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/form.tsx
+++ b/src/components/form/form.tsx
@@ -10,6 +10,7 @@ import React from 'react';
 import { render, unmountComponentAtNode } from 'react-dom';
 import JSONSchemaForm from 'react-jsonschema-form';
 import retargetEvents from 'react-shadow-dom-retarget-events';
+import { FormError, ValidationStatus } from './form.types';
 import {
     ArrayFieldTemplate,
     FieldTemplate,
@@ -41,11 +42,22 @@ export class Form {
     @Event()
     public change: EventEmitter<object>;
 
+    /**
+     * Emitted when the validity of the form changes, or when
+     * a change is made to an invalid form
+     */
+    @Event()
+    public validate: EventEmitter<ValidationStatus>;
+
     @Element()
     private host: HTMLElement;
 
+    private form: any;
+    private isValid = true;
+
     public constructor() {
         this.handleChange = this.handleChange.bind(this);
+        this.setForm = this.setForm.bind(this);
     }
 
     public render() {
@@ -55,10 +67,12 @@ export class Form {
     protected componentDidLoad() {
         this.reactRender();
         retargetEvents(this.host.shadowRoot);
+        this.validateForm(this.value);
     }
 
     protected componentDidUpdate() {
         this.reactRender();
+        this.validateForm(this.value);
     }
 
     protected componentDidUnload() {
@@ -80,6 +94,7 @@ export class Form {
                     FieldTemplate: FieldTemplate,
                     ArrayFieldTemplate: ArrayFieldTemplate,
                     ObjectFieldTemplate: ObjectFieldTemplate,
+                    ref: this.setForm,
                 },
                 []
             ),
@@ -87,7 +102,25 @@ export class Form {
         );
     }
 
+    private setForm(form: any) {
+        this.form = form;
+    }
+
     private handleChange(event: any) {
         this.change.emit(event.formData);
+    }
+
+    private validateForm(value: object) {
+        const errors: FormError[] = this.form.validate(value).errors;
+        const status: ValidationStatus = {
+            valid: errors.length === 0,
+            errors: errors,
+        };
+
+        if (this.isValid !== status.valid || !status.valid) {
+            this.validate.emit(status);
+        }
+
+        this.isValid = status.valid;
     }
 }
--- a/src/components/form/widgets/date-picker.ts
+++ b/src/components/form/widgets/date-picker.ts
@@ -1,3 +1,4 @@
+import moment from 'moment/moment';
 import React from 'react';
 import { DateType } from '../../date-picker/date.types';
 
@@ -53,7 +54,7 @@ export class DatePicker extends React.Component {
         }
     }
 
-    private handleChange(event: CustomEvent<boolean>) {
+    private handleChange(event: CustomEvent<Date>) {
         const props = this.props;
         event.stopPropagation();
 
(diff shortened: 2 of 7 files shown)


src/components/form/form.tsx
```diff
@@ -41,11 +42,21 @@ export class Form {
     @Event()
     public change: EventEmitter<object>;
 
+    /**
+     * Emitted when the validity of the form changes
```
adrianschmidt: > Emitted when the validity of the form changes

That's not quite true as far as I can see. If I read the code correctly, it will be emitted any time the validity changes, but also any time the form is updated and it fails validation, regardless of whether it was already invalid before the update.

Not that we necessarily have to specify to that level of detail. I just wanted to check that I've read the code correctly, and that the code is as intended? I assume we need to send the event whenever the form is invalid, regardless of prior state, because the error messages might have changed?

jgroth: Yeah that was what I intended, I think I forgot to push the updated docstring, becuase I'm quite sure I wrote an update to it at some point :smile: 
But you agree that it should work that way? It seems correct, right?

src/components/form/widgets/date-picker.ts
```diff
@@ -53,15 +54,16 @@ export class DatePicker extends React.Component {
         }
     }
 
-    private handleChange(event: CustomEvent<boolean>) {
+    private handleChange(event: CustomEvent<Date>) {
```
adrianschmidt: Oops… I guess I missed this when reviewing the PR for the datepicker widget?

jgroth: Yeah, copy + paste mistake :frowning_face:

src/components/form/widgets/date-picker.ts
```diff
@@ -53,15 +54,16 @@ export class DatePicker extends React.Component {
...
         const props = this.props;
         event.stopPropagation();
 
         if (!props.onChange) {
             return;
         }
 
-        props.onChange(event.detail);
+        const dateString = moment(event.detail).format('YYYY-MM-DD');
```
adrianschmidt: This seems to work on my machine, and I've got American dates because I have my browser set to English. So the hardcoded format here doesn't seem to cause any issues even though the format shown might be different. I'm not quite sure what happens to the value once submitted though? It's sent to the server I assume? Does it get converted to a full JSON date first?

src/components/form/form.tsx
adrianschmidt: Yes, that seems correct to me 👍

src/examples/form/form.tsx
```diff
@@ -1,60 +1,116 @@
...
         color: {
             type: 'string',
-            title: 'Color',
-            enum: ['red', 'green', 'blue'],
+            title: 'Favorite cape color',
+            description: 'Preferably matching your boots',
+            oneOf: [
+                {
+                    type: 'string',
+                    const: 'red',
```
adrianschmidt: Is the property name `const` decided by the form-lib we're using?

src/components/form/widgets/date-picker.ts
jgroth: No the date submitted will be in the format specified here as a string. I tried `date.toISOString()` first, but the form would never be valid since the format is specified as `date` in the schema and must be a valid representation according to the "full-date" in RFC 3339 https://tools.ietf.org/html/rfc3339#section-5.6

src/examples/form/form.tsx
jgroth: It's part of the JSON schema spec

src/components/form/widgets/date-picker.ts
jgroth: Maybe there is a better way than using Moment? :thinking:

src/examples/form/form.tsx
```diff
@@ -1,60 +1,116 @@
...
-        date: {
-            type: 'string',
-            title: 'Date',
-            format: 'date',
-            description: 'Select a nice date',
+        newsletters: {
+            type: 'boolean',
+            title: 'I want to receive daily newsletters!',
+            default: false,
+            const: true,
```
adrianschmidt: > `const: true`

This makes me a bit confused… what does this actually indicate? That the value is `true` when this is checked? Or that the value has to be `true` for it to validate? I don't quite understand how this relates to what I thought `const` meant on the options above…

jgroth: `const: true` means that the value of the field must equal to `true` in order to be valid. From the example with the options above, all the options are part of a `oneOf`, so the field there will be valid if the field is equal to any of those values. Luckily, you would normally not write the JSON schema by hand :dizzy_face:

src/components/form/widgets/date-picker.ts
adrianschmidt: Hmm… flatpickr is already "validating" the value, I think, because if I enter something that can't be parsed as a date, and tab out of the field, the field is emptied. And if I enter something that can be interpreted as a date, the field is formatted to the "correct" format for that date.

So, I guess the only validation we really get to do on this is to check that it's not empty, because if it's not empty, it must be a valid date. At least I think so… 😄

src/examples/form/form.tsx
adrianschmidt: Yeah, I assumed so. Then I won't pester you with complaints about using a reserved word as a property name 😄

(I'm pretty surprised nothing in our lint-and-build chain balks at that…)

src/components/form/widgets/date-picker.ts
jgroth: Hmm yeah maybe :thinking: Since the value from the event is a `Date` object, it must be a valid date I guess :smile:

jgroth: But we cannot put that object in the form value, since the schema wants a `string`

src/examples/form/form.tsx
adrianschmidt: Ah, so that's what the `anyOf` thing means…

How come the option field is still valid if no option is selected? Because it's not required?

jgroth: Yeah, it's not required :slightly_smiling_face:

src/components/form/widgets/date-picker.ts
adrianschmidt: > Maybe there is a better way than using Moment? 🤔

For transforming dates, I think we should stick to Moment 👍

PR #568, opened 2020-01-24
feat(form): add template for object fields

jgroth:
### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/templates/object-field.ts
+++ b/src/components/form/templates/object-field.ts
@@ -1,1 +1,68 @@
-export const ObjectFieldTemplate = undefined; // TODO implement object field template
+import React from 'react';
+import { ObjectFieldProperty, ObjectFieldTemplateProps } from './types';
+
+export const ObjectFieldTemplate = (props: ObjectFieldTemplateProps) => {
+    const id = props.idSchema.$id;
+    if (id === 'root' || !isCollapsible(props.schema)) {
+        return renderFieldWithTitle(props);
+    }
+
+    if (isCollapsible(props.schema)) {
+        return renderCollapsibleField(props);
+    }
+
+    return renderProperties(props.properties);
+};
+
+function renderFieldWithTitle(props: ObjectFieldTemplateProps) {
+    return React.createElement(
+        'div',
+        {},
+        renderTitle(props.title),
+        renderDescription(props.description),
+        renderProperties(props.properties)
+    );
+}
+
+function renderCollapsibleField(props: ObjectFieldTemplateProps) {
+    return React.createElement(
+        'limel-collapsible-section',
+        {
+            header: props.title,
+        },
+        renderDescription(props.description),
+        renderProperties(props.properties)
+    );
+}
+
+function renderProperties(properties: ObjectFieldProperty[]) {
+    return properties.map(element => element.content);
+}
+
+function renderTitle(title: string) {
+    if (!title) {
+        return;
+    }
+
+    return React.createElement(
+        'h1',
+        { className: 'mdc-typography mdc-typography--headline1' },
+        title
+    );
+}
+
+function renderDescription(description: string) {
+    if (!description) {
+        return;
+    }
+
+    return React.createElement(
+        'p',
+        { className: 'mdc-typography mdc-typography--body1' },
+        description
+    );
+}
+
+function isCollapsible(schema: any) {
+    return !!schema.lime?.collapsible;
+}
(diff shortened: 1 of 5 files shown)


src/components/form/templates/object-field.ts
```diff
@@ -1 +1,63 @@
-export const ObjectFieldTemplate = undefined; // TODO implement object field template
+import React from 'react';
+import { ObjectFieldProperty, ObjectFieldTemplateProps } from './types';
+
+export const ObjectFieldTemplate = (props: ObjectFieldTemplateProps) => {
+    const id = props.idSchema.$id;
+    if (id === 'root' || !props.schema.collapsible) {
```
jgroth: `collapsible` is not part of the JSON schema specification, I just made it up here as an example of how we could display object fields. 
I think it would be neat to have the possibility to have collapsible fields, and I think we might need it for array items as well in order to render them in a nice way.
I'm not sure if this is the correct way to do this though... The spec allows custom properties to be specified in the schema, but it is recommended that a meta schema is specified as well if custom properties are added. I think we should discuss how to handle this properly

adrianschmidt: 👍

adrianschmidt: Yeah, collapsible sections will be very nice to have, I think. I wonder if we might want to make the collapsible and non-collapsible ones look more similar in the UX, since the schemas for them are identical apart from the `collapsible: true` property?

I guess we don't have to decide that in this PR though.

src/components/form/templates/object-field.ts
```diff
@@ -1 +1,63 @@
...
+    );
+}
+
+function renderCollapsibleField(props: ObjectFieldTemplateProps) {
+    return React.createElement(
+        'limel-collapsible-section',
+        {
+            header: props.title,
+        },
+        renderProperties(props.properties)
```
adrianschmidt: I think adding the description here works pretty well:

image

```suggestion
        renderDescription(props.description),
        renderProperties(props.properties)
```

src/components/form/templates/types.ts
```diff
@@ -0,0 +1,68 @@
+// These are defined by react-jsonschema-form and added
+// here as interfaces for type hinting
```
adrianschmidt: Shouldn't these interfaces be exported further? Or does that happen automatically?

adrianschmidt: Also, is there any way we could make use of them in the examples?

jgroth: They should not be exposed to consumers if that's what you mean. They are only internal and used by the form component. If we one day find another library for rendering forms from a schema and decide to use that instead, we will be in big trouble if these are exposed :smile:

adrianschmidt: Ah, ok. So these don't describe the schema format that our form-component takes, only the data it passes on to the lib we're using?

jgroth: Yeah, and the data we get from it as well :slightly_smiling_face:

jgroth: Ideally, these would have been defined in the library itself, but they are only described in the docs as far as I can see

adrianschmidt: 👍

src/examples/form/nested-form.tsx
```diff
@@ -0,0 +1,78 @@
...
+                    title: 'Country',
+                    description: 'It might also be a planet',
+                },
+            },
+        },
+        data: {
+            type: 'object',
+            collapsible: true,
+            title: 'Data',
+            description: 'Some additional data we would like you to submit',
```
adrianschmidt: I added this line as well, so there's a description in a collapsible-section in the example.

src/components/form/templates/object-field.ts
adrianschmidt: Approved, but if you feel we should discuss this, I guess it's probably best to do that before merging… 🙂

adrianschmidt: AFK discussion: We'll put our own properties in a dict called `lime`, effectively prefixing them, like this: `props.schema.lime.collapsible`

This is the fastest way to secure the code against name conflicts with potential future standard properties.

PR #579, opened 2020-01-28
feat(form): add component for array items

jgroth:
fix Lundalogik/crm-feature#962

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/collapsible-section/collapsible-section.tsx
+++ b/src/components/collapsible-section/collapsible-section.tsx
@@ -1,5 +1,6 @@
 import { Component, Event, EventEmitter, h, Prop } from '@stencil/core';
 import { dispatchResizeEvent } from '../../util/dispatch-resize-event';
+import { Action } from './action';
 
 /**
  * @slot - Content to put inside the collapsible section
@@ -21,6 +22,12 @@ export class CollapsibleSection {
     public header: string;
 
     /**
+     * Actions to place to the far right inside the header
+     */
+    @Prop()
+    public actions: Action[];
+
+    /**
      * Emitted when the section is expanded
      */
     @Event()
@@ -32,8 +39,15 @@ export class CollapsibleSection {
     @Event()
     private close: EventEmitter<void>;
 
+    /**
+     * Emitted when an action is clicked inside the header
+     */
+    @Event()
+    private action: EventEmitter<Action>;
+
     constructor() {
         this.onClick = this.onClick.bind(this);
+        this.renderActionButton = this.renderActionButton.bind(this);
     }
 
     public render() {
@@ -54,9 +68,10 @@ export class CollapsibleSection {
                             })()}
                         </svg>
                     </span>
-                    <h5 class="mdc-typography mdc-typography--headline5">
+                    <h2 class="mdc-typography mdc-typography--headline2">
                         {this.header}
-                    </h5>
+                    </h2>
+                    {this.renderActions()}
                 </header>
                 <div class="body">
                     <slot />
@@ -76,4 +91,32 @@ export class CollapsibleSection {
             this.close.emit();
         }
     }
+
+    private renderActions() {
+        if (!this.actions) {
+            return;
+        }
+
+        return (
+            <div class="actions">
+                {this.actions.map(this.renderActionButton)}
+            </div>
+        );
+    }
+
+    private renderActionButton(action: Action) {
+        return (
+            <limel-icon-button
+                icon={action.icon}
+                label={action.label}
+                disabled={action.disabled}
+                onClick={this.handleActionClick(action)}
+            />
+        );
+    }
+
+    private handleActionClick = (action: Action) => (event: MouseEvent) => {
+        event.stopPropagation();
+        this.action.emit(action);
+    };
 }
--- a/src/components/form/templates/array-field-simple-item.ts
+++ b/src/components/form/templates/array-field-simple-item.ts
@@ -1,0 +1,79 @@
+import React from 'react';
+import { ArrayFieldItem } from './types';
+
+export class SimpleItemTemplate extends React.Component {
+    private props: { item: ArrayFieldItem; index: number };
+    private refs: { removeButton: any; moveUpButton: any; moveDownButton: any };
+
+    private removeHandler: (event: any) => void;
+    private moveUpHandler: (event: any) => void;
+    private moveDownHandler: (event: any) => void;
+
+    public componentDidMount() {
+        const { item, index } = this.props;
+        const removeButton: HTMLLimelButtonElement = this.refs.removeButton;
+        this.removeHandler = item.onDropIndexClick(index);
+        removeButton.addEventListener('click', this.removeHandler);
+
+        const upButton: HTMLLimelButtonElement = this.refs.moveUpButton;
+        this.moveUpHandler = item.onReorderClick(index, index - 1);
+        upButton.addEventListener('click', this.moveUpHandler);
+
+        const downButton: HTMLLimelButtonElement = this.refs.moveDownButton;
+        this.moveDownHandler = item.onReorderClick(index, index + 1);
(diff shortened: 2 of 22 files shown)


src/components/collapsible-section/collapsible-section.tsx
```diff
@@ -20,6 +21,9 @@ export class CollapsibleSection {
     @Prop()
     public header: string;
 
+    @Prop()
+    public actions: Action[];
```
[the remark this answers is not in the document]
jgroth: Yeah sorry, I forgot about that :slightly_smiling_face: 
Also, we should be able to use the same interface in other components. I think the input field or some other component has the same concept of an action as a trailing icon if I remember correctly. I just put the interface here for now and then we can move it later if we want to use it in other places

src/components/form/templates/types.ts
```diff
@@ -66,3 +66,7 @@ export interface ArrayFieldItem {
     onReorderClick: (index: number, newIndex: number) => (event: any) => void;
     readonly: boolean;
 }
+
+export interface Runnable {
+    run: (event: any) => void;
```
adrianschmidt: Any particular reason why actions in general wouldn't benefit from a `run` property?

src/components/form/templates/array-field-simple-item.ts
```diff
@@ -0,0 +1,79 @@
...
+        );
+    }
+
+    private renderRemoveButton(item: ArrayFieldItem) {
+        // tslint:disable-next-line:no-duplicate-string
+        return React.createElement('limel-button', {
+            icon: 'trash',
+            disabled: !item.hasRemove,
+            ref: 'removeButton',
+        });
```
adrianschmidt: I tried using `limel-icon-button` instead, and I like the result much better:

image

We should preferably supply labels for the buttons as well though (they won't be visible when using `limel-icon-button`), as those will be used by assistive tech, and we're going to have to move towards assistive tech compatibility anyway. But I guess that would mean either adding translation functionality to lime-elements, or requiring translated strings be input by the consuming client… So, I guess that'll be a matter for a future update regardless… 😕

src/components/form/templates/array-field.ts
```diff
@@ -1 +1,63 @@
...
+    }
+
+    private renderAddButton() {
+        if (!this.props.canAdd) {
+            return;
+        }
+
+        return React.createElement('limel-button', {
+            outlined: true,
+            label: 'Add item',
```
adrianschmidt: This string should definitely be translated though, and I think it should be provided by the consumer.

jgroth: Yeah, I didn't know what to do about it right now but we need to be able to translate it somehow :+1:

src/components/form/templates/common.ts
```diff
@@ -0,0 +1,94 @@
...
+ * Find the first string item in a nested structure.
+ *
+ * If an object has the key 'title' it will have priority, followed by the key 'name'
+ *
+ * @param {*} data the data to find the title for
+ * @param {*} schema schema for the item
+ *
+ * @returns {string} describing title of the data
+ */
+export function findTitle(data: any, schema: any) {
```
adrianschmidt: This should be reasonably easy to write some tests for, right? If so, I think we should.

Don't we already have schema data structures somewhere, that we could use? In lime-admin perhaps? I could swear I've seen similar tests very recently 🙂

adrianschmidt: Well, I guess those were probably tests for the lime-admin schema, so maybe not exactly cut-and-paste to here, but testing this should be very similar I assume?

src/components/form/templates/array-field-simple-item.ts
jgroth: Yeah I tried them both and just went with one of them for now as I didn't want to decide this by myself :smile: Let's have a discussion during the meeting this afternoon. 
I'm not sure I like that these are bigger, but they have less margin to the left/right which I think is good :slightly_smiling_face: 
Also, there might be other icons we can use. It also require icons8 at the moment, which is not shipped with lime elements which might be an issue

src/components/form/templates/common.ts
jgroth: Haha yeah we have more or less the same code in lime-admin, I made som improvements to it here. But we should definitely have tests for it :+1:

src/components/collapsible-section/collapsible-section.tsx
```diff
@@ -76,4 +85,31 @@ export class CollapsibleSection {
...
+        return (
+            <div class="actions">
+                {this.actions.map(this.renderActionButton)}
+            </div>
+        );
+    }
+
+    private renderActionButton(action: Action) {
+        return (
+            <limel-button
```
adrianschmidt: In the screenshot using `lime-icon-button`, I changed this as well of course.

src/components/form/templates/array-field-simple-item.ts
adrianschmidt: I didn't immediately like the bigger icons either. They are easier to see, but the are a bit big in comparison to the text. Although, I think we should maybe use slightly bigger text in the headers for collapsible-section…

That the regular button can't have a label for assistive tech without also visibly showing that label suggests to me that the icon-button is the right component to use for buttons with no text.

Regarding icons8, I agree that it's not ideal. I had a question from Fredrik the other day, if there was any way for an add-on to supply its own icon to a limel-button, and there isn't at the moment, but it would be really nice if, apart from a name of an icon, the consumer could also supply either a url for where to load the icon from, or the image-data itself. For a url, we could simply look at the first character of the `icon` string, and if it's a slash, we load the image from that url, instead of the normal way. I'm pretty sure image data has some known prefix as well, which we could detect. That way, we would only need to make changes to limel-icon, not to all the components using it. And then we could embed some icons that we are allowed to use, like I did with the chevron for collapsible-section.

src/components/form/templates/types.ts
jgroth: I thought about it, but it didn't seem like the right solution. There is only one listener for the action event here, but there might be several listeners in other cases and they might want to do different things. Here I added the function to the action object mostly because it seemed like the easiest solution, instead of mapping the action id to a different function inside the handler. 

I hope that made sense :smile:

adrianschmidt: Yeah, adding the function rather than mapping the id was really nice! I did the mapping thing when I temporarily modified the example for collapsible-section in order to try it, and mapping felt a lot clunkier.

src/style/variables.scss
```diff
@@ -39,8 +39,8 @@ $mdc-theme-text-secondary-on-background: rgba(0, 0, 0, 0.54);
 @import "@limetech/mdc-theme/mdc-theme";
 
 $mdc-typography-styles-headline1: (
-    font-size: pxToRem(16),
-    line-height: pxToRem(14)
+    font-size: pxToRem(22),
+    line-height: pxToRem(22)
```
jgroth: At first when I was chaning the size in the header of the collapsible section I thought it was kind of strange that material had pretty much the same sizes for all the headlines, and they looked quite similar.

Now I found these :smile: I find it a bit strange though that we have set all of them to pretty much the same size as they start to loose their meaning if they are all the same. I guess we should go through these at some point and give them a proper size so we can use them

jgroth: Oh, and I took the liberty of changing one of them now that we did not seem to use anywhere apart from in the new form component

adrianschmidt: Haha! Yeah, I have no idea why they are all the same 😂 I guess maybe it was done in some effort to have the new components from lime-elements fit in with the old stuff from the webclient back in the day… 🤔

PR #581, opened 2020-01-29
feat(collapsible-section): background-color and padding can now be set using CSS variables

adrianschmidt:
re Lundalogik/crm-feature#968

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/collapsible-section/collapsible-section.scss
+++ b/src/components/collapsible-section/collapsible-section.scss
@@ -1,5 +1,12 @@
 @import '../../style/variables';
 
+/**
+ * @prop --closed-header-background-color: background color for header when closed
+ * @prop --open-header-background-color: background color for header when open
+ * @prop --body-background-color: background color for body
+ * @prop --body-padding: padding for body
+ */
+
 :host {
     display: block;
 }
@@ -13,11 +20,9 @@
 }
 
 section {
-    background-color: #ffffff;
-
     > header {
         align-items: center;
-        background-color: #ffffff;
+        background-color: var(--closed-header-background-color, #ffffff);
         cursor: pointer;
         display: flex;
         padding-left: pxToRem(10);
@@ -26,7 +31,7 @@ section {
         height: pxToRem(60);
 
         > .expander {
-            color: #a6a6a6;
+            color: rgba(0,0,0,0.33);
             height: pxToRem(30);
             margin-right: pxToRem(5);
             width: pxToRem(30);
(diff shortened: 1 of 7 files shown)


src/components/collapsible-section/collapsible-section.scss
```diff
@@ -34,21 +39,33 @@ section {
...
+
+        &:before,
+        &:after {
+            content: " ";
+            display: table;
+        }
+
+        &:after {
+            clear: both;
+        }
```
adrianschmidt: Clearfix baby!

src/components/collapsible-section/collapsible-section.scss
```diff
@@ -13,17 +20,15 @@
 }
 
 section {
-    background-color: #ffffff;
```
adrianschmidt: This was stupid. Clearfix is the right way to do it. Bad past-Adrian!

adrianschmidt: To clarify, I had to remove this, because with semitransparent background for the header and body, this effectively "reset" the "base-background", resulting in nested collapsible-sections not appearing as expected, but like this instead:

image

src/components/collapsible-section/collapsible-section.scss
adrianschmidt: To clarify, after removing the section background colour on line 16, clearfixing was needed to ensure that the `.body` element contains any margins on descendant elements, or this would happen:

image

src/components/collapsible-section/collapsible-section.scss
```diff
@@ -1,5 +1,12 @@
 @import '../../style/variables';
 
+/**
+ * @prop --limel-collapsible-section-header-closed-background-color: background color for header when closed
+ * @prop --limel-collapsible-section-header-open-background-color: background color for header when open
+ * @prop --limel-collapsible-section-body-background-color: background color for body
+ * @prop --limel-collapsible-section-body-padding: padding for body
```
jgroth: These are quite long and it would be nice if we could make them a bit shorter
We have been very inconsistent with the naming of our css variables :smile: The only other ones that have a prefix are the ones in `limel-grid` and they are only prefixed with `lime` _not_ `limel`

Would the following be too generic?
```
--closed-header-background-color
--open-header-background-color
--body-background-color
--body-padding
```

src/components/collapsible-section/collapsible-section.scss
jgroth: I'm not that familiar with clearfix, but I thought it was used to clear floats :thinking: I don't think we are using that here, right? Is it used for something else as well?

src/components/collapsible-section/collapsible-section.scss
adrianschmidt: Oh, yeah… we don't actually need to prefix them, because we keep them local to the component element anyway (or possibly the parent of it, but same-same). I forgot about that. Yeah, your suggestion is much nicer!

src/components/collapsible-section/collapsible-section.scss
adrianschmidt: Yeah, it clears the margins of the `<p>` element. Those are what jut out and creates the space between the body and the header in the screenshot above. They also jut out at the bottom of the body, but that's less apparent when looking at this screenshot. Compare to the screenshot on line 16 though, and you'll see that the bottom margin has been cleared there (and is thus contained within the body, instead of hanging outside it).

PR #589, opened 2020-02-06
fix(text-field): show error icon

Befkadu1:
### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -127,6 +127,9 @@ export class InputField {
     @State()
     private isFocused: boolean = false;
 
+    @State()
+    private isModified: boolean = false;
+
     @Element()
     private limelInputField: HTMLElement;
 
@@ -164,7 +167,6 @@ export class InputField {
 
     public render() {
         const additionalProps = this.getAdditionalProps();
-
         return [
             <label
                 class={`
@@ -173,7 +175,7 @@ export class InputField {
                     ${this.disabled ? 'mdc-text-field--disabled' : ''}
                     ${this.required ? 'mdc-text-field--required' : ''}
                     ${
-                        this.trailingIcon
+                        this.getIcon()
                             ? 'mdc-text-field--with-trailing-icon'
                             : ''
                     }
@@ -245,6 +247,7 @@ export class InputField {
 
     private onBlur() {
         this.isFocused = false;
+        this.isModified = true;
     }
 
     private renderHelperLine() {
@@ -267,8 +270,20 @@ export class InputField {
         if (this.helperText === null || this.helperText === undefined) {
             return;
         }
-
-        return <p class="mdc-text-field-helper-text">{this.helperText}</p>;
+        return (
+            <p
+                class={`
+            mdc-text-field-helper-text
+            ${
+                this.isInvalid()
+                    ? 'mdc-text-field-helper-text--validation-msg'
+                    : ''
+            }
+        `}
+            >
+                {this.helperText}
+            </p>
+        );
     }
 
     private renderCharacterCounter() {
@@ -282,8 +297,31 @@ export class InputField {
         return <div class="mdc-text-field-character-counter">{label}</div>;
     }
 
+    private getIcon() {
+        if (this.isInvalid()) {
+            return 'high_importance';
+        }
+
+        return this.trailingIcon;
+    }
+
+    private isInvalid() {
+        if (this.invalid) {
+            return true;
+        }
+
+        if (!this.isModified) {
+            return false;
+        }
+
+        const element = this.limelInputField.shadowRoot.querySelector('input');
+
+        return !(element && element.checkValidity());
+    }
+
     private renderTrailingIcon() {
-        if (!this.trailingIcon) {
+        const icon = this.getIcon();
+        if (!icon) {
             return;
         }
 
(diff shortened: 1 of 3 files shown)


src/components/input-field/input-field.tsx
```diff
@@ -68,6 +68,12 @@ export class InputField {
     @Prop({ reflectToAttr: true })
     public trailingIcon: string;
 
+    /**
+     * Trailing icon to show to the far right in the field
+     */
+    @Prop({ reflectToAttr: true })
+    public trailingErrorIcon: string;
```
jgroth: I don't think we need the ability to customize the error icon, the default one should be enough for now. If there is demand for a custom icon we can add it later.

src/components/input-field/input-field.tsx
```diff
@@ -204,6 +214,7 @@ export class InputField {
                     {this.label}
                 </span>
                 {this.renderTrailingIcon()}
+                {this.renderTrailingErrorIcon()}
```
jgroth: It should not be possible to have both the regular icon and the error icon. If there is an error, the error icon should replace the regular icon

src/components/input-field/input-field.tsx
```diff
@@ -268,7 +279,24 @@ export class InputField {
...
+            }
+            
+        `}
+            >
+                {this.helperText}
+            </p>
+        );
+    }
+
+    onValidateMinValueLength() {
```
jgroth: It should not be the responsibility of this component to perform validation since it can never know when value is valid, thats why we have the `invalid` property. It's true that we can have some validation base on the `minlength`, `maxlength`, `min`, `max` and `required` properties, but that is handled by underlying `input` tag and can be checked with the either the `:invalid` CSS property (https://developer.mozilla.org/en-US/docs/Web/CSS/:invalid) or the Constraint Validation API (https://developer.mozilla.org/en-US/docs/Web/API/HTMLObjectElement/checkValidity)

The only other thing we can do to decide if the field is valid or not is to check the `invalid` property that is sent into the component. `invalid` should always take precedence

src/components/input-field/input-field.tsx
```diff
@@ -300,6 +328,35 @@ export class InputField {
...
+                onKeyPress={this.handleIconKeyPress}
+                onClick={this.handleIconClick}
+                class={`
+                    ${
+                        this.isFocused
+                            ? 'mdc-text-field-icon-color'
+                            : 'mdc-text-field__icon'
+                    }
+                
+                `}
```
jgroth: It's possible to write this in a nicer way using objects instead of string literals :slightly_smiling_face: 
```jsx
<i class={{
    'this-class-is-applied': true,
    'this-class-is-not-applied': false
}} />
```

src/enums/icon.ts
```diff
@@ -0,0 +1,4 @@
+export enum IconType {
```
jgroth: I like that you think about stuff like this :+1: 
However, in this particular case I'm not sure how useful this file will be. We do not have a lot of icon names that are hard coded. The `EMAIL` one in here is not used so it should be removed, which just leaves the `IMPORTANCE` one left.
What do you think about using a constant inside `input-field.tsx` instead?

src/examples/input-field/input-field-error-icon.tsx
```diff
@@ -0,0 +1,35 @@
...
+    constructor() {
+        this.onChange = this.onChange.bind(this);
+    }
+
+    public render() {
+        return (
+            <limel-input-field
+                label="Text Field"
+                minlength={MIN_LENGTH}
+                helperText={this.onHelperText()}
```
jgroth: The function needs to be either renamed or removed. `on` indicates that this is an event or callback which is not the case, and since the text never changes we can just insert the text here instead
```
helperText="..."
```

src/examples/input-field/input-field-error-icon.tsx
```diff
@@ -0,0 +1,35 @@
...
+            />
+        );
+    }
+
+    private onChange(event) {
+        this.value = event.detail;
+    }
+
+    private onHelperText() {
+        return 'Min 6 characters!';
```
jgroth: What do you think about "Please enter at least 6 characters"?

jgroth:
We also forgot to add the `mdc-text-field-helper-text--persistent` class to the helper text when the field is invalid  :slightly_smiling_face:

src/enums/icon.ts
Befkadu1: In this case, you are right that we don't have to use the enum.  It is good to use enums for the future I guess.

src/components/input-field/input-field.scss
```diff
@@ -84,3 +84,10 @@ $mdc-text-field-fullwidth-bottom-line-color: $lime-text-field-bottom-line-color;
     width: pxToRem(24);
     height: pxToRem(24);
 }
+
+label.mdc-text-field.mdc-text-field--invalid + .mdc-text-field-helper-line {
+    .mdc-text-field-helper-text {
+        color: $mdc-theme-error;  
+        opacity: 1;
```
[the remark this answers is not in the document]
Befkadu1: if we use `mdc-text-field-helper-text--validation-msg`  class, we don't need to use opacity.  We can even remove the css 
`label.mdc-text-field.mdc-text-field--invalid + .mdc-text-field-helper-line {
    .mdc-text-field-helper-text {
        color: $mdc-theme-error;  
        opacity: 1;
   }
}
`

PR #599, opened 2020-02-17
Improve collapsible section

hannahu:
fix Lundalogik/lime-admin#84

This PR includes fixes for:
- Expand collapsible section when adding a new collapsible field to reveal its content (also updated default title to _New item_ instead of the array index)
- How to choose a section title:
    1. Take field with title or name key
    2. Take required field
    3. Take first occurrence of string field with value

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/templates/array-field-collapsible-item.ts
+++ b/src/components/form/templates/array-field-collapsible-item.ts
@@ -2,6 +2,7 @@ import { Action } from '@limetech/lime-elements';
 import React from 'react';
 import { findTitle } from './common';
 import { ArrayFieldItem, Runnable } from './types';
+import { isEmpty } from 'lodash-es';
 
 interface CollapsibleItemProps {
     /**
@@ -32,10 +33,12 @@ interface CollapsibleItemProps {
 
 export class CollapsibleItemTemplate extends React.Component {
     public refs: { section: any };
+    private isOpen: boolean;
 
     constructor(public props: CollapsibleItemProps) {
         super(props);
         this.handleAction = this.handleAction.bind(this);
+        this.isOpen = isEmpty(props.data);
     }
 
     public componentDidMount() {
@@ -56,13 +59,15 @@ export class CollapsibleItemTemplate extends React.Component {
     }
 
     public render() {
-        const { data, schema, index, formSchema } = this.props;
+        const { data, schema, formSchema } = this.props;
+
         return React.createElement(
             'limel-collapsible-section',
             {
-                header: findTitle(data, schema, formSchema) || index,
+                header: findTitle(data, schema, formSchema) || 'New item',
                 className: 'limel-form-array-item--object',
                 ref: 'section',
+                'is-open': this.isOpen,
             },
             this.props.item.children
         );
--- a/src/components/form/templates/common.ts
+++ b/src/components/form/templates/common.ts
@@ -1,4 +1,4 @@
-import { get } from 'lodash-es';
+import { get, isEmpty } from 'lodash-es';
 import React from 'react';
 import { isArrayType, isObjectType } from '../schema';
 
@@ -27,9 +27,11 @@ export function renderDescription(description: string) {
 }
 
 /**
- * Find the first string item in a nested structure.
+ * Find a suitable title for a nested structure.
  *
- * If an object has the key 'title' it will have priority, followed by the key 'name'
+ * If an object has the key 'title' it will have priority, followed by the key 'name'.
+ * If 'title' nor 'name' is found, a required item will be considered.
+ * Otherwise the first occurrence of a nonempty string is chosen.
  *
  * @param {*} data the data to find the title for
  * @param {*} fieldSchema schema for the item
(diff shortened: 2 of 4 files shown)


adrianschmidt:
@hannahu If this shouldn't close Lundalogik/lime-admin#84, it shouldn't use the keyword `fix`. You can use `re` instead.

src/components/form/templates/common.ts
```diff
@@ -50,28 +50,41 @@ export function findTitle(data: any, fieldSchema: any, formSchema: any) {
...
+    return newData;
 }
 
 function findFirstEntry(data: any) {
+    const firstKey = Object.keys(data)[0];
     const entries = [
         ['title', data.title],
         ['name', data.name],
-        ...Object.entries(data),
+        [firstKey, data[firstKey]],
```
jgroth: Will this work if `data[firstKey]` is falsy?

src/components/form/templates/array-field-collapsible-item.ts
```diff
@@ -56,14 +59,20 @@ export class CollapsibleItemTemplate extends React.Component {
     }
 
     public render() {
-        const { data, schema, index, formSchema } = this.props;
+        const { data, schema, formSchema } = this.props;
+        const inputs = {
```
jgroth: We might want to name this `props`, since that is what it is called in both React and Stencil

src/components/form/templates/array-field-collapsible-item.ts
```diff
@@ -56,14 +59,20 @@ export class CollapsibleItemTemplate extends React.Component {
...
+            header: findTitle(data, schema, formSchema) || 'New item',
+            className: 'limel-form-array-item--object',
+            ref: 'section',
+        };
+
+        if (this.isOpen) {
+            inputs['is-open'] = true;
```
jgroth: We can add this to the object above to get rid of this `if`-statement
```
{
    'is-open': this.isOpen
}
```

src/components/form/templates/common.ts
```diff
@@ -50,28 +50,41 @@ export function findTitle(data: any, fieldSchema: any, formSchema: any) {
         return findSchemaTitle(data, fieldSchema);
     }
 
+    const subSchema = findSubSchema(fieldSchema, formSchema);
+
+    data = sortdDataByProperties(data, subSchema.properties);
```
jgroth:
```suggestion
    data = sortDataByProperties(data, subSchema.properties);
```

src/components/form/templates/common.ts
```diff
@@ -50,28 +50,41 @@ export function findTitle(data: any, fieldSchema: any, formSchema: any) {
...
         return null;
     }
 
     const [key, value] = firstEntry;
-    const subSchema = findSubSchema(fieldSchema, key, formSchema);
+    return findTitle(value, subSchema.properties[key], formSchema);
+}
 
-    return findTitle(value, subSchema, formSchema);
+function sortdDataByProperties(data: any, properties: object) {
```
jgroth:
```suggestion
function sortDataByProperties(data: any, properties: object) {
```

hannahu:
> @hannahu ~If this shouldn't close [Lundalogik/lime-admin#84](https://github.com/Lundalogik/lime-admin/issues/84), it shouldn't use the keyword `fix`. You can use `re` instead.~
> 
> Actually, I might be wrong about that. Johan thought that it doesn't autoclose the parent issue if the issue has more PR:s connected to it.

Good input, no it won't close the issue if there are more PRs but I wanted to try your `re` suggestions since I'll merge the other PR connected. I couldn't get your suggestion to work, the PR disappears from the issue?

src/components/form/templates/common.ts
hannahu: ddddddddddd

adrianschmidt:
> I couldn't get your suggestion to work, the PR disappears from the issue?

Yeah, I noticed that too. So keep the `fix` 🙂

hannahu:
> > I couldn't get your suggestion to work, the PR disappears from the issue?
> 
> Yeah, I noticed that too. So keep the `fix`

Yes but I don't want the issue to close when this PR closes but then I've to reopen it I guess :D

src/components/form/templates/array-field-collapsible-item.ts
hannahu: I couldn't get that to work before but will try again,  maybe I changed something on the way

hannahu: Ah it's the same now, but still doesn't work. Will try something else.

adrianschmidt:
> Yes but I don't want the issue to close when this PR closes but then I've to reopen it I guess :D

Or open a new connected PR before this one is merged and closed, that's possible?

src/components/form/templates/common.ts
```diff
@@ -50,28 +52,53 @@ export function findTitle(data: any, fieldSchema: any, formSchema: any) {
...
     const entries = [
         ['title', data.title],
         ['name', data.name],
         ...Object.entries(data),
     ];
 
+    if (subSchema.required) {
+        const requiredEntry = getRequiredEntry(data, subSchema);
+        const requiredEntryIndex = 2;
+        entries[requiredEntryIndex] = requiredEntry;
```
jgroth: This will overwrite the first property from data :scream: 
What do you think about adding a new method called `getRequiredEntries` and write the code above like this?
```typescript
const entries = [
    ['title', data.title],
    ['name', data.name],
    ...getRequiredEntries(),
    ...Object.entries(data),
    ];
```

src/components/form/templates/common.ts
```diff
@@ -79,22 +79,24 @@ function findFirstEntry(data: any, subSchema: any) {
     const entries = [
         ['title', data.title],
         ['name', data.name],
+        getRequiredEntry(data, subSchema) || [null, null],
```
jgroth: I think it would be a bit cleaner if `getRequiredEntry` returned `[null, null]`

PR #602, opened 2020-02-19
fix(input-field-number): blurred onWheel event on input field

terryilebode6:
fix Lundalogik/crm-feature#974
### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -193,6 +193,7 @@ export class InputField {
                     required={this.required}
                     disabled={this.disabled}
                     type={this.type}
+                    onWheel={this.handleWheel}
                     {...additionalProps}
                     value={this.value}
                 />
@@ -477,6 +478,11 @@ export class InputField {
 
         if (isSpace || isEnter) {
             this.action.emit();
+        }
+    }
+    private handleWheel(event: MouseEvent) {
+        if (this.type === 'number') {
+            event.preventDefault();
         }
     }
 }
--- a/src/examples/dialog/dialog-form.tsx
+++ b/src/examples/dialog/dialog-form.tsx
@@ -66,6 +66,7 @@ export class DialogFormExample {
                             required={true}
                             invalid={!this.ageValid()}
                             onChange={this.ageOnChange}
+                            type="number"
                         />
                     </p>
                     <p>


src/examples/dialog/dialog-form.tsx
```diff
@@ -25,6 +25,16 @@ export class DialogFormExample {
     @State()
     private isConfirmationOpen = false;
 
+    @State()
+    private formatNumber = true;
```
jgroth: This does not need to be a `State()` variable, or even a variable at all since we never change it

src/examples/dialog/dialog-form.tsx
```diff
@@ -25,6 +25,16 @@ export class DialogFormExample {
...
+
+    @State()
+    private value;
```
jgroth: Same here, this is never changed so we don't need to have it as a variable

src/examples/input-field/input-field-number.tsx
```diff
@@ -85,4 +91,7 @@ export class InputFieldNumberExample {
     private toggleRequired() {
         this.required = !this.required;
     }
+    private blurAway(){
+        this.el.blur()
```
jgroth: This only fixes the issue in this example, if I were to use the number input field in my own component, the problem would still be there. 
Also, this has the side effect of removing focus from the field when scrolling which is not a good user experience

src/examples/dialog/dialog-form.tsx
```diff
@@ -34,10 +44,21 @@ export class DialogFormExample {
         this.closeConfirmation = this.closeConfirmation.bind(this);
         this.onConfirmPositive = this.onConfirmPositive.bind(this);
         this.onConfirmNegative = this.onConfirmNegative.bind(this);
+        
+        this.scrollStop=this.scrollStop.bind(this);
     }
 
     public render() {
         return [
+            <limel-input-field
```
jgroth: We can probably remove this field
We should refactor the existing example instead and change the "age" input field in the dialog to `type="number"`

src/examples/dialog/dialog-form.tsx
```diff
@@ -34,10 +44,21 @@ export class DialogFormExample {
...
+                label="Number Field Label"
+                value={this.value}
+                type="number"
+                formatNumber={this.formatNumber}
+
+                onWheel={() => { this.el.blur() }}
```
jgroth: `this.el` here will be the entire example component, not just the input field, which means we are trying to blur the whole component. Is that what we want? :thinking:

src/examples/dialog/dialog-form.tsx
```diff
@@ -149,4 +170,10 @@ export class DialogFormExample {
         this.isOpen = true;
         this.isConfirmationOpen = false;
     }
+   
+    
+    private scrollStop(e:Event){
```
jgroth: This function is never called

src/examples/input-field/input-field-number.tsx
```diff
@@ -21,6 +21,9 @@ export class InputFieldNumberExample {
     @State()
     private value;
 
+    @Element()
+    private el: HTMLElement;
```
jgroth: This does not seem to be used so we can remove it :slightly_smiling_face:

src/components/input-field/input-field.tsx
```diff
@@ -479,4 +480,9 @@ export class InputField {
             this.action.emit();
         }
     }
+    private handleWheel(event) {
```
jgroth: We should add a type to the event
```suggestion
    private handleWheel(event: MouseEvent) {
```

src/components/input-field/input-field.tsx
```diff
@@ -479,4 +480,9 @@ export class InputField {
...
+        if (this.type === 'number') {
+            event.target.blur();
```
jgroth: I think `preventDefault` would be better since it does not have the side effect of removing focus from the field
```suggestion
            event.preventDefault();
```

adrianschmidt:
Nice! First PR merged to a production repo I think? Great to see, and welcome for real @terryilebode6! 🍻🙂

PR #604, opened 2020-02-20
fix(limel-picker): render picker dropdown in document.body

BregenzerK:
fix Lundalogik/crm-feature#971

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu-surface/menu-surface.scss
+++ b/src/components/menu-surface/menu-surface.scss
@@ -1,7 +1,9 @@
 @import '../../style/variables';
-@import "@limetech/mdc-menu-surface/mdc-menu-surface";
+@import '@limetech/mdc-menu-surface/mdc-menu-surface';
+@import '@limetech/mdc-elevation/mdc-elevation';
 
 .mdc-menu-surface {
     width: 100%;
     max-height: 100%;
 }
+
--- a/src/components/menu-surface/menu-surface.tsx
+++ b/src/components/menu-surface/menu-surface.tsx
@@ -61,14 +61,15 @@ export class MenuSurface {
     }
 
     public render() {
+        const classList = {
+            'mdc-menu-surface': true,
+            'mdc-menu-surface--open': this.open,
+            'mdc-elevation-transition': true,
+            'mdc-elevation--z4': true,
+        };
+
         return (
-            <div
-                class={{
-                    'mdc-menu-surface': true,
-                    'mdc-menu-surface--open': this.open,
-                }}
-                tabindex="-1"
-            >
+            <div class={classList} tabindex="-1">
                 <slot />
             </div>
         );
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -21,6 +21,7 @@ import {
     TAB,
     TAB_KEY_CODE,
 } from '../../util/keycodes';
+import { createRandomString } from '../../util/random-string';
 
 const SEARCH_DEBOUNCE = 500;
 const CHIP_SET_TAG_NAME = 'limel-chip-set';
@@ -130,6 +131,7 @@ export class Picker {
 
     private debouncedSearch;
     private chipSet: HTMLLimelChipSetElement;
+    private portalId: string;
 
     constructor() {
         this.handleTextInput = this.handleTextInput.bind(this);
@@ -140,15 +142,14 @@ export class Picker {
         this.handleInteract = this.handleInteract.bind(this);
         this.handleListChange = this.handleListChange.bind(this);
         this.handleStopEditAndBlur = this.handleStopEditAndBlur.bind(this);
+        this.handleSurfaceDismissed = this.handleSurfaceDismissed.bind(this);
+
+        this.portalId = createRandomString();
     }
 
     @Watch('value')
-    public onChangeValue(newValue, oldValue) {
+    public onChangeValue(newValue = [], oldValue = []) {
         this.chips = this.createChips(this.value);
-        if (!this.multiple) {
-            return;
-        }
-
         if (newValue.length <= oldValue.length) {
             return;
         }
@@ -161,13 +162,8 @@ export class Picker {
             this.searcher,
             SEARCH_DEBOUNCE
         );
-        this.element.addEventListener('blur', this.handleStopEditAndBlur);
         this.chipSet = this.element.shadowRoot.querySelector(CHIP_SET_TAG_NAME);
         this.chips = this.createChips(this.value);
-    }
-
-    public componentDidUnload() {
-        this.element.removeEventListener('blur', this.handleStopEditAndBlur);
     }
 
     public async componentWillUpdate() {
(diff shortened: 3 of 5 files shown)


src/components/menu-surface/menu-surface.tsx
```diff
@@ -30,6 +30,9 @@ export class MenuSurface {
     @Prop()
     public open = false;
 
+    @Prop()
+    public cssClasses: string[] = [];
```
jgroth: I'm not sure about this :thinking: It makes it a bit unclear for the consumer what we actually can do to change the style of the menu. Maybe we should have one or more properties with explicit names that can be set, and set the css classes internally inside the menu instead. What do you think?

jgroth: Or if it can be solved with css properties it would be even better I guess :slightly_smiling_face:, but that might not be possible here

src/components/menu-surface/menu-surface.scss
```diff
@@ -1,7 +1,30 @@
...
+    &:not(.display-full-list) {
+        max-height: pxToRem(250);
+        overflow-y: auto;
+    }
+}
+
+.empty-result-message {
+    color: var(--lime-light-grey, #{$lime-light-grey});
+    text-align: center;
+}
```
jgroth: Should these classes be part of the menu surface component? It feels like they are part of the picker or the spinner. Would it be possible to refactor these so they belong to the component that uses them? I hope that made sense :slightly_smiling_face:

src/components/picker/picker.tsx
```diff
@@ -140,15 +142,14 @@ export class Picker {
...
+        this.handleListClosed = this.handleListClosed.bind(this);
+
+        this.portalId = createRandomString();
     }
 
     @Watch('value')
-    public onChangeValue(newValue, oldValue) {
+    public onChangeValue(newValue = [], oldValue = []) {
         this.chips = this.createChips(this.value);
-        if (!this.multiple) {
```
jgroth: Do we need to remove this or am I missing something? :slightly_smiling_face:  
If `multiple` is `false`, neither `newValue` or `oldValue` will be arrays and will not have the `.length` property

src/components/picker/picker.tsx
```diff
@@ -252,94 +248,70 @@ export class Picker {
...
+        );
+
+        cssList.push('dropdown--list');
+        if (this.displayFullList || this.items.length <= ITEM_LIMIT_NO_SCROLL) {
+            cssList.push('display-full-list');
+        }
+        return this.renderPortal(true, limelList, cssList);
+    }
 
+    private renderPortal(isVisible, content = null, cssList: string[] = []) {
```
jgroth: I don't think we need the `isVisible` argument, it seems we always call this in one of two ways, either
```typescript
this.renderPortal(false)
```
or
```typescript
this.renderPortal(true, content)
```
so we should be able to just check if the `content` argument is set or not.
Also, `this.renderPortal(false)` makes it sound like we are not rendering a portal, so it's a bit confusing :slightly_smiling_face:

src/components/picker/picker.tsx
```diff
@@ -252,94 +248,70 @@ export class Picker {
...
-                    onChange={this.handleListChange}
-                    type="selectable"
-                    items={this.items}
-                />
-            </div>
+            <limel-portal visible={isVisible} containerId={this.portalId}>
+                <limel-menu-surface
+                    open={isVisible}
+                    cssClasses={cssList}
+                    onDismiss={this.handleListClosed}
```
jgroth: What do you think about naming this `handleMenuClosed`?

src/components/menu-surface/menu-surface.tsx
BregenzerK: agree, I didn't like the solution too much either, but I also didn't x properties for x different styling combinations. I'll look into doing it as css variables but otherwise we could have one property as object with boolean like this 
```
{dropdown-list: boolean;
list-full: boolean}
```
probably different names but I guess you get the idea

src/components/menu-surface/menu-surface.scss
BregenzerK: good catch, that might actually not be needed anymore because I'm setting it as inline css on the paragraph (in the picker) 

the problem is that html wise it's not part anymore of the picker, that's why it can't be in the css file for limel-picker

src/components/picker/picker.tsx
BregenzerK: the if statement made sure, that the chipset wouldn't set focus when selecting a value one a picker that only accepts one value. that makes sense, but with the dropdown list rendered in the body it gives you worse keyboard support, because the tab position will stay on limel-list and if you then try to tab to the next field you end up on the url field of your browser (in chrome at least). Do you have a better idea how we can change the tab position? I tried to set the tabindex on the chipset but that didn't work

src/components/picker/picker.tsx
BregenzerK: not sure. it makes sense as a listener on the menu-surface but as a method name in limel-picker it's rather confusing because the picker doesn't really operate with a menu.

src/components/menu-surface/menu-surface.tsx
jgroth: Does it have to be styled inside the menu surface? Would it be possible to have the styling on the components that are affected? :thinking:

jgroth: We can always add styling as a `style` attribute if don't find a better solution

BregenzerK: I should be able to wrap the content that I render inside the menu surface with a div and set the css classes on that. that shouldn't need any changes in the surface component.

src/components/picker/picker.tsx
jgroth: Maybe `handleSurfaceClosed`? `handleSurfaceDismissed`?

src/components/menu-surface/menu-surface.tsx
BregenzerK: shit, it doesn't work that way. it has to be inside a shadow dom if we don't want to use purely inline style. I'll look into a combination of concreter props and inline styling

jgroth: If we want to send in styling it might make more sense to do it on the portal than on the menu surface, since that is the component that moves its children to somewhere else in the DOM :thinking: 
That means we could do something like this

```tsx
<limel-portal styling={{}}>
    <my-component />
</limel-portal>
```
And inside the portal
```tsx
render() {
    return [
        <style>{this.styling}</style>,
        <slot />
    ];
}
```

What do you think? Does it make sense?

BregenzerK: I think that makes to our own styles on the dropdown list. but not too much for using material's elevation classes. My current idea is to have a prop on the surface like this
```
    /**
     * use material design's elevation styling
     */
    @Prop()
    public elevationDesign: boolean;
```
and use the styling you suggested for the rest

src/components/menu-surface/menu-surface.tsx
```diff
@@ -30,6 +30,12 @@ export class MenuSurface {
     @Prop()
     public open = false;
 
+    /**
+     * use material design's elevation styling
+     */
+    @Prop()
+    public elevationDesign: boolean;
```
jgroth: I will remove this, I don't think we should expose material internals and we should be able to just set the elevation all the time. It's probably just an inconsistency that it was not set for the `select` component

src/components/picker/picker.tsx
```diff
@@ -252,94 +248,77 @@ export class Picker {
      */
     private renderDropdown() {
```
jgroth: This method has a bit too much responsibility so I'm breaking it up since we are poking around in it anyway :slightly_smiling_face:

src/components/picker/picker.tsx
```diff
@@ -252,94 +248,77 @@ export class Picker {
...
+                onChange={this.handleListChange}
+                type="selectable"
+                items={this.items}
+            />
+        );
+
+        if (!this.displayFullList && this.items.length > ITEM_LIMIT_NO_SCROLL) {
+            const styling = {
+                'max-height': '250px',
+                'overflow-y': 'auto',
```
jgroth: The `overflow-y` property made the shadows around the dropdown disappear so I removed it. The scrolling should be handled by the list anyway so the property should not be needed here

Issue #571, opened 2020-01-27
The auto-publish of documentation fails because the aws cli-tool is not installed in the Docker container


adrianschmidt:
There are probably a few possible ways to do this.

- The `publish-docs.js` script is using the `aws` cli-command, which should be the `aws-cli` python package.
- There is also an aws node-module, but that doesn't seem to have the `sync` command on the `s3` module, so some changes would probably be necessary in order to use that.
- It might be possible to put the aws stuff in a separate Docker container, and use that from our main container. There's probably even already a container available that can be used.

jensgustafsson:
I suggest that we refactor the script `publish-docs.js` in order to solve this issue.

Perhaps we could rename the script to `prepare-publish-docs.js` and remove the actual `publish` command (the `aws s3 sync` command).

If so, we could solve this issue by using two different docker images and creating a command (for instance in the makefile) for the publish step that would look something like this:

```Makefile

.PHONY: publish-docs
publish-docs: prepare-docs
    ## Mount the `pwd`/output folder and publish the content in the folder
    docker run docker run -v `pwd`/output:/output -e AWS_ACCESS_KEY_ID="<id>" -e AWS_SECRET_ACCESS_KEY="<key>" -e AWS_DEFAULT_REGION="<region>" --rm mesosphere/aws-cli s3 sync . s3://lime-documentation-lime-elements --and-extra-config-stuff-if-needed

.PHONY: prepare-docs
prepare-docs:
    ##Do all the node stuff and output the docs bundle to /output in the container which
    ## is mapped to `pwd`/output on the host machine
    docker run -v `pwd`/output:/output --rm <node-image> prepare-publish-docs prepare
```

And then we can publish the docs simply by running:

```bash
make publish-docs
```

adrianschmidt:
That sounds like a very good solution! Thanks @jensgustafsson!

jensgustafsson:
I have put some thinking in to this and have an idea of a potential solution.

If we treat the "the publish docs" step in semantic release as an upload of the compiled docs bundle to the release in github (as an asset), then we could add a "post" build step that always fetches the "docs release" from github and syncs it to s3. What do you think about that @adrianschmidt ?

adrianschmidt:
> I have put some thinking in to this and have an idea of a potential solution.
> 
[...]

PR #607, opened 2020-02-26
ci(semantic-release): disable publishing of docs to aws, because it doesn't work anyway

adrianschmidt:
This is a temporary fix. Everything should work just as it has so far, but with release-builds
passing instead of failing at the very last command in the last step. Docs still need to be
published to aws manually, but they should be built and pushed to github by the release-build.

re #571

### DIFF
--- a/release.config.js
+++ b/release.config.js
@@ -11,7 +11,7 @@ module.exports = {
             '@semantic-release/exec',
             {
                 publishCmd:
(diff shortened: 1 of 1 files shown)


adrianschmidt:
Not really anything to test here, except keeping an eye on the next release-build on master. So I'm merging this.

PR #615, opened 2020-03-02
feat(config): add feature switches to config object

jgroth:
Connected to Lundalogik/crm-feature#983

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/global/config.ts
+++ b/src/global/config.ts
@@ -1,5 +1,6 @@
 export class Config {
     public iconPath = '';
+    public featureSwitches: any = {};
 }
 
 const config = new Config();
--- a/src/components/config/config.tsx
+++ b/src/components/config/config.tsx
@@ -9,10 +9,18 @@ export class Config {
     @Prop()
     public config: object;
 
+    public componentDidLoad() {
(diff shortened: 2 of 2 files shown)


src/global/config.ts
```diff
@@ -1,5 +1,6 @@
 export class Config {
     public iconPath = '';
+    public featureSwitches = {};
```
adrianschmidt: I think this has to be typed as `any`, or the compiler will complain about properties not existing on type `Object`…

jgroth: Oh, yeah! :smile: 
Good catch!

adrianschmidt: Confirmed in testing, so I ⚡️'d it 😄

jgroth: I approve! :heavy_check_mark: :smile:

Issue #87, opened 2018-09-13
Add a tabs-component


adrianschmidt:
## New feature motivation

We're using tabs in the webclient, and it's not unlikely that plugins might want to use them as well. They're also a somewhat complicated component, since the handling of tabs that overflow the width of the page can be done in several ways, all of which are complicated, except for the alternative to just wrap the overflowing tabs to a new line, but that is butt-ugly.

## New feature description

Add a tabs-component :)

## New feature implementation

The tabs-component in lime-webclient is a bit extra complicated, because it's been built such that the row of tabs and the "tab content" sections do not have to be attached. That is, you can choose to place the row of tabs in one place on the page, and the "pages" somewhere else entirely (unless it's been changed or my memory lets me down). IMHO, that's probably unnecessary, and lower complexity should be a higher priority.

PR #616, opened 2020-03-02
Tab bar 

Befkadu1:
Connected to Lundalogik/lime-elements#87

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tab-bar/examples/tab-bar.tsx
+++ b/src/components/tab-bar/examples/tab-bar.tsx
@@ -1,0 +1,79 @@
+import { Component, h, State } from '@stencil/core';
+import { Tab } from '../tab.types';
+
+@Component({
+    tag: 'limel-example-tab-bar',
+    shadow: true,
+})
+export class TabBarExample {
+    @State()
+    private text: string = 'Joker';
+
+    @State()
+    private tabs: Tab[] = [
+        {
+            id: 1,
+            text: 'Joker',
+            icon: 'joker',
+            active: true,
+            iconColor: 'var(--lime-green)',
+        },
+        {
+            id: 2,
+            text: 'Parasite',
+            icon: 'insect',
+            iconColor: 'var(--lime-magenta)',
+        },
+        {
+            id: 3,
+            text: 'Harriet',
+            icon: 'administrator_female',
+            iconColor: 'var(--lime-orange)',
+        },
+        {
+            id: 4,
+            text: 'Bombshell',
+            icon: 'surprised',
+            iconColor: 'var(--lime-blue)',
+        },
+        {
+            id: 5,
+            text: 'Judy',
+            icon: 'female',
+            iconColor: 'var(--lime-deep-red)',
+        },
+        {
+            id: 6,
+            text: 'Jojo Rabbit',
+            icon: 'rabbit',
+            iconColor: 'var(--lime-yellow)',
+        },
+        {
+            id: 7,
+            text: 'Little Women',
+            icon: 'female',
+            iconColor: 'var(--lime-deep-red)',
+        },
+    ];
+
+    constructor() {
+        this.onChange = this.onChange.bind(this);
+    }
+
+    public render() {
+        return [
+            <limel-tab-bar tabs={this.tabs} onChangeTab={this.onChange} />,
+            <p>Tab: {this.text}</p>,
+        ];
+    }
+
+    private onChange(event: CustomEvent<Tab>) {
+        this.text = event.detail.text;
+        this.tabs = this.tabs.map(tab => {
+            if (tab.id === event.detail.id) {
+                return event.detail;
+            }
+            return tab;
+        });
+    }
+}
--- a/src/components/tab-bar/tab-bar.scss
+++ b/src/components/tab-bar/tab-bar.scss
@@ -1,0 +1,12 @@
+@import '../../style/variables';
+@import '@limetech/mdc-tab-bar/mdc-tab-bar';
+@import '@limetech/mdc-tab-scroller/mdc-tab-scroller';
+@import '@limetech/mdc-tab-indicator/mdc-tab-indicator';
+@import '@limetech/mdc-tab/mdc-tab';
+:host {
+  display: block;
+}
+
+.mdc-tab-bar__icon {
+  margin-right: .025rem;
+}
--- a/src/components/tab-bar/tab-bar.tsx
+++ b/src/components/tab-bar/tab-bar.tsx
@@ -1,0 +1,154 @@
+import {
+    Component,
+    h,
+    Prop,
+    Element,
+    EventEmitter,
+    Event,
+} from '@stencil/core';
+import { MDCTabBar, MDCTabBarActivatedEvent } from '@limetech/mdc-tab-bar';
+import { strings } from '@limetech/mdc-tab-bar/constants';
+import { Tab } from './tab.types';
+import config from '../../global/config';
+
+const { TAB_ACTIVATED_EVENT } = strings;
+
+@Component({
+    tag: 'limel-tab-bar',
+    styleUrl: 'tab-bar.scss',
+    shadow: true,
+})
+export class TabBar {
+    @Element()
+    private element: HTMLElement;
+
+    /**
+     * List of tabs to display
+     */
+    @Prop()
+    public tabs: Tab[];
+
+    private mdcTabBar: MDCTabBar;
+
+    @Event()
+    private changeTab: EventEmitter<Tab>;
+
+    constructor() {
+        this.handleTabActivated = this.handleTabActivated.bind(this);
+        this.renderTab = this.renderTab.bind(this);
+    }
+
+    public componentDidLoad() {
+        this.setup();
+    }
+
+    private setup() {
+        const element = this.element.shadowRoot.querySelector('.mdc-tab-bar');
+        if (!element) {
+            return;
+        }
+
+        this.mdcTabBar = new MDCTabBar(element);
+
+        // Workaround for shadow dom support for material
+        (this
+            .mdcTabBar as any).foundation_.adapter_.getFocusedTabIndex = () => {
+            const tabElements = this.getTabElements();
+            const activeElement = this.element.shadowRoot.activeElement;
+            return tabElements.indexOf(activeElement);
+        };
+
+        this.setupListeners();
+    }
+
+    private getTabElements() {
+        return [].slice.call(
+            this.element.shadowRoot.querySelectorAll('.mdc-tab')
+        );
+    }
+
+    private setupListeners() {
+        this.mdcTabBar.listen(TAB_ACTIVATED_EVENT, this.handleTabActivated);
+    }
+
+    private handleTabActivated(event: MDCTabBarActivatedEvent) {
+        const index = event.detail.index;
+        const oldSelectedTab = this.tabs.find(tab => tab.active === true);
+
+        if (oldSelectedTab) {
+            this.changeTab.emit({ ...oldSelectedTab, active: false });
+        }
+
+        this.changeTab.emit({ ...this.tabs[index], active: true });
+    }
+
+    public componentDidUnload() {
+        this.mdcTabBar.unlisten(TAB_ACTIVATED_EVENT, this.handleTabActivated);
+        this.mdcTabBar.destroy();
+    }
+
+    render() {
+        const featFlag = config.featureSwitches.enableTabs;
+        if (!featFlag) {
+            return;
+        }
+
+        return (
+            <div class="mdc-tab-bar" role="tablist">
+                <div class="mdc-tab-scroller">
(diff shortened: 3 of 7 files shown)


src/components/tab-bar/tab.types.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * True if the tab should be selected.
+     */
+    selected?: boolean;
+
+    /**
+     * A text shown when a tab is selected.
+     */
+    content?: string;
```
jgroth: This should not be part of the actual tab, but instead be provided by a `<slot>` element

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,139 @@
...
+    private renderTabs() {
+        return (
+            this.tabs.map((tab, index) => {
+                return (
+                    <button class={`
+                        mdc-tab
+                        ${tab.selected ? 'mdc-tab--active' : ''}
+                    `}
+                        role="tab"
+                        aria-selected="true" tabindex={index}
```
adrianschmidt: I know the code isn't done yet, but I'll comment now or I'll forget to check later:

This isn't right. `aria-selected` should only be true if this is the selected tab, and `tabindex` should be set to `-1` for all but the selected tab. On the selected tab, no tabindex should be set, if I read the docs correctly.

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+    }
+
+    public componentDidLoad() {
+        this.element.focus();
+        this.setup();
+    }
+
+    private setup() {
+        this.element.focus();
```
jgroth: I don't think we should always set focus on the element when it's created. Instead, we should call `this.mdcTabBar.activateTab()` if one of the given tabs are active

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,139 @@
...
+                class="mdc-text-field__icon"
+                name={tab.icon}
+                badge={true}
+                style={style}
+                size="small"
+            />
+        )
+    }
+
+    private renderTabs() {
```
jgroth: What do you think about renaming this to `renderTab` and only have it render a single tab? It will removing some unnecessary nesting in this method and make it easier to read

The render method above would then only have to do this instead
```
{this.tabs.map(this.renderTab)}
```

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,139 @@
...
+                    <div class="mdc-tab-scroller__scroll-area mdc-tab-scroller__scroll-area--scroll">
+                        <div class="mdc-tab-scroller__scroll-content">
+                            {this.renderTabs()}
+                        </div>
+                    </div>
+                </div>
+
+                <div class="content">
+                    <slot>{this.selectedTab && this.selectedTab.content}</slot>
+                </div>
```
jgroth: I think we might want to consider having a separate component that handles the interaction between slotable content and the active tab in a tab bar.

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+        this.setupListeners();
+    }
+
+    private getTabElements() {
+        return [].slice.call(this.element.shadowRoot.querySelectorAll('.mdc-tab'));
+    }
+
+    private setupListeners() {
+        this.mdcTabBar.listen(TAB_ACTIVATED_EVENT, this.handleAction);
```
jgroth: `handleAction` is a little vague here since there is no concept of actions in this context. What about `handleTabActivated`?

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,139 @@
...
+import { strings } from '@limetech/mdc-tab-bar/constants';
+import { Tab } from './tab.types';
+const { TAB_ACTIVATED_EVENT } = strings;
+
+@Component({
+    tag: 'limel-tab-bar',
+    styleUrl: 'tab-bar.scss',
+    shadow: true
+})
+export class TabBar {
```
jgroth: Since this will probably be worked on for a while during the cycle, we can add a feature flag that will prevent it from being used until we are ready to release it :slightly_smiling_face:

src/components/tab-bar/tab-bar.e2e.ts
```diff
@@ -0,0 +1,11 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('limel-tab-bar', () => {
```
jgroth: We should remove this unless we write specific tests for it

src/components/tab-bar/tab.types.ts
```diff
@@ -0,0 +1,26 @@
...
+
+    /**
+     * Name of the icon to use.
+     */
+    icon?: string;
+
+    /**
+     * True if the tab should be selected.
+     */
+    selected?: boolean;
```
jgroth: What do you think about `active` instead of `selected`? It's the term material uses and we should probably try to minimize confusion by having another name for it when they are so similar.
We should also clarify that only one tab at a time can be active

src/components/tab-bar/tab.types.ts
```diff
@@ -0,0 +1,26 @@
+export interface Tab {
+    /**
+     * Name of the tab.
+     */
+    name: string;
```
jgroth: We might want to consider renaming this to `text` to be consistent with our other interfaces with similar functionality. Types like `ListItem`, `Chip` and `Option` all have the `text` property

src/components/tab-bar/tab.types.ts
```diff
@@ -0,0 +1,26 @@
+export interface Tab {
```
jgroth: It might be a good idea to also have an `id` property so consumers have a way of identifying their tabs

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+    private getTabElements() {
+        return [].slice.call(this.element.shadowRoot.querySelectorAll('.mdc-tab'));
+    }
+
+    private setupListeners() {
+        this.mdcTabBar.listen(TAB_ACTIVATED_EVENT, this.handleAction);
+    }
+
+    private handleAction(event: MDCTabBarActivatedEvent) {
```
jgroth: We should let consumers know that another tab has been selected by emitting an event from this component

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,139 @@
...
+    private renderIcon(tab: Tab) {
+        const style = {};
+        if (tab.iconColor) {
+            style['background-color'] = tab.iconColor;
+            style['color'] = 'white';
+        }
+
+        return (
+            <limel-icon
+                class="mdc-text-field__icon"
```
jgroth: Is this copy/paste? :slightly_smiling_face: 
Maybe `mdc-tab-bar__icon`?

src/components/tab-bar/tab-bar.tsx
Befkadu1: This code was for testing. The PR is pushed only for you guys to check if you guys workaround for shadow dom support

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,142 @@
...
+        );
+    }
+
+    private renderTab(tab: Tab) {
+        return (
+            <button
+                class={`
+                mdc-tab
+                ${tab.active ? 'mdc-tab--active' : ''}
+            `}
```
jgroth: I think inline string literals for attributes makes the code a bit messy hand hard to read. I suggest we declare them as an object at the top of the function instead, e.g.
```tsx
private renderTab(tab: Tab) {
    const classList = {
        'mdc-tab': true,
        'mdc-tab--active': !!tab.active
    };

    return (
        <button class={classList}>...</button>
    );
}
```

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,142 @@
...
+            >
+                <span class="mdc-tab__content">
+                    {this.renderIcon(tab)}
+                    <span class="mdc-tab__text-label">{tab.text}</span>
+                </span>
+                <span
+                    class={`
+                        mdc-tab-indicator
+                        ${tab.active ? 'mdc-tab-indicator--active' : ''}
+                    `}
```
jgroth: Same here regarding inline string literals for class attribute

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+    private setupListeners() {
+        this.mdcTabBar.listen(TAB_ACTIVATED_EVENT, this.handleTabActivated);
+    }
+
+    private handleTabActivated(event: MDCTabBarActivatedEvent) {
+        const index = event.detail.index;
+        const selectedTab = this.tabs[index];
+        this.selectedTab = selectedTab;
+        this.changeTab.emit({ ...this.selectedTab, active: true });
```
jgroth: The list component emits two events when selection changes, one for the old item telling the consumer that the item is no longer selected and one for the new selected item. We should probably do the same here :thinking:

src/components/tab-bar/tab-bar.scss
```diff
@@ -0,0 +1,16 @@
...
+  display: block;
+}
+
+.mdc-tab-bar__icon {
+  margin-right: 1rem;
+}
+
+.content {
+  margin-top: 2rem;
+}
```
jgroth: This can be removed now :slightly_smiling_face:

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,155 @@
...
+    @Prop()
+    public tabs: Tab[];
+
+    private mdcTabBar: MDCTabBar;
+
+    @Event()
+    private changeTab: EventEmitter<Tab>;
+
+    @State()
+    private selectedTab = this.tabs[0];
```
jgroth: I'm not sure we need this variable. We should be able to get the old selected tab in `handleTabActivated` by looping through the given `this.tabs` and finding the one with `active === true`. Having two variables for the same thing makes it harder to keep them in sync

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    private getTabElements() {
+        return [].slice.call(
+            this.element.shadowRoot.querySelectorAll('.mdc-tab')
+        );
+    }
+
+    private setupListeners() {
+        this.mdcTabBar.listen(TAB_ACTIVATED_EVENT, this.handleTabActivated);
```
jgroth: We need to unlisten for the events when this component is destroyed

src/components/tab-bar/tab-bar.tsx
```diff
@@ -0,0 +1,155 @@
...
+        const style = { color: '' };
+        if (tab.iconColor) {
+            style.color = tab.iconColor;
+        }
+
+        return (
+            <limel-icon
+                class="mdc-tab-bar__icon"
+                name={tab.icon}
+                badge={true}
```
jgroth: Should this still be `true`? We don't have the background for the icon any more

src/components/tab-bar/examples/tab-bar.tsx
```diff
@@ -0,0 +1,58 @@
...
+    },
+];
+
+@Component({
+    tag: 'limel-example-tab-bar',
+    shadow: true,
+})
+export class TabBarExample {
+    public render() {
+        return <limel-tab-bar tabs={tabs} />;
```
jgroth: I think we can improve this example a bit by using the event in some way as well. Just having a small text label below the tabs displaying the name of the current tab should be enough :slightly_smiling_face:

jgroth:
We should clean up the commit history as well before merging. I'm not sure why there is so much code in the first commit by greenkeeper 63822f5 :thinking:

adrianschmidt:
> We should clean up the commit history as well before merging. I'm not sure why there is so much code in the first commit by greenkeeper [63822f5](https://github.com/Lundalogik/lime-elements/commit/63822f5658362f8aa28a9a31b4e4eb017c73ce41) 🤔

I'm guessing stuff was accidentally squashed into an earlier commit, and the stuff changed by the greenkeeper PR doesn't show up in the diff here, since it has already been merged to master.

That commit should simply be renamed.

Since it's a new feature, I think it should all go into a single commit. All except the one called "feat(config): initialize config with feature switches from localStorage".

PR #620, opened 2020-03-04
chore(package): allow new commit type `switched` for non-released functionality

adrianschmidt:


package.json
```diff
@@ -63,7 +63,7 @@
     "@types/react": "16.9.24",
     "commitizen": "4.0.3",
     "cross-env": "7.0.0",
-    "cz-conventional-changelog": "3.1.0",
+    "cz-conventional-changelog": "git+https://github.com/Lundalogik/cz-conventional-changelog.git",
```
adrianschmidt: I've opened a PR to the parent repo here: commitizen/cz-conventional-changelog#114

I guess we'll see if they accept it. I think we might need to write tests to get it accepted, but I don't really have time to do that right now.

PR #625, opened 2020-03-05
feat(badge): add badge component

Befkadu1:
Fix Lundalogik/crm-feature#978

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/badge/badge.tsx
```diff
@@ -0,0 +1,54 @@
...
+    shadow: true,
+})
+export class Badge {
+    /**
+     * Shows numbers in the badge
+     */
+    @Prop()
+    public badgeCounter: number;
+
+    showRoundedNumber() {
```
jgroth: Did you look into using a library for this? I found this and it looks like it does what we want :slightly_smiling_face: 
https://www.npmjs.com/package/number-abbreviate

src/components/badge/badge.tsx
```diff
@@ -0,0 +1,54 @@
...
+            badgeCounter = badgeCounter.slice(0, numbersToRemove);
+        }
+        return `${badgeCounter}${getParam.postfix}`;
+    }
+
+    filter(digitLength: number) {
+        let devisor;
+        let postfix;
+        if (digitLength > FIVE_DIGIT_LENGTH) {
+            devisor = ONE_MILION;
```
hannahu: Just some misspelling. "divisor" and "ONE_MILLION"

src/components/badge/badge.scss
```diff
@@ -0,0 +1,17 @@
...
+
+.badge-container {
+    display: inline-block;
+    font-size: 0.7rem;
+    min-width: 1.2rem;
+    height: 1.2rem;
+    border-radius: 1.2rem;
+    line-height: 1.2rem;
+    text-align: center;
+    background-color: #EEF1F5;
```
jgroth: It would be nice if this was configurable. We should use a CSS variable for it :slightly_smiling_face: 
We might need one for the color of text as well

src/components/badge/badge.tsx
```diff
@@ -0,0 +1,54 @@
...
+    tag: 'limel-badge',
+    styleUrl: 'badge.scss',
+    shadow: true,
+})
+export class Badge {
+    /**
+     * Shows numbers in the badge
+     */
+    @Prop()
+    public badgeCounter: number;
```
jgroth: I think we should rename this. `number`? `label`?

src/components/tab-bar/tab.types.ts
```diff
@@ -23,4 +23,9 @@ export interface Tab {
      *  Color of the icon.
      */
     iconColor?: string;
+
+    /**
+     * Counts the number of badge
+     */
+    badgeCounter?: number;
```
jgroth: What do you think of just `badge`?

src/components/badge/badge.tsx
adrianschmidt: Nice lib! I think we should override their default units though, because `m` is "milli" not "mega"… (`k`, `M`, `G`, and `T` is definitely less language dependant than `k`, `m`, `b`, and `t`)

Although, to be honest, I wouldn't be surprised if someone from sales or something forces us to use `t` for "tusen" 😩

We'll probably have to enable translations for the units eventually.

src/components/badge/badge.tsx
Befkadu1: I removed this code.

src/components/badge/badge.tsx
Befkadu1: Changed to label

src/components/tab-bar/tab-bar.tsx
```diff
@@ -143,6 +143,7 @@ export class TabBar {
                 <span class="mdc-tab__content">
                     {this.renderIcon(tab)}
                     <span class="mdc-tab__text-label">{tab.text}</span>
+                    <limel-badge label={tab.badge} />
```
jgroth: This needs to be conditional. If there is no badge for the tab, this should not be rendered

src/components/badge/badge.scss
```diff
@@ -0,0 +1,25 @@
...
+    font-size: 0.7rem;
+    min-width: 1.2rem;
+    height: 1.2rem;
+    border-radius: 1.2rem;
+    line-height: 1.2rem;
+    text-align: center;
+    background-color:  var(--badge-background-color, #EEF1F5);
+    color:  var(--badge-text-color, #7D899C);
+    padding: 0 0.3rem;
+    margin-left: 0.25rem;
```
adrianschmidt, line 24: Is there any specific reason why none of the size values use the `pxToRem` function, and why almost all of them are set at values that result in fractions of pixels at the default size of 16px/rem? @Befkadu1

adrianschmidt: Tweaking this in #678

PR #627, opened 2020-03-05
Tab panel

jgroth:
fix Lundalogik/crm-feature#988 
fix Lundalogik/crm-feature#979

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -154,7 +154,7 @@ export class Picker {
             return;
         }
 
-        this.chipSet.setFocus(true);
+        this.chipSet?.setFocus(true);
     }
 
     public componentDidLoad() {
--- a/src/components/tab-bar/examples/tab-bar.tsx
+++ b/src/components/tab-bar/examples/tab-bar.tsx
@@ -1,5 +1,5 @@
 import { Component, h, State } from '@stencil/core';
-import { Tab } from '../tab.types';
+import { Tab } from '@limetech/lime-elements';
 
 @Component({
     tag: 'limel-example-tab-bar',
--- a/src/components/tab-panel/examples/tab-panel-content.scss
+++ b/src/components/tab-panel/examples/tab-panel-content.scss
@@ -1,0 +1,19 @@
+.text {
+    display: flex;
+    align-items: center;
+    margin-bottom: 1rem;
+}
+
+.container {
+    display: flex;
+    align-items: center;
+    flex-direction: column;
+}
+
+p {
+    margin: 0;
+}
+
+limel-icon {
+    margin-right: 1rem;
+}
--- a/src/components/tab-panel/examples/tab-panel-content.tsx
+++ b/src/components/tab-panel/examples/tab-panel-content.tsx
@@ -1,0 +1,95 @@
+import {
+    Component,
+    h,
+    Prop,
+    EventEmitter,
+    Event,
+    State,
+    Watch,
+} from '@stencil/core';
+import { Tab } from '@limetech/lime-elements';
+
+const LOAD_TIME = 1000;
+
+@Component({
+    tag: 'limel-example-tab-panel-content',
+    shadow: true,
+    styleUrl: 'tab-panel-content.scss',
+})
+export class TabPanelContentExample {
+    @Prop()
+    public tab: Tab;
+
+    @Event()
+    private changeTab: EventEmitter<Tab>;
+
+    @State()
+    private votes = 0;
+
+    @State()
+    private loaded = false;
+
+    constructor() {
+        this.vote = this.vote.bind(this);
+    }
+
+    public render() {
+        if (!this.loaded) {
+            return <limel-spinner />;
+        }
+
+        const style = {
+            backgroundColor: this.tab.iconColor,
+            color: 'white',
+        };
+
+        return (
+            <div class="container">
+                <div class="text">
+                    <limel-icon
+                        badge={true}
+                        size="large"
+                        name={this.tab.icon}
+                        style={style}
+                    />
+                    <p>
+                        {this.tab.text} has received {this.votes} votes!
+                    </p>
+                </div>
+                <limel-button
+                    outlined={true}
+                    icon="star"
+                    label="Vote"
+                    onClick={this.vote}
+                />
+            </div>
+        );
+    }
+
+    /**
+     * Load the tab contents the first time the tab is activated
+     */
+    @Watch('tab')
+    protected watchTab() {
+        if (!this.loaded && this.tab.active) {
+            this.loadTabContent();
+        }
+    }
+
+    /**
+     * Simulate loading the tab content
+     */
+    private loadTabContent() {
+        setTimeout(() => {
+            this.loaded = true;
+        }, LOAD_TIME);
+    }
+
+    private vote() {
+        this.votes++;
+        this.changeTab.emit({
+            ...this.tab,
+            badge: this.votes,
+        });
+    }
+}
--- a/src/components/tab-panel/examples/tab-panel.tsx
+++ b/src/components/tab-panel/examples/tab-panel.tsx
@@ -1,0 +1,59 @@
+import { Component, h, State } from '@stencil/core';
+import { Tab } from '@limetech/lime-elements';
+
+@Component({
+    tag: 'limel-example-tab-panel',
+    shadow: true,
+    styleUrl: 'tab-panel.scss',
+})
+export class TabPanelExample {
+    @State()
+    private tabs: Tab[] = [
+        {
+            id: 'joker',
+            text: 'Joker',
+            icon: 'joker',
+            active: true,
+            iconColor: 'var(--lime-green)',
+        },
+        {
+            id: 'parasite',
+            text: 'Parasite',
+            icon: 'insect',
+            iconColor: 'var(--lime-magenta)',
+        },
+        {
+            id: 'harriet',
+            text: 'Harriet',
+            icon: 'administrator_female',
+            iconColor: 'var(--lime-orange)',
+        },
+    ];
+
+    constructor() {
+        this.handleChangeTab = this.handleChangeTab.bind(this);
+    }
+
+    public render() {
+        return [
+            <limel-tab-panel
+                tabs={this.tabs}
+                onChangeTab={this.handleChangeTab}
+            >
+                <limel-example-tab-panel-content id="joker" />
+                <limel-example-tab-panel-content id="parasite" />
+                <limel-example-tab-panel-content id="harriet" />
+            </limel-tab-panel>,
+        ];
+    }
+
+    private handleChangeTab(event: CustomEvent<Tab>) {
+        this.tabs = this.tabs.map(tab => {
+            if (tab.id === event.detail.id) {
+                return event.detail;
+            }
+
+            return tab;
+        });
+    }
+}
--- a/src/components/tab-panel/tab-panel.mdx
+++ b/src/components/tab-panel/tab-panel.mdx
@@ -1,0 +1,36 @@
+---
+name: Tab panel
+route: /tab-panel
+menu: Components
+---
+
+# Tab panel 
+
+### 🚧 This feature is work in progress. It can be tested out by setting the feature switch `enableTabs` 🚧
+
+The `tab-panel` component uses the `tab-bar` component together with custom slotted components and will display the 
+content for the currently active tab. Each slotted component must have an id equal to the id of the corresponding tab
+it belongs to.
+
+The `tab-panel` component will automatically set each tab configuration on the corresponding slotted component as a
+property named `tab` so that the component can take action upon that. Sometimes it might be desirable to not load data
+or render anything until the tab is active. 
+
+The slotted components can also emit the `changeTab` event to update anything inside the actual tab, e.g. to change
+the icon, color or badge.
+
+<limel-props name="limel-tab-panel" />
+
+## Example
+
+This example illustrates how to add custom components inside the `tab-panel`. Each component will simulate loading the
+data it needs once the tab has been activated and then display the actual content. If the button is pressed, the 
+component will emit the `changeTab` event to change the badge inside the corresponding tab.
+
+<limel-example name="limel-example-tab-panel" />
+
+### Custom component
+
+This is the component that is used for each tab in the example above.
+
+<limel-example name="limel-example-tab-panel-content" path="tab-panel" code-only={true} />
--- a/src/components/tab-panel/tab-panel.tsx
+++ b/src/components/tab-panel/tab-panel.tsx
@@ -1,0 +1,93 @@
+import {
+    Component,
+    Element,
+    h,
+    Prop,
+    EventEmitter,
+    Event,
+    Host,
+} from '@stencil/core';
+import { Tab } from '../tab-bar/tab.types';
+
+/**
+ * @slot - Content to put inside the `tab-panel`. Each slotted element must have the `id` attribute equal to the id of the tab it belongs to.
+ */
+@Component({
+    tag: 'limel-tab-panel',
+    shadow: true,
+})
+export class TabPanel {
+    @Prop({ mutable: true })
+    public tabs: Tab[] = [];
+
+    @Element()
+    private host: HTMLElement;
+
+    @Event()
+    protected changeTab: EventEmitter<Tab>;
+
+    private slotElements: HTMLElement[];
+
+    constructor() {
+        this.handleChangeTabs = this.handleChangeTabs.bind(this);
+        this.setSlotElements = this.setSlotElements.bind(this);
+        this.setTabStatus = this.setTabStatus.bind(this);
+    }
+
+    protected componentDidLoad() {
+        const slot = this.getSlot();
+        slot.addEventListener('slotchange', this.setSlotElements);
+        this.setSlotElements();
+        this.tabs.forEach(this.setTabStatus);
+    }
+
+    protected componentDidUnload() {
+        const slot = this.getSlot();
+        slot.removeEventListener('slotchange', this.setSlotElements);
+    }
+
+    public render() {
+        return (
+            <Host onChangeTab={this.handleChangeTabs}>
+                <div class="tab-panel">
+                    <limel-tab-bar tabs={this.tabs} />
+                    <div class="tab-content">
+                        <slot />
+                    </div>
+                </div>
+            </Host>
+        );
+    }
+
+    private setSlotElements() {
+        const slot = this.getSlot();
+        this.slotElements = [].slice.call(slot.assignedElements());
+    }
+
+    private setTabStatus(tab: Tab) {
+        const element = this.slotElements.find(e => e.id === tab.id);
+        if (tab.active) {
+            element.style.display = '';
+        } else {
+            element.style.display = 'none';
+        }
+
+        element['tab'] = tab; // tslint:disable-line:no-string-literal
+    }
+
+    private handleChangeTabs(event: CustomEvent<Tab>) {
+        this.tabs = this.tabs.map(tab => {
+            if (tab.id === event.detail.id) {
+                return event.detail;
+            }
+
+            return tab;
+        });
+
+        this.setTabStatus(event.detail);
+    }
+
+    private getSlot(): HTMLSlotElement {
+        return this.host.shadowRoot.querySelector('slot');
+    }
+}
--- a/src/components/tab-panel/examples/tab-panel.scss
+++ b/src/components/tab-panel/examples/tab-panel.scss
@@ -1,0 +1,7 @@
+limel-example-tab-panel-content {
+    display: flex;
+    height: 10rem;
+    background-color: #eee;
+    justify-content: center;
+    align-items: center;
+}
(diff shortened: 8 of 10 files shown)


src/components/tab-panel/tab-panel.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+    private setTabStatus(tab: Tab) {
+        const element = this.slotElements.find(e => e.id === tab.id);
+        if (tab.active) {
+            element.style.display = '';
+        } else {
+            element.style.display = 'none';
+        }
+
+        element['tab'] = tab; // tslint:disable-line:no-string-literal
```
jgroth: It feels quite nice that this is done automatically, but I'm not 100% sure. If it's not done here, the user has to send in the tab manually for each component

src/components/tab-panel/examples/tab-panel-content.tsx
```diff
@@ -0,0 +1,101 @@
...
+    h,
+    Prop,
+    EventEmitter,
+    Event,
+    State,
+    Watch,
+} from '@stencil/core';
+import { Tab } from '@limetech/lime-elements';
+
+const LOAD_TIME = 1500;
```
Befkadu1: LOAD_TIME = 1500 isn't it too much to wait? It is just example 500ms might be ok. Of course this is a personal preference, I leave this for you.

src/components/tab-panel/examples/tab-panel.tsx
```diff
@@ -0,0 +1,59 @@
...
+        this.handleChangeTab = this.handleChangeTab.bind(this);
+    }
+
+    public render() {
+        return [
+            <limel-tab-panel
+                tabs={this.tabs}
+                onChangeTab={this.handleChangeTab}
+            >
+                <limel-example-tab-panel-content id="joker" />
```
Befkadu1: What do you think we loop through the tabs array and the id of this tag`limel-example-tab-panel-content` becomes dynamic.

src/components/tab-panel/examples/tab-panel-content.tsx
```diff
@@ -0,0 +1,101 @@
...
+        this.votes++;
+        const text = this.getMovieTitle();
+        this.changeTab.emit({
+            ...this.tab,
+            text: [text, this.votes].join(' '),
+        });
+    }
+
+    private getMovieTitle() {
+        return this.tab.text.split(' ').shift();
```
Befkadu1: I don't a point to have this `split(' ').shift()`. What if you have a text `Jojo Rabbit`? It will cut off the `Rabit` part. I suggest to retun `this.tab.text`

src/components/picker/picker.tsx
```diff
@@ -154,7 +154,7 @@ export class Picker {
             return;
         }
 
-        this.chipSet.setFocus(true);
+        this.chipSet?.setFocus(true);
```
adrianschmidt: Whaa… I assume this incorporates an automatic null-check, but I've never seen it before. New js-magic? Or something in typescript or Stencil?

src/components/tab-bar/examples/tab-bar.tsx
```diff
@@ -1,5 +1,5 @@
 import { Component, h, State } from '@stencil/core';
-import { Tab } from '../tab.types';
+import { Tab } from '@limetech/lime-elements';
```
adrianschmidt: 👏

src/components/tab-panel/examples/tab-panel-content.scss
```diff
@@ -0,0 +1,19 @@
+.text {
+    display: flex;
+    align-items: center;
+    margin-bottom: 1rem;
```
adrianschmidt:
```suggestion
@import '../../style/variables';

.text {
    display: flex;
    align-items: center;
    margin-bottom: pxToRem(16);
```

src/components/tab-panel/examples/tab-panel-content.scss
```diff
@@ -0,0 +1,19 @@
...
+    align-items: center;
+    flex-direction: column;
+}
+
+p {
+    margin: 0;
+}
+
+limel-icon {
+    margin-right: 1rem;
```
adrianschmidt:
```suggestion
    margin-right: pxToRem(16);
```

src/components/tab-panel/examples/tab-panel-content.scss
adrianschmidt: Scratch that. I didn't realise this was an example.

src/components/picker/picker.tsx
jgroth: It's new in Typescript 3.7, but it's also going to land in ES sometime in the future, I think the draft is in the final stages :slightly_smiling_face:

src/components/tab-panel/examples/tab-panel-content.tsx
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * Simulate loading the tab content
+     */
+    private loadTabContent() {
+        setTimeout(() => {
+            this.loaded = true;
+        }, LOAD_TIME);
+    }
+
+    public render() {
```
adrianschmidt: I think functions should be sorted with public before protected, and protected before private. The consultants look very much to our documentation and examples for guidance, so I think it's important to keep the examples as clean and well-formatted as any production code.

Befkadu1:
👏

src/components/tab-panel/tab-panel.mdx
```diff
@@ -0,0 +1,36 @@
...
+data it needs once the tab has been activated and then display the actual content. If the button is pressed, the 
+component will emit the `changeTab` event to change the badge inside the corresponding tab.
+
+<limel-example name="limel-example-tab-panel" />
+
+### Custom component
+
+This is the component that is used in the example above for each tab.
+
+<limel-example name="limel-example-tab-panel-content" path="tab-panel"/>
```
adrianschmidt: I guess there's no way to get this to not show a perpetual spinner, is there? 😣

jgroth: No I don't think there is an easy way :disappointed: 
The code is the most important part anyway since the component is displayed above as well

src/components/tab-panel/examples/tab-panel.tsx
jgroth: Yeah maybe :thinking: I think the most important part is to make the example intuitive and easy to understand, and I'm not sure which is better in that sense
What do you think @adrianschmidt ? Hard code it like this or do it in a more dynamic way for the example?

src/components/tab-panel/examples/tab-panel-content.tsx
jgroth: This is just temporary. We should update the example to use the badge once that is merged :slightly_smiling_face:

src/components/tab-panel/tab-panel.mdx
adrianschmidt: True. Imma see if I can make some magic in a different branch 😛

src/components/tab-panel/examples/tab-panel-content.tsx
jgroth: 1000 it is! :smile:

src/components/tab-panel/examples/tab-panel.tsx
adrianschmidt: It's probably better to keep it simple. Anyone creating tabs from an array ought to know how to convert this into a loop, and if not, it's probably a good thing that they'll have to ask someone. But for anyone not well acquainted with loops, and whose only goal is to create three static tabs, the current example is going to be much more helpful than one with a loop.

src/components/tab-panel/examples/tab-panel-content.tsx
jgroth: Fixed! However, I don't agree that there should be a general rule to order stuff in public > protected > private order. Code that belong together should stay together :slightly_smiling_face: If a public method uses a private one, I should not have to scroll to the bottom just to read it :slightly_smiling_face: There should be a natural flow of reading the code if possible

adrianschmidt: I both agree and disagree. All rules can be broken if the circumstances justify it. But I think there's value in putting the stuff that constitutes the external interface bundled together at the top. That gives the reader of the code a quick overview of the component and what it does. Any non-trivial implementation details are typically quite nice to put in well-named private functions further down. Whatever specific detail you are interested in, you can just jump to in any modern editor.

Of course it's a trade off, and I'm not saying it's a rule set in stone. But I think it's a good guideline.

src/components/tab-panel/tab-panel.mdx
adrianschmidt: Added option for code-only examples in #630 

I've rebased this onto that, and used the new option in the above example.

src/components/tab-panel/examples/tab-panel.tsx
Befkadu1: I am not sure if I understand it correctly but who is going to use lime-elements/this example (I am wondering to know our audience if this person is going to have a difficulty to understand basic javascript operation such as `map`)? We have implemented loops in our code as I see it in other examples, for example in Button component. 

I get the point It would be simple to understand the code if we don't use loop. When I see the code, I felt it is a redundant code.

src/components/tab-panel/tab-panel.tsx
adrianschmidt: This is used by `tab-panel-content.tsx` right?

Off the top of my head, I can't come up with any reason we shouldn't do this automatically, but please do share your concerns, there are probably things I haven't thought of!

jgroth: Yes, any content inside the tab-panel will receive the corresponding tab configuration as a `Prop() tab`. The component needs to declare a property named `tab` or it won't work, so it's kind of implicit. If the component already has another prop named `tab`, we will override it with this line of code.
We can make it more explicit by not doing this and forcing the consumer to send in the tab manually when using the tab-panel. It might make it slightly harder for us in the context of the webclient when we want to dynamically render custom web-components from consultants. In that case, we might have to send it in on the context or something like that :thinking:

adrianschmidt: Ah, ok. I understand your doubts then, but yes, it seems pretty nice to have it done automatically. I guess we could define an interface `TabPanel` which defines that it must have a `tab` property of type `Tab`? That wouldn't fix the "if it already exists we'll overwrite it"-part, but it ought to make it explicit enough at least?

jgroth: Yeah maybe :thinking: 
There's no way to enforce it though, so I wonder how useful it would be :thinking: 
The probability of placing a component in the tab-panel with a property named `tab` and having it be something else than the actual tab seems pretty slim though :smile:

adrianschmidt: Is there any way we can check if it's truthy before it's set the first time, and throw an explicit error if it is? To decrease the risk that an overwrite is missed and causes bugs?

jgroth: Yeah sure, throwing an error might not be necessary though... we could just ignore setting it

src/components/tab-panel/tab-panel.tsx
```diff
@@ -0,0 +1,93 @@
...
+
+/**
+ * @slot - Content to put inside the `tab-panel`. Each slotted element must have the `id` attribute equal to the id of the tab it belongs to.
+ */
+@Component({
+    tag: 'limel-tab-panel',
+    shadow: true,
+})
+export class TabPanel {
+    @Prop({ mutable: true })
```
adrianschmidt: Does this really have to be mutable?

jgroth: No, but it simplifies the code **alot**

My first implementation was without it being mutable and it forces us to have two arrays of the tabs and keep them in sync. Since we have to listen for the `changeTab` event to be able to display the active content (and also if the event comes from within the inner components), we must trigger a re-render. If the consumer does not add a `changeTab` listener and updates the tabs accordingy, it will break this component.

I think mutable might sound worse than it actually is. It will not mutate the consumer's reference to the tabs, only our local copy inside this component. When/if the consumer updates their reference, this one will be updated accordingly as well

src/components/tab-panel/tab-panel.tsx
adrianschmidt: Because we assume it's already set to the right thing? If that was the case, overwriting it shouldn't be a problem either, I think? I'm thinking of the potential case where someone built a component that takes some completely different input on a property called `tab` and that component is later placed into our tabs. Unless that component immediately fails because the prop was replaced, the bug could be easy to miss, since you're probably going to test that the tabs you just set up work as intended, but you might not thoroughly test the components you put in them, if they've always worked before… 🤔

src/components/tab-panel/tab-panel.tsx
adrianschmidt: > It will not mutate the consumer's reference to the tabs, only our local copy inside this component.

Ah, that's nice, I didn't know that 👍

I was partially asking for "should we update it with the event-handler-loop we usually do" part that you answered, but I was also partially asking because I wasn't sure if a "non-mutable" Prop decorator actually stopped internal changes to the *content* of the array, or only stopped internal replacing of the whole object itself. If all we're doing is updating properties on the objects contained in the array, I'd be pretty surprised if that required the `mutable` flag, but maybe it does?

jgroth: I don't think it will stop us changing the contents of the array, but that would also not trigger a re-render. Also, if we were to mutate the actual items in the array, that would probably also update the references that the consumer is holding onto :dizzy_face:

jgroth: Have not tested that though, but it seems like a really bad practice :smile:

adrianschmidt: Ah! Yeah, of course! I didn't think of that… 😳

👌

src/components/tab-panel/tab-panel.tsx
jgroth: If you have a component like that you could always wrap it inside another component. 

We have documented the behavior in the docs so I think I'm leaning towards just having it the way it is now. It seems we need a lot of book keeping if we want to keep track of which components to set the tab on... :slightly_frowning_face:

src/components/tab-panel/tab-panel.tsx
```diff
@@ -0,0 +1,93 @@
...
+    public tabs: Tab[] = [];
+
+    @Element()
+    private host: HTMLElement;
+
+    @Event()
+    protected changeTab: EventEmitter<Tab>;
```
adrianschmidt: I don't really understand how this manages to pass the compiler… 🤔

According to [this answer on StackOverflow](), `protected` works like this in typescript:

> The **protected** modifier acts much like the **private** modifier with the exception that members declared **protected** can also be accessed within deriving classes.

I don't really know what a *deriving class* is in typescript, but unless the definition includes "stuff in the same folder" or "stuff with the same prefix", I really can't figure out how the compiler determines that `TabPanelContentExample` is a "deriving class" of `TabPanel`…? 😂 

Regardless, it works, so I'm fine with it 😄

adrianschmidt: Completely regardless of `protected` vs `private`, I think `eventEmitter` is more important than `host`, so I'd place `eventEmitter` above `host` 🙂

src/components/tab-panel/tab-panel.tsx
adrianschmidt: Yeah, exactly. I don't think we should track it. I just mean that if it's simple for us to check if there's something there, for example in `componentWillLoad`, and throw an error with a helpful message, then I think it would be helpful to do that.

If it's not easy to check in a place where we know we haven't set anything ourselves yet, then I think it's fine to just do it the way we do it now, overwriting if there's a conflict.

src/components/tab-panel/tab-panel.tsx
jgroth: Hmm :thinking: `TabPanelContentExample` is not derived from `TabPanel` so thats a bit weird haha. I think deriving class is probably just a sub-class of another :slightly_smiling_face: 

Maybe we should document this in some way... we should not have to declare this at all since the event comes from the `tab-bar`, but the compiler complains that "there is no such event" if you specify the handler on `tab-panel` like in our example component. It's still valid though since all events propagates through the DOM.

I had it as `private` first, but then I got warnings about an unused property. I'm not sure why I did not have it as `public` though :thinking: Maybe I didn't want it to end up in the documentation, but now I see that it's there anyway

jgroth: Yeah I agree, I usually place the events after the props :+1: 

Interestingly though, I found this the other day and they have the `Element` almost above everything else :smile: 
https://stenciljs.com/docs/style-guide

I'm not saying that we should follow that 100%, but some of the points are quite interesting. Like placing the watchers right next to prop that is being watched for instance :slightly_smiling_face:

adrianschmidt: Oh, yeah, this isn't even the one used on [line 92 in tab-panel-content.tsx](https://github.com/Lundalogik/lime-elements/pull/627/files?file-filters%5B%5D=.js&file-filters%5B%5D=.mdx&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-68164d3e9ca96b930b2967af27e2561cR92)… I just searched the diff for a .emit, and that was the only one I found so I didn't really look further into it.

Ok, but that explains why `protected` works here 😄

So, the compiler complains about [line 41 in tab-panel.tsx](https://github.com/Lundalogik/lime-elements/pull/627/files?file-filters%5B%5D=.js&file-filters%5B%5D=.mdx&file-filters%5B%5D=.ts&file-filters%5B%5D=.tsx#diff-910770ff727a8a4c863417a2b6566e32R41) unless this is declared here?

Yeah, that's annoying… 😖

On the other hand, since the "normal" way of consuming `tab-bar` will be through this component, it doesn't seem wildly inappropriate that this component documents the `changeTab` event, does it? From the perspective of someone reading the docs I mean.

jgroth: And now I also found this in case we *do want* to enforce some rules :thinking: 
https://www.npmjs.com/package/tslint-stencil

jgroth: No I guess not :slightly_smiling_face: It could just as well have been emitted from this component :slightly_smiling_face:

src/components/tab-panel/tab-panel.mdx
```diff
@@ -0,0 +1,36 @@
...
+or render anything until the tab is active. 
+
+The slotted components can also emit the `changeTab` event to update anything inside the actual tab, e.g. to change
+the icon, color or badge.
+
+<limel-props name="limel-tab-panel" />
+
+## Example
+
+This example illustrates how to add custom components inside the `tab-panel`. Each component will simulate to load the
```
adrianschmidt:
```suggestion
This example illustrates how to add custom components inside the `tab-panel`. Each component will simulate loading the
```

src/components/tab-panel/tab-panel.mdx
```diff
@@ -0,0 +1,36 @@
...
+data it needs once the tab has been activated and then display the actual content. If the button is pressed, the 
+component will emit the `changeTab` event to change the badge inside the corresponding tab.
+
+<limel-example name="limel-example-tab-panel" />
+
+### Custom component
+
+This is the component that is used in the example above for each tab.
```
adrianschmidt:
```suggestion
This is the component that is used for each panel in the example above.
```

adrianschmidt:
```suggestion
This is the component that is used for each tab in the example above.
```

adrianschmidt: I don't know which one is better… 🤔

src/components/tab-panel/tab-panel.mdx
```diff
@@ -0,0 +1,36 @@
...
+
+# Tab panel 
+
+### 🚧 This feature is work in progress. It can be tested out by setting the feature switch `enableTabs` 🚧
+
+The `tab-panel` component uses the `tab-bar` component together with custom slotted components and will display the 
+content for the currently active tab. Each slotted component must have an id equal to the id of the corresponding tab
+it belongs to.
+
+The `tab-panel` component will automatically send each tab configuration to the corresponding slotted component as a
```
adrianschmidt:
```suggestion
The `tab-panel` component will automatically set each tab configuration on the corresponding slotted component as a
```

src/components/tab-panel/examples/tab-panel-content.tsx
```diff
@@ -0,0 +1,95 @@
...
+            <div class="container">
+                <div class="text">
+                    <limel-icon
+                        badge={true}
+                        size="large"
+                        name={this.tab.icon}
+                        style={style}
+                    />
+                    <p>
+                        {this.tab.text} as received {this.votes} votes!
```
adrianschmidt:
```suggestion
                        {this.tab.text} has received {this.votes} votes!
```

jgroth:
I did nothing else than squashing :slightly_smiling_face:

PR #628, opened 2020-03-05
Improving lime elements

Kiarokh:
Fixing a richer and more mature foundation for a globally shared design system for all Lime products.

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -1,4 +1,4 @@
-@import '../../style/variables';
+@import '../../style/internal/mdc-variables';
 @import "@limetech/mdc-button/mdc-button";
 
 :host {
--- a/src/color-system.mdx.bak
+++ b/src/color-system.mdx.bak
@@ -1,0 +1,27 @@
+---
+name: Color system
+route: /color-system
+menu: Design Guidelines
+---
+
+# Color system
+
+
+## Extended color palette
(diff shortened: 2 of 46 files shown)


src/components/button/button.scss
```diff
@@ -1,5 +1,8 @@
-@import '../../style/variables';
 @import "@limetech/mdc-button/mdc-button";
+@import '../../style/mdc-imports';
+@import '../../style/mdc-typography';
```
adrianschmidt: The order of these imports are important, because the variables have to be imported before the component styles to have any effect. So the the button doesn't quite look like we want it to with this change 😄

(I'll fix that of course, since I'm working on the PR. But I wanted to let you know for the future 👍)

adrianschmidt: image
image
image

adrianschmidt: Ah, the last two screenshots was actually because I had my watcher running while fiddling with the files. But the all-caps on the first screenshot is there in the PR.

Or that change was intentional?

PR #630, opened 2020-03-06
Code only examples

adrianschmidt:
### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/examples/example.tsx
+++ b/src/examples/example.tsx
@@ -33,7 +33,7 @@ export class Example {
         const tsxUrl = `${BASE_URL}public/stencil/components/${path}/examples/${type}.tsx`;
 
         this.fetchData(tsxUrl).then(tsxData => {
-            this.tsxCode = prism.highlight(tsxData, prism.languages.tsx);
+            this.tsxCode = prism.highlight(tsxData, prism.languages.tsx, 'tsx');
 
             const styleUrlMatch = tsxData.match(/styleUrl: '(.*)'/);
             if (styleUrlMatch) {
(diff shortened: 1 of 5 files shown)


src/examples/example.tsx
```diff
@@ -68,6 +65,20 @@ export class Example {
...
+        return this.codeOnly
+            ? null
+            : [
+                  <limel-config
+                      config={{ iconPath: `${BASE_URL}public/stencil/` }}
+                  />,
+                  <div class="example">
+                      <ExampleComponent />
+                  </div>,
+              ];
```
jgroth: It's a bit hard to read this code and I think it's because of the ternary operator. I think we should avoid using the it unless the expression is on a single line. 

```suggestion
    if (this.codeOnly) {
        return;
    }

    const ExampleComponent = this.name;    
    return [
        <limel-config
            config={{ iconPath: `${BASE_URL}public/stencil/` }}
        />,
        <div class="example">
            <ExampleComponent />
        </div>
    ];
```

adrianschmidt: Yeah, that makes sense 👍

adrianschmidt: :zap:

Interesting… I think the conversation was auto-resolved because I committed the suggestion…? At least I don't think I resolved it manually…

PR #633, opened 2020-03-10
Fix dev build

adrianschmidt:
When running `npm run watch` (or `npm start`), lime-elements is built in dev-mode. However, StencilJS does not output the loader files necessary to load lime-elements in Angular when built in this mode. So we add `npm run watch:prod` for building the prod-build with a running watcher, thus gaining the benefits of auto-build while also being able to load the result in lime-webclient.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "cm": "git-cz",
     "dev": "cross-env-shell SASS_PATH=node_modules \"stencil build --dev --docs\"",
     "watch": "cross-env-shell SASS_PATH=node_modules \"stencil build --dev --watch --docs\"",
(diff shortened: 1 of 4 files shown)


adrianschmidt:
@Kiarokh: 🎁 for you! With this you will no longer have to re-run the build manually each time you changed something in lime-elements and want to see the result in the webclient 🙂

adrianschmidt:
Related bug report to Stencil: ionic-team/stencil#2258

(This PR is a workaround for the bug though, so we don't need to wait for any action from the Stencil team.)

Kiarokh:
@adrianschmidt awesome man. 🎉great job!

PR #634, opened 2020-03-10
Move lint to GitHub action

adrianschmidt:


Makefile
```diff
@@ -5,17 +5,6 @@ build:
...
-.PHONY: lint
-lint:
-	@# Lints all applicable files.
-	docker run --rm -w /lime $(DOCKER_IMAGE) npm run lint
-
-.PHONY: commitlint
-commitlint: HASH=''
-commitlint:
-	@# Lint the commit message of the commit with the given HASH
-	docker run --rm -w /lime $(DOCKER_IMAGE) npx commitlint -f $(HASH)^ -t $(HASH)
```
adrianschmidt: @jensgustafsson Stavros told me that you usually have opinions when he tries to remove make commands related to Docker, so I should sync with you about this 🙂

adrianschmidt: The commitlint command is already run in GitHub Actions instead of in Jenkins. It's using a 3rd party action, and I've actually already removed the dependency on commitlint. I had just forgotten to clean up the makefile. I really don't think we have all that much use for being able to run commitlint locally. I'd be very surprised if anyone but me has done it even once 😄

adrianschmidt: The lint command is moving from Jenkins to GitHub Actions in this PR, but as you can see in the diff, I haven't used any 3rd party actions for the linting. The workflow is just running the scripts we already have defined in our package.json, since we need to run those locally all the time anyway.

PR #639, opened 2020-03-16
fix(styles): fix broken css variables for shadows, introduced in v26.11.0

adrianschmidt:
Import the scss function `pxToRem` so that values are converted correctly.

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/style/shadows.scss
```diff
@@ -1,3 +1,4 @@
+@import './functions';
```
adrianschmidt: I want to avoid ever importing anything from the `internal` folder in the stylesheets that go out externally. That's why I moved this file.

PR #640, opened 2020-03-16
Action menu on list items

### LINKED ISSUE #622: Add ellipsis menu to list items in lime-elements

hannahu:
fix Lundalogik/crm-feature#989
### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/list/list-renderer.tsx
+++ b/src/components/list/list-renderer.tsx
@@ -6,7 +6,6 @@ import { RadioButtonTemplate } from './radio-button/radio-button.template';
 
 export class ListRenderer {
     private defaultConfig: ListRendererConfig = {
-        isOpen: true,
         badgeIcons: false,
     };
 
@@ -70,7 +69,7 @@ export class ListRenderer {
         return (
             <ul
                 class={classNames}
-                aria-hidden={(!this.config.isOpen).toString()}
+                aria-hidden={(this.config.type === 'menu').toString()}
                 role={role}
                 aria-orientation="vertical"
             >
--- a/src/components/list/list-renderer.tsx
+++ b/src/components/list/list-renderer.tsx
@@ -6,7 +6,6 @@ import { RadioButtonTemplate } from './radio-button/radio-button.template';
 
 export class ListRenderer {
     private defaultConfig: ListRendererConfig = {
-        isMenu: false,
         isOpen: true,
         badgeIcons: false,
     };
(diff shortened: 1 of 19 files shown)


src/components/list/list.tsx
```diff
@@ -43,11 +47,18 @@ export class List {
     @Prop()
     public iconSize: IconSize = 'small';
 
+    /**
+     * Set to `true` if the list is open
+     */
+    @Prop()
+    public isOpen: boolean;
```
jgroth: Just adding a note here so I do not forget about it. We should investigate if we really need to add this, it's not very intuitive that a list has an `open` property. If we really need it, we might want to mark it as `internal` and add to the documentation that it's only used for the menu

src/components/menu/menu.tsx
```diff
@@ -39,6 +30,12 @@ export class Menu {
     @Prop({ reflectToAttr: true })
     public disabled = false;
 
+    /**
+     * Decides if the menu should open right or left. Defaults to right.
+     */
+    @Prop({ reflectToAttr: true })
+    public openDirection: OpenDirection = 'right';
```
adrianschmidt: Defaulting to the right is a good choice I think, but it does kind of make this change breaking…

image

I'm fixing the above in Lundalogik/lime-webclient#1247

hannahu: There are more issues with the menu... opened an issue yesterday in the regular board:
https://github.com/Lundalogik/crm-feature/issues/1008

adrianschmidt: Ah, ok. Good that they've been caught 🙂

src/components/list/list-renderer.tsx
```diff
@@ -70,7 +69,7 @@ export class ListRenderer {
         return (
             <ul
                 class={classNames}
-                aria-hidden={(!this.config.isOpen).toString()}
+                aria-hidden={(this.config.type === 'menu').toString()}
```
adrianschmidt: This doesn't seem quite right. Won't the menu list always be hidden from ARIA users?

src/components/menu/menu.tsx
```diff
@@ -61,61 +58,39 @@ export class Menu {
...
-                {this.disabled ? <div class="menu-disabled" /> : null}
+                <limel-portal
+                    visible={this.open}
+                    containerId={this.portalId}
+                    openDirection={this.openDirection}
+                >
+                    <limel-menu-surface
+                        open={this.open}
+                        onDismiss={this.onClose}
+                    >
```
adrianschmidt: This version doesn't seem to be using a scrim, so how do we detect clicks outside the menu to make the menu close?

PR #642, opened 2020-03-17
Remove obsolete fix for dev-build

adrianschmidt:
This reverts commit 8a3e4fb39a8a48c4a1da8325cc5fd3590646a1d9 from #633.


adrianschmidt:
FYI @Kiarokh: The updated version of Stencil fixes the bug in the dev build, so you can now run `npm start` and still have your local lime-elements installed in lime-webclient. So this PR removes the `npm run watch:prod` script, as it's no longer needed.

PR #644, opened 2020-03-18
fix(icon): log to console instead of raising error when no container found

adrianschmidt:
Sometimes the watcher for `name` triggers rendering of the svg before there's a container element present. The error raised gives no clue to which icon is causing trouble, so check if the element is present, and if not, log a message including the name of the icon to the console.

### DIFF
--- a/src/components/icon/icon.tsx
+++ b/src/components/icon/icon.tsx
@@ -45,7 +45,7 @@ export class Icon {
             return;
         }
         const svgData = await this.loadSvg(name);
-        this.renderSvg(svgData);
+        this.renderSvg(svgData, name);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/components/icon/icon.tsx
```diff
@@ -65,7 +65,15 @@ export class Icon {
...
-        this.host.shadowRoot.querySelector('div.container').innerHTML = svgData;
+    private renderSvg(svgData: string, name: string) {
+        const container = this.host.shadowRoot.querySelector('div.container');
+        if (container) {
+            container.innerHTML = svgData;
+        } else {
+            // tslint:disable-next-line:no-console
+            console.error(
+                `limel-icon: no container found when trying to render svg! icon: ${name}`
+            );
```
adrianschmidt: I don't know why the bug appears, and I only saw it while testing some other stuff. I couldn't see any icons that where missing, and once I had added this to try to find out what the problem was, it never logged anything…

I thought it might be a reasonable thing to have anyway. We could of course throw an error instead of just logging, but make sure to include the icon name in the error message.

PR #656, opened 2020-03-22
Prettier default arrow parens

adrianschmidt:
Since version 1.9, Prettier has had an option to always wrap arrow function arguments with parentheses. In version 2.0, this behavior has become the default.

```js
// Input
const fn = (x) => y => x + y;

// Prettier 1.19
const fn = x => y => x + y;

// Prettier 2.0
const fn = (x) => (y) => x + y;
```

At first glance, avoiding parentheses in the isolated example above may look like a better choice because it results in less visual noise. However, when Prettier removes parentheses, it becomes harder to add type annotations, extra arguments, default values, or a variety of other things. Consistent use of parentheses provides a better developer experience when editing real codebases, which justifies the change.

You are encouraged to use the new default value, but if the old behavior is still preferred, please configure Prettier with { "arrowParens": "avoid" }.

### DIFF
--- a/rollup-plugin-docz-starter.ts
+++ b/rollup-plugin-docz-starter.ts
@@ -20,7 +20,7 @@ function doczStarter() {
             const docz = spawn('npx', ['docz', 'dev']);
             started = true;
 
-            docz.on('close', code => {
+            docz.on('close', (code) => {
                 console.log('📔 - Closing Docz...', code);
             });
         },
--- a/src/components/checkbox/checkbox.template.tsx
+++ b/src/components/checkbox/checkbox.template.tsx
@@ -10,7 +10,9 @@ interface CheckboxTemplateProps {
     label?: string;
 }
 
-export const CheckboxTemplate: FunctionalComponent<CheckboxTemplateProps> = props => {
+export const CheckboxTemplate: FunctionalComponent<CheckboxTemplateProps> = (
+    props
+) => {
     return (
         <div class="mdc-form-field ">
             <div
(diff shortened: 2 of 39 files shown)


rollup-plugin-docz-starter.ts
```diff
@@ -20,7 +20,7 @@ function doczStarter() {
             const docz = spawn('npx', ['docz', 'dev']);
             started = true;
 
-            docz.on('close', code => {
+            docz.on('close', (code) => {
```
adrianschmidt: I know this is probably going to bug some people out. Prettier has changed their default setting for whether or not to use parenthesis in arrow functions with a single argument, and I think their reasoning is pretty good to be honest ([link](https://prettier.io/blog/2020/03/21/2.0.0.html#change-default-value-for-arrowparens-to-always-7430httpsgithubcomprettierprettierpull7430-by-kachkaevhttpsgithubcomkachkaev)).

No one will have to change how they write these, since the linter will autofix them anyway.

src/components/checkbox/checkbox.template.tsx
```diff
@@ -10,7 +10,9 @@ interface CheckboxTemplateProps {
     label?: string;
 }
 
-export const CheckboxTemplate: FunctionalComponent<CheckboxTemplateProps> = props => {
+export const CheckboxTemplate: FunctionalComponent<CheckboxTemplateProps> = (
+    props
+) => {
```
adrianschmidt: This is a little unfortunate. I assume it's an indirect result of the added parenthesis, as the added line length probably caused Prettier to break the line. Not the end of the world though, imo.

PR #664, opened 2020-03-25
fix(icon): remove unnecessary `console.error` logging

adrianschmidt:
While the `renderSvg` function is called too early, regularly if not always, it seems it always gets called again, making the icon appear correctly. Thus, the error-level logging is unnecessary.

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/icon/icon.tsx
+++ b/src/components/icon/icon.tsx
@@ -45,7 +45,7 @@ export class Icon {
             return;
         }
         const svgData = await this.loadSvg(name);
-        this.renderSvg(svgData, name);
+        this.renderSvg(svgData);
     }
 
     /**
@@ -65,15 +65,10 @@ export class Icon {
      * will remain the way they look in the file.
      * Therefore we inject the svg as inline markup instead.
      */
-    private renderSvg(svgData: string, name: string) {
+    private renderSvg(svgData: string) {
         const container = this.host.shadowRoot.querySelector('div.container');
         if (container) {
             container.innerHTML = svgData;
-        } else {
-            // tslint:disable-next-line:no-console
-            console.error(
-                `limel-icon: no container found when trying to render svg! icon: ${name}`
-            );
         }
     }
 }


src/components/icon/icon.tsx
```diff
@@ -65,15 +65,10 @@ export class Icon {
      * will remain the way they look in the file.
      * Therefore we inject the svg as inline markup instead.
      */
-    private renderSvg(svgData: string, name: string) {
+    private renderSvg(svgData: string) {
```
adrianschmidt: What the heck? Building this locally (with `npm run build`) fails, because I forgot to remove the `name` argument in the call. But it passes `npm run build` in the CI job in GitHub Actions?! That's not good…

adrianschmidt: (Now I've pushed the fix, so it should be green. But [this run](https://github.com/Lundalogik/lime-elements/runs/532907419?check_suite_focus=true) shouldn't have been green.)

src/components/icon/icon.tsx
```diff
@@ -65,15 +65,7 @@ export class Icon {
...
-        if (container) {
-            container.innerHTML = svgData;
-        } else {
-            // tslint:disable-next-line:no-console
-            console.error(
-                `limel-icon: no container found when trying to render svg! icon: ${name}`
-            );
-        }
+    private renderSvg(svgData: string) {
+        this.host.shadowRoot.innerHTML = svgData;
```
adrianschmidt: This bypasses the problem completely 🙂

jgroth: Unfortunately this will not work :slightly_frowning_face: This will overwrite *all* the elements inside the host, including the `<style>` tag generated by Stencil. If we did not have any styles it would be fine though :smile:

adrianschmidt: Ah crap… good catch!

adrianschmidt: image

😂😂😂

adrianschmidt: ⚡️

PR #665, opened 2020-03-25
feat(tabs): new tab-bar and tab-panel components

jgroth:
There are no more tasks right now for the tab-bar and tab-panel and the interfaces are probably how we want them, so we can probably remove the feature switches now :slightly_smiling_face: 

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tab-bar/tab-bar.scss
+++ b/src/components/tab-bar/tab-bar.scss
@@ -3,10 +3,11 @@
 @import '@limetech/mdc-tab-scroller/mdc-tab-scroller';
 @import '@limetech/mdc-tab-indicator/mdc-tab-indicator';
 @import '@limetech/mdc-tab/mdc-tab';
+
 :host {
-  display: block;
+    display: block;
 }
 
 .mdc-tab-bar__icon {
-  margin-right: .5rem;
+    margin-right: .5rem;
 }
--- a/src/components/tab-bar/tab-bar.tsx
+++ b/src/components/tab-bar/tab-bar.tsx
@@ -10,7 +10,6 @@ import {
 import { MDCTabBar, MDCTabBarActivatedEvent } from '@limetech/mdc-tab-bar';
 import { strings } from '@limetech/mdc-tab-bar/constants';
 import { Tab } from './tab.types';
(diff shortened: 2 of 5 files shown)


src/components/tab-bar/tab-bar.scss
```diff
@@ -3,10 +3,11 @@
...
 @import '@limetech/mdc-tab/mdc-tab';
+
 :host {
-  display: block;
+    display: block;
 }
 
 .mdc-tab-bar__icon {
-  margin-right: .5rem;
+    margin-right: .5rem;
```
adrianschmidt: Someone doesn't have their editorconfig support set up 🙂

src/components/tab-bar/tab-bar.tsx
```diff
@@ -72,7 +71,7 @@ export class TabBar {
     }
 
     private setup() {
-        const element = this.element.shadowRoot.querySelector('.mdc-tab-bar');
+        const element = this.host.shadowRoot.querySelector('.mdc-tab-bar');
```
adrianschmidt: We use `host` in `tab-panel` and I think it's less confusing to not have both `element` and `this.element`.

src/components/tab-bar/tab-bar.tsx
```diff
@@ -21,21 +20,21 @@ const { TAB_ACTIVATED_EVENT } = strings;
...
-    private setupMdc = false;
-
     @Event()
     private changeTab: EventEmitter<Tab>;
 
+    @Element()
+    private host: HTMLElement;
+
+    private mdcTabBar: MDCTabBar;
+    private setupMdc = false;
```
adrianschmidt: I think it's better to have external stuff above the internal stuff. It makes it easier to see the important stuff.

jgroth:
I am! :smile: 
:heavy_check_mark:

PR #678, opened 2020-04-02
fix(badge): make badges with one digit appear like a full circle

Kiarokh:
Badges that have only 1 digit had a stretched shape. This both looks unusual and wastes a bit of space.

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/badge/badge.scss
+++ b/src/components/badge/badge.scss
@@ -5,9 +5,8 @@
  * @prop --badge-text-color: badge text color
  */
 
-
 :host {
-  display: block;
+    display: block;
 }
 
 * {
--- a/src/components/badge/badge.scss
+++ b/src/components/badge/badge.scss
@@ -10,6 +10,10 @@
   display: block;
 }
 
+* {
+    box-sizing: border-box;
+}
+
 .badge-container {
     display: inline-block;
     font-size: pxToRem(11);
--- a/src/components/badge/badge.scss
+++ b/src/components/badge/badge.scss
@@ -12,14 +12,14 @@
 
 .badge-container {
     display: inline-block;
-    font-size: 0.7rem;
-    min-width: 1.2rem;
-    height: 1.2rem;
-    border-radius: 1.2rem;
-    line-height: 1.2rem;
+    font-size: pxToRem(11);
+    min-width: pxToRem(20);
+    height: pxToRem(20);
+    border-radius: pxToRem(20);
+    line-height: pxToRem(20);
     text-align: center;
     background-color:  var(--badge-background-color, #EEF1F5);
     color:  var(--badge-text-color, #7D899C);
-    padding: 0 0.3rem;
-    margin-left: 0.25rem;
+    padding: 0 pxToRem(5);
+    margin-left: pxToRem(4);
 }
--- a/src/components/badge/badge.scss
+++ b/src/components/badge/badge.scss
@@ -9,6 +9,10 @@
     display: block;
 }
 
+:host([hidden]) {
+    display: none;
+}
+
 * {
     box-sizing: border-box;
 }


src/components/badge/badge.scss
```diff
@@ -13,7 +13,7 @@
 .badge-container {
     display: inline-block;
     font-size: 0.7rem;
-    min-width: 1.2rem;
+    min-width: .6rem;
```
adrianschmidt, line 16: Small sidenote: I don't think this commit should be a `refactor`, it's either a `feat` or a `fix` 🙂

`refactor` is for changes to the code that has no impact on how the code functions (including no visual impact). Commits of type `refactor` will not be listed in the changelog.

src/components/badge/badge.scss
```diff
@@ -5,21 +5,24 @@
  * @prop --badge-text-color: badge text color
  */
 
-
 :host {
   display: block;
 }
 
+* {
+    box-sizing: border-box;
```
adrianschmidt, line 13: Before:
image
After:
image

(Well, the change in vertical alignment comes from the tweaked sizes below, but mostly, setting the box-sizing did the job.)

src/components/badge/badge.scss
```diff
@@ -5,21 +5,28 @@
...
 
 .badge-container {
     display: inline-block;
-    font-size: 0.7rem;
-    min-width: 1.2rem;
-    height: 1.2rem;
-    border-radius: 1.2rem;
-    line-height: 1.2rem;
+    font-size: pxToRem(11);
+    min-width: pxToRem(20);
```
Befkadu1, line 23: Any advantages changing rem -> pxToRem(11)?

src/components/badge/badge.scss
```diff
@@ -5,21 +5,28 @@
...
 :host {
-  display: block;
+    display: block;
+}
+
+:host([hidden]) {
+    display: none;
+}
+
+* {
```
Befkadu1, line 16: If I may ask, why didn't you add the setting `box-sizing` in `.badge-container` class? I just wondered

src/components/badge/badge.scss
adrianschmidt: The advantages, as I see them, is:

- (I think most of us tends to think in pixels, and if you do) `pxToRem(10)` is much easier to understand than `0.625rem`
- Since the vast majority of users have the default font size of 16px in their browser, using `pxToRem(x)`, where `x` is an integer, ensures that the end result is a rem value that translates to an integer number of pixels for that vast majority of users.

Nowadays, whole pixels are (at least almost always) rendered in a very predictable manner across different browsers and platforms. With most of our target browsers now using variants of Chromium, I imagine fractions of pixels are probably rendered a lot more similarly across browsers than only a few years ago, but I wouldn't trust the outcome on one machine to necessarily translate across the spectrum.

src/components/badge/badge.scss
adrianschmidt: Two reasons:

- Mainly to make it easy to copy and paste to other files. (I'd like to have this in every component, but I'm not sure if doing that might screw up the stylings we get from MDC.)
- Secondarily, just in case we add more elements some time in the future, not applying the same box-sizing to all elements in the component could be extremely confusing.

adrianschmidt: You may always ask! 😉
Asking is always good 👍

src/components/badge/badge.scss
adrianschmidt: Before:
image
Only changing the box-sizing:
image
Both changing the box-sizing and using whole pixels:
image

Notice especially that the red badge isn't round in the middle image.

PR #679, opened 2020-04-02
Elevated chips

Kiarokh:
ezgif-3-0916cde1c41d

This is start of a process for unifying the look & feel and behaviour of clickable elements such as Chips, Buttons, Select dropdowns, etc... starting with Chips

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -1,3 +1,4 @@
+@import '../../style/mixins';
 @import '../../style/internal/mdc-variables';
 
 $mdc-textarea-background: $mdc-theme-background;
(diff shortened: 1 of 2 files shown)


src/style/mixins.scss
```diff
@@ -5,3 +5,20 @@
     }
   }
 }
+
+
+@mixin is-elevated-clickable() {
```
adrianschmidt, line 10: This is a pretty good example of a `refactor` type commit. If it would have replaced existing copy-and-pasted code in several places with this mixin, it would have been a spot-on `refactor`. Since it only adds a mixin for use in upcoming commits, this could also be marked as a `chore`, it doesn't really matter.

src/components/chip-set/chip-set.scss
```diff
@@ -22,6 +24,15 @@ $mdc-text-field-fullwidth-bottom-line-color: $lime-text-field-bottom-line-color;
...
  */
 
+.mdc-chip {
+    background-color: $mdc-chip-background-color;
+    @include is-elevated-clickable;
+
+    &:hover {
+        background-color: $mdc-chip-background-color;
+    }
+}
```
adrianschmidt, line 34: This on the other hand is either a `feat` or a `fix`. I'd consider it a `feat`, since it's not really fixing a bug… even though it's a *really* small feature 😄

adrianschmidt: (I'll fix)

adrianschmidt: ⚡️

PR #680, opened 2020-04-02
Improve chip styles

Kiarokh:
These changes align leading icons more correctly within the chips, since they are slightly smaller than the chip itself, then there should be equal margin around the chip:
old: Screen Shot 2020-04-02 at 13 42 50
 new:Screen Shot 2020-04-02 at 13 35 59

... and also makes the remove (x) buttons on chips better:
Screen Shot 2020-04-02 at 13 40 16

It's hard to know whether the cursor is hovering the (x) buttons that remove chips, or whether it's hovering the chip itself. These two of course result in totally different consequences. Therefore it's important to give proper feedback to users, when they hover on the (x) buttons. Also, since their previous size was too small, it was really hard to target them on touch screen devices. Now increased size makes it slightly easier.

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -33,7 +33,7 @@ $mdc-text-field-fullwidth-bottom-line-color: $lime-text-field-bottom-line-color;
 
 limel-icon.mdc-chip__icon.mdc-chip__icon--leading {
     background-color: var(--icon-background-color, transparent);
-    margin-left: pxToRem(-12) !important;
+    margin-left: pxToRem(-11) !important;
     color: var(--icon-color, rgba(0,0,0,.54));
 }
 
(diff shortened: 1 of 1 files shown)


src/components/chip-set/chip-set.scss
```diff
@@ -22,6 +24,15 @@ $mdc-text-field-fullwidth-bottom-line-color: $lime-text-field-bottom-line-color;
  * @prop --input-chip-set-selected-chip-color: Color of the highlight around selected chips in input chip-sets.
  */
 
+.mdc-chip {
+    background-color: $mdc-chip-background-color;
+    @include is-elevated-clickable;
+
+    &:hover {
+        background-color: $mdc-chip-background-color;
```
adrianschmidt, line 32: Commits that change how components look shouldn't be of type `refactor` :)

(I'll fix.)

PR #683, opened 2020-04-03
chore: remove `.docz` folder at start of `npm run watch` (and `npm start`)

adrianschmidt:
Since the introduction of the doczStarter rollup plugin, docz has trouble starting at times.
Removing the `.docz` folder before running the watcher usually resolves this problem. Add removal of
the folder to the beginning of the script, so it doesn't have to be done manually all the time.


package.json
```diff
@@ -18,7 +18,7 @@
     "build": "cross-env-shell NODE_ENV=prod SASS_PATH=node_modules \"stencil build --config stencil.config.dist.ts\"",
     "cm": "git-cz",
     "dev": "cross-env-shell SASS_PATH=node_modules \"stencil build --dev --docs\"",
-    "watch": "cross-env-shell SASS_PATH=node_modules \"stencil build --dev --watch --docs\"",
+    "watch": "shx rm -rf .docz/ && cross-env-shell SASS_PATH=node_modules \"stencil build --dev --watch --docs\"",
```
adrianschmidt, line 21: `shx` is the cli utility for `shelljs` (the latter of which we already use in `publish-docs.js`). This should make the `rm` command work on Windows too (although I haven't actually tested it).

adrianschmidt:
Quick-check ok from @jgroth, so merging.

PR #684, opened 2020-04-03
Better tabs, inspired by Chrome

Kiarokh:
fix Lundalogik/lime-elements#690

Vanilla tabs from Material Design have two major problems. Firstly, they are ugly and lifeless, too flat and rigid. Secondly, the active tab indicator (the accented bottom border) is not a decent visual sign to indicate which tab is active. It looks like a divider, but what you want to do is actually connecting the active tab to its contents, not separating it from its contents, using a divider. A better UI design for tabs is actually something similar to what Google Chrome has. In Chrome, the active tab is visually distinct due to its background color, therefore users notice it directly. It also looks connected to what is rendered below it, with a minimalist design.
**GIF Example: Tab Bar**
tab-bar

**GIF Example: Tab Panel**
tab-panel

Here I tried to do the same thing inspired by Google Chrome, also taking into account the upcoming dark-mode. 

Dark mode will require proper usage of css variables on visual elements that can dynamically switch colors. This is specially important for outer curved edges on the active tab; which were pretty tricky to make with pure css. 

**Note** that in this PR, I'm not using css variables, and rather use colors as sass variables, copied as HEX directly from our paletts. The reason is that we still haven't started the dark-mode project, and need to do some finalization / optimization in our palettes.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tab-bar/tab-bar.scss
+++ b/src/components/tab-bar/tab-bar.scss
@@ -4,10 +4,102 @@
 @import '@limetech/mdc-tab-indicator/mdc-tab-indicator';
 @import '@limetech/mdc-tab/mdc-tab';
 
+$tab-content-background-color: #fff;
+$tab-background-color: #f6f6f7;
+$tab-border-radius: .65rem;
+$tab-active-outer-edge-curve-size: .75rem;
+$tab-separator-width: .125rem;
+$tab-separator-background-color: #e8e8ea;
+
 :host {
     display: block;
 }
 
 .mdc-tab-bar__icon {
     margin-right: .5rem;
+}
+
+.mdc-tab-indicator {
+    .mdc-tab-indicator__content  {
+        border: none;
+    }
+}
+
+.mdc-tab-scroller__scroll-content {
+    padding: .5rem $tab-active-outer-edge-curve-size 0 $tab-active-outer-edge-curve-size;
+    background-color: $tab-background-color;
+}
+
+.mdc-tab__ripple {
+    box-sizing: border-box;
+    border-radius: $tab-border-radius;
+    border: {
+        style: solid;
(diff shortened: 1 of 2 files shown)


Kiarokh:
Also we still need to add some guidelines in lime elements documentations regarding usage of colors for tabs. It's important that consultants don't change them so that we can easily transition to dark-mode later. When this is merged and exposed to Web Client, we need to do a bit of tweaks for Lists and Activity Feed colors (as they will be rendered on a kind of off-white background color now).

Kiarokh:
I'll slightly tweak colors here. So please wait a bit for my update.

adrianschmidt:
@Befkadu1 Did you look at the commit message for this PR? Because this is not a refactor…

src/components/tab-bar/tab-bar.scss
```diff
@@ -4,10 +4,102 @@
 @import '@limetech/mdc-tab-indicator/mdc-tab-indicator';
 @import '@limetech/mdc-tab/mdc-tab';
 
+$tab-content-background-color: #fff;
+$tab-background-color: #f6f6f7;
+$tab-border-radius: .65rem;
+$tab-active-outer-edge-curve-size: .75rem;
+$tab-separator-width: .125rem;
+$tab-separator-background-color: #e8e8ea;
```
adrianschmidt, line 12: Should we document these, or do we want to keep them semi-secret?

@Kiarokh @jgroth

PR #696, opened 2020-04-14
fix(picker): handle updates to the `searcher` property

adrianschmidt:
A debounced version of the searcher function was created and stored when the component was created, but there was no handling of updates, and no handling for missing values. The debounced version will now only be created if the value of the `searcher` property is a function, and if the value is updated, a new debounced version will be created, replacing the old one.

re Lundalogik/crm-feature#1017

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [x] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


adrianschmidt:
To actually *fix* the linked issue, we obviously need to update to the new version of lime-elements created when this is merged 🙂

Issue #276, opened 2019-02-01
limel-input-field: add keyboard support to autocomplete


BregenzerK:
## New feature description

Choosing one of the suggestions should be possible by using the UP and DOWN arrow and ENTER should select an option

PR #697, opened 2020-04-15
add keyboard support for autocomplete on input-field

Befkadu1:
fix Lundalogik/lime-elements#276
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -9,12 +9,19 @@ import {
     State,
 } from '@stencil/core';
 import {
+    ARROW_DOWN,
+    ARROW_DOWN_KEY_CODE,
+    ARROW_UP,
+    ARROW_UP_KEY_CODE,
     ENTER,
     ENTER_KEY_CODE,
     SPACE,
     SPACE_KEY_CODE,
+    TAB,
+    TAB_KEY_CODE,
 } from '../../util/keycodes';
 import { InputType } from './input-field.types';
+import { ListItem } from '@limetech/lime-elements';
 
 @Component({
     tag: 'limel-input-field',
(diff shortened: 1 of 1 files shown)


src/components/input-field/input-field.tsx
```diff
@@ -408,46 +421,48 @@ export class InputField {
         );
     }
 
-    private renderDropdown() {
-        const filteredCompletions = this.filterCompletions(this.value);
+    private onKeyDown() {
+        this.showCompletions = true;
```
BregenzerK, line 425: currently you have to tab to select the list first before you can use the arrow key. wouldn't it be more natural to select the first item already by pressing `DOWN`? also if you tab twice the next input gets selected but the list stays open. check the `handleInputKeyDown` method in the picker, I think we need something similar here

src/components/input-field/input-field.tsx
```diff
@@ -408,46 +421,48 @@ export class InputField {
...
+    }
+
+    private handleCompletionChange(event: CustomEvent<ListItem>) {
+        this.showCompletions = false;
+        event.stopPropagation();
+        if (!event.detail) {
+            return;
+        }
+
+        this.showCompletions = false;
```
BregenzerK, line 435: the input should still be on focus after you selected something, shouldn't it?

src/components/input-field/input-field.tsx
adrianschmidt: The behaviour for this should match the behaviour for similar elements.

I was going to write "like limel-menu" and then realised that limel-menu doesn't close if you tab away from it while it's open 😞

Then I checked limel-select, and that works pretty well, except that it loses focus when tabbing out of it (at least sometimes), so the focus ends up in the address bar instead of on the next element.

But, generally, we don't want to open the menu by tabbing, because that will make it a horrible experience to tab past a number of these elements. Instead, `arrow-up` and `arrow-down` should typically open the menu, and probably `enter` should too. Tab should tab away from the element, and if the menu is open when tabbing away, the menu should close.

adrianschmidt: Note also that on macOS, `arrow-up` and `arrow-down` should normally place the caret (the text input marker) at the beginning and end of the text, respectively. So we might need to consider how that works together with the menu. We probably don't want to prevent that default behaviour, at least not when the menu isn't open… that's less important than getting the tabbing behaviour right though 🙂

src/components/input-field/input-field.tsx
```diff
@@ -436,6 +488,14 @@ export class InputField {
         this.change.emit(event.detail.text);
     }
 
+    onKeyDownForList(event: KeyboardEvent) {
```
BregenzerK, line 491: `private` missing

src/components/input-field/input-field.tsx
```diff
@@ -421,12 +429,56 @@ export class InputField {
...
+        const isForwardTab =
+            (event.key === TAB || event.keyCode === TAB_KEY_CODE) &&
+            !event.altKey &&
+            !event.metaKey &&
+            !event.shiftKey;
+        const isUp =
+            event.key === ARROW_UP || event.keyCode === ARROW_UP_KEY_CODE;
+        const isDown =
+            event.key === ARROW_DOWN || event.keyCode === ARROW_DOWN_KEY_CODE;
+
```
BregenzerK, line 452: when the focus is on the input and you tab backwards the list stays visible

PR #700, opened 2020-04-16
ci(lint): add linter for scss files

adrianschmidt:
`prettier`, which we already have installed, can beautify scss files too 🥳

## Review:


package.json
```diff
@@ -32,6 +32,8 @@
     "lint:specs:fix": "npx tslint --project . -t stylish --fix -c tslint.specs.js 'src/**/*.{spec,e2e}.{ts,tsx}'",
     "lint:src": "npx tslint --project . -t stylish -c tslint.js 'src/**/*.{ts,tsx}'",
     "lint:src:fix": "npx tslint --project . -t stylish --fix -c tslint.js 'src/**/*.{ts,tsx}'",
+    "lint:scss": "prettier -c **/*.scss",
+    "lint:scss:fix": "prettier --write **/*.scss",
```
adrianschmidt, line 36: When running this, it will rewrite scss files inside the `.docz` folder too, but if I change to `src/**/*.scss` it doesn't run on the files in the `src/components/` folder for some reason I really can't understand.

@jgroth Do you have any idea?

Anyone else?

BregenzerK: have you tried to ignore the `.docz` path/ add a `.prettierignore`? https://prettier.io/docs/en/cli.html#--ignore-path

adrianschmidt: Ah, of course! 🙏

adrianschmidt: Adding `--ignore-path ./.gitignore` worked, and seems like quite a silvery bullet to me 😁

src/style/color-palette.scss
```diff
@@ -1,271 +1,271 @@
...
     --color-light-blue-darker: #01579b;
 
-    --color-cyan-lighter: #80deea;
-    --color-cyan-light: #26c6da;
-    --color-cyan-default: #00acc1;
-    --color-cyan-dark: #00838f;
+    --color-cyan-lighter: #b2ebf2;
+    --color-cyan-light: #4dd0e1;
+    --color-cyan-default: #00bcd4;
+    --color-cyan-dark: #0097a7;
```
adrianschmidt, line 69: This looks alarming, but it's actually just git matching the light mode of the updated file to the dark mode of the old file, because the indentation changed… 😅

PR #721, opened 2020-04-22
Action menu on list items (again)

jensgustafsson:
Connected to: Lundalogik/crm-feature#989

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [x] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu/menu.tsx
+++ b/src/components/menu/menu.tsx
@@ -1,16 +1,7 @@
 import { ListItem, ListSeparator } from '@limetech/lime-elements';
-import { Corner, MDCMenu } from '@limetech/mdc-menu';
-import {
-    Component,
-    Element,
-    Event,
-    EventEmitter,
-    h,
-    Prop,
-    Watch,
-} from '@stencil/core';
-import { ListRenderer } from '../list/list-renderer';
-import { ListRendererConfig } from '../list/list-renderer-config';
+import { Component, Event, EventEmitter, h, Prop } from '@stencil/core';
+import { createRandomString } from '../../util/random-string';
+import { OpenDirection } from './menu.types';
 
 /**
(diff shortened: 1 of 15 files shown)


src/components/menu/menu.tsx
```diff
@@ -124,28 +99,33 @@ export class Menu {
...
     private onTriggerClick = () => {
+        if (this.disabled) {
+            return;
+        }
         this.open = !this.open;
     };
+
+    private onListChange = (event) => {
+        this.items = this.items.map((item: ListItem) => {
+            if (item.value === event.detail.value) {
```
jensgustafsson, line 122: I think value has to be a primitive object in order for this to work

jgroth: We should just have to compare the items I guess :thinking: 
```suggestion
            if (item === event.detail) {
```

jensgustafsson: Then I could probably remove most of my changes in the web client :)

jgroth: :smile:

src/components/portal/portal.tsx
```diff
@@ -101,6 +102,7 @@ export class Portal {
         const content = slot.assignedElements();
 
         this.container = document.createElement('div');
+        this.container.style.opacity = '0';
```
jgroth, line 105: Super minor, but since we have a `showContainer`, we could have a `hideContainer` as well :smile:

jensgustafsson: Yea thats true! I will change that :)

PR #723, opened 2020-04-22
fix: prevent argument mismatching in case of value being an array

BregenzerK:
re Lundalogik/crm-feature#1057

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/templates/array-field-collapsible-item.ts
+++ b/src/components/form/templates/array-field-collapsible-item.ts
@@ -38,6 +38,8 @@ export class CollapsibleItemTemplate extends React.Component {
     constructor(public props: CollapsibleItemProps) {
         super(props);
         this.handleAction = this.handleAction.bind(this);
+        this.isDeepEmpty = this.isDeepEmpty.bind(this);
+
         this.isOpen = this.isDeepEmpty(props.data);
     }
 


src/components/form/templates/array-field-collapsible-item.ts
```diff
@@ -38,6 +38,8 @@ export class CollapsibleItemTemplate extends React.Component {
     constructor(public props: CollapsibleItemProps) {
         super(props);
         this.handleAction = this.handleAction.bind(this);
+        this.isDeepEmpty = this.isDeepEmpty.bind(this);
```
adrianschmidt, line 41: This accomplishes the same thing, and doesn't run the same risk of someone (probably future me) accidentally reverting the change in an attempt to "clean up" 😄

src/components/form/templates/array-field-collapsible-item.ts
```diff
@@ -38,6 +38,8 @@ export class CollapsibleItemTemplate extends React.Component {
     constructor(public props: CollapsibleItemProps) {
```
adrianschmidt, line 38: For the changelog, it would be helpful to have a scope in the commit message:

`fix(form): prevent argument mismatching in case of value being an array`

src/components/form/templates/array-field-collapsible-item.ts
BregenzerK: seriously? :laughing: I thought about that. but then why is the recursive call working as long as it's not array of objects?

adrianschmidt: I would assume because we're in `array-field-collapsible-item.ts`, so this code isn't used unless we're working with an array? Not sure though, I haven't investigated any further.

(And yes, my first reaction when I saw your change was "WAT? How is this changing anything?" — which is also precisely why I foresee my future self coming across that and trying to "clean it up", thereby breaking it again 😂)

BregenzerK: not really, `array-field-collapsible-item` is always only one item (the array gets handle in `array-field`). for section though we have an array of controls (so a collapsible in a collapsible) and even that works as long as `isDeepEmpty` finds a value before reaching the controls array. but when you add a new section it has to check all of it and that's when it failed

adrianschmidt: I have no idea then 😄

Good find either way! 👏👏

Issue #725, opened 2020-04-23
Limel-menus also get the new striped styles from lists


Kiarokh:
The "even/odd" styling and hover effects of list items `<li>`s are not necessary for menus, but since limel-menus are using limel-lists, they are being affected by it. Solution is adding a custom class which can be used optionally only on lists that need such a style.

PR #726, opened 2020-04-23
New props for lists that need striped styles

Kiarokh:
Fix: Lundalogik/lime-elements#725

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/list/list.mdx
+++ b/src/components/list/list.mdx
@@ -8,23 +8,32 @@ menu: Components
 
 <limel-props name="limel-list" />
 
+## Custom styles
+
+| Class name              | Description                                                                                     |
+| ----------------------- | ----------------------------------------------------------------------------------------------- |
+| `has-striped-rows`      | Adds distinct styling by which every other row (list item) gets a darker background.            |
+| `has-interactive-items` | Adds more affordance by applying an elevated visual effect on list item, when they are hovered. |
+
+
+
 ## Examples
 
 ### Basic list with separator
 
-<limel-example name="limel-example-list" /> 
+<limel-example name="limel-example-list" />
 
 ### List with secondary text
 
-<limel-example name="limel-example-list-secondary" path="list" /> 
+<limel-example name="limel-example-list-secondary" path="list" />
 
 ### List with selectable items
 
-<limel-example name="limel-example-list-selectable" path="list" /> 
+<limel-example name="limel-example-list-selectable" path="list" />
 
 ### List with icons
 
-<limel-example name="limel-example-list-icons" path="list" /> 
+<limel-example name="limel-example-list-icons" path="list" />
 
 ### List with badge icons
 
@@ -49,3 +58,12 @@ menu: Components
 ### List with action menu
 
 <limel-example name="limel-example-list-action" path="list" />
+
+### List with custom styles
+Adding the `has-striped-rows` class to the list will make the items more distinct by adding different background colors to even and odd rows.
+
+Also, by taking advantage of the `has-interactive-items`, hovering on a list item which is not `disabled` will display an elevated visual effect, giving it more affordance and a solid feeling of interactivity.
+
+**Note:** to get both effects, you need to apply both of these classes.
+
+<limel-example name="limel-example-list-striped" path="list" />
--- a/src/components/list/examples/list-striped.tsx
+++ b/src/components/list/examples/list-striped.tsx
@@ -1,0 +1,30 @@
+import { ListItem } from '@limetech/lime-elements';
+import { Component, h } from '@stencil/core';
+
+@Component({
+    tag: 'limel-example-list-striped',
+    shadow: true,
+})
+export class StripedListExample {
+    private items: Array<ListItem<number>> = [
+        {
+            text: 'King of Tokyo',
+            secondaryText: '2-6 players',
+            value: 1,
+            disabled: true,
+        },
+        { text: 'Smash Up!', secondaryText: '2-4 players', value: 2 },
+        { text: 'Pandemic', secondaryText: '2-4 players', value: 3 },
+        { text: 'Catan', secondaryText: '3-4 players', value: 4 },
+        { text: 'Ticket to Ride', secondaryText: '2-5 players', value: 5 },
+    ];
+
+    public render() {
+        return (
+            <limel-list
+                items={this.items}
+                class="has-striped-rows has-interactive-items"
+            />
+        );
+    }
+}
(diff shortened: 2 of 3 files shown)


src/components/list/list.mdx
```diff
@@ -8,23 +8,31 @@ menu: Components
 
 <limel-props name="limel-list" />
 
+## Custom styles
+
+| Class name | Description                                                                                     |
+| -------- | -------------------------------------------------------------------- |
+| `has-striped-rows` | Adds distinct styling and more interactivity to list items |
```
jgroth, line 15: Do you think we need  `has` in the name? :thinking: 
Maybe `striped-rows` is enough?

jgroth, line 15: The description is very generic, maybe make it a bit more specific like "Adds every other row a darker background" or something like that. What do you think?

jgroth, line 15: The class name also does not indicate that it has any effect on the hovering. Should it still be in the same class or do you think it is a good use case to separate them?

Kiarokh: Yes I think having such words makes the classes more "descriptive" and easier to understand, specially for people who don't know our intentions of having such a class. Personally I'd recommend having them. But open for discussion.

some examples for this naming convention:

`dark` vs. `has-dark-background`
`disabled` vs. `is-disabled`
`dimmed` vs. `gets-dimmed-while-loading`
`hover-zoom` vs. `gets-zoomed-on-hover`

Kiarokh: Being short, abstract and concise is sometimes not that good. In these cases I think descriptive class names makes the code more understandable.

Kiarokh: I like your suggestion. 👌will you add it or should I do some fixups?

Kiarokh: There may be a point in separating them actually, and this will increase our customization possibilities. On the other hand, I couldn't imagine a case where only one is needed, not the other one. But it can happen. Not really sure how to decide here. What do you think?

jgroth: For me it does not have any meaning in this case and is just clutter :smile: Since you have added the class you would expect it to be applied. We could flip the table and look at it another way, having a class named `has-not-striped-rows` does not make much sense, but is almost the same thing
We are all different I suppose but this is my view on it, you can keep it or change it if you want :slightly_smiling_face:

jgroth: You can do it :slightly_smiling_face:

jgroth: Hmm I'm not sure :thinking: 
I think we should probably change the class name if we keep them together, but I'm not sure what the name should be then

adrianschmidt: `dark` vs. `has-dark-background`

I think `dark-background` is more informative than `dark`, but I don't think `has-dark-background` is more informative than `dark-background`. The "has" is conveyed by the fact that the class is applied.

`disabled` vs. `is-disabled`

Same thing here. If it's not disabled, then the class just isn't there.

The last two examples have more merit, in my opinion, but I would skip the `gets-` for the same reason as above.

`dimmed-while-loading` and `zoomed-on-hover` are expressive, but not unnecessarily verbose, in my opinion 🙂

Kiarokh: I understand what you mean, and I'm not trying to defend the class name. Mainly want to pick your brain on the subject. 

Now my examples weren't so well-thought 😛. It takes some time to find good examples. But imagine you see this imaginary code:

`<button class="lime-button is-in-top-header has-accented-background-color has-icon-left gets-fancy-effect-on-hover" ····`
versus
`<button class="lime-button top-header accented-background icon-left fancy-effect" ····`

For me personally, it feels I'm reading a sentence that describes how the button would look and behave like. I can more quickly build an assumption around the styles and expect things. Also expect for example that the creators have a solid idea regarding buttons which are located in the top header -they must be behaving or looking differently compared to similar buttons-; or there is a visual effect that only gets applied on hover not in other states - so if I want to inspect it I must hover. And so forth... In the minimal example, you almost have to find the classes and read their respective styles to build this understanding.

I don't know...., I kindda like this type of naming convention more, as a UX designer. But I totally also understand your points, and know that abbreviations and short-codes and so forth are always more desired in coding mindset. And of course I'm aware that too long words make reading the sentence also too hard eventually.

I also think the same way for class-name conventions that describe the structure. For example seeing this class `container` will immediately 😵💀me. I would wonder container of what you ☀️of a 🏖?! 😄 Why don't you specify more? ’button-group-container`? `top-menu-container`? can't you put something that is easy to track/understand/target? 😜

Anyhow, I have no good bullet-proof argument or naming convention to recommend. Open for suggestions, change, new conventions, anything you guys say.

Kiarokh: `has-dark-background`
`gets-dark-background-when-focused`
`applies-dark-background-to-something`
`buys-dark-background-for-dinner-😁`

adrianschmidt: Yeah, good points I think.

> abbreviations and short-codes and so forth are always more desired in coding mindset

I'm actually mostly against using abbreviations. They should only be used when it helps the *reader*, not the *writer*.

I'm fine with either `has-striped-rows` or `striped-rows`. I didn't want to but in on the decision, just add my thoughts to the conversation 👍🙂

adrianschmidt: I think they should be split. I imagine not all lists will have items that do something when clicked, and then I assume they shouldn't zoom when hovered? But you might still want to have a striped list, I imagine. So I'd go with something like `has-striped-rows` and `has-interactive-items`

Kiarokh: Totally right. Abbreviations should be forbidden, unless everybody knows what you are talking about, e.g. SMS or Covid-19. Nobody knows what it stands for, but everyone know what it is. 

As you beautifully said, it helps the reader!

A bit more thoughts on the topic...
Since a class name such as `striped-rows` can be also perceived as "being" something (being striped rows [maybe striped rows is a component in itself, for instance `is-striped-rows`]), it's not favorable. 

But `has-striped-rows` automatically includes and expresses a few things about the component: 
1: whatever this thing is, it has rows
2: and the rows are striped
3-ish: it probably can have rows which aren't striped too

So, I get a feeling that we're almost on the same side. I suggest sticking to this convention :)

Kiarokh:
I didn't do any fixups, rather pushed a few updates. But don't know if everything went as expected. Do you see the recent changes?

PR #737, opened 2020-04-28
fix(picker): ensure picker doesn't crash if its value is set to null

jensgustafsson:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -203,6 +203,8 @@ export class Picker {
     @Watch('value')
     protected onChangeValue(newValue = [], oldValue = []) {
         this.chips = this.createChips(this.value);
+        newValue = newValue || [];
+        oldValue = oldValue || [];
         if (newValue.length <= oldValue.length) {
             return;
         }
--- a/src/util/dom.ts
+++ b/src/util/dom.ts
@@ -7,6 +7,10 @@
  * @returns {boolean} true if the element is a descendant of the parent element, false otherwise
  */
 export function isDescendant(element: Node, parent: Node) {
+    if (!parent) {
+        return false;
+    }
+
     if (parent.contains(element)) {
         return true;
     }


src/components/picker/picker.tsx
```diff
@@ -203,6 +203,8 @@ export class Picker {
     @Watch('value')
     protected onChangeValue(newValue = [], oldValue = []) {
         this.chips = this.createChips(this.value);
+        newValue = newValue || [];
```
Befkadu1, line 206: newValue = [] and oldValue = [] by default. Right? I think we don't have to set it again.

Befkadu1:
Still picker crashes when portalElement=null at line 378. It crashes here, `parent.contains(element)` when `parent` is null in dom.ts file. We should make sure that `parent` not null.

Screenshot 2020-04-29 at 17 44 10

src/components/picker/picker.tsx
jensgustafsson: Unfortunately not. If `null` is provided as an argument then `newValue` will be `null` and not `[]`

jensgustafsson:
> Still picker crashes when `portalElement=null` at `isDescendant(element as HTMLElement, portalElement as HTMLElement)` around line 378 . It crashes here, `parent.contains(element)` when `parent` is null in dom.ts file. We should make sure that `parent` is not null.
> 
> Screenshot 2020-04-29 at 17 44 10

Maybe you can show me remotely tomorrow?

Befkadu1:
> > Still picker crashes when `portalElement=null` at `isDescendant(element as HTMLElement, portalElement as HTMLElement)` around line 378 . It crashes here, `parent.contains(element)` when `parent` is null in dom.ts file. We should make sure that `parent` is not null.
> > Screenshot 2020-04-29 at 17 44 10
> 
> Maybe you can show me remotely tomorrow?

Ok

PR #738, opened 2020-04-29
Fix focus chipset

hannahu:
Fix Lundalogik/crm-feature#1019

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -201,15 +201,8 @@ export class Picker {
     }
 
     @Watch('value')
-    protected onChangeValue(newValue = [], oldValue = []) {
+    protected onChangeValue() {
         this.chips = this.createChips(this.value);
-        newValue = newValue || [];
-        oldValue = oldValue || [];
-        if (newValue.length <= oldValue.length) {
-            return;
-        }
-
-        this.chipSet?.setFocus(true);
     }
 
     @Watch('searcher')
@@ -423,6 +416,10 @@ export class Picker {
             this.change.emit(newValue);
             this.items = [];
         }
+
+        if (this.multiple) {
+            this.chipSet?.setFocus(true);
+        }
     }
 
     /**
@@ -449,6 +446,10 @@ export class Picker {
                     return `${item.value}` === chip.id;
                 });
             });
+        }
+
+        if (!this.multiple && !newValue) {
+            this.chipSet?.setFocus(true);
         }
 
         this.change.emit(newValue);


src/components/picker/picker.tsx
```diff
@@ -203,7 +203,12 @@ export class Picker {
     @Watch('value')
     protected onChangeValue(newValue = [], oldValue = []) {
         this.chips = this.createChips(this.value);
-        if (newValue.length <= oldValue.length) {
+
+        if (!this.multiple) {
```
jgroth, line 207: This won't work, we would still have the bug for pickers that have `multiple` set to true.

We have to fix that we do not set focus on the element when the value is changed. We want to do it when selecting an item from the dropdown I guess, but not _everytime_ the value is changed :dizzy_face:

src/components/chip-set/chip-set.tsx
```diff
@@ -323,7 +323,9 @@ export class ChipSet {
                         'mdc-text-field--disabled':
                             this.readonly || this.disabled,
                         'mdc-text-field--required': this.required,
-                        'force-float': !!(this.value.length || this.editMode),
+                        'mdc-floating-label--float-above': !!(
+                            this.value.length || this.editMode
+                        ),
```
jgroth, line 328: I think this might also mess stuff up. When the picker has a value the label looks fine. If I focus the element and then blur it again, the label will overlap with the chip

hannahu: > I think this might also mess stuff up. When the picker has a value the label looks fine. If I focus the element and then blur it again, the label will overlap with the chip

Ahh thought I tested everything but yes you're right...

src/components/picker/picker.tsx
hannahu: Let's see if you're happy with this solution :D
- The element won't get focus when the chipSet isn't defined (from start)
- The new card component in the webclient we've never used messed some stuff up and therefore I'm removing it

src/components/picker/picker.tsx
```diff
@@ -436,7 +435,7 @@ export class Picker {
         this.handleSearchResult(query, result);
     }
 
-    private handleChange(event: CustomEvent<Chip | Chip[]>) {
+    private handleRemoveChip(event: CustomEvent<Chip | Chip[]>) {
```
jgroth, line 438: This does not make sense to rename :smile: 
This will be called when a chip is added as well, `handleChangeChip` would be better name if you want to rename it :slightly_smiling_face:

PR #743, opened 2020-05-03
Use popperjs in limel-portal for positioning

jensgustafsson:
resolves https://github.com/Lundalogik/crm-feature/issues/1069

Things to do before merge:

- [ ] Remove the commit that changes the example code. (See commit history)
- [ ] Move z-index hack to a separate class in the webclient (since we dont understand why its needed and it doesn't seem to be needed in lime-elements it self.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/portal/portal.tsx
+++ b/src/components/portal/portal.tsx
@@ -1,5 +1,7 @@
 import { Component, Element, h, Prop, Watch } from '@stencil/core';
 import { OpenDirection } from '../menu/menu.types';
+import { createPopper, Instance, OptionsGeneric } from '@popperjs/core';
+import { FlipModifier } from '@popperjs/core/lib/modifiers/flip';
 
 /**
  * The portal component provides a way to render children into a DOM node that exist outside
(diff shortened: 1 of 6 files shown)


src/components/portal/portal.tsx
```diff
@@ -128,50 +135,28 @@ export class Portal {
...
-
-        this.container.style.top = `${y}px`;
-        this.container.style.width =
-            rect.width > 0 ? `${rect.width}px` : `${this.width}px`;
         this.container.style.height = `${containerHeight}px`;
         this.container.style.display = 'block';
         if (!this.visible) {
             this.container.style.display = 'none';
         }
+        this.container.style.zIndex = '7'; // Required for rendering elements above the AddDialog in the webclient
```
adrianschmidt, line 149: We could maybe set this to some reasonable default, and provide a way to override it using a property?

adrianschmidt: Ah, or, probably, this would make it overridable, without having to expose a property on both limel-portal and the component using it…

```suggestion
        this.container.style.zIndex = 'var(--dropdown-z-index, 7)'; // Required for rendering elements above the AddDialog in the webclient
```

jgroth:
I think the solution to this problem is using the Intersection Observer API, but maybe thats out of scope for this PR
https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API
scroll

adrianschmidt:
> [developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API)

# 😮🤩🥳

Issue #747, opened 2020-05-05
make input fields backgrounds transparent


Kiarokh:
![image](https://user-images.githubusercontent.com/35954987/81065809-74b21a80-8edc-11ea-96f0-064616430ce6.png)

The input fields that are also used in pickers and chipsets come with a white background. This messes up styling when the background of their containers already has a color. For now, we make everything transparent, until we have edit mode & view mode

PR #749, opened 2020-05-05
Transparent input field background

Kiarokh:
fix: https://github.com/Lundalogik/lime-elements/issues/747

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -287,7 +287,7 @@ export class ChipSet {
 
     private renderInputChips() {
         return (
-            <div
+            <label
                 class={{
                     'mdc-text-field': true,
                     'mdc-text-field--disabled': this.readonly || this.disabled,
@@ -330,7 +330,7 @@ export class ChipSet {
                     {this.label}
                 </label>
                 <div class="mdc-line-ripple" />
-            </div>
+            </label>
         );
     }
 
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -287,7 +287,7 @@ export class ChipSet {
 
     private renderInputChips() {
         return (
-            <label
+            <div
                 class={{
                     'mdc-text-field': true,
                     'mdc-text-field--disabled': this.readonly || this.disabled,
@@ -330,7 +330,7 @@ export class ChipSet {
                     {this.label}
                 </label>
                 <div class="mdc-line-ripple" />
-            </label>
+            </div>
         );
     }
 
--- a/src/components/input-field/input-field.scss
+++ b/src/components/input-field/input-field.scss
@@ -35,7 +35,8 @@ $mdc-text-field-fullwidth-bottom-line-color: $lime-text-field-bottom-line-color;
     @include mdc-states-focus-opacity(0, true);
 }
 
-label {
+label,
+div {
     &.mdc-text-field {
         transition: background-color 0.2s ease;
 
--- a/src/components/input-field/input-field.scss
+++ b/src/components/input-field/input-field.scss
@@ -35,6 +35,29 @@ $mdc-text-field-fullwidth-bottom-line-color: $lime-text-field-bottom-line-color;
     @include mdc-states-focus-opacity(0, true);
 }
 
+label {
+    &.mdc-text-field {
+        transition: background-color 0.2s ease;
+
+        &:not(.mdc-text-field--disabled) {
+            // this has to be written this way to be able to override MD styles
+            background-color: transparent;
+
+            &:hover {
+                background-color: #f1f1f3;
+            }
+        }
+
+        &.mdc-text-field--focused {
+            background-color: #f6f6f7;
+        }
+
+        &.mdc-text-field--disabled {
+            background-color: transparent;
+        }
+    }
+}
+
 .mdc-text-field__formatted_input {
     display: None;
 }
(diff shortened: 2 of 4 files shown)


Befkadu1:
The 1st commit type doesn't seem correct. Probably `fix`

Kiarokh:
@Befkadu1 the background color doesn't fix anything or adds a new feature. It's just inside the documentations itself for preview purposes only. So next time anyone adds a component that has a background color, they will notice it easier.
image

adrianschmidt:
> > The 1st commit type doesn't seem correct. Probably `fix`
>
> Befkadu1 the background color doesn't fix anything or adds a new feature. It's just inside the documentations itself for preview purposes only. So next time anyone adds a component that has a background color, they will notice it easier.

Since it doesn't change anything in the released code, you are correct that it should not be `fix` or `feat` 👍

I think the best type for that commit would be `docs` 🙂

src/components/input-field/input-field.scss
```diff
@@ -35,6 +35,29 @@ $mdc-text-field-fullwidth-bottom-line-color: $lime-text-field-bottom-line-color;
...
 
+label {
+    &.mdc-text-field {
+        transition: background-color 0.2s ease;
+
+        &:not(.mdc-text-field--disabled) {
+            // this has to be written this way to be able to override MD styles
+            background-color: transparent;
+
+            &:hover {
```
Befkadu1, line 46: This hover effect is implemented only in the limel-chip-set. Should we need to implement the hover effect in the other fields as well?

src/components/chip-set/chip-set.tsx
```diff
@@ -330,7 +330,7 @@ export class ChipSet {
                     {this.label}
                 </label>
                 <div class="mdc-line-ripple" />
-            </div>
+            </label>
```
adrianschmidt, line 333: Why change this to a `label`? There is already a label on line 320, which is inside this, and I'm not sure it's valid to nest label elements. This label would also include all the chips, and their respective remove-buttons, something [MDN says we shouldn't do](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label#Accessibility_concerns).

adrianschmidt: Ah, just read the commit message:

> This makes the HTML structure consistent, like mdc-text-field. They are also wrapped in <label>s. This way we can also apply the same styles to both.

I think it's more important to create well formed and accessible html than to make styling more convenient for us.

src/components/input-field/input-field.scss
```diff
@@ -35,6 +35,29 @@ $mdc-text-field-fullwidth-bottom-line-color: $lime-text-field-bottom-line-color;
...
+                background-color: #f1f1f3;
```
adrianschmidt, line 47: If the point of this PR is to make input fields look good on backgrounds other than white, we can't use a solid hover colour.

Befkadu1:
> @Befkadu1 the background color doesn't fix anything or adds a new feature. It's just inside the documentations itself for preview purposes only. So next time anyone adds a component that has a background color, they will notice it easier.
> image

What I mean is I would suggest the commit message to be `docs(example): give a slight grey background color to example areas`.  What do you think?

src/components/input-field/input-field.scss
Kiarokh: It should be implemented in input-field which is also used in the chipset. I could see it in both places when I tested. Sounds strange that you don't see it. R U sure?

src/components/chip-set/chip-set.tsx
Kiarokh: Do you recommend to change it back to div? @adrianschmidt

adrianschmidt: In the absence of any real recommendation, other than "don't make it a label", I guess so 😄

src/components/input-field/input-field.scss
Kiarokh: You're absolutely right about not having solid colors. But colors aren't things we have really decided yet. At some point of time, we need to go through all colors and start a reform which takes into account the dark mode as well. It's gonna be a very complex task and requires a lot of decisions and tests and checks. First thing to do there is to decide for a couple of color palettes, and where and how to use them. We're having a separate discussion about them in the redesign and marketing teams right now.

The point of this PR is to remove the initial background color from the input fields (when they're not focused or active) so that we can have them on backgrounds with other colors.

For active and hover states, we need other good conventions (even perhaps shared with other elements such as radio buttons and checkboxes). For now it's simply too abstract to think about and design for. Especially because there are other indicators (the accented bottom border) to show where a field is focused or not. So I think we can let go of it

Kiarokh:
regarding `docs` or `fix`, let me know what you decide. I'll do the fixup

src/components/input-field/input-field.scss
adrianschmidt: Ok, as long as you've made a conscious decision, and didn't just overlook it, I'm happy 🙂

adrianschmidt:
> regarding docs or fix, let me know what you decide. I'll do the fixup

Bef and I agree, use `docs` 🙂

PR #758, opened 2020-05-12
feat(tab-bar): add buttons when bar can be scrolled 

adrianschmidt:
fix Lundalogik/crm-feature#1022

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tab-bar/tab-bar.scss
+++ b/src/components/tab-bar/tab-bar.scss
@@ -10,6 +10,7 @@ $tab-border-radius: 0.65rem;
 $tab-active-outer-edge-curve-size: 0.75rem;
 $tab-separator-width: 0.125rem;
 $tab-separator-background-color: #e8e8ea;
+$tab-scroller-fade-width: 65;
 
 :host {
     display: block;
@@ -22,6 +23,70 @@ $tab-separator-background-color: #e8e8ea;
 .mdc-tab-indicator {
     .mdc-tab-indicator__content {
         border: none;
+    }
+}
+
+.mdc-tab-scroller {
+    position: relative;
+    overflow: hidden;
+
+    .scroll-fade,
+    .scroll-button {
+        position: absolute;
+        transition-property: transform;
+        transition-duration: 250ms;
+        transition-timing-function: ease-out;
+    }
+
+    .scroll-fade {
+        top: 0;
+        height: 100%;
+        width: pxToRem($tab-scroller-fade-width);
+        pointer-events: none;
+
+        &.left {
+            transform: translate3d(pxToRem(0 - $tab-scroller-fade-width), 0, 0);
+            left: 0;
+            background: linear-gradient(
+                270deg,
+                rgba(246, 246, 247, 0) 0%,
+                rgba(246, 246, 247, 0.8) 40%,
+                rgba(246, 246, 247, 0.8) 100%
+            );
+        }
+
+        &.right {
+            transform: translate3d(pxToRem($tab-scroller-fade-width), 0, 0);
+            right: 0;
+            background: linear-gradient(
+                90deg,
+                rgba(246, 246, 247, 0) 0%,
+                rgba(246, 246, 247, 0.8) 40%,
+                rgba(246, 246, 247, 0.8) 100%
+            );
+        }
+    }
+
+    .scroll-button {
+        --icon-background-color: #ffffff;
+        top: pxToRem(4);
+
+        &.left {
+            transform: translate3d(pxToRem(0 - $tab-scroller-fade-width), 0, 0);
+            left: pxToRem(4);
+        }
+
+        &.right {
+            transform: translate3d(pxToRem($tab-scroller-fade-width), 0, 0);
+            right: pxToRem(4);
+        }
+    }
+
+    &.scroll-left .scroll-fade.left,
+    &.scroll-left .scroll-button.left,
+    &.scroll-right .scroll-fade.right,
+    &.scroll-right .scroll-button.right {
+        transform: translate3d(0, 0, 0);
     }
 }
 
--- a/src/components/tab-bar/tab-bar.scss
+++ b/src/components/tab-bar/tab-bar.scss
@@ -6,9 +6,9 @@
 
 $tab-content-background-color: #fff;
 $tab-background-color: #f6f6f7;
-$tab-border-radius: 0.65rem;
-$tab-active-outer-edge-curve-size: 0.75rem;
-$tab-separator-width: 0.125rem;
+$tab-border-radius: pxToRem(10);
+$tab-active-outer-edge-curve-size: pxToRem(12);
+$tab-separator-width: pxToRem(2);
 $tab-separator-background-color: #e8e8ea;
 $tab-scroller-fade-width: 65;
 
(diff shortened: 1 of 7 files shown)


hannahu:
Nice!!
Some notes/comments (I know it's a draft):
- From start a scrollbar is shown and no "shadow"
- When reaching the end, a scrollbar is shown and there's an extra space to the right
- Sometimes it's hard to see/understand that there's more content, maybe some less fading will do it

jensgustafsson:
I think I found an issue with it :/

See video below:


issue

hannahu:
> I think I found an issue with it :/
> 
> See video below:
> 
> issue
> 
> The tabs after "Judy" is not shown if you "scroll" with the scrollbar.

I don't think there should be a scrollbar at all 🤔

adrianschmidt:
> > I think I found an issue with it :/
> >
> > The tabs after "Judy" is not shown if you "scroll" with the scrollbar.
> 
> I don't think there should be a scrollbar at all 🤔
> 
> Then that's another issue we need to solve :P

Haha! No, there shouldn't be a scrollbar 😂

I missed it was there because my Mac doesn't show it if I don't scroll in that container, and that's not the container normally being scrolled… but as I said on standup, my wife managed to find it by scrolling in precisely the right spot to hit the 1px high line that triggers it 😆

I'm pretty sure I know what the problem is 👍

src/components/tab-bar/tab-bar.tsx
```diff
@@ -87,9 +94,11 @@ export class TabBar {
         };
 
         this.setupListeners();
+        this.handleScroll();
```
adrianschmidt, line 97: Is it clear enough why we'd want to do this here, or should I add a comment? (This is in the `setup` function. Expand the code to see the context.)

src/components/tab-bar/tab-bar.scss
```diff
@@ -25,6 +25,46 @@ $tab-separator-background-color: #e8e8ea;
     }
 }
 
+.mdc-tab-scroller {
+    position: relative;
+    overflow: hidden;
```
adrianschmidt, line 30: > When reaching the end, a scrollbar is shown and there's an extra space to the right

⚡

src/components/tab-bar/tab-bar.tsx
adrianschmidt, line 97: > From start a scrollbar is shown and no "shadow"

⚡

src/components/tab-bar/tab-bar.scss
```diff
@@ -25,6 +25,46 @@ $tab-separator-background-color: #e8e8ea;
...
+        pointer-events: none;
+        transition-property: transform;
+        transition-duration: 250ms;
+        transition-timing-function: ease-out;
+    }
+
+    &::before {
+        transform: translate3d(-10rem, 0, 0);
+        left: 0;
+        background: linear-gradient(270deg, rgba(246,246,247,0) 0%, rgba(246,246,247,.9) 80%, rgba(246,246,247,.9) 100%);
```
adrianschmidt, line 50: > Sometimes it's hard to see/understand that there's more content, maybe some less fading will do it

That's always a bit of a problem when using fading. Using buttons with arrows instead, like @jensgustafsson suggested, ought to solve that problem though.

adrianschmidt:
@hannahu @jensgustafsson I made some updates. Please test again. I'll start looking at using buttons instead of a fade, but the scrolling issues still applies, so please test those.

src/components/tab-bar/tab-bar.tsx
```diff
@@ -113,6 +123,25 @@ export class TabBar {
...
+            this.scrollerElement.classList.add('scroll-left');
+        } else {
+            this.scrollerElement.classList.remove('scroll-left');
+        }
+
+        if (scrollRight > 40) {
+            this.scrollerElement.classList.add('scroll-right');
+        } else {
+            this.scrollerElement.classList.remove('scroll-right');
+        }
```
adrianschmidt, line 142: This circumvents Stencil's render-loop. I'm not sure if we want to do that, or if we would rather have these as State-properties, which trigger a re-render? Using State-props would be more consistent, but I'd be surprised if doing it this way wasn't more performant…

Thoughts?

src/components/tab-bar/tab-bar.scss
jensgustafsson: I think we should try buttons. I think the fading looks really great but unfortunately that doesn't make me understand that there is more tabs if I scroll

src/components/tab-bar/tab-bar.tsx
jensgustafsson: I think its clear enough :)

src/components/tab-bar/tab-bar.tsx
jensgustafsson: I think it's okay but wouldn't mind a second opinion from @jgroth

src/components/tab-bar/tab-bar.scss
adrianschmidt: I'm working it now 🙂

@Kiarokh posted this design on the Issue, which I'm going to see if I can pull off:

image

src/components/tab-bar/tab-bar.tsx
```diff
@@ -100,6 +109,7 @@ export class TabBar {
 
     private setupListeners() {
         this.mdcTabBar.listen(TAB_ACTIVATED_EVENT, this.handleTabActivated);
+        this.scrollArea.addEventListener('scroll', this.handleScroll);
```
jgroth, line 112: We should probably make this listener passive
https://developers.google.com/web/updates/2016/06/passive-event-listeners

adrianschmidt: Will do! 👍

src/components/tab-bar/tab-bar.scss
adrianschmidt: Now there are buttons. Not yet quite as nice looking as those in the mock, but functional at least 🙂

src/components/tab-bar/tab-bar.tsx
adrianschmidt: ⚡

jensgustafsson:
Found an issue with the arrows in combination with the new sticky header in the webclient:

File

When you scroll the arrows is shown above the sticky header. I'm not sure if it should be considered to be a bug here or in the header

adrianschmidt:
> Found an issue with the arrows in combination with the new sticky header in the webclient:
> 
> When you scroll the arrows is shown above the sticky header. I'm not sure if it should be considered to be a bug here or in the header

This one should be :zap:'ed now, as I've removed the use of `z-index` completely.

hannahu:
Looks reaaaaally nice!! I know the showed mock has large buttons but don't you think the arrows are a bit too large or is it just me? :D

src/components/tab-bar/tab-bar.tsx
adrianschmidt: I've changed this to use State-props instead in #1181, because we've also added the `disabled` attribute to the buttons in that PR.

Issue #754, opened 2020-05-11
Sync input styles & layouts in Lime Elements


Kiarokh:
We need to make sure when components align correctly, when placed beside each other or under each other. This applies to components such as input fields, sliders, pickers, chipsets, etc...

Examples of miss-alignments are displayed below:
![Screen Shot 2020-04-17 at 11 48 18](https://user-images.githubusercontent.com/35954987/81572682-cc042f00-93a3-11ea-95eb-10aa5316f91a.png)

![image](https://user-images.githubusercontent.com/35954987/81572919-1d142300-93a4-11ea-920e-e47cbd460f25.png)

![image](https://user-images.githubusercontent.com/35954987/81572932-21d8d700-93a4-11ea-9c8a-24cb58588b7c.png)

Kiarokh:
The way **chips** and **select** is visually rendered is a bit different than before. I don't think it's a breaking change. But please make sure to test this with real examples in the CRM. I have mostly tested it with Lime Elements.

**Note:** This PR doesn't fix the issue with the slider (as shown in the examples)!

Kiarokh:
Maybe I should also add some tips and descriptions to the docs.

Issue #728, opened 2020-04-24
Too small margin over limel-select


sofiehelgesson:
## Current behavior

I have a dialog with both input fields and limel-selects and the margin for limel-select is smaller than for the limel-input
![image](https://user-images.githubusercontent.com/24894627/80194990-87466d00-861b-11ea-9685-9542fb2e8e12.png)

## Expected behavior

the same margin for both limel-input and limel-select

## Environment

- **lime-elements** version: 22.3.0

adrianschmidt:
This issue is very similar to #626. At least in this one, the label doesn't actually *overlap* the preceding component 😄

Ping @Kiarokh: These two issues seem like something you could probably find a good solution for?

Kiarokh:
Tnx Adrian. Yes I'll check both of them up ;)

Kiarokh:
@adrianschmidt & @JohannesHansen I'm fixing this issue and unifying all these different fields. 

But I wanted to pick your brains about another thing regarding the layout of these fields. In the attached picture above, we see that all fileds are tightly placed after each other. There is no distance between them, which doesn't provide a good UI design. I believe that the reason is the absence of `mdc-text-field-helper-text` under the fields in this custom implementation.

In LimeElements' own "Form" example, these divs exist (but they're empty), which will automatically make the layout more organized. 

If we remove them, same problem happens there too.
<img width="1534" alt="Screen Shot 2020-05-08 at 13 58 38" src="https://user-images.githubusercontent.com/35954987/81403896-a9b8aa00-9134-11ea-83b0-6c4d56d0e873.png">

That's also one of the reasons why they will render like this [in the other issue](https://github.com/Lundalogik/lime-elements/issues/626):
![image](https://user-images.githubusercontent.com/35954987/81403928-bc32e380-9134-11ea-832e-bf7c8eb65483.png)
_________________
Anyhow, in reality, we cannot expect the empty helper text containers to be present in the `dom` all the time, thus we may easily get this unpractical form layouts everywhere. Unless the consultants make sure to do some styling themselves, which is probably something that we really want to avoid; right?

So I'm thinking about having some default margins (all sides) in all these types of fields which can be used in forms. Such as inputs, sliders, pickers, selects, (more...?) 

Default margins will be there with or without help texts. In other words, all of these elements will have the same height and some distance from each other, thus would align correctly when placed beside each other in a row. And they will also have some vertical distance when placed in a column. 

Do you also thing this is the right track to go?

[...]

PR #765, opened 2020-05-15
Unify input field layouts

Kiarokh:
fix: https://github.com/Lundalogik/lime-elements/issues/754
fix: https://github.com/Lundalogik/lime-elements/issues/728

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [x] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -1,19 +1,10 @@
-@import '../../style/mixins';
 @import '../../style/internal/mdc-variables';
 @import '../input-field/input-field';
-
-$mdc-textarea-background: $mdc-theme-background;
-// $mdc-textarea-disabled-background: $mdc-theme-background;
-
-$mdc-text-field-background: $mdc-theme-background;
-// $mdc-text-field-disabled-background: $mdc-theme-background;
-
-$mdc-text-field-bottom-line-idle: $lime-text-field-bottom-line-color;
-$mdc-text-field-bottom-line-hover: $lime-text-field-bottom-line-color;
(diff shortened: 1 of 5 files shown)


Kiarokh:
@adrianschmidt I used `fix` for two different issues here. Hope this is correctly done, and as you meant. ;)

adrianschmidt:
> @adrianschmidt I used `fix` for two different issues here. Hope this is correctly done, and as you meant. ;)

Looks like GitHub linked the issues correctly 👍

terryilebode6:
the commit marked as a chore should have a scope, so something like chore(lime-elements): .......

Kiarokh:
To test this properly, we need to have Limel- Select, -Picker, and -Input 

1- beside each other, and 
2- under each other. 

And see how they are laid out, and whether they align correctly or not. A good place to do that is in the object cards (Fullscreen view Vs. Related view). But I don't know which object has all these 3 fields together. (optimally, we also need a Slider in there too, because my next task will be unifying the layout of the slider with these 3 guys)

Another solution is to make an example in limel-form and put all of these together in the form and test it there.

adrianschmidt:
> the commit marked as a chore should have a scope, so something like chore(lime-elements): .......

A scope isn't required, and adding the scope `lime-elements` is completely redundant, since that doesn't narrow anything down at all. So I think that chore is fine without a scope.

terryilebode6:
while placed horizontally , the items still look not aligned properly  
image

PR #771, opened 2020-05-18
Search input field

Befkadu1:
fix: Lundalogik/crm-feature#1109
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/examples/input-field-search.tsx
+++ b/src/components/input-field/examples/input-field-search.tsx
@@ -1,0 +1,31 @@
+import { Component, h, State } from '@stencil/core';
+
+@Component({
+    tag: 'limel-example-input-field-search',
+    shadow: true,
+    styleUrl: 'input-field.scss',
+})
+export class InputFieldSearchExample {
+    @State()
+    private value;
+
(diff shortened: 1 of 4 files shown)


Befkadu1:
Regarding the support for the leadingIcon, it is already mentioned on top in `Properties`. I am not sure if it is needed to mention about it (leadingIcon) on the  example's title. It is not mentioned on the trailing icons examples and shouldn't we need to be consistent ? If it is necessary I will add it in the title.

hannahu:
> Regarding the support for the leadingIcon, it is already mentioned on top in `Properties`. I am not sure if it is needed to mention about it (leadingIcon) on the example's title. It is not mentioned on the trailing icons examples and shouldn't we need to be consistent ? If it is necessary I will add it in the title.

Ah true, good input!

PR #775, opened 2020-05-22
fix(select): dont always focus select on a react update

### LINKED ISSUE #647: limel-select pulls focus away to it when its values have been updated
## Current behavior

When limel-select gets an update it pulls focus to it. This can cause some unexpected behavior because changing unrelated inputs causes the focus to be pulled to the limel-select. The offending code can be found in select.ts

```javascript
    public componentDidUpdate() {
        if (this.menuOpen) {
            this.setMenuFocus();
        } else {
            this.setTriggerFocus();
        }
    }
```
The limel-select element could be rerendered for a number of reason including the parent being rerendered due to an unrelated change. In my use case editing a limel-input-field caused focus to be pulled to the limel-select element.

## Expected behavior

Focus should not be pulled to limel-select unless the user clicks on the element

## Environment

- **lime-elements** version: 26.11.2
- Framework used: StencilJS 1.10.0
- Logs: N/A

sampsonbryce:
fix #647 

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [x] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/select/select.tsx
+++ b/src/components/select/select.tsx
@@ -154,8 +154,6 @@ export class Select {
     public componentDidUpdate() {
         if (this.menuOpen) {
             this.setMenuFocus();
-        } else {
-            this.setTriggerFocus();
         }
     }
 
@@ -246,6 +244,7 @@ export class Select {
 
         this.change.emit(option);
         this.menuOpen = false;
+        this.setTriggerFocus();
     }
 
     private openMenu() {


src/components/select/select.tsx
```diff
@@ -154,8 +154,6 @@ export class Select {
     public componentDidUpdate() {
         if (this.menuOpen) {
             this.setMenuFocus();
-        } else {
-            this.setTriggerFocus();
```
adrianschmidt, line 158: Yeah, we should not set focus on each update, but we should also make sure focus is set after the user selects an item, at least if done with the keyboard. If setting focus on every update is a significantly worse bug than losing focus on selection, I think a trade-off is ok, so long as we open another issue to fix the new problem. (Assuming that problem even appears.)

BregenzerK: I agree with Adrian here. 
to explain the problem a bit further: when you inspect the options in the select you'll see that their position in the DOM is not next to the input field but on the body. we had to do that to better position the list in dialogs especially when the dialog needs a scrollbar. Because of the position the 'next' available focusable input will be your browser url bar when you select an option with the keyboard and press tab (without explicitly focusing the input field).

Could you explain the bug you faced, that would be fixed with this change?

sampsonbryce: @BregenzerK The bug i faced was in lime admin 2, when editing a field value, the form object is updated and all select elements are updated. So all select elements have componentDidUpdate method called and so regardless of where my cursor was it would be moved to the last select element on the page. Meaning i would have to reclick into an input-field after typing every character.

Here is a video of the issue. I am entering one character into the input field and then focus is drawn to the select element. I then press enter to trigger the dropdown on the select element to show that my focus is there.
ezgif com-video-to-gif(1)

BregenzerK: oh, that's not good. @adrianschmidt shouldn't the update hook be only triggered when the component is updated? I would excepted that it would ignore updates from other components in the same form?! :thinking:

jgroth: I think we fixed an issue similar to this recently in the picker where it got focus when it shouldn't. @hannahu was the one who fixed it I believe :slightly_smiling_face:

adrianschmidt: It should…

But in lime-admin we've got a tree structure for the schema and a tree structure for the values… I'd bet that the whole tree of values gets replaced when any one value changes, and that triggers an update on every component. Or, if not that precisely, something very similar.

hannahu: Here's my PR for that, haven't looked if it would be applicable here. Can have a look later if needed.
https://github.com/Lundalogik/lime-elements/pull/738

hannahu: > I think we fixed an issue similar to this recently in the picker where it got focus when it shouldn't. @hannahu was the one who fixed it I believe

Here's my PR for that, haven't looked if it would be applicable here. Can have a look later if needed.
#738

jgroth: The solution was basically to move the focusing of the element from the watcher (update lifecycle hook in this case) to the change listener

sampsonbryce: This issue might be a side effect of some additional properties that we pass to the limel-select in lime admin 2 due to integrating with react-jsonschema-form. Which may be why it hasn't popped up before.

Regardless, the core of the issue is that if i programmatically update the select fields value, focus should not be pulled to it. Like if i click a button to auto select a value for the select element, that doesn't mean i should be moving my focus to the select element.

BregenzerK: Hey @sampsonbryce, sorry for pausing this for so long, we were focusing on cycle stuff. 

can you look into the fix that Hanna mentioned for the picker and see if you can apply the same thing to the select? We will otherwise most likely work on limel-form in the next cycle during the fall and attack the problem then

sampsonbryce: @BregenzerK I took a look at made an additional change. We now set focus on the select element again on menu change (when a value is selected) unless the select is in multiple mode. This should address @adrianschmidt concern. I can't think of another use case where we would need to set focus. Focus is already on the select element after select when using the native select.

PR #777, opened 2020-05-22
Populate nested defaults

sampsonbryce:
Fixes for:
* A fix for default values not propagating more than one dependency deep. rjsf-team/react-jsonschema-form#1703

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [x] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/form.tsx
+++ b/src/components/form/form.tsx
@@ -9,7 +9,7 @@ import {
 } from '@stencil/core';
 import React from 'react';
 import { render, unmountComponentAtNode } from 'react-dom';
-import JSONSchemaForm from 'react-jsonschema-form';
+import JSONSchemaForm from './jsonschema-form/jsonschema-form';
 import retargetEvents from 'react-shadow-dom-retarget-events';
 import { FormError, ValidationStatus } from './form.types';
 import {
--- a/src/components/form/jsonschema-form/jsonschema-form.tsx
+++ b/src/components/form/jsonschema-form/jsonschema-form.tsx
@@ -1,0 +1,130 @@
+import React from 'react';
+import JSONSchemaForm from 'react-jsonschema-form';
+import {
+    retrieveSchema,
+    getDefaultFormState,
+} from 'react-jsonschema-form/lib/utils';
+import { isEqual } from 'lodash-es';
+
+/**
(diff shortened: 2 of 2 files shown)


adrianschmidt:
FYI, the review checklist is meant for the reviewer to keep track of what they've looked at and should look at. So you shouldn't fill that out when you open a PR. Sadly, the reason I suspect you (Bryce) filled it out, is that no one but me seems to use it when reviewing… 😕

src/components/form/adapter/base-adapter.ts
```diff
@@ -0,0 +1,211 @@
+import React from 'react';
```
adrianschmidt, line 1: Apart from some stuff like the added rule for prettier (odd, I thought we already had a rule for trailing commas 🤔), this seems to be one new feature, so I don't think there should be 17 separate commits. Any fixes of bugs in the new features should be squashed into the feature commit for example. Since the feature was never released with the bug, there's no reason for a separate changelog entry for the bug fix, nor is there any need for a separate commit for that fix.

I haven't looked at this in detail yet, I'm really just trying to catch up on a bunch of emails from the last few days, and saw this. I'll put it in review on our board, and someone else might review it, but I wanted to add this comment 🙂

sampsonbryce: @adrianschmidt You can take this off the board for now i've moved it back to a draft commit I have to update this PR based on changes and feedback in https://github.com/Lundalogik/lime-elements/pull/781

sampsonbryce:
> FYI, the review checklist is meant for the reviewer to keep track of what they've looked at and should look at. So you shouldn't fill that out when you open a PR. Sadly, the reason I suspect you (Bryce) filled it out, is that no one but me seems to use it when reviewing… confused

Good to know! Ill stop doing that i thought is was more a checklist or creating the PR. :)

sampsonbryce:
Closing as the issue was caused by a misconfiguration.

PR #781, opened 2020-05-25
Custom Components for Schema Fields

sampsonbryce:
A modification of PR https://github.com/Lundalogik/lime-elements/pull/773 to simplify the custom components.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [x] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/form/adapters/base-adapter.ts
```diff
@@ -0,0 +1,211 @@
...
+            if (isPrimitive(elementProps[propName])) {
+                // Dash case required here but not when setting props manually in `setComponentProperty`.
+                // It is assumed that camcelCase props are converted to dash-case, but that is not the case
+                // and we end up with props using 'thiscase' <- bad
+                primitiveProps[toDashCase(propName)] = elementProps[propName];
+            }
+        }
+
+        return primitiveProps;
+    }
```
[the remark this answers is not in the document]
jgroth: Oh, and I stumbled upon an alternative to our implementation of `toDashCase` :smile: There's a function called `kebabCase` in Lodash :stuffed_flatbread: :smile:

src/components/form/adapters/base-adapter.ts
```diff
@@ -0,0 +1,211 @@
...
+            prevEvents: {},
+        };
+
+        this.component = React.createRef();
+
+        this.setComponentProperty = this.setComponentProperty.bind(this);
+    }
+
+    // Sets a property on the web component
+    setComponentProperty([key, value]: [string, any]) {
```
jgroth, line 43: I get that this is invoked in the loop further down when getting the entries from an object, but it looks a bit weird to have the signature like this. Do you think we can improve it? Something like
```typescript
setComponentProperty(key: string, value: any)
```
Maybe it will be natural because of some other proposed changes? :thinking:

src/components/form/adapters/base-adapter.ts
```diff
@@ -0,0 +1,211 @@
...
+        const element = this.component.current;
+        element[key] = value;
+    }
+
+    /**
+     * Get the handler for an event.
+     *
+     * The handler stop propagation otherwise. Should be used instead of just binding
+     * the handler passed in the `events` prop due to this reason
```
jgroth, line 52: I think we maybe should rephrase this a bit to make it easier to understand

src/components/form/adapters/base-adapter.ts
```diff
@@ -0,0 +1,211 @@
...
+        public props: {
+            name: string;
+            events?: any;
+            elementProps?: any;
+        }
+    ) {
+        super(props);
+
+        this.state = {
+            prevEvents: {},
```
jgroth, line 34:
```suggestion
            previousEvents: {},
```

src/components/form/adapters/base-adapter.ts
```diff
@@ -0,0 +1,211 @@
...
+                eventsChanged = true;
+            }
+        }
+
+        if (eventsChanged) {
+            this.setState({ prevEvents: events });
+        }
+
+        this.removeEvents(eventsToRemove);
+        this.setEvents(eventsToAdd);
```
jgroth, line 121: Do we still need to do all this even if nothing has changed?

src/components/form/adapters/base-adapter.ts
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    /**
+     * Detect if event handlers have changed. If they
+     * have changed, store them in the state, and remove and
+     * re-add the changed event handlers
+     *
+     * @returns {void}
+     */
+    updateEvents() {
```
jgroth, line 101: Do you think we could simplify this method a bit? It has a lot of responsibilities right now. It filters out old events, new events, remove old listeners and sets new listeners. I get what the method does, but I have to read it a few times to fully understand it

src/components/form/fields/schema-field.ts
```diff
@@ -0,0 +1,76 @@
...
+        required: required,
+        readonly: readonly,
+        name: name,
+        disabled: disabled,
+        title: schema.title,
+        helperText: schema.description, // TODO: pull error from errorSchema
+        formInfo: {
+            schema: schema,
+            rootSchema: registry.formContext.schema,
+            errorSchema: errorSchema,
```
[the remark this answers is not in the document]
sampsonbryce: im not 100% what it looks like yet i haven't gotten it to populate in testing yet. The documentation for it can be found here: https://react-jsonschema-form.readthedocs.io/en/latest/advanced-customization/custom-widgets-fields/#field-props

src/components/form/fields/schema-field.ts
```diff
@@ -0,0 +1,76 @@
...
+    };
+};
+
+const renderCustomComponent = (props: FieldProps) => {
+    const handleChange = (event) => {
+        event.stopPropagation();
+        props.onChange(event.detail);
+    };
+
+    const { name, props: overridenFieldProps } = getCustomComponent(
```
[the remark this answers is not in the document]
sampsonbryce: yep copy paste error. I renamed everything from overriden to just custom something... ill update

src/components/form/adapters/base-adapter.ts
```diff
@@ -0,0 +1,211 @@
...
+    }
+
+    /**
+     * Remove events on the web component
+     *
+     * @param {object} events Event names to handlers
+     *
+     * @returns {void}
+     */
+    removeEvents(events: object) {
```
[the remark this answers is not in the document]
sampsonbryce: Ya i suppose listener is a little more explicit in this context

src/components/form/adapters/base-adapter.ts
sampsonbryce: we do not

src/components/form/adapters/base-adapter.ts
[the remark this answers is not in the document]
sampsonbryce: Good catches!

src/components/form/adapters/widget-adapter.ts
```diff
@@ -0,0 +1,136 @@
...
+
+        return React.createElement(LimeElementsAdapter, {
+            name: name,
+            elementProps: {
+                value: value,
+                label: this.getLabel(),
+                disabled: disabled,
+                required: this.isRequired(),
+                invalid: this.isInvalid(),
+                'helper-text': this.getHelperText(),
```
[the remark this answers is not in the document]
sampsonbryce: I don't think so. LimeElementsAdapter is intended to be low level and really just handle whatever it takes to go from a React environment to a web component environment and could theoretically be used for any application that needed that functionality. These props are all specific to widgets and rjsf so i keep them all here.

src/components/form/fields/schema-field.ts
```diff
@@ -0,0 +1,76 @@
...
+    } = props;
+
+    return {
+        value: formData,
+        required: required,
+        readonly: readonly,
+        name: name,
+        disabled: disabled,
+        title: schema.title,
+        helperText: schema.description, // TODO: pull error from errorSchema
```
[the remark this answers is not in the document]
sampsonbryce: Ya I need to do more testing with errorSchema first to see what it really does/looks like

src/components/form/adapters/base-adapter.ts
sampsonbryce: As much as im for being more explicit with variables names, as i've reworked the code based on your suggestions, i use prev/next do identify previous props/events and the next props/events and its really pretty :rose: to look at cause they are both 4 characters :p So i would rather not do this but lemme know if you think its a really good idea :)

src/components/form/adapters/widget-adapter.ts
```diff
@@ -0,0 +1,127 @@
...
+        super(props);
+
+        this.handleBlur = this.handleBlur.bind(this);
+        this.initState();
+    }
+
+    private initState() {
+        if (this.hasValue()) {
+            this.state.modified = true;
+        }
```
jgroth, line 32: Is this the correct behavior? If the form is initialized with a value, should it be considered to be modified? :thinking:

src/components/form/adapters/base-adapter.ts
```diff
@@ -0,0 +1,296 @@
...
+ * @param {any} value the value to check
+ *
+ * @returns {boolean} Whether or not the value is primitive
+ */
+const isPrimitive = (value) =>
+    isNil(value) ||
+    isNumber(value) ||
+    isString(value) ||
+    isBoolean(value) ||
+    isSymbol(value);
```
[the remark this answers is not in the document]
sampsonbryce: I have not i just added it as it is listed as a primitive by MDN.

src/components/form/adapters/widget-adapter.ts
sampsonbryce: that was the existing behavior before i broke LimeElementsAdapter into LimeElementsAdapter and LimeElementsWidgetAdapter. I believe you wrote it.

jgroth: Haha that's very possible, but it does not mean it is correct :smile:

sampsonbryce: i think the difference would be, on a blank empty form we doe not immediately show validation messages until the user blurs the element, but when loading a form with existing data, we set modified to true which means errors will be shown immediately. It sounds like good behavior to me

src/components/form/adapters/base-adapter.ts
adrianschmidt: Got here because I had to see what all the flowers were about 😄

I agree that the alignment of
```
nextSomething
prevSomething
```
warrants an exception to the rule of avoiding abbreviations 😊

src/components/form/form.types.ts
```diff
@@ -33,3 +35,67 @@ export interface FormError {
...
+
+    /**
+     * The name of the current property
+     */
+    name?: string;
+
+    /**
+     * The title of the current property
+     */
+    title?: string;
```
jgroth, line 68: We might have to change this to something else. All these will become props on the component, which means they will become HTML attributes as well. `title` is already a defined HTML attribute and it might add an ugly native tooltip to the element :dizzy_face:

src/components/form/form.types.ts
```diff
@@ -33,3 +35,67 @@ export interface FormError {
...
+        /**
+         * A tree of errors for this property and its children
+         */
+        errorSchema?: object;
+
+        /**
+         * The value of the whole form
+         */
+        rootValue?: any;
+    };
```
jgroth, line 95: It might be nice to have an interface for this as well so you could refer to it in your own code when implementing `FormComponent` :thinking:

src/components/form/form.types.ts
sampsonbryce: Saw that was planning on relabeling to label. :)

src/components/form/form.types.ts
sampsonbryce: agreed

src/components/form/fields/schema-field.ts
```diff
@@ -0,0 +1,157 @@
...
+const getCustomComponent = (
+    schema
+): { name: string; props: { [key: string]: any } } => {
+    const name = schema.lime?.component?.name;
+    const props = schema.lime?.component?.props || {};
+
+    return { name: name, props: props };
+};
+
+export class SchemaField extends React.Component<FieldProps> {
```
sampsonbryce, line 20: This component has a lot of the same functions as WidgetAdapter and looks very similar in many cases. However, most of the functions work slightly differently due to the differences between a widget and a field in rjsf. I thought about maybe trying to abstract away the similarities but it seemed excessive for a one off use case and I think it would have a negative output on the complexity and readability of the code. Just wanted to point that out since i figured it would be mentioned in code review :)

src/components/form/form.types.ts
```diff
@@ -33,3 +35,72 @@ export interface FormError {
...
+
+    /**
+     * Whether or not the current property is disabled
+     */
+    disabled?: boolean;
+
+    /**
+     * The name of the current property
+     */
+    name?: string;
```
jgroth, line 63: Maybe this belongs to the `FormInfo` interface? :thinking: 
All of the other properties here are used to change the appearance/behavior of the field in some way, but this seems to be more of a "metadata" type field. What do you think?

sampsonbryce: ya makes sense

src/components/form/fields/schema-field.ts
```diff
@@ -0,0 +1,157 @@
+import { LimeElementsAdapter } from '../adapters';
+import JSONSchemaField from 'react-jsonschema-form/lib/components/fields/SchemaField';
+import React from 'react';
+import { FieldProps } from './types';
+import { isEmpty, capitalize } from 'lodash-es';
+
+const hasCustomComponent = (schema): boolean => {
+    return Boolean(schema.lime?.component?.name);
+};
```
jgroth, line 9: I'm not sure if we should have some error handling here or just ignore the custom rendering if the component does not exist :thinking: 
We can check the existance of a component by calling `customElements.get(schema.lime?.component?.name)`

sampsonbryce: hmm ya good idea i would go with throwing an error i think. silent failures are awful to debug.

jgroth:
Great work! I think it looks good and is ready to be merged, but I would like someone else from the team to have a quick look at it as well :slightly_smiling_face: 
We should also squash the commits before we merge it

src/components/form/adapters/widget-adapter.ts
```diff
@@ -11,6 +11,17 @@ interface WidgetAdapterProps {
     events?: { [key: string]: (event: any) => void };
 }
 
+/**
+ * A widget is a concept in react-jsonschema-form (RJSF).
+ * It represents a HTML tag for the user to enter data, eg. input, select, etc.
+ *
+ * You're using a widget when you want to render a field in a form
+ * in a specific way. For boolean fields you could for instance use a `radio` or
```
sampsonbryce, line 19: This wording makes it seem like users can override widgets when they can't. Users custom components are rendered as Fields. We use the widget adapter exclusively so we can use lime-elements instead of the default input/select/checkbox/etc fields that rjsf provides.

jgroth:
Needs to be squashed before merge :slightly_smiling_face:

sampsonbryce:
> Needs to be squashed before merge slightly_smiling_face

done

PR #788, opened 2020-05-28
fix(input-field): sync invalid state to underlying mdctextfield

### LINKED ISSUE #791: input-field not respecting invalid prop
## Current behavior

The mdcTextField will not respect the invalid class put onto the containing label component and will instead reset the classes based on its own calculation of whether or not the component is valid, likely based on native input attributes such as required and the type.

### Steps to reproduce
* Pull master
* Get lime-elements into the browser somehow
* Insert `<limel-text-field invalid help-text="help text" value=""></limel-input-field>` into the dom
* The text field should be red signifying that it is invalid (since the invalid prop is set). 
* Click to focus the input field
* Click away to blur the input field
* Note that the red styling due to the invalid state is missing due to loss of focus. This is incorrect the red styling should remain since the invalid prop is set

## Expected behavior

If I set the invalid prop to true the field should be shown as invalid regardless of the internal validation of the text field.

## Environment

- **lime-elements** version: 26.27.0
- Framework used: N/A the issue shows when using lime-elements in an empty html page.
- Logs: N/A

## Proposed Solutions
https://github.com/Lundalogik/lime-elements/pull/788

sampsonbryce:
The mdcTextField will not respect the invalid class put onto the containing label component
and will instead reset the classes based on its own calculation of whether or not the component
is valid, likely based on native input attributes such as required and the type. This sync update
forces the mdcTextField into an invalid state if the invalid prop is set, otherwise it
lets mdcTextField determine if the component is valid or not.

## Steps to reproduce
* Pull master
* Get lime-elements into the browser somehow
* Insert `<limel-text-field invalid help-text="help text" value=""></limel-input-field>` into the dom
* The text field should be red signifying that it is invalid (since the invalid prop is set). 
* Click to focus the input field
* Click away to blur the input field
* Note that the red styling due to the invalid state is missing due to loss of focus. This is incorrect the red styling should remain since the invalid prop is set

fix #791

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/examples/input-field-error-icon.tsx
+++ b/src/components/input-field/examples/input-field-error-icon.tsx
@@ -7,25 +7,50 @@ const MIN_LENGTH = 6;
 })
 export class InputFieldErrorIconExample {
     @State()
-    private value;
+    private valueNative: string;
+
+    @State()
+    private valueConsumer: string;
 
     constructor() {
-        this.onChange = this.onChange.bind(this);
+        this.onChangeNative = this.onChangeNative.bind(this);
+        this.onChangeConsumer = this.onChangeConsumer.bind(this);
+        this.isInvalid = this.isInvalid.bind(this);
     }
 
     public render() {
-        return (
+        return [
             <limel-input-field
-                label="Text Field"
+                label="Text Field with native validation"
                 minlength={MIN_LENGTH}
                 helperText="Please enter at least 6 characters!"
-                value={this.value}
-                onChange={this.onChange}
-            />
-        );
+                value={this.valueNative}
+                onChange={this.onChangeNative}
+            />,
+            <limel-input-field
+                label="Text Field with consumer validation"
+                type={'email'}
+                invalid={this.isInvalid()}
+                helperText="Please enter an email with the domain 'test.com'"
+                value={this.valueConsumer}
+                onChange={this.onChangeConsumer}
+            />,
+        ];
     }
 
-    private onChange(event) {
-        this.value = event.detail;
+    private onChangeNative(event: CustomEvent<string>) {
+        this.valueNative = event.detail;
+    }
+    private onChangeConsumer(event: CustomEvent<string>) {
+        this.valueConsumer = event.detail;
+    }
+
+    private isInvalid() {
+        const substringLength = 9;
+        return !!(
+            this.valueConsumer &&
+            this.valueConsumer.substr(-substringLength, substringLength) !==
+                '@test.com'
+        );
     }
 }
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -188,6 +188,12 @@ export class InputField {
         }
     }
 
+    public componentDidUpdate() {
+        if (this.invalid) {
+            this.mdcTextField.valid = false;
+        }
+    }
+
     public render() {
         if (this.type === 'textarea') {
             return this.renderTextArea();


src/components/input-field/input-field.tsx
```diff
@@ -570,4 +574,17 @@ export class InputField {
...
         }
     }
+
+    private syncInvalidToTextField() {
+        if (!this.mdcTextField) {
+            return;
+        }
+
+        if (!this.mdcTextField.valid) {
+            return;
```
adrianschmidt, line 584: > This sync update forces the mdcTextField into an invalid state if the invalid prop is set, otherwise it lets mdcTextField determine if the component is valid or not.

Hmm… 🤔

What happens if, for some reason, the consumer's validation says the input is valid, but the MDC component says it's not? The component would render as invalid, I assume, but the value of limel-input-field's `invalid` property would be `false, right? That might mean that a Save-button that normally becomes disabled when there are validation errors will be enabled.

Regardless, that should reasonably be a much less severe bug than the bug this fixes, but we might want to try it out so we know.

src/components/input-field/input-field.tsx
```diff
@@ -570,4 +574,17 @@ export class InputField {
...
+        }
+
+        const isValid = !this.isInvalid();
```
adrianschmidt, line 587: The `isInvalid` function looks a little weird to me, especially when `syncInvalidToTextField` is added to the mix.

Seems like we're now juggling the `invalid` property set by the consumer, the validity reported by the *native* element, and the validity reported by the MDC component, and it's spread out over two different functions where one is calling the other, but both are called by other code at different points in the life cycle. I think it's too complex and inconsistent. If we really need to consider all these three values, we need to do it consistently, in a single place, and make sure all those values are considered every time any of them are updated.

src/components/input-field/input-field.tsx
sampsonbryce: Ya im not 100% sure how we could make the save button disabled if the MDC internal validation fails. From what i understand of this code the MDC internal validation is not exported in any way so theres no way to access it outside of limel-input-field right?

src/components/input-field/input-field.tsx
sampsonbryce: Ya it is kinda hard to follow. I'm not 100% sure the best way to go about this. If you want I can port this PR to an issue and link this as a potential solution so that the best way to fix this can be discussed outside of the context of a specific implementation.

jgroth: I'm not 100% sure, but I think there is only the `invalid` property and the valid status of the native element. Material is using the native validity reported as well if I remember correctly. The `invalid` property should probably always have precedence over anything reported by the element

adrianschmidt: > If you want I can port this PR to an issue and link this as a potential solution

Yeah, that would be great. It's generally a good idea to create a bug report first, even if you already have a PR ready to go (not that I always do it myself 😖).

sampsonbryce: issue: https://github.com/Lundalogik/lime-elements/issues/791

BregenzerK:
I looked into the issue (because I also didn't like the sync method too much). I think we have two cases that we want to support: 
- the native validation depending on the type and properties like `minlength`
- a consumer validation, that ends up defining the `invalid` property based on consumer logic that the native validation check can't take into consideration

In my opinion we just need to overwrite the native valid state if the consumer 'knows better' to slso support the second case. I also adjust the example to show both cases.

What do you think?
@adrianschmidt @jgroth

adrianschmidt:
> In my opinion we just need to overwrite the native valid state if the consumer 'knows better' to also support the second case. I also adjust the example to show both cases.

Yeah, that sounds good to me 👍

src/components/input-field/examples/input-field-error-icon.tsx
```diff
@@ -7,25 +7,49 @@ const MIN_LENGTH = 6;
...
+        this.valueConsumer = event.detail;
+    }
+
+    private isInvalid() {
+        const substringLength = 9;
+        return !(
+            this.valueConsumer &&
+            this.valueConsumer.substr(-substringLength, substringLength) ===
+                '@test.com'
+        );
```
adrianschmidt, line 53: I added this small improvement of the example validator.

If you think it makes the example harder to understand, feel free to veto it :)

src/components/input-field/input-field.tsx
```diff
@@ -188,6 +188,12 @@ export class InputField {
         }
     }
 
+    public componentDidUpdate() {
+        if (this.invalid) {
+            this.mdcTextField.valid = false;
```
adrianschmidt, line 193: Wow, nice simplification! 👏

src/components/input-field/examples/input-field-error-icon.tsx
BregenzerK: looks fine, I wasn't going for absolute perfection on the validator, just wanted to make a point ;) but if that is exacter then go with it

Issue #789, opened 2020-05-29
"hidden" input field in the chipset breaks the UI


Kiarokh:
###Problem
When the picker is inactive, the the input field gets a class called `hidden`. But it still physically exists in the DOM and breaks the UI by adding extra space under the chips (in some cases, depending on the width of the placeholder text and the width of picker's container).
![image](https://user-images.githubusercontent.com/35954987/83232796-06384380-a18e-11ea-92e3-372f07638d3f.png)

![image](https://user-images.githubusercontent.com/35954987/83233257-bc9c2880-a18e-11ea-9b09-9a0920fa3f0f.png)

![image](https://user-images.githubusercontent.com/35954987/83232807-0a646100-a18e-11ea-8b20-5b01601820ea.png)

We can't add a style such as `display: none;` to it when `hidden`, because then users cannot use keyboard navigation (tab) to activate the picker.
We can't use techniques such as adding `width: 0; height: 0;` neither, because that creates other weird problems as soon as it is reactivated using tab or clicking in the field.

_________________
### Solution in the linked PR
Use `position: absolut;`, `opacity: 0;` and `z-index: -100;`to really hide the input field and render it below all other chips which are visible.

![image](https://user-images.githubusercontent.com/35954987/83232740-ee60bf80-a18d-11ea-9f87-3e043901c3ff.png)

###Note
User doesn't have to click on the `<input>` itself to activate it. They click on the container of it -the picker- to activate the input field and make it  `focused`.

PR #790, opened 2020-05-29
fix(chip-set): render input really "hidden", when it's hidden.

Kiarokh:
fix: https://github.com/Lundalogik/lime-elements/issues/789

Use `position: absolut;`, `opacity: 0; `and `z-index: -100;`to really hide the input field and render it below all other chips which are visible.

Note
User doesn't have to click on the `<input>` itself to activate it. They click on the container of it -the picker- to activate the input field and make it `focused`.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -121,7 +121,9 @@ limel-icon.mdc-chip__icon.mdc-chip__icon--leading {
             // This class is added, as soon as there is a chip selected and displayed
             // This input field should not be visually visible as it breaks the UI in some cases
             // But it should be rendered to be able to tab between fields and do other keyboard commands /Kia
-            padding: 0;
+            opacity: 0;
+            position: absolute;
+            z-index: -100;
         }
     }
 }


src/components/chip-set/chip-set.scss
```diff
@@ -121,7 +121,9 @@ limel-icon.mdc-chip__icon.mdc-chip__icon--leading {
             // This class is added, as soon as there is a chip selected and displayed
             // This input field should not be visually visible as it breaks the UI in some cases
             // But it should be rendered to be able to tab between fields and do other keyboard commands /Kia
-            padding: 0;
+            opacity: 0;
+            position: absolute;
+            z-index: -100;
```
BregenzerK, line 126: just out of curiosity: is `-100`just a magical high enough number or is there a particular reason?

PR #796, opened 2020-06-02
feat(table): created simple table component

terryilebode6:
fix Lundalogik/crm-feature#1148
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/table/table.tsx
```diff
@@ -0,0 +1,45 @@
...
+    tag: 'limel-table',
+    styleUrl: 'table.scss',
+    shadow: true,
+})
+export class Table {
+    /**
+     * table data to be displayed
+     */
+    @Prop()
+    public tableData: any[] = [];
```
jgroth, line 18: `any[]` will mean that it will be possible to send in an array of strings, which will fail. It should probably be set to `object[]`

jgroth, line 18: What will happen if the consumer updates the data that is being passes in here, will the table display the new data?

src/components/table/table.tsx
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * object for column definitions
+     * {
+     *    title -required
+     *    field- required
+     * }
+     */
+    @Prop()
+    public columns: Column[] = [];
```
jgroth, line 27: Same thing here. If the consumer updates the columns, will the table display the new headers? :slightly_smiling_face:

src/components/table/table.tsx
```diff
@@ -0,0 +1,45 @@
...
+     * table data to be displayed
+     */
+    @Prop()
+    public tableData: any[] = [];
+    /**
+     * object for column definitions
+     * {
+     *    title -required
+     *    field- required
+     * }
```
jgroth, line 24: I think we should remove these. I think absolutely we should document this in some way but this is not the right place. We have not written it like this for any other components, e.g. `ListItem` in the list component

src/components/table/table.scss
```diff
@@ -0,0 +1,8 @@
+@import '../../../node_modules/tabulator-tables/src/scss/tabulator.scss';
+:host {
+    display: block;
+    height: 100px;
```
jgroth, line 4: We cannot hard-code the height like this :slightly_smiling_face:

src/components/table/examples/table.tsx
```diff
@@ -0,0 +1,53 @@
...
+export class TableExample {
+    public tableData: any[] = [
+        {
+            id: 1,
+            name: 'Billy Bob',
+            age: 12,
+            gender: 'male',
+            height: 95,
+            col: 'red',
+            dob: '14/05/2010',
```
jgroth, line 17: I think we should avoid abbreviations like this, even if this is just an example component

src/components/table/table.tsx
```diff
@@ -0,0 +1,45 @@
...
+     */
+    @Prop()
+    public columns: Column[] = [];
+
+    @Element()
+    el: HTMLElement;
+
+    private tabulator;
```
jgroth, line 32: We should add a type to this property as well

jgroth:
Also, since this is work in progress and will not be solved by just this PR, I think we should add this behind a feature switch

src/components/table/examples/table.tsx
```diff
@@ -0,0 +1,57 @@
...
+        { title: 'Height', field: 'height' },
+        { title: 'Favourite Color', field: 'color' },
+        { title: 'Date Of Birth', field: 'dateOfBirth' },
+    ] as any;
+
+    render() {
+        if (!getFeatureSwitches().showTable) {
+            return <div> warning : construction on-going </div>;
+        }
+        return [<limel-table data={this.tableData} columns={this.columns} />];
```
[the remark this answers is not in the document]
terryilebode6: I added a button in here to test update functionality, forgot to take them out :)

src/util/feature-switches.ts
```diff
@@ -0,0 +1,13 @@
...
+
+function getLocalFeatures(storage: Storage) {
+    const features: Record<string, boolean> = {};
+    for (let i = 0, max = storage.length; i < max; i++) {
+        const key = storage.key(i);
+        const value = storage.getItem(key);
+        features[key] = value === 'true';
+    }
+    return features;
+}
```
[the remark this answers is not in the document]
terryilebode6: ahh , nice , will use that instead

Issue #808, opened 2020-06-12
Hide spinners in buttons to avoid constantly running animations


Kiarokh:
In buttons, there seems to be a `<limel-spinner>` with an infinite rotation `animation`. When the spinner is supposed to be hidden, it actually only gets an `opacity: 0;` style, which means the animation is still running and consuming processing power, consuming battery on phones.
![spinner2](https://user-images.githubusercontent.com/35954987/84496607-880b8f00-acad-11ea-80a2-5f24ea357a0e.gif)
![spinner](https://user-images.githubusercontent.com/35954987/84496617-8b067f80-acad-11ea-822c-d864ae44f4a2.gif)
This can be more problematic when there are several buttons on the screen.

I used `display: none;` instead of hiding it with `opacity: 0;`, so it won't get rendered.

PR #809, opened 2020-06-12
perf(button): hide spinners to avoid constantly running animations

Kiarokh:
fix: https://github.com/Lundalogik/lime-elements/issues/808

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -20,6 +20,7 @@ button {
 
     limel-spinner {
         opacity: 0;
+        display: none;
         position: absolute;
     }
 
@@ -54,6 +55,7 @@ button {
         }
         limel-spinner {
             opacity: 1;
+            display: block;
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/components/button/button.scss
```diff
@@ -65,5 +67,8 @@ button {
         svg {
             opacity: 1;
         }
+        limel-spinner {
+            display: block;
+        }
```
adrianschmidt, line 72: With this addition under `.just-loaded`, the spinner properly fades out when loading ends, but is properly hidden 2 seconds later, when the checkmark disappears.

Kiarokh: awesome!

Issue #810, opened 2020-06-12
Improve button styles


Kiarokh:
This PR
- adds elevation effect and interactivity to buttons 
- unifies paddings of primary and secondary buttons. 
- Also buttons that only have an icon won't occupy so much space horizontal space.

PR #811, opened 2020-06-12
fix(button): improve button styles

Kiarokh:
fix: https://github.com/Lundalogik/lime-elements/issues/810
Add elevation effect and interactivity to buttons + unify paddings of primary and secondary buttons. Also buttons that only have an icon won't occupy so much space.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/button/button.scss
```diff
@@ -14,6 +16,13 @@
 }
 
 button {
+    @include is-elevated-clickable();
```
BregenzerK, line 19: is it on purpose that the button has different styles for `:active` and `:focus`?
active/ clicked: 
image
focused but but active (closed alert): 
image

button needs to lose its focus to turn back to the neutral state.

Kiarokh: Hmm... I haven't added any styles for `focus` state, unless MD has something. However the styles for `active` state is there to indicate what happens when you press/click the button. It will go down with an animation. To test it, click the button but don't immediately release your click.

Kiarokh: We will at some point make some unified styles for focused states for all components. Today, we have no good and nice design for them. Already had some thought exchange regarding this with @adrianschmidt

BregenzerK: > To test it, click the button but don't immediately release your click.

that's fine, but it looks like we capture the same active state, when the button opens the alert (I haven't tested whether it's the same with a regular dialog). Is that on purpose?

adrianschmidt: Yeah, MDC has two different looks for focused and active. The very subtle change of colour between focused and blurred is probably MDC's way to try to balance the need for focus indication for keyboard users with the need for a focus indicator that doesn't look very ugly or weird for mouse users, since the focus will stay on the button after clicking (which is a browser behaviour, that we can't really do much about).

There's a lot of different factors to consider around focus and focus indication, several of which are conflicting, so just let it be like it is for now, and we'll sit down and do some good design work specifically for focus handling on all of our components at some point. If you're interested, you'd be more than welcome to join that effort!

adrianschmidt: Oh, my "yeah" was referring to Kia's replies. I didn't see your reply until now Katja.

If I'm understanding your question correctly, I think the OK-button becomes focused immediately as the dialog opens, right? And yes, that's on purpose, because it lets the user close the dialog by pressing Enter.

BregenzerK: I just wanted to make sure that this was intentionally and not a bug. If that's the case I'm totally fine with the changes

PR #812, opened 2020-06-12
feat(tab-panel): make it possible to scroll inside the content of a tab

jgroth:
fix Lundalogik/crm-feature#1141

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tab-bar/tab-bar.scss
+++ b/src/components/tab-bar/tab-bar.scss
@@ -14,6 +14,7 @@ $tab-scroller-fade-width: 65;
 
 :host {
     display: block;
+    position: relative;
 }
 
 .mdc-tab-bar__icon {
--- a/src/components/tab-panel/tab-panel.scss
+++ b/src/components/tab-panel/tab-panel.scss
@@ -1,0 +1,22 @@
+@import '../../style/internal/variables';
+
+:host(limel-tab-panel) {
+    display: block;
+    height: 100%;
+}
+
+.tab-panel {
+    // contains both limel-tab-bar and tab-content
+    height: 100%;
+    position: relative;
+    display: flex;
+    flex-direction: column;
+}
+
+.tab-content {
+    height: 100%;
+    flex: 1;
+    overflow-y: auto;
+    -webkit-overflow-scrolling: touch;
+    box-sizing: border-box;
+}
--- a/src/components/tab-panel/examples/tab-panel.scss
+++ b/src/components/tab-panel/examples/tab-panel.scss
@@ -1,8 +1,12 @@
 $tab-content-background-color: #fff;
 
+limel-tab-panel {
+    height: 20rem;
+}
+
 limel-example-tab-panel-content {
     display: flex;
-    height: 10rem;
+    height: 100%;
     background-color: $tab-content-background-color;
     justify-content: center;
     align-items: center;
--- a/src/components/tab-panel/tab-panel.tsx
+++ b/src/components/tab-panel/tab-panel.tsx
@@ -15,6 +15,7 @@ import { Tab } from '../tab-bar/tab.types';
  */
 @Component({
     tag: 'limel-tab-panel',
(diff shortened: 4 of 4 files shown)


src/components/tab-panel/tab-panel.scss
```diff
@@ -1,15 +1,20 @@
-:host(limel-tab-panel) {
-    display: block;
```
jgroth: I don't think we can remove this, did you test the tab panel in the documentation? Nothing is visible :disappointed:

src/components/tab-panel/tab-panel.scss
```diff
@@ -1,15 +1,20 @@
...
+    position: relative;
     display: flex;
     flex-direction: column;
 }
 
 .tab-content {
-    height: 100%;
-    overflow: auto;
-    flex: 1;
+    position: absolute;
```
jgroth, line 11: Just a small question in order for me to understand and learn :slightly_smiling_face: 
What is the benefit of doing it this way with absolute positioning?

src/components/tab-bar/tab-bar.scss
```diff
@@ -14,6 +14,8 @@ $tab-scroller-fade-width: 65;
 
 :host {
     display: block;
+    position: relative;
+    z-index: 3;
```
jgroth, line 18: Do we have to use `z-index` here? I would prefer if we could do without it

src/components/tab-panel/tab-panel.scss
jgroth, line 11: Also, if we do it this way it seems you are forced to set a height on the tab-panel as a consumer when using it or it will not be usable, which I'm not a fan of :slightly_frowning_face:

adrianschmidt: Related point: we should perhaps make it configurable whether the tab panel should have a fixed height (or max-height, or whatever way we choose to go about it) and scroll any overflow inside it, or have the panel height extend to show all content, like it does now. It might be that other consumers of the code might want or need the page to scroll rather than the tab content. (Assuming it's not a very complicated thing to achieve, but it doesn't seem to me like it would be?)

jgroth: Previously it would not scroll at all, it would just grow with it's content. It should probably still do that if you do not set a height, but if you do then the content should be scrollable

jgroth: I think we mean the same thing :slightly_smiling_face:

src/components/tab-panel/tab-panel.scss
Kiarokh: in the webclient that `display: block;` was causing it to not render for some reason. That's why I removed it. But I figured having a `height: 100%;` would show it when it's a `block`. So now this is back

src/components/tab-bar/tab-bar.scss
Kiarokh: image
Yes, this is because hovered list items (as well as other things which may have a z-index) will render on top of the tab-bar. And this is because we have an absolute position on `.tab-content`. But now since that's removed, this is not needed anymore. So I'll revet this.

src/components/tab-panel/tab-panel.scss
Kiarokh: well, the short answer is: "That was the only way I could make it work while throwing things around in different places." 🙃
But now when I test with `position: relative` and `flex` and so forth, that works too. Probably something else we did solved the issues. So there is no point in having absolute position anymore. So I change this too.

jgroth:
I approve the changes from @Kiarokh :heavy_check_mark: :smile:

Issue #764, opened 2020-05-15
Focus on datepicker


sofiehelgesson:
## Current behavior

The first time I pick a date it gets marked with the green colour which stays even when I go to another field. In the picture below I first picked a date and then I choose "Sales Status" and then both got marked.

If I go back to the datepicker and then to a new field the colour goes away but not the first time.

![image](https://user-images.githubusercontent.com/24894627/82049408-923e6b80-96b6-11ea-89a5-75b8056c2991.png)

## Expected behavior

that it gets grey, when leaving to another field

## Environment

- **lime-elements** version: 22.3.0

Kiarokh:
@sofiehelgesson i think this issue must be solved when we solved this one: https://github.com/Lundalogik/crm-feature/issues/1019

Can you please verify, and close consequently close the issue if it's solved? Thanks. :)

Kiarokh:
@jensgustafsson can this be something that will automatically get solved when you switch to the new date pickers?

jensgustafsson:
> @jensgustafsson can this be something that will automatically get solved when you switch to the new date pickers?

I hope so 🙏

Kiarokh:
@jensgustafsson @adrianschmidt should we close this issue, as it's addressed here: https://github.com/Lundalogik/lime-elements/issues/815 ? or should we link `fix:`es to them?

adrianschmidt:
#823 (the PR for #815) already mentions that it might fix this issue too. I'll test if it does, and if so, I'll link them so that PR closes this issue when merged.

jensgustafsson:
I've been looking in to this issue today and found out that the bug was partially fixed by #815 but still exists under a specific condition.

Right now I'm leaning towards closing this issue and marking it as "wont fix" as I believe we have other issues that's more important to fix.

Scenario when the bug still appears:

1. Set the value in the datepicker to `null` (same as setting empty string in the input field of the picker)
2. Open the picker and select a new value.

PR #823, opened 2020-06-18
Datepickers in portals

### LINKED ISSUE #815: Update limel-date-picker to use a portal

jensgustafsson:
#### It renders the flatpickr inside a portal

Why?

* We want to make it possible to render the flatpickr correctly inside a limel-form in the webclient.
* We want to render the date picker correctly inside dialogs.

#### It moves the flatpickr container in to a separate web component (`limel-flatpickr-adapter`)

Why?

* Otherwise we can't move the flatpickr component (and keep styling etc) via the limel-portal to the body element in the DOM.

#### Fixes overflow issues

This PR ensures that the flatpickr calendar is rendered above the input field if there is not enough space under the input field on the screen.

#### Removes support for changing the date via the input field directly

Why?

It didn't work on master -> Therefore it was already "dead" code.

#### Comments

This PR might address https://github.com/Lundalogik/lime-elements/issues/764#issuecomment-646289037.

fix #815 
fix Lundalogik/crm-feature#1112

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [x] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS


src/components/date-picker/date-picker.tsx
```diff
@@ -98,123 +89,88 @@ export class DatePicker {
...
+            `#${this.portalId} limel-date-picker-calendar-container`
+        )
 
-    public componentDidUnload() {
-        this.picker.destroy();
+        calendarElement.inputElement = inputElement;
+
+        setTimeout(() => {
+            this.showPortal = false;
+        }, 100); // Without this timeout value the width is set to zero on the picker in the webclient
```
jensgustafsson, line 126: I would like to remove this comment but I also want people to understand why I have chosen the timeout value...

In the best of worlds I would want to completely get rid of the timeout function.

adrianschmidt: Hmm… yeah, this didn't work with a timeout of 0 I guess?

Just in case you or anyone else reading this doesn't already know why I'm asking:
A timeout of zero is completely ok to use, because all that does is letting everything that's "behind" this function run before the timeout triggers. So there's no timing-issue with doing that (no pun intended). But if a non-zero timeout is necessary, that means we do have a timing-issue, and if the process we are waiting for happens to run too slow, we will be triggering the timeout too early.

jensgustafsson: In lime-elements it worked fine but not in the webclient. Super confusing tbh. But I saw that there is another timeout in the datepickers as well which also has a value defined.

It seems like that was added to support changing the date directly in the input field, which we no longer support. (At least not in master). Maybe we could remove that timeout and then this timeout could be run without a value specified.

src/components/date-picker/pickers/Picker.ts
```diff
@@ -101,38 +87,8 @@ export abstract class Picker {
         });
     }
 
-    private handleCloseForFlatpickr(selectedDates) {
```
jensgustafsson, line 104: "Dead code"

This piece of code are not working on master and I don't see us fix it in a near future.

src/components/date-picker/date-picker.tsx
jensgustafsson: ⚡ 

I found a better solution which made it possible to get rid off the timeout value :)

src/components/date-picker/date-picker.tsx
```diff
@@ -195,26 +149,61 @@ export class DatePicker {
...
                 />
+                <limel-portal
+                    containerId={this.portalId}
+                    visible={this.showPortal}
+                    containerStyle={portalContainerStyle}
+                >
+                    <limel-flatpickr-adapter
+                        type={this.type}
+                        value={this.value}
+                        onReady={this.hidePortalElement}
```
adrianschmidt, line 164: It seems like the ready event isn't always emitted from all datepickers…

I'm wondering if there's some other way of getting the size of the portal set correctly… 🤔

jensgustafsson: A potential but hacky fix to this came to my mind.

What if we trigger a "fake" ready event from a function wrapped in a `setTimeout` that always times out after X <time-units> unless an actual ready event is dispatched from flatpickr. (When/if the ready event is emitted from flatpickr we cancel the timeout). Doing that will ensure that we always set the portal to "not visible".

adrianschmidt: The thought of that kind of makes my skin crawl… 😄

Sure we can try. But I'm worried we're building ourselves a real can of worms here. I think a fix like that is likely to introduce some very intermittent bugs.

But, flatpickr has an inline-mode, where the picker is just rendered inline, without any of flatpickr's own menu guff… maybe that could be helpful. It's also possible to not have it open when the user clicks the input, so we can instead catch that click, open the portal, and then open flatpickr.

I don't know how that might mess with the native-picker support on mobile though.

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.tsx
```diff
@@ -175,6 +176,11 @@ export class DatePickerCalendar {
         this.flatPickrCreated = true;
     }
 
+    @Method()
+    public open() {
+        this.picker.open();
```
adrianschmidt, line 181: This commit ("do not open flatpickr on click, but use timeout after portal has opened") is likely not needed. I just wanted to push it before removing it just in case it turned out it was necessary after all. After some quick testing, it seems not to be.

src/components/date-picker/date-picker.tsx
```diff
@@ -98,91 +85,44 @@ export class DatePicker {
...
+    /**
+     * A flag that is set to true when the calendar is rendered.
+     *
+     * The portal needs to be visible in order for flatpickr to set
+     * the correct size when it's attached. That's why showPortal
+     * initially is set to true.
+     *
+     * The flag is switched to false as soon as the component has
+     * loaded.
+     */
```
adrianschmidt, line 102:
```suggestion
```

src/components/date-picker/date-picker.tsx
```diff
@@ -98,91 +85,44 @@ export class DatePicker {
...
+    @State()
+    private showPortal: boolean = false;
```
adrianschmidt, line 104:
```suggestion
    private showPortal = false;
```

src/components/date-picker/date-picker.tsx
```diff
@@ -195,26 +135,63 @@ export class DatePicker {
...
+    private showPortalElement(event) {
+        this.showPortal = true;
+        const inputElement = this.textField.shadowRoot.querySelector('input');
+        setTimeout(() => {
+            this.datePickerCalendar.inputElement = inputElement;
+        });
         event.stopPropagation();
-        this.formattedValue = event.detail;
+
+        document.addEventListener('mousedown', this.eventListener, { passive: true });
```
adrianschmidt, line 174: This means that if the `mousedown` event is caught somewhere and its propagation is stopped, the portal won't be closed. I'm guessing the risk of that happening is pretty small, since we're usually using with the `click` event instead. But thought I'd mention it.

Maybe the portal could also use a scrim? One problem with a scrim is that if the menu is "dropping out of" an input field, the field will normally be underneath the scrim, and thus not possible to click on without closing the dropdown. But perhaps we can move the whole input field into the scrim in those cases? (It's currently not a problem for datepicker at the moment, since it's not possible to directly input text into the input anyway.)

adrianschmidt, line 174: The menu doesn't close when tabbing out of the input, so we probably need a listener for that too.

src/components/date-picker/date-picker.tsx
```diff
@@ -98,91 +85,44 @@ export class DatePicker {
...
-        this.formattedValue = this.picker.formatDate(this.value);
     }
 
-    public componentDidUnload() {
-        this.picker.destroy();
+    private eventListener = (event: any) => {
+        const element = document.querySelector(`#${this.portalId}`);
+        if (!element.contains(event.target)) {
+            this.hidePortalElement();
+        }
```
adrianschmidt, line 115: I'd move this one down below `hidePortalElement`, around line 183.

src/components/date-picker/date-picker.tsx
```diff
@@ -98,91 +85,44 @@ export class DatePicker {
...
-
-        this.picker.init(this.input, this.container, this.value);
-        this.formattedValue = this.picker.formatDate(this.value);
     }
 
-    public componentDidUnload() {
-        this.picker.destroy();
+    private eventListener = (event: any) => {
+        const element = document.querySelector(`#${this.portalId}`);
+        if (!element.contains(event.target)) {
```
adrianschmidt, line 113:
```suggestion
    private eventListener = (event: MouseEvent) => {
        const element = document.querySelector(`#${this.portalId}`);
        if (!element.contains(event.target as Node)) {
```

src/components/date-picker/date-picker.tsx
jensgustafsson: This seems to be the code that handles this in flatpickr: https://github.com/flatpickr/flatpickr/blob/00036f843b2882fbf810b8e4639a02c34a28ad7d/src/index.ts#L1427

What event do you think we need to listen to for that to be fixed?

src/components/date-picker/date-picker.tsx
```diff
@@ -195,26 +125,63 @@ export class DatePicker {
...
+    private handleCalendarChange(event) {
+        const date = event.detail;
+        this.formattedValue = this.dateFormatter.formatDateByType(
+            date,
+            this.type
+        );
         event.stopPropagation();
-        this.formattedValue = event.detail;
+        if (this.type !== 'datetime' && this.type !== 'time') {
+            this.textField.blur();
```
adrianschmidt, line 182: I think this means the next tab press would focus the first focusable element on the page. Not an issue at the moment, since these aren't keyboard accessible anyway, but just mentioning it for the future.

src/components/date-picker/dateFormatter.ts
```diff
@@ -0,0 +1,56 @@
+import 'moment/locale/da';
+import 'moment/locale/fi';
+import 'moment/locale/nb';
+import 'moment/locale/sv';
```
adrianschmidt, line 4: This reminds me… we probably need to import French here too, right? And Dutch too soon.

(An issue for a different PR obviously. Just writing it down while I'm thinking of it.)

src/components/date-picker/pickers/Picker.ts
```diff
@@ -101,38 +88,8 @@ export abstract class Picker {
...
-                        pickerDate = null;
-                        this.flatpickr.clear();
-                    }
-                }
-                this.change.emit(pickerDate);
-                resolve(pickerDate);
-            }, timeout);
-        });
+    protected getMomentLang() {
+        return this.language === 'no' ? 'nb' : this.language;
```
adrianschmidt, line 92: With the new DateFormatter, are we using this `language` property anywhere?

src/components/date-picker/date-picker.tsx
adrianschmidt: They bind that handler to `touchstart`, `click`, and `focus` here: https://github.com/flatpickr/flatpickr/blob/00036f843b2882fbf810b8e4639a02c34a28ad7d/src/index.ts#L427

src/components/portal/portal.tsx
```diff
@@ -77,6 +77,9 @@ export class Portal {
     @Prop()
     public parent: HTMLElement = document.body;
 
+    @Prop()
+    public inheritParentWidth = false;
```
adrianschmidt, line 81: Note to self: Since `limel-portal` isn't documented for external use, this isn't a breaking change.

src/components/portal/portal.tsx
```diff
@@ -143,33 +146,30 @@ export class Portal {
...
         const rect: any = this.host.getBoundingClientRect();
-        const viewportHeight = this.getViewportHeight();
-        const containerHeight = viewportHeight - rect.y;
 
-        this.container.style.height = `${containerHeight}px`;
-        this.container.style.display = 'block';
-        if (!this.visible) {
+        if (this.visible) {
+            this.container.style.display = 'block';
+            this.container.style.opacity = '1';
```
adrianschmidt, line 155: Do we need this opacity stuff still?

src/components/picker/picker.tsx
```diff
@@ -344,6 +344,7 @@ export class Picker {
                 visible={!!content}
                 containerId={this.portalId}
                 containerStyle={styling}
+                inheritParentWidth={true}
```
adrianschmidt, line 347: Note to self: limel-input-field with autocomplete would need this too, but that doesn't seem to use limel-portal yet, so it's fine for now.

src/components/date-picker/date-picker.scss
```diff
@@ -1,233 +1,13 @@
...
-    --lime-on-primary-color,
-    #{$mdc-theme-on-primary}
-);
-
-.container {
-    display: inline-block;
-    position: relative;
-    width: 100%;
+limel-portal {
+    height: 3.5rem;
```
adrianschmidt, line 2: The portal seems to adjust its height to the content, so I'm not quite sure why this height is set. Is it so that flatpickr will render correctly the first time, and the portal then adjusts after that?

src/components/date-picker/date-picker.scss
```diff
@@ -1,233 +1,13 @@
...
+    z-index: -999999;
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    right: 0;
+    left: 0;
```
adrianschmidt, line 8: This styling seems a little odd to me:

1. Setting both the `bottom` and `top` values should override the `height` value.
2. Since `position` is `absolute`, this will make the `limel-portal` element the exact same size as the closest positioned parent, which seems a little odd. Is the intention to set the element to the full size of the viewport? If so, `position: fixed` should be used.
3. This is setting the size of the `limel-portal` element, not the size of the `div` located in the root of the `body`, where the content is actually rendered.

I'm guessing this might all be left over from earlier attempts at making flatpickr render correctly?

src/components/date-picker/date-picker.scss
```diff
@@ -1,233 +1,13 @@
...
-            }
-        }
-    }
-
-    // The column with week numbers (including the title)
-    .flatpickr-weekwrapper {
-        width: pxToRem(40);
-    }
+div.container {
+    position: relative;
```
adrianschmidt, line 12: Not sure if this is needed either, if the position styling of limel-portal isn't needed.

src/components/menu-surface/menu-surface.scss
```diff
@@ -14,4 +14,5 @@
 .mdc-menu-surface {
     width: var(--menu-surface-width, auto);
     max-height: 100%;
+    position: inherit;
```
adrianschmidt, line 17: I'm not sure I understand what this does 🤔

That is, I understand that it inherits the value of the `position` style from the parent (or closest ancestor where it's set perhaps, not sure), but I don't understand what doing that accomplishes? 🙂

src/components/portal/portal.tsx
jensgustafsson: And since it has a default value I guess its just a minor change anyway? :)

src/components/date-picker/date-picker.scss
jensgustafsson: It's because I wanted it to render above the input element and not "over it" when rendering upwards

src/components/menu-surface/menu-surface.scss
jensgustafsson: Without this, the `.mdc-menu-surface` gets static positioning, which makes the portal get a size of zero.

src/components/date-picker/date-picker.scss
jensgustafsson: I don't feel to confident in writing css so if you think we should change this I'm all ears. (This change is related to the same as I explained in this comment btw:

> It's because I wanted it to render above the input element and not "over it" when rendering upwards

src/components/menu-surface/menu-surface.scss
jensgustafsson: So I say inherit to remove the static positioning that `mdc` adds somewhere deep down in its internal code

src/components/date-picker/date-picker.tsx
```diff
@@ -98,94 +85,35 @@ export class DatePicker {
...
-        this.picker.init(this.input, this.container, this.value);
-        this.formattedValue = this.picker.formatDate(this.value);
-    }
-
-    public componentDidUnload() {
-        this.picker.destroy();
     }
 
     public render() {
+        const inputProps = {
```
jensgustafsson, line 113: Nice improvement!

src/components/menu-surface/menu-surface.scss
adrianschmidt: Static positioning makes it get a size of zero? That's odd 🤔

My thought was that it would be safer to set a fixed value, like `relative`, or `absolute`, because I assumed we wouldn't want it to inherit `fixed` for example. But maybe it actually needs to have the same value not to break… Whatever, if it works it works. Leave it in 😄

adrianschmidt: And resolve this if you're happy with that reply 😛

src/components/date-picker/date-picker.scss
adrianschmidt: Ah, of course! We shouldn't need the `height` value since we've got the `top:0` and `bottom:0` though, so I'd propose removing `height`.

src/components/date-picker/date-picker.scss
adrianschmidt: This is the better way to do it, than setting `height` 👌

src/components/date-picker/date-picker.tsx
adrianschmidt: 🙏

src/components/portal/portal.tsx
adrianschmidt: In the technical sense, yes. But I'd still consider it breaking if it would break the consumer's UI. Which, well, it might or it mightn't I guess. Good thing we can ignore that regardless! 😄

src/components/portal/portal.tsx
jensgustafsson: I think this is a good fix but it's no longer related to this PR. I say we keep it.

src/components/date-picker/pickers/Picker.ts
```diff
@@ -35,13 +40,15 @@ export abstract class Picker {
 
     public init(element: HTMLElement, container: HTMLElement, value?: Date) {
         let config: flatpickr.Options.Options = {
-            allowInput: true,
```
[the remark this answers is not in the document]
jensgustafsson: I like your fix, I approve it :)

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.tsx
```diff
@@ -0,0 +1,189 @@
...
+    private redrawFlatpickr() {
+        this.picker.redraw();
+    }
+
+    public componentDidUnload() {
+        this.picker.destroy();
+    }
+
+    public render() {
+        return <div class="container" />;
```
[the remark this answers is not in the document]
jensgustafsson: ⚡

src/components/date-picker/date-picker.scss
jensgustafsson: :zap

src/components/menu-surface/menu-surface.scss
jensgustafsson: I was considering setting it to `relative` but decided against as inherit would ensure that it gets what's used by the parent. 

Then I press the resolve button :D

src/components/date-picker/date-picker.scss
jensgustafsson: This is indeed needed, otherwise the calender is not shown at all

src/components/menu-surface/menu-surface.scss
adrianschmidt: Ok, if `relative` works, I think we should set it to that. Letting the child inherit `absolute` from the parent seems very likely to break things… 🙂

src/components/portal/portal.tsx
adrianschmidt: Ah, yeah, this was a fix for if the portal is created with `visible` already `true` right? Yeah, then keep it 👍

src/components/date-picker/date-picker.tsx
adrianschmidt: Bump

src/components/date-picker/pickers/Picker.ts
adrianschmidt: Bump

jensgustafsson: We are using the `getMomentLang` I guess?

src/components/date-picker/date-picker.tsx
jensgustafsson: I think I need some help with solving this :)

src/components/menu-surface/menu-surface.scss
adrianschmidt: I changed this to `relative`.

adrianschmidt:
This needs testing on mobiles and stuff, but frankly, I don't want to do that before the demo 😆
So I'm just going to wait and do that later. There's no hurry.

adrianschmidt:
For some reason, all pickers are using flatpickr's calendar instead of native controls on mobile, both Android and iOS 😕

src/components/date-picker/date-picker.tsx
```diff
@@ -110,9 +110,13 @@ export class DatePicker {
 
     public componentWillLoad() {
         this.useNative = isIOSDevice() || isAndroidDevice();
-        this.formattedValue = this.dateFormatter.formatDateByType(
+        this.format = this.format
+            ? this.format
+            : this.dateFormatter.getDateFormat(this.type);
```
adrianschmidt, line 115: Nice. But I don't quite get how this works since `this.format` is an `@Prop()`-decorated property, and those aren't supposed to be possible to change from inside the component without configuring them to be mutable… 🤔

I'm going to need to update this for the native-picker stuff anyway, so leave as-is, and I might update to make use of `@State() internalFormat` or something.

jensgustafsson: You're right about that! Sounds like a good plan. Could it be that it's possible to set just because it's not sent in as a prop?

adrianschmidt: Maybe… 🤷‍♂️

jensgustafsson:
Screenshot 2020-07-05 at 14 04 18

☝️ I found a bug that this PR will add to the webclient. If the menu for "go to table" has a height > the client height it will not be possible to see the items in the very bottom off the menu. (It's not possible to scroll and overflowing elements are hidden)

I've added a fix for that in my latest commit that we could discuss tomorrow. I'm not "super proud" of the fix but couldn't come up with a better solution Let's discuss it tomorrow!

If we use my fix we will have to set the height from the webclient.

adrianschmidt:
> EDIT: I don't like my solution and will re-implement the fix with another solution. If you want to review the code I suggest you wait just a little longer :)

I don't like it either 😝

Not sure what the new fix is you're working on, but it'd be great if we could fix this inside the affected lime-elements component (or components). And it seems to me like it ought to be possible to do that in some way… 🤔 Let's talk about solutions today!

src/components/menu-surface/menu-surface.tsx
```diff
@@ -163,4 +167,24 @@ export class MenuSurface {
...
+            }
+
+            // We don't know if the surface will be rendered "upwards" or "downwards".
+            // We therefore calculate the space above or below the surface and subtract
+            // that from its height.
+            const spaceAboveSurface = Math.max(top, 0);
+            const spaceBelowSurface = Math.max(vh - bottom, 0);
+            const extraCosmeticSpace = 16;
+            const maxHeight = vh - Math.min(spaceAboveSurface, spaceBelowSurface) - extraCosmeticSpace;
+            this.host.style['height'] = `${maxHeight}px`
```
adrianschmidt, line 187: I'm not entirely sure I understand what this is doing, or trying to do.

Because of the `if` on line 175, if we get here, either `spaceAboveSurface` or `spaceBelowSurface` must be 0, which gives us a constant `maxHeight` of `vh - extraCosmeticSpace`.

It seems like we're meaning to consider the placement of the menu-trigger on the page, and crop the height of the menu to the remaining space above or below?

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.tsx
```diff
@@ -0,0 +1,180 @@
...
+        // issue for us.
+        const ONE_SECOND = 1000;
+        const TEN_PIXELS = 10;
+        setTimeout(() => {
+            const { width } = this.container.getBoundingClientRect();
+            if (width < TEN_PIXELS) {
+                this.destroyFlatpickr();
+                this.createFlatpickr();
+            }
+        }, ONE_SECOND);
```
adrianschmidt, line 152: Inside this timeout, we need to check if the portal is still open. Because if it's not, we're destroying a perfectly good flatpickr, and creating a new one in a circumstance where it's guaranteed to fail… 😆

jensgustafsson: haha! That's true! Do you wanna push a fix for it? 🙏

jensgustafsson: I pushed a fix for that. It turns out that the flatpickr element still has its width set when the portal is closed. So by looking at that element instead of the container we can decide if we should rerender.

adrianschmidt: Ah, that sounds great! I'll test it again today 👍

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.tsx
```diff
@@ -144,7 +144,8 @@ export class DatePickerCalendar {
         const ONE_SECOND = 1000;
         const TEN_PIXELS = 10;
         setTimeout(() => {
-            const { width } = this.container.getBoundingClientRect();
+            const flatpickrElement = this.container.querySelectorAll('div.flatpickr-calendar')[0];
```
adrianschmidt, line 147:
```suggestion
            const flatpickrElement = this.container.querySelector('div.flatpickr-calendar');
```

This will both be faster, and not throw an error if nothing is found (although, an error will still be thrown on the next line if that happens…)

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.tsx
```diff
@@ -0,0 +1,188 @@
...
+                break;
+        }
+    }
+
+    public componentDidUpdate() {
+        if (this.flatPickrCreated) {
+            this.redrawFlatpickr();
+        } else {
+            this.createFlatpickr();
+        }
```
adrianschmidt, line 137: I think we might not need this anymore, but I'm leaning towards leaving it until we've merged, and then look at whether it's possible to remove afterwards 🙂

PR #833, opened 2020-06-24
Feat low density table

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/1149

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.mdx
+++ b/src/components/table/table.mdx
@@ -6,7 +6,7 @@ menu: Components
 
 # Table
 
-<h1> This component is under construction</h1> 
+<h1> This component is under construction</h1>
 
 <limel-props name="limel-table" />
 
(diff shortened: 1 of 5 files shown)


src/components/table/table.mdx
```diff
@@ -6,7 +6,9 @@ menu: Components
 
 # Table
 
-<h1> This component is under construction</h1> 
+<h1> This component is under construction</h1>
+This component is based on [Tabulator](http://tabulator.info/), an open source library for generating interactive tables. To get a better understanding of technical possibilities make sure to check out their [documentations](http://tabulator.info/docs/4.7).
+__Don't we need something like this? Otherwise we have to write some documentations about features that we allow.__
```
jgroth, line 11: No, we don't want to expose the fact that we are using a specific library since that should be transparent to the consumer. If we choose to switch to another library, it should still work without the consumer having to update their code. Also, it's not possible anyway to send in properties to Tabulator that we are not explicitly allowing so it's not very useful for them to read that documentation. I we were to just act as a proxy and send anything from the consumer in to Tabulator, we would create a lock-in that would be very hard to get out of if we want to switch in the future

Kiarokh: Thanks for the answer @jgroth I see what you mean. 

I feel we have to have some documentations regarding the properties that are explicitly allowed, or whatever helps people setup and use the table. For example the "Merger" add-on needed tables. They made one. Optimally we want them to use this one of course. Can they easily do that? What do they need to know? What are their options? These things we must explain :) 

I'll remove this part and push again

jgroth: Yeah, the properties that are set directly on the table can be seen on the documentation page, but a lot of options are nested inside other data structures and unfortunately they are not visible in the documentation page and are only listed as `Column<any>[]` and similar :disappointed: We have yet to find a good way to include them here (that's also part of the reason I started on that personal documentation/dev-environment i demoed a few weeks ago :smile: )

The documentation for these options are listed in this file https://github.com/Lundalogik/lime-elements/blob/master/src/components/table/table.types.ts

For someone using this component, they will get information about what properties are available and what they do directly in their code editor. Optimally they would also be listed somewhere else, but at least this is better than nothing :slightly_smiling_face:

Kiarokh: Sounds great! Yes, we need a way to document and share these things. It's great that you are already on it.

adrianschmidt: @jgroth: @jensgustafsson expressed yesterday that he wouldn't mind working on your docs-app on the side. Is it public on GitHub already perhaps? I think it would be fantastic if we could replace docz with something that fits us better, even if it isn't fancy yet 🙂

jgroth: I made it public just now! :smile: 
https://github.com/jgroth/kompendium

PR #837, opened 2020-06-25
docs(props): refresh visual styles with dark-mode support

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/examples/props.scss
+++ b/src/examples/props.scss
@@ -1,3 +1,5 @@
+@import url('https://fonts.googleapis.com/css2?family=Source+Code+Pro&display=swap');
+
 h2 {
     margin: 50px 0 20px;
     line-height: 1.5em;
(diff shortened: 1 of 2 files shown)


src/examples/props.scss
```diff
@@ -6,45 +8,74 @@ h2 {
...
+    tbody {
+        td {
+            padding: 8px 12px 12px 12px;
+            vertical-align: top;
+            font-size: 14px;
+        }
+        tr {
+            display: table-row;
+            &:nth-child(even) {
+                background-color: var(--contrast-300);
```
adrianschmidt, line 57: Looking very nice! But, something a little weird happens when I go back to the landing page after visiting one of the component pages:

image

The reason it only happens after going to a component page and back is probably that this stylesheet isn't loaded until a component page is visited. But why only some of the rules seem to apply I have no idea.

Switching the props-component to use shadow-dom solves it, so I'll add a small commit doing that.

Kiarokh: @adrianschmidt i think those tables for some reason are not using the same tables which are used in props, because they have this weird `fsncwJ` class name and their own CSS. Do you know where they come from?
image

adrianschmidt: :zap:

adrianschmidt: > Do you know where they come from?

Yeah, they are created from our markdown:

image

The styling applied is whatever the default docz theme applies. So that probably had some selectors with higher specificity and that's why only some of the styling was applied to that table.

adrianschmidt: Just don't spend *too* much time styling the docz-app. It's really nice to have it look better, but hopefully we'll get Johan's [kompendium](https://github.com/jgroth/kompendium) to a state where it can replace docz sometime within the foreseeable future 🙂

But stuff that is very generic, like these new styles for tables, we can just lift straight over to kompendium if we like to, so it's not wasted effort.

Kiarokh: Awesome! No I won't spend too much time. This one was a big pain to look at every time, and I just thought it's so generic for tables, so we can move it to Johan's solution as well.

I'll just so some quick fixes to markdown tables too, so they sync

Kiarokh:
I fixed the problem for Markdown generated tables and pushed again. But now since they are inside shadow doms, the styles won't come outside for the markdown tables anymore :) 
image

I think you know how to fix it Adrian.

PR #846, opened 2020-06-29
Add `'nb'` for Norsk Bokmål as permitted language option for date-picker

adrianschmidt:
`no` is a non-standard option used in a lot of Lime CRM databases. We convert `no` to `nb` already,
but we should of course also support using the standard language code for Norsk Bokmål.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [x] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/date-picker/pickers/Picker.ts
+++ b/src/components/date-picker/pickers/Picker.ts
@@ -31,6 +31,7 @@ export abstract class Picker {
         this.handleClose = this.handleClose.bind(this);
         this.parseDate = this.parseDate.bind(this);
         this.formatDate = this.formatDate.bind(this);
+        this.getFlatpickrLang = this.getFlatpickrLang.bind(this);
     }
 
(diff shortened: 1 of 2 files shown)


src/components/date-picker/pickers/Picker.ts
```diff
@@ -87,6 +90,10 @@ export abstract class Picker {
         }
     }
 
+    protected getFlatpickrLang() {
+        return this.language === 'nb' ? 'no' : this.language;
```
adrianschmidt, line 94: > looks like flatpicker doesn't support `nb`. I changed the language property in the localized example to `nb` and the weekdays and months are translated to English not Nowegish

:zap:

PR #855, opened 2020-07-03
fix(form): use Ajv to validate instead of rjsf

jgroth:
Sometimes, the validator in rjsf incorectly reports validation errors when dependencies are used
within the schema. The reported error did not make any sense since it complained about properties
that did not exist, neither in the schema or the data that was validated. Using Ajv to validate the
form data does not report any errors, even though rjsf also uses Ajv under the hood.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4436,14 +4436,21 @@
       }
     },
     "ajv": {
-      "version": "6.10.0",
-      "resolved": "https://registry.npmjs.org/ajv/-/ajv-6.10.0.tgz",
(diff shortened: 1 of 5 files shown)


jgroth:
This fix should allow us to enable the frontend validation of the form in lime-admin again. Now the submit button is always enabled, even if the form is invalid

PR #866, opened 2020-07-07
feat(table): added props factory function to table component definiti…

terryilebode6:
fix Lundalogik/lime-webclient#1396
…on interface

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/table/table.types.ts
```diff
@@ -37,6 +37,8 @@ export interface TableComponentDefinition {
      * Properties to send to the component
      */
     props?: Record<string, any>;
+
+    propFactory?: (data: object) => Record<string, any>;
```
jgroth, line 41: We need to add documentation here so people know how to use it :slightly_smiling_face:

src/components/table/columns.ts
```diff
@@ -72,6 +72,12 @@ export function createCustomComponent(
 
     const element = document.createElement(column.component.name);
     let props: object = column.component.props || {};
+    if (column.component.propFactory) {
+        props = {
+            ...props,
+            ...column.component.propFactory(data),
```
jgroth, line 78: We should also add a test for this

src/components/table/table.types.ts
jgroth: We might also want to rename this to `propsFactory` to match the `props` property

src/components/table/table.types.ts
```diff
@@ -37,6 +37,11 @@ export interface TableComponentDefinition {
      * Properties to send to the component
      */
     props?: Record<string, any>;
+
+    /**
+     * factory function that returns Properties to send to the component
+     */
```
jgroth, line 43: I think we can improve this a bit. We need to explain the input parameter and the return value. Maybe something like this? Feel free to modify :slightly_smiling_face: 

```suggestion
    /**
     * Factory for creating properties dynamically for a custom component.
     *
     * The properties returned from this function will be merged with the
     * `props` properties when the component is created.
     *
     * @param {*} data the data for the current row
     *
     * @returns {object} properties for the component
     */
```

PR #871, opened 2020-07-08
fix(date-picker): improve handling of focus on the input field

### LINKED ISSUE #764: Focus on datepicker

jensgustafsson:
This commit ensures that the input element has focus when the datepicker closes. It also fixes a bug
where it could look like several input fields had focus at the same time.

fix #764

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/date-picker/date-picker.tsx
+++ b/src/components/date-picker/date-picker.tsx
@@ -284,7 +284,6 @@ export class DatePicker {
         );
         event.stopPropagation();
         if (this.type !== 'datetime' && this.type !== 'time') {
-            this.textField.blur();
             this.hideCalendar();
         }
         this.change.emit(date);
--- a/src/components/date-picker/pickers/Picker.ts
+++ b/src/components/date-picker/pickers/Picker.ts
@@ -33,6 +33,7 @@ export abstract class Picker {
 
         this.getWeek = this.getWeek.bind(this);
         this.handleClose = this.handleClose.bind(this);
+        this.handleOnClose = this.handleOnClose.bind(this);
         this.parseDate = this.parseDate.bind(this);
         this.formatDate = this.formatDate.bind(this);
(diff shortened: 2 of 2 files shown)


src/components/date-picker/date-picker.tsx
```diff
@@ -284,7 +285,6 @@ export class DatePicker {
         );
         event.stopPropagation();
         if (this.type !== 'datetime' && this.type !== 'time') {
-            this.textField.blur();
```
jensgustafsson, line 287: This had no effect as the textfield is not an input field but a limel-input-field.

src/components/date-picker/date-picker.tsx
```diff
@@ -259,6 +259,7 @@ export class DatePicker {
         setTimeout(() => {
             this.showPortal = false;
         });
+        this.textField.shadowRoot.querySelector('input').focus();
```
jensgustafsson, line 262: this fixes the bug Sofie mentioned in #764. It also makes our tabbing work better as we no longer have focus on the root dom element after the datepicker has closed

adrianschmidt: Looks good. Since we do it in `hideCalendar`, I assume this doesn't mean we set focus immediately on creation, but please make sure to check that @BregenzerK.

PR #872, opened 2020-07-09
fix(menu): when dropdown is rendered upwards from trigger, do not cover the trigger

adrianschmidt:
Instead of the dropdown extending upwards from the bottom of the trigger, covering the trigger
element, make it extend upwards from the top of the trigger.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu/menu.scss
+++ b/src/components/menu/menu.scss
@@ -35,12 +35,12 @@ $menu-item-background-color: #ebebeb;
         }
 
         &:not(.limel-portal--fixed) {
+            z-index: -999999;
             position: absolute;
-
-            :host(*[open-direction='left']) & {
-                right: 0;
-                bottom: 0;
-            }
+            top: 0;
+            right: 0;
+            bottom: 0;
+            left: 0;
         }
     }
 }


adrianschmidt:
Note! This only fixes the issue with the button being covered when the dropdown is extended upwards instead of downwards. The issue with the menu not being visible at all when too long is a separate issue for a separate PR 🙂

src/components/menu/menu.scss
```diff
@@ -35,12 +35,12 @@ $menu-item-background-color: #ebebeb;
         }
 
         &:not(.limel-portal--fixed) {
+            z-index: -999999;
```
adrianschmidt, line 38: This doesn't affect the z-index of the dropdown itself, only the z-index of the `limel-portal` element, which the dropdown is positioned after. We need this to have a negative z-index to not end up covering the trigger and making the trigger impossible to click.

PR #876, opened 2020-07-09
style: use ESLint instead of TSLint

jgroth:
re Lundalogik/crm-feature#1060

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/badge/badge.spec.ts
```diff
@@ -1,35 +1,19 @@
...
-    const badge = new Badge();
-    it('builds', () => {
-        expect(badge).toBeTruthy();
-    });
-
     it('Badge: returns 950 (3 digits)', () => {
-        const expectedCounter = (badge.label = 950);
-        const actualNumber = badge.showRoundedNumber();
-        expect(actualNumber).toBe(expectedCounter);
+        expect(abbreviate(950)).toBe(950);
```
adrianschmidt, line 5: Oh yes, this is so much better! 🙏

src/components/table/examples/birds.tsx
```diff
@@ -155,7 +155,7 @@ export const data: Bird[] = [
         origin: ['Europe', 'Asia'],
     },
     {
-        name: "Wilson's Snipe",
+        name: "Wilson's Snipe", // eslint-disable-line quotes
```
adrianschmidt, line 158: Isn't there a setting that automatically allows double quotes if it helps avoid escaping single quotes in the string?

jgroth: Yeah maybe :thinking: 
I think there were two rules conflicting each other, so maybe one of them should just be disabled? :thinking:

src/components/badge/badge.spec.ts
jgroth: This should probably have been in a separate `refactor` commit since the changes were a lot more than just formatting, but they ended up in the same one in the end

src/components/button/examples/button.tsx
```diff
@@ -59,17 +59,18 @@ export class ButtonExample {
...
-                    onChange={event => { // tslint:disable-line:jsx-no-lambda prettier
-                        this.props = {
-                            ...this.props,
-                            ...{
-                                [control.property]:
-                                    (event.detail && control.value) || null,
-                            },
-                        };
-                    }}
+                    onChange={this.handleChange}
```
adrianschmidt, line 62: This doesn't work for some reason 😞

src/components/button/examples/button.tsx
```diff
@@ -59,17 +59,18 @@ export class ButtonExample {
...
                 />
             );
         });
     }
+
+    private handleChange = (control) => (event: CustomEvent) => {
```
adrianschmidt, line 68: The double arrow functions don't look right, I'm guessing that's why the button doesn't update anymore.

src/components/badge/badge.spec.ts
adrianschmidt: Yeah, I thought about that too, but in the end, neither `refactor` nor `style` triggers a release or shows up in the changelog, and we currently aren't using release branches with hotpatches, so I'm ok with leaving them together.

adrianschmidt: I did split the switch from tslint to eslint out from the style-changes though. I used to think those belonged together, but actually, should we ever wish to cherry pick the tslint->eslint commit, we might very well not want to have the style changes too.

src/components/checkbox/checkbox.tsx
```diff
@@ -73,7 +73,7 @@ export class Checkbox {
         this.formField.input = this.mdcCheckbox;
     }
 
-    public componentDidUnload() {
+    public disconnectedCallback() {
         this.mdcCheckbox.destroy();
         this.formField.destroy();
     }
```
adrianschmidt, line 79: Hmm… with this fix, I think we'll have to run the setup on `connectedCallback` too, right?

src/components/date-picker/pickers/MonthPicker.tsx
```diff
@@ -123,7 +125,8 @@ export class MonthPicker extends Picker {
         return (
             <div
                 className="datepicker-month"
-                onClick={() => { // tslint:disable-line:jsx-no-lambda prettier
+                // eslint-disable-next-line react/jsx-no-bind
+                onClick={() => {
```
adrianschmidt, line 129: We really should fix this one, as we don't want people reading these docs to copy this behaviour. I'll probably fix it, so view this mostly as a "note to self".

src/components/date-picker/pickers/QuarterPicker.tsx
```diff
@@ -131,7 +134,8 @@ export class QuarterPicker extends Picker {
             <div
                 className="datepicker-quarter"
                 id={`datepicker-quarter-${quarter}`}
-                onClick={() => { // tslint:disable-line:jsx-no-lambda prettier
+                // eslint-disable-next-line react/jsx-no-bind
+                onClick={() => {
```
adrianschmidt, line 138: Note to self 🙂

src/components/date-picker/pickers/YearPicker.tsx
```diff
@@ -94,7 +97,8 @@ export class YearPicker extends Picker {
         return (
             <div
                 className="datepicker-year"
-                onClick={() => { // tslint:disable-line:jsx-no-lambda prettier
+                // eslint-disable-next-line react/jsx-no-bind
+                onClick={() => {
```
adrianschmidt, line 101: Note to self 🙂

src/components/dialog/dialog.tsx
```diff
@@ -114,7 +116,7 @@ export class Dialog {
             : '';
     }
 
-    public componentDidUnload() {
+    public disconnectedCallback() {
         this.mdcDialog.unlisten('MDCDialog:opened', this.handleMdcOpened);
         this.mdcDialog.unlisten('MDCDialog:closed', this.handleMdcClosed);
         this.mdcDialog.unlisten('MDCDialog:closing', this.handleMdcClosing);
```
adrianschmidt, line 122: Again, we probably need to run the setup on `connectedCallback`.

src/components/file/file.tsx
```diff
@@ -82,7 +82,7 @@ export class File {
         );
     }
 
-    public componentDidUnload() {
+    public disconnectedCallback() {
         if (this.mdcTextField) {
             this.mdcTextField.destroy();
         }
```
adrianschmidt, line 88: `connectedCallback`

src/components/badge/badge.spec.ts
adrianschmidt: (I might still end up splitting this out, we'll see 😄)

src/components/form/examples/custom-picker.tsx
```diff
@@ -7,21 +7,39 @@ import { ListItem } from 'src/components/list/list-item.types';
...
+     */
     @Prop()
     public value: number;
 
+    /**
+     * Label to display next to the input field
+     */
     @Prop()
-    public title: string;
+    public label: string;
```
adrianschmidt, line 20: This is a breaking change!

src/components/checkbox/checkbox.tsx
jgroth: Hmm... maybe :thinking:  Good catch either way! :slightly_smiling_face: 
I guess it depends on if `disconnectedCallback` behaves any different than `componentDidUnload` :thinking: 
The reason for this change was that the Stencil ruleset complained about `componentDidUnload` being public, and after some research I found out that `componentDidUnload` will be deprecated in the future. It's not event documented in the Stencil documentation any longer.

However, `connectedCallback` will be called whenever the element is inserted into the DOM while `componendWillLoad` is only called once. So yeah on a second thought I guess that means we should update all of them if we use `disconnectedCallback` like this :smile:

src/components/form/examples/custom-picker.tsx
adrianschmidt: Oh, wait, no, this is just an example… 😄

jgroth: It's just in the example so it's fine :slightly_smiling_face: 
I think the example was not working before :smile:

src/components/form/form.tsx
```diff
@@ -88,7 +88,7 @@ export class Form {
         this.validateForm(this.value);
     }
 
-    protected componentDidUnload() {
+    protected disconnectedCallback() {
         const rootElement = this.host.shadowRoot.querySelector('.root');
         unmountComponentAtNode(rootElement);
     }
```
adrianschmidt, line 94: `connectedCallback`

src/components/button/examples/button.tsx
jgroth: I thought I tested it :disappointed:

src/components/checkbox/checkbox.tsx
adrianschmidt: > after some research I found out that `componentDidUnload` will be deprecated in the future. It's not event documented in the Stencil documentation any longer.


I thought `componentDidUnload` didn't even exist, but this explains why I once started using it 😄

> However, `connectedCallback` will be called whenever the element is inserted into the DOM while `componendWillLoad` is only called once. So yeah on a second thought I guess that means we should update all of them if we use `disconnectedCallback` like this 😄

👍

src/components/dialog/dialog.tsx
jgroth: One note about this is that `connectedCallback` will run before the first render and then we will not be able to setup the mdc components. We might have to use `componentDidRender` instead and just set a flag in `connectedCallback` if we should initialize the mdc components or not :thinking:

src/components/form/examples/custom-picker.tsx
adrianschmidt: > I think the example was not working before 😄

You are absolutely correct! 😄

src/components/icon-button/icon-button.tsx
```diff
@@ -60,7 +59,7 @@ export class IconButton {
         this.mdcIconButtonRipple.unbounded = true;
     }
 
-    public componentDidUnload() {
+    public disconnectedCallback() {
```
adrianschmidt, line 62: disconnectedCallback

src/components/input-field/input-field.tsx
```diff
@@ -182,7 +182,7 @@ export class InputField {
         });
     }
 
-    public componentDidUnload() {
+    public disconnectedCallback() {
```
adrianschmidt, line 185: disconnectedCallback

src/components/linear-progress/linear-progress.tsx
```diff
@@ -31,7 +31,7 @@ export class LinearProgress {
         this.mdcLinearProgress.progress = this.value;
     }
 
-    public componentDidUnload() {
+    public disconnectedCallback() {
```
adrianschmidt, line 34: disconnectedCallback

src/components/list/list-renderer.tsx
```diff
@@ -189,11 +191,11 @@ export class ListRenderer {
...
     private renderIcon(config: ListRendererConfig, item: ListItem) {
-        const style = {};
-        if (item.iconColor) {
-            config.badgeIcons
-                ? (style['--icon-background-color'] = item.iconColor)
-                : (style['color'] = item.iconColor); // tslint:disable-line:no-string-literal
+        const style: any = {};
+        if (item.iconColor && config.badgeIcons) {
+            style['--icon-background-color'] = item.iconColor;
+        } else if (item.iconColor && !config.badgeIcons) {
```
adrianschmidt, line 197: If `item.iconColor && config.badgeIcons` is falsy and `item.iconColor` is truthy, then we already know that `config.badgeIcons` is falsy 🙂 

```suggestion
        } else if (item.iconColor) {
```

Unless you added it as way to harden against bugs appearing as a result of future code changes?

src/components/portal/portal.tsx
```diff
@@ -11,9 +11,9 @@ import { FlipModifier } from '@popperjs/core/lib/modifiers/flip';
  *
  * * Events might not bubble up as expected since the content is moved out to another DOM node
  * * Any styling that is applied to content from the parent will be lost, if the content is
- *   just another web compoent it will work without any issues
- * * When the node is moved in the DOM, `componentDidUnload`, `disconnectedCallback` and `connectedCallback`
- *   will be invoked, so `componentDidUnload` can not be used as a destructor (which is the wrong behavior anyway)
+ * just another web compoent it will work without any issues
+ * * When the node is moved in the DOM, `disconnectedCallback` and `connectedCallback`
+ * will be invoked, so `disconnectedCallback` can not be used as a destructor (which is the wrong behavior anyway)
```
adrianschmidt, line 16: disconnectedCallback

We might have to look into this comment, as `disconnectedCallback` should probably actually be used as a destructor 😄

src/components/portal/portal.tsx
```diff
@@ -77,11 +79,14 @@ export class Portal {
     @Prop()
     public parent: HTMLElement = document.body;
 
+    /**
+     *
```
adrianschmidt, line 83: Note to self: add a description here (Jens and I added this in the date-picker branch).

src/components/portal/portal.tsx
```diff
@@ -77,11 +79,14 @@ export class Portal {
...
+     */
     @Prop()
     public inheritParentWidth = false;
```
adrianschmidt, line 86: Note to self: add a description here (Jens and I added this in the date-picker branch).

src/components/select/select.tsx
```diff
@@ -281,7 +281,7 @@ export class Select {
         const element: HTMLSelectElement = this.host.shadowRoot.querySelector(
             'select.limel-select__native-control'
         );
-        const options = Array.apply(null, element.options)
+        const options = Array.apply(null, element.options) // eslint-disable-line prefer-spread
```
adrianschmidt, line 284: I assume all this is doing is making a copy of the array so we don't mutate the original one? Wouldn't the spread operator do the same thing here?

src/components/select/test/select.test-wrapper.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @stencil/required-jsdoc */
```
adrianschmidt, line 1: We should perhaps just disable this rule for all test-files?

src/components/slider/slider.tsx
```diff
@@ -15,31 +15,58 @@ import {
     styleUrl: 'slider.scss',
 })
 export class Slider {
+    /**
+     * Set to `true` id input is disabled
```
adrianschmidt, line 19: Is there a typo in this description, or am I just having a brain freeze? 😄

adrianschmidt: Ah! "id" should be "if", right? 😂

```suggestion
     * Set to `true` to disable the input
```

maybe?

src/components/slider/slider.tsx
```diff
@@ -15,31 +15,58 @@ import {
...
     /**
      * Default value: 1.
      * The factor that the properties `value`, `valuemax` and `valuemin` are multiplied by.
      * On `change` divides the value by the factor, so the original format stays the same
      */
     @Prop({ reflectToAttr: true })
     public factor: number = 1;
+
+    /**
+     * Label to display nex to the input
```
adrianschmidt, line 33:
```suggestion
     * Label to display next to the input
```

src/components/slider/slider.tsx
```diff
@@ -54,7 +81,7 @@ export class Slider {
         this.mdcSlider.disabled = this.disabled;
     }
 
-    public componentDidUnload() {
+    public disconnectedCallback() {
```
adrianschmidt, line 84: disconnectedCallback

src/components/snackbar/snackbar.tsx
```diff
@@ -74,12 +74,11 @@ export class Snackbar {
         this.mdcSnackbar.listen('MDCSnackbar:closing', this.handleMdcClosing);
     }
 
-    public componentDidUnload() {
+    public disconnectedCallback() {
```
adrianschmidt, line 77: disconnectedCallback

src/components/switch/switch.tsx
```diff
@@ -49,7 +61,7 @@ export class Switch {
     /**
      * @returns {void}
      */
-    public componentDidUnload() {
+    public disconnectedCallback() {
```
adrianschmidt, line 64: disconnectedCallback

src/components/tab-bar/tab-bar.tsx
```diff
@@ -61,7 +64,7 @@ export class TabBar {
         this.setupMdc = false;
     }
 
-    public componentDidUnload() {
+    public disconnectedCallback() {
```
adrianschmidt, line 67: disconnectedCallback

src/components/tab-panel/examples/tab-panel-content.tsx
```diff
@@ -17,9 +17,15 @@ const LOAD_TIME = 1000;
...
 })
 export class TabPanelContentExample {
+    /**
+     * The tab that this component belongs to
+     */
     @Prop()
     public tab: Tab;
 
+    /**
+     * Emitted when the vote buttons is clicked to update the badge in the tab
```
adrianschmidt, line 27:
```suggestion
     * Emitted when the vote button is clicked to update the badge in the tab
```

src/components/tab-panel/tab-panel.tsx
```diff
@@ -43,7 +49,7 @@ export class TabPanel {
         this.tabs.forEach(this.setTabStatus);
     }
 
-    protected componentDidUnload() {
+    protected disconnectedCallback() {
```
adrianschmidt, line 52: disconnectedCallback

src/components/table/examples/birds.tsx
adrianschmidt: :zap:

src/components/button/examples/button.tsx
adrianschmidt: :zap:

src/components/portal/portal.tsx
jgroth: The Stencil documentation says that you should not use it as a destructor
> Called every time the component is disconnected from the DOM, ie, it can be dispatched more than once, DO not confuse with a "onDestroy" kind of event.

src/components/list/list-renderer.tsx
adrianschmidt: :zap:

src/components/portal/portal.tsx
adrianschmidt: Yeah, but is there something else to use then? We can't just leak instead… 😕

adrianschmidt: In the example, they certainly seem to clean up in `disconnectedCallback`…
https://stenciljs.com/docs/component-lifecycle#example

I think perhaps they just mean "don't assume the component will be gone forever after this". Or that comment was just forgotten and left in the docs when `componentDidUnload` was deprecated…

src/components/badge/format.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @param {number} value the number to abbreviate
+ *
+ * @returns {string} abbreviated number
+ */
+export function abbreviate(value: number): string {
+    if (typeof value !== 'number') {
+        return '';
+    }
+
+    const units = ['k', 'M', 'B', 'T'];
```
adrianschmidt, line 15: We really should make sure these are translated, as "billion" and "trillion" are *very* different things in English compared to the rest of the world… 😣

src/components/portal/portal.tsx
```diff
@@ -11,9 +11,10 @@ import { FlipModifier } from '@popperjs/core/lib/modifiers/flip';
  *
  * * Events might not bubble up as expected since the content is moved out to another DOM node
  * * Any styling that is applied to content from the parent will be lost, if the content is
- *   just another web compoent it will work without any issues
- * * When the node is moved in the DOM, `componentDidUnload`, `disconnectedCallback` and `connectedCallback`
- *   will be invoked, so `componentDidUnload` can not be used as a destructor (which is the wrong behavior anyway)
+ * just another web compoent it will work without any issues
+ * * When the node is moved in the DOM, `disconnectedCallback` and `connectedCallback`
+ * will be invoked, so if `disconnectedCallback` is used to do any tear-down, the
+ * appropriate setup will have to be done again on `connectedCallback`
```
adrianschmidt, line 17: I updated this a little bit, to kind of reflect our current view on `disconnectedCallback`…

src/components/slider/slider.tsx
jgroth: Haha yeah there was a typo, but your version is good as well :smile:

src/components/select/test/select.test-wrapper.tsx
jgroth: Yeah that should be fine :slightly_smiling_face:

src/components/select/select.tsx
jgroth: I don't think the spread operator will work here because `element.options` is a `DOMCollection` or something similar

src/components/portal/portal.tsx
jgroth: Yeah I think you can do cleanup in `disconnectedCallback`, but you have to assume that the component might be added to the DOM again and initialize everything again

PR #885, opened 2020-07-14
chore: remove use of deprecated config option `excludeSrc`

adrianschmidt:
Use the `exclude` option in `tsconfig.json` instead.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


adrianschmidt:
The obvious worry with this change is that the different builds wouldn't include everything that's necessary, or that the dist-build would include lots of stuff that it shouldn't.

- I've run the docz-app locally, and that seems to work.
- I've published the docs [here](http://lime-documentation-lime-elements.s3-website-eu-west-1.amazonaws.com/versions/0.0.0-dev/), and that also seems to work.
- I've built the production build locally, and tried it in the web client, and that seems to work.
- I've checked the size of the production build output from this branch vs master, and that had a little surprise for me:
  image
  It's over 3 MB smaller, and has almost 1000 fewer files! I'm about to look into exactly which files are no longer there, but since it's working when loaded in the web client, this seems like a positive thing 🙂

adrianschmidt:
Wow, the production build from master actually includes all the example files, and the production build from this branch doesn't. So that's really good news!

PR #886, opened 2020-07-15
fix(input-field): ensure that the label floats if value of the native input field changes

jensgustafsson:
# How to repro issue

1. Open the docz application and route to the docs for `limel-input`
2. Open dev-tools and inspect the native input field of an empty `limel-input`-field
3. Open the console and type `$0.value = 'whatever you want'`
4. Click outside the input field. The label will now appear over the text value and not above it as expected.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -231,7 +231,9 @@ export class InputField {
                     class={`
                         mdc-floating-label
                         ${
-                            this.value || this.isFocused
+                            this.value ||
+                            this.isFocused ||
+                            this.mdcTextField?.value
                                 ? 'mdc-floating-label--float-above'
                                 : ''
                         }


src/components/input-field/input-field.tsx
```diff
@@ -231,7 +231,9 @@ export class InputField {
                     class={`
                         mdc-floating-label
                         ${
-                            this.value || this.isFocused
+                            this.value ||
+                            this.isFocused ||
+                            this.mdcTextField?.value
```
adrianschmidt, line 236: This isn't a bug that affects normal use of the component. When `this.value` is set programmatically, the label is correctly updated. When setting the value of the native input using the console, not only does the label not update, but `this.value` also doesn't update.

Since programmatically setting the value on the native input field isn't the right way to use the component, I don't think we should spend time on making the component update the value if that's done (you can probably already make it update by triggering a `change` event on the native element), and if we don't change the underlying behaviour, I don't think we should change the behaviour for the label either, as that just camouflages the fact that the value isn't updated correctly.

setting input value through console

jensgustafsson:
This PR was created in order to fix a problem with flatpickr. Flatpickr automatically changes the raw input element inside limel-input-field. Thus, the input-field is not aware off it. I will solve this issue by creating a separate input field for the input field and mirror the data instead. Closing... :)

adrianschmidt:
Can the branch be deleted?

PR #891, opened 2020-07-17
enable linting of config files again

adrianschmidt:
## Review:


adrianschmidt:
@jgroth I made some updates here to. Ping fyi.

jgroth:
Great!
However, I suggest we keep the name of the main config file to `.eslintrc.js`. Editor plugins will have an easier time finding this file, otherwise you will probably have to configure it yourself and having the "default" name will just make it easier :slightly_smiling_face:

jgroth:
Created a PR for the filename issue :slightly_smiling_face: 
#904

I noticed that we could make it a bit more simple by just moving one of the config files to the `src` directory as mentioned here 
https://eslint.org/docs/user-guide/configuring#configuration-cascading-and-hierarchy
It would allow us to just run the linter once instead of both for the src and config files :slightly_smiling_face:

Issue #852, opened 2020-07-02
Tab-bar crashes when previous tabs are undefined


JohannesHansen:
## Current behavior
If the second argument to `tabsChanged` is undefined, the component crashes and renders the text "undefined". This seems to happen quite frequently when using the component in a react wrapper.

## Expected behavior
Component shouldn't crash and should not render "undefined"

## Environment

- **lime-elements** version: 26.35.0
- Framework used: React

adrianschmidt:
@jgroth I'm almost certain I've seen this as an issue on the board during the last cycle? Wasn't this fixed already? Or am I confusing it with something similar?

jgroth:
I'm not sure I have seen it here, but there has been a similar issue in another component. It should be quite easy to fix by giving default values to the arguments, i.e.
```ts
protected tabsChanged(newTabs: Tab[] = [], oldTabs: Tab[] = [])
```

adrianschmidt:
I'm putting this one straight on the Feature board, since it seems like a good task for someone to grab during summer.

PR #894, opened 2020-07-22
fix(tab-bar): fix tab-bar crashes when previous tabs are undefined

Befkadu1:
fix: Lundalogik/lime-elements#852

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tab-bar/tab-bar.tsx
+++ b/src/components/tab-bar/tab-bar.tsx
@@ -26,7 +26,7 @@ export class TabBar {
      * List of tabs to display
      */
     @Prop()
-    public tabs: Tab[];
+    public tabs: Tab[] = [];
 
     /**
      * Emitted when a tab has been changed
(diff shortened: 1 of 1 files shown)


src/components/tab-bar/tab-bar.tsx
```diff
@@ -26,7 +26,7 @@ export class TabBar {
      * List of tabs to display
      */
     @Prop()
-    public tabs: Tab[];
+    public tabs: Tab[] = [];
```
adrianschmidt, line 29: I tested the PR by updating one of the examples to initially have supply `undefined` and then change to an array of tabs when I clicked a button. The component still crashed, before it got to `tabsChanged`. So I searched the file for `.map`, and there's another map function in the render function (line 78). By adding a default value here too, the crashing is resolved.

I tried it with only this default value, and no default values for the `tabsChanged` function, and it still didn't crash, even when the tabs where updated. But I think it's more prudent to have default values in both places, just in case.

adrianschmidt:
I updated the commit message and added `fix #852`:

image

By adding that, semantic-release automatically adds a link to the Issue in the changelog entry, like this:

image

PR #895, opened 2020-07-23
fix(input-field): correct placement of label for input-field type of …

### LINKED ISSUE #722: Missing "upper label" for input field of type textarea
## Current behavior
Input Field of Type Textarea
When we prefill the input field with some text the "upper label", or whatever it is called, is not visible.

Skärmavbild 2020-04-22 kl  14 34 04

When we focus the input field, the upper label appears.

Skärmavbild 2020-04-22 kl  16 18 43

## Expected behavior

The upper label should be visible from the beginning without focus the input field.

## Environment

- **lime-elements** version: v26.18.0
- lime-dev-app

Befkadu1:
…textarea  when the value is prefilled

fix #722

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.scss
+++ b/src/components/input-field/input-field.scss
@@ -26,7 +26,10 @@
     @include mdc-states-focus-opacity(0, true);
 
     .mdc-floating-label {
-        top: $floating-label-top-value;
+        &:not(.textarea-label) {
+            top: $floating-label-top-value;
+        }
+
         right: pxToRem(
             8
         ); //This is a hack to force the label truncate when container is too little. Otherwise due to position: absolute it won't. Kia
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -260,6 +260,12 @@ export class InputField {
             'mdc-text-field--required': this.required,
         };
 
+        const labelClassList = {
+            'mdc-floating-label': true,
+            'textarea-label': true,
+            'mdc-floating-label--float-above': !!this.value || this.isFocused,
+        };
+
         return [
             <div class={classList}>
                 {this.renderCharacterCounter()}
@@ -278,7 +284,7 @@ export class InputField {
                 <div class="mdc-notched-outline">
                     <div class="mdc-notched-outline__leading" />
                     <div class="mdc-notched-outline__notch">
-                        <label htmlFor="textarea" class="mdc-floating-label">
+                        <label htmlFor="textarea" class={labelClassList}>
                             {this.label}
                         </label>
                     </div>


src/components/input-field/input-field.tsx
```diff
@@ -260,6 +260,11 @@ export class InputField {
             'mdc-text-field--required': this.required,
         };
 
+        const labelClassList = {
+            'mdc-floating-label textarea-label': true,
```
adrianschmidt, line 264: I would like this tiny change, just to make it a little easier to read:

```suggestion
            'mdc-floating-label': true,
            'textarea-label': true,
```

src/components/input-field/input-field.scss
```diff
@@ -26,7 +26,10 @@
...
 
     .mdc-floating-label {
-        top: $floating-label-top-value;
+        &:not(.textarea-label) {
+            top: $floating-label-top-value;
+        }
+
         right: pxToRem(
             8
         ); //This is a hack to force the label truncate when container is too little. Otherwise due to position: absolute it won't. Kia
```
adrianschmidt, line 35: There's still this small visual bug left, but it's been on master at least since May, and probably longer, and I don't think we should try to fix it in this PR, because this PR fixes a much uglier issue. So I think it's better to get this one merged, and look into the remaining bug separately.

Video showing the remaining bug:
Small visual bug in textarea

I'm putting this comment on this line, because while this looks like it might be the cause, I'm pretty sure it's not, because I've tried removing it, and that makes no difference.

The thing is that something, presumably something in the MDC component, is setting the width of the element, and setting it to be too small. Once the input is blurred, the width is set again, this time to the correct value. I can't quite get my head around why setting this width would be necessary at all, but I still have an idea for what the reason it's set too small *might* be. We have changed the font used, from the default *Open Sans* to *Roboto*. I don't think we've changed the default font-size (although I'm not sure), but I think we might have fiddled with the amount the label is shrunk when floating, making it slightly bigger than it was by default. Again, I'm not sure, but I think I have some vague memory of possibly doing that at some point… If MDC is setting the width by measuring the current width, and then calculating what the new width ought to be, using values for how much it's supposed to shrink that it's getting from somewhere other than our CSS, then that might explain the bug. It might also mean it's *very* difficult to fix… 😕

Befkadu1: For some reason the width of `mdc-notched-outline__notch` is set

adrianschmidt: Ping @Kiarokh and @magnusfagerlund for opinions on whether we should open an issue for this, and spend time trying to fix it, or whether we consider it small enough that we'll just leave it be until a customer reports it?

Kiarokh: @adrianschmidt , yes it's MD that is dynamically putting a `width` on the `mdc-notched-outline__notch` that contains the label. And it miscalculates the width. All because they want to remove the `border-top` from that middle section, when the label is transformed  to the top position. (They don't put background color on the label itself to cover/hide the border below, on the textarea [or any other "outlined" input field]. This is because they want these elements to work fine on colored backgrounds too.)

The funny thing is that the content of the `<label>` is already slightly sticking out of that `mdc-notched-outline__notch` div (which may be the reason for miscalculation of the width, when the field is focused). But because of the `width: auto` it's not clipped out in this mode. 
image image

When the label is moved to top, it will also get a `transform: scale(0.75)`. Since `scale` is just a visual effect and doesn't really change the physical `width` of the element, they cannot use the `width: auto` anymore; that will generate this effect (some empty space on the right side of the label):
image and that's why they have to apply the width dynamically. The problem is in their calculations.

_____

Are we using the latest version of their component? This doesn't happen [in their examples](https://material-components.github.io/material-components-web-catalog/#/component/text-field?type=outlined). 

This can be slightly improved with some simple CSS hacks (`.mdc-notched-outline__notch { wdith: auto !important;}` for example), but it won't look so good as I show above. I'd say it's not worth fixing via CSS hacks.

It's better to see whether we can update our components and if that fixes this issue. 

At some point, I want to switch all our other text inputs to the outlined version of MD. This is when we do the Edit mode / View mode. Perhaps we can look into this again when we do that big change.

Kiarokh: another weird thing is that the label is suppose to truncated when the width is smaller than the label. But somehow, that width is not small enough to trigger truncate :/

adrianschmidt: > Are we using the latest version of their component?

No. We're using version 4, and the latest is 7.0.0. Updating to a new major version tends to be a bit of a hassle, since usually we need to make a fair number of updates in lime-elements to make all components work again.

PR #901, opened 2020-08-07
fix(tab-bar): make sure scrollArea &  mdcTabBar exist before removing…

Befkadu1:
… the element

make sure that scrollArea and mdcTabBar exist before removing the element when destroying the
component

fix Lundalogik/crm-feature#1238

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/tab-bar/tab-bar.tsx
```diff
@@ -66,6 +66,10 @@ export class TabBar {
 
     // eslint-disable-next-line @stencil/own-methods-must-be-private
     public componentDidUnload() {
+        if (!this.setupMdc) {
```
jgroth, line 69: This will almost always be `false` so by having this condition we will never run the teardown when the component is destroyed.

It would be better to have checks for `this.scrollArea` and `this.mdcTabBar` in the `teardown` method.

PR #913, opened 2020-08-19
feat(table): add row selection feature on table view

### LINKED ISSUE #912: Add Select Row feature on table view
Row selection allows users to select row that you can then take action on. This allows rows to be selected  by clicking on a row.

Befkadu1:
Row selection allows users to select row that you can then take action on. This allows rows to be
selected by clicking on a row.

fix #912

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


adrianschmidt:
I think it's fairly likely that, some time in the future, we will want to be able to select multiple rows by checking checkboxes. If we ever do, we will probably regret calling *this* functionality "select"… 🤔

There is no visual indication that a row is selected in the example. If this is meant only as a step towards being able to click a row to open the related-view, perhaps it's better to call it something like "clickable"?

Thoughts @Kiarokh @jgroth (and anyone else)?

src/components/table/table.tsx
```diff
@@ -71,6 +78,18 @@ export class Table {
...
+     * Emitted when a raw is selected
+     */
+    @Event()
+    public selectRow: EventEmitter<number>;
+
+    /**
+     * Emitted when `mode` is `local` and a new page has been set
+     */
+    @State()
+    public prevSelectedRawId: number;
```
jgroth, line 91: Right now we allow the user to select multiple rows since the `selectable` property is a `number`, but we only store a single row here. I think we need to store all the rows that are selected :thinking: 

Also, I think it should be possible to have rows pre-selected when the table is created, so this should probably be a `@Prop` instead. However, in the `limel-list` component `selected` is part of the data that is given to the component so we might want to do something similar here. I'm not sure what the best approach is :thinking:

src/components/table/table.tsx
```diff
@@ -71,6 +78,18 @@ export class Table {
     @Event()
     public load: EventEmitter<TableParams>;
 
+    /**
+     * Emitted when a raw is selected
+     */
+    @Event()
+    public selectRow: EventEmitter<number>;
```
jgroth, line 85: We should also have an event for when rows have been deselected. In the `limel-list` component all this is handled in the `change` event. If we want to be consistent, we might want to rename this.

jgroth (changes requested):

jgroth:
> I think it's fairly likely that, some time in the future, we will want to be able to select multiple rows by checking checkboxes. If we ever do, we will probably regret calling _this_ functionality "select"… 
> 
> There is no visual indication that a row is selected in the example. If this is meant only as a step towards being able to click a row to open the related-view, perhaps it's better to call it something like "clickable"?
> 
> Thoughts @Kiarokh @jgroth (and anyone else)?

Yes I agree. The code here actually allows for multiple rows to be selected, but there is no visual indication of selected rows. We need to add that :slightly_smiling_face:

adrianschmidt:
> Yes I agree. The code here actually allows for multiple rows to be selected, but there is no visual indication of selected rows. We need to add that 🙂

Ok, so this is actually about selecting rows then? If so, I don't think a row should be selected just from clicking anywhere on the row. I think there should be a checkbox in a column at the far left of the row. Otherwise we will get conflicts between clicking to select and clicking to open in related-view.

Befkadu1:
> > Yes I agree. The code here actually allows for multiple rows to be selected, but there is no visual indication of selected rows. We need to add that 🙂
> 
> Ok, so this is actually about selecting rows then? If so, I don't think a row should be selected just from clicking anywhere on the row. I think there should be a checkbox in a column at the far left of the row. Otherwise we will get conflicts between clicking to select and clicking to open in related-view.
> 
> Edit: Unless I'm completely missing something here 😄

It is possible to allow one row at a time. What I am about to do is to set `selectable = 1`.  That is what I thought but if we need multi row select feature we will do that in the future.

Befkadu1:
> clicking to select and clicking to open in related-view.

This PR does Clicking to select and clicking to open in related-view for now. Isn't it same in List view? If you click one of the object in the list view, it means it opens the related view. Correct me if I am wrong

adrianschmidt:
> > clicking to select and clicking to open in related-view.
> 
> This PR does Clicking to select and clicking to open in related-view for now. Isn't it same in List view? If you click one of the object in the list view, it means it opens the related view. Correct me if I am wrong

Hmm… that's true… 🤔

I'm not sure it was a good idea to use "selected" for highlighting that an item is shown in the related-view there either, but I guess it won't be a problem as long as we're only *either* opening the item in the related-view, *or* allowing multi-select. As soon as we want to combine those two behaviours, I think it's going to be really annoying if clicking an item both selects it and opens it in the related-view.

Befkadu1:
> > > clicking to select and clicking to open in related-view.
> > 
> > 
> > This PR does Clicking to select and clicking to open in related-view for now. Isn't it same in List view? If you click one of the object in the list view, it means it opens the related view. Correct me if I am wrong
> 
> Hmm… that's true… 🤔
> 
> I'm not sure it was a good idea to use "selected" for highlighting that an item is shown in the related-view there either, but I guess it won't be a problem as long as we're only _either_ opening the item in the related-view, _or_ allowing multi-select. As soon as we want to combine those two behaviours, I think it's going to be really annoying if clicking an item both selects it and opens it in the related-view.


Or we can have this feature. Selecting and clicking is different here

Screenshot 2020-08-19 at 14 32 37

src/components/table/examples/table-selectable-row.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { persons, columns, Person } from './persons';
+
+@Component({
+    tag: 'limel-example-table-selectable-row',
+    styleUrl: 'table.scss',
+    shadow: true,
+})
+export class TableExampleSelectableRow {
+    @State()
+    private selectedRaw: Person;
```
hannahu, line 12: R**a**w and on the lines 30 and 36 as well :)

Befkadu1: Haha

src/components/table/examples/persons.tsx
```diff
@@ -0,0 +1,134 @@
...
+        stamina: '40%',
+        placeOfBirth: 'Orasas',
+        sign: '🔮',
+        dateOfBirth: '05/10/1945',
+        role: 'Chief Entertainment Officer',
+        activated: false,
+    },
+];
+
+export const columns: Column[] = [
```
jgroth, line 124: This is quite useful to have in the example itself since it configures how the table is used. We can use the same data in several tables but the config might differ :slightly_smiling_face:

src/components/table/examples/table-selectable-row.tsx
```diff
@@ -0,0 +1,38 @@
+import { Component, h, State } from '@stencil/core';
+import { Column } from '../table.types';
+import { persons, columns, Person } from './persons';
+
+@Component({
+    tag: 'limel-example-table-selectable-row',
+    styleUrl: 'table.scss',
+    shadow: true,
+})
+export class TableExampleSelectableRow {
```
jgroth, line 10: We should rename this example since the functionality now is called "Activate" instead of "Select"

src/components/table/table.scss
```diff
@@ -69,6 +69,11 @@ $tabulator-arrow-color-active: var(--lime-turquoise);
         background-color: #fff;
     }
 
+    &.tabulator-selected,
+    &.tabulator-selected:hover {
+        background-color: rgba(#26a69a, 0.16) !important;
```
jgroth, line 74: It would be great if we can do this without `!important` :slightly_smiling_face:

src/components/table/table.tsx
```diff
@@ -71,6 +71,12 @@ export class Table {
     @Event()
     public load: EventEmitter<TableParams>;
 
+    /**
+     * Emitted when a row is activated
+     */
+    @Event()
+    public activate: EventEmitter<any>;
```
jgroth, line 78: We should not use `any` here, it does not help the consumer in any way in knowing what data will be part of the event

src/components/table/table.tsx
```diff
@@ -205,6 +213,37 @@ export class Table {
         this.changePage.emit(page);
     }
 
+    private onClickRow(event: MouseEvent, row: Tabulator.RowComponent): void {
+        // deselect all activated rows
+        this.tabulator.deselectRow();
+        row.select();
```
jgroth, line 219: We can not actually select the row in the table like this, since that will conflict when we actually want to select rows later. This is not a "select" functionality, just an "activate" functionality

src/components/table/table.tsx
```diff
@@ -205,6 +213,37 @@ export class Table {
...
+        const object = row.getData();
+
+        const activatedObject: any = this.data.find((item: any) => {
+            return !!item.activated;
```
jgroth, line 223: We can not do it like this. This forces the consumer to have a field in their data that is named `activated`. If we want to store the `activated` state with the data, we need an interface for the data and also nest the actual data deeper in that structure since the consumer might have a field named `activated` that is used for something else. Also, what happens if the consumer sets `activated` to true on all the rows? As a consumer I would expect all rows to be activated, but nothing will happen. Even if I just set one of the rows to active when data is given to the table, it will not activate the row since we only have logic for it when we are clicking.

How about adding a new `@Prop` to the table that tells us what row is currently active? That way we can have the user activate a row by changing the input, and the event will tell the user when a new row has been activated.

jgroth (changes requested):

src/components/table/examples/table-selectable-row.tsx
Kiarokh: Rawwww 🐯

src/components/table/table.scss
Kiarokh: First: change the class name from `.tabulator-selected` to `.activated`. I think it's good to remove the word `tabulator` from the class name, so we know in future that this is not a native Tabulator function or something.

Kiarokh: The `!important" is used because there is already a `background-color` style for normal `:hover` on line 60.

So, either we should make this rule more specific here. For example: 

````css
#tabulator-table {
    .tabulator-table {
        .tabulator-row {
            &.tabulator-selected,
            &.tabulator-selected:hover {
                 background-color: rgba(#26a69a, 0.16);
            }
        [...]
````
or add another identifier for tables that have this capability ON, e.g. "`.has-activateable-rows`" and make the CSS bound to that.

**OR this is preferred solution according to me:**
refine the code for hover on line 57, like this:
````css
.tabulator-row {
    &.tabulator-selectable {
        transition: background-color 0.2s ease;
        &:not(.activated) {
            &:hover {
                background-color: $table-header-background-color;
            }
        }
    }
````
Note: Test the above code. I guess I've written it correctly, but make sure to test.

File: src/components/table/table.tsx (hunk 8 above)
Comment (Kiarokh) [reply src/components/table/table.tsx:219]: * a "selected" row (using checkboxes) can be "activated" a well, and activating shouldn't deselect it.
* multiple rows can be "selected"
* only 1 row at a time can be "activated". Activating a row must deactivate any other row that is currently active

File: src/components/table/partial-styles/tabulator-custom-styles.scss [hunk 10]
```diff
@@ -3,9 +3,13 @@
         &.tabulator-selectable {
             @include is-flat-clickable();
             &:hover {
-                background-color: #fff;
                 z-index: 1;
             }
+            &:not(.activated) {
```
jgroth, line 8:
```suggestion
            &:not(.active) {
```

src/components/table/table.scss
```diff
@@ -57,8 +57,10 @@ $tabulator-arrow-color-active: var(--lime-turquoise);
 .tabulator-row {
     &.tabulator-selectable {
         transition: background-color 0.2s ease;
-        &:hover {
-            background-color: $table-header-background-color;
+        &:not(.activated) {
```
jgroth, line 60:
```suggestion
        &:not(.active) {
```

src/components/table/table.tsx
```diff
@@ -205,6 +226,19 @@ export class Table {
         this.changePage.emit(page);
     }
 
+    private onClickRow(_, row: Tabulator.RowComponent): void {
+        this.activeRow = row.getData();
+        this.activate.emit(this.activeRow);
+    }
+
+    private reformat(row) {
```
jgroth, line 234:
```suggestion
    private formatRow(row: Tabulator.RowComponent) {
```

src/components/table/table.tsx
```diff
@@ -205,6 +226,19 @@ export class Table {
...
+        if (this.activeRow === row.getData()) {
+            row.getElement().classList.add('activated');
+        } else {
+            row.getElement().classList.remove('activated');
```
jgroth, line 238:
```suggestion
            row.getElement().classList.add('active');
        } else {
            row.getElement().classList.remove('active');
```

src/components/table/table.tsx
```diff
@@ -98,6 +112,11 @@ export class Table {
         this.tabulator = new TabulatorTable(table, options);
     }
 
+    @Watch('activeRow')
+    public activatedRowChanged() {
+        this.tabulator.redraw(true);
```
jgroth, line 117: If you scroll down in a table and click on a row, the table will scroll to the top. I *guess* that's its because of this line :disappointed:

jgroth (commented):
Code looks good, just some naming suggestions, and the annoying scroll bug :disappointed: 

Maybe we should add a `deactivate` as well that triggers when you click on an active row?

PR #914, opened 2020-08-19
Spacier tables

Kiarokh:
Trying to make the new tables look more similar to the old tables, as internally people experience the old tables as less boring 🙃
This PR also removes a layout misalignment in the tables with custom styles.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/partial-styles/tabulator-custom-styles.scss
+++ b/src/components/table/partial-styles/tabulator-custom-styles.scss
@@ -19,11 +19,20 @@
 
     .tabulator-table {
         background-color: transparent;
-        padding: pxToRem(8);
+    }
+
+    // This provides space for box-shadow effect. Otherwise the overflow properties will cut them out
+    .tabulator-headers {
+        padding: 0 pxToRem(8);
+    }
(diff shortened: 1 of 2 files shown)


BregenzerK (approved):
looks good, just double-checking: is it on purpose that the box-shadow effect is set for the low-density tables? wouldn't the effect be nice on the "regular" tables as well? this won't make a difference in the webclient, since we're (currently) not using the low-density tables there

Kiarokh:
Good question @BregenzerK 
You answer is in Lime Element: 
image

The normal hover style (row gets a darker BG color) improves the experience a bit for users. You can more easily follow the row as you hover on it and see all cells that belong to what you hovered. 

But this other shadow style indicates for the user that they can actually click to get trigger an action.

BregenzerK (approved):

PR #920, opened 2020-08-24
Invert eslint config

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


.eslintrc.js
```diff
@@ -15,7 +15,6 @@ module.exports = {
     parserOptions: {
         ecmaVersion: 2020,
         sourceType: 'module',
-        project: './tsconfig.lint.json',
```
jgroth, line 18: Is this not required? :thinking: The default seems to be `undefined` and the docs says the following 
> This setting is required if you want to use rules which require type information

adrianschmidt: It's required for some rules (which will tell you if it's missing, and fail the lint), but we're only using those rules in the overrides that have a project defined 👍

I found this as the recommended way to lint both files included in a project and files not included, since once set as the default, there's no way to remove the `project` setting from an override, only change it to something different.

jgroth (approved):

PR #931, opened 2020-08-31
feat(chip-set): add delete all chips button

maltja:
fix Lundalogik/crm-feature#1329

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -150,6 +150,9 @@ export class ChipSet {
         this.inputFieldOnChange = this.inputFieldOnChange.bind(this);
         this.handleKeyDown = this.handleKeyDown.bind(this);
         this.inputHidden = this.inputHidden.bind(this);
+        this.handleDeleteAllIconClick = this.handleDeleteAllIconClick.bind(
+            this
+        );
     }
 
     /**
@@ -228,6 +231,7 @@ export class ChipSet {
         const classes = {
             'mdc-chip-set': true,
             disabled: this.disabled,
+            'mdc-text-field--with-trailing-icon': true,
         };
         if (this.type) {
             classes[`mdc-chip-set--${this.type}`] = true;
@@ -293,6 +297,7 @@ export class ChipSet {
                     'mdc-text-field': true,
                     'mdc-text-field--disabled': this.readonly || this.disabled,
                     'force-invalid': this.isInvalid(),
+                    'chips-exists': this.value.length !== 0,
                 }}
                 onClick={this.handleTextFieldFocus}
             >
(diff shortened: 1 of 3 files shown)


src/components/chip-set/chip-set.tsx
```diff
@@ -295,6 +304,8 @@ export class ChipSet {
                     'force-invalid': this.isInvalid(),
                 }}
                 onClick={this.handleTextFieldFocus}
+                onMouseEnter={this.onMouseEnter}
+                onMouseLeave={this.onMouseLeave}
```
jgroth, line 308: I don't think we need these :slightly_smiling_face: 
We should be able to hide the button with only CSS rules and not have to rely on any JS code

jgroth (changes requested):

jgroth (dismissed):
Good work! :confetti_ball:

jgroth (approved):

adrianschmidt:
@jgroth Methinks you didn't test keyboard navigation for this… 😉

You can tab to the clear-button, but there doesn't seem to be any way to actually activate it. I've tried both \<enter\>  and \<space\>.

image

Also, if you don't hover the component when tabbing to the clear-button, it looks a bit broken:

image

I'd be fine with just taking this button out of the tab-flow, since you can already delete the chips by pressing backspace when the input is focused, and unless you have an unreal amount of chips selected, that's pretty fast and easy. That also avoids the downside of having to tab twice to get past this component. What do you say @Kiarokh?

Regardless, it shouldn't be possible to tab to the button when there are no chips selected.

adrianschmidt:
Found some more small issues:

When disabled, it's still possible to tab to the button:

image

When read-only, it's not only possible to tab to the button, but it also becomes visible when any chip is hovered:

image

(There's also a small issue with the label cutting part of the chips, both when disabled and when read-only, but I don't think that's got anything to do with this PR. @Kiarokh: If I recall correctly, you moved the label a little bit a while back? It might have happened then. Can you take a look at it?)

Kiarokh:
@adrianschmidt I think:
* user should be able to tab to the (x) button
* (x) should be none-existing when there are no chips as you said
* (x) should have a `title` and `area-label`

adrianschmidt:
Ok. I'll add the above in an issue.

jgroth:
> @jgroth Methinks you didn't test keyboard navigation for this… 

That is a fair assumption! :smile:

Kiarokh:
@adrianschmidt I think we have addressed issues above here: https://github.com/Lundalogik/lime-elements/pull/938

Kiarokh:
oops, except the keyboard interactions (**Enter** / Space)

PR #932, opened 2020-09-01
fix(tab-bar): save tab state by making it mutable

jgroth:
If tabs are changed rapidly, Stencil will schedule the update of the tabs prop. This will leave us
in a state were the tabs prop still have the old value of the tabs when a new tab is selected, and
thus we will not find the old active tab and cannot deactivate it. To fix this, the tabs prop is
made mutable and we store any changes made to the tabs before we emit the changed tabs in the
`changeTab` event.

fix Lundalogik/crm-feature#1327

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/tab-bar/tab-bar.tsx
+++ b/src/components/tab-bar/tab-bar.tsx
@@ -10,7 +10,8 @@ import {
 import { MDCTabBar, MDCTabBarActivatedEvent } from '@limetech/mdc-tab-bar';
 import { strings } from '@limetech/mdc-tab-bar/constants';
 import { Tab } from './tab.types';
-import { isEqual } from 'lodash-es';
+import { isEqual, difference } from 'lodash-es';
+import { setActiveTab } from './tabs';
 
 const { TAB_ACTIVATED_EVENT } = strings;
 const SCROLL_DISTANCE_ON_CLICK_PX = 150;
@@ -25,7 +26,7 @@ export class TabBar {
     /**
      * List of tabs to display
      */
-    @Prop()
+    @Prop({ mutable: true })
     public tabs: Tab[] = [];
 
     /**
@@ -172,13 +173,13 @@ export class TabBar {
 
     private handleTabActivated(event: MDCTabBarActivatedEvent) {
         const index = event.detail.index;
-        const oldSelectedTab = this.tabs.find((tab) => tab.active === true);
+        const newTabs = setActiveTab(this.tabs, index);
 
-        if (oldSelectedTab) {
-            this.changeTab.emit({ ...oldSelectedTab, active: false });
-        }
+        difference(newTabs, this.tabs).forEach((tab: Tab) => {
+            this.changeTab.emit(tab);
+        });
 
-        this.changeTab.emit({ ...this.tabs[index], active: true });
+        this.tabs = newTabs;
     }
 
     private handleScroll() {
--- a/src/components/tab-bar/tabs.ts
+++ b/src/components/tab-bar/tabs.ts
@@ -1,0 +1,26 @@
+import { Tab } from './tab.types';
+
+/**
+ * Set a tabs `active` state to true in a list of tabs. The previous tab with
+ * `active` set to true will have it set to `false` instead.
+ *
+ * @param {Tab[]} tabs list of tabs
+ * @param {number} index the index of the tab to set to active
+ *
+ * @returns {Tab[]} a copy of the list of tabs with the changed tabs replaced
+ */
+export function setActiveTab(tabs: Tab[], index: number): Tab[] {
+    const oldSelectedTabIndex = tabs.findIndex((t) => t.active === true);
+    const result = [...tabs];
+
+    if (oldSelectedTabIndex !== -1) {
+        result[oldSelectedTabIndex] = {
+            ...tabs[oldSelectedTabIndex],
+            active: false,
+        };
+    }
+
+    result[index] = { ...tabs[index], active: true };
+
+    return result;
+}
--- a/src/components/tab-bar/tab-bar.e2e.ts
+++ b/src/components/tab-bar/tab-bar.e2e.ts
@@ -1,0 +1,72 @@
+import {
+    newE2EPage,
+    E2EPage,
+    E2EElement,
+    EventSpy,
+} from '@stencil/core/testing';
+import { Tab } from './tab.types';
+
+describe('limel-tab-bar', () => {
+    let page: E2EPage;
+    let component: E2EElement;
+    const tabs: Tab[] = [
+        {
+            id: 'foo',
+            active: true,
+        },
+        {
+            id: 'bar',
+        },
+        {
+            id: 'baz',
+        },
+    ];
+
+    beforeEach(async () => {
+        page = await newE2EPage();
+        await page.setContent('<limel-tab-bar></limel-tab-bar>');
+        component = await page.find('limel-tab-bar');
+        component.setProperty('tabs', tabs);
+        await page.waitForChanges();
+    });
+
+    describe('when bar tab is clicked', () => {
+        let spy: EventSpy;
+        beforeEach(async () => {
+            spy = await page.spyOnEvent('changeTab');
+            const buttons = await page.findAll('limel-tab-bar >>> button');
+            await buttons[1].click();
+        });
+
+        it('emits an event with the old inactive tab', () => {
+            expect(spy).toHaveReceivedEventTimes(2);
+            const event = spy.events.find((e) => e.detail.id === 'bar');
+            expect(event.detail).toEqual({
+                id: 'bar',
+                active: true,
+            });
+        });
+
+        it('emits an event with the active tab', () => {
+            expect(spy).toHaveReceivedEventTimes(2);
+            const event = spy.events.find((e) => e.detail.id === 'foo');
+            expect(event.detail).toEqual({
+                id: 'foo',
+                active: false,
+            });
+        });
+    });
+
+    describe('when foo tab is clicked', () => {
+        let spy: EventSpy;
+        beforeEach(async () => {
+            spy = await page.spyOnEvent('changeTab');
+            const buttons = await page.findAll('limel-tab-bar >>> button');
+            await buttons[0].click();
+        });
+
+        it('does not emit an event', () => {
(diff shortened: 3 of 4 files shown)


src/components/tab-bar/tab-bar.tsx
```diff
@@ -175,12 +175,25 @@ export class TabBar {
...
+        this.changeTab.emit(tab);
+    }
+
+    private mapTab = (newTab: Tab) => (oldTab: Tab) => {
+        if (newTab.id === oldTab.id) {
+            return newTab;
+        }
+
+        return oldTab;
+    };
```
adrianschmidt, line 195: 🤯

Ok, so after a few minutes of trying to parse this code, what I think it's basically doing is setting the `active` property of the previously selected tab to `false` and that of the newly selected tab to `true`, and it also makes sure to recreate the whole array, in order to trigger a re-render…

Am I somewhat close?

jgroth: I don't think this by itself will trigger a re-render, we just need to make sure that our saved tabs (`this.tabs`) are updated in case a new tab is clicked before `this.tabs` have had a chance to be updated by the consumer. I think that the consumer update is synchronous (i.e. the consumer _thinks_ that they have updated the props before a new tab is selected), but it seems that Stencil queues up some work so the data inside the component is not updated until later :disappointed: When this happens we will be in a state were multiple tabs are active at once

adrianschmidt: Yeah, I understand the problem. I'm just not quite sure I understand what this solution is actually doing. I find it incredibly hard to follow this short piece of code 😅

Would it be possible to refactor to something that might be a little easier to follow? If not, please add some explanatory comments, to give the rest of us a fighting chance 😄

jgroth: The solution saves the new state of the tabs in `this.tabs` so that we can be sure that it is up to date in case there is a new click before they are updated by the consumer.

I'm not sure how to refactor it, I was struggling a bit with the naming of the `mapTab` function. Maybe if we find a better name for that it will be more clear? :thinking: What part of the code do you think needs to be clarified? :slightly_smiling_face: 
We could also wrap line 185 above in another function with a name that perhaps could make it easier to understand

adrianschmidt: If we can write some unit tests for `handleTabActivated`, I could take a look at refactoring it. Also, if we can't refactor it, tests would be really helpful for explaining how it works. It looks to me like we could break it out into a pure function pretty easily, and maybe put it in its own file and import it here?

Do you wanna, or should I? 🙂

jgroth: Yeah I can try and see if I can write some tests for it. I'm not sure how easy it will be making this a pure function though, since it both emits an event and saves state data :thinking: But maybe I'm misunderstanding something :smile:

adrianschmidt: Ah, yeah, it emits an event too…

*(later…)*

I think this would work:

```ts
    private handleTabActivated(event: MDCTabBarActivatedEvent) {
        const results = this.inner(event);
        results.forEach(({tabs, changedTab}) => {
            this.changeTab.emit(changedTab);
            this.tabs = tabs;
        });
    }
    
    private inner(event: MDCTabBarActivatedEvent) {
        const index = event.detail.index;
        const oldSelectedTab = this.tabs.find((tab) => tab.active === true);
        let resultForOld: { tabs: Tab[]; changedTab: Tab; };

        if (oldSelectedTab) {
            resultForOld = this.replaceTab({ ...oldSelectedTab, active: false });
        }

        const resultForNew = this.replaceTab({ ...this.tabs[index], active: true });
        
        if (resultForOld) {
            return [resultForOld, resultForNew];
        }
        return [resultForNew];
        
    }

    private replaceTab(changedTab: Tab) {
        const tabs = this.tabs.map(this.mapTab(changedTab));
        return {tabs, changedTab};
    }

    private mapTab = (newTab: Tab) => (oldTab: Tab) => {
        if (newTab.id === oldTab.id) {
            return newTab;
        }

        return oldTab;
    };
```

Then the `inner` function should be easy to test.

jgroth: It's still accessing `this.tabs` though which makes it impure :slightly_frowning_face: 
I'm not sure if a unit test here would help with understanding the root cause of this issue, and I'm afraid that we make the code harder to understand if we refactor it around that :thinking: I've been trying to write an E2E test but it seems to be very flaky and Stencil does not seem to let me write the tests the way I want to :disappointed: I will have another go at it after lunch
I have at least renamed the functions a bit, so maybe it's easier to read now? Please let me know :slightly_smiling_face:

adrianschmidt: > It's still accessing `this.tabs` though which makes it impure 🙁

Ah, crap, I missed that. But that should be easy to fix. Hang on!

adrianschmidt:
```ts
    private handleTabActivated(event: MDCTabBarActivatedEvent) {
        const results = this.inner(event, this.tabs);
        results.forEach(({tabs, changedTab}) => {
            this.changeTab.emit(changedTab);
            this.tabs = tabs;
        });
    }

    private inner(event: MDCTabBarActivatedEvent, incomingTabs: Tab[]) {
        const index = event.detail.index;
        const oldSelectedTab = incomingTabs.find((tab) => tab.active === true);
        let resultForOld: { tabs: Tab[]; changedTab: Tab; };

        if (oldSelectedTab) {
            resultForOld = this.replaceTab(incomingTabs, { ...oldSelectedTab, active: false });
        }

        incomingTabs = resultForOld?.tabs || incomingTabs;

        const resultForNew = this.replaceTab(incomingTabs, { ...incomingTabs[index], active: true });

        if (resultForOld) {
            return [resultForOld, resultForNew];
        }
        return [resultForNew];

    }

    private replaceTab(incomingTabs: Tab[], changedTab: Tab) {
        const tabs = incomingTabs.map(this.mapTab(changedTab));
        return {tabs, changedTab};
    }

    private mapTab = (newTab: Tab) => (oldTab: Tab) => {
        if (newTab.id === oldTab.id) {
            return newTab;
        }

        return oldTab;
    };
```

adrianschmidt: Note, I'm not saying the above is better. It's just that it's testable. Once there are tests for it, I can refactor.

jgroth: Ok, I refactored it again and added both some E2E tests and some unit tests. I really don't feel that the tests illustrates the bug that this PR is solving though :smile: But at least there are some tests now! :partying_face: 
Please let me know if you think this is better or not :slightly_smiling_face:

adrianschmidt: Great!

I pushed a new fixup with a small refactor of `tabs.ts`. It still does the same thing, I just made things a little more explicit, and added some comments.

👍 or 👎?

🙂

src/components/tab-bar/tabs.ts
```diff
@@ -0,0 +1,47 @@
...
+    });
+
+    return result.map(selectTab);
+}
+
+function getMapCompatibleReplaceTabFunction(updatedTab: Tab) {
+    return replaceTab;
+
+    // eslint-disable-next-line prefer-arrow/prefer-arrow-functions
+    function replaceTab(existingTab: Tab) {
```
adrianschmidt, line 36: The lint rule wants this to like as follows:

```ts
function getMapCompatibleReplaceTabFunction(updatedTab: Tab) {
    const replaceTab = (existingTab: Tab) => {
        if (updatedTab.id === existingTab.id) {
            // This is the tab we want to update, so return the
            // updated tab instead of the existing one.
            return updatedTab;
        }

        // This is not the tab we are looking for,
        // so just return it untouched.
        return existingTab;
    };

    return replaceTab;
}
```

But I think that makes it less readable, because it lures me into reading the details of the inner function first, and trying to juggle more details in my mind at once.

jgroth: Hmm I'm not sure about this refactor, to me this is less readable :slightly_frowning_face: 
But, reading your refactor makes me think I know what is causing the confusion for you :slightly_smiling_face: I'm guessing that it is the curried mapping function? We could remove that whole function probably and use `splice` in the `setActiveTab` function instead. I normally don't like `splice` since it mutates the array, but if we make a copy of it first it should be fine.

So something like this maybe?

```ts
export function setActiveTab(tabs: Tab[], index: number): Tab[] {
    const oldSelectedIndex = tabs.findIndex((t) => t.active === true);    
    let result = [...tabs];

    if (oldSelectedIndex) {
        result.splice(oldSelectedIndex, 1, { ...tabs[oldSelectedIndex], active: false });
    }

    result.splice(index, 1, { ...tabs[index], active: true });

    return result;
}
```

jgroth: `splice` should have a performance benefit as well, O(1) vs O(n) 
Although it's probably very minor since the number of items will be quite small :sweat_smile:

jgroth: And we don't actually have to use `splice` lol
We could just use
```ts
result[index] = { ...tabs[index], active: true }
```
:upside_down_face:

adrianschmidt: The curried function itself isn't really a problem for me, but the combination of the arrow-syntax and the initial call of the function inside the map() made it a stumbling block. Then the use of `oldSelectedTab` and `oldTab` to mean *completely* different things came in and kicked me when I was down 😆

Not to toot my own horn here, but I think I'm relatively well acquainted with higher order programming in javascript, so if I'm having trouble following the code, I think there's a good chance other people might too.

Regarding the splice-idea, I hadn't thought of using `splice` specifically, but I did think about whether we could maybe just find the relevant elements and flip their `active` value, and then use spread to make a copy of the array to return, if that's needed to trigger any re-rendering?

adrianschmidt (approved):
👌

PR #934, opened 2020-09-01
fix(form): fix resetting dependent fields during schema changes in lists

### LINKED ISSUE #933: Reset dependent fields not working in lists
## Current behavior

Resetting dependent fields when the fields schema changes due to changes in another field does not work inside lists

## Expected behavior

Resetting dependent fields should work regardless of whether or not it is in a list or not

## Environment

- **lime-elements** version: 28.0.1
- Framework used: Angular
- Logs: N/A

## Cause
We added support for resetting dependent fields in the SchemaField component that the react-jsonschema-form form component uses. This resetDependentFields function is dependent on having access to the raw schema. react-jsonschema-form does not pass a raw schema to the SchemaField components rendered as children of the ArrayField component. Instead it passes a modified schema has the dependencies resolved for the current state of the form data. Therefore this modified schema is not able to be used to effectively determine changes in schemas in dependent fields.

## PR's
https://github.com/Lundalogik/lime-elements/pull/934

sampsonbryce:
fix #933

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


jgroth (approved):
Great work! Especially with all the comments explaining the problem :slightly_smiling_face: 

One thing for the next PR though... There were some code here that just had been moved from one file to another. It makes it easier for the reviewer if those kind of changes are in a separate refactor-commit, so you can focus on the real changes instead of just moved code :slightly_smiling_face:

PR #938, opened 2020-09-02
Fix usability of clear all chips button

### LINKED ISSUE #936: chip-set: keyboard-navigation for delete-all button broken
## Current behavior

1. You can tab to the clear-button, but there doesn't seem to be any way to actually activate it. I've tried both \<enter\>  and \<space\>.
image
2. If you don't hover the component when tabbing to the clear-button, it looks a bit broken:
image
3. It's possible to tab to the button when there are no chips selected.
4. When disabled, it's still possible to tab to the button:
image
5. When read-only, it's not only possible to tab to the button, but it also becomes visible when any chip is hovered:
image

(Ignore the issue with the label cutting part of the chips, when disabled or read-only, that's an unrelated problem, and belongs in a separate issue.)

## Expected behavior

1. Pressing <enter> should activate the button.
2. The button should appear correctly when tabbed to, whether or not the component is being hovered.
3. The button should be taken out of the tab-flow when the component has no chips. We should probably use `display:none`, so it's also hidden from assistive tech.
4. Regardless of if the component is empty or has chips, it should not be possible to tab to the button when the component is disabled.
5. Same as for disabled, no tabbing to the button when the component is read-only, and also don't show the button on hover.

## Environment

- **lime-elements** version: v28.1.0 (introduced in #931)
- Framework used: reproducible in docz-app
- Logs:

Kiarokh:
fix #936 

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -601,7 +601,8 @@ export class ChipSet {
 
     private renderClearAllChipsButton() {
         return (
-            <i
+            <a
+                href=""
                 onClick={this.handleDeleteAllIconClick}
                 class="mdc-text-field__icon clear-all-button"
                 tabindex="0"
@@ -611,7 +612,8 @@ export class ChipSet {
             />
         );
     }
-    private handleDeleteAllIconClick() {
+    private handleDeleteAllIconClick(event: Event) {
+        event.preventDefault();
         this.change.emit([]);
     }
 
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -340,7 +340,7 @@ export class ChipSet {
                     />
                 </div>
                 {this.renderLeadingIcon()}
-                {this.renderRemoveAllChipsButton()}
+                {this.renderClearAllChipsButton()}
                 <label
                     class={{
                         'mdc-floating-label': true,
@@ -598,7 +598,8 @@ export class ChipSet {
             />
         );
     }
-    private renderRemoveAllChipsButton() {
+
+    private renderClearAllChipsButton() {
         return (
             <i
                 onClick={this.handleDeleteAllIconClick}
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -601,7 +601,7 @@ export class ChipSet {
         return (
             <i
                 onClick={this.handleDeleteAllIconClick}
-                class="mdc-text-field__icon delete-all-icon-button"
+                class="mdc-text-field__icon clear-all-button"
                 tabindex="0"
                 role="button"
             />
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -149,6 +149,7 @@ export class ChipSet {
     private mdcChipSet: MDCChipSet;
     private mdcTextField: MDCTextField;
     private handleKeyDown = handleKeyboardEvent;
+    private clearAllLabel = 'Clear chips';
 
     constructor() {
         this.renderChip = this.renderChip.bind(this);
@@ -604,6 +605,8 @@ export class ChipSet {
                 class="mdc-text-field__icon clear-all-button"
                 tabindex="0"
                 role="button"
+                title={this.clearAllLabel}
+                aria-label={this.clearAllLabel}
             />
         );
     }
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -67,3 +67,26 @@
         transform: translate3d(0, 0.05rem, 0);
     }
 }
+
+@mixin clear-all-button() {
+    // NOTE: you may need to specify "position: absolute" and align the position where you use this mixin
+    @include is-flat-clickable;
+
+    cursor: pointer;
+
+    height: pxToRem(20);
+    width: pxToRem(20);
+    border-radius: 50%;
+
+    background: {
+        color: #babac0;
+        repeat: no-repeat;
+        position: center;
+        size: pxToRem(12);
+        image: url("data:image/svg+xml; utf8, <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 32 32'><defs/><path fill='rgb(255,255,255)' d='M7.219 5.781L5.78 7.22 14.563 16 5.78 24.781 7.22 26.22 16 17.437l8.781 8.782 1.438-1.438L17.437 16l8.782-8.781L24.78 5.78 16 14.563z'/></svg>");
+    }
+
+    &:hover {
+        background-color: #757580;
+    }
+}
--- a/src/components/chip-set/chip-set.e2e.ts
+++ b/src/components/chip-set/chip-set.e2e.ts
@@ -359,7 +359,7 @@ describe('limel-chip-set', () => {
             beforeEach(async () => {
                 spy = await page.spyOnEvent('change');
                 const deleteAllIconButton: E2EElement = await page.find(
-                    'limel-chip-set >>> .delete-all-icon-button'
+                    'limel-chip-set >>> .clear-all-button'
                 );
                 await chipSet.hover();
                 await deleteAllIconButton.click();
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -78,6 +78,9 @@ limel-icon.mdc-chip__icon.mdc-chip__icon--leading {
     &.mdc-chip-set--input {
         padding-top: $padding-top-of-chipset-input;
         padding-bottom: 0;
+        padding-right: pxToRem(
+            32
+        ); // This leaves space for "clear all" button and avoids overlapping with chips
         width: 100%;
 
         .mdc-chip--selected {
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -176,32 +176,25 @@ limel-icon.mdc-chip__icon.mdc-chip__icon--leading {
     @extend .mdc-text-field--invalid;
 }
 
-.has-chips:hover {
-    .delete-all-icon-button {
-        -webkit-appearance: none;
-        cursor: pointer;
+.clear-all-button {
+    transition: opacity 0.3s ease;
+    @include clear-all-button;
 
-        @include is-flat-clickable;
-        height: pxToRem(20);
-        width: pxToRem(20);
-        border-radius: 50%;
+    position: absolute;
+    right: pxToRem(8);
+    top: pxToRem(26);
 
-        position: absolute;
-        right: pxToRem(8);
-        top: 0;
-        bottom: 0;
-        margin: auto;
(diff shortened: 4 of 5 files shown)


src/components/chip-set/chip-set.tsx
```diff
@@ -572,9 +578,11 @@ export class ChipSet {
         return (
             <i
                 onClick={this.handleDeleteAllIconClick}
-                class="mdc-text-field__icon delete-all-icon-button"
+                class="mdc-text-field__icon clear-all-icon-button"
                 tabindex="0"
                 role="button"
+                title={this.clearAllLabel}
+                area-label={this.clearAllLabel}
```
adrianschmidt, line 585:
```suggestion
                aria-label={this.clearAllLabel}
```

src/style/mixins.scss
```diff
@@ -67,3 +67,26 @@
...
+    background: {
+        color: #babac0;
+        repeat: no-repeat;
+        position: center;
+        size: pxToRem(12);
+        image: url("data:image/svg+xml; utf8, <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 32 32'><defs/><path fill='rgb(255,255,255)' d='M7.219 5.781L5.78 7.22 14.563 16 5.78 24.781 7.22 26.22 16 17.437l8.781 8.782 1.438-1.438L17.437 16l8.782-8.781L24.78 5.78 16 14.563z'/></svg>");
+    }
+
+    &:hover {
+        background-color: #757580;
```
adrianschmidt, line 90: While we're at it, shouldn't we use the new color-palette for these background colours? 🙂

src/components/chip-set/chip-set.tsx
```diff
@@ -82,6 +82,12 @@ export class ChipSet {
     @Prop({ reflectToAttr: true })
     public searchLabel: string;
 
+    /**
+     * Clear all label to display as `title` and `area-label` for icon button that clears all chips
```
adrianschmidt, line 86:
```suggestion
     * Label to display as `title` and `aria-label` for button that clears all chips
```

src/components/chip-set/chip-set.tsx
```diff
@@ -572,9 +578,11 @@ export class ChipSet {
         return (
             <i
```
adrianschmidt, line 579: I think this should be a link. Then there should be no need for the `tabindex="0"`, and it ought to be possible to trigger it by tabbing to it and pressing \<enter\>.

Alternatively, we might want to use a limel-icon-button, but I'm guessing that might be a little too bloated for use here?

src/components/chip-set/chip-set.tsx
Kiarokh: typical me :D

src/style/mixins.scss
Kiarokh: I think we will go through these color things later, one by one in each component.

src/components/chip-set/chip-set.tsx
Kiarokh: have you tried to see whether having it as a link would work? Otherwise I'lll test tomorrow ;)

Kiarokh: @adrianschmidt this didn't work unfortunately. :( maybe for a link to work, we need a URL? donno

Kiarokh:
@adrianschmidt the tabbing behavior seems to be inconsistent. Sometimes it works (specially when the field is "focused" and you're adding or removing chips) But some times it doesn't (like when you tab into the field from outside). We couldn't really figure out how and why.

Kiarokh:
> It's because unless you hover the component, the button is set to `display:none` as soon as the input loses focus.

But we don't want the (x) to show up when the field is not being used (is not focused). However, in some situations as soon as you tab, the field will lose focus. Right now I'm not sure how to fix this problem. Perhaps the easiest way is to constantly show the (x) button as long as `.has-chips` class exists, even if the field is not in focus.

adrianschmidt:
> > It's because unless you hover the component, the button is set to `display:none` as soon as the input loses focus.
> 
> But we don't want the (x) to show up when the field is not being used (is not focused). However, in some situations as soon as you tab, the field will lose focus. Right now I'm not sure how to fix this problem. Perhaps the easiest way is to constantly show the (x) button as long as `.has-chips` class exists, even if the field is not in focus.

Invisible and `display:none` are not the same thing. But I can fix it later. Busy with some other things at the moment, but tell me once you've done what you want and can, and I'll look at the remaining bits.

Kiarokh:
> By "invisible", I didn't mean `visibility:hidden` or any other specific styling, just that there are ways to make something be not possible to see, while still possible to focus. 

No problem Adrian. Using `visibility:hidden` was actually a smart idea, even if you didn't mean it 😄. `https://developer.mozilla.org/en-US/docs/Web/CSS/visibility` they say:
image

Anyhow, I later realized that you can still tab into the (x) even if there are no chips (due to lack of CSS rules), so in the [last fixup commit](https://github.com/Lundalogik/lime-elements/pull/938/commits/7ee6cb3005a85457ce1b3b0bec540ee8373af194), I fixed this. I also found a bug which is hard to explain. The styles caused the (x) to jump up and down so you couldn't really click it. This happened when the first line in the chipset was filled with chips and the `<input>` (where you type in) was placed on the next line. 

Anyhow, it's now fixed by having an absolut position for the (x) which only considers `right` and `top`. (doesn't middle align itself based on the height of the chipset). This fix gives a more pleasing UI as well.

src/components/chip-set/chip-set.tsx
```diff
@@ -82,6 +82,12 @@ export class ChipSet {
     @Prop({ reflectToAttr: true })
     public searchLabel: string;
 
+    /**
+     * Label to display as `title` and `aria-label` for button that clears all chips
+     */
+    @Prop({ reflectToAttr: true })
+    public clearAllLabel: string;
```
jgroth, line 89: Is this to make it translatable or do we want consumers to be able to change the label to something completely different? 

If it's only for translation purposes, I would suggest to remove this for now and just have it in English. We need to discuss how to handle translations in Lime Elements at a broader scope before it becomes unmaintainable. I would prefer if we could solve it without having props for all translation strings on a component level. Thoughts? :slightly_smiling_face:

adrianschmidt: > Is this to make it translatable or do we want consumers to be able to change the label to something completely different?

Ping @Kiarokh ☝️ 

> If it's only for translation purposes, I would suggest to remove this for now and just have it in English. We need to discuss how to handle translations in Lime Elements at a broader scope before it becomes unmaintainable. I would prefer if we could solve it without having props for all translation strings on a component level. Thoughts? 🙂

It seems we already have some temporary hardcoded translations in Lime Elements, used for the date-picker. If it's for translation, I would suggest adding the translations there for now.

I absolutely agree that we should discuss how to handle translations generally 👍

Kiarokh: > Ping @Kiarokh ☝️ 

thanks

> Is this to make it translatable

yes :) 

> handle translations in Lime Elements at a broader scope

would be awesome, decent, amazing, professional ;) 

> solve it without having props for all translation strings on a component level

you guys know the best how to do it optimally

adrianschmidt: Great! I'll add some hardcoded translations 👍

adrianschmidt: Ah, crap… no, that's just going to require us to send in the language code instead. I'm going with English only for now.

Unfortunately, since that property has already been merged and released, this is now becoming a breaking change 😭

adrianschmidt: I decided to deprecate the property, and give it a default value in English. That way we can avoid making a breaking change for now.

src/components/chip-set/chip-set.tsx
```diff
@@ -83,10 +83,10 @@ export class ChipSet {
     public searchLabel: string;
 
     /**
-     * Label to display as `title` and `aria-label` for button that clears all chips
+     * ***DEPRECATED*** Label to display as `title` and `aria-label` for button that clears all chips
```
jgroth, line 86:
```suggestion
     * Label to display as `title` and `aria-label` for button that clears all chips
     * @deprecated
```

jgroth:
I'm not sure why GitHub is not letting me respond in the thread so I'm forced to write my reply here instead...

> Unfortunately, since that property has already been merged and released, this is now becoming a breaking change :sob:

Has it been merged? :thinking:  It's part of this PR is it not? I'm confused now :smile:

maltja:
It doesn't seem to work for me to focus on the clear all chips button with tab in Firefox and Safari. Chrome works perfect.

adrianschmidt:
> I'm not sure why GitHub is not letting me respond in the thread so I'm forced to write my reply here instead...
> 
> > Unfortunately, since that property has already been merged and released, this is now becoming a breaking change 😭
> 
> Has it been merged? 🤔 It's part of this PR is it not? I'm confused now 😄

Oh, yeah, it is! I thought it had just been renamed in this one, but introduced in the PR where the button was added. I'll just remove the prop then! 🎉

adrianschmidt:
> It doesn't seem to work for me to focus on the clear all chips button with tab in Firefox and Safari. Chrome works perfect.

Good catch! I'll take a look at it 👍

src/components/chip-set/chip-set.tsx
adrianschmidt: Resolving this as I removed the property.

src/components/chip-set/chip-set.tsx
adrianschmidt: Sorry for being confused. The property was introduced in this PR, so I just removed it and used a hardcoded value in English instead.

adrianschmidt:
> > It doesn't seem to work for me to focus on the clear all chips button with tab in Firefox and Safari. Chrome works perfect.
> 
> Good catch! I'll take a look at it 👍

I could get it to work in Firefox if I changed it from an `<a>` to a `<button>`, but then I lost the hover effect. Safari refused to focus it whatever I did, but I think Safari generally doesn't focus on any links or buttons?

I've now spent most of the day on this without really getting anywhere, and I think we should just merge this as it's clearly better than what's currently on master. If we wish to spend more time polishing it later, we can do so in a separate PR.

Kiarokh:
I also think we should merge it. We will at some point go through keyboard navigation and focus styles of all components. These things we can certainly improve later.

However, it might be a good idea to already create an issue regarding Safari in this board https://github.com/orgs/Lundalogik/projects/61 so we don't forget them.

adrianschmidt:
It actually works in Safari, but you have to enable this accessibility option in Safari's preferences:

image

jgroth (approved):

PR #947, opened 2020-09-04
perf(table): improve tables' loading speed in chrome

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/1361

## Affects Chrome 85+
At this time, `content-visibility: auto;` works in Chrome 85 and above. (as far as I know [Kia]). More here: https://web.dev/content-visibility/

## performance test
Tested with an object with 60 deals. I think in an object with a lot of table rows (`.tabulator-row`), this will have an even more notable impact. Performance analysis shows from the second the page is refreshed > until the user clicks on the *deals* tab and its content is loaded. 

### Reload page > Go to deals tab (With this change)
Screen Shot 2020-09-04 at 16 25 22

### Reload page > Go to deals tab  (Without this)
 change
Screen Shot 2020-09-04 at 16 26 24

______________

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.scss
+++ b/src/components/table/table.scss
@@ -55,6 +55,8 @@ $tabulator-arrow-color-active: var(--lime-turquoise);
 }
 
 .tabulator-row {
+    content-visibility: auto; //Currently (2020-Sep-04), this works only in Chrome 85+ https://web.dev/content-visibility/
+
     &.tabulator-selectable {
         transition: background-color 0.2s ease;
         &:not(.active) {


Kiarokh:
Because we tested the connected PRs (2020-09-07) in Chrome and it seems they create a blinking effect while scrolling too fast in tabulator rows. This happens even when scrolling back to rows which are already loaded once. This may be due to early and poor implementations of the feature by Chrome. 

It probably interferes with Tabulator's own lazy inbuilt loading.

PR #956, opened 2020-09-15
perf(portal): create popper instance when portal is displayed

jgroth:
The popper instance has an event listener on scroll, which causes severe performance issues when
there are a lot of portals on a page e.g. menues in list items etc. The popper instance does not
have to be created until something is actually displayed and needs to be positioned.

## Review:

### Browsers tested:

Windows:
- [x] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/portal/portal.tsx
+++ b/src/components/portal/portal.tsx
@@ -73,16 +73,16 @@ export class Portal {
     @Watch('visible')
     protected onVisible() {
         if (!this.visible) {
+            this.hideContainer();
+            this.styleContainer();
+            this.destroyPopper();
+
             return;
         }
 
-        if (!this.popperInstance) {
-            return;
-        }
-
-        setTimeout(() => {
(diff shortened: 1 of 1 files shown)


src/components/portal/portal.tsx
```diff
@@ -73,16 +73,16 @@ export class Portal {
     @Watch('visible')
```
adrianschmidt, line 73: Yay! This seems to work well, and it also seems to have the added benefit that menus with `open-direction="left"` no longer opens to the right and then jumps to the left the first time they are opened! 👍

adrianschmidt (approved):

PR #957, opened 2020-09-15
docs(design-guidelines): add guidelines for size rhythms

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


adrianschmidt:
I've self-requested a review, since I want to have a look at this, but anyone else is of course also welcome and encouraged to review or just generally have a look! 👍😊

Kiarokh:
Thanks @adrianschmidt 
I'd love to get feedback specially from you regarding these design guidelines. Let me know how we can improve them.

adrianschmidt:
Since I'm only just now starting to look at this, I've rebased this on the latest master and squashed the fixups. (Still use fixups for any new changes though, of course 🙂)

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
+# Size rhythms
+
+When designing components and layouts, you will have to specify size of many properties such as `padding`, `margin`, `grid-gap`, `width`, `height`, and so forth.
```
adrianschmidt, line 3:
```suggestion
When designing components and layouts, you will have to specify the size of many properties such as `padding`, `margin`, `grid-gap`, `width`, `height`, and so forth.
```

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+
+To generate a consistent layout and visual expression, we only use sizes that are multiples of four (4 * x). For example 4, 8, 12, 16, 20, and so on.
+
+It is very important to follow this convention, both when creating new components for Lime Elements, and when creating bigger components and layouts which use these components. When the entire interface is calculated based on the single core value of 4, elements of the UI will align auto-magically.
+
+For instance, if you want to create a custom row of buttons (not directly using Lime Elements) with this sizing convention, its `height` and `line-height` can be 40 units (to vertically align the text inside the button), the `padding-left` and `padding-right` can be 12 units, and its `margin-left` and `margin-right` which set the distance to the next or previous buttons can be for example 4 or 2, or in this example 6.
+
+Even though 6 is not a multiple of 4, it sums up to a 12 units distance between two buttons in a row; and since 12 is a multiple of 4, the result will be moa👌.
```
adrianschmidt, line 11: What does "moa" mean?

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+
+<limel-example-size></limel-example-size>
+
+
+## REM or PX
+We only use `rem` units in our designs system, based on browsers' default base unit of measurement (where `16px` = `1rem`). In such a system, multiples of 4px (our core size unit) can easily be calculated, as they increase sequentially by `0.25`.
```
adrianschmidt, line 17:
```suggestion
We only use `rem` units in our design system. `1rem` equals the font size set in the browser, and defaults to `16px` in most browsers. In such a system, multiples of 4px (our core size unit) can easily be calculated, as they increase sequentially by `0.25`.
```

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+
+* 24 px = 1.5 rem
+* 20 px = 1.25 rem
+* 16 px = 1 rem
+* 8 px = 0.5 rem
+* 4 px = 0.25 rem
```
adrianschmidt, line 23: It was a little confusing to me when the preceding paragraph spoke of increasing in multiples of .25, to then come to this list where we start at 1.5 and go down.

```suggestion
* 4 px = 0.25 rem
* 8 px = 0.5 rem
* 16 px = 1 rem
* 20 px = 1.25 rem
* 24 px = 1.5 rem
```

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+
+However, while designing a UI, it may be easier to think in pixels, as rem is more abstract. By using a custom SASS function like below, you can effortlessly leave the job of unit conversion to the computer.
+
+````css
+/// first define this converter function in your SCSS ···
```
adrianschmidt, line 28:
```suggestion
// first define this converter function in your SCSS …
```

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+@function pxToRem($px) {
+    @return #{$px/16}rem;
+}
+
+// ··· and then use it to ease calculations
```
adrianschmidt, line 33:
```suggestion
// … and then use it to ease calculations
```

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+.example {
+    width: pxToRem(4); // will be compiled to 0.25rem
+}
+````
+
+If you are developing new components for Lime Elements, you should know that this function is already available and can be easily imported into your `.scss` files via: `@import '../../style/mixins';`
+
+So if you think something should be rendered 40px wide on the screen, you you don't need to calculate the respective value in REM. Just type `width: pxToRem(40);` in your SCSS code.
```
adrianschmidt, line 41: I'm not sure we should include documentation only directed at developers of lime-elements itself here. We could perhaps include it on the *Contributing* page…

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+
+**Note** write `1px` not `0,0625rem`!
+
+Anything that should get rendered in `1px` on a screen (such as a border) should be written in pixels. It's very important not to write it in its rem equivalent (`0,0625rem`). This is because depending on the physical pixel density of screens, a `0,0625rem` thick border for example, may get rendered like a semi-transparent line; or even worse, never get rendered in some situations.
+
+## Exceptions in sizing rhythms
+There are exceptions where we do not have to follow the abovementioned sizing convention (multiples of 4). These exceptions are:
```
adrianschmidt, line 48:
```suggestion
There are exceptions where we do not have to follow the above-mentioned sizing convention (multiples of 4). These exceptions are:
```

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+}
+````
+
+If you are developing new components for Lime Elements, you should know that this function is already available and can be easily imported into your `.scss` files via: `@import '../../style/mixins';`
+
+So if you think something should be rendered 40px wide on the screen, you you don't need to calculate the respective value in REM. Just type `width: pxToRem(40);` in your SCSS code.
+
+**Note** write `1px` not `0,0625rem`!
+
+Anything that should get rendered in `1px` on a screen (such as a border) should be written in pixels. It's very important not to write it in its rem equivalent (`0,0625rem`). This is because depending on the physical pixel density of screens, a `0,0625rem` thick border for example, may get rendered like a semi-transparent line; or even worse, never get rendered in some situations.
```
adrianschmidt, line 45:
```suggestion
Anything that should get rendered in `1px` on a screen (such as a border) should be written in pixels. It's very important not to write it in its rem equivalent (`0,0625rem`). This is because depending on the physical pixel density of the screen, a `0,0625rem` thick line might be rendered as a semi-transparent line, or, even worse, not be rendered at all.
```

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+
+## Exceptions in sizing rhythms
+There are exceptions where we do not have to follow the abovementioned sizing convention (multiples of 4). These exceptions are:
+* Numbers `1` and `2` are allowed to use when defining some visual details such as borders.
+* `font-size`: how fonts are rendered depend on the typography and design of the font itself. Fonts do not follow the same logic of simple geometric shapes of the UI. Therefore, using incremental rhythm based on 4 for font sizes (e.g. for H1, H2, H3, etc...) may generate huge font sizes which won't be usable. Also other numbers such as 14 or 13 may appear more appealing for a balanced and easily readable text, compared to 16, or 12.
```
adrianschmidt, line 50: Should the line-height be adjusted to make the block of text still line up with the 4* rhythm?

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+* `border-radius`: how a value of 4 is rendered as a border radius is slightly too-little for the type of visual expression we want to achieve for example on a small clickable element like a button. This is while a value of 8 may be too much. Thus, the designer may pick an arbitrary value that helps the visual expression.
```
adrianschmidt, line 51:
```suggestion
* `border-radius`: a border radius of 4 is slightly too little for the type of visual expression we want to achieve for example on a small clickable element like a button. This is while a value of 8 may be too much. Thus, the designer may pick an arbitrary value that helps the visual expression.
```

This whole bullet point is a bit rambling, and hard to follow. Can it be simplified?

How about something like this?

```suggestion
* `border-radius`: for small elements, jumps of 4 may be too large for the border-radius. Pick the value that best helps the visual expression, even if it's not a multiple of 4.
```

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+* `box-shadow`, `drop-shadow`, `blur`: How far a shadow reaches until it's faded out, or how blurry something looks should also be flexible depending on the visual expression that you want to achieve.
+
+## Be believer, but not a fanatic!
```
adrianschmidt, line 54:
```suggestion
## Be a believer, but not a fanatic!
```

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+This is about having a consistent rhythm in the UI, not about religiously following hard guidelines. In some cases, you will find yourself in situations in which some small detail in the UI is not following the 4-units rhythm, while everything else does.
+
+Let's look at the example below:
+
+<limel-example-size-edge-case></limel-example-size-edge-case>
+
+Here you see that the distance between the picture and top, bottom, and left of the rows is 6 units. However, the `line-height` or `height` of the picture itself (and other details) are multiplies of 4. Now this is a fantasy example for illustrative purposes, but cases like this may happen in real life when designing components and layouts. For example you may choose to use 6 units for distance on a certain type of screen to achieve a higher visual density.
```
adrianschmidt, line 61:
```suggestion
Here you see that the distance between the picture and top, bottom, and left of the rows is 6 units. However, the `line-height` or `height` of the picture itself (and other details) are multiples of 4. Now this is a made up example for illustrative purposes, but cases like this may happen in real life when designing components and layouts. For example you may choose to use 6 units for distance on a certain type of screen to achieve a higher visual density.
```

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+
+So, if you can justify that 6 works fine for your design, and rest of your structure is following the 4x rhythm; then go for it! In this example, the accumulated height of the row (which is more decisive for the general rhythm) will be 24 (4 * 6). So you don't "have to" change 6 to 4 or 8 so that all numbers become multiplies of 4.
```
adrianschmidt, line 63:
```suggestion
So, if you can justify that 6 works fine for your design, and the rest of your structure is following the 4x rhythm; then go for it! In this example, the accumulated height of the row (which is more decisive for the general rhythm) will be 24 (4 * 6). So you don't "have to" change 6 to 4 or 8 so that all numbers become multiplies of 4.
```

src/size-rhythms.md
```diff
@@ -0,0 +1,65 @@
...
+
+The only thing you should be mindful about is whether you are getting too many of these exceptions within the same component. In that case, something is wrong in your design.
```
adrianschmidt, line 65:
```suggestion
The only thing you should be mindful of is whether you are getting too many of these exceptions within the same component. In that case, something is wrong in your design.
```

adrianschmidt (changes requested):

src/size-rhythms.md
Kiarokh: it's a "sound" you make when something is perfect and you appreciate it 😄  while you do this gesture -> 👌 with you hand in front of your mouth. Like:
😚👌 moa....

But let's change it to. 😂

src/size-rhythms.md
Kiarokh: Good idea! Will you update the contribution page?

src/size-rhythms.md
Kiarokh: Well… Sometimes you wanna have a button that's 40px high, and an easy way to do so is using `line-height: 40px`. So, in this case you wanna play with line-height. But for blocks of text it won't be needed. It's generally impossible to get dynamic sections of the page to follow these rules. Specially given the responsive layouts and various screen sizes.

For example a block of text with `width: 100%`:
| ←4px padding→| ····· content here ····· | ←4px padding→| 

The content section can be anything in width (which doesn't have to be a multiple of 4), depending on the screen size. 

Do you think we should add this to "not be a fanatic" examples?

src/size-rhythms.md
Kiarokh: OK, what about this:

* `border-radius`: when it comes to border radius, multiples of 4 may render too large or too little. Pick the value that best helps the visual expression you want to achieve, even if it's not a multiple of 4.

src/size-rhythms.md
adrianschmidt: Nah… We need to balance being explicit enough for people to understand, and terse enough for people to actually read it…

src/size-rhythms.md
adrianschmidt: 👍 with this tiny change:

```suggestion
* `border-radius`: when it comes to border radius, multiples of 4 may render too large or too small. Pick the value that best helps the visual expression you want to achieve, even if it's not a multiple of 4.
```

("small" instead of "little")

src/size-rhythms.md
adrianschmidt: I'll take a look at it.

adrianschmidt: I'll add this:

> ### Styling
>
> Please read the Design Guidelines section of the documentation, look at existing code, and feel free to ask any questions on the issue or PR you are working on.

Any contributor who is thorough enough to read that file, is likely to look at existing code and find the pxToRem function. And if they don't, we'll just point it out in the PR review.

adrianschmidt: (In other words, I still think we should remove the above two paragraphs.

```suggestion
```

adrianschmidt: I pushed a fixup removing this 👍

src/size-rhythms.md
adrianschmidt: Yeah, I think it's better to change it. We don't need to be stick-up-the-butt formal in our communication, but for one-way communication like this (we're it's not necessarily easy or even possible for the reader to ask for clarification), I think we should prioritise clarity over being casual and personal 🙂

adrianschmidt (approved):

PR #962, opened 2020-09-24
feat(table): add default sorting

### LINKED ISSUE #961: Add default sorting feature in tableview (lime-elements)

Befkadu1:
fix #961

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -55,6 +55,13 @@ export class Table {
     public totalRows: number;
 
     /**
+     * The initial sorted columns
+     */
+
+    @Prop()
+    public sorting: ColumnSorter[] = [];
+
+    /**
      * Active row in the table
      */
     @Prop({ mutable: true })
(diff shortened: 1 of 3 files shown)


src/components/table/table.tsx
```diff
@@ -53,6 +53,13 @@ export class Table {
     @Prop()
     public totalRows: number;
 
+    /**
+     * The initial sorted columns
+     */
+
+    @Prop()
+    public defaultSortedColumns: Tabulator.Sorter[];
```
jgroth, line 61: I don't think we should name it like this, something like `sorting` would be better. We might want to add support for changing the sorting by changing this property in the future and then the name `sorting` makes more sense.

Also, we cannot expose `Tabulator.Sorter` to consumers since that is internal to the implementation of the `limel-table` component. There is an interface named `ColumnSorter` in `src/components/table/table.types.ts`, use that one instead :slightly_smiling_face:

jgroth (changes requested):

jgroth (approved):

PR #967, opened 2020-10-01
Kompendium

jgroth:
fix Lundalogik/crm-feature#1437

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [x] Edge
- [ ] Firefox

Linux:
- [x] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


package.json
```diff
@@ -88,6 +87,8 @@
     "prettier": "2.0.5",
     "prismjs": "1.21.0",
     "puppeteer": "1.19.0",
+    "react": "^16.13.1",
+    "react-dom": "^16.13.1",
```
adrianschmidt, line 91: Why is react suddenly needed?

publish-docs.js
```diff
@@ -125,31 +125,21 @@ function checkOutBranch() {
...
-        replace.sync(options);
-
-        const options2 = {
-            files: ['src/examples/example.tsx', 'src/examples/props.tsx'],
-            from: /const BASE_URL = '\/';/g,
-            to: `const BASE_URL = '/versions/${version}/';`,
-        };
-        replace.sync(options2);
-
         const options3 = {
```
adrianschmidt, line 128: I guess we can update the `options3` and `options4`. Maybe `indexReplaceOptions` and `stencilConfigReplaceOptions`?

adrianschmidt: (And yes, I'm the one guilty of naming them `optionX` in the first place 😆)

package.json
jgroth: It was included implicitly by Docz before, and the `react-jsonschema-form` library in the form component needs it. I think it was only set as a peer dependency from `react-jsonschema-form`

publish-docs.js
jgroth: We could also just have `options` and reuse that variable for the different cases

publish-docs.js
```diff
@@ -20,7 +20,7 @@ const dryRun = !!argv.dryRun && argv.dryRun !== 'false';
```
adrianschmidt, line 20: We are copying the output to the wrong place when publishing, but I'm on it. (Run `npm run docs:publish -- --dryRun --noTeardown` to check the output.) I think I have a fix though, but will run it before pushing.

publish-docs.js
```diff
@@ -161,13 +151,6 @@ function build() {
         };
         replace.sync(options4);
 
-        const options5 = {
-            files: ['src/index.mdx'],
-            from: /<version\\>/g,
-            to: `${version}`,
-        };
-        replace.sync(options5);
```
adrianschmidt, line 169: Needed to update the version here:
image

(But had to be changed from `index.mdx` to `index.md`. Will push fix soon!)

adrianschmidt: Also, while we're at it, the info shown there (under *Home/Lime Elements*) I would quite like to see on the landing page instead.

jgroth: Yeah, great input!
I think I just used the readme in the beginning when there was no support for separate md-files, but you should definitely be able to add something else as the landing page

.dockerignore
```diff
@@ -1,7 +1,7 @@
 Dockerfile
```
adrianschmidt, line 1: The commit `feat: use Kompendium instead of Docz` should perhaps be called `docs: use Kompendium instead of Docz`? We don't really add anything to the published package…

publish-docs.js
adrianschmidt: ⚡

publish-docs.js
adrianschmidt: :zap:

.dockerignore
adrianschmidt: ⚡

adrianschmidt (approved):

PR #972, opened 2020-10-05
docs(select): improve docs for the `Option` type, clarifying that `value` must be a `string`

### LINKED ISSUE #919: limel-select with multiple selection broken on phone
## Current behavior

it opens the native selector/ dropdown where you can't select multiple options and returns undefined when selecting one.
reported on the forum: https://forum.lime-crm.com/t/error-in-web-component-on-mobile/2814

## Environment

- **lime-elements** version: 26.27.4
- Framework used: stencil

adrianschmidt:
fix #919

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


Befkadu1:
What do you think if we use this code `value={String(option.value)}` in `select.template.tsx`? If the user uses numeric values on some iOS devices, it might avoid breaks.

src/components/select/option.types.ts
```diff
@@ -1,5 +1,28 @@
...
+
+    /**
+     * The unique value of the option. Should always be the same for any given
+     * option, regardless of localization. The type `T` defaults to `string`,
+     * but can be set to any type that extends `string` (using `Option<type>`),
+     * for example an enum of specific strings.
+     *
+     * Note the value of this property *must* be a string!
+     * If the value you wish to use is, for example, numeric, convert it to a
+     * string before sending it to limel-select. Using numeric values does work
```
Befkadu1, line 18: Using numeric values work

adrianschmidt:
> What do you think if we use this code `value={String(option.value)}` in `select.template.tsx`? If the user uses numeric values on some iOS devices, it might avoid breaks.

Hmm… yeah, since we don't actually seem to be using any data from the change-event, that might well work. I was thinking we'd have problems with comparing values and stuff, but it looks like we never actually use the value that's been set. I'll try it 👍

src/components/select/option.types.ts
adrianschmidt: I disagree. I'm fairly certain that "using numeric values does work in many desktop browsers" is correct. The singular "does" is referring to the verb "using" (which I assume is uncountable) not the plural noun "values".

adrianschmidt:
> > What do you think if we use this code `value={String(option.value)}` in `select.template.tsx`? If the user uses numeric values on some iOS devices, it might avoid breaks.
> 
> Hmm… yeah, since we don't actually seem to be using any data from the change-event, that might well work. I was thinking we'd have problems with comparing values and stuff, but it looks like we never actually use the value that's been set. I'll try it 👍

No, unfortunately that doesn't seem to help 😞

Befkadu1 (approved):

PR #977, opened 2020-10-06
Update Stencil to v2

### LINKED ISSUE #975: Update Stencil to v2
## New feature motivation

We should stay up to date, and there are nice new improvements as well 😊

## New feature description

There shouldn't be any functional changes.

## New feature implementation

There are a small number of breaking changes in v2, described in the [changelog](https://github.com/ionic-team/stencil/blob/master/CHANGELOG.md#-200-2020-08-31). It seems like they should be pretty straightforward to address.

adrianschmidt:
fix #975

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


package.json
```diff
@@ -93,7 +93,8 @@
     "semantic-release": "^17.1.1",
     "shelljs": "0.8.4",
     "showdown": "1.9.1",
-    "shx": "0.3.2"
+    "shx": "0.3.2",
+    "typescript": "^4.0.3"
```
adrianschmidt, line 97: Stencil <2 had a dependency on typescript. That dependency has been removed, which meant that other dependencies we have failed because typescript wasn't available. So I had to add this dev-dependency.

src/components/tab-panel/tab-panel.tsx
```diff
@@ -51,25 +51,36 @@ export class TabPanel {
...
     }
 
-    // eslint-disable-next-line @stencil/own-methods-must-be-private
-    protected componentDidUnload() {
+    public disconnectedCallback() {
         const slot = this.getSlot();
         slot.removeEventListener('slotchange', this.setSlotElements);
     }
 
+    constructor() {
```
jgroth, line 78: I'm not sure we should move the constructor this far down in the component :thinking: 
We usually have it at the top everywhere else

package.json
jgroth: Some thoughts... (not really related to this PR :smile: )
Should we not be able to put almost _all_ of our dependencies as dev-dependencies? For a consumer that installes lime-elements, all the dependencies will be installed as well. That should not be needed since we compile all our code into a distributable package :thinking: Everything listed here in `package.json` is only for building lime-elements, not consuming it :slightly_smiling_face:

src/components/tab-panel/tab-panel.tsx
adrianschmidt: Hmm… I don't know how it ended up all the way down here. I agree, and I'll move it back up 👍

package.json
adrianschmidt: Almost all our dependencies are dev-dependencies… 😄

src/components/tab-panel/tab-panel.tsx
adrianschmidt: Because there were two different commits that involved the same code, I couldn't move it with a fixup without creating a really messy conflict when rebasing it in the future. So did the change in an interactive rebase. The force-push I just did is only changing this one file.

adrianschmidt: ⚡

package.json
jgroth: Yeah but I think most if not all of the other ones probably could be as well :slightly_smiling_face:

jgroth (approved):

package.json
adrianschmidt:
```json
  "dependencies": {
    "@limetech/material-components-web": "^4.0.1-p4.0.0.1",
    "@popperjs/core": "^2.5.3",
    "@types/tabulator-tables": "^4.7.1",
    "ajv": "^6.12.2",
    "awesome-debounce-promise": "^2.1.0",
    "chart.js": "^2.9.3",
    "flatpickr": "^4.6.3",
    "html-escaper": "^3.0.0",
    "jsx-dom": "^6.4.17",
    "kompendium": "^0.7.0",
    "moment": "^2.26.0",
    "number-abbreviate": "^2.0.0",
    "react-jsonschema-form": "^1.8.1",
    "react-shadow-dom-retarget-events": "^1.0.11",
    "tabulator-tables": "^4.8.2"
  }
```

Really? `kompendium` should be a dev-dependency, but apart from that, it would seem to me like all of these are needed by the final build? Unless of course the tree-shaking pulls in and embeds any code that's used… 🤔

Hmm… yeah, I guess it would be worth a try to just move everything to dev-deps and see if it works 👍

Different branch though, obviously.

jgroth: Yeah I think all of those will be compiled into our package. No need for consumers of lime-elements to install those :smile: Just a theory though :smile:

PR #992, opened 2020-10-21
Form layout

jgroth:
fix Lundalogik/crm-feature#1479

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [x] Edge
- [x] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/form/examples/form-layout.tsx
+++ b/src/components/form/examples/form-layout.tsx
@@ -1,0 +1,59 @@
+import { ValidationStatus } from '@limetech/lime-elements';
+import { Component, h, State } from '@stencil/core';
+import { schema } from './schema-layout';
+
+/**
+ * Layout
+ *
+ * @link schema-layout.ts
+ */
+@Component({
+    tag: 'limel-example-form-layout',
+    shadow: true,
+})
+export class FormLayoutExample {
+    @State()
+    private formData: object = {};
+
+    @State()
+    private valid = true;
+
+    constructor() {
+        this.handleFormChange = this.handleFormChange.bind(this);
+        this.handleFormValidate = this.handleFormValidate.bind(this);
+        this.handleSubmit = this.handleSubmit.bind(this);
+    }
+
+    public render() {
+        return [
+            <limel-form
+                onChange={this.handleFormChange}
+                onValidate={this.handleFormValidate}
+                value={this.formData}
+                schema={schema}
+            />,
+            <br />,
+            <limel-button
+                label="Submit"
+                primary={true}
+                disabled={!this.valid}
+                onClick={this.handleSubmit}
+            />,
+            <limel-example-value value={this.formData} />,
+        ];
+    }
+
+    private handleFormChange(event) {
+        this.formData = event.detail;
+    }
+
+    private handleFormValidate(event: CustomEvent<ValidationStatus>) {
+        this.valid = event.detail.valid;
+        console.log(event.detail);
+    }
+
+    private handleSubmit() {
+        const json = JSON.stringify(this.formData, null, '    ');
+        alert(`Sending information to villains...\n\n${json}`);
+    }
+}
--- a/src/components/form/form.types.ts
+++ b/src/components/form/form.types.ts
@@ -130,6 +130,12 @@ export interface LimeSchemaOptions {
      * should implement the `FormComponent` interface
      */
     component?: FormComponentOptions;
+
+    /**
+     * When specified on an object it will render the sub components with the
+     * specified layout
+     */
+    layout?: FormLayoutOptions;
 }
 
 /**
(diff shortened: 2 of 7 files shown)


src/components/form/form.types.ts
```diff
@@ -104,3 +104,73 @@ export interface FormInfo {
...
+    /**
+     * The type of layout to use
+     */
+    type: FormLayoutType;
+
+    /**
+     * When specified on a component within a layout, the component will take
+     * up the full width of the form
+     */
+    fullWidth?: boolean;
```
jgroth, line 167: Not 100% sure about the API here :thinking: 
Tried having `span` property here instead but it seemed a bit complex

src/components/form/form.types.ts
```diff
@@ -104,3 +104,73 @@ export interface FormInfo {
...
+}
+
+// eslint-disable-next-line no-shadow
```
jgroth, line 170: No idea why this was needed since there is no shadowing of `FormLayoutType`, it might be a bug in ESLint. No matter what I named it, it always complained

adrianschmidt: Yeah, I think it's a bug in ESLint. It seems to complain for all top level exports that aren't `interface`, or something like that at least. I've seen the same thing before.

src/components/form/templates/field.ts
```diff
@@ -1,12 +1,20 @@
 import React from 'react';
+import { FormLayoutOptions } from '../form.types';
 
 export const FieldTemplate = (props) => {
     const { classNames, children } = props;
+    const layout: FormLayoutOptions = props.schema.lime?.layout;
+    const style: any = {};
+
+    if (layout?.fullWidth) {
+        style.gridColumn = '1/-1';
```
adrianschmidt, line 10: Had to google this one 😄

If I understand MDN's description properly, this means the item will start at the first column and end at the last column (precisely as expected). Or, more technically, the `1` means it should start at the first column as counted from the start, and the `-1` means it should end at the first column from the end… 😅

jgroth: Yeah I think that's right :slightly_smiling_face:  It's not the most intuitive syntax and that's why I was hesitant to have `span` on the interface.
Maybe we could have `span: number | 'all'` or something like that though because I'm not 100% happy about the `fullWidth: boolean` either :disappointed:

src/components/form/form.types.ts
adrianschmidt: Yeah, with `span` you'd have to know how many columns there are I think. I think if we want to give more fine-grained control than just "one column or full width" later on, we might look into `span`. Or we might just go to full configuration at that point, more like the start page grid…

src/components/form/templates/field.ts
adrianschmidt: Hmm… yeah, that's a pretty neat idea, if we can make that work.

On the other hand, do we really need it right now? Perhaps we could do `span: 'all'` for now? That would just work exactly like the `fullWidth: boolean` for now, but it would be ready to grow into `span: number | 'all'`…

adrianschmidt (approved):

src/components/form/examples/form-layout.tsx
```diff
@@ -0,0 +1,57 @@
+import { ValidationStatus } from '@limetech/lime-elements';
+import { Component, h, State } from '@stencil/core';
+import { schema } from './schema-layout';
+
+/**
+ * @link schema-layout.ts
```
adrianschmidt, line 6:
```suggestion
/**
 * Layout
 *
 * @link schema-layout.ts
```

src/components/form/templates/field.ts
jgroth: No we don't really need it right now, but we also want to make sure that it will be somewhat compatible with future changes so we don't have to make it breaking right away :smile:

adrianschmidt: But if we make it `span?: 'all'` now, changing it to `span?: number | 'all'` later shouldn't be breaking I think? That change should only be breaking to the implementer, but not to the consumer (I think?), and since the implementation is in this same package, we should be able to make that change without it being breaking.

That way we don't have to spend time making number values work right now 🙂

jgroth: Yeah let's go with the `span` option instead, it feels more natural and I think we should be able to implement numbers without major issues in the future :slightly_smiling_face:

adrianschmidt: 😍

src/components/form/form.types.ts
```diff
@@ -104,3 +104,73 @@ export interface FormInfo {
...
+ * ```ts
+ * const schema = {
+ *     type: 'object',
+ *     lime: {
+ *         collapsible: true,
+ *     },
+ * };
+ * ```
+ */
+export interface LimeSchemaOptions {
```
adrianschmidt, line 121: Not for this PR, but:

It's *really* nice to have this in the docs, but it's kind of hard to find, since the `schema` property is just typed as `{ [key: string]: any; id?: string; }`. Perhaps we can either expand that typing a little in the future, or just add some text with links to helpful "inner" typings… 🙂

adrianschmidt (approved):

PR #994, opened 2020-10-23
fix(switch): ensure no exception is thrown in the console

jensgustafsson:
fix: Lundalogik/crm-feature#1514

This fix ensures that the switch value is not updated if the
component has not loaded.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/switch/switch.tsx
+++ b/src/components/switch/switch.tsx
@@ -107,6 +107,10 @@ export class Switch {
 
     @Watch('value')
     protected valueWatcher(newValue, oldValue) {
+        if (!this.mdcSwitch) {
+            return;
+        }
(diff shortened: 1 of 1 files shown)


adrianschmidt (approved):
I can't replicate the original problem, so I can't really test this, but I've tested against regressions and I'm happy with how the code looks (and the change is simple enough) 👍

PR #999, opened 2020-10-23
fix(table): ensure component is completely rendered when loaded

jensgustafsson:
fix: https://github.com/Lundalogik/crm-feature/issues/1450

Tabulator requires that the html element it's rendered inside
has a size before it's created, otherwise it doesn't consider
it self renderedy completely. (the callback "renderComplete"
is never run).

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -122,6 +122,7 @@ export class Table {
         this.onClickRow = this.onClickRow.bind(this);
         this.formatRow = this.formatRow.bind(this);
         this.updateMaxPage = this.updateMaxPage.bind(this);
+        this.initTabulatorComponent = this.initTabulatorComponent.bind(this);
         this.pool = new ElementPool(document);
         this.columnFactory = new ColumnDefinitionFactory(this.pool);
     }
@@ -139,7 +140,8 @@ export class Table {
         const table: HTMLElement = this.host.shadowRoot.querySelector(
             '#tabulator-table'
         );
-        this.tabulator = new TabulatorTable(table, options);
+
+        this.initTabulatorComponent(table, options);
     }
 
     public disconnectedCallback() {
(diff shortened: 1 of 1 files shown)


src/components/table/table.tsx
```diff
@@ -175,6 +177,28 @@ export class Table {
...
+     * has a size before it's created, otherwise it doesn't consider
+     * it self renderedy completely. (the callback "renderComplete"
+     * is never run).
+     *
+     * @param table {HTMLElement}
+     * @param options {abulator.Options}
+     *
+     * @returns {void}
+     */
+    private initTabulatorComponent(
```
Befkadu1, line 191: What if we call the name `initTabulatorTable` since `Tabulator` is a class ?

Befkadu1:
The tables not loading on phones. I verified it on browserstack. 
Device: Iphone 11
Safari

Screenshot 2020-10-28 at 09 59 26

Befkadu1 (changes requested):
The tables not loading on phones. This shows only on browserstack. We probably need to check it on the real phone
Device: Iphone 11
Safari
Screenshot 2020-10-28 at 10 41 17

As you see on the log, there is a problem regarding `ResizeObserver`.

jensgustafsson:
Good catch @Befkadu1!

src/components/table/table.tsx
```diff
@@ -175,6 +177,28 @@ export class Table {
...
     }
 
+    /*
+     * Tabulator requires that the html element it's rendered inside
+     * has a size before it's created, otherwise it doesn't consider
+     * it self renderedy completely. (the callback "renderComplete"
+     * is never run).
+     *
+     * @param table {HTMLElement}
+     * @param options {abulator.Options}
```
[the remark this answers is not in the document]
jensgustafsson: ⚡

jensgustafsson:
> The tables not loading on phones. This shows only on browserstack. We probably need to check it on the real phone
> Device: Iphone 11
> Safari
> Screenshot 2020-10-28 at 10 41 17
> 
> As you see on the log, there is a problem regarding `ResizeObserver`.

⚡

src/components/table/table.tsx
```diff
@@ -175,6 +177,35 @@ export class Table {
...
+    private initTabulatorComponent(
+        table: HTMLElement,
+        options: Tabulator.Options
+    ) {
+        // Some browsers does not implement the ResizeObserver API...
+        // If that's the case lets just create the table no
+        // matter if its rendered or not.
+        if (!('ResizeObserver' in window)) {
+            this.tabulator = new TabulatorTable(table, options);
+        }
```
Befkadu1, line 200: Should we add `return; ` to avoid the error logs.

Befkadu1: It means we are going to have a broken table if the browsers don't implement the ResizeObserver API?

src/components/table/table.tsx
```diff
@@ -175,6 +177,35 @@ export class Table {
...
+     * @param table {HTMLElement}
+     * @param options {Tabulator.Options}
+     *
+     * @returns {void}
+     */
+    private initTabulatorComponent(
+        table: HTMLElement,
+        options: Tabulator.Options
+    ) {
+        // Some browsers does not implement the ResizeObserver API...
```
Befkadu1, line 195:
```suggestion
        // Some browsers do not implement the ResizeObserver API...
```

src/components/table/table.tsx
jensgustafsson: You are right!

Yes but it works after you have sorted the table and the problem doesn't seem to appear in the webclient on iphone :)

Befkadu1 (approved):

PR #1001, opened 2020-10-27
feat(input-field): add support for regex pattern

FredrikWallstrom:
Required by https://github.com/Lundalogik/limepkg-esign/pull/530

Commit one:
* We want to use the pattern attribute provided on the input tag.
Pattern is not supported for all available input types, for example not for number. How should we handle this? Seems like pattern is not applied when I'm using it with `number` type, so it don't break at least. Is that enough?

Commit two:
* Discover some weirdness. 
I was trying to change `this.value` programmatically from our add-on without keystrokes so it matched the pattern. However, it turned out that `mdcTextValue` was out of sync with `this.value`, so the UI displayed that the value did not match the pattern even though it was a match!

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/examples/input-field-pattern.tsx
+++ b/src/components/input-field/examples/input-field-pattern.tsx
@@ -1,0 +1,32 @@
+import { Component, h, State } from '@stencil/core';
+
+/**
+ * Input Field with pattern
+ */
+@Component({
+    tag: 'limel-example-input-field-pattern',
+    shadow: true,
+})
+export class InputFieldPatternExample {
+    @State()
+    private value;
+
+    constructor() {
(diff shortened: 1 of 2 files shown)


jgroth (approved):
Great work! I updated the documentaion a bit to clarify that it's only used for certain values of `type`

FredrikWallstrom:
> Great work! I updated the documentaion a bit to clarify that it's only used for certain values of `type`

Thanks!

PR #1010, opened 2020-10-30
Docs bring back guidelines for tabs

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tab-bar/tab-bar.tsx
+++ b/src/components/tab-bar/tab-bar.tsx
@@ -18,6 +18,21 @@ const SCROLL_DISTANCE_ON_CLICK_PX = 150;
 const HIDE_SCROLL_BUTTONS_WHEN_SCROLLED_LESS_THAN_PX = 40;
 
 /**
+ * Tabs are great to organize information hierarchically in the interface and divide it into distinct categories. Using tabs, you can create groups of content that are related and at the same level in the hierarchy.
+ * :::warning
+ * Tab bars should be strictly used for navigation at the top levels.
+ * They should never be used to perform actions, or navigate away from the view which contains them.
+ * :::
+ * An exception for using tab bars in a high level of hierarchy is their usage in modals. This is because modals are perceived as a separate place and not a part of the current context. Therefore you can use tab bars in a modal to group and organize its content.
+ * A tab bar can contain an unlimited number of tabs. However, depending on the device width and width of the tabs, the number of tabs that are visible at the same time will vary. When there is limited horizontal space, the component shows a left-arrow and/or right-arrow button, which scrolls and reveals the additional tabs. The tab bar can also be swiped left and right on a touch-device.
+ *:::tip Other things to consider
+ * * Never divide the content of a tab using a nested tab bar.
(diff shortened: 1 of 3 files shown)


src/components/tab-bar/tab-bar.tsx
```diff
@@ -18,6 +18,21 @@ const SCROLL_DISTANCE_ON_CLICK_PX = 150;
...
+ * An exception for using tab bars in a high level of hierarchy is their usage in modals. Since modals are perceived as a separate place and not a part of the current context. Therefore you can even use tab bars in a modal to group and organize their content.
+ * A tab bar can contain unlimited number of tabs. However, depending on the device width and width of the tabs, the number of visible tabs varies. When there is limited horizontal space, the component shows a "next" or "back" button, which scrolls and reveals the additional tabs.
+ *:::tip Other things to consider
+ * * Never divide the content of a tab using a nested tab bar.
+ * * Never place two tab bars within the same screen.
+ * * Never use background color for icons in tabs.
+ * * Avoid having long labels for tabs.
+ * * A tab will never be hidden or get disabled, even if there is no content under it.
+ * :::
+ *
```
adrianschmidt, line 35:
```suggestion
 * Tabs are great to organize information hierarchically in the interface and divide it into distinct categories. Using tabs, you can create groups of content that are related and at the same level in the hierarchy.
 * :::warning
 * Tab bars should be strictly used for navigation at the top levels.
 * They should never be used to perform actions, or navigate away from the view which contains them.
 * :::
 * An exception for using tab bars in a high level of hierarchy is their usage in modals. This is because modals are perceived as a separate place and not a part of the current context. Therefore you can use tab bars in a modal to group and organize its content.
 * A tab bar can contain an unlimited number of tabs. However, depending on the device width and width of the tabs, the number of tabs that are visible at the same time will vary. When there is limited horizontal space, the component shows a left-arrow and/or right-arrow button, which scrolls and reveals the additional tabs. The tab bar can also be swiped left and right on a touch-device.
 *:::tip Other things to consider
 * * Never divide the content of a tab using a nested tab bar.
 * * Never place two tab bars within the same screen.
 * * Never use background color for icons in tabs.
 * * Avoid having long labels for tabs.
 * * A tab will never be removed or get disabled, even if there is no content under it.
 * :::
 *
```

adrianschmidt (commented):
Really nice!

I assume I've reviewed this before, but I seemed to have missed some minor things, so I made a suggestion for a change.

src/components/tab-bar/tab-bar.tsx
Kiarokh: nice! :)

Kiarokh:
@adrianschmidt added a fixup

adrianschmidt (approved):

PR #1011, opened 2020-10-30
fix(input-field): yet *another* attempt at fixing the issue of scrolling numeric input fields

adrianschmidt:
fix Lundalogik/crm-feature#1538

## Review:

### Browsers tested:

Windows:
- [x] Chrome
- [x] Edge
- [x] Firefox

Linux:
- [x] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [x] Firefox
- [x] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -256,6 +256,7 @@ export class InputField {
                     disabled={this.disabled}
                     type={this.type}
                     pattern={this.pattern}
+                    onWheel={this.handleWheel}
                     onKeyDown={this.onKeyDown}
                     {...additionalProps}
                     value={this.value}
@@ -627,5 +628,17 @@ export class InputField {
         if (isSpace || isEnter) {
             this.action.emit();
         }
+    }
+
+    private handleWheel() {
+        // This empty event handler is here to circumvent a bug.
+        // In some browsers (Chrome for example), hovering the input with
+        // the input focused, and scrolling, will both change the value
+        // AND scroll the page. We would prefer to never change the value
+        // on scroll, instead always scrolling the page, but since we
+        // haven't found a way to do that, this is the next best thing, as
+        // it prevents the page from being scrolled, but only in the
+        // circumstances when the value is changed by the scrolling.
+        // Please test THOROUGHLY if you remove this event handler 😄
     }
 }


adrianschmidt:
Browsers I've tested in:

A) The behaviour we want; scroll never changes the value, always scrolls the page.
B) The next best thing; when hovering a focused input, scrolling changes the value, but at least the page doesn't scroll. In all other circumstances, scrolling scolls the page, and does not change the value.

Windows:
- [x] Chrome (B)
- [x] Edge (B)
- [x] Firefox (B)
- [x] Opera (A)

Linux:
- No Linux on BrowserStack 😞

macOS:
- [x] Chrome (B)
- [x] Firefox (A)
- [x] Safari (A)

Mobile:
- [x] Chrome on Android (touch doesn't initiate scroll-events, don't know if it's possible with some accessory)
- [x] iOS (touch doesn't initiate scroll-events, don't know if it's possible with some accessory)

src/components/input-field/input-field.tsx
```diff
@@ -624,4 +625,17 @@ export class InputField {
...
+    private handleWheel() {
+        if (this.type === 'number') {
+            // In some browsers (Chrome for example), hovering the input with
+            // the input focused, and scrolling, will both change the value
+            // AND scroll the page. We would prefer to never change the value
+            // on scroll, instead always scrolling the page, but since we
+            // haven't found a way to do that, this is the next best thing, as
+            // it prevents the page from being scrolled, but only in the
+            // circumstances when the value is changed by the scrolling.
+            // Please test THOROUGHLY if you remove it 😄
```
jgroth, line 637: Maybe remove this last line before we merge :smile:

src/components/input-field/input-field.tsx
```diff
@@ -624,4 +625,17 @@ export class InputField {
...
+            return true;
```
jgroth, line 638: What does this do? :thinking: I tried looking it up but didn't find anything

jgroth (dismissed):
Tested it in Chrome, Safari and Firefox on Mac and Windows, and Chrome on Linux. Also tested on Safari on iPhone and Chrome on Android. The behavior was the same everywhere except for Firefox on Mac where the number did not change even if the field was focused while scrolling

src/components/input-field/input-field.tsx
adrianschmidt: It doesn't seem to do anything. The hack works regardless of whether we return `true`, `false`, or nothing. But since returning `false` from an event-listener used to mean approximately `preventDefault`, and we don't wan't to do that, I was thinking `return true` was probably the closest thing to something that makes sense… but maybe not having a return statement at all makes it clearer that it is the event-listener itself that seems to fix the problem, rather than what is returned from it…?

src/components/input-field/input-field.tsx
adrianschmidt: No, that's meant to be left in, in case anyone gets the idea to fiddle with this sometime in the future…

But maybe the whole comment needs some improvement 🙂

src/components/input-field/input-field.tsx
adrianschmidt: Oh, by the way, there shouldn't be any need for the `if` either I guess… that's just left there from when we where doing `event.preventDefault` and stuff…

I'll experiment a little 😄

src/components/input-field/input-field.tsx
jgroth: Oh yeah right, I read it as something completely different haha :smile:

src/components/input-field/input-field.tsx
adrianschmidt: Pushed a fixup.

adrianschmidt:
Oh, by the way, I noticed that the behaviour on master is different if you're in the docs or testing in lime-webclient. At least it was for me in Chrome on macOS. I can't figure out why there's a difference though…

Anyhow, with this fix the behaviour is the same in both cases.

jgroth (dismissed):

jgroth (approved):

PR #1019, opened 2020-11-03
Input field refactor and bugfixes

adrianschmidt:
The commit `fix(input-field): do not trigger action from leadingIcon` could be considered breaking. It could also be considered to just be a bugfix that's only breaking if someone has exploited that bug to do something we never intended. But I'm not entirely sure that's fair, because I'm not entirely sure we never intended the leading icon to trigger actions… Opinions welcome!

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/examples/input-field-icon-both.tsx
+++ b/src/components/input-field/examples/input-field-icon-both.tsx
@@ -1,0 +1,40 @@
+import { Component, h, State } from '@stencil/core';
+
+/**
+ * Input Field with Leading and Trailing Icons & Action
+ */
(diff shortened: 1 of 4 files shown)


jgroth (approved):
Great work! :slightly_smiling_face:

PR #1023, opened 2020-11-03
Allow set limel-button size

### LINKED ISSUE #1004: limel-button doesn't change size when changing the size of the `<limel-button>` element
## Current behavior

When setting the height and width of `<limel-button>`, only the container element changes size:

image

## Expected behavior

The button should follow the size of the container element:

image

## Environment

- **lime-elements** version: 29.3.1 but probably affects all versions
- Framework used: reproducible in docs
- Logs:

anderssonjohan:
fixes #1004

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -21,6 +21,9 @@ button {
     &.mdc-button {
         min-width: pxToRem(36);
         padding: 0 pxToRem(12);
+        height: 100%;
+        min-height: pxToRem(36);
+        width: 100%;
     }
 
     .label {
--- a/src/components/button/examples/button.scss
+++ b/src/components/button/examples/button.scss
@@ -3,6 +3,5 @@ limel-switch {
 }
 
 limel-button {
-    display: block;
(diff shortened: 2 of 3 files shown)


src/components/button/button.scss
```diff
@@ -21,6 +21,9 @@ button {
     &.mdc-button {
         min-width: pxToRem(36);
         padding: 0 pxToRem(12);
+        height: 100%;
+        min-height: pxToRem(36);
```
adrianschmidt, line 25: I changed this to using `pxToRem`. My bad, since I used plain rem in my comment on the issue 🙂

adrianschmidt (approved):
Great!

I updated the commit messages slightly, so that they don't mention MDC and so the fix for the docs is typed as `docs`.

src/components/button/button.scss
anderssonjohan: I'm only thinking rem now so my bad not noticing the use of px in the rules next to my additions! 😁

PR #1024, opened 2020-11-04
feat(form): add support for different types of date-pickers

maltja:
fix Lundalogik/crm-feature#1584

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/widgets/date-picker.ts
+++ b/src/components/form/widgets/date-picker.ts
@@ -1,6 +1,5 @@
 import moment from 'moment/moment';
 import React from 'react';
-import { DateType } from '../../date-picker/date.types';
 import { WidgetProps } from './types';
 import { LimeElementsWidgetAdapter } from '../adapters';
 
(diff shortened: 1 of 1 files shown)


src/components/form/widgets/date-picker.ts
```diff
@@ -17,7 +17,8 @@ export class DatePicker extends React.Component {
 
     public render() {
         const props: WidgetProps = this.props;
-        const type = getDateType(props.schema);
+        this.props.schema.format = getDateFormat(props.schema);
```
jgroth, line 20: I don't think we should change this here :thinking: The consumer has given us a schema and it will probably be confusing and might lead to bugs if we change it.
We should either fix this issue in the backend when we serialize the schema or in the webclient when we add the property controls
I think I prefer the backend solution :slightly_smiling_face:

jgroth (changes requested):

src/components/form/widgets/date-picker.ts
maltja: Ah that is true, I will try to do it in the backend then 😄

jgroth (approved):

jgroth (approved):

PR #1025, opened 2020-11-04
feat(input-field): add option `showLink` for type `email`, `tel`, and `url`

adrianschmidt:
For inputs of type `email`, `tel`, and `url`, `showLink` can be set to `true` to show a trailing icon with a `mailto:`,`tel:`, or normal link, respectively. The default icon can be overridden using the `trailingIcon` property.

fix Lundalogik/crm-feature#1565

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -37,6 +37,7 @@ interface LinkProperties {
  * @exampleComponent limel-example-input-field-icon-leading
  * @exampleComponent limel-example-input-field-icon-trailing
  * @exampleComponent limel-example-input-field-icon-both
+ * @exampleComponent limel-example-input-field-showlink
  * @exampleComponent limel-example-input-field-error-icon
  * @exampleComponent limel-example-input-field-textarea
  * @exampleComponent limel-example-input-field-search
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -24,6 +24,11 @@ import {
 import { InputType } from './input-field.types';
 import { ListItem } from '@limetech/lime-elements';
 
+interface LinkProperties {
+    href: string;
+    target?: string;
+}
+
 /**
  * @exampleComponent limel-example-input-field-text
  * @exampleComponent limel-example-input-field-text-inline
@@ -155,6 +160,14 @@ export class InputField {
      */
     @Prop()
     public completions: string[] = [];
+
+    /**
+     * For inputs of type `email`, `tel`, and `url`, set this to `true` to show
+     * a trailing icon with a `mailto:`,`tel:`, or normal link, respectively.
+     * The default icon can be overridden using the `trailingIcon` property.
+     */
+    @Prop({ reflect: true })
+    public showLink = false;
 
     /**
      * Emitted when the input value is changed.
@@ -450,8 +463,6 @@ export class InputField {
     }
 
     private renderIcons() {
-        const trailingIcon = this.getTrailingIcon();
-
         const html = [];
 
         if (this.leadingIcon) {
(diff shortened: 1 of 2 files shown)


src/components/input-field/input-field.tsx
```diff
@@ -462,7 +473,37 @@ export class InputField {
...
+                    href={link}
+                    class="mdc-text-field__icon"
+                    tabindex={this.disabled ? '-1' : '0'}
+                    role="button"
+                    {...props}
+                >
+                    <limel-icon name={trailingIcon} />
+                </a>
+            );
+        } else if (trailingIcon) {
```
adrianschmidt, line 506: I think there may be room for some refactoring to make this nicer, but it's what I have for tonight 😅

src/components/input-field/input-field.tsx
```diff
@@ -25,6 +25,9 @@ import { InputType } from './input-field.types';
 import { ListItem } from '@limetech/lime-elements';
 
 /**
+ * @exampleComponent limel-example-input-field-email
+ * @exampleComponent limel-example-input-field-tel
+ * @exampleComponent limel-example-input-field-url
```
jgroth, line 30: Is there a reason why these are before the more basic examples? :slightly_smiling_face:

src/components/input-field/input-field.tsx
```diff
@@ -462,7 +473,37 @@ export class InputField {
             );
         }
 
-        if (trailingIcon) {
+        if (
+            !this.isInvalid() &&
+            this.showLink &&
+            ['email', 'tel', 'url'].includes(this.type)
```
jgroth, line 479: Maybe move this to a separate function? How about `hasLink` or something like that?

src/components/input-field/input-field.tsx
```diff
@@ -462,7 +473,37 @@ export class InputField {
...
+                case 'email':
+                    link = `mailto:${this.value}`;
+                    break;
+                case 'tel':
+                    link = `tel:${this.value}`;
+                    break;
+                default:
+                    link = `${this.value}`;
+                    props.target = '_blank';
+            }
```
jgroth, line 493: This can also be extracted to a spearate function :slightly_smiling_face: 
E.g. `getLink`

src/components/input-field/input-field.tsx
jgroth: Yeah both `if`/`else` blocks can probably be extracted to something like
```ts
if (!this.isInvalid() && this.hasLink()) {
    html.push(this.renderLinkIcon());
} else if (trailingIcon) {
    html.push(this.renderTrailingIcon());
}
```

src/components/input-field/examples/input-field-url.tsx
```diff
@@ -0,0 +1,78 @@
+import { Component, h, State, Watch } from '@stencil/core';
```
jgroth, line 1: These three examples are quite similar. What do you think of combining them into a single one? :thinking:

src/components/input-field/input-field.tsx
adrianschmidt: LOL!

Yes, because I wanted them to be easily accessible while developing, and then totally forgot about moving them down.

Good catch! 🙏

src/components/input-field/examples/input-field-url.tsx
adrianschmidt: Yeah, I've considered it.

I think combining them makes the live example better, but it also makes the code-side of the example worse, as it becomes more complex. I'll ask the add-ons team what they think.

src/components/input-field/examples/input-field-tel.tsx
```diff
@@ -0,0 +1,85 @@
...
+                label="Tel Field"
+                helperText="Please enter a phone number!"
+                minlength={MIN_LENGTH}
+                value={this.value}
+                required={this.required}
+                invalid={this.invalid}
+                disabled={this.disabled}
+                onChange={this.changeHandler}
+                type="tel"
+                trailingIcon="dog"
```
adrianschmidt, line 49: Oops 😆

src/components/input-field/input-field.tsx
adrianschmidt: ⚡

src/components/input-field/input-field.tsx
adrianschmidt: ⚡

src/components/input-field/input-field.tsx
adrianschmidt: ⚡

src/components/input-field/input-field.tsx
adrianschmidt: ⚡

src/components/input-field/examples/input-field-url.tsx
adrianschmidt: Katja was about as undecided as I am… 🤷‍♂️

src/components/input-field/examples/input-field-tel.tsx
adrianschmidt: ⚡

src/components/input-field/input-field.tsx
```diff
@@ -24,6 +24,11 @@ import {
 import { InputType } from './input-field.types';
 import { ListItem } from '@limetech/lime-elements';
 
+interface linkProperties {
```
jgroth, line 27: Interfaces usually start with a capital letter
```suggestion
interface LinkProperties {
```

src/components/input-field/examples/input-field-url.tsx
jgroth: Haha ok, I will let you decide what you think is best. My argument for combining them is that there will be less code to maintain and I don't think the example would be _that_ much more complex :slightly_smiling_face:

src/components/input-field/input-field.tsx
adrianschmidt: ⚡

src/components/input-field/examples/input-field-url.tsx
adrianschmidt: I've pushed a fixup where all three types still have their own fields, but they've been combined into a single example. Not sure that's any better. Showing only one field but with a select for the type might be better, but I'm a little too tired to take on doing that right now. So try this one out and see what you think.

jgroth: I think this looks good! :+1:

jgroth (approved):

jgroth (approved):

PR #1028, opened 2020-11-09
feat(button): add `has-reduced-presence` helper class

adrianschmidt:
Adding `has-reduced-presence` to limel-button hides the button when it is disabled and not either loading or showing the checkmark after loading.

fix: Lundalogik/crm-feature#1581

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -3,6 +3,7 @@ import { Component, Element, h, Prop, Watch } from '@stencil/core';
 /**
  * @exampleComponent limel-example-button
  * @exampleComponent limel-example-button-click
+ * @exampleComponent limel-example-button-reduce-presence
  */
 @Component({
     tag: 'limel-button',
--- a/src/components/button/examples/button-reduce-presence.tsx
+++ b/src/components/button/examples/button-reduce-presence.tsx
@@ -1,0 +1,51 @@
+import { Component, h, State } from '@stencil/core';
+
+/**
+ * Reduce Presence
+ *
+ * This example is identical to the one above, except that here, the
+ * `has-reduced-presence` class has been set to `true`. This will hide the
+ * button when it is disabled. However, it will also make sure that the button
+ * remains visible while the loading animation is ongoing. When the animation is
+ * done and the checkmark has been shown, the button will hide.
+ *
+ * Read more in the [Design Guidelines](/#/DesignGuidelines/decluttering.md)
+ */
+@Component({
+    tag: 'limel-example-button-reduce-presence',
+    shadow: true,
+})
+export class ButtonReducePresenceExample {
+    @State()
+    private loading = false;
+
+    @State()
+    private disabled = false;
+
+    public render() {
+        return (
+            <limel-button
+                class="has-reduced-presence"
+                label="Click me!"
+                primary={true}
+                loading={this.loading}
+                disabled={this.disabled}
+                onClick={this.onClick}
+            />
+        );
+    }
+
+    private onClick() {
+        this.disabled = true;
+        this.loading = true;
+
+        const TIME_LOADING = 1000;
+        const TIME_DISABLED = 5000;
+        setTimeout(() => {
+            this.loading = false;
+            setTimeout(() => {
+                this.disabled = false;
+            }, TIME_DISABLED);
+        }, TIME_LOADING);
+    }
+}
--- a/src/decluttering.md
+++ b/src/decluttering.md
@@ -1,0 +1,24 @@
+# Reducing UI clutter and cognitive load
+
+A user interface with too many elements requires more processing from users' brains. When fewer things on the screen try to attract the user's attention, it becomes easier for the user to "consume" the UI. This is what we mean when we talk about "reducing cognitive load". In the end, our goal is for the user to spend their mental energy on the problem at hand, not at the tool they are using to solve it (our UI).
+
+## Buttons
+
+Buttons are particularly strong elements in the UI, since they are meant to perform important actions. Thus, an effective way of reducing clutter is to hide buttons that aren't useful at the moment. This most commonly applies to disabled buttons.
+
+Disabled buttons can convey important information, but when they don't, they should be hidden. Here are some examples:
+
+1. **When there are invalid fields, or empty required fields**
+   We can anticipate that the user might try to save their changes, and showing a disabled Save button is part of telling the user that there is something they need to do before saving is possible. This should be combined with other clear visual hints on what to do, like highlighting the invalid field and displaying a validation error message.
+2. **When the user has made no changes**
+   This example can be divided into two groups:
+   - *The user is in a distinct "flow" with discrete steps, and is prevented from continuing to the next step.* Common examples are so called "wizards". In this case, a disabled Save or Continue button should be kept visible, for the same reason as in example 1.
+   - *Making and saving changes is just one of the possible "things to do" on the current page.* Examples include forms that are used both for displaying and changing information (common for configuration or settings pages), or a feed, with an input for adding new posts. In this case, a disabled Save button isn't useful. We don't even know that the user has any intention of using it. Once the user updates the information, or starts writing in the input field, it's time to display the Save button, along with any other controls that might have also been hidden.
+
+:::note
+Keep in mind that a SAVE button can be disabled but visible, to tell the
+users that requirements are not met, or there are errors.
+
+Such cases should be accompanied with other clear visual hints where the
+errors or progress-blockers are located in the UI.
+:::
--- a/guides.ts
+++ b/guides.ts
@@ -7,6 +7,10 @@ export default [
     },
     {
         name: 'DesignGuidelines',
-        children: ['src/color-system.md', 'src/size-rhythms.md'],
+        children: [
+            'src/color-system.md',
+            'src/decluttering.md',
+            'src/size-rhythms.md',
+        ],
     },
 ] as Guide[];
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -84,3 +84,20 @@ button {
         }
     }
 }
+
+:host(.has-reduced-presence) {
+    button {
+        transition: opacity 300ms ease-in-out;
+
+        &[disabled]:not(.loading):not(.just-loaded) {
+            opacity: 0;
+
+            limel-icon,
+            .label {
+                // Avoid fading in the label while
+                // simultaneously fading out the button.
+                opacity: 0;
+            }
+        }
+    }
+}
--- a/src/components/button/examples/button-click.tsx
+++ b/src/components/button/examples/button-click.tsx
@@ -5,7 +5,7 @@ import { Component, h, State } from '@stencil/core';
  *
  * The click handler in this example sets the attributes `loading` and
  * `disabled` to `true`. After 1 second, the `loading` attribute is set to
- * `false` again. After another 4 seconds, the button is once again enabled.
+ * `false` again. After another 5 seconds, the button is once again enabled.
  *
  * When the `loading` attribute changes from `true` to `false`, the button
  * automatically displays a checkmark icon for 2 seconds. Note that our click
@@ -39,7 +39,7 @@ export class ButtonClickExample {
         this.loading = true;
 
         const TIME_LOADING = 1000;
-        const TIME_DISABLED = 4000;
+        const TIME_DISABLED = 5000;
         setTimeout(() => {
             this.loading = false;
             setTimeout(() => {


src/components/button/examples/button-hide-when-disabled.tsx
```diff
@@ -0,0 +1,51 @@
...
+/**
+ * Hide When Disabled
+ *
+ * This example is identical to the one above, except that here, the
+ * `hideWhenDisabled` property has been set to `true`, hiding the button when
+ * it is disabled and not either loading or showing the checkmark after loading.
+ * This is useful for limiting clutter in the interface.
+ *
+ * Note that sometimes, a button can provide useful information to the user,
+ * even when it is disabled.
```
adrianschmidt, line 12: @Kiarokh might want to add some instructions here?

src/components/button/button.tsx
```diff
@@ -54,6 +55,14 @@ export class Button {
     @Prop({ reflect: true })
     public loading = false;
 
+    /**
+     * Set to `true` to make the button automatically hide when disabled.
+     * While loading or showing the checkmark after loading, the button will
+     * still be visible.
+     */
+    @Prop({ reflect: true })
+    public hideWhenDisabled = false;
```
jgroth, line 64: My main argument against this is that it makes the API less clean. If the consumer wants to hide the button they can just hide it themselves. They know when they want to disable it and therefore also know when they do not want to display it. 

Also, why even set it to `disabled` in the first place if they are going to hide it? :thinking: `disable` will loose it's meaning since the button is not visible anyway. Another approach would be to use the global html `hidden` attribute instead if they want to hide it. Thoughts? :slightly_smiling_face:

adrianschmidt: The main reason was to not have the consumer account for the time the "loading done" checkmark is shown after loading. Granted, we should give the consumer some more control over that, because it really should only be shown if something was being saved or sent or similar, and the process was successful.

But, handing off the *whole* thing to the consumer won't really work, because the checkmark isn't just a different icon for the label, it's replacing the label, and there are fading transitions between the label, spinner, checkmark, and back to the label again. And we want the timing to consistent in every place the checkmark is shown.

We can take care of all of those things, and present them to the consumer through a very simple interface, where they essentially just "tick the boxes" for the features they want, and don't have to care about timing and transitions and stuff.

Kiarokh: I agree with @adrianschmidt. The benefits are that:
1. We're in control, if anything changes in future. Let's say an animation length changes. Or the progress reflects the actual time the process is taking.
1. Consumers don't have to think and find solutions. They just say I want this feature. All they have to focus on is when to make the button disable.

Kiarokh: Regarding the HTML attribute `<button hidden>`. It's up to you Adrian. This won't be a CSS-only-fix anymore though, as a script should add the `hidden` attribute, after the animation has run.

adrianschmidt: We can use the `hidden` attribute to hide the button, instead of adding a class `.hidden`.

But we can't use the `hidden` attribute instead of the `hideWhenDisabled` because they do different things. `hideWhenDisabled` doesn't hide the button. It just tells the button that when it's disabled, *if* there is no other information it needs to convey, like showing the spinner or the checkmark, then it should hide itself.

adrianschmidt: Personally I don't think it's a problem that the API becomes larger, since these more complicated options are optional. But we could provide a `limel-button` with a much simpler API, and a `limel-complex-button` or something, that provides all the special options…

jgroth: Hmm ok, I had forgotten about all the spinner stuff I think, and I'm not even sure if I get how all that works :smile: 

I agree that it's too much work for the consumer to take care of that, they should not have to worry about the timing of things etc. I'm also not against making the API larger, I'm mostly against mixing concepts that seem unrelated :slightly_smiling_face: Maybe it's just a naming thing? Maybe we can find a better name for the property, because I still think that `disabled` and `hidden` are two separate things and should not be coupled like this :thinking: Why set the state to disabled when you want to hide it?

adrianschmidt: Ah, now I understand where you're coming from 😄

If your primary goal is to hide the button, then no, you shouldn't use this property. Then you should just set the "native" `hidden` attribute.

Instead, think of this as setting an alternative disabled-style for the button. Don't think of this as setting a *state*, think of it as setting a *type* of button, just like `primary`. And `disabled` shouldn't be set to hide the button, `disabled` should be set to disable the button, and, if appropriate based on the circumstances, the alternative disabled-styling will be applied.

adrianschmidt: We can definitely rename the property, if we can think of a better name 👍

jgroth: I had another thought about this yesterday. Do we never want to show a disabled button on the card? What about when the form is invalid? We are not using the validation status of the form at the moment, but if we do that do we want to hide the button when the form has been edited but is invalid or do we want to show the disabled button instead? Or should we display some other indicator that the form has been altered but is invalid? I guess this might affect how we implement this :thinking:

adrianschmidt: Yeah, you are absolutely correct. If the form is invalid, we want to disable the button, but keep it visible. So if we would go with my implementation, there would be cases where the consumer would flip `hideWhenDisabled` to `false`.

I understand your concerns, and I think I mostly agree, but I can't come up with a better alternative at the moment. Do you have any ideas?

jgroth: I think it's good if we keep things as simple as possible and let the consumer be in control. We cannot stop anyone from using the `hidden` attribute and that will hide the button no matter what (I think?), so if we want to have a nicer experience and be in control in regards to animations, spinners etc. we have to use something else. What about just having something like `show` or `hide`? Setting that could hide the button but only when all the animations etc. are done. Will that solve our problems? :slightly_smiling_face:

adrianschmidt: So, basically just cut `WhenDisabled` from `hideWhenDisabled`? I'm not sure that's any less confusing, but yeah, I guess…

Maybe we should call it something completely different, and maybe not even add it as a property, but as a helper-class. Similar to `has-low-density` and `has-interactive-rows` on limel-table?

Something like `has-reduced-presence`?

jgroth: Yeah that might be a better idea :slightly_smiling_face:

Kiarokh: or something like `hides-when-not-needed` 🙃

adrianschmidt: I would have liked `hides-when-not-needed` if the button itself would always know if it's needed or not. But it won't, unless we give it knowledge of the form and its validity. I'd like to avoid that 🙂

adrianschmidt: I've renamed the property, but haven't changed it to a helper-class yet.
I'll add another fixup for that, and we can see which one we like better.

adrianschmidt: Ah, no, making it into a helper-class makes things more difficult, because we can't just style directly off of that, because of the shadowDom boundary… so we'd have to set up some javascript to look for that class, and since the class may have to be removed and added again at times, we'd have to have some kind of watcher and stuff.

Unless I'm forgetting something because it's Friday afternoon, I think it's best we just go with the property.

src/components/button/examples/button-reduce-presence.tsx
```diff
@@ -0,0 +1,97 @@
...
+ * be saved.
+ * 2. **When the button tries to stop the user from proceeding** –
+ * Like a CONTINUE button in a form or a wizard which stops users from
+ * "continuing" because requirements are not met, required fields are not
+ * filled, or a form has errors.
+ *
+ *:::tip
+ * Sometimes it makes a lot of sense to simply hide a button that *has no
+ * functionality*, instead of leaving it `disabled` but visible.
+ * A good example of such a case is the SAVE and DISCARD buttons in a view.
```
adrianschmidt, line 25: > the SAVE and DISCARD buttons in a view

That only makes sense in the context of Lime CRM Web Client. Can we come up with an example that can be understood without that context? Could it just say "the SAVE and DISCARD buttons on a form" instead?

src/components/button/examples/button-reduce-presence.tsx
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Reduce Presence
+ *
+ * This example is identical to the one above, except that here, the
+ * `reducePresence` property has been set to `true`. This will hide the button
+ * when it is disabled. However, it will also make sure that the button remains
+ * visible while the loading animation is ongoing. When the animation is done
+ * and the checkmark has been shown, the button will hide.
+ *
+ * ###### Usage of `reducePresence` property
```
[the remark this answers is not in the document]
Kiarokh: It's never great to be honest. We should just remove it. 

I also wonder another thing. First line in these "comments" in the docs always becomes a `h5` that's why we have to write other headings under it as `h6`. @jgroth Can we set the first line to be a normal `p` instead, or at least set it to be `h2`? `h5` is brutal, and makes writing a good documentation really hard.

src/components/button/examples/button-reduce-presence.tsx
Kiarokh: By "view" I meant **any** view. But if you think it's less confusing, yes we can use "form" instead. ;)

src/components/button/examples/button-reduce-presence.tsx
adrianschmidt, line 12: I think everything from this heading down is a bit too long-form and philosophical to be a good fit here in the docs for the example. I think it might fit better under the design guidelines, and then we could have a more concise one-paragraph description here, with a link to the longer text.

src/components/button/button.tsx
jgroth: Hmm is that true? :thinking: 
I don't think we added any code for the classes that we added for the list, I think this should work even if we have the shadow boundary

src/components/button/examples/button-reduce-presence.tsx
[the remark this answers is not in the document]
jgroth: Yeah that should not be hard to fix I think, I'm not sure how it become like this with `h5` :smile:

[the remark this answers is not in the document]
adrianschmidt: I think I know, kind of 🙂 

Take this example:

image

- Button group is h1
- Usage and Layout are h2
- Examples is h3
- Icon only is h5

So for one, there's a gap between the h3 of *Examples* and the h5 of each example's heading. But also, *Examples* shouldn't be h3, because it's not a subsection of *Layout*.

So *Examples* should be promoted to an h2, and each example's first line should be promoted to h3 👍

src/components/button/examples/button-reduce-presence.tsx
adrianschmidt: Well, if that's the case, we can just drop the "in a view" part completely:

> A good example of such a case are SAVE and DISCARD buttons.

src/components/button/button.tsx
adrianschmidt: I'm not sure… I'll have a look at the tables 😄

adrianschmidt: What the heck?!

image

I didn't think it was possible to nest stuff inside the `:host` selector…

I swear I've tried using

```scss
:host {
	display: box;

	&[hidden] {
		display: none;
	}
}
```

But maybe it's just the `&` that doesn't work for the `:host`… 🤔

I've been living a LIE!!! 🤦‍♂️

adrianschmidt: Makes me the only problem was that I didn't add parenthesis… since I didn't do that in the example above… 😭

adrianschmidt: ⚡

jgroth: Yeah I think  you need those parentheses for some reason :upside_down_face:

src/components/button/examples/button-reduce-presence.tsx
Kiarokh: Hmm... 🤔  Good idea. What do we call that section in Design Guidelines? 
Perhaps: "**Reducing UI clutter**"

Can you make this sections? Then we can add even more info, examples, and guidelines there.

adrianschmidt: Yeah, I'm already working on it actually. Which I should have said of course. I hope we didn't do any double work? 🙂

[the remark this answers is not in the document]
adrianschmidt: I'm resolving this thread (the one about heading levels in the docs), since it's something we'll have to fix in Kompendium.

src/decluttering.md
```diff
@@ -0,0 +1,24 @@
+# Decluttring the UI and reducing cognitive load
```
hannahu, line 1: Missed an `e` here hehe,  Declutt`e`ring

hannahu, line 1: Was going to say an **e** is missed but _Decluttring_ is a word?

hannahu (approved):

src/decluttering.md
hannahu, line 1: I kinda don't like the _Decluttring_ word here. At first I didn't understand what it meant and also that the sentence sounds weird. Would it be possible to improve it? I saw that Kia mentioned `Reducing UI clutter` but you like this better?

hannahu (changes requested):

src/decluttering.md
Kiarokh: We could say "Reducing UI clutter and user's cognitive load"

src/components/button/examples/button-reduce-presence.tsx
Kiarokh: Sure

src/decluttering.md
adrianschmidt: Decluttring is a typo. It should be decluttering. But I'm absolutely fine with rephrasing to avoid the word 🙂

adrianschmidt: Pushed a fix with the heading "Reducing UI clutter and cognitive load".

hannahu (approved):

adrianschmidt:
I squashed the fixups, which dismissed the approval, but the code is the same, so I'll merge.

PR #1029, opened 2020-11-09
fix(table): only escape html in table for custom components once

maltja:
html for custom components were escaped twice, therefore, the escape characters were shown.

fix: Lundalogik/crm-feature#1588

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


adrianschmidt:
I'm trying to review this, but can't reproduce it, because my relation columns just show the `id` 🙁

image

image

@jensgustafsson Any ideas?

src/components/table/columns.ts
```diff
@@ -99,7 +99,7 @@ export function formatCell(
         value = column.formatter(value, data);
     }
 
-    if (typeof value === 'string') {
+    if (typeof value === 'string' && !column.component) {
         value = escape(value);
```
adrianschmidt, line 103: It looks like we are simply not escaping strings that are rendered by a custom component, but leave escaping up to the component?

If that's the case, I think the commit should be called `fix(table): do not escape html for custom components` (don't forget to edit the commit message body too of course).

adrianschmidt: Or are we escaping somewhere else too?

jgroth: I don't think we are escaping anywhere else so this is basically turning it off :scream:

adrianschmidt: Yeah, I had that scream-feeling too for a little while yesterday, but then checked the belongsto-component, and for that one it's "automatically" escaped because we of how we inject the string inside a link in jsx.

The risky thing with this is that it leaves escaping or sanitising completely up to the custom component. Maybe we could keep the escaping here, and any component that needs non-escaped html will have to un-escape it. That way we make it an active choice.

If that's possible?

Does `html-escaper` have an `unescape` function? @maltja

maltja: yes they have an unescape function 🙂  @jgroth @adrianschmidt  Should we unescape it in all our own custom components that already escapes the HTML string instead? Thanks for looking at it so I don't introduce some "risky" code 😅 .

adrianschmidt: Yeah, I think that would be the more prudent way to go about it. What do you say @jgroth?

If we've got any documentation about custom components, we can mention there that any string values are escaped and how to unescape it again, when necessary.

jgroth: Yeah I'm thinking about what the best solution to this whole problem is :thinking: Who should be responsible to escape all the strings? It should probably be at a level so that nobody else has to think about it, and if we add it here we might only fix it for the table :thinking: 
Maybe this should be solved in the formatters in lime-crm-components? Those are meant to format values before being displayed which seems very related to escaping strings. Thoughts?

adrianschmidt: Where in the chain are those formatters used? Are they end-of-the-chain html elements? Or are they affecting the values closer to the db, before the values are passed on to any custom components?

adrianschmidt: This might be interesting for this conversation… we don't escape our search results:

image

Compare:

image

jgroth: The formatters are used by the views so they run before something might have been sent to a custom component. We might event want to do the escaping before that, I don't think we should burden our appers and force them to have to think about stuff like that, if they make a custom view/component and get stuff from the databse, it should probably be escaped already :thinking:

adrianschmidt: Well, then I'm all for it! 👍

I guess we should still document how to unescape, for situations when that's necessary?

adrianschmidt: So, to recap:

1. If possible, we would like to move escaping from here to the formatters, "further up" in the chain (closer to the DB).
2. If unescaping is required, it should be done by the "end component", the one that actually renders stuff.
3. The change made in this PR won't be required.

Doing just bullet 2 will fix Lundalogik/crm-feature#1588, regardless of whether we do bullet 1 now, later, or never.

Agree or disagree? 🙂

@maltja @jgroth

adrianschmidt: PR for bullet 2 here: Lundalogik/lime-crm-components#217

adrianschmidt:
Replaced by Lundalogik/lime-crm-components#217

src/components/table/columns.ts
jgroth: I spent some time now debugging this to see what's going on and what we need to do in order to fix this. I think the short answer is to re-open this PR and close the other one.

The long answer...
I wondered above who should be responsible for escaping all the values, and the answer to that is the one that is rendering that value. In this case Stencil is the one responsible for rendering so Stencil should make sure that all values are safe. If that was not the case we would probably be in much worse state since we are pretty much not escaping values _anywhere_ in our components. I stepped through the code and found that Stencil does something like this
```ts
elm = newVNode.$elm$ = doc.createTextNode(newVNode.$text$);
```
i.e. Stencil is creating text nodes for everything that should be rendered as text which will make it impossible to inject any scripts into it.

So why was this a problem in the table? Why did we have to escape the values here in the first place? The answer is that Stencil is not rendering stuff inside the table, Tabulator is doing that. Stencil is just responsible for rendering the container for the table and everything after that is up to Tabulator.

When we render stuff in the table we use a custom Tabulator formatter for pretty much every value, and from the Tabulator documentation we can read
> The function must return the contents of the cell, either the text value of the cell, valid HTML or a DOM node.

And since HTML is a valid return type from these formatters we need to escape any data that we want to return or we will be vulnerable to XSS injection. The exception to this however are the custom components, since these are written in Stencil we are again entering Stencil-land where everything is taken care of for us.

Sorry for not realizing this sooner and creating extra work for you @adrianschmidt 
Security is hard and it's always worth taking some extra time to understand how things work. Hopefully we all learned something here, at least I did :slightly_smiling_face:

adrianschmidt: No worries @jgroth! Thanks for spending the time to dig into it! ❤️

adrianschmidt (approved):

PR #1031, opened 2020-11-10
fix(table): show no pagination when only one page exists

maltja:
fix: Lundalogik/crm-feature#1551

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/partial-styles/tabulator-paginator.scss
+++ b/src/components/table/partial-styles/tabulator-paginator.scss
@@ -1,50 +1,90 @@
-.tabulator .tabulator-footer {
-    .tabulator-paginator {
-        display: grid;
-        grid-auto-flow: column;
-        justify-content: end;
+.tabulator {
+    .tabulator-footer {
+        transition: transform 0.5s ease-out, opacity 0.35s ease; //For some reason the footer is not animated when it gets hidden/removed
 
-        .tabulator-pages {
-            margin: 0 pxToRem(8);
+        .tabulator-paginator {
+            display: grid;
+            grid-auto-flow: column;
+            justify-content: end;
 
-            .tabulator-page {
-                margin: 0 pxToRem(0);
-            }
-        }
+            .tabulator-pages {
+                margin: 0 pxToRem(8);
 
-        .tabulator-page {
-            &:not([disabled]) {
-                @include is-flat-clickable;
-            }
-            position: relative;
-            height: pxToRem(28);
-            min-width: pxToRem(28);
-            padding: 0 pxToRem(8);
-            margin: 0;
-
-            border: none;
-            border-radius: pxToRem(40);
-
-            color: #303042;
-            background-color: transparent;
-
-            &:hover {
-                color: #303042;
-                background-color: #fafafb;
-            }
-
-            &.active {
-                box-shadow: var(--button-shadow-inset);
-                background-color: #fff;
-            }
-
-            &:disabled {
-                cursor: not-allowed;
-
-                &:hover {
-                    background-color: transparent;
+                .tabulator-page {
+                    margin: 0 pxToRem(0);
                 }
             }
+
+            .tabulator-page {
+                &:not([disabled]) {
+                    @include is-flat-clickable;
+                }
+                position: relative;
+                height: pxToRem(28);
+                min-width: pxToRem(28);
+                padding: 0 pxToRem(8);
+                margin: 0;
+
+                border: none;
+                border-radius: pxToRem(40);
+
+                color: #303042;
+                background-color: transparent;
+
+                &:hover {
+                    color: #303042;
+                    background-color: #fafafb;
+                }
+
+                &.active {
+                    box-shadow: var(--button-shadow-inset);
+                    background-color: #fff;
+                }
+
+                &:disabled {
+                    cursor: not-allowed;
+
+                    &:hover {
+                        background-color: transparent;
+                    }
+                }
+            }
+        }
+    }
+}
+
+.tabulator:not(.has-pagination) {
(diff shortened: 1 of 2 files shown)


src/components/table/partial-styles/tabulator-paginator.scss
```diff
@@ -49,6 +51,18 @@
     }
 }
 
+.has-no-pagination {
+    .tabulator-tableHolder {
+        height: 100% !important;
+        max-height: 100% !important;
```
Befkadu1, line 57: Even if it is not recommended to use `!important`. I can see that I don't see other option to do it in other way

src/components/table/table.tsx
```diff
@@ -385,9 +385,12 @@ export class Table {
     }
 
     render() {
+        const pagination =
+            this.totalRows <= this.pageSize ? 'has-no-pagination' : null;
+
```
Befkadu1, line 390: If I would do like this
```
        const pagination = {
            'has-no-pagination': this.totalRows <= this.pageSize,
        };
```

Befkadu1:
I am not sure if this comment is related with this PR but I would like to mention it here. When we remove the pagination, it becomes clear that we need to fix the margin/padding by 0.5rem at the bottom of the table.
@Kiarokh should we need to create an issue for this?

Screenshot 2020-11-11 at 13 34 46

maltja:
@Befkadu1 I think it can be included in this PR. I will add some padding/margin. 🙂

src/components/table/table.tsx
```diff
@@ -385,9 +385,12 @@ export class Table {
...
         return (
             <div id="tabulator-container">
-                <div id="tabulator-table" />
+                <div id="tabulator-table" class={pagination} />
```
adrianschmidt, line 393: If you use Bef's suggestion for `const pagination` above, I think you need this too:

```suggestion
                <div id="tabulator-table" class={{pagination}} />
```

maltja:
I'm not sure that we actually need padding/margin. If you look at the details and history view we have no padding there. What do you think? @Kiarokh @Befkadu1

src/components/table/partial-styles/tabulator-paginator.scss
Kiarokh: Yes. It's such an isolated case. So it won't affect anything else.

Kiarokh:
We don't need extra padding or margin at the bottom. 

And if we would need for some reason, it should be done in the crm-components repo in table-view component. ;)

src/components/table/partial-styles/tabulator-paginator.scss
```diff
@@ -49,6 +51,18 @@
...
+    }
+    .tabulator-footer {
+        visibility: hidden; //Display:none does not work here because of the tabulators dynamic calculations, which requires manual resizing.
+        opacity: 0;
+        transform: translateY(pxToRem(40));
+    }
```
Befkadu1, line 63: We can not remove the `tabulator-footer` class because for certain limetypes for example `deal` we have to show the `calculation holder` (I am not sure of the name, see the pic for clarification ). The class `.tabulator-calcs-holder` in side `.tabulator-footer` should not be hidden

Screenshot 2020-11-12 at 10 16 44

Befkadu1 (changes requested):

src/components/table/partial-styles/tabulator-paginator.scss
Befkadu1: I don't think 
```
height: 100% !important;
max-height: 100% !important;
```
This values are correct when we have `the calculation holder tab` at the bottom. See the attached picture with the calculation holder tab` in the other comment

src/components/table/partial-styles/tabulator-paginator.scss
Kiarokh: @maltja we can instead hide the pagination div (inside the footer), instead of hiding the entire footer div. However, the problem again is that Tabulator calculates the height of the table and inlines it differently when there is **a)** a footer only with pagination **b)** a footer with both pagination and calculations row

To fix this with CSS, we need to know when the footer contains calculations and add an extra class to the tabulator container, like what you are doing with `has-no-pagination`. You could add something like `has-calculations`.

Then we can address it with CSS

Kiarokh: Also, I prefer positive naming more :) instead of `has-no-xyz` I'd prefer `has-xyz` :) But it's a minor thing.

maltja: Yes I also think we can solve it like that, I will take a look at it after the lime-project env meeting 🙂

src/components/table/table.tsx
```diff
@@ -387,8 +387,18 @@ export class Table {
...
+                        'has-pagination': this.totalRows > this.pageSize,
+                        'has-calculations': hasCalculations(this.columns),
+                    }}
+                />
             </div>
         );
     }
 }
+
+function hasCalculations(columns: Column[]): boolean {
```
Befkadu1, line 402: Shouldn't we write this function in the `Table`  class? And what do you think if we make it private function?
The other thing is I may be wrong but what this function does is to check if at least one of the columns has `aggregator`. What do you think if we name it `hasAggregator()`

Befkadu1 (approved):

PR #1032, opened 2020-11-10
perf(dev-server): create link to node_modules instead of copying 5000+ files

jgroth:
Just copying the icons takes 30+ seconds on my machine when starting the dev-server. This PR creates a symlink to node_modules instead to speed things up a bit

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,7 @@
     "eslint-plugin-sonarjs": "^0.5.0",
     "jest": "^26.5.2",
     "jest-cli": "^26.5.2",
+    "lnk": "^1.1.0",
     "lodash-es": "4.17.15",
     "prettier": "^2.1.2",
     "prismjs": "1.21.0",
--- a/stencil.config.docs.ts
+++ b/stencil.config.docs.ts
@@ -23,12 +23,12 @@ export const config: Config = {
             copy: [
                 { src: 'style/color-palette-extended.css' },
                 {
-                    src: '../node_modules/@lundalogik/lime-icons8/assets/',
-                    dest: 'assets/',
+                    src: '../node_modules/@lundalogik/lime-icons8',
+                    dest: 'node_modules/@lundalogik/lime-icons8',
                 },
                 {
-                    src: '../node_modules/kompendium/dist/',
-                    dest: 'assets/kompendium/',
+                    src: '../node_modules/kompendium/',
+                    dest: 'node_modules/kompendium/',
                 },
             ],
         },
--- a/publish-docs.js
+++ b/publish-docs.js
@@ -132,6 +132,7 @@ function build() {
                 /="\/build/g,
                 /="\/style/g,
                 /="\/assets/g,
+                /="\/node_modules/g,
                 /\/kompendium.json/g,
             ],
             to: [
@@ -139,6 +140,7 @@ function build() {
                 `="/versions/${version}/build`,
                 `="/versions/${version}/style`,
                 `="/versions/${version}/assets`,
+                `="/versions/${version}/node_modules`,
                 `/versions/${version}/kompendium.json`,
             ],
         };
--- a/rollup-plugin-link-node-modules.js
+++ b/rollup-plugin-link-node-modules.js
@@ -1,0 +1,20 @@
+/* eslint-disable no-undef */
+const lnk = require('lnk');
+const fs = require('fs');
+
+const TIMEOUT = 500;
+
+module.exports = function linkNodeModules() {
+    return {
+        name: 'link-node-modules',
+        generateBundle: async () => {
+            while (!fs.existsSync('www')) {
+                await new Promise((resolve) => setTimeout(resolve, TIMEOUT));
+            }
+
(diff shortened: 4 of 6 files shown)


stencil.config.docs.ts
```diff
@@ -23,12 +23,12 @@ export const config: Config = {
             copy: [
                 { src: 'style/color-palette-extended.css' },
                 {
-                    src: '../node_modules/@lundalogik/lime-icons8/assets/',
-                    dest: 'assets/',
+                    src: '../node_modules/@lundalogik/lime-icons8',
+                    dest: 'node_modules/@lundalogik/lime-icons8',
```
adrianschmidt, line 27: I guess we're doing this linking for "production" builds of the docs too? If so, it will probably not work on AWS. But we want to move back to GitHub Pages anyway, now that we have Kompendium. So I guess we'll have to do that at the same time…

package.json
```diff
@@ -83,6 +83,7 @@
     "eslint-plugin-sonarjs": "^0.5.0",
```
adrianschmidt, line 83: Note: I had to delete my build output folders (`.kompendium`, `dist`, `www`) to get this to work. Otherwise it got stuck on something, and after I canceled the build, the node_modules would be missing, so I had to run `npm install` again 😄

stencil.config.docs.ts
jgroth: We are still copying all the files for the docs build, it's just copied to another location now so that we have the same structure as the dev build

package.json
jgroth: Hmm ok :thinking:

adrianschmidt, line 83: I don't actually get much performance benefit from this, because the copying seems to run in parallel with "generate lazy", which still takes the same amount of time:

On master:
```
[46:11.6]  copy started ...
[46:11.6]  generate lazy started ...
[46:33.9]  copy finished (5138 files) in 22.33 s
[46:46.4]  generate lazy finished in 34.83 s
```

On this branch:
```
[06:53.5]  copy started ...
[06:53.5]  generate lazy started ...
[06:53.9]  copy finished (13 files) in 436 ms
[07:26.3]  generate lazy finished in 32.80 s
```

adrianschmidt: That ☝️ in itself wouldn't be a problem, since this still brings a benefit on Linux, but:

After trying around a bit more, it seems I have to run npm install every time I stopped the dev-server with ctrl-c, otherwise the node-modules are missing, and I also have to remove the build output folders before running `npm start` again, because otherwise the build process seems to just end up in some endless loop, spinning up my computer fan to maximum… 😢

stencil.config.docs.ts
adrianschmidt: Ah, ok 👍

package.json
jgroth: :dizzy_face: omg that seems like a real hassle haha :smile: 
I can put this back in draft, it seems like getting the tests to work was a real hassle as well and don't want to spend several more hours on it right now... I have tried getting the mocking to work, but it seems to be really hard to get right when running in puppeteer :thinking:  
Even something like this did not seem to work :disappointed: 
```ts
    await page.addScriptTag({
        content: `
            window.fetch = () => {
                return Promise.resolve({
                    text: () => Promise.resolve('${emptySvg}')
                });
            };
        `,
    });
```
It was worth a try at least! But if it is only marginally faster there is no point in doing this. I just looked at the `copy finished (5138 files) in 22.33 s` when testing this, I should have clocked it as well :smile:

adrianschmidt: 😞

But yeah, keep it as a draft. It would be really nice if we can improve performance a bit. It would also be really nice to have the different versions of the published docs use the same copy of limel-icons8, instead of having one copy for each version, and fixing that seems like it might be similar to fixing this?

Btw, you are linking the whole node_modules folder in this PR. Have you tried only linking the folder where the icons reside?

jgroth: Yeah I linked the whole node_modules folder since that was the fastest and easiest way :upside_down_face: It would not be that hard to make the rollup plugin take a config of what to create links to instead

anderssonjohan:
@jgroth @adrianschmidt In my local dev in another package I don't use the www folder at all. That makes the dev server serve from repo root and I include the icons and lime elements from `/node_modules/... `. No copying needed for local dev! (I have dist but no www)

jgroth:
I think it might also be possible to solve this with the new custom request listener introduced in Stencil 2.5.0 :slightly_smiling_face: 
https://github.com/ionic-team/stencil/blob/master/CHANGELOG.md#-250-2021-03-22

adrianschmidt:
We no longer depend on lime-icons8, so I think we can close this?

PR #1034, opened 2020-11-11
Add optional form layouts

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/1527

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/form/templates/grid-layout.ts
```diff
@@ -0,0 +1,75 @@
...
+            this.setState({
+                maxColumns: maxColumns,
+            });
+        }
+    }
+
+    public render() {
+        const layout: FormLayoutOptions = this.props.options;
+        const classes = ['limel-form-layout--grid'];
+        const columns = Math.min(layout.columns || 1, this.state.maxColumns);
```
jgroth, line 53: Default should be 5

src/components/form/templates/grid-layout.ts
```diff
@@ -0,0 +1,75 @@
...
+
+        classes.push(`layout-${columns}-columns`);
```
jgroth, line 55: There seems to be some bug with this as well :thinking:

src/components/form/templates/grid-layout.ts
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
+import { FormLayoutOptions, GridLayoutOptions } from '../form.types';
+
+const MAX_COLUMNS = 5;
+const MAX_COLUMN_WIDTH = 15;
+const REM_TO_PX = 16;
+const COLUMN_COUNT = '--number-of-columns';
+
+export class GridLayout extends React.Component<LayoutProps, LayoutState> {
```
jgroth, line 9: Right now this layout just works in the form, but we could probably quite easily move this implementation to the `limel-grid` component instead and use that one in here. The reason for not doing it right now was that I did not realize that until later and it might also introduce breaking changes in the existing component. We can always do it later if we decide not to do it now :slightly_smiling_face:

Kiarokh: I don't really follow. You mean we should move these new layout possibilities/options (spaning and dense layout) to `limel-grid` instead?

jgroth: I'm just saying we could, there's not much difference between using them outside or inside of a form :slightly_smiling_face:

Kiarokh: That's right. Specially because that way of `grid-area` template thingy is quite hard to work with, isn't responsive, and not so sustainable. 

But how can can we transition them to use this new way? It's a breaking change and people won't be so happy if we change it I guess?

Maybe we can deprecate that and make a new limel-grid component with all these features? 

It'd also be nice to use it on the CRM dashboard

Kiarokh:
By the way. I'm not sure what will happen to fields such as chipset, which can grow in height. We will probably need to make tweaks in them (?). 🤔

adrianschmidt:
> By the way. I'm not sure what will happen to fields such as chipset, which can grow in height. We will probably need to make tweaks in them (?). 🤔

Very good point!

adrianschmidt (commented):
Just a partial review. This is how far I got yesterday 🙂

src/components/form/templates/grid-layout.ts
adrianschmidt: (In other words, I think we should keep this separate from limel-grid for now 🙂)

jgroth:
> > By the way. I'm not sure what will happen to fields such as chipset, which can grow in height. We will probably need to make tweaks in them (?). thinking
> 
> Very good point!

This is not limited to chip sets. As soon as a custom component is used (which a chip-set will be if it's used), anything can happen to the layout since the components can grow/shrink however they see fit

src/components/form/examples/form-layout.tsx
```diff
@@ -4,6 +4,49 @@ import { schema } from './schema-layout';
...
+ * columns will also decrease.
+ *
+ * :::tip
+ * You can divide a form into collapsible or none-collapsible sections,
+ * and for each section specify a different layout.
+ * :::
+ *
+ * In this example, each collapsible section has its on `colSpan`.
+ * However, since the layout is responsive, make sure to change the browser
+ * window size to see how their responsive layout changes.
```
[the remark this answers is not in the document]
jgroth: No not at the moment, not sure how easy that will be to fix, it's probably mostly a CSS issue so @Kiarokh might have a better answer :slightly_smiling_face: 
As a workaround, while not perfect, you could always put the form in a full screen dialog

adrianschmidt:
> > > By the way. I'm not sure what will happen to fields such as chipset, which can grow in height. We will probably need to make tweaks in them (?). thinking
> > 
> > 
> > Very good point!
> 
> This is not limited to chip sets. As soon as a custom component is used (which a chip-set will be if it's used), anything can happen to the layout since the components can grow/shrink however they see fit

I tried changing the size of the textarea in the example for "stretching fields", and while I could change the inner textarea, the outer fram stayed the same size… Not sure if we need to incorporate row-height change from https://github.com/Lundalogik/lime-webclient/pull/1590?

src/components/form/examples/form-layout.tsx
```diff
@@ -4,6 +4,49 @@ import { schema } from './schema-layout';
...
+ * items per row is not possible, the form will automatically change the layout
+ * and fit in 3 items per row. As the container's width decreases, number of
+ * columns will also decrease.
+ *
+ * :::tip
+ * You can divide a form into collapsible or none-collapsible sections,
+ * and for each section specify a different layout.
+ * :::
+ *
+ * In this example, each collapsible section has its on `colSpan`.
```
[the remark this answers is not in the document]
Kiarokh: Thanks @adrianschmidt . I'll do a fixup.
********
I can just say i **DO HATE 🤬** Mackbooks' new freaking [butterfly keyboard](https://www.thenextweb.com/plugged/2019/07/07/what-hell-apple-butterfly-keyboard-keys-design/). It's disgusting. Some letters randomly get typed twice. And some never get typed. Now add this to mmy (⬅ look, as I'm typing) dyslexia. And it feels you're typing on some disgusting low quality toy-keyboard and I can't stand the sound of it neither. It's like someone is chewing with open mouth. 😭 Blä🤮. 

Seems many people have this problem when you google it.

src/components/form/examples/form-layout.tsx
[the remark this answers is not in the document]
Kiarokh: Good point! Can we load the examples in **fullscreen dialogues** instead?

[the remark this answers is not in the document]
Kiarokh: Ah just saw that @jgroth have also written that 😂

src/components/form/examples/form-layout.tsx
```diff
@@ -4,6 +4,49 @@ import { schema } from './schema-layout';
 
 /**
  * Layout
+ * By default, each item in a limel-form will be rendered in a single row; and
+ * each row occupies the entire available width of the form's container.
+ *
+ * This default layout may work fine on small screens or narrow containers,
+ * but on larger screens it won't usually produce a nice layout. Thus we
+ * recommend that you choose an appropriate responsive layout for your form.
```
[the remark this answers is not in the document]
Kiarokh: Oh good catch. 😂 This means when they don't specify `layout: grid`. Then it will be single rows which are 100% wide (1 column). 

But if the specify `layout: grid`, then we default on 5 columns. We can for sure improve how we explain it here.

src/components/form/examples/form-layout.tsx
```diff
@@ -4,6 +4,49 @@ import { schema } from './schema-layout';
 
 /**
  * Layout
+ * By default, each item in a limel-form will be rendered in a single row; and
+ * each row occupies the entire available width of the form's container.
```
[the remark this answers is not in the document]
Kiarokh: nice 👌 tnx

src/components/form/examples/form-layout.tsx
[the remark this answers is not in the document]
adrianschmidt: Yeah, I have the same shitty keyboard on mine. I have to take the keys off every now and then to fix stuck keys… (if you do, it's important to open them on the top side, as there are hooks on the bottom side that would break if you do it wrong) https://apple.stackexchange.com/questions/272038/macbook-pro-2016-keyboard-key-stuck-how-to-remove-key

src/components/form/examples/form-layout.tsx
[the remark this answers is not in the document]
adrianschmidt: It might be that I was just confused by the commit called `feat(object-field): give limel-form a css-grid layout by default`. But I think that refers to the css-property `grid`, which isn't the same thing as what we mean by `layout: grid` here, right?

I'll read it again with those glasses on, and see if it makes sense then 👍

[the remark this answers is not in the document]
Kiarokh: Totally right. Both of them have `display: grid;` in their CSS. (because we wanted to use some advantages of grid, like the `grid-gap` even for the one columned layout.

Kiarokh:
> Not sure if we need to incorporate row-height change from Lundalogik/lime-webclient#1590?

We can look at it. I did a quick test with `grid-auto-rows: minmax(auto, var()--min-height-of-one-row);` and then changed the height of the example. But it didn't help. I'll have to look at it when my head is more fresh

src/components/form/examples/form-span-fields.tsx
```diff
@@ -0,0 +1,215 @@
...
+ *         },
+ *     },
+ * };
+ * ```
+ *
+ * :::note
+ * Sometimes order of fields are important in the way users perceive the form.
+ * If you choose to have the default auto-reordering behavior of the forms,
+ * make sure to test yout forms layout in different screen sizes to see whether
+ * you can mitigate unwanted layout changes.
```
adrianschmidt, line 81:
```suggestion
 * Sometimes, the order of fields are important for the way users perceive the form.
 * If you choose to use the default auto-reordering behavior, make sure to test your
 * form's layout in different screen sizes to see whether you can mitigate unwanted
 * layout changes.
```

adrianschmidt: I think this ☝️ is a very good reason for not having `dense` default to `true`. How many users of limel-form is actually going to do that?

src/components/form/examples/form-span-fields.tsx
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * You can avoid unwanted results by changing order of fields so that they render
+ * appropriately on different screens, or by dividing the form into more
+ * sections.
```
adrianschmidt, line 85:
```suggestion
 * Some unwanted results can be avoided by changing the order of the fields,
 * so that they render appropriately on different screens, or by dividing
 * the form into more sections.
```

src/components/form/examples/form-span-fields.tsx
```diff
@@ -0,0 +1,215 @@
...
+ * ###### Stretching a field vertically
+ * All standard elements that can be used in forms such as `limel-input`,
+ * `limel-select`, `limel-slider`, etc. have a fixed UI in the horizontal axis
+ * and therefore it does not really make sense to stretch them in height,
+ * and we strongly recommend you not to!
+ *
+ * But there are some exceptions. One of them is the `textarea`.
+ * It's always recommended to assign at least two rows for a textarea, to improve
+ * its usability. Since this is a component that is supposed to show multi-lined
+ * pieces of text.
```
adrianschmidt, line 97:
```suggestion
 * Most standard elements that can be used in forms, such as `limel-input`,
 * `limel-select`, `limel-slider`, etc, have a fixed height, and therefore
 * it does not really make sense to stretch them vertically, and we strongly
 * recommend you not to!
 *
 * But there are some exceptions. One of them is `limel-input-field` with `type='textarea'`.
 * It's always recommended to assign at least two rows for a textarea, to improve
 * its usability. Since this is a component that is supposed to show multi-lined
 * pieces of text.
```

src/components/form/examples/form-span-fields.tsx
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * Also, if you make create a custom component for your form –let's say a map–,
+ * you will have the possibility of using `rowSpan`, to increase the height of
+ * your custom component.
```
adrianschmidt, line 101:
```suggestion
 * Also, if you make create a custom component for your form—let's say
 * a map—you will have the possibility of using `rowSpan` to increase the height of
 * your custom component.
```

adrianschmidt: It's not visible in the code, but those "hyphens" are this kind: —

I find it ugly myself, but the "correct" way of using them is actually _without_ spaces… and they shouldn't be used with commas, as they act as a kind of "super-comma".

src/components/form/examples/form-span-fields.tsx
```diff
@@ -0,0 +1,215 @@
...
+ * ````ts
```
adrianschmidt, line 102:
```suggestion
 * ```ts
```

src/components/form/examples/form-span-fields.tsx
```diff
@@ -0,0 +1,215 @@
...
+ *                 },
+ *                 layout: {
+ *                       colSpan: 3,
+ *                       rowSpan: 2,
+ *                 },
+ *             },
+ *         },
+ *     },
+ * };
+ * ````
```
adrianschmidt, line 123:
```suggestion
 * ```
```

src/components/form/examples/form-span-fields.tsx
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * :::warning
+ * Custom web-components that you include in the forms should not have hard-coded
+ * `width` or `height` values! Otherwise they will stretch out of their cell and break
+ * the UI. Make sure that such components are internally designed to be responsive,
+ * and that their `:host` and their eventual wrapping container has the following
+ * styles:
```
adrianschmidt, line 130:
```suggestion
 * Custom web-components that you include in the forms should not have hard-coded
 * `width` or `height` values! Otherwise they will stretch out of their cell and break
 * the UI. Make sure that such components are internally designed to be responsive,
 * and that their `:host` and any potential wrapping container has the following styles:
```

In English, "eventual" means something that is certain to happen, but at some uncertain time in the future.

src/components/form/examples/form-span-fields.tsx
```diff
@@ -0,0 +1,215 @@
...
+ * ````scss
```
adrianschmidt, line 131:
```suggestion
 * ```scss
```

src/components/form/examples/form-span-fields.tsx
```diff
@@ -0,0 +1,215 @@
...
+ * :host {
+ *   display: block; // or another suitable property
+ *   width: 100%;
+ *   height: 100%;
+ * }
+ * .my-component {
+ *   width: 100%;
+ *   height: 100%;
+ * }
+ * ````
```
adrianschmidt, line 141:
```suggestion
 * ```
```

src/components/form/examples/form-span-fields.tsx
```diff
@@ -0,0 +1,215 @@
...
+ * ````scss
+ * :host {
+ *   display: block; // or another suitable property
+ *   width: 100%;
+ *   height: 100%;
+ * }
+ * .my-component {
+ *   width: 100%;
+ *   height: 100%;
+ * }
```
adrianschmidt, line 140:
```suggestion
 * :host {
 *     display: block; // or another suitable property
 *     width: 100%;
 *     height: 100%;
 * }
 * :host([hidden]) {
 *     display: none;
 * }
 * .my-component {
 *     width: 100%;
 *     height: 100%;
 * }
```

adrianschmidt (commented):
Again, this is today's "yield". To be continued… 😄

src/components/form/examples/form-span-fields.tsx
```diff
@@ -0,0 +1,215 @@
...
+ * layout in 3 columns instead. Therefore, the second field has to jump
+ * to the next line to still be able to stretch through 3 columns.
+ * This will leave 2 empty cells in the first row, right after the first field.
+ *
+ * To avoid having these empty cells in the UI, limel-form will automatically
+ * bring up two other fields that span only 1 column, and places them in the
+ * available empty cells in the first row.
+ *
+ * However, you can disable this functionality by setting `dense` to fall in the
+ * options for the grid layout
```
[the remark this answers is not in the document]
Kiarokh: I don't think it's a big deal if fields switch place. (This might be just an uninformed assumption though.) We can wait to see if people complain, then we can default to the other option. It's not like the form will pull up the last filed in a section and display it first. The algorithms will pick the next best choice. So its location will be naturally close and still make sense, or play nicely with users' memory about the location of field. (if they have any memory of it).

Anyhow, I believe when forms are made in a way that they have to reorder (which means someone intentionally and deliberately did design them that way and used col/row-spans) if reordering form fields mess up users' heads, it's the designer's responsibility and fault.

We have given them both options, and they have to think about how they are using it.

There are so many of these scenarios when we design generic components. We can't mitigate every possible mistake that consumers of them components can make.

src/components/form/examples/form-span-fields.tsx
Kiarokh: so there should be a `space` after them, not before them?

Kiarokh: What I have learnt is that: when you put something between these `–` characters, it means the meaning with or without "that thing which is between the `–`s" will have the same message/meaning. In other words, you can drop that part without making your sentence meaningless (it's like an inline footnote). But if your original sentence has commas, then commas should still be where they are with or without the "inline footnotes". But I have no academic reference to support wha I'm saying. So feel free to change ;)

Kiarokh:
@adrianschmidt is there a way to add all your corrections at once as a fixup? I don't know how to do it. Can you please do it :) I think they are all great and have no comments about them.

src/components/form/examples/form-span-fields.tsx
adrianschmidt: No spaces around them at all (which is why I had to put the line-break between the words inside them).

More info here: https://www.dictionary.com/e/em-dash/

Reading the _When to use an em dash_ section on that page, it seems like it would be more correct to just use commas in the instance above… 🤔

adrianschmidt:
> @adrianschmidt is there a way to add all your corrections at once as a fixup? I don't know how to do it. Can you please do it :) I think they are all great and have no comments about them.

Go to the _Files changed_ tab, then click the _Add suggestion to batch_ button for each change you wish to commit together :+1:

src/components/form/examples/form-span-fields.tsx
adrianschmidt: > What I have learnt is that: when you put something between these `–` characters, it means the meaning with or without "that thing which is between the `–`s" will have the same message/meaning. In other words, you can drop that part without making your sentence meaningless (it's like an inline footnote). But if your original sentence has commas, then commas should still be where they are with or without the "inline footnotes". But I have no academic reference to support wha I'm saying. So feel free to change ;)

My only knowledge of how to use em dashes is from reading English language newspaper articles. Newspapers love using them, while seemingly no one else does 😄

And yes, I think I've mostly seen them used as you describe, infixing a kind of footnote. Which is why I was a little surprised to not see that use listed on dictionary.com. I say keep the em dashes! 👍

Kiarokh:
oh… hallelujah 🎉😇

src/components/form/templates/grid-layout.ts
```diff
@@ -0,0 +1,92 @@
...
+        const element = this.elementRef.current;
+        const width = element.getBoundingClientRect().width;
+        const maxColumns = Math.min(
+            MAX_COLUMNS,
+            Math.max(1, Math.floor(width / (MAX_COLUMN_WIDTH * REM_TO_PX)))
+        );
+
+        const columns = this.getColumnCount(maxColumns);
+        if (element.style.getPropertyValue(COLUMN_COUNT) !== String(columns)) {
+            element.style.setProperty(COLUMN_COUNT, String(columns));
```
adrianschmidt, line 43:
```suggestion
        if (element.style.getPropertyValue(COLUMN_COUNT) !== `${columns}`) {
            element.style.setProperty(COLUMN_COUNT, `${columns}`);
```

I'm not sure it really makes any difference, but the advice at least used to be to avoid the object constructors for types that has literals…

Either way, I kind of think this looks nicer anyway 🙂

adrianschmidt: Errrr… wtf happened to that suggestion???

In case it renders better when reloaded, here's a screenshot:

image

# 🧐 😂

This is what I _actually_ meant:

image

adrianschmidt: Update: yeah, it rendered ok after a page reload 😅

src/components/input-field/input-field.scss
```diff
@@ -20,6 +20,15 @@
     display: none;
 }
 
+:host([type='textarea']) {
+    height: inherit; // This makes it possible for textarea in limel-form to stretch & span through rows
```
adrianschmidt, line 24:
```suggestion
    height: inherit; // This makes it possible for textarea in limel-form to stretch & span rows
```

src/components/input-field/input-field.scss
```diff
@@ -20,6 +20,15 @@
...
+
+    .mdc-text-field.mdc-text-field--textarea {
+        height: var(--textarea-height, 100%);
+        resize: vertical;
+    }
```
adrianschmidt, line 29: This isn't working as expected when used in limel-form 🙁

See the following 2 gifs.

Multi-column:

textarea in limel-form 1

Single-column:

textarea in limel-form 2

Kiarokh:
By the way, this is how Fullscreen modals open in Lime Elements. They will be below the navigation.
image

src/components/form/examples/form-layout.tsx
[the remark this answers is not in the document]
adrianschmidt: I'd say let's postpone that for now. The docs tell the reader to try different window sizes anyway.

src/components/form/examples/form-span-fields.tsx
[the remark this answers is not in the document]
adrianschmidt: Yeah, good enough for me 👍 😄

adrianschmidt:
I'm happy with the code in the fixups, so I'll squash those and rebase on latest master.

src/components/input-field/input-field.scss
adrianschmidt: Bump on this.

@Kiarokh

Kiarokh:
@adrianschmidt awesome!

What about these two suggestions:

1. Using `max-height` and `height`, instead of only `height`
> I guess you mean adding both `max-height` and `height`? Because min-height should still let them grow. And we did actually use min-height in the beginning but that wouldn't let the textarea to expand to fill all the rows. That's why we used height instead.

2. Loading layout examples in fullscreen dialogues.
I tried this, but I was certainly doing something wrong and couldn't make it work. It should be easy for you guys to fix it think.

src/components/form/form.scss
```diff
@@ -10,12 +10,8 @@
     }
 }
 
-.limel-form-layout--grid {
+.limel-form-layout--default {
+    --gap: 1rem;
```
adrianschmidt, line 14: Does this mean the consumer can override this by setting their own value for `--gap`?

If so, should we document that?

src/components/form/form.scss
```diff
@@ -10,12 +10,8 @@
...
     display: grid;
-    grid-template-columns: repeat(auto-fill, minmax(18.75rem, 1fr));
-    gap: 0.5rem;
```
adrianschmidt, line 16: Except for the change from 0.5rem to 1rem gap, I think this commit is more of a refactor than a feature… 🤔

(And except for the possibility for the consumer to change `--gap`, if that's now possible.)

adrianschmidt:
> @adrianschmidt awesome!
> 
> What about these two suggestions:
> 
> 1. Using `max-height` and `height`, instead of only `height`
> 
> > I guess you mean adding both `max-height` and `height`? Because min-height should still let them grow. And we did actually use min-height in the beginning but that wouldn't let the textarea to expand to fill all the rows. That's why we used height instead.
> 
> 1. Loading layout examples in fullscreen dialogues.
>    I tried this, but I was certainly doing something wrong and couldn't make it work. It should be easy for you guys to fix it think.

It would be very helpful if you could write these replies in their respective threads, so we can keep the discussions threaded. Neither of them are resolved, so they aren't that hard to find 🙂

Kiarokh:
> This isn't working as expected when used in limel-form 🙁



@adrianschmidt great that you noticed this. I pushed new fixups now, fixing both the resize issue and that `textarea` grows larger than its container.

src/components/form/form.scss
Kiarokh: This wasn't my initial intention. the `var` was there only to be used in the `calc`s. I don't think we should expose this for external tweaks, at the moment at least.

src/components/form/form.scss
Kiarokh: I have no objections.

src/components/form/templates/field.ts
```diff
@@ -1,19 +1,47 @@
...
-    if (layout?.span) {
-        classNames += ` limel-form-layout-span--${layout.span}`;
+    if (colSpan) {
+        classNames += ` limel-form-layout-colspan--${colSpan}`;
+    }
+
+    if (rowSpan) {
+        style = {
+            gridRow: `span ${rowSpan}`,
+            height: `calc(var(--min-height-of-one-row) * ${rowSpan})`,
```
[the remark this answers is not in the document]
Kiarokh: Min height wouldn't help. We initially started with that. But we came to conclusion that we should write `height` for text area to work for example. I also tested having both `max-height` and `height`. But no difference that I could see. Seems `height` is enough for now. Or what do you think?

src/components/input-field/input-field.scss
```diff
@@ -22,6 +22,19 @@
...
+:host([type='textarea']) {
+    height: inherit; // This makes it possible for textarea in limel-form to stretch & span rows
+
+    .mdc-text-field.mdc-text-field--textarea {
+        height: var(--textarea-height, 100%);
+    }
+    textarea.mdc-text-field__input {
+        resize: vertical;
+        min-height: $height-of-mdc-text-field - 0.5rem;
+        max-height: 100%; // This will prevent the field to grow taller than its container, in grid forms for instance
```
adrianschmidt, line 34: Nice.

This makes the resizing work as expected for "stand-alone" textareas, while the resizing becomes somewhat useless for textareas in forms…

I think that's good enough for this to be mergeable. We'll still have to take a look at other possibly expanding components, like limel-picker. We don't have an example for any of these in limel-form though (only a custom-picker, that isn't using `multiple` and therefore would need some tweaking), and I'm a little too tired to take on adding one right now.

@jgroth The layout stuff obviously isn't used yet, so if there are some things that don't yet work with that, I don't think that's the end of the world. But the one thing I'm a little concerned about is if the change to using a single column grid layout for the default layout might mean that limel-forms using the default layout also have the same problem with components that changes height?

src/components/form/examples/form-span-fields.tsx
```diff
@@ -0,0 +1,218 @@
...
+ * }
+ * :host([hidden]) {
+ *     display: none;
+ * }
+ * .my-component {
+ *     width: 100%;
+ *     height: 100%;
+ * }
+ * ```
+ * :::
```
adrianschmidt, line 145: I think the documentation becomes waaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaay too long to navigate easily with all this text and these super long examples… 😞

Maybe we could put them in collapsibles or dialogs or something.

It wasn't a problem when I was reading the whole thing from top to bottom, but now that I was trying to actually find my way around to specific examples to look for one that had a picker or something, it was extremely frustrating.

src/components/input-field/input-field.scss
Kiarokh: Yes, we should add a multi-line picker to one of the examples and play around with it. It's a bit hard for me to add fields to example forms, unless I can steal it from another example :D Can you guys do it plz…?

I don't think there is a **css-only** solution for fields that change size on the go. Probably some script should dynamically apply inline styles (for `grid-row` and `height`) directly on the element, to make the grid-cell span in rows (if user drags it to expand, or if it automatically expands). Such scrip should overwrite what's already set by the consumer in the schema.

#### one more thing
MD offers something that we may at some point use too, called "Multi-line fields". So it'd be nice if we're prepared for it.
multi-line fields

Anyhow, I don't think that the "default layout" has any problems with fields that increase in height. That default layout pretty much only cares about the gaps, at the moment.

src/components/form/examples/basic-form.tsx
```diff
@@ -1,9 +1,9 @@
 import { ValidationStatus } from '@limetech/lime-elements';
 import { Component, h, State } from '@stencil/core';
-import { schema } from './schema';
+import { schema } from './basic-schema';
```
adrianschmidt, line 3: I renamed a bunch of these files, so it would be easier to know which belong together, and which example they provide.

src/components/form/form.tsx
```diff
@@ -27,10 +27,11 @@ import { isInteger } from './validators';
 /**
  * @exampleComponent limel-example-form
  * @exampleComponent limel-example-nested-form
- * @exampleComponent limel-example-form-layout
  * @exampleComponent limel-example-list-form
  * @exampleComponent limel-example-dynamic-form
  * @exampleComponent limel-example-custom-component-form
+ * @exampleComponent limel-example-form-layout
+ * @exampleComponent limel-example-form-span-fields
```
adrianschmidt, line 34: I moved these to the bottom, just to have them interfere less with the other examples, since these have such long descriptions.

src/components/form/examples/form-span-fields.tsx
Kiarokh: I think that's the downside, or better said the "trade-off" of having guidelines. I always think things are better to be explained well, rather than abstractly mentioned. Specially because we want this to be a stand-alone source of truth for consultants, who have different skill levels.

People can always choose not to read if they want to. But when they actually want to read, there should be content that clarifies things well.

The purpose of "guidelines" is this very thing; not just displaying examples I'd say. So we're satisfying the purpose. If it gets too long, it's not necessarily a problem. (Of course it's better if we can say the same things with less words).

However, when it comes to these forms examples, it feels they can be too long and too many. I don't know if we can merge them somehow. 

Also it'd be nice if Kompendium added more anchor-links to different sections of the documentation and made them available as floating navigation menu somehow, or added them in the nav menu.

adrianschmidt: I mostly agree actually 😄

But, we're mixing api-documentation, examples, and guidelines. It's getting hard to find your way around, and that makes it easy to miss important things.

Adding a way of navigating would improve things, for sure 👍

Also, if we want these longer descriptions in more places, we might want to move the api-docs up, above the examples again, where they were when we used **docz**…

Anyway, it's a work in progress. Sorry for letting my frustration shine through in my comment yesterday. I was tired and just felt "this is completely impossible to work with", but that's not true 😄

Kiarokh: Great! No worries dude. You absolutely have a point! Maybe we need two separate places for guidelines and actual API stuff. It's probably hard to do sometimes, as these are guidelines about options that consultants have when using the components (which overlap with our "design guidelines" sometimes).

Kiarokh:
Another thing I came to think of that we should test in the new form layouts is `limel-select`, which lets you choose multiple choices. The picture below is from another branch and this behaviour (expanding the column) may not happen in our layout branch. But we must test this.
image

src/components/form/examples/basic-form.tsx
```diff
@@ -1,11 +1,11 @@
 import { ValidationStatus } from '@limetech/lime-elements';
```
adrianschmidt, line 1: > > > By the way. I'm not sure what will happen to fields such as chipset, which can grow in height. We will probably need to make tweaks in them (?). thinking
> > 
> > 
> > Very good point!
> 
> This is not limited to chip sets. As soon as a custom component is used (which a chip-set will be if it's used), anything can happen to the layout since the components can grow/shrink however they see fit

I've tested this with pickers (I hacked the custom-component a little, see the two tmp-commits I've added), and the resizing doesn't create any issues with getting cut off or overflowing above or beneath other components. The height of the row the item is in is simply increased, so there are holes created around it (unless it's alone in its row of course), but that's not breaking any functionality.

I'm placing this comment on a line in the diff so we can thread any future comments, and I'll mark the previous comments quoted here as resolved, to make it easier to see which comments are still left to resolve.

adrianschmidt: I've now added a few textareas as well, and tested around a bit, and it appears that a component is free to grow so long as the rowSpan has not been defined for that component.

As soon as `rowSpan` is set, even if set to `1`, the component's height cannot be changed.

I'm pushing another tmp-commit with a few more textareas and another picker.

src/components/form/templates/field.ts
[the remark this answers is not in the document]
adrianschmidt: This is good enough for now, but I'm leaving this unresolved, as I might want to take a quick look at it later on. (And it would be near impossible to find among all other resolved comments 😅)

src/components/form/form.scss
adrianschmidt: It doesn't seem like it can be set from the outside anyway (not by just setting the value on the `limel-form` element at least). I guess we would have to define it on `:host` for that to be possible. So we're all good 👍

src/components/input-field/input-field.scss
adrianschmidt: I've fixed the resizing issue for textareas 🙂

src/components/form/examples/basic-form.tsx
Kiarokh: @adrianschmidt thanks. It sounds great!

This sounds like a good addition to the documentations:
> a component is free to grow so long as the rowSpan has not been defined for that component. As soon as `rowSpan` is set, even if set to `1`, the component's height cannot be changed.

adrianschmidt: Good point! 👍

src/components/form/examples/form-layout.tsx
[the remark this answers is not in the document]
Kiarokh: It's a great 💡! 

And it'd be nice to have an option in Kompendium to toggle an example to fullscreen and back its original size. (if that's what you also mean)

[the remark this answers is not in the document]
jgroth: Yeah I have thought about adding that for a while, a small button next to each example that will just display the example and nothing else. I think it would be great when you are developing or debugging something as there will be less noise on the screen and less stuff to load :slightly_smiling_face:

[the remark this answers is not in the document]
Kiarokh: it should be easy to do. Just give me a toggle-able class on the container. I'll do the rest. Also merge my last PR in Kompendium already 😄 !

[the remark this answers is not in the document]
jgroth: Haha yeah sorry, I will take a look at it. I've been less motivated to work on it now when I work from home all the time :disappointed:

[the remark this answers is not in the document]
adrianschmidt: I hope neither of you feel like you can only work on Kompendium on your spare time. We use it in our work at Lime, so changes motivated by what we do at Lime can of course be developed and contributed as part of our work at Lime 🙂

I want to start working a bit on Kompendium too, but haven't felt like I've had the time for it yet. I guess I'll just have to make time at some point 😄

src/components/form/templates/field.ts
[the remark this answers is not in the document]
adrianschmidt: ⚡

src/components/form/examples/basic-form.tsx
adrianschmidt: ⚡

Can you take a look @Kiarokh? (It's in the latest fixup at the time of writing.)

I added this note:

```diff
+ * :::note
+ * If you do *not* set the `rowSpan` for a component, it can stretch vertically
+ * within its row, and the row will simply expand with the component.
+ *
+ * If you *do* set a `rowSpan`, even if you set it to `1`, the component is
+ * fixed to that height. What happens to any potential overflow depends on the
+ * component.
+ * :::
```

But with that, I didn't think the passage about always assigning at least two columns to textareas made as much sense anymore, so I changed it like this:

```diff
  * it does not really make sense to stretch them vertically, and we strongly
  * recommend you not to!
  *
- * But there are some exceptions. One of them is `limel-input-field` with `type='textarea'`.
- * It's always recommended to assign at least two rows for a textarea, to improve
- * its usability. Since this is a component that is supposed to show multi-lined
- * pieces of text.
+ * But there are some exceptions. One of them is `limel-input-field` with
+ * `type='textarea'`.
  *
  * Also, if you create a custom component for your form—let's say a map—you
  * can use `rowSpan` to increase the height of your custom component.
```

src/components/form/form.scss
```diff
@@ -10,12 +10,112 @@
     }
 }
 
+.limel-form-layout--default {
+    --gap: 1rem;
+    display: grid;
```
adrianschmidt, line 15: Putting this here, so we can have a thread:

> (Kiarokh) Another thing I came to think of that we should test in the new form layouts is `limel-select`, which lets you choose multiple choices. The picture below is from another branch and this behaviour (expanding the column) may not happen in our layout branch. But we must test this.
> image

Confirmed. I just hard-coded the width of a textarea, and it squashed the other columns:

image

Hopefully there's something similar to `flex-grow: 0` we can use to stop that from happening… I'm looking at it now.

src/components/form/form.scss
```diff
@@ -10,12 +10,112 @@
...
+        grid-column-gap: calc(var(--gap) + 0.75rem);
+    }
+    &[style*='--number-of-columns:3;'],
+    &.layout-3-columns {
+        grid-column-gap: calc(var(--gap) + 0.5rem);
+    }
+    &[style*='--number-of-columns:2;'],
+    &.layout-2-columns {
+        grid-column-gap: calc(var(--gap) + 0.25rem);
+    }
```
adrianschmidt, line 46: Why is there a bigger gap when there are more columns? 🤔

Is it to make the columns easier to distinguish? I guess maybe my real question is: "Why is there a _smaller_ gap when there are fewer columns?" 😄

src/components/form/form.scss
adrianschmidt: ⚡ 

See line 29 in this file:

```scss
    // Using `minmax(0, 1fr)` below, instead of just `1fr` ensures that
    // components that are wider than its column doesn't make the column grow
    // and squash the other columns. /Ads
    grid-template-columns: repeat(var(--number-of-columns), minmax(0, 1fr));
```

adrianschmidt: I haven't looked at what happens with a select with loads of options selected, but if there's a problem, it's something that will have to be fixed in limel-select (or whatever other component might be affected), and I don't think we have to fix any such issues in _this_ PR.

src/components/form/form.scss
```diff
@@ -10,12 +10,116 @@
...
+    gap: var(--gap);
+}
+
 .limel-form-layout--grid {
+    --gap: 1rem;
+    --min-height-of-one-row: 4.71rem;
     display: grid;
-    grid-template-columns: repeat(auto-fill, minmax(18.75rem, 1fr));
-    gap: 0.5rem;
+    padding: var(--gap);
```
adrianschmidt, line 23: This padding is going to be a headache:

image

adrianschmidt: This example shows the problem perfectly:

image

The left side of the text and the left side of the first column doesn't line up, and making them line up while this padding is still here is going to be a _lot_ messier than just letting the consumer add the necessary whitespace, when necessary.

If we think this specific amount of padding will often be useful, we can add an optional helper-class that adds the padding.

src/components/form/examples/basic-form.tsx
Kiarokh: You're right! The next section is better to be removed. But I'm just thinking if now somehow magically due to latest changes, whether applying an inline style for `min-height` works better than `height`? This way, rows won't be "locked".

First time we tested, `min-height` wouldn't help the `textarea` to visually and vertically expand inside its container grid cell. But maybe now it does.

Kiarokh: Also something off-topic, but I must say, I really dislike the way this `–` character is supposed to be used when I think about it... 😞 

> [···] form—let's say a map—you [···]

It has serious usability problems in my opinion 😄 I'd read those tow words before and after that character as attached words. 😢 
the difference between this: `-` (normal dash) and this: `–` longer dash is simply too little to notice for normal people (like myself) - –. Is there only one standard for writing this?

src/components/form/form.scss
Kiarokh: you're right. Let's make new issues for them to fix.

src/components/form/form.scss
Kiarokh: I agree. In a dream world, consumers should put their own padding. But I bet that they don't, because this is a pure UI-design questions; and they don't/can't care about these stuff.

Can we use a var with a default-and-editable value and a fallback (like we do in many other components)? Do you know what I mean? We do things like this: `background-color: var(--icon-background-color, transparent);` in many places. 

Can we do like: `padding: var(--gap, 1rem);`

src/components/form/form.scss
Kiarokh: Good question and it tickles, because I try to be smart here. But I'm not sure if I am 😄 

My assumption is:
When users choose a 5 columned layout, it means that the UI has enough space to fit-in 5 columns. Therefore, the UI can as well be more "airy" (to improve readability of the content and provide more space for the eyes to rest).

(Same thing when they don't deliberately specify number of columns, but the form decides that it can display 5 columns because there is enough space. This will essentially make the UI more compact on smaller containers or devices.)

What do you think?

src/components/form/form.scss
adrianschmidt: Yeah, ok, that would work, yes. Good idea!

adrianschmidt: Hmm… by the way… is it intentional that this padding is only applied to `.limel-form-layout--grid`, not to `.limel-form-layout--default`? Or is it just an oversight?

adrianschmidt: AFK discussion sum-up: We'll remove the padding for now, so we can get this merged. We can always add it later.

src/components/form/examples/basic-form.tsx
adrianschmidt: Do you mean this fix?
image

Kiarokh: yes exactly

src/components/form/form.scss
adrianschmidt: Rule of thumb: If it feels clever, it's a red flag.

Reducing complexity, keeping things simple, is the most important thing we can do. We need to be minimalist not only in our visual design, but in terms of code complexity as well. Let's go with the smaller gap for now.

src/components/form/examples/basic-form.tsx
adrianschmidt: Way ahead of you 😉

src/components/form/form.scss
Kiarokh: > Let's go with the smaller gap for now.

If this means you wanna delete those lines, it's fine for me. ;)

src/components/form/form.types.ts
```diff
@@ -154,21 +154,54 @@ export interface FormComponentOptions {
     props?: Record<string, any>;
 }
 
-export interface FormLayoutOptions {
+export interface FormLayoutOptions<T = FormLayoutType.Default> {
```
adrianschmidt, line 157: Considering the change on line 138 (see below), I was a little worried this might be a breaking change, but I tried removing the `<any>` and the prod-build was still successful, so it seems to me this change is non-breaking.

```diff
-    layout?: FormLayoutOptions;
+    layout?: FormLayoutOptions<any>;
```

(Just adding this as a bit of a record for posterity, in case it turns out this actually does break stuff 😄)

src/components/form/form.types.ts
```diff
@@ -154,21 +154,54 @@ export interface FormComponentOptions {
...
+     */
+    // eslint-disable-next-line no-magic-numbers
+    columns?: 1 | 2 | 3 | 4 | 5;
+
+    /**
+     * Attempts to fill in holes earlier in the grid, if smaller items come up
+     * later. This may cause items to appear out-of-order, when doing so would
+     * fill in holes left by larger items. Defaults to `true`
+     */
+    dense: boolean;
```
adrianschmidt, line 195: @jgroth Can't this also be optional?

jgroth: Yes! It should be! I'm not sure why it's not, must be a mistake :sweat_smile:

src/components/form/form.scss
adrianschmidt: Yes, sorry, wasn't very clear there. I've indeed removed them. I've just forgotten to push 😄

src/components/form/examples/form-layout.tsx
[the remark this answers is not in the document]
Kiarokh: Good point! It's something that Lime is directly benefiting from. Even if it wasn't, you could still argue that it brings free publicity for the company 😄 

But we should present it decently also, (which was my latest PR in Kompendium).

adrianschmidt (approved):

PR #1048, opened 2020-11-20
fix(input-field): hide character counter when not needed & more

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/1614

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/examples/input-field-textarea.tsx
+++ b/src/components/input-field/examples/input-field-textarea.tsx
@@ -41,7 +41,7 @@ export class InputFieldTextareaExample {
                 <limel-flex-container justify="end">
                     <limel-checkbox
                         onChange={this.setEnabled}
-                        label="Enabled"
+                        label="Disabled"
                     />
                     <limel-checkbox
                         onChange={this.setRequired}
--- a/src/components/input-field/input-field.scss
+++ b/src/components/input-field/input-field.scss
@@ -131,3 +131,12 @@ input.mdc-text-field__input {
         margin: auto;
     }
 }
+
+.mdc-text-field-character-counter {
+    transition: opacity 0.2s ease;
+    opacity: 0;
+
+    .mdc-text-field--focused + .mdc-text-field-helper-line & {
+        opacity: 1;
+    }
+}
--- a/src/components/input-field/input-field.scss
+++ b/src/components/input-field/input-field.scss
@@ -115,6 +115,7 @@ div {
 .mdc-text-field.mdc-text-field--textarea {
     textarea {
         height: var(--textarea-height);
+        resize: vertical;
     }
 }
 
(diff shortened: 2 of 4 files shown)


src/components/input-field/examples/input-field-textarea.tsx
```diff
@@ -41,7 +41,7 @@ export class InputFieldTextareaExample {
                 <limel-flex-container justify="end">
                     <limel-checkbox
                         onChange={this.setEnabled}
-                        label="Enabled"
+                        label="Disabled"
```
adrianschmidt, line 44: Good catch 👍

src/components/input-field/input-field.scss
```diff
@@ -131,3 +132,12 @@ input.mdc-text-field__input {
         margin: auto;
     }
 }
+
+.mdc-text-field-character-counter {
+  transition: opacity 0.2s ease 0.5s;
```
adrianschmidt, line 137: Nice, but I find it pretty annoying that the helper-line doesn't have the same transition. We should either add it there too, or remove it from here.

adrianschmidt (changes requested):
One small remark, but otherwise some *very* nice changes! 👏

src/components/input-field/input-field.scss
Kiarokh: Thanks. Well, this was actually intentional. I wanted things to happen with slight delays, so that user can actually follow what is being added to the UI. Because we also have the border animation when field gets focused. But it's not hard to remove the `0.5s` delay. 

What do you think?

adrianschmidt: I think we should definitely remove the delay. Animations call attention, and when something happens .5 seconds after I did something, it draws my attention, and that shouldn't be abused for drawing attention to something for no good reason. For no good reason, because more attentive users will notice the character counter even if it's displayed or faded in simultaneously with the border and the helper text, and less attentive users will probably notice it when they start typing and the counter starts ticking up. Even the least attentive users ought to notice it when it hits the max-count, and (presumably) turns red.

adrianschmidt: I think the fading should be used to *lessen* the impact of the showing and hiding of the label, not increase it.

Kiarokh: fair. I'll fix

Kiarokh: fixup pushed

adrianschmidt (approved):

PR #1053, opened 2020-12-01
docs(form): add example for `propsFactory`

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/examples/props-factory-form.tsx
+++ b/src/components/form/examples/props-factory-form.tsx
@@ -1,0 +1,57 @@
+import { Component, h, State } from '@stencil/core';
+import { schema } from './props-factory-schema';
+
+/**
+ * Using `propsFactory`
+ *
+ * @link props-factory-schema.ts
+ * @link props-factory-picker.tsx
+ */
+@Component({
+    tag: 'limel-example-props-factory-form',
+    shadow: true,
+})
+export class PropsFactoryFormExample {
+    @State()
+    private formData: object = {
+        hero: 1001,
+    };
+
+    private anObjectToInject = {
+        someProp: 'The object was successfully injected!',
+    };
+    private aStringToInject = 'The string was successfully injected!';
+
+    constructor() {
+        this.propsFactory = this.propsFactory.bind(this);
+        this.handleFormChange = this.handleFormChange.bind(this);
+    }
+
+    public render() {
+        return [
+            <limel-form
+                onChange={this.handleFormChange}
+                value={this.formData}
+                schema={schema}
+                propsFactory={this.propsFactory}
+            />,
+            <limel-example-value value={this.formData} />,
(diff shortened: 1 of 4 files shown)


src/components/form/examples/props-factory-form.tsx
```diff
@@ -44,7 +44,7 @@ export class PropsFactoryFormExample {
             'limel-example-props-factory-picker'
         ) {
             console.log('propsFactory - subSchema:', subSchema);
-            const properties = { injected: this.somePropertyToInject };
+            const properties = { injectedObject: this.somePropertyToInject };
```
adrianschmidt, line 47: I would like to show that props should be camelcased when set by the propsFactory, but I couldn't come up with a better name than `injectedObject`, which might just make people assume we want them to use Hungarian notation… 🤔

jgroth: Maybe one of these will do? :slightly_smiling_face: 
- `injectedData`
- `myData`
- `componentData`

adrianschmidt: ⚡

src/components/form/examples/props-factory-form.tsx
```diff
@@ -0,0 +1,57 @@
...
+})
+export class PropsFactoryFormExample {
+    @State()
+    private formData: object = {
+        hero: 1001,
+    };
+
+    private somePropertyToInject = {
+        message: 'Success!',
+    };
```
jgroth, line 22: Is there a reason to why this is nested inside an object? It might just make this look more complicated than it is :thinking:  If a message needs to be injected into the component, a property called `message` is probably better. What do you think? :slightly_smiling_face: Same reasoning for the property in the other component

adrianschmidt: Yeah, but this isn't supposed to just be a message, it's a placeholder for some complicated object that needs inserting, like `platform`…

Maybe we should change the names to indicate that more clearly? Or give it some more dummy properties or something? I'm not sure…

jgroth: You can inject whatever you want, either a simple message or a complicated object. I guess the intended use is to put properties here that cannot be expressed in the schema :thinking: The most simple thing here would probably be an object that is an instance of a class or just an object with a method.

But yeah, it's probably just a naming issue. If we can come up with something more realistic I think it would be great :slightly_smiling_face:

jgroth (dismissed):

Issue #1065, opened 2020-12-08
Icon button component affected by MDC Ripple bugs


anderssonjohan:
## Current behavior

1. Focused state makes button look stuck

This makes a button look like this when it's being clicked.

![mdcripple-background-focused-sticky](https://user-images.githubusercontent.com/435885/101535333-c9e56b80-3998-11eb-94dc-c8c196cbf4b1.gif)

2. `:hover` is sticky on iOS so ripple never goes away until other element is focused

## Expected behavior

1. Focused state issue

Workaround: Remove CSS class manually in click handler

```javascript
onClick={(ev) => {
    const e: any = ev.currentTarget;
    const mdcButton = e.shadowRoot.querySelector('button');
    setTimeout(() => mdcButton.classList.remove('mdc-ripple-upgraded--background-focused'), 100);
}}
```

![mdcripple-background-focused-workaround](https://user-images.githubusercontent.com/435885/101535750-69a2f980-3999-11eb-87a4-e761ba9f0f31.gif)

This focused state behavior is not a feature in MDC on other platforms than MDC Web and is reported over and over in various places. In every place I've found it has been closed because of accessibility reasons.
https://github.com/material-components/material-components-web/issues/2337#issuecomment-370544239

This accessibility feature is of course important, but only if you have a device where you can tab around AND you tabbed to the element, ie. not clicked or used touch input.
The expected behavior should be something like this comment suggests:
https://github.com/material-components/material-components-web/issues/1951#issuecomment-359084163

Example of expected behavior using the Github comment linked above. The focused state only appears when the button is tabbed to, NOT when it's clicked using the pointer.

![focusstyle](https://user-images.githubusercontent.com/435885/101534100-fac4a100-3996-11eb-886c-c5889536af09.gif)

Suggested fix: Remove focused CSS class when button is clicked, not tabbed to

```typescript
@Listen('click')
private removeFocusedStateOnClick() {
    const mdcButton = this.host.shadowRoot.querySelector('.mdc-icon-button');
    mdcButton.classList.remove('mdc-ripple-upgraded--background-focused');
}
```

2. The sticky `:hover` state on iOS

When the workaround for the focused state is in place, the sticky ripple still exists on mobile devices.
This is due to the `:hover` state not being removed. Some very old references to this issue:
https://bugs.webkit.org/show_bug.cgi?id=158517

[...]

PR #1066, opened 2020-12-08
fix(icon-button): remove sticky mdc ripple effects

anderssonjohan:
fix: #1065 

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [x] Edge
- [x] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/icon-button/icon-button.tsx
+++ b/src/components/icon-button/icon-button.tsx
@@ -13,6 +13,12 @@ import { IconSize } from '@limetech/lime-elements';
     styleUrl: 'icon-button.scss',
 })
 export class IconButton {
+    public constructor() {
+        this.removeFocusedStyleOnClick = this.removeFocusedStyleOnClick.bind(
+            this
+        );
+    }
+
     /**
      * The icon to display.
      */
@@ -56,6 +62,13 @@ export class IconButton {
         }
     }
 
+    private removeFocusedStyleOnClick() {
+        const mdcButton = this.host.shadowRoot.querySelector(
+            '.mdc-icon-button'
+        );
+        mdcButton.classList.remove('mdc-ripple-upgraded--background-focused');
+    }
+
     public connectedCallback() {
         this.initialize();
     }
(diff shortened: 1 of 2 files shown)


adrianschmidt:
And of course he's already opened a PR for it too… 😄 This is amazing ❤️

I'll look at it as soon as I can, but that might not be until cooldown.

anderssonjohan:
@adrianschmidt No rush! Please note that I changed from using `@Listen('click')` (in my issue) to using add/remove event handler due to our eslint rule. However, I couldn't find the reason for why `@Listen` should be avoided for vDOM events. I hoped to find some explanations but I could only find this: https://github.com/ionic-team/stencil-eslint/issues/10#issuecomment-674301520

adrianschmidt:
> @adrianschmidt No rush! Please note that I changed from using `@Listen('click')` (in my issue) to using add/remove event handler due to our eslint rule. However, I couldn't find the reason for why `@Listen` should be avoided for vDOM events. I hoped to find some explanations but I could only find this: [ionic-team/stencil-eslint#10 (comment)](https://github.com/ionic-team/stencil-eslint/issues/10#issuecomment-674301520)

Yeah, I have no idea… we should perhaps disable that rule?

@jgroth do you have any idea?

jgroth:
Hmm no I don't know :thinking: 
However I read your comment in the other thread and would guess that the reason is something like that :slightly_smiling_face: 
It feels like we could probably disable it if we want

adrianschmidt (approved):

src/components/icon-button/icon-button.tsx
```diff
@@ -72,10 +85,12 @@ export class IconButton {
 
         this.mdcIconButtonRipple = new MDCRipple(element);
         this.mdcIconButtonRipple.unbounded = true;
+        this.host.addEventListener('click', this.removeFocusedStateOnClick);
```
adrianschmidt, line 88: Since activating the button by pressing enter on a keyboard creates a synthetic `click` event, this removes the styling even if the user presses enter, instead of clicking. That's not optimal, but I think we can live with it. If the user interacts with the element, they hopefully know "where they are", despite the visual style disappearing.

The actual focus isn't removed, just the styling. (Which, by the way, makes me want to rename `removeFocusedStateOnClick` to `removeFocusedStyleOnClick`. I think I'll do that.)

adrianschmidt (approved):

Issue #1059, opened 2020-12-07
limel-file: expose icon/ icon-color of chip 


BregenzerK:
## New feature motivation
Different icons and colors can help to easier differentiate between different file types/ formats. We also have scenarios where the populated value is picked in the step ahead and therefore not directly accessible anymore to the user. To avoid confusions it would be good if the value couldn't be remove anymore in this case
 
## New feature description
When populating the component with a file value, the consumer should be able to overwrite the default icon and its color and set removeable to false

## New feature implementation

enable overwrite for `icon`, `iconColor` and `removeable` in the [default chip](https://github.com/Lundalogik/lime-elements/blob/master/src/components/file/file.tsx#L14)

PR #1075, opened 2020-12-14
feat(limel-file): support custom icon and icon-color

joelricci:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

closing #1059

### DIFF
--- a/src/components/file/file.tsx
+++ b/src/components/file/file.tsx
@@ -143,26 +143,26 @@ export class File {
 
         return [
             <input
-                id={this.fileInputId}
-                type="file"
-                onChange={this.handleFileChange}
                 hidden={true}
+                id={this.fileInputId}
+                onChange={this.handleFileChange}
+                type="file"
             />,
             <limel-chip-set
                 disabled={this.disabled}
                 label={this.label}
-                required={this.required}
-                onKeyDown={this.handleKeyDown}
-                onKeyUp={this.handleKeyUp}
-                type="input"
-                value={chipArray}
-                onClick={this.handleFileSelection}
+                leadingIcon="upload_to_cloud"
                 onChange={this.handleChipSetChange}
-                onInteract={this.preventAndStop}
-                onDrop={this.handleFileDrop}
+                onClick={this.handleFileSelection}
                 onDragEnter={this.preventAndStop}
                 onDragOver={this.preventAndStop}
-                leadingIcon="upload_to_cloud"
+                onDrop={this.handleFileDrop}
+                onInteract={this.preventAndStop}
+                onKeyDown={this.handleKeyDown}
+                onKeyUp={this.handleKeyUp}
+                required={this.required}
+                type="input"
+                value={chipArray}
             />,
         ];
     }
--- a/src/components/file/file.tsx
+++ b/src/components/file/file.tsx
@@ -15,9 +15,8 @@ const DEFAULT_FILE_CHIP: Chip = {
     id: null,
     text: null,
     removable: true,
-    icon: 'note',
-    iconColor: 'var(--lime-light-grey)',
 };
+const DEFAULT_ICON = 'note';
 
 /**
  * This component lets end-users select a *single* file from their device
@@ -49,10 +48,12 @@ const DEFAULT_FILE_CHIP: Chip = {
  *:::
  *
  * @exampleComponent limel-example-file
+ * @exampleComponent limel-example-file-custom-icon
  */
 @Component({
     tag: 'limel-file',
     shadow: true,
+    styleUrl: 'file.scss',
 })
 export class File {
     /**
@@ -164,6 +165,9 @@ export class File {
                       ...DEFAULT_FILE_CHIP,
                       text: this.value.filename,
                       id: this.value.id,
+                      icon: this.value.icon || DEFAULT_ICON,
+                      iconFillColor: this.value.iconColor,
+                      iconBackgroundColor: this.value.iconBackgroundColor,
                   },
               ]
             : [];
--- a/src/components/file/file.tsx
+++ b/src/components/file/file.tsx
@@ -131,16 +131,6 @@ export class File {
     }
 
     public render() {
-        const chipArray = this.value
-            ? [
-                  {
-                      ...DEFAULT_FILE_CHIP,
-                      text: this.value.filename,
-                      id: this.value.id,
-                  },
-              ]
-            : [];
-
         return [
             <input
                 hidden={true}
@@ -162,9 +152,21 @@ export class File {
                 onKeyUp={this.handleKeyUp}
                 required={this.required}
                 type="input"
-                value={chipArray}
+                value={this.chipArray}
             />,
         ];
+    }
+
+    private get chipArray() {
+        return this.value
+            ? [
+                  {
+                      ...DEFAULT_FILE_CHIP,
+                      text: this.value.filename,
+                      id: this.value.id,
+                  },
+              ]
+            : [];
     }
 
     private handleKeyDown(event: KeyboardEvent) {
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -578,7 +578,19 @@ export class ChipSet {
 
     private renderIcon(chip: Chip) {
         const style = {};
+        if (chip.iconFillColor) {
+            style['--icon-color'] = chip.iconFillColor;
+        }
+
+        if (chip.iconBackgroundColor) {
+            style['--icon-background-color'] = chip.iconBackgroundColor;
+        }
+
         if (chip.iconColor) {
+            // eslint-disable-next-line no-console
+            console.warn(
+                '`Chip.iconColor` is deprecated. Use `Chip.iconBackgroundColor` to set the background color of the icon.'
+            );
             style['--icon-background-color'] = chip.iconColor;
         }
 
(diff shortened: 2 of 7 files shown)


src/components/file/examples/file-custom.tsx
```diff
@@ -0,0 +1,51 @@
+import { FileInfo } from '@limetech/lime-elements';
```
adrianschmidt, line 1: We usually leave off the `limel-` part in the commit message scope. So for consistency, I suggest renaming the commit `feat(file): support custom icon and icon-color` 🙂

joelricci: Fixed :)

src/components/file/file.tsx
```diff
@@ -131,38 +160,28 @@ export class File {
...
                 onDragEnter={this.preventAndStop}
                 onDragOver={this.preventAndStop}
-                leadingIcon="upload_to_cloud"
+                onDrop={this.handleFileDrop}
+                onInteract={this.preventAndStop}
+                onKeyDown={this.handleKeyDown}
+                onKeyUp={this.handleKeyUp}
+                required={this.required}
+                type="input"
+                value={this.chipArray}
```
adrianschmidt, line 184: Are you using some automatic formatter that sorts attributes alphabetically?

I don't mind having them sorted alphabetically, but then we need to have that done by prettier, so we don't get commits like this, with one small change, and lots of completely unrelated stylistic changes… 😕

You'll have to either remove these changes, or put them in their own commit (it can still be in this PR, but must be a separate commit).

adrianschmidt (changes requested):

src/components/file/file.tsx
joelricci: Sorry, my ocd just can't stand unsorted properties (and imports) so I do that manually but agree it should be done by the formatter. I will fix it.

joelricci: Should be good now @adrianschmidt

src/components/file/file.tsx
```diff
@@ -143,26 +143,26 @@ export class File {
...
```
adrianschmidt, line 184: At this point, this should still be `value={this.chipArray}`. It should only change in the next commit.

Sorry for being a pain in the butt…

joelricci: I'm aware of that but couldn't get VSC to cooperate and hoped you wouldn't catch it :D

joelricci: I can spend another 15 minutes to fix it if you really want.

src/components/file/file.tsx
adrianschmidt: For the javascript imports, the order probably doesn't matter, but please note that in the scss files, the import order is important, so please don't change the order of imports there…

src/components/file/file.tsx
```diff
@@ -67,6 +68,18 @@ export class File {
...
+     * Optional file icon.
+     */
+    @Prop()
+    public fileIcon: string;
+
+    /**
+     * Optional file icon color.
+     */
+    @Prop()
+    public fileIconColor: string;
```
adrianschmidt, line 81: Since this sets the background colour of the icon "badge", we will probably want to be able to set the icon colour too, before too long. I'm thinking we should probably add that while we're at it. What do you think?

src/components/file/file.tsx
adrianschmidt: I really do. But I can fix it for you 😄

adrianschmidt: Done 😉

src/components/file/examples/file-default.tsx
```diff
@@ -21,9 +24,9 @@ export class FileExample {
         return [
             <limel-file
                 label="Attach a file"
-                value={this.value}
-                required={this.required}
                 onChange={this.handleChange}
+                required={this.required}
+                value={this.value}
```
adrianschmidt, line 29: This is another change that doesn't belong in this commit. I'll fix it for you this time, but for next time, if you absolutely have to sort stuff, you know to do it beforehand 🙂

src/components/file/file.tsx
joelricci: That was my initial thought as well. However I backed on that when I saw that the `chip-set` component doesn't seem to have support for that other than via CSS since the `iconColor` property on the `Chip` type actually sets the _background colour_ which is a bit confusing. So don't know what's the best approach here.

adrianschmidt: Hmm… I wanted to rename this to something like `fileIconBackgroundColor`, but found this in `chip.types.ts`:

image

@jgroth do you remember why we had to override `--icon-background-color` in the first place?

joelricci: Hence the confusion ↑

adrianschmidt: Ah, yeah, I hadn't seen your comment 😆

src/components/file/examples/file-default.tsx
joelricci: 🙏

src/components/file/file.tsx
```diff
@@ -93,6 +106,22 @@ export class File {
     private chipSet;
     private mdcTextField;
 
+    private get chipArray() {
```
adrianschmidt: I don't know if there's any benefit to using a getter as opposed to just creating the chip in the render function, apart from making the code look a little nicer?

It does expose the chip outside the component (since `private` is a typescript lie 😄), which I'm inclined to see as a negative.

I reverted this change in my first fixup, but that was mainly because I wanted to keep the changes from `master` to be minimal. I can always bring back the refactor to using a getter in a separate commit.

Thoughts @jgroth?

src/components/file/file.tsx
adrianschmidt: Turns out both colours can be set using the appropriate css variables. So I revert part of your change, and added proper documentation for how to use those variables.

src/components/file/file.tsx
jgroth: There's no way for an outsider to get hold of the instance where all the methods are declared since the component instance (`this`) is not the same as the DOM element, so `private`/`public` does not really matter in most cases. To expose a method on the DOM element you will have to decorate it with `@Method`

src/components/file/file.tsx
jgroth: I guess that this can be used if you want different colors for the chips if you have multiple ones as part of a chip set. Setting the CSS property will set the same color on all of the chips

adrianschmidt: Ah, yeah, that makes sense 👍

`limel-file` only supports one chip at the moment though, so we can use the css variables in this case 🙂

src/components/file/file.tsx
adrianschmidt: Hmm… yeah, I can't seem to get that node… but is that really not possible? If so, that's great!

I guess it's not really a big problem in this case anyway, so let's go with that getter 👍

adrianschmidt (approved):

src/components/file/file.tsx
```diff
@@ -67,6 +68,12 @@ export class File {
     @Prop({ reflect: true })
     public label: string;
 
+    /**
+     * Optional file icon.
+     */
+    @Prop()
+    public fileIcon: string;
```
jgroth, line 75: Do you think we need the `file` prefix on this property?  What does it add compared to just `icon`? :slightly_smiling_face:

adrianschmidt: Currently nothing, but since we have the option of leading and trailing icons in `limel-input-field`, it doesn't seem impossible that we might add some option for that here as well at some point, and if that happens, it will be helpful that this is already named appropriately. So I think we should keep it. Thoughts?

jgroth: Maybe it shoud be part of the `FileInfo` interface now that I think a bit more about it :thinking:

jgroth: That way we could support multiple files with different icons in the future

adrianschmidt: It is already part of the FileInfo api… and of course, that means we can have the consumer change it when `value` is updated… didn't think of that. Thanks! 😄

adrianschmidt: Ah… no… it's not part of the `FileInfo` interface, it's part of the `Chip` interface, but that's not exposed. I was confusing them.

Yeah, we should add that.

adrianschmidt: @joelricci How urgent is the custom-icon feature for you? I'm just thinking of how I should prioritise this versus some other things in the feature-cycle…

joelricci: @adrianschmidt Probably not super-urgent. Let me check with the team...

joelricci: So it's not very urgent, but would appreciate if someone could keep me updated on the progress as we have other issues and PR we want to do depending on it. Also let me know if I can help or you need any input. Thanks for the help gents!

adrianschmidt: Yeah, absolutely! I would like to get it done asap as well.

adrianschmidt: :zap:

@jgroth Can you please do a final review of this, since I've made quite a few changes to it myself?

jgroth: It looks good! :slightly_smiling_face:

jgroth (dismissed):

adrianschmidt (approved):
Ok:ed by jgroth above. I just squashed the fixups.

PR #1076, opened 2020-12-14
fix(form): emit undefined from date-picker when cleared

jgroth:
fix Lundalogik/crm-feature#1643

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/widgets/date-picker.ts
+++ b/src/components/form/widgets/date-picker.ts
@@ -47,6 +47,12 @@ export class DatePicker extends React.Component {
             return;
         }
 
+        if (!event.detail) {
+            props.onChange(undefined);
+
+            return;
+        }
+
         const formatMapping = {
             'date-time': 'YYYY-MM-DDTHH:mm:ss',
             date: 'YYYY-MM-DD',


maltja:
It seems that it is not possible to save an empty/cleared date. If you save it and then reload the page the old date shows up in the date-picker. However, I tried to save another date and that didn't work either (then I got a 400 bad request error) , so maybe it's another issue? 🤔

jgroth:
> It seems that it is not possible to save an empty/cleared date. If you save it and then reload the page the old date shows up in the date-picker. However, I tried to save another date and that didn't work either (then I got a 400 bad request error) , so maybe it's another issue?

Yeah, that does not seem related to this fix :thinking: 
It's probably an issue in the webclient

maltja:
> > It seems that it is not possible to save an empty/cleared date. If you save it and then reload the page the old date shows up in the date-picker. However, I tried to save another date and that didn't work either (then I got a 400 bad request error) , so maybe it's another issue?
> 
> Yeah, that does not seem related to this fix 🤔
> It's probably an issue in the webclient

Make sense, I create an new issue for that 🙂

maltja (approved):

PR #1078, opened 2020-12-15
fix(menu-surface): resize the menu's height when the host element's h…

maltja:
…eight changes

fix Lundalogik/crm-feature#1635

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu-surface/menu-surface.tsx
+++ b/src/components/menu-surface/menu-surface.tsx
@@ -41,6 +41,7 @@ export class MenuSurface {
     private host: HTMLLimelMenuSurfaceElement;
 
     private menuSurface: MDCMenuSurface;
+    private observer: IResizeObserver;
 
     constructor() {
         this.handleDocumentClick = this.handleDocumentClick.bind(this);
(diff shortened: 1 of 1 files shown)


src/components/menu-surface/menu-surface.tsx
```diff
@@ -61,7 +61,7 @@ export class MenuSurface {
     }
 
     public componentDidUpdate() {
-        this.ensureMenuFitsInViewPort();
+        this.createResizeObserver(this.host);
```
jgroth, line 64: Do we have to create a new observer every time the component is updated? :thinking: We should be able to just create one and then remove it when the component is destroyed

maltja: I also thought about that, now the observer is removed in the timeout function. I don't know what is best, only have it when it is resizing(updating) or have it there all the time? but do you think it is better that the observer is always there? 🙂

jgroth: It will allow us to remove the ugly `setTimeout` :sweat_smile:

maltja: haha 😆 I'll do a fix-up then!

jgroth (dismissed):

PR #1083, opened 2020-12-17
fix(date-picker): add timezone to date-time format type

Befkadu1:
fix Lundalogik/crm-feature#1690

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/widgets/date-picker.ts
+++ b/src/components/form/widgets/date-picker.ts
@@ -54,7 +54,7 @@ export class DatePicker extends React.Component {
         }
 
         const formatMapping = {
-            'date-time': 'YYYY-MM-DDTHH:mm:ss',
+            'date-time': 'YYYY-MM-DDTHH:mm:ssZ',
             date: 'YYYY-MM-DD',
             time: 'HH:mm:ss',
         };


src/components/form/widgets/date-picker.ts
```diff
@@ -54,11 +54,16 @@ export class DatePicker extends React.Component {
...
-            'date-time': 'YYYY-MM-DDTHH:mm:ss',
+            'date-time': 'YYYY-MM-DDTHH:mm:ssZ',
             date: 'YYYY-MM-DD',
             time: 'HH:mm:ss',
         };
-        const dateString = moment(event.detail).format(
+        const value =
+            props.schema.format !== 'date-time'
+                ? event.detail
+                : `${event.detail}Z`;
```
jgroth, line 64: I do not understand why this is needed, and it seems to mess up the time that you select. It will offset the time more than it should. Can we not just remove this?

jgroth: This also converts `event.detail` from an `object` to a `string` with `Z` at the end, that might be what causes the weird behavior

Befkadu1: Yea it should be removed.

jgroth (dismissed):

jgroth (approved):

PR #1084, opened 2020-12-17
fix(input-field): improve action for incomplete url:s, improve target for relative url:s

maltja:
I copied the lime-link-helper.service.ts file from the webclient repo that was used in the old card and did some small refactoring changes to it. ~~The type of the input-field cannot be url because then the full url is required. Therefore, it is set to type = 'text'.~~

fix Lundalogik/crm-feature#1628

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -23,7 +23,7 @@ import {
 } from '../../util/keycodes';
 import { InputType } from './input-field.types';
 import { ListItem } from '@limetech/lime-elements';
-import { getHref } from './link-helper';
+import { getHref, getTarget } from './link-helper';
 
 interface LinkProperties {
     href: string;
@@ -514,7 +514,7 @@ export class InputField {
                 break;
             default:
                 props.href = getHref(this.value);
-                props.target = '_blank';
+                props.target = getTarget(this.value);
         }
 
         return props;
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -23,6 +23,7 @@ import {
 } from '../../util/keycodes';
 import { InputType } from './input-field.types';
 import { ListItem } from '@limetech/lime-elements';
+import { getHref } from './link-helper';
 
 interface LinkProperties {
     href: string;
@@ -512,7 +513,7 @@ export class InputField {
                 props.href = `tel:${this.value}`;
                 break;
             default:
-                props.href = `${this.value}`;
+                props.href = getHref(this.value);
                 props.target = '_blank';
         }
 
--- a/src/components/input-field/link-helper.ts
+++ b/src/components/input-field/link-helper.ts
@@ -1,0 +1,53 @@
+export function getHref(value: string) {
+    const href = value ? String(value.trim()) : '';
+    if (isValid(href)) {
+        return href;
+    }
+
+    return prependProtocol(href);
+}
+
+export function getTarget(value: string) {
+    const url = getHref(value);
+    if (isRelativeLink(url)) {
+        return '_self';
+    }
+
+    return '_blank';
+}
+
+export function prependProtocol(input: string) {
+    if (!input) {
+        return input;
+    }
+
+    return 'https://' + input;
+}
+
+function isValid(href: string) {
+    return (
+        hasKnownProtocol(href) ||
+        isRelativeLink(href) ||
+        hasRelativeProtocol(href)
+    );
+}
+
+function hasKnownProtocol(input: string) {
+    const knownProtocols = ['ftp', 'ftps', 'https', 'http'];
+
+    return knownProtocols.some((knownProtocol) => {
+        return input.indexOf(knownProtocol + '://') === 0;
+    });
+}
+
+function isRelativeLink(input: string) {
+    if (hasRelativeProtocol(input)) {
+        return false;
+    }
+
+    return input.startsWith('/') || input.startsWith('#');
+}
+
+function hasRelativeProtocol(input: string) {
+    return input.startsWith('//');
+}
--- a/src/components/input-field/examples/input-field-showlink.tsx
+++ b/src/components/input-field/examples/input-field-showlink.tsx
@@ -1,4 +1,4 @@
-import { Component, h, State, Watch } from '@stencil/core';
+import { Component, h, State } from '@stencil/core';
 
 /**
  * With `showLink=true`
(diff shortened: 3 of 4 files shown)


src/components/input-field/input-field.tsx
```diff
@@ -512,8 +513,8 @@ export class InputField {
                 props.href = `tel:${this.value}`;
                 break;
             default:
-                props.href = `${this.value}`;
-                props.target = '_blank';
+                props.href = `${limeLinkHelperService.getHref(this.value)}`;
```
Befkadu1, line 516: Is it obligatory to make it as a string here?

maltja: Nope, you are right, I will fix it 🙂

src/components/input-field/input-field.tsx
```diff
@@ -723,6 +724,14 @@ export class InputField {
...
     }
 
+    private getType() {
+        if (this.type === 'url') {
+            return 'text';
+        }
+
+        return this.type;
+    }
+
```
jgroth, line 734: I'm not sure this is the correct place for this bugfix. This will more or less remove the feature of having URL fields altogether, so it will no longer work for consumers using this field outside of a form either.

This should probably be fixed on a schema level since it is in the schema that we have defined that we should use a URL field. However, I think it is a missed opportunity to not use validation for stuff like this. If a URL is required here we should indicate if the user has entered something that is not valid. Sure, it might be annoying at the moment but we are not using the validation status of the form to determine if a user can save or not. The field will also not display an error until it is interacted with in some way.

Maybe you discussed all this in the project before I joined. What are your thoughts? :slightly_smiling_face:

jgroth (changes requested):

src/components/input-field/input-field.tsx
maltja: Isn't the only difference between the types, url and text, that url is validated? 

I think it is nice to only have to write the url without specifying the protocol. I thought that it should always be the same for the user if they should specify the full url or not. Otherwise, I think it might be confusing if they sometimes only need to write a part of the url and in another part of the application must write the entire part of the url. That is why I put it in lime-elements.  So I think if we don't want to fix it in lime-element we shouldn't fix it at all, so it is consistent thought the application. 🙂

maltja: I don't think we discussed it really, we just compared it with the old card design and then said it was a bug 😅

jgroth: > Isn't the only difference between the types, url and text, that url is validated?

Yes, but with this change we remove that possibility altogether so this will be a breaking change :slightly_smiling_face: Also, `type='url'` will add an icon that you can click to navigate to that URL.

> I think it is nice to only have to write the url without specifying the protocol.

But then it's not a URL :smile:

> I thought that it should always be the same for the user if they should specify the full url or not. Otherwise, I think it might be confusing if they sometimes only need to write a part of the url and in another part of the application must write the entire part of the url. That is why I put it in lime-elements. So I think if we don't want to fix it in lime-element we shouldn't fix it at all, so it is consistent thought the application. :slightly_smiling_face:

Yes that is true, but it might also depend in what context the user is in when doing that :thinking: 
Either way, I'm not sure how much you have already discussed around this issue, but if this should be fixed I do not think we should put it on this level. Lime Elements can be used by other parties as well, and those might want the URL `type` even if we decide not to use it in the webclient :slightly_smiling_face:

jgroth: > I don't think we discussed it really, we just compared it with the old card design and then said it was a bug

Haha ok :sweat_smile:

maltja: Okey, I can move it then or should we wait with this one until the issue is discussed. I don't know what I should do. If I shouldn't fix it at all or if I should move it? 🤔

maltja: The problem if we don't fix it is that, if a user has an invalid url (that worked before the new card), it will not work to click on the trailing icon to go to that page.

jgroth: Yeah ok, but that we could probably fix though so that invalid URLs are still clickable
We cannot know the correct URL but we could still guess and just prefix it with either `http` or `https`

maltja:
What I did in the fixup: 
1. Copied/Used the link-helper from the crm-component repo instead (that code there was refactored and improved compared to the weblient repo).
2. We continue to use the url type which means that the url in the input-field will be validated and have a red border if invalid. However, this will only happen if the user click on the input-field.
3. If a user has written an invalid url (for example, without a protocol) the link will still work (this is why the link-helper is used).

src/components/input-field/link-helper.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    return prependProtocol(href);
+}
+
+function isValid(href) {
+    return (
+        hasKnownProtocol(href) ||
+        isRelativeLink(href) ||
+        hasRelativeProtocol(href) ||
+        hasMailtoProtocol(href)
```
adrianschmidt, line 15: Since we have a separate type for email addresses, I'm thinking we probably don't want to allow them here.

src/components/input-field/input-field.tsx
```diff
@@ -23,6 +23,7 @@ import {
 } from '../../util/keycodes';
```
adrianschmidt, line 23: > The type of the input-field cannot be url because then the full url is required. Therefore, it is set to type = 'text'.

That doesn't seem to be true anymore, so I'm removing those two sentences from the PR description.

adrianschmidt (approved):

adrianschmidt (changes requested):
Will update the commit message.

adrianschmidt (dismissed):

src/components/input-field/input-field.tsx
adrianschmidt: I'm going to merge this without that fix, for now, since there are still improvements here.

src/components/input-field/link-helper.ts
adrianschmidt: :zap:

adrianschmidt (approved):

PR #1086, opened 2020-12-23
docs: load stylesheet for light-mode only, as we don't fully support dark-mode yet

adrianschmidt:
Since we don't support dark-mode in most components, many of the examples in the docs look completely broken when viewed in dark-mode. This PR makes sure we only load the light-mode version of our colour variables.

Some examples on the _Guidelines / Color System_-page require both dark-mode and light-mode to be supported, so the reader can switch between them and see how that affects the colours. For those examples, we now specifically import the complete set of our colour variables instead.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/color-system/examples/color-palette-extended.css
+++ b/src/components/color-system/examples/color-palette-extended.css
@@ -1,0 +1,308 @@
+/*
+ * To be able to use colors with alpha, they are written in RGB.
+ * Comments with HEX codes in front of the values are used as reference only.
+ * Example of use:
+ * solid: `color: rgb(var(--contrast-100))`
+ * transparent: `color: rgba(var(--contrast-100), 0.5)`
+ */
+
+:host {
+    /* Lime Technologies Brand Colors (Do not have dark/light mode variants) */
+    --lime-brand-color-deep-red: 240, 87, 80; /* #f05750 */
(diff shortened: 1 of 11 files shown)


src/components/color-system/examples/color-palette-extended.css
```diff
@@ -0,0 +1,308 @@
+/*
+ * To be able to use colors with alpha, they are written in RGB.
+ * Comments with HEX codes in front of the values are used as reference only.
+ * Example of use:
+ * solid: `color: rgb(var(--contrast-100))`
+ * transparent: `color: rgba(var(--contrast-100), 0.5)`
+ */
+
+:host {
```
adrianschmidt, line 9: This file is just a copy of `lime-elements/src/style/color-palette-extended.css`, but with `:host` substituted for `:root` here and on line 169.

This fix is only needed until we enable dark-mode in the docs again, and the fix has been confined to a single commit, which can be reverted at that time.

Kiarokh (approved):
Really nice work! Thanks @adrianschmidt

PR #1088, opened 2021-01-04
hide slider when portal opens on top of it

Befkadu1:
fix Lundalogik/crm-feature#1657

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/slider/slider.scss
+++ b/src/components/slider/slider.scss
@@ -117,7 +117,6 @@ $colorGrey: #999999;
 
 .mdc-slider__thumb-container {
     top: 0;
-    z-index: 1;
 
     .mdc-slider:not(.mdc-slider--disabled) & {
         &:before,


src/components/slider/slider.scss
```diff
@@ -117,7 +117,6 @@ $colorGrey: #999999;
 
 .mdc-slider__thumb-container {
     top: 0;
-    z-index: 1;
```
adrianschmidt, line 120: @Kiarokh This z-index was introduced by you in #1018. Can you test this and make sure something else doesn't break from removing it?

(If we can remove it, that's really good. The fewer instances of z-index we need, the better.)

Befkadu1: > @Kiarokh This z-index was introduced by you in #1018. Can you test this and make sure something else doesn't break from removing it?
> 
> (If we can remove it, that's really good. The fewer instances of z-index we need, the better.)

I let @Kiarokh know about it and he was fine with it. Can you confirm it here?

Kiarokh: slider
the z-index was there to make sure that the `mdc-slider__thumb-container` isn't covering the visualizations of start and end point. :(

Kiarokh: I actually don’t think we need to fix this bug, since it seems to only happen in that specific Lime Elements example. Something we understood after investigations. 

Portal adds a dynamic z-index on dropdown menus. Which seems to be always higher than 1. So in the webclient this issue won't happen.

PR #1090, opened 2021-01-07
style(eslint): disallow multiline ternary operator

jgroth:
It's far too easy to abuse the ternary operator to make multiline statements in the code that is hard to read. I suggest we add an ESLint rule to not allow them at all

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dialog/dialog.tsx
+++ b/src/components/dialog/dialog.tsx
@@ -129,12 +129,15 @@ export class Dialog {
         this.mdcDialog.listen('MDCDialog:closed', this.handleMdcClosed);
         this.mdcDialog.listen('MDCDialog:closing', this.handleMdcClosing);
 
-        this.mdcDialog.scrimClickAction = this.closingActions.scrimClick
-            ? 'close'
-            : '';
-        this.mdcDialog.escapeKeyAction = this.closingActions.escapeKey
-            ? 'close'
-            : '';
+        this.mdcDialog.scrimClickAction = '';
+        if (this.closingActions.scrimClick) {
+            this.mdcDialog.scrimClickAction = 'close';
+        }
+
+        this.mdcDialog.escapeKeyAction = '';
+        if (this.closingActions.escapeKey) {
+            this.mdcDialog.escapeKeyAction = 'close';
+        }
     }
 
     public disconnectedCallback() {
--- a/src/components/input-field/examples/input-field-number.tsx
+++ b/src/components/input-field/examples/input-field-number.tsx
@@ -32,6 +32,11 @@ export class InputFieldNumberExample {
     }
 
     public render() {
+        let formatLabel = 'Format number';
+        if (this.formatNumber) {
+            formatLabel = 'Unformat number';
+        }
+
         return [
             <limel-input-field
                 label="Number Field Label"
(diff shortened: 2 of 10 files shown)


src/components/dialog/dialog.tsx
```diff
@@ -129,12 +129,13 @@ export class Dialog {
...
-        this.mdcDialog.escapeKeyAction = this.closingActions.escapeKey
-            ? 'close'
-            : '';
+        if (this.closingActions.scrimClick) {
+            this.mdcDialog.scrimClickAction = 'close';
+        }
+
+        if (this.closingActions.escapeKey) {
+            this.mdcDialog.escapeKeyAction = 'close';
+        }
```
adrianschmidt, line 138: Have you tested these? My guess would be that scrimClick and escapeKey defaults to `close` and we would need to set them to `''` to prevent that behaviour?

jgroth: Yeah I tested it and it did not appear to have any effect. Even setting it to `''` did not seem to do anything. There might have been some change in Material that affected this behavior so we should probably look into it

src/components/input-field/examples/input-field-number.tsx
```diff
@@ -32,6 +32,11 @@ export class InputFieldNumberExample {
     }
 
     public render() {
+        let formatLabel = 'Format number';
+        if (this.formatNumber) {
+            formatLabel = 'Unformat number';
+        }
```
adrianschmidt, line 38: Personally, I don't think this is any better than the ternary that was removed (I actually think this is ever so slightly worse, because it's removed from the context where it's used), but if this is the price we have pay for getting rid of some of the much worse ternary-uses, then I guess it's a price I'm willing to have us pay… 😅

jgroth: Yeah my thoughts exactly :sweat_smile: 
There might be some cases where it would make sense to have it on multiple lines, but as I said it's far too easy to abuse and we have some horrible examples of that :smile:

src/components/dialog/dialog.tsx
jgroth: I take that back! :smile:

src/components/portal/portal.tsx
```diff
@@ -172,10 +177,11 @@ export class Portal {
...
         if (this.inheritParentWidth) {
-            this.container.style.width =
-                rect.width > 0
-                    ? `${rect.width}px`
-                    : `${this.getContentWidth(this.container)}px`;
+            const containerWidth = this.getContentWidth(this.container);
+            this.container.style.width = `${containerWidth}px`;
+            if (rect.width > 0) {
+                this.container.style.width = `${rect.width}px`;
+            }
```
adrianschmidt, line 184: Hmm… I think this might trigger two DOM accesses if `rect.width > 0` is true…

I'd like to split out the addition of typings to a separate commit, and also rename that bloody  `rect` variable, so I'll add my suggestion in a fixup.

adrianschmidt (approved):
👍

adrianschmidt:
If you're happy with my additions, feel free to squash the fixup, and either merge or ping me and I'll approve again 😄

jgroth:
It looks good! :slightly_smiling_face:

adrianschmidt (approved):

Issue #1033, opened 2020-11-11
Lime elements fails to initialize in non-Chromium Edge


JohannesHansen:
## Current behavior

Lime elements fails to initialize in Edge. 

## Environment

- **lime-elements** version: 29.6.2
- Framework used: Vanilla.js
- Browser: Microsoft Edge 44.19041.423.0

adrianschmidt:
This is almost certainly because we updated to Stencil 2 a while back, and did not enable building of legacy js-formats, which was disabled by default in Stencil 2. We need to enable that again.

adrianschmidt:
Technically, there are two ways we can choose between:

1. Enable "legacy" build-output. This *should* just be a matter of changing two booleans in the build-config, and we obviously need to test that it worked as well. So it's a *little* work, but if it works as expected, it shouldn't be much.

2. Decide that lime-elements has the same very strict browser requirement that Lime CRM Web Client has, which means Lime Go also has to restrict their browser requirement. It also means that any outward facing site, like the eSign portal (which have to support a wider range of browsers because not all users are customers), will be unable to use lime-elements, or will have to manage using lime-elements for compatible browsers, and provide other solutions for non-compatible browsers.

I *strongly* consider option 2 to be the *wrong* way to go for lime-elements, but as we've agreed that new issues should go through triage, I'm laying out the options here 🙂

Ping @magnusfagerlund

JohannesHansen:
Note that this issue currently prevents Lime Go from using Lime Elements 29 and above.

magnusfagerlund:
Lets go with option 1. Maybe something for cool-down?

PR #1097, opened 2021-01-08
fix: support legacy browsers again

adrianschmidt:
Support for some older browsers was removed in lime-elements v29.0.0.
That support has now been added again.

fix: #1033

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/stencil.config.dist.ts
+++ b/stencil.config.dist.ts
@@ -19,6 +19,18 @@ export const config: Config = {
         },
     },
     plugins: [sass()],
+    buildEs5: 'prod',
+    extras: {
+        /* See docs at https://stenciljs.com/docs/config-extras */
+        appendChildSlotFix: true,
+        cloneNodeFix: true,
+        cssVarsShim: true,
+        dynamicImportShim: true,
+        safari10: true,
+        scriptDataOpts: false,
+        shadowDomShim: true,
+        slotChildNodesFix: true,
+    },
     tsconfig: './tsconfig.dist.json',
     globalStyle: 'src/global/core-styles.scss',
 };
--- a/stencil.config.dist.ts
+++ b/stencil.config.dist.ts
@@ -22,14 +22,11 @@ export const config: Config = {
     buildEs5: 'prod',
     extras: {
         /* See docs at https://stenciljs.com/docs/config-extras */
-        appendChildSlotFix: true,
-        cloneNodeFix: true,
         cssVarsShim: true,
         dynamicImportShim: true,
-        safari10: true,
-        scriptDataOpts: false,
+        scriptDataOpts: true,
         shadowDomShim: true,
-        slotChildNodesFix: true,
+        initializeNextTick: true,
     },
     tsconfig: './tsconfig.dist.json',
     globalStyle: 'src/global/core-styles.scss',
--- a/stencil.config.dist.ts
+++ b/stencil.config.dist.ts
@@ -26,7 +26,6 @@ export const config: Config = {
         dynamicImportShim: true,
         scriptDataOpts: true,
         shadowDomShim: true,
-        initializeNextTick: true,
     },
     tsconfig: './tsconfig.dist.json',
     globalStyle: 'src/global/core-styles.scss',
--- a/stencil.config.dist.ts
+++ b/stencil.config.dist.ts
@@ -22,9 +22,7 @@ export const config: Config = {
     buildEs5: 'prod',
     extras: {
         /* See docs at https://stenciljs.com/docs/config-extras */
-        cssVarsShim: true,
         dynamicImportShim: true,
-        scriptDataOpts: true,
         shadowDomShim: true,
     },
     tsconfig: './tsconfig.dist.json',
--- a/package-lock.json
+++ b/package-lock.json
@@ -982,6 +982,7 @@
         "jest-resolve": "^26.6.2",
         "jest-util": "^26.6.2",
         "jest-worker": "^26.6.2",
+        "node-notifier": "^8.0.0",
         "slash": "^3.0.0",
         "source-map": "^0.6.0",
         "string-length": "^4.0.1",
@@ -1769,6 +1770,7 @@
       "integrity": "sha512-41P5BCXPEIT0BlZCxAXR5mLsBmPTWqGs93rnwN2BBVxxAFtGouRk8H8nVx5wE2UTU3dbpXd2GnUGXKAXffn7RA==",
       "dev": true,
       "dependencies": {
+        "@commitlint/load": ">6.1.1",
         "chalk": "^2.4.1",
         "commitizen": "^4.0.3",
         "conventional-commit-types": "^3.0.0",
@@ -3862,6 +3864,7 @@
       "dependencies": {
         "anymatch": "~3.1.1",
         "braces": "~3.0.2",
+        "fsevents": "~2.1.2",
         "glob-parent": "~5.1.0",
         "is-binary-path": "~2.1.0",
         "is-glob": "~4.0.1",
@@ -4441,6 +4444,7 @@
       "integrity": "sha512-yAYxeGpVi27hqIilG1nh4A9Bnx4J3Ov+eXy4koL3drrR+IO9GaWPsKjik20ht608Asqi8TQPf0mczhEeyAtMzg==",
       "dev": true,
       "dependencies": {
+        "@commitlint/load": ">6.1.1",
         "chalk": "^2.4.1",
         "commitizen": "^4.0.3",
         "conventional-commit-types": "^3.0.0",
@@ -4944,7 +4948,8 @@
         "esprima": "^4.0.1",
         "estraverse": "^4.2.0",
         "esutils": "^2.0.2",
-        "optionator": "^0.8.1"
+        "optionator": "^0.8.1",
+        "source-map": "~0.6.1"
       },
       "bin": {
         "escodegen": "bin/escodegen.js",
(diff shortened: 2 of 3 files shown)


stencil.config.dist.ts
```diff
@@ -18,6 +18,18 @@ export const config: Config = {
...
+        /* See docs at https://stenciljs.com/docs/config-extras */
+        appendChildSlotFix: true,
+        cloneNodeFix: true,
+        cssVarsShim: true,
+        dynamicImportShim: true,
+        safari10: true,
+        scriptDataOpts: false,
+        shadowDomShim: true,
+        slotChildNodesFix: true,
+    },
```
adrianschmidt, line 32: This ought to add support for non-Chromium Edge and some other older browsers.

@JohannesHansen Can you try this branch in Lime Go, and see if it works?

adrianschmidt: For clarification, I've tried to test it via the documentation and via the Lime CRM web client, but both fail to load for reasons that seem to not be directly related to lime-elements. But since Lime Go already works in non-Chromium Edge, I think it makes sense to test there.

JohannesHansen: > @JohannesHansen Can you try this branch in Lime Go, and see if it works in a non-Chromium version of Edge?

@adrianschmidt I'll have a look!

JohannesHansen: @adrianschmidt Still not working unfortunately. I'm getting a fairly unhelpful `SCRIPT5022: SCRIPT5022: Syntax error` that might be to blame. I'll dig around a bit more.

adrianschmidt: I think there was one of the things they switched from default `true` to default `false` that I didn't switch back, because from the description it didn't seem to be anything that we would need… I'm not sure though. I'll have a look again.

adrianschmidt: > Still not working unfortunately. I'm getting a fairly unhelpful SCRIPT5022: SCRIPT5022: Syntax error suggesting a missing polyfill that might be to blame. I'll dig around a bit more.

Ah crap… that's the same problem I get when running the Kompendium app in Edge 18. I was hoping that was Kompendium not being compatible with Edge 18, but if it appears when this has been installed in Lime Go, then it's in lime-elements itself… 😞

But at least I then know that it's something I need to look into! 👍

adrianschmidt: > @adrianschmidt Still not working unfortunately. I'm getting a fairly unhelpful `SCRIPT5022: SCRIPT5022: Syntax error` suggesting a missing polyfill that might be to blame. I'll dig around a bit more.

Did you really test in Lime Go, or did you test by runnings the docs in this repo? Because the SCRIPT5022 error seemed to me to come from Kompendium, not from Lime Elements… 🤔

Anyway, I've now updated this branch a bit, and tested it with the following simple test page:

```html
<!DOCTYPE html>
<html dir="ltr" lang="en">

<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <meta http-equiv="X-UA-Compatible" content="ie=edge">
    <base href="/">
    <title>Testing Lime Elements</title>
</head>

<body class="mdc-typography" style="--badge-background-color: blue; --badge-text-color: yellow;">
    <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500" rel="stylesheet" />

    <link href="/node_modules/@limetech/lime-elements/dist/collection/style/color-palette-extended-light-mode-only.css" rel="stylesheet" />
    <link href="/node_modules/@limetech/lime-elements/dist/lime-elements/lime-elements.css" rel="stylesheet" />
    <script type="module" src="/node_modules/@limetech/lime-elements/dist/lime-elements/lime-elements.esm.js"></script>
    <script nomodule src="/node_modules/@limetech/lime-elements/dist/lime-elements/lime-elements.js"></script>
    <limel-spinner></limel-spinner>
    <limel-badge class="badge" label="999"></limel-badge>
    <limel-button label="Hello, World!" primary></limel-button>
</body>
</html>
```

It works in Edge 18, Edge 17, and Safari 11.1

I obviously can't say for absolute certain that there isn't some problem that will only appear when using a specific component, but loading lime-elements and showing a spinner or a button should work. If you still can't get it to work, let's look at it together over a screen share some time.

JohannesHansen: > Did you really test in Lime Go, or did you test by runnings the docs in this repo? Because the SCRIPT5022 error seemed to me to come from Kompendium, not from Lime Elements… 🤔

Yes, I linked it locally and built it. I'll try it again, this makes me suspect an issue with how we initialize the scripts.

adrianschmidt: Oh, that reminds me. Make sure you run the production build `npm run build`. If you run any of the commands that use the dev-build, it won't output the stuff needed for those older browsers.

JohannesHansen: @adrianschmidt it seems to work now! The app no longer crashes.

However there seems to be more issues with the components themselves; weird outlines and broken dropdowns were some obvious ones I encountered. 
image

Perhaps we should face the facts and officially drop support?

adrianschmidt: ### 😲

That looks like it's the css that isn't working properly, but I don't know what we might have changed that could cause things to look like that… 🤔

lime-elements v28.5.6 is the latest release before we removed the support for Edge 18. Do the components above work and look alright with that version?

JohannesHansen: > lime-elements v28.5.6 is the latest release before we removed the support for Edge 18. Do the components above work and look alright with that version?

The outlined input fields were introduced later so they work fine. The dropdowns are actually still broken in that version, wich is... curious. I'm fairly certain that is the same version we've been running for a while but I haven't noticed that issue before.

adrianschmidt: I just spent a few more hours on this.

I managed to load a temporary test-page with some components in Edge 18, and could check that `limel-input-field` rendered correctly. However, I also tried a date-picker, and the dropdown for that was unfortunately completely broken.

I then spent some time on trying to get the lime-elements docs working in Edge 18, but didn't really get anywhere. I think that supporting Edge 18 is just too much work for it to be worth the effort. It's not just that we need to put in the work right now, but we'll have to continuously test all our future changes in Edge 18.

If it's at all an option for Lime Go to drop the Edge 18 support, I think we should close this issue, and any products that need to support Edge 18, IE11, or any other legacy browsers, simply shouldn't use lime-elements.

Toughts?
pings:
@eketorp for general product strategy
@BregenzerK for input from the eSign portal (which is "public-facing", so should probably support legacy browsers, and not rely on lime-elements)

JohannesHansen: I agree. I've talked a bit to support about this as well, and they don't believe it's too problematic to just ask the users to upgrade/switch browser.

However, it could be worth considering adding some browser detection so we can display a meaningful error message to users - or should we leave that to the individual apps to implement?

adrianschmidt: I think that's the responsibility of each app.

But we should perhaps add some note to the docs, stating what we support.

BregenzerK: I think @justFredster is the better person to ask in terms of browser requirements in the e-sign portal. and just to double-check: Edge 18 is pretty outdated, right?? the latest is like 88 or something? I would be considered about completely removing Edge support, but I think it's reasonable that our customers have to update the browser versions

justFredster: I think you got this covered @BregenzerK . I think we can require a fairly late version of the browser you use. But we can't be to extreme about it. I would however like it if we made it more clear in the eSign documentation, which browsers the signing portal supports. Can you add that based on the outcome of this PR @BregenzerK ? Maybe even include the information in the error message shown if the browser is unsupported (I think we have something like that in the signing portal).

BregenzerK: created issue for it: https://github.com/Lundalogik/limepkg-esign/issues/651. bottom line e-sign should be fine with removing IE 11 and Edge 18 support.

adrianschmidt: Edge 18 was the last version of Edge using Microsoft's own rendering engine. It was first released in 2015. It was superseded by Edge 79, the first version of Edge built on Chromium, in January 2020. Wikipedia says that

> Microsoft plans to stop releasing security patches for Edge Legacy after March 9, 2021, and release a security update on April 13, 2021, which will replace Edge Legacy with Chromium-based Edge.

So, yeah, I guess it's more reasonable than I first thought to not support Edge 18, even in the public facing portal part of the e-sign add-on 👍

And it certainly feels like we should be able to require Edge 79 or newer in all of our products that are not "public facing", like Lime Go.

By "public facing", I mean that it is used, or potentially used, by someone who is not our customer. For Lime CRM and Lime Go, the only way you get access is to become a customer, and then we can simply say that "this software has these requirements". This in contrast to the e-sign portal, where someone who is a customer of ours may send a link to a third party, who may not even know that Lime exists, and that third party should still be able to sign the document. Which is why I have been of the opinion that the e-sign portal should support older browsers. But if Microsoft are end-of-life:ing Edge 18, and even plan to release a patch that updates Edge 18 to a Chromium based version, in just a few months, then I think we shouldn't worry too much about it.

PR #1099, opened 2021-01-13
fix(chip-set): emit startEdit event when text field gets focus

### LINKED ISSUE #1064: limel-picker: picker for multiple values closes the dropdown after the first value is selected, and it's difficult to open it again (quite a lot annoying, and _might_ prevent users from doing their work)
## Current behavior

In a picker for multiple values, after an item has been picked, the picker is focused, but the dropdown is closed. To open the dropdown again, you have to either type something, or blur the picker and then focus it again (either by clicking outside it and then on it again, or by tabbing away and then back to it).

1. In a limel-picker that is configured with `multiple=true`, select an item

- The picker stays focused, but the dropdown closes
- Clicking on the picker does not open the dropdown again, unless the picker is first blurred
- Pressing the down-arrow key or the enter key does nothing

## Expected behavior

For a picker that allows multiple items to be selected, the dropdown should probably just stay open. If we do actually want it to close, it should open again if the user clicks on the picker, without the need to blur the picker first. I think it should also open if the user presses the down-arrow key or the enter key on their keyboard.

## Environment

- **lime-elements** version: 29.8.0

hannahu:
fix #1064

This code was removed in this commit https://github.com/Lundalogik/lime-elements/commit/4e473a131dab790feb909ead0ed846795aa0ccf5

Is there any reason for us to consider that only one _startEdit_ event is sent?

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -201,6 +201,7 @@ export class ChipSet {
         }
 
         this.host.shadowRoot.querySelector('input').focus();
+        this.startEdit.emit();
     }
 
     /**


adrianschmidt:
> This code was removed in this commit 4e473a1
>
> Is there any reason for us to consider that only one startEdit event is sent?

@maltja and @jgroth wrote and reviewed that commit it looks like, so: Ping! 😊

jgroth:
I think there were multiple events being emitted when there should only have been one :thinking:

maltja:
> I think there were multiple events being emitted when there should only have been one 🤔

Yes, I also think that was why I did like that. And I don't think that I knew then that you should be able to pick multiple values in the dropdown menu. 😅

adrianschmidt:
I think it makes sense that we should emit another `startEdit` event if we are already in "edit mode".

But it seems we're a little bit out of synk between the chip-set and the picker, since the picker closes the menu and seems to think we're no longer in edit mode, while the chip-set obviously still considers itself to be in edit mode…

I'm not sure if maybe one should be changed to leave or stay in edit mode, or if we should de-couple the opening of the menu from the `startEdit` event, and listen to some other event (or something) for that?

src/components/chip-set/chip-set.tsx
```diff
@@ -401,10 +401,6 @@ export class ChipSet {
      * @returns {void}
      */
     private handleTextFieldFocus() {
-        if (this.editMode) {
-            return;
-        }
-
```
jgroth, line 407: This reintroduces the bug that was solved in #985

I had a look through the code, and it seems like we are not emitting `startEdit` when we explicitly set the focus through the `setFocus` method, maybe that will solve this? E.g.

```diff
    @Method()
    public async setFocus(emptyInput: boolean = false) {
        this.editMode = true;
        if (emptyInput) {
            this.textValue = '';
        }

        this.host.shadowRoot.querySelector('input').focus();
+       this.startEdit.emit();
    }
```

PR #1100, opened 2021-01-14
Picker clear all

### LINKED ISSUE #1062: limel-picker: Clicking the button to remove all chips when the dropdown is open breaks the picker
## Current behavior

Clicking the x icon to the right (is it even needed at all?) makes it impossible to get the drop down to close again by clicking outside the drop down.

97188322-acbb5b80-17a3-11eb-9f9e-c8604dd008bb

1. In a limel-picker that is configured with `multiple=true`, select one or more items
2. Open the dropdown again (you may have to click outside first, then on the picker again, this is a separate bug: #1064)
3. Hover the remove-button on the right edge of the picker, note that it does indicate that it can be clicked:
   image
   Important! Use the remove-button for removing all chips at once, _not_ the one on the chip itself (see separate bug concerning that one: #1063)
4. Click on the button

- The chips remain in the picker
- The picker is blurred, but the dropdown remains open
- Clicking outside the picker does not close the dropdown
- Clicking on the input does not focus the picker again
- Clicking on the remove button of a chip does nothing
- Clicking on the remove-all button does nothing
- At this point, there is only one way to close the picker: you have to select an item in the dropdown, once you do, the picker works as normal again

## Expected behavior

- All chips should be removed
- Ideally, if the chips are removed when the picker is open, it should stay open, and if the chips are removed when the picker is closed, it should probably stay closed… if that can't be achieved, I'm not sure which is better, always closing it or always opening it…? /@adrianschmidt

## Environment

- **lime-elements** version: 29.7.1

hannahu:
fix #1062

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu-surface/menu-surface.tsx
+++ b/src/components/menu-surface/menu-surface.tsx
@@ -119,34 +119,12 @@ export class MenuSurface {
     }
 
     private preventClickEventPropagation() {
-        // When the menu surface is open, we want to stop the `click` event from propagating
-        // when clicking outside the surface itself. This is to prevent any dialog that might
-        // be open from closing, etc. However, when dragging a scrollbar no `click` event is emitted,
-        // only mousedown and mouseup. So we listen for `mousedown` and attach a one-time listener
-        // for `click`, so we can capture and "kill" it.
-        document.addEventListener('click', this.stopEvent, {
-            capture: true,
-            once: true,
-        });
-        // We also capture and "kill" the next `mouseup` event.
+        // When the menu surface is open, we want to stop the `mouseup` event from propagating
+        // when for example dragging a scrollbar. So we capture and "kill" the `mouseup` event.
         document.addEventListener('mouseup', this.stopEvent, {
             capture: true,
             once: true,
         });
-        // If the user dragged the scrollbar, no `click` event happens. So when we get the
-        // `mouseup` event, remove the handler for `click` if it's still there.
-        // Otherwise, we would catch the next click even though the menu is no longer open.
-        document.addEventListener(
-            'mouseup',
-            () => {
-                document.removeEventListener('click', this.stopEvent, {
-                    capture: true,
-                });
-            },
-            {
-                once: true,
-            }
-        );
     }
 
     private stopEvent(event) {


jgroth (commented):
The fix works, but it's a little bit unintuitive... Normally we would use click instead of mousdown for clickable actions, and also that this fix is done in a separate component than where the bug is :thinking:

hannahu:
> The fix works, but it's a little bit unintuitive... Normally we would use click instead of mousdown for clickable actions, and also that this fix is done in a separate component than where the bug is

There's now a solution in the menu-surface and the other stuff is removed hihi.
To be able to interact with components belonging to the menu-surface (in this case the picker), the only way I see is to remove the preventing of click events on the document. The comment said the clicks were prevented here to prevent dialogs from closing when clicking on their close button or similar, but isn't that the wrong place to prevent it? We can't know what click events to stop here and not right?

jgroth:
This will change the behavior when you have a select/picker in a dialog and click outside the dialog when dismissing a menu surface. Previously we would not close the dialog, with this change the dialog will be closed which I think will be quite annoying

jgroth:
So I guess we have to decide which behavior is more annoying :sweat_smile: Not sure if we can both have the cake and eat it :thinking: :cake:

adrianschmidt:
> This will change the behavior when you have a select/picker in a dialog and click outside the dialog when dismissing a menu surface. Previously we would not close the dialog, with this change the dialog will be closed which I think will be quite annoying

Yeah, I think that's quite a lot worse than the picker bug…

I haven't dug into this one at all, but maybe I can take a look and see if I can come up with any ideas?

hannahu:
> > This will change the behavior when you have a select/picker in a dialog and click outside the dialog when dismissing a menu surface. Previously we would not close the dialog, with this change the dialog will be closed which I think will be quite annoying
> 
> Yeah, I think that's quite a lot worse than the picker bug…
> 
> I haven't dug into this one at all, but maybe I can take a look and see if I can come up with any ideas?

I'll have a chat with Kia later, I can ask you if I need more inputs :) thanks!

PR #1102, opened 2021-01-18
fix(tab-panel): emit resize event to trigger component re-layout when changing tabs

adrianschmidt:
Content inside the newly activated tab may need to redraw once visible, so we use the resize event
trick.

fix: Lundalogik/crm-feature#1696
fix: Lundalogik/crm-feature#1697

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [x] Firefox
- [x] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tab-panel/tab-panel.tsx
+++ b/src/components/tab-panel/tab-panel.tsx
@@ -9,6 +9,7 @@ import {
     Watch,
 } from '@stencil/core';
 import { Tab } from '../tab-bar/tab.types';
+import { dispatchResizeEvent } from '../../util/dispatch-resize-event';
 
 /**
  * The `limel-tab-panel` component uses the `limel-tab-bar` component together
@@ -133,6 +134,10 @@ export class TabPanel {
         });
 
         this.setTabStatus(event.detail);
+
+        // Content inside the newly activated tab may need to redraw once
+        // visible, so we use the resize event trick. /Ads
+        setTimeout(dispatchResizeEvent);
     }
 
     private getSlot(): HTMLSlotElement {


src/components/tab-panel/tab-panel.tsx
```diff
@@ -133,6 +134,12 @@ export class TabPanel {
         });
 
         this.setTabStatus(event.detail);
+
+        // Content inside the newly activated tab may need to redraw once
+        // visible, so we use the resize event trick. /Ads
+        setTimeout(() => {
+            dispatchResizeEvent();
```
adrianschmidt, line 141: Note to reviewer: Before trying this fix, I was able to very consistently replicate both the issue with the slider and the issue with missing labels, by loading a deal, then going to dashboard, and then using the browser's back and forward buttons to step back and forth to the deal card again.

When I tried that again after adding this fix, I was not able to reproduce. However, I did try removing the fix again, just to double check, and while I _was_ able to reproduce the bug, it only happened on something like 1 out of 10 tries, so not at all consistently. Thus, I'm not sure this solves the bug 100 % of the time, but I'm hoping it does.

jgroth: You can try the `requestAnimationFrame` function instead. It will run in the next paint cycle instead of as soon as possible, so it might affect this since it seems like it's connected to stuff being rendered

adrianschmidt: Nice suggestion! I'm wondering if that might not send the event too early, before the panel has become visible. But I'll certainly try it, and thanks for pointing me to it regardless, as I'm sure that function will come in very handy in other places too! 🙏

adrianschmidt: `requestAnimationFrame` seems to work perfectly fine 👍

⚡

jgroth: Nice! :slightly_smiling_face:

maltja: The requestAnimationFrame doesn't work when I tested it. However, it worked before the fixup for me (I might have broken it even with the first fix but I couldn't reproduce it so I guess the risk is very small). What I did to test it: 
1. Go to a deal.
2. Change tab to History.
3. Resize the window. 
4. Go back to the overview -> The slider is broken.

adrianschmidt: Ah! Yeah, that's a great way to consistently reproduce it 👍

I'll have another go at fixing it!

adrianschmidt: I changed it back to setTimeout, and that works in your test case too Malin 👍

⚡

maltja (dismissed):

adrianschmidt:
Approved by maltja above. I just squashed the fixups.

PR #1104, opened 2021-01-19
add a support for having action buttons inside the column header

Befkadu1:
fix: Lundalogik/crm-feature#1749

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/columns.ts
+++ b/src/components/table/columns.ts
@@ -21,6 +21,10 @@ export class ColumnDefinitionFactory {
             field: column.field,
         };
 
+        if (column.headerComponent) {
+            definition.titleFormatter = formatHeader(column);
+        }
+
         if (column.component?.name || column.formatter) {
             definition.formatter = createFormatter(column, this.pool);
             definition.formatterParams = column as object;
@@ -33,6 +37,39 @@ export class ColumnDefinitionFactory {
         return definition;
     }
 }
+
+/**
+ * Formats the header of the column
+ *
+ * @param {Column} column the configuration for the column
+ *
+ * @returns {string | HTMLElement} custom component that renders a column header
+ */
+export const formatHeader = (column: Column) => (): string | HTMLElement => {
+    const element = document.createElement('div');
+    element.setAttribute('class', 'lime-col-title__custom-component');
+    const titleText = document.createElement('span');
+    titleText.setAttribute('class', 'title-component-text');
+    titleText.innerText = column.title;
+
+    const header = document.createElement(column.headerComponent.name);
+    header.setAttribute('class', 'title-component-slot');
+    let props: object = column.headerComponent.props || {};
+    if (column.headerComponent.propsFactory) {
+        props = {
+            ...props,
+            // we pass null to propsFactory function because no data in column header
+            ...column.headerComponent.propsFactory(null),
+        };
+    }
+
+    Object.assign(header, props);
+    element.appendChild(titleText);
+
+    element.appendChild(header);
+
+    return element;
+};
 
 /**
  * Create a formatter to be used to format values in a column
--- a/src/components/table/examples/header-menu.tsx
+++ b/src/components/table/examples/header-menu.tsx
@@ -1,0 +1,24 @@
+import { Component, h, Prop } from '@stencil/core';
+import { ListItem } from 'src/components/list/list-item.types';
+
+@Component({
+    tag: 'limel-example-header-menu',
+    shadow: true,
+})
+export class HeaderMenu {
+    @Prop()
+    public items: ListItem[];
+
+    @Prop()
+    public icon: string;
+
+    private menuOpen = true;
+
+    public render() {
+        return (
+            <limel-menu items={this.items} open={this.menuOpen}>
+                <limel-icon slot="trigger" name={this.icon} size="x-small" />
+            </limel-menu>
+        );
+    }
+}
--- a/src/components/table/examples/table-header-menu.tsx
+++ b/src/components/table/examples/table-header-menu.tsx
@@ -1,0 +1,73 @@
+import { Component, h } from '@stencil/core';
+import { Column } from '../table.types';
+import { persons, Person } from './persons';
+
+/**
+ * Column header menu
+ *
+ * @link persons.ts
+ * @link header-menu.tsx
+ */
+@Component({
+    tag: 'limel-example-table-header-menu',
+    styleUrl: 'table.scss',
+    shadow: true,
+})
+export class TableExampleHeadermenu {
+    private tableData: Person[] = persons;
+    private columns: Column[] = [
+        {
+            title: 'Name',
+            field: 'name',
+            headerComponent: {
+                name: 'limel-example-header-menu',
+                props: {
+                    icon: 'menu',
+                    items: [
+                        { text: 'All' },
+                        { text: 'Me' },
+                        { text: 'Contains' },
+                        { text: 'Equals' },
+                        { text: 'Not equals' },
+                    ],
+                },
+            },
+        },
+        {
+            title: 'Age',
+            field: 'age',
+        },
+        {
+            title: 'Kind',
+            field: 'kind',
+        },
+        {
+            title: 'Height',
+            field: 'height',
+        },
+        {
+            title: 'Stamina',
+            field: 'stamina',
+        },
+        {
+            title: 'Place of Birth',
+            field: 'placeOfBirth',
+        },
+        {
+            title: 'Sign',
+            field: 'sign',
+        },
+        {
+            title: 'Date of Birth',
+            field: 'dateOfBirth',
+        },
+        {
+            title: 'Role',
+            field: 'role',
+        },
+    ];
+
+    public render() {
+        return [<limel-table data={this.tableData} columns={this.columns} />];
+    }
+}
--- a/src/components/table/table.types.ts
+++ b/src/components/table/table.types.ts
@@ -23,6 +23,11 @@ export interface Column<T extends object = any> {
      * Type of aggregator to use for the column
      */
     aggregator?: ColumnAggregatorType | ColumnAggregatorFunction<T>;
+
+    /**
+     * A component used to render inside the column header
+     */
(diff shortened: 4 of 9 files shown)


src/components/table/columns.ts
```diff
@@ -34,6 +38,27 @@ export class ColumnDefinitionFactory {
...
 
+export const formatTitle = (column: Column) => (): string | HTMLElement => {
+    const element = document.createElement('div');
+    element.innerText = column.title;
+
+    const icon = document.createElement('limel-icon');
+    icon.setAttribute('name', column.menuIcon);
+    icon.setAttribute('size', 'x-small');
+    icon.setAttribute('class', 'tabulator-header-menu-button');
+    icon.setAttribute('style', 'padding: 0 0 0 5px;');
```
Befkadu1, line 49: The styling part will be handled by @Kiarokh

src/components/table/columns.ts
```diff
@@ -34,6 +38,27 @@ export class ColumnDefinitionFactory {
     }
 }
 
+export const formatTitle = (column: Column) => (): string | HTMLElement => {
+    const element = document.createElement('div');
+    element.innerText = column.title;
+
+    const icon = document.createElement('limel-icon');
+    icon.setAttribute('name', column.menuIcon);
+    icon.setAttribute('size', 'x-small');
```
Befkadu1, line 47: I am not sure if we set the icon size here

src/components/table/columns.ts
```diff
@@ -34,6 +38,27 @@ export class ColumnDefinitionFactory {
...
+    icon.setAttribute('style', 'padding: 0 0 0 5px;');
+
+    element.appendChild(icon);
+    icon.addEventListener('click', headerClickListener);
+
+    return element;
+};
+
+function headerClickListener(event: MouseEvent | KeyboardEvent) {
+    event.stopPropagation();
```
Befkadu1, line 58: We will configure the event when we get this https://github.com/Lundalogik/crm-feature/issues/1748 done

src/components/table/columns.ts
jgroth: We should probably do it in the CSS instead of inline :slightly_smiling_face:

src/components/table/columns.ts
jgroth: How would the consumer set the size with the implementation right now? :slightly_smiling_face: It would be great with an example!
(On a side note I'm not sure the consumer should be allowed to change the size, but that's another discussion)

src/components/table/columns.ts
jgroth: I don't think we need to wait for that PR in order to emit an event. They should be independent. A consumer _might_ use the popover to display something, but they might also do something completely different

src/components/table/examples/table-header-menu.tsx
```diff
@@ -0,0 +1,38 @@
...
+ * Header menu
+ *
+ * @link persons.ts
+ */
+@Component({
+    tag: 'limel-example-table-header-menu',
+    styleUrl: 'table.scss',
+    shadow: true,
+})
+export class TableExampleHeadermenu {
```
jgroth, line 15: It would be great to display something when the menu icon is clicked in this example :slightly_smiling_face:

src/components/table/table.types.ts
```diff
@@ -23,6 +23,11 @@ export interface Column<T extends object = any> {
      * Type of aggregator to use for the column
      */
     aggregator?: ColumnAggregatorType | ColumnAggregatorFunction<T>;
+
+    /**
+     * Action menu icon used to display a menu component when clicked
```
jgroth, line 28: This is not 100% true, a consumer may choose what they want to do when the icon is clicked. Maybe we should rename this property to something else to indicate that :thinking: What do you think? Maybe `headerIcon` or something similar is better?

src/components/table/columns.ts
Befkadu1: I was thinking if we can add a property called `headerIconSIze` to the column interface

Kiarokh: No we should totally set the size ourselves! It's 100% ours.

src/components/table/columns.ts
Kiarokh: yes, I'll put all the styles in a CSS file. You can remove the inline styles. Put them in the respective file for now just as a reference, and I'll tweak them.

src/components/table/examples/table-header-menu.tsx
```diff
@@ -0,0 +1,173 @@
...
+                        { text: 'Equals' },
+                        { text: 'Not equals' },
+                    ],
+                },
+            },
+        },
+        {
+            title: 'Age',
+            field: 'age',
+            headerMenu: {
```
jgroth, line 38: Maybe we do not need a menu component for each column in this example. Since the menu items does not do anything at the moment, this whole example is mostly a big chunk of duplicated menu items that does not add much to the reader. What do you think?

src/components/table/examples/header-menu.tsx
```diff
@@ -0,0 +1,25 @@
+import { Component, h, Prop } from '@stencil/core';
+import { TableComponent } from '../table.types';
+
+@Component({
+    tag: 'limel-example-header-menu',
+    shadow: true,
+})
+export class HeaderMenu implements TableComponent {
+    @Prop()
+    public data: any;
```
jgroth, line 10: We should not use `any` here. It makes it very hard for the reader to understand what it is used for

src/components/table/columns.ts
```diff
@@ -34,6 +38,27 @@ export class ColumnDefinitionFactory {
...
+export const formatTitle = (column: Column) => (): string | HTMLElement => {
+    const element = document.createElement('div');
+    element.innerText = column.title;
+    if (!column.headerMenu.component) {
+        return;
+    }
+
+    const menu = document.createElement(column.headerMenu.component.name);
+    menu.setAttribute('class', 'tabulator-header-menu-button');
+    menu.data = column.headerMenu.component;
```
jgroth, line 56: What is this doing? Will every component have a prop named `data`?

jgroth (changes requested):

src/components/table/columns.ts
```diff
@@ -34,6 +39,32 @@ export class ColumnDefinitionFactory {
...
+    if (!column.headerComponent) {
+        return;
+    }
+
+    const menu = document.createElement(column.headerComponent.name);
+    menu.setAttribute('class', 'tabulator-header-menu-button');
+
+    const props = column.headerComponent.props;
+
+    forIn(props, (value, key) => (menu[key] = value));
```
jgroth, line 61: We should not have to depend on an external library for this. The following will have the same behaviour
```suggestion
    Object.assign(menu, props);
```

src/components/table/columns.ts
```diff
@@ -34,6 +39,32 @@ export class ColumnDefinitionFactory {
...
+ * @param {Column} column the configuration for the column
+ *
+ * @returns {string | HTMLElement} custom component that renders a column header
+ */
+export const formatHeader = (column: Column) => (): string | HTMLElement => {
+    const element = document.createElement('div');
+    element.innerText = column.title;
+    if (!column.headerComponent) {
+        return;
+    }
```
jgroth, line 54: Do we need this check? :thinking: 
We are already checking if the column has specified a header component above in the `create` function

src/components/table/columns.ts
```diff
@@ -34,6 +39,32 @@ export class ColumnDefinitionFactory {
...
+
+    const menu = document.createElement(column.headerComponent.name);
```
jgroth, line 56: I don't think we should call this `menu` :slightly_smiling_face:

src/components/table/columns.ts
```diff
@@ -34,6 +39,32 @@ export class ColumnDefinitionFactory {
...
+    menu.setAttribute('class', 'tabulator-header-menu-button');
```
jgroth, line 57: What is this class used for? We could probably use a better name for it as well. What do you think? :slightly_smiling_face:

src/components/table/columns.ts
```diff
@@ -34,6 +39,32 @@ export class ColumnDefinitionFactory {
...
+
+    const props = column.headerComponent.props;
+
+    forIn(props, (value, key) => (menu[key] = value));
+
+    element.appendChild(menu);
+
+    return element;
```
jgroth, line 65: According to the interface, `headerComponent` is of the type `TableComponentDefinition`
However, `TableComponentDefinition` has a property called `propsFactory`. That is not taken care of in this function right now. Can we get that to work as well? :slightly_smiling_face:

Befkadu1: Is this to use for example `context`?

jgroth: Yes! :slightly_smiling_face:

src/components/table/columns.ts
Befkadu1: This would be good to have it for styling

jgroth: But who is using it? We do not know what component the user is adding, and they can style their component on their own :thinking:

src/components/table/columns.ts
```diff
@@ -49,18 +48,20 @@ export class ColumnDefinitionFactory {
...
-    const menu = document.createElement(column.headerComponent.name);
-    menu.setAttribute('class', 'tabulator-header-menu-button');
-
-    const props = column.headerComponent.props;
+    const header = document.createElement(column.headerComponent.name);
+    let props: object = column.headerComponent.props || {};
+    if (column.headerComponent.propsFactory) {
+        props = {
+            ...props,
+            // we pass null to propsFactory function because no data in column header
```
jgroth, line 57: We should clarify this in the interface documentation as well, that when the factory is used for header components there will be no data available

jgroth (approved):

Issue #1071, opened 2020-12-10
limel-input-field: refactor to make more DRY 🙂


adrianschmidt:
Now that the structure of all input fields are very similar, I think we can refactor this and `renderTextArea()`, so that we don't repeat so much code.

_Originally posted by @adrianschmidt in https://github.com/Lundalogik/lime-elements/pull/1057#discussion_r539323592_

PR #1108, opened 2021-01-21
input-field: fix small bug, add tests, refactor, add small feature to input-field

adrianschmidt:
- fix(input-field): correctly link the label to the input or textarea element
- test(input-field): add tests for input-field
- refactor(input-field): duplicate less code
- refactor(input-field): add some missing typings
- feat(input-field): add `type='urlAsText'`, allowing urls that are not fully qualified

fix: #1071
fix: Lundalogik/crm-feature#1595

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.e2e.ts
+++ b/src/components/input-field/input-field.e2e.ts
@@ -1,0 +1,248 @@
+import { E2EElement, E2EPage, newE2EPage } from '@stencil/core/testing';
+import { InputType } from './input-field.types';
+
+describe('limel-input-field', () => {
+    let page: E2EPage;
+    let limelInput: E2EElement;
+    let inputContainer: E2EElement;
+    let label: E2EElement;
+    let outline: E2EElement;
+
+    const types: Array<{ name: InputType; [key: string]: any }> = [
+        { name: 'email' },
+        { name: 'number' },
+        { name: 'password' },
+        { name: 'search' },
+        { name: 'tel' },
+        { name: 'text' },
+        { name: 'textarea', nativeSelector: 'limel-input-field>>>textarea' },
+        { name: 'url' },
+    ];
+
+    types.forEach((type) =>
+        describe(`with type="${type.name}"`, () => {
+            const expectedInputId = 'tf-input-element';
+            beforeEach(async () => {
+                page = await createPage(`
+                    <limel-input-field
+                        type="${type.name}"
+                        label="Test"
+                    ></limel-input-field>
+                `);
+                limelInput = await page.find('limel-input-field');
+                inputContainer = await page.find(
+                    'limel-input-field>>>div.mdc-text-field'
+                );
+            });
+            if (type.name === 'textarea') {
+                it('uses the `textarea` style', () => {
+                    expect(inputContainer).toHaveClass(
+                        'mdc-text-field--textarea'
+                    );
+                });
+            } else {
+                it('uses the `outlined` style', () => {
+                    expect(inputContainer).toHaveClass(
+                        'mdc-text-field--outlined'
+                    );
+                });
+            }
+
+            it('is NOT considered invalid', () => {
+                expect(inputContainer).not.toHaveClass(
+                    'mdc-text-field--invalid'
+                );
+            });
+            it('is NOT disabled', () => {
+                expect(inputContainer).not.toHaveClass(
+                    'mdc-text-field--disabled'
+                );
+            });
+            it('is NOT required', () => {
+                expect(inputContainer).not.toHaveClass(
+                    'mdc-text-field--required'
+                );
+            });
+            it('does NOT have a leading icon', () => {
+                expect(inputContainer).not.toHaveClass(
+                    'mdc-text-field--with-leading-icon'
+                );
+            });
+            it('does NOT have a trailing icon', () => {
+                expect(inputContainer).not.toHaveClass(
+                    'mdc-text-field--with-trailing-icon'
+                );
+            });
+            describe('the native input', () => {
+                let nativeInput: E2EElement;
+
+                beforeEach(async () => {
+                    const selector =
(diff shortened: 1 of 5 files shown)


src/components/input-field/input-field.tsx
```diff
@@ -443,11 +445,27 @@ export class InputField {
...
+        if (this.type === 'url' && this.isUrlInvalid(this.value)) {
+            return true;
+        }
+
         const element = this.getInputElement();
 
         return !(element && element.checkValidity());
     }
 
+    private isUrlInvalid(url: string) {
```
jgroth, line 457: I'm not doing a full review of this, just leaving some thoughts here :slightly_smiling_face: 

It feels a bit unfortunate that we have to remove the native validation of URL fields and that this now does not have the same functionality as before, only because we are having issues with it in one place in the webclient. Perhaps it would be possible to solve the underlying problem behind this issue from crm-components instead by always passing `invalid = false`, basically turning off validation for the problematic fields? :thinking:

jgroth: We could also add a new `type`, e.g. `lime-url`, `relative-url` or something like that :thinking:

adrianschmidt: I don't think we can, firstly because I don't think always passing in `false` will make any difference (we will treat the field as invalid if _either_ the input or the native check says it's invalid), and secondly because I already tried using `<input type="url">` and doing the following on lines 448-450:
```
        if (this.type === 'url') {
            return this.isUrlInvalid(this.value);
        }
```
That completely bypasses the native validation, and does make the component consider the field valid (showing the correct icon and not blocking saving of the form), but the underlying MDC component still sees the failed native validation and styles the input with a red border and does a little shake animation of the label when blurring the input. So I don't think we're getting around that 😞

The main reason I tried that is that I would still like mobile devices to show the appropriate keyboard for url inputs, rather than the generic keyboard. I guess we could make it possible to opt-out of the native validation for the url type (still by using `text` instead of `url` as the native type, but then at least you still get validation by default).

Thoughts?

adrianschmidt: > We could also add a new `type`, e.g. `lime-url`, `relative-url` or something like that :thinking: 

Ah, yeah, that's a great idea! While a bit wordy, it could be `non-validated-url`…

jgroth: Hmm ok, I thought that the passed `invalid` always took precedence over anything else :thinking: Maybe it only does if you set it to `true`? :smile:

adrianschmidt: > Maybe it only does if you set it to `true`? :smile: 

Yes, exactly 🙂

adrianschmidt: ⚡

maltja (approved):
Looks good, I gave some small comments but nothing that needs to be fixed 🙂

src/components/input-field/input-field.e2e.ts
```diff
@@ -0,0 +1,379 @@
...
+                    );
+                    expect(icon).toEqualAttribute('name', 'external_link');
+                });
+            })
+        );
+
+        const invalidUrls = [
+            { input: 'justoneword' },
+            { input: 'some words with spaces' },
+            // { input: 'an.email.address@example.com' }
```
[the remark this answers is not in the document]
adrianschmidt: Ah, that's just cruft I forgot to remove. I wanted to test that it didn't accept an email address, but then realised that I would almost certainly not be able to make it _not_ incorrectly treat certain url:s as invalid because they contain an email address or something… So I decided to keep it simple.

src/components/input-field/input-field.tsx
```diff
@@ -458,11 +456,27 @@ export class InputField {
...
+        if (!url) {
+            return false;
+        }
+
+        if (url.indexOf('.') === -1 && url.indexOf('/') !== 0) {
```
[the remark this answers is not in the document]
adrianschmidt: Ah, yeah… All these newfangled functions all the kids use nowadays… 😝

Definitely better! I'll update it to match the current millennium 👍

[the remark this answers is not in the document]
adrianschmidt: ⚡

src/components/input-field/input-field.tsx
```diff
@@ -257,47 +266,53 @@ export class InputField {
...
+        properties.class = 'mdc-text-field__input';
+        properties.onInput = this.handleChange;
+        properties.onFocus = this.onFocus;
+        properties.onBlur = this.onBlur;
+        properties.required = this.required;
+        properties.disabled = this.disabled;
+
+        const forceInvalid = this.type === 'urlAsText' && this.isInvalid();
+
+        const containerClassList = {
```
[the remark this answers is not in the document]
adrianschmidt: ⚡

src/components/input-field/input-field.e2e.ts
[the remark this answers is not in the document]
adrianschmidt: Removed!

⚡

src/components/input-field/input-field.tsx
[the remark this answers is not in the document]
adrianschmidt: I had changed `url.indexOf('.') === -1` to `url.includes('.')` instead of `!url.includes('.')`, which caused the tests to fail. I've now pushed a fix for that 🙂

maltja (approved):

maltja (approved):

PR #1109, opened 2021-01-22
Popover component

filipharry:
fix: https://github.com/Lundalogik/crm-feature/issues/1748

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/popover-surface/popover-surface.scss
+++ b/src/components/popover-surface/popover-surface.scss
@@ -1,0 +1,47 @@
+@import '../../style/functions';
+
+:host {
+    --popover-border-radius: 0.75rem;
+}
+
+.limel-popover-surface {
+    position: relative;
+    display: flex;
+    flex-direction: column;
+
+    width: var(--popover-surface-width, auto);
+    max-height: calc(100vh - 2rem);
+    max-width: calc(100vw - 2rem);
+
+    border-radius: var(--popover-border-radius);
+    box-shadow: var(--shadow-depth-16);
+    background-color: var(
+        --popover-body-background-color,
+        rgb(var(--contrast-100))
+    );
+
+    slot {
+        box-sizing: border-box;
+        display: block;
+        width: 100%;
+    }
+
+    slot[name='header'] {
+        border-radius: var(--popover-border-radius) var(--popover-border-radius)
+            0 0;
+        background-color: var(
+            --popover-header-background-color,
+            rgb(var(--contrast-300))
+        );
+    }
+    slot[name='body'] {
+    }
+    slot[name='footer'] {
+        border-radius: 0 0 var(--popover-border-radius)
+            var(--popover-border-radius);
+        background-color: var(
+            --popover-footer-background-color,
+            rgb(var(--contrast-500))
+        );
+    }
+}
--- a/src/components/popover-surface/popover-surface.tsx
+++ b/src/components/popover-surface/popover-surface.tsx
@@ -1,0 +1,69 @@
+import { Component, h, Prop, Element } from '@stencil/core';
+
+/**
+ * @slot - Content to put inside the surface
+ * @private
+ */
+
+@Component({
+    tag: 'limel-popover-surface',
+    shadow: true,
+    styleUrl: 'popover-surface.scss',
+})
+export class PopoverSurface {
+    /**
+     * True if the popover surface is open, false otherwise
+     */
+    @Prop()
+    public open = false;
+
+    /**
+     * Portal Id
+     */
+    @Prop()
+    public portalId: string;
+
+    /**
+     * Content to render
+     */
+    @Prop()
+    public contentCollection: HTMLCollection;
+
+    @Element()
+    private host: HTMLLimelPopoverSurfaceElement;
+
+    public componentDidLoad() {
+        this.appendElement();
+    }
+
+    public render() {
+        const classList = {
+            'limel-popover-surface': true,
+            'limel-popover-surface--open': this.open,
+        };
+
+        return (
+            <div class={classList} tabindex="-1">
+                <slot name="header" />
+                <slot name="body" />
+                <slot name="footer" />
+            </div>
+        );
+    }
+
+    private appendElement() {
+        const portalSlots = this.host.shadowRoot.querySelectorAll('slot');
+        Array.from(this.contentCollection).forEach((child) => {
+            if (child instanceof HTMLElement) {
+                const slot = child.getAttribute('slot');
+                const element = Array.from(portalSlots).find(
+                    (el) => el.name === slot
+                );
+
+                if (element) {
+                    element.appendChild(child);
+                }
+            }
+        });
+    }
+}
--- a/src/components/popover/popover.tsx
+++ b/src/components/popover/popover.tsx
@@ -1,0 +1,158 @@
+import {
+    Component,
+    h,
+    Prop,
+    Element,
+    Event,
+    EventEmitter,
+    Watch,
+} from '@stencil/core';
+import { createRandomString } from '../../util/random-string';
+import { zipObject } from 'lodash-es';
+
+/**
+ * A popover is an impermanent layer that is displayed on top of other content
+ * when user taps an element that triggers the popover. This element can be
+ * practically anything, a button, piece of text, and icon, etc.
+ *
+ * Popovers are nonmodal, thus they don't have a semi-transparent backdrop layer
+ * that darkens the content below them. Also for the same reason, they can be
+ * dismissed by tapping or clicking another part of the screen, but not by a
+ * clicking a button or link on the popover itself.
+ *
+ * :::important
+ * The component is emitting a close event when you click outside its container.
+ * However, it’s up to you as consumer to react properly on this event and
+ * hide the component (preferably by setting the property `open` to `false`).
+ * :::
+ *
+ * ## Usage
+ * Use a popover to show **options** or **information** related to the trigger
+ * onscreen. A typical use case for popovers is a tooltip, in which you show
+ * help text or contextual information to users.
+ *
+ * Popovers are most appropriate on larger screens but can be used on smaller
+ * screens too, as long as their content is responsive and takes into account
+ * the context they are displayed on.
+ *
+ * When a popover is displayed, interactions with other controls are blocked,
+ * until user dismisses the popover.
+ *
+ *
+ * ## Layout
+ * Popovers have 3 slots, `header`, `body`, and `footer`; in which you can
+ * import custom web-components. You can choose to only use `body`, or only
+ * `header` and `body` for example.
+ *
+ * :::note
+ * You must make sure that web-components that you import into these slots have
+ * a `width: 100%;` style and horizontally stretch across the popover.
+ *
+ * However, `width` of the popover can be controlled by specifying a CSS variable
+ * of `--popover-surface-width`. If you don't specify any width, the popover
+ * will grow as wide as its content.
+ *
+ * There are other style variables as well for changing the default background
+ * colors of `header`, `body`, and `footer`.
+ * :::
+ *
+ * :::important
+ * Do not make a popover too big. They should never take over the entire screen.
+ * If your content is that big, you should probably be using a Modal instead.
+ * :::
+ *
+ *
+ * @slot - header - Content to put inside the header surface
+ * @slot - body - Content to put inside the body surface
+ * @slot - footer - Content to put inside the footer surface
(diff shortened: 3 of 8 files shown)


jgroth (commented):
Great work! My comments are mostly minor, but there is one issue with having multiple popovers :slightly_smiling_face:

jensgustafsson (commented):
I think we need to close the popover component when clicking outside

src/components/popover/popover.tsx
```diff
@@ -0,0 +1,73 @@
...
+                    <slot name="popover__header" />
+                    <slot name="popover__body" />
+                    <slot name="popover__footer" />
+                </limel-popover-surface>
+            </limel-portal>
+        );
+    }
+
+    private appendElement() {
+        const content = document.body.querySelectorAll('slot');
```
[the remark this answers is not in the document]
filipharry: `this.host.querySelectorAll('slot')` returns an empty list for the context of popover. However I've moved this logic to the popover-surface component instead and it works well in that context, even with multiple popovers on the same page.

src/components/popover-surface/popover-surface.tsx
```diff
@@ -0,0 +1,39 @@
...
+            'mdc-menu': true,
+            'mdc-menu-surface': true,
+            'mdc-menu-surface--open': this.open,
+            'mdc-elevation-transition': true,
+            'mdc-elevation--z4': true,
+        };
+
+         return (
+             <div class={classList} tabindex="-1">
+                 <slot />
```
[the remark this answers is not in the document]
filipharry: I'm not entirely sure on this, adding those makes it so nothing renders so something else would be needed aswell.

src/components/popover/popover.tsx
```diff
@@ -0,0 +1,73 @@
...
+    public render() {
+        return (
+            <limel-portal
+                visible={this.visible}
+                containerId={this.portalId}
+            >
+                <limel-popover-surface open={this.visible}>
+                    <slot name="popover__header" />
+                    <slot name="popover__body" />
+                    <slot name="popover__footer" />
```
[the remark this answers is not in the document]
filipharry: Had a chat with Kia about this, we liked it so left it like that and I've updated the documentation aswell.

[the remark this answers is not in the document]
jgroth: Hmm ok, it will make it less consistent to the other components since none of those with slots have a prefix :slightly_smiling_face:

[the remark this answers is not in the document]
jensgustafsson: I think I agree with Johan :) Why did you and @Kiarokh prefer the prefixing?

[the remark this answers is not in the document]
jgroth: Another suggestion is to not have a slot name at all for the body slot, we do this for the dialog component

src/components/popover/popover.tsx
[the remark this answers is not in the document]
jgroth: Hmm ok, maybe I'm a bit confused of what is actually going on here and what elements are being moved :sweat_smile: 
If this is supposed to be the elements within the portal container, we could maybe make it a bit clearer by adding the container ID to the query selector, e.g

```ts
const portalSlotElements = document.body.querySelectorAll(`#${this.portalId} slot`);
```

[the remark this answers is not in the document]
jensgustafsson: Could we perhaps use the portal id? It was a while ago I worked with the portal but I think it's possible to set the id attribute to a specific id. Then we could just do: `querySelector('portal-id')`

src/components/popover/popover.tsx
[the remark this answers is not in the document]
filipharry: I've chosen to remove the prefixes in order the keep the naming convention similar to our other components.

src/components/popover/popover.tsx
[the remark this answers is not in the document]
filipharry: Yep, this works! Thanks for the tip!

Kiarokh:
> @jensgustafsson I think we need to close the popover component when clicking outside

I agree. I think that's the most expected behaviour. 

But we should investigate if there are use-cases where you want to close a popover by delibratly pressing a button (like a **save** or **close** or similar). Perhaps in those cases, user will use modals (?).

jensgustafsson:
> I think we need to close the popover component when clicking outside

Fixed!

jensgustafsson:
> > @jensgustafsson I think we need to close the popover component when clicking outside
> 
> I agree. I think that's the most expected behaviour.
> 
> But we should investigate if there are use-cases where you want to close a popover by deliberately pressing a button (like a **save** or **close** or similar). Perhaps in those cases, user will use modals (?).

For now it auto closes but if we find a need I guess we could add a prop that you could set if it should not auto close since that will be what we're using in most (all) cases

src/components/popover/popover.tsx
```diff
@@ -0,0 +1,77 @@
...
+                    <slot name="footer" />
+                </limel-popover-surface>
+            </limel-portal>
+        );
+    }
+
+    private globalClickListener({ target }) {
+        const clickedInside = Boolean(target.closest(`#${this.portalId}`));
+        if (this.visible && !clickedInside) {
+            this.visible = false;
```
jgroth, line 73: This will not do anything since this prop is not mutable.
However, I think it's a good idea that we set it ourselves and not forcing the user to listen to the `close` event.

I had a thought regarding behaviour similar as this the other day, and I think it could be a good idea to do something like this if we have a "default" behavior, e.g.

```ts
const event = this.close.emit();
if (!event.defaultPrevented) {
    this.visible = false;
}
```

This way we can have a default behavior that will close the popover, and the consumer can opt out of it by calling `event.preventDefault()`. They also still have the possibility to control the state from outside by setting the `visible` prop. We could probably do the same for the dialog and menu components etc. What do you think? :slightly_smiling_face:

src/components/popover/popover.tsx
```diff
@@ -42,4 +66,12 @@ export class Popover {
             </limel-portal>
         );
     }
+
+    private globalClickListener({ target }) {
+        const clickedInside = Boolean(target.closest(`#${this.portalId}`));
```
jgroth, line 71: Nice didn't know you could do stuff like this with `closest` :slightly_smiling_face:

src/components/popover/popover.tsx
```diff
@@ -0,0 +1,77 @@
...
+                >
+                    <slot name="header" />
+                    <slot name="body" />
+                    <slot name="footer" />
+                </limel-popover-surface>
+            </limel-portal>
+        );
+    }
+
+    private globalClickListener({ target }) {
```
jgroth, line 70: How about making this a bit more explicit? What do you think about
```suggestion
    private globalClickListener(event: MouseEvent) {
```

src/components/popover/popover.tsx
```diff
@@ -19,13 +26,30 @@ export class Popover {
...
     private host: HTMLLimelPopoverElement;
 
     private portalId: string;
 
     constructor() {
         this.portalId = createRandomString();
+        this.globalClickListener = this.globalClickListener.bind(this);
+    }
+
+    public componentDidLoad() {
```
jgroth, line 45: This is only called once, but `disconnectedCallback` can be called multiple times, so we should probably use `connectedCallback` instead :thinking:

jensgustafsson: The problem with `connectedCallback` is that it's called too early in the current implementation. Thus a close event is emitted before it has rendered.

But you had some other really good comments that might resolve this so I will look in to it.

Ideally I would prefer to use `disconnectedCallback` :)

jgroth: Aha! Yeah I guess it maybe will emit when you try to open it? :smile:

jgroth: Or maybe not since I now saw that we checked for `visible` as well before emitting :thinking:

src/components/popover/popover.tsx
jensgustafsson: > This will not do anything since this prop is not mutable.

Are you suggesting that we make this prop mutable or that we add some other state variable to the component?

jgroth: If we are keeping this line we need to make it mutable, otherwise we can just remove it :slightly_smiling_face: 
It depends on if we want to have a default behavior or not

jgroth: Adding another state variable will probably just make it hard to keep in sync with `visible`

jensgustafsson: > I had a thought regarding behaviour similar as this the other day, and I think it could be a good idea to do something like this if we have a "default" behavior, e.g.
> 
> ```ts
> const event = this.close.emit();
> if (!event.defaultPrevented) {
>     this.visible = false;
> }
> ```
> 
> This way we can have a default behavior that will close the popover, and the consumer can opt out of it by calling `event.preventDefault()`. They also still have the possibility to control the state from outside by setting the `visible` prop. We could probably do the same for the dialog and menu components etc. What do you think?

I like your idea!

But on the same time I'm not completely understanding it :)

I thought that the emitted event was fired asynchronously. Thus -> The if statement shouldn't work according to me. (But it seems to do)

Can you explain it for me? :)

src/components/popover/popover.tsx
jensgustafsson: Unless you send in visible as a prop to the component 😛

src/components/popover/popover.tsx
jgroth: No I think the events are sent synchronously :thinking: Otherwise this would not work :sweat_smile: 

If we do it like this we can have a default behavior where the popover closes without forcing the user to listen to the `close` event. If they don't want to have the default behavior they can prevent it by invoking `event.preventDefault()`. I think this is a bit more clean than having multiple events like `beforeClose` and stuff like that to prevent some behavior.

On the other hand I'm not sure if it will be possible to open the popover again without listening to the `close` event and updating the local `isOpen` state in the consumer component :thinking:

src/components/popover/popover.tsx
jgroth: We could also setup the listener when the popover opens, and then remove the listener when it has been closed :slightly_smiling_face: That way we won't have to listen for events and have code running when there is no need for it

src/components/popover/popover.tsx
jensgustafsson: Then everything makes sense! Thanks!🤸‍♂️🌟

src/components/popover-surface/popover-surface.scss
```diff
@@ -0,0 +1,42 @@
+@import '../../style/functions';
+
+/**
+ * @prop --popover-surface-width: Width of the popover surface. Defaults to `auto`
+ * @prop --popover-header-background-color: Background color of popover header, deafutls to `--contrast-300`.
+ * @prop --popover-body-background-color: Background color of popover body, deafutls to `--contrast-100`.
+ * @prop --popover-footer-background-color: Background color of popover footer, deafutls to `--contrast-500`.
```
jgroth, line 7: These should be moved to `popover.scss` otherwise they will not appear in the documentation

src/components/popover/popover.tsx
jgroth: Seems like it is hard to open it again after it has been closed if we have the default behavior and the consumer is not using a close listener, so I'm not sure there is any benefit to it :thinking: If we would have it like that I think we need to have an imperative API with a method on the element named `open()` that you can invoke to open it, but we don't have methods like that in other components and I believe it's best practice to not have them. I just removed this line for now and maybe we can revisit this feature in the future

jensgustafsson: what about switching name `visible` -> `open`?

jgroth: It has been renamed :slightly_smiling_face:

Kiarokh: I don’t understand the technical difficulties you are facing. 

But usually consumers don’t care that much about these small details. 

Thus the risk is that they don't do anything about it, and we open several popovers, without closing the previous ones. This messes up the screen. 

I think it’s good to have a default behaviour for clicking outside which is close, and let them override it if they have to. That’s the normal and universal behaviour of popovers.

jensgustafsson (approved):

PR #393, opened 2019-08-20
Improvements limel-picker

hannahu:
fix #363

Issue #363, opened 2019-05-07
Minor improvements to limel-picker (badge size, positioning, help text)


magnusfagerlund:
1. Decrease the size of the icon (see image)
2. Add help text "Search and connect to [limetype]" (see images)
3. Lower the suggestions menu so that you see the line underneath (see image)
4. Always show 5 suggestions if there are (never show scrollbar)
![limelpicker](https://user-images.githubusercontent.com/7524499/57289061-f6e5e400-70ba-11e9-8404-f2cee8504362.png)

adrianschmidt:
#393 fixes 1) and 3), and adds options for 2) and 4). To actually fix 2) and 4), some additional changes are needed in lime-webclient.

PR #1110, opened 2021-01-26
fix(portal): popper is updated when container height is updated

maltja:
The popper needs to recalculate on which side the container should flip at when the container's
dimension changes

fix Lundalogik/crm-feature#1722
fix Lundalogik/crm-feature#1705

Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [x] Edge
- [x] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/menu-surface/menu-surface.tsx
+++ b/src/components/menu-surface/menu-surface.tsx
@@ -41,7 +41,6 @@ export class MenuSurface {
     private host: HTMLLimelMenuSurfaceElement;
 
     private menuSurface: MDCMenuSurface;
-    private observer: IResizeObserver;
 
     constructor() {
         this.handleDocumentClick = this.handleDocumentClick.bind(this);
@@ -95,13 +94,6 @@ export class MenuSurface {
         window.addEventListener('resize', this.handleResize, {
             passive: true,
         });
-
-        if ('ResizeObserver' in window) {
-            const observer = new ResizeObserver(() => {
-                this.ensureMenuFitsInViewPort();
-            });
-            observer.observe(this.host);
-        }
     }
 
     private teardown() {
@@ -111,10 +103,6 @@ export class MenuSurface {
         });
         this.host.removeEventListener('keydown', this.handleKeyDown);
         window.removeEventListener('resize', this.handleResize);
-
-        if (this.observer) {
-            this.observer.unobserve(this.host);
-        }
     }
 
     private handleDocumentClick(event) {
@@ -175,31 +163,5 @@ export class MenuSurface {
             event.stopPropagation();
             this.dismiss.emit();
         }
-    }
-
-    private ensureMenuFitsInViewPort() {
-        this.host.style.height = 'auto';
-        setTimeout(() => {
-            const viewHeight = Math.max(
-                document.documentElement.clientHeight || 0,
-                window.innerHeight || 0
-            );
-            const { top, bottom } = this.host.getBoundingClientRect();
-            if (viewHeight > bottom && top > 0) {
-                // The surface is rendered inside the viewport :)
-                return;
-            }
-
-            // Set the height of the surface so that it fits either above or
-            // below the trigger, depending on where there is more space.
-            const spaceAboveTopOfSurface = Math.max(top, 0);
-            const spaceBelowTopOfSurface = Math.max(viewHeight - top, 0);
-            const extraCosmeticSpace = 16;
-            const maxHeight =
-                viewHeight -
-                Math.min(spaceAboveTopOfSurface, spaceBelowTopOfSurface) -
-                extraCosmeticSpace;
-            this.host.style.height = `${maxHeight}px`;
-        });
     }
 }
--- a/src/components/menu/menu.scss
+++ b/src/components/menu/menu.scss
@@ -28,19 +28,4 @@ $menu-item-background-color: #ebebeb;
 
 .mdc-menu-surface--anchor {
     position: relative;
-
-    limel-portal {
-        &.limel-portal--fixed {
-            position: fixed;
-        }
-
-        &:not(.limel-portal--fixed) {
-            z-index: -999999;
-            position: absolute;
-            top: 0;
-            right: 0;
-            bottom: 0;
-            left: 0;
-        }
-    }
 }
--- a/src/components/picker/picker.scss
+++ b/src/components/picker/picker.scss
@@ -2,4 +2,9 @@
 
 :host {
     position: relative;
+    display: block;
+}
+
+:host([hidden]) {
+    display: none;
 }
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -25,7 +25,6 @@ import { createRandomString } from '../../util/random-string';
 
 const SEARCH_DEBOUNCE = 500;
 const CHIP_SET_TAG_NAME = 'limel-chip-set';
-const ITEM_LIMIT_NO_SCROLL = 5;
 
 /**
  * @exampleComponent limel-example-picker
@@ -98,6 +97,10 @@ export class Picker {
 
     /**
      * True if the dropdown list should be displayed without cutting the content
+     *
+     * @deprecated This was used for a workaround, and isn't needed any
+     * longer. Setting it has no effect, and the property will be removed
+     * in the next major version.
      */
     @Prop()
     public displayFullList: boolean = false;
@@ -206,7 +209,7 @@ export class Picker {
                 emptyInputOnBlur={false}
                 {...props}
             />,
-            <div class="mdc-menu-surface--anchor">{this.renderDropdown()}</div>,
+            this.renderDropdown(),
         ];
     }
 
@@ -262,13 +265,8 @@ export class Picker {
      */
     private renderDropdown() {
         const content = this.getDropdownContent();
-        const styling = {};
 
-        if (this.isScrollableDropdown()) {
-            styling['max-height'] = '250px';
-        }
-
-        return this.renderPortal(content, styling);
+        return this.renderPortal(content);
     }
 
     private getDropdownContent() {
@@ -337,29 +335,20 @@ export class Picker {
         );
     }
 
-    private isScrollableDropdown() {
-        if (this.displayFullList) {
-            return false;
-        }
-
-        if (!this.items || !this.items.length) {
-            return false;
-        }
-
-        return this.items.length > ITEM_LIMIT_NO_SCROLL;
-    }
-
-    private renderPortal(content = null, styling = {}) {
+    private renderPortal(content = null) {
         return (
             <limel-portal
                 visible={!!content}
                 containerId={this.portalId}
-                containerStyle={styling}
                 inheritParentWidth={true}
             >
                 <limel-menu-surface
                     open={!!content}
-                    style={{ '--menu-surface-width': '100%' }}
+                    style={{
+                        '--menu-surface-width': '100%',
+                        'max-height': 'inherit',
+                        display: 'flex',
+                    }}
                 >
                     {content}
                 </limel-menu-surface>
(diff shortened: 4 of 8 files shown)


src/components/picker/picker.tsx
```diff
@@ -98,6 +97,9 @@ export class Picker {
 
     /**
      * True if the dropdown list should be displayed without cutting the content
+     *
+     * @deprecated This functiallity dosen't work anymore and should be
+     * handled by the searcher insted
```
adrianschmidt, line 102: Small update on this. The picker wasn't actually limiting the number of items in the list, and so this property wasn't to avoid that. Instead, this property was there to allow the dropdown to become high enough to show all items in the list, regardless of whether that made the dropdown extend outside the screen. Thus, if not all items fit on screen, the user would scroll the whole page instead of the content of the dropdown.

It seems this hasn't worked lately though. Compare the _With icons and displaying full list without cutting content_ example of version [30.0.1](http://lime-documentation-lime-elements.s3-website-eu-west-1.amazonaws.com/versions/30.0.1/#/component/limel-picker/) (working) and version [30.0.2](http://lime-documentation-lime-elements.s3-website-eu-west-1.amazonaws.com/versions/30.0.2/#/component/limel-picker/) (broken). Type an `a` in the input to get a long list of results.

As we found previously, this property was originally added in commit 7f0a31a9e51f76acb866058d7aadde43771e0cd7, but I finally managed to find the PR for that commit: #393, fixing issue #363.

We should test the stuff listed in that issue, to make sure we don't reintroduce any of those problems.

adrianschmidt: > to make sure we don't reintroduce any of those problems

(I don't think we will, to be fair. There's a lot of things that are different about the dropdown now: we calculate the size, we can pop it either above or below the trigger, you can scroll the page while the dropdown is visible, etc.)

adrianschmidt: Oh, yeah, what I was _going_ to say was that we should update this comment a bit. Perhaps something like

```suggestion
     * @deprecated This was used for a workaround, and isn't needed any
     * longer. Setting it has no effect, and the property will be removed
     * in the next major version.
```

maltja: nice that you found the issues! 🎉  For me both of the versions are broking, I don't know if I do anything wrong when I tested them. 

"If not all items fit on screen, the user would scroll the whole page instead of the content of the dropdown."
I don't see why we want to be able to switch between scrolling in the whole page and scrolling in the dropdown menu. 🤔

maltja: Yes I agree and will update it 🙂

adrianschmidt: > nice that you found the issues! 🎉 For me both of the versions are broking, I don't know if I do anything wrong when I tested them.

Hmm… I might have linked the wrong versions or something. I'll take a look and post a video showing what I mean.

> "If not all items fit on screen, the user would scroll the whole page instead of the content of the dropdown."
> I don't see why we want to be able to switch between scrolling in the whole page and scrolling in the dropdown menu. 🤔

I think back when that issue was created, we couldn't pop the dropdown _above_ the trigger, so we always just had the space _below_ the trigger to work with. That meant that if the trigger was placed at the bottom of the page, the dropdown would become tiny, sometimes smaller than a single list-item in height. Coupled with the fact that back then, when a dropdown was visible, the page behind it couldn't be scrolled, and we couldn't place the dropdown above the item instead of below it, this led to some workarounds that perhaps doesn't seem very obvious today 😄

Today though, all of those compounding problems have been fixed, and we should never end up in the situation that the dropdown is opened downward and shrunk to less than a single item in height, because we'll just be popping the dropdown upwards instead 👍

adrianschmidt: The first example shown is v30.0.1 where the dropdown becomes longer than the page (thus what the prop is supposed to do), and the second example is v30.0.2 where the dropdown is shorter and gets a scrollbar.

https://user-images.githubusercontent.com/154725/106133121-1f522380-6165-11eb-84b5-dda4173101e2.mov

To clarify, we should still remove the behaviour in the first example 🙂

src/components/menu-surface/menu-surface.tsx
```diff
@@ -176,30 +164,4 @@ export class MenuSurface {
             this.dismiss.emit();
         }
     }
-
-    private ensureMenuFitsInViewPort() {
```
adrianschmidt, line 180: > refactor(picker): move calculation of dropdown menu height from menu-surface to picker

Hmm… I haven't looked at the code yet, but I think we'll be needing this calculation in all dropdowns, not just those belonging to limel-picker. So if it can't be done in `menu-surface`, we might want to extract it to a helper or something, that can be used in limel-picker, limel-menu, etc.

Again, I haven't looked at your changes yet, I just saw the commit message and wanted to make a note of it before logging off for the day.

maltja:
I fixed the commit messages that was wrong and confusing 😅. I also added the suggested changes to the deprecated displayFullList property.

src/components/picker/picker.tsx
```diff
@@ -99,6 +97,10 @@ export class Picker {
 
     /**
      * True if the dropdown list should be displayed without cutting the content
+     *
+     * @deprecated This was used for a workaround, and isn't needed any
+     * longer. Setting it has no effect, and the property will be removed
+     * in the next major version.
```
adrianschmidt, line 103: We don't need to set this commit as a breaking change, since the functionality was broken in an earlier version, and all we do in this PR is fixing it the bug, albeit in a way that makes the property unnecessary.

The breaking change comes when we actually remove the property.

I can update the commit message, since I have the branch checked out anyway.

maltja: ah okey, thanks! 🙂

src/components/picker/picker.tsx
```diff
@@ -26,6 +26,7 @@ import { createRandomString } from '../../util/random-string';
 const SEARCH_DEBOUNCE = 500;
```
adrianschmidt, line 26: It feels like the changes to `menu-surface.tsx` and `portal.tsx` in this commit belong together (and might belong together with the first commit), while the changes to `picker.tsx` feel like they might belong in their own commit.

I assume the changes to `picker.tsx` are needed as a result of moving the dropdown height calculation from menu-surface to portal?

src/components/picker/picker.tsx
```diff
@@ -349,17 +345,20 @@ export class Picker {
...
-                containerStyle={styling}
                 inheritParentWidth={true}
             >
                 <limel-menu-surface
                     open={!!content}
-                    style={{ '--menu-surface-width': '100%' }}
+                    style={{
+                        '--menu-surface-width': '100%',
+                        'max-height': 'inherit',
+                        display: 'flex',
```
adrianschmidt, line 350: I was just about to comment that these should be set in the scss file, but that doesn't work because of the portal I guess?

src/components/picker/picker.scss
```diff
@@ -2,4 +2,5 @@
 
 :host {
     position: relative;
+    display: block;
 }
```
adrianschmidt, line 6: When setting the `display` property of the `:host` selector, that will override the browser styles that hides the element if the attribute `hidden` is set. So we need to add that functionality back:

```suggestion
}

:host([hidden]) {
    display: none;
}
```

src/components/menu/menu.scss
```diff
@@ -28,19 +28,4 @@ $menu-item-background-color: #ebebeb;
```
adrianschmidt, line 28: The changes to `menu.scss` and `portal.scss` in `fix(portal): give portal exact size of its parent` seem like they might belong together with the changes to `portal.tsx` and `menu-surface.tsx` in the first two commits.

src/components/select/select.scss
```diff
@@ -11,6 +11,12 @@
     display: none;
```
adrianschmidt, line 11: I like that all the changes to limel-select are contained in a single commit. I think all the changes to limel-picker also belong in a single commit, and the changes to limel-portal, limel-menu, and limel-menu-surface belong in a single commit (which should come before the others).

I'll move stuff around a little on a separate branch, and you can see if you like it better my way or not 🙂

src/components/select/select.scss
```diff
@@ -11,6 +11,12 @@
...
 }
 
+:host {
+    display: block;
+    position: relative;
+    height: $height-of-mdc-text-field;
+}
```
adrianschmidt, line 18:
```suggestion
}

:host([hidden]) {
    display: none;
}
```

src/components/picker/picker.tsx
```diff
@@ -26,6 +26,7 @@ import { createRandomString } from '../../util/random-string';
 const SEARCH_DEBOUNCE = 500;
 const CHIP_SET_TAG_NAME = 'limel-chip-set';
 const ITEM_LIMIT_NO_SCROLL = 5;
+const MAX_HEIGHT_DROP_DOWN = 320;
```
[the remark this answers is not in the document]
maltja: Yes, I don't know how to fix that 😕 Do you know how I can fix that? 😅

src/components/picker/picker.tsx
maltja: yes, I agree 🙂

src/components/picker/picker.tsx
maltja: I tried to fix this with Kia but we didn't manage to get it to work. It's because it is set to the container so it is not in the right place in the HTML to be in the Portal's scss file. But I guess we could set it in the portal.tsx file 🤔

src/components/picker/picker.scss
maltja: ah okey, what does that mean exactly? That it is not possible to hide the element with my changes? 🤔

src/components/select/select.scss
maltja: I agree 🙂

src/components/picker/picker.tsx
adrianschmidt: No, don't worry about it. It's much better that it's set here than in the portal.

src/components/picker/picker.scss
adrianschmidt: No, you can still hide it anyway, but if you write

```html
<p hidden>This should be hidden.</p>
```

that paragraph will indeed be hidden. However, if you have styled it with `display:block` via css, and _then_ add the `hidden` attribute, the css wins. Which perhaps wouldn't normally matter much, but since we're building a component library, that's meant to be used in many circumstances by different people, we want our components to behave like native components (within reason).

src/components/select/select.scss
adrianschmidt: Ah, this one already had this the `:host([hidden])`, but it comes before this new code. It shouldn't matter for the rendering, but consistency is nice when reading the code, so I'll swap the order around 🙂

src/components/picker/picker.scss
adrianschmidt: :zap:

src/components/picker/picker.tsx
adrianschmidt: :zap:

src/components/picker/picker.tsx
adrianschmidt: ⚡

src/components/select/select.scss
adrianschmidt: ⚡

src/components/menu/menu.scss
adrianschmidt: ⚡

src/components/picker/picker.tsx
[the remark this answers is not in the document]
adrianschmidt: ⚡

src/components/select/select.scss
adrianschmidt: ⚡

adrianschmidt (approved):

PR #1112, opened 2021-01-28
fix(popover): do not emit close event when clicking inside nested portals

jgroth:
If the popover contains e.g. a limel-select, the dropdown of the select will be placed inside a
limel-portal while the popover content is inside another portal. Clicking inside the select dropdown
should not emit a close event, se we have to go back through all the portals to the source element
tree when deciding if the event should be emitted or not.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/portal/portal.tsx
+++ b/src/components/portal/portal.tsx
@@ -141,6 +141,9 @@ export class Portal {
         this.container = document.createElement('div');
         this.container.setAttribute('id', this.containerId);
         this.container.setAttribute('class', 'limel-portal--container');
+        Object.assign(this.container, {
+            portalSource: this.host,
+        });
 
         content.forEach((element: HTMLElement) => {
             this.container.appendChild(element);
--- a/src/components/popover/popover.tsx
+++ b/src/components/popover/popover.tsx
@@ -9,6 +9,7 @@ import {
 } from '@stencil/core';
 import { createRandomString } from '../../util/random-string';
 import { zipObject } from 'lodash-es';
+import { portalContains } from '../portal/contains';
 
 /**
  * A popover is an impermanent layer that is displayed on top of other content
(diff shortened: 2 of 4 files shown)


src/components/portal/portal.tsx
```diff
@@ -141,6 +141,9 @@ export class Portal {
         this.container = document.createElement('div');
         this.container.setAttribute('id', this.containerId);
         this.container.setAttribute('class', 'limel-portal--container');
+        Object.assign(this.container, {
+            portalSource: this.host,
+        });
```
jensgustafsson, line 146: Why do we need to use `Object.assign` here?

jensgustafsson (approved):
Nice implementation! Easy to follow and pretty straight forward 🌟

I only had one comment, and that was a question :)

Approving!

src/components/portal/portal.tsx
jgroth: It's kind of a cheat to trick the compiler :smile: 

The following will give a compiler error since `portalSource` is not part of `HTMLElement`
```ts
this.container.portalSource = this.host;
```

The following will give a lint error since we should use the syntax from the first example :upside_down_face: 
```ts
this.container['portalSource'] = this.host;
```

jgroth: We could introduce another internal web-component, e.g. `limel-portal-container` instead of using a regular `div` for the container which would make this a bit more clear, but that's for a future project if we want to do that :slightly_smiling_face:

jensgustafsson: Thanks! I guessed it boiled down to that but wanted to ensure that I didn't misunderstand some parts

jgroth: Another component might also separate the logic a bit which could be nice :thinking: 
The portal would be responsible for moving the elements in the DOM and the container would be responsible for positioning them

PR #1113, opened 2021-01-28
feat(header): add header component

filipharry:
fix: https://github.com/Lundalogik/crm-feature/issues/1761

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/header/examples/header-narrow.tsx
+++ b/src/components/header/examples/header-narrow.tsx
@@ -1,0 +1,34 @@
+import { Component, h } from '@stencil/core';
+/**
+ * Narrow headers
+ * Sometimes your UI design may require having a narrower header.
+ * This will be easy to achieve by sending in the class of `is-narrow`
+ * to your component.
+ *
+ * This will render the header icon smaller, and reduces the font size of
+ * the `heading`.
+ *
+ * :::tip
+ * Keep in mind that headers are programmed to grow in height, depending
+ * on their content. So if you have large custom components in the `actions`
+ * slot or use both `heading` and `subheading`, they will still force the header
+ * to appear tall.
+ * :::
+ *
+ */
+@Component({
+    tag: 'limel-example-header-narrow',
+    shadow: true,
+    styleUrl: 'header-narrow.scss',
+})
+export class HeaderExample {
+    public render() {
+        return (
+            <limel-header
+                class="is-narrow"
+                icon="ok"
+                heading="This is a narrow header"
+            />
+        );
+    }
+}
--- a/src/components/header/examples/header-responsive.tsx
+++ b/src/components/header/examples/header-responsive.tsx
@@ -1,0 +1,92 @@
+import { Component, h } from '@stencil/core';
+import { Action } from '@limetech/lime-elements';
+/**
+ * How Responsive layout of header works
+ * However, sometimes you may need to make the layout be responsive and split
+ * into two rows, at a break point.
+ *
+ * To activate this responsive layout, you can simply add the `has-responsive-layout`
+ * class to your `limel-header` component.
+ *
+ * This makes a few changes in the layout. Firstly, both the left side (icon and
+ * headings) and right side (actions slot) will occupy 50% of the total header
+ * width each. However, the width of left and right side will never become smaller
+ * than `22rem`.
+ *
+ * :::tip
+ * The value of `22rem` is the default breakpoint. But you can easily change it
+ * by tweaking the `--header-responsive-breakpoint` variable in your component.
+ * :::
+ *
+ */
+
+@Component({
+    tag: 'limel-example-header-responsive',
+    shadow: true,
+    styleUrl: 'header-responsive.scss',
+})
+export class HeaderExample {
+    private actions = [
+        {
+            id: '1',
+            icon: 'refresh',
+            label: 'Refresh',
+        },
+        {
+            id: '2',
+            icon: 'delete',
+            label: 'Delete',
+            disabled: true,
+        },
+        {
+            id: '3',
+            icon: 'edit',
+            label: 'Edit',
+        },
+    ];
+
+    constructor() {
+        this.renderActionButton = this.renderActionButton.bind(this);
+    }
+
+    public render() {
+        return (
+            <limel-header
+                class="has-responsive-layout"
+                icon="resize_horizontal"
+                heading="This header is responsive"
+                subheading="Resize the container of this header to see how it works"
+            >
+                {this.renderActions()}
+            </limel-header>
+        );
+    }
+
+    private renderActions() {
+        if (!this.actions) {
+            return;
+        }
+
+        return (
+            <div class="actions">
+                {this.actions.map(this.renderActionButton)}
+            </div>
+        );
+    }
+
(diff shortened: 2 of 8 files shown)


jensgustafsson (commented):
The implementation looks good to me!

But I think we need to add some docs that describes how and when this component should be used.

Also, should we perhaps hide it in the docs until we have tried using it in one of our components? Then its easier to make breaking changes as we don't have to treat them as breaking :)

src/components/header/examples/header.tsx
```diff
@@ -0,0 +1,48 @@
+import { Component, h, State } from '@stencil/core';
+import { Action } from '@limetech/lime-elements';
+
+@Component({
+    tag: 'limel-example-header',
+    shadow: true,
+})
+export class HeaderExample {
+    @State()
```
Befkadu1, line 9: I am not sure if it is a `State` here

src/components/header/examples/header.tsx
```diff
@@ -0,0 +1,48 @@
...
+        },
+        {
+            id: '3',
+            icon: 'edit',
+            label: 'Edit',
+        },
+    ];
+
+    constructor() {
+        this.handleAction = this.handleAction.bind(this);
```
Befkadu1, line 30: No necessary to bind this for this example

src/components/header/header.tsx
```diff
@@ -0,0 +1,91 @@
+import { Component, h, Prop, Event, EventEmitter } from '@stencil/core';
+import { Action } from '../collapsible-section/action';
```
Befkadu1, line 2: If we need to use this component in the global table and object card. We don't need to import the Action from `collabsible-section`. This is the interface we need 
``` 
export interface Action {
    /**
     * The command to send to the commandbus when the action is activated
     */
    command: object;

    /**
     * Icon of the action
     */
    icon?: string;

    /**
     * Label to display next to the action
     */
    label?: string;

    /**
     * True if the action is disabled
     */
    disabled?: boolean;
}

File: src/components/header/examples/header.tsx (hunk 1 above)
Comment (adrianschmidt) [reply src/components/header/examples/header.tsx:10]: No, since they don't change, the State decorator isn't necessary, and just wires up unnecessary overhead.

File: src/components/header/examples/header.tsx (hunk 2 above)
Comment (adrianschmidt) [reply src/components/header/examples/header.tsx:31]: Correct, since we don't access `this` inside `handleAction`. But since any real world `handleAction` would almost certainly access `this`, and this is an example that people will look at to learn, maybe we should still bind it… 🤔

The best argument I can see against it is that it might be confusing to some if we use `bind` where we don't need it.

Just my 2¢, I'll leave the decision to you people 😊

File: src/components/header/examples/header-narrow.tsx [hunk 4]
```diff
@@ -0,0 +1,35 @@
+import { Component, h } from '@stencil/core';
+import { Action } from '@limetech/lime-elements';
```
Befkadu1, line 2: remove unused import

src/components/header/examples/header.tsx
filipharry: Fixed!

src/components/header/examples/header.tsx
filipharry: Decided to leave this here just for this example.

src/components/header/examples/header-narrow.tsx
filipharry: Fixed!

src/components/header/examples/header.scss
```diff
@@ -0,0 +1,6 @@
+:host(limel-example-header) {
+    --header-background-color: rgb(var(--contrast-1100));
```
Befkadu1, line 2: With this background color setting it is difficult to know which icon-buttons are `enabled/disabled`

Screenshot 2021-02-02 at 08 56 10

src/components/header/examples/header.tsx
Befkadu1: What I wanted to say is it might be confusing to some if we use bind where we don't need it.

Befkadu1 (changes requested):

src/components/header/examples/header.tsx
filipharry: Yeah that make sense, I removed this binding in the example.

src/components/header/examples/header.scss
Kiarokh: fixed

jgroth (commented):
Looks great! Good job!
I think it is ready to be merged, but we should probably fix the typo and also squash all the commits

src/components/header/header.tsx
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * For example, when a header is placed on top of a card, it should quickly
+ * explain the card to the user. When placed on top of a modal, it should easily
+ * clarify what the modal is about. When displayed on top of a fullscreen view,
+ * it should indicate where in the system users are, and what part of the app
+ * they are looking at.
+ *
+ *
+ * ## Layout
+ * The vital information in a header is usually manifested if form of an icon,
```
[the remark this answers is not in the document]
filipharry: Fixed!

jgroth (dismissed):

Kiarokh:
I tend to have many typos in my texts 😄

src/components/header/examples/header-responsive.tsx
```diff
@@ -0,0 +1,92 @@
...
+ * However, sometimes you may need to make the layout be responsive and split
+ * into two rows, at a break point.
+ *
+ * To activate this responsive layout, you can simply add the `has-responsive-layout`
+ * class to your `limel-header` component.
+ *
+ * This makes a few changes in the layout. Firstly, both the left side (icon and
+ * headings) and right side (actions slot) will occupy 50% of the total header
+ * width each. However, the width of left and right side will never become smaller
+ * than `22rem`.
```
Befkadu1, line 14: I think the default breakpoint is 12rem

Befkadu1 (changes requested):

src/components/header/examples/header-responsive.tsx
```diff
@@ -0,0 +1,92 @@
...
+ *
+ * :::tip
+ * The value of `22rem` is the default breakpoint. But you can easily change it
```
Befkadu1, line 17: The default breakpoint is 12rem

src/components/header/examples/header-responsive.tsx
Kiarokh: No, the default is `352px` or `22rem` 👇

````scss
grid-template-columns: repeat(
        auto-fit,
        minmax(
            clamp(50%, var(--header-responsive-breakpoint, pxToRem(352)), 100%),
            1fr
        )
    );
````

However in the example I have overwritten in with `12rem`

Befkadu1 (approved):

PR #1114, opened 2021-01-29
Table component events

jgroth:
This PR makes it possible to add event listeners to custom components of the table view, both in the header and in the cells

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/columns.ts
+++ b/src/components/table/columns.ts
@@ -2,6 +2,7 @@ import { Column, ColumnSorter, ColumnAggregatorFunction } from './table.types';
 import Tabulator from 'tabulator-tables';
 import { escape } from 'html-escaper';
 import { ElementPool } from './element-pool';
+import { pickBy, negate } from 'lodash-es';
 
 export class ColumnDefinitionFactory {
     constructor(private pool: ElementPool) {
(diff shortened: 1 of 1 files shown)


src/components/table/columns.ts
```diff
@@ -181,13 +182,37 @@ export function createCustomComponent(
...
+    if (typeof value !== 'function') {
+        return false;
+    }
+
+    return /^on[A-Z]/.test(key);
+}
+
+function getEventName(propertyName: string): string {
+    // eslint-disable-next-line no-magic-numbers
+    return propertyName.charAt(2).toLowerCase() + propertyName.slice(3);
```
jensgustafsson, line 213: I like the solution but I suggest that we add some kind of comment or docstring that explains what's going on. It took quite a bit of time for me to realise what we're doing :)

jensgustafsson (changes requested):

src/components/table/columns.ts
jgroth: 7a9f5ff

jensgustafsson (dismissed):
👏

jensgustafsson (approved):

PR #1116, opened 2021-02-01
fix(popover): remove unnecessary slots

filipharry:
fix: https://github.com/Lundalogik/crm-feature/issues/1791

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/popover-surface/popover-surface.tsx
+++ b/src/components/popover-surface/popover-surface.tsx
@@ -18,12 +18,6 @@ export class PopoverSurface {
     public open = false;
 
     /**
-     * Portal Id
-     */
-    @Prop()
-    public portalId: string;
-
-    /**
      * Content to render
      */
     @Prop()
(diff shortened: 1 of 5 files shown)


src/components/popover-surface/popover-surface.tsx
```diff
@@ -44,26 +38,16 @@ export class PopoverSurface {
 
         return (
             <div class={classList} tabindex="-1">
-                <slot name="header" />
                 <slot name="body" />
```
jgroth, line 41: Do we still need the name of the slot here? :thinking:

jgroth (commented):
Looks good! :slightly_smiling_face: 
Just had a small question, otherwise I think we can merge it!

src/components/popover-surface/popover-surface.tsx
filipharry: I tried removing the named slot but then it will not render anything, I couldn't figure out why so I left it as is.

jgroth: Aha! I guess because then the content you put inside will go to the slot in the slot and that might cause some weird behavior :thinking: We would probably get the same behavior if we do something like this in the parent popover component
```tsx
function render() {
    return <slot slot="body" />
}
```

jgroth (approved):

PR #1123, opened 2021-02-05
fix(form): vertically align switches and checkboxes

maltja:
fix Lundalogik/crm-feature#1629

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [x] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/form/form.scss
+++ b/src/components/form/form.scss
@@ -1,4 +1,5 @@
 @import '../../style/internal/mdc-variables';
+@import '../../style/internal/shared_input-select-picker';
 
 /**
  * @prop --form-body-padding: space around content of the form
@@ -109,5 +110,15 @@
 
     .limel-form-layout-colspan--all {
         grid-column: 1/-1;
+    }
+}
+
+.form-group {
+    limel-checkbox,
+    limel-switch {
+        margin-left: pxToRem(8);
+        min-height: calc(
+            var(--min-height-of-one-row) - #{$height-of-mdc-helper-text-block}
+        ); //This is because we have to align these items to top and make sure they align correctly even in one-columned layouts
     }
 }
--- a/src/components/switch/switch.scss
+++ b/src/components/switch/switch.scss
@@ -2,6 +2,11 @@
 
 @import '@limetech/mdc-switch/mdc-switch';
 
+:host {
+    display: inline-flex;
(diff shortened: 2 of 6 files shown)


Kiarokh:
@maltja I have pushed a few more changes, and a fixup to your original commit. I think someone else should review this now, since I have been making changes too

src/components/form/form.scss
```diff
@@ -111,3 +111,13 @@
         grid-column: 1/-1;
     }
 }
+
+.form-group {
+    limel-checkbox,
+    limel-switch {
+        margin-left: pxToRem(8);
+        min-height: calc(
+            var(--min-height-of-one-row) - #{pxToRem(15)} /*height of helper text*/
```
adrianschmidt, line 120: If we don't have a variable for "height of helper text", maybe we should add one? 🙂

src/components/switch/switch.scss
```diff
@@ -2,8 +2,14 @@
...
 @import '@limetech/mdc-switch/mdc-switch';
 
+:host {
+    display: inline-flex;
+    align-items: center;
+}
+
 .mdc-switch {
     @include mdc-switch-toggled-on-color(primary);
+    line-height: normal;
```
adrianschmidt, line 12: Was this intended to go on the label? I've only checked in Chrome on macOS, but switching this on and off makes no difference, at least not where I'm checking. I tried adding `line-height:normal` to the label though, and that changed the alignment a pixel or two, the result of which I thought looked better.

adrianschmidt (commented):
Very nice! I left two small comments.

src/components/form/form.scss
Kiarokh: great Idea. I'll do a fixup

src/components/switch/switch.scss
Kiarokh: awesome catch! yes, it was for the label 😅

Kiarokh:
Thanks @adrianschmidt I pushed two fixups

adrianschmidt (approved):

Issue #1128, opened 2021-02-08
input-field: Remove what little (rather useless) input validation there is for type `urlAsText`


adrianschmidt:
Spawned from discussion about #1111, when we decided that doing that one fully wasn't worth it, at least not right now, and a much quicker solution is to just remove the rather useless validation I added originally.

PR #1129, opened 2021-02-08
fix(input-field): remove unnecessary validation when using `urlAsText`

adrianschmidt:
fix: #1128

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.e2e.ts
+++ b/src/components/input-field/input-field.e2e.ts
@@ -340,36 +340,6 @@ describe('limel-input-field', () => {
                 });
             })
         );
-
-        const invalidUrls = [
-            { input: 'justoneword' },
-            { input: 'some words with spaces' },
-        ];
-
-        invalidUrls.forEach((url) =>
-            describe(`with a value of '${url.input}'`, () => {
-                beforeEach(async () => {
-                    const nativeInput = await page.find(
-                        'limel-input-field>>>input'
-                    );
-                    nativeInput.focus();
-                    limelInput.setProperty('value', url.input);
-                    nativeInput.press('Tab');
-                    await page.waitForChanges();
-                });
-                it('IS considered invalid', () => {
-                    expect(inputContainer).toHaveClass(
-                        'mdc-text-field--invalid'
-                    );
-                });
-                it('has a trailing icon indicating the field is invalid', async () => {
-                    const icon = await page.find(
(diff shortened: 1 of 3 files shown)


maltja (commented):
Should we keep the urlAsText type, now when it is only a text type and doesn't do any validation? Maybe we should update the comment as well? 🤔 

```
     * **Note** regarding type `url`: `limel-input` uses the native validation
     * built into the browser for many types of input fields. The native
     * validation for `url` is very strict, and does not allow relative urls,
     * nor any other formats that are not a "fully qualified" url. To allow
     * such urls, use the type `urlAsText` instead. This will use `text` as the
     * type for the native input element, and only apply minimal validation.
    
    @Prop({ reflect: true })
    public type: InputType = 'text';
```

adrianschmidt:
> Should we keep the urlAsText type, now when it is only a text type and doesn't do any validation? Maybe we should update the comment as well? 🤔

Haha! Doh! 🤦‍♂️ No, we should't 😂

Thanks for catching that!

src/components/input-field/input-field.types.ts
```diff
@@ -11,5 +11,4 @@ export type InputType =
     | 'textarea'
     | 'time'
     | 'url'
-    | 'urlAsText'
```
adrianschmidt, line 14: Don't merge this yet. It's currently a breaking change, and I want to make it non-breaking.

adrianschmidt:
> > Should we keep the urlAsText type, now when it is only a text type and doesn't do any validation? Maybe we should update the comment as well? 🤔
> 
> Haha! Doh! 🤦‍♂️ No, we should't 😂
> 
> Thanks for catching that!

I changed my mind again, because we want to be able to use the `showLink` property, and while we could of course enable that on type=text, we then still need to add a protocol if missing, to make the link work, and then it starts being a little weird to just use the `text` type… Doing it this way keeps things a little more separate, and hopefully lower the risk of collisions when making changes in the future.

maltja (approved):

maltja (approved):

PR #1130, opened 2021-02-08
fix(collapsible-section): fix chopped letters in section-headers

maltja:
I think this bug was introduced by using `overflow:hidden` in fix(collapsible-section): use better default colors for header and body 7b7527fe02144a0b8561a7f0192cf3b569d5d569 . I don't know if there was a reason for adding overflow hidden but everything seems to work when remove it. 

fix Lundalogik/crm-feature#1739

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/collapsible-section/collapsible-section.scss
+++ b/src/components/collapsible-section/collapsible-section.scss
@@ -1,4 +1,5 @@
 @import '../../style/internal/mdc-variables';
+@import '../../style/mixins';
 
 /**
  * @prop --closed-header-background-color: background color for header when closed
@@ -49,12 +50,11 @@
 }
 
 .section__header__title {
+    @include truncate-text;
+    line-height: normal;
+
     justify-self: flex-start;
     padding-right: pxToRem(12);
-
-    overflow: hidden;
-    white-space: nowrap;
-    text-overflow: ellipsis;
 
     user-select: none; // mostly to improve experience on Android, where tapping on sections selects the text too
 }


Kiarokh:
@maltja  the `overflow: hidden;` is a part of a 3-lines CSS trick which makes text be laid out on one line and  be truncated. Without that, this is what happens:

https://user-images.githubusercontent.com/35954987/107485498-22e1a380-6b84-11eb-9d6b-8dfe8850ec61.mov

But it's completely fine not to know that. Which is why I made this mixin for it in the `mixins.scss` file. When the mixin is used, it'll be easier to decipher the code. 
````scss
@mixin truncate-text() {
    overflow: hidden;
    white-space: nowrap;
    text-overflow: ellipsis;
}
````

I'll use the mixin now instead, and then I'll see if we can find a better solution for this, most probably using a taller `line-height` ;)

Kiarokh:
OK, now I've pushed a `fixup` @maltja, you can check it out.

maltja:
> @maltja the `overflow: hidden;` is a part of a 3-lines CSS trick which makes text be laid out on one line and be truncated. Without that, this is what happens:
> 
>  Screen.Recording.2021-02-10.at.09.38.23.mov 
> But it's completely fine not to know that. Which is why I made this mixin for it in the `mixins.scss` file. When the mixin is used, it'll be easier to decipher the code.
> 
> ```css-scss
> @mixin truncate-text() {
>     overflow: hidden;
>     white-space: nowrap;
>     text-overflow: ellipsis;
> }
> ```
> 
> I'll use the mixin now instead, and then I'll see if we can find a better solution for this, most probably using a taller `line-height` ;)

Ah, I see, thanks for explaining it 😅

Kiarokh:
in future, I'll make a new PR, replacing the mixin with those three lines everywhere in the code.

Kiarokh (approved):

PR #1136, opened 2021-02-10
feat(select): enable adding icons to select options

Kiarokh:
according to size rhythm guidelines
fix: https://github.com/Lundalogik/crm-feature/issues/1790
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [x] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/select/examples/select-with-icons.tsx
+++ b/src/components/select/examples/select-with-icons.tsx
@@ -1,0 +1,70 @@
+import { Option } from '@limetech/lime-elements';
+import { Component, h, State } from '@stencil/core';
+/**
+ * Select with icons for options
+ */
+@Component({
+    shadow: true,
+    tag: 'limel-example-select-with-icons',
+})
+export class SelectExample {
+    @State()
+    public value: Option;
+
+    private options: Option[] = [
+        {
+            text: 'Batman',
+            value: 'bat',
+            icon: 'batman_old',
+            iconColor: 'rgb(var(--color-black))',
+        },
+        {
+            text: 'Iron Man',
+            value: 'iron',
+            disabled: true,
+            icon: 'iron_man',
+            iconColor: 'rgb(var(--color-coral-default))',
+        },
+        {
+            text: 'Spider-Man',
+            value: 'spider',
+            icon: 'spiderman_head',
+            iconColor: 'rgb(var(--color-red-default))',
+        },
+        {
+            text: 'Superman',
+            value: 'super',
+            icon: 'superman',
+            iconColor: 'rgb(var(--color-blue-default))',
+        },
+        {
+            text: 'Wonder Woman',
+            value: 'wonder',
+            icon: 'wonder_woman',
+            iconColor: 'rgb(var(--color-yellow-darker))',
+        },
+    ];
+
+    constructor() {
+        this.onChange = this.onChange.bind(this);
+    }
+
+    public render() {
+        return (
+            <section>
+                <limel-select
+                    label="Favorite hero"
+                    helperText="If you see a lack of diversity, it's our icon-provider's fault"
+                    value={this.value}
+                    options={this.options}
+                    onChange={this.onChange}
+                />
+                <limel-example-value value={this.value} />
+            </section>
+        );
+    }
+
+    private onChange(event) {
+        this.value = event.detail;
+    }
+}
--- a/src/components/select/select.template.tsx
+++ b/src/components/select/select.template.tsx
@@ -80,8 +80,11 @@ const SelectValue: FunctionalComponent<
             class={containerClassList}
             onKeyPress={props.onTriggerPress}
         >
-            <div class="limel-select__selected-text">
-                {getSelectedText(props.value)}
+            <div class="limel-select__selected-option">
+                {getSelectedIcon(props.value)}
+                <span class="limel-select__selected-option__text">
+                    {getSelectedText(props.value)}
+                </span>
             </div>
             <i class="mdc-select__dropdown-icon" />
             <label class={labelClassList}>{props.label}</label>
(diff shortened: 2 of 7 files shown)


jensgustafsson (approved):

src/components/select/select.template.tsx
```diff
@@ -210,3 +213,16 @@ function getSelectedText(value: Option | Option[]): string {
 
     return value.text;
 }
+
+function getSelectedTextIcon(value: any) {
+    if (!value) {
```
adrianschmidt, line 218:
```suggestion
    if (!value || !value.icon) {
```

That ought to fix the empty icon when there's no name. But I can test it while I review. Gotta do another review first though.

adrianschmidt (commented):
A few smallish comments. Good job!

src/components/select/examples/select.tsx
```diff
@@ -13,9 +13,31 @@ export class SelectExample {
...
+            iconColor: 'rgb(var(--color-coral-default))',
+        },
+        {
+            text: 'Batman',
+            value: 'bat',
+            icon: 'batman_old',
+            iconColor: 'rgb(var(--color-black))',
+        },
+        {
+            text: 'Superman',
```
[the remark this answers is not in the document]
adrianschmidt: Oh, and we're going to use icons in the narrow style select too, right? Maybe we don't need to have one example for every possible combination, but perhaps we should have one basic example, one example for each "extra", and then a single example at the bottom where all the extras are combined, and can be toggled on and off…? 🤔

What do you think?

(You don't have to build that, I could probably whip something up pretty quickly…)

src/components/select/select.template.tsx
```diff
@@ -210,3 +215,23 @@ function getSelectedText(value: Option | Option[]): string {
 
     return value.text;
 }
+
+function getSelectedTextIcon(value: any) {
```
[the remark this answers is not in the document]
Kiarokh: I have no objections. I'll change it in a new fixup

src/components/select/examples/select.tsx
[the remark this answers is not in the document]
Kiarokh: Thanks for all great ideas and feedback. I pushed a few new fixups. 

> a single example at the bottom where all the extras are combined, and can be toggled on and off…?

This is really nice to have. But up to you if you wanna make it. 
_Also I think we should create and use a common component for toggling options in examples across Lime Elements. We're inconsistently using switches, filter chips, buttons, etc… But that's another story ;)_

src/components/select/examples/select-with-icons.tsx
```diff
@@ -0,0 +1,72 @@
...
+        this.onChange = this.onChange.bind(this);
+        this.toggleEnabled = this.toggleEnabled.bind(this);
+    }
+
+    public render() {
+        return (
+            <section>
+                <limel-select
+                    label="Favorite hero"
+                    helperText="If you don't see a heroine, it's our icon-provider's fault"
```
adrianschmidt, line 55: 😂👌

src/components/select/examples/select.tsx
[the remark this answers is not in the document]
adrianschmidt: You're welcome, and right back at ya!

Totally agree that we should have some consistency. Not sure how to actually implement such a component though… 🤔

src/components/select/examples/select-with-icons.tsx
Kiarokh: 😅

src/components/select/select.template.tsx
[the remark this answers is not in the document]
adrianschmidt: This one seems to have been missed, so I'm fixing it 🙂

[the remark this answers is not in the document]
adrianschmidt: Ah, ok, no I was just too fast 😂

src/components/select/examples/select-with-icons.tsx
adrianschmidt: I found Wonder Woman at least, so I added her and tweaked the helper-text 🤷‍♂️

src/components/select/examples/select.tsx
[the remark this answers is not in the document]
adrianschmidt: > > a single example at the bottom where all the extras are combined, and can be toggled on and off…?
>
> This is really nice to have. But up to you if you wanna make it.

Yeah… I'll not, for now at least. Too late in the day 😄

src/components/select/examples/select-with-icons.tsx
Kiarokh: `helperText="If you see just one heroine, it's our icon-provider's fault"` ?
😄

Kiarokh: Oh I see what you wrote. Nice! 
Maybe the new icons8 has some up to date icons 😄 we could probably also use a cat icon and pretend it's Catwoman :P

adrianschmidt: Yeah, I went with "lack of diversity", because they're not just men by overwhelming majority, they're also all white… 🙄

Kiarokh: Totally right! 👌 
This forces us to reuse the same cat icon and call it Black Panther then 😸 and change the helper text to:
`If you see a lack of diversity, it's both our icon-provider's and Hollywood's fault`

Kiarokh: I don't think Catwoman and Black Panther have any specific sign or icon.

adrianschmidt (approved):

Issue #1107, opened 2021-01-20
Use consumer overridable css variables for all* z-indexes


adrianschmidt:
Place the default values in a single file, so that it's easy to see what other values might need to be checked when making a change.

*) After closer inspection, it turns out pretty much all z-indexes that a consumer might wish to override are coming from MDC. That by no means makes them impossible to override, but it requires a little more work. So the PR #1137 collects all z-indexes we define ourselves in a central file, but does not make them all overridable. It does add a consumer-overridable `--dialog-z-index` to the `limel-dialog` component though.

adrianschmidt:
@eketorp I'm putting this in the maintenance column on the cool down board, but also in the triage inbox, in case you fancy putting it in a cycle or something 🙂

Issue #1020, opened 2020-11-03
Reasonable z-index for limel-dialog


anderssonjohan:
Question: Should we increase the z-index of limel-dialog?

When using limel-dialog, it's easy to have it displayed underneath something else because of the quite low z-index value used. The dialog uses the value 7, the same as the main navigation of Lime CRM Webclient.
Other frameworks, like Bootstrap, uses 1040 for modals+backdrops. Also, if you use tooltip components etc you need to have even higher z-index to have them render properly to not be hidden by the modal.
The next question is if a modal should render over something like the main navigation of Lime CRM and the dialog backdrop shouldn't?

My humble thoughts about z-indexes among these components, ranked in highest-first order:

* (tooltips)
* menu
* dialog (and FABs, off-canvas menus, but neither should render at the same time)
* fixed positioned navigation (like Kompendium)
* other content

Some simple examples of this issue can be seen with Kompendium running the fullscreen dialog example, where the main navigation uses a higher z-index compared to the limel-dialog's and is rendered on top of the fullscreen dialog.

<img width="742" alt="Screenshot 2020-11-03 at 22 05 02" src="https://user-images.githubusercontent.com/435885/98040575-c5b9b180-1e20-11eb-88fe-0132f0d68c0a.png">
<img width="303" alt="Screenshot 2020-11-03 at 22 04 38" src="https://user-images.githubusercontent.com/435885/98040579-c7837500-1e20-11eb-9a5b-625f3d9eca5c.png">

Some more reading/background on the z-index (wars, if you will):
https://coder-coder.com/z-index-isnt-working/
 

## Environment

- **lime-elements** version: 29.4.3
- Framework used: 
- Logs:

adrianschmidt:
Yeah the 🤬🤬🤬🤬🤬🤬🤬🤬🤬🤬🤬🤬🤬🤬🤬🤬 z-index is always a 🤬 problem. We've had a low z-index because it was needed for something, I don't remember exactly why, if it was lime-webclient that required it or conflict with docz that we used before kompendium or something.

This is yet another of those over-arching design questions we need to do some work on some time.

I'm perfectly happy not to have a longwinded discussion about this during our dev-cycle, but if anyone else fancies it, then go ahead.

adrianschmidt:
I just added #1107, which will make z-indexes overridable by the consumer. We may still want to tweak our default values after that, but it's a good starting point for very little effort.

adrianschmidt:
[...]

Issue #979, opened 2020-10-07
Select menu appears underneath dialog (seems to only be an issue in the docs)


adrianschmidt:
## Current behavior

Update: It seems like this isn't a problem when lime-elements is used in lime-webclient, so this might only happen in the kompendium docs.

When a limel-select is placed inside a limel-dialog, and the select menu is opened, the menu appears underneath the dialog and the dialog scrim:

![image](https://user-images.githubusercontent.com/154725/95314221-525c6880-0891-11eb-9b10-cb47e7d0c3da.png)

## Expected behavior

The menu should be on top:

![image](https://user-images.githubusercontent.com/154725/95314353-7ddf5300-0891-11eb-8c26-12779ce51776.png)

## Environment

- **lime-elements** version: 28.5.6 (verified ok in 28.5.5)
- Framework used: bug exists in docs [example](http://lime-documentation-lime-elements.s3-website-eu-west-1.amazonaws.com/versions/28.5.6/#/component/limel-select)
- Logs:

anderssonjohan:
Did not see this issue when writing #1020, probably a related z-index war! :)

adrianschmidt:
I just added #1107, which will make z-indexes overridable by the consumer. We may still want to tweak our default values after that, but it's a good starting point for very little effort.

PR #1137, opened 2021-02-11
refactor z-indexes, add css property `--dialog-z-index` to `limel-dialog`, fix some issues in docs, add css property `--dropdown-z-index` to `limel-select`, `limel-menu`, and `limel-picker`

adrianschmidt:
fix: #1107
fix: #1020
fix: #979

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/date-picker/date-picker.scss
+++ b/src/components/date-picker/date-picker.scss
@@ -1,5 +1,7 @@
+@import '../../style/internal/variables';
+
 limel-portal {
-    z-index: -999999;
+    z-index: $limel-date-picker--portal--z-index;
     position: absolute;
     top: 0;
     bottom: 0;
--- a/src/components/date-picker/date-picker.scss
+++ b/src/components/date-picker/date-picker.scss
@@ -1,14 +1,5 @@
 @import '../../style/internal/variables';
 
-limel-portal {
-    z-index: $limel-date-picker--portal--z-index;
-    position: absolute;
-    top: 0;
-    bottom: 0;
-    right: 0;
-    left: 0;
-}
-
 div.container {
     position: relative;
 }
--- a/src/components/menu/menu.tsx
+++ b/src/components/menu/menu.tsx
@@ -84,6 +84,9 @@ export class Menu {
     }
 
     public render() {
+        const dropdownZIndex = getComputedStyle(this.host).getPropertyValue(
+            '--dropdown-z-index'
+        );
         const portalClasses = {
             'limel-portal--fixed': this.fixed,
         };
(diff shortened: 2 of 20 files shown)


anderssonjohan (dismissed):

src/components/date-picker/date-picker.scss
```diff
@@ -1,5 +1,7 @@
+@import '../../style/internal/variables';
+
 limel-portal {
```
Kiarokh, line 3: this whole block of styles with `limel-portal` is unnecessary now, since we have moved these styles to `portal.scss` line 1. This happened after the latest fixes of the portal. 

Also, there is no need for having a `z-index` here really I think. Probably `pointer-events: none;` would have been a better solution from the beginning to avoid blocking interactions (which is now a part of styles in `portal.scss`).

src/components/table/partial-styles/tabulator-arrow.scss
```diff
@@ -25,7 +25,7 @@
         bottom: 0;
     }
     &:before {
-        z-index: 1;
+        z-index: $limel-table--tabulator-arrow--before--z-index;
```
Kiarokh, line 28: I've double checked now. There is no need for `z-indexes` for `:before` and `:after` here for the arrows, as the browser renders them already correctly. Checked with latest chrome and firefox. Let's remove them.

adrianschmidt: ⚡

src/components/date-picker/date-picker.scss
adrianschmidt: ⚡

Kiarokh (approved):

Kiarokh:
I approved but cannot merge

src/components/menu/menu.tsx
```diff
@@ -84,6 +84,9 @@ export class Menu {
     }
 
     public render() {
+        const dropdownZIndex = getComputedStyle(this.host).getPropertyValue(
+            '--dropdown-z-index'
+        );
```
adrianschmidt, line 89: Ping @jgroth: I'd like your opinion on whether you think this is ok to do on every render…

Maybe we should just do it when the component loads or something? You're probably not going to change it throughout the lifetime of the component, right? 🤔

jgroth: This is more safe in case it would change, but I agree that it feels very unlikely that you would change it once it has been set.

Also, I don't think that this would impact performance in a way that you would notice it. The component will not render several times per second, more likely only when you open/close it.

adrianschmidt: Ok, that's great. Then let's go with this for now 👍

PR #1138, opened 2021-02-12
feat(slider): color slider to visualize percentage value

maltja:
fix Lundalogik/crm-feature#1636

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/widgets/slider.ts
+++ b/src/components/form/widgets/slider.ts
@@ -13,6 +13,7 @@ export class Slider extends React.Component {
         const schema = props.schema;
         let factor;
         let unit;
+        const additionalProps = getAdditionalProps(props.schema);
 
         if (isPercent(schema)) {
             factor = 100; // eslint-disable-line no-magic-numbers
@@ -32,6 +33,7 @@ export class Slider extends React.Component {
                 step: schema.multipleOf,
                 factor: factor,
                 unit: unit,
+                ...additionalProps,
             },
         });
     }
@@ -52,4 +54,17 @@ function isPercent(schema: any): boolean {
     return (
         schema.multipleOf < 1 && schema.minimum === 0 && schema.maximum === 1
     );
+}
+
+function getAdditionalProps(schema: any) {
+    let props: any = {};
+
+    if (schema.lime?.component?.props) {
+        props = {
+            ...props,
+            ...schema.lime.component.props,
+        };
+    }
+
+    return props;
 }
--- a/src/components/slider/examples/slider-multiplier-percentage-colors.tsx
+++ b/src/components/slider/examples/slider-multiplier-percentage-colors.tsx
@@ -1,0 +1,49 @@
+import { Component, h, State } from '@stencil/core';
+
+/**
+ * With percentage colors
+ * You can add the `displays-percentage-colors` class to your slider component
+ * and it will automatically visualize current percentage colors in real-time.
+ *
+ * The colors change with intervals of 10 as users drags the slider pin.
+ * The color spectrum is not modifiable, and looks like red → orange → yellow
+ * → green → teal.
+ */
+@Component({
+    tag: 'limel-example-slider-multiplier-percentage-colors',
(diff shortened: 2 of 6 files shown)


src/components/form/widgets/slider.ts
```diff
@@ -32,6 +33,7 @@ export class Slider extends React.Component {
                 step: schema.multipleOf,
                 factor: factor,
                 unit: unit,
+                ...additionalProps,
```
adrianschmidt, line 36: I think this, in combination with https://github.com/Lundalogik/lime-webclient/pull/1704/files#diff-1fa34b1dd5ce6645bced6e46655ca008385d2f1784a20b1d265c1a6203bd07a8R79, will overwrite any other classes set on `<limel-slider>`… for example if the component would set any classes on itself via the `Host` element…

Perhaps it's not actually a problem though… it's probably not work making a workaround for at the moment anyway… 🤔

It's internal implementation regardless, so we can just change it later if it turns out to be a problem…

src/components/slider/examples/slider-multiplier-percentage-colors.tsx
```diff
@@ -0,0 +1,49 @@
+import { Component, h, State } from '@stencil/core';
+
+/**
+ * With percentage colors
+ * You can add the `displays-percentage-colors` class to your slider component
+ * and it will automatically visualize current percentage colors in real-time.
+ *
+ * The colors change with intervals of 10 as users drags the slider pin.
+ * The color spectrum is not modifieable, and looks like red → orange → yellow
```
adrianschmidt, line 9:
```suggestion
 * The color spectrum is not modifiable, and looks like red → orange → yellow
```

adrianschmidt, line 9: Since the colours are defined on `:host` in `percentage-color.scss`, I'm pretty sure they can actually be modified… but maybe we don't want to document that in case we want to change the implementation?

adrianschmidt: ⚡

adrianschmidt (approved):

adrianschmidt (approved):

src/components/slider/examples/slider-multiplier-percentage-colors.tsx
maltja: I don't know. I think @Kiarokh said that we would have to write the css a little bit different so that it can be modified.

adrianschmidt: Hmm… 🤔

Regardless, it's probably best to claim it's not possible to modify to begin with, then if we get requests to make it possible to modify, we can consider if we just wish to expose the ten steps we currently have, or make the steps modifiable too, and go from there, without potentially having to make a breaking change 👍

Kiarokh: I can easily imagine that some people want other color spectrums, e.g. from cold to hot (blue to red) or even the same spectrum we have but in reverse. 

It's super easy to make the colors externally modifiable. It's only some CSS work and a bit of additional documentation.

However, it will be only for intervals of 10. So they can't for example switch to another color when they reach 35, and to another color at 75.

If people ask for it, we can easily add it. (or we just add this possibility in future ourselves)

adrianschmidt: It's already possible:

image

But yes, the variable has to be defined right on the `limel-slider` element, and by changing the CSS a little we can make it possible to override the colours by defining the variables on any ancestor element.

Regardless, let's keep that and the documentation as an easy feature to add in the future 🙂

PR #1145, opened 2021-02-17
docs(popover): explain how to pop on the correct side

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/1849
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/popover/popover.tsx
+++ b/src/components/popover/popover.tsx
@@ -39,6 +39,16 @@ import { portalContains } from '../portal/contains';
  * When a popover is displayed, interactions with other controls are blocked,
  * until user dismisses the popover.
  *
+ * :::warning
+ * In order for the popover to pop on the correct side of the element that
+ * triggers it, you need to wrap both of them in a container and make sure that
+ * the container:
+ * 1. is visually as big as the element that triggers the popover
+ * 1. has a `position` property specified (this normally can be `position: relative;`).
+ *
+ * Check the example to see how `:host` is displayed as `inline-block` to be as
+ * big as the button which is inside it.
+ * :::
  *
  * ## Layout
  * Popovers has only one slot in which you can import a custom web-component.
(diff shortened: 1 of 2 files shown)


src/components/popover/popover.tsx
```diff
@@ -39,6 +39,16 @@ import { portalContains } from '../portal/contains';
  * When a popover is displayed, interactions with other controls are blocked,
  * until user dismisses the popover.
  *
+ * :::warning
+ * In order for the popover to pop on the correct side of the element that
+ * triggers it, you need to wrap both of them in a container and make sure that
+ * the container:
+ * 1. is visually as big as the element that triggers the popover
```
adrianschmidt, line 46: Can't we provide the container ourselves, and put the trigger in a slot. I'm pretty sure that's what we do in limel-menu, and that works just fine…

Then we wouldn't need this big warning.

maltja: We also thought about that. So maybe we should try to do that instead 🙂

Kiarokh: I don't know how or where we should do this :/ can you suggest a solution in code?

maltja: I can try to do it 🙂

adrianschmidt: Give me a holler tomorrow if you don't manage, but if you look at limel-menu, I think it'll be pretty straightforward 👍

adrianschmidt:
I'll be adding a different solution to this problem.

PR #1147, opened 2021-02-18
Add trigger slot to popover

maltja:
fix: Lundalogik/crm-feature#1849

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [x] Firefox
- [x] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/popover/examples/popover.tsx
+++ b/src/components/popover/examples/popover.tsx
@@ -16,22 +16,27 @@ export class PopoverExample {
 
     public render() {
         return [
-            <limel-button
-                primary={true}
-                label="Click me!"
-                onClick={this.openPopover}
-            />,
             <limel-popover open={this.isOpen} onClose={this.onPopoverClose}>
+                <limel-button
+                    slot="trigger"
+                    primary={true}
+                    label="Click me!"
+                    onClick={this.openPopover}
+                />
                 <p style={{ margin: '0.5rem 1rem' }}>Content</p>
             </limel-popover>,
         ];
     }
 
-    private openPopover() {
+    private openPopover(event: MouseEvent) {
+        event.stopPropagation();
+        console.log('opening');
         this.isOpen = true;
     }
 
-    private onPopoverClose() {
+    private onPopoverClose(event: CustomEvent) {
+        event.stopPropagation();
+        console.log('closing');
         this.isOpen = false;
     }
 }
--- a/src/components/popover/popover.tsx
+++ b/src/components/popover/popover.tsx
@@ -102,7 +102,7 @@ export class Popover {
     private setupClickHandler() {
         if (this.open) {
             document.addEventListener('click', this.globalClickListener, {
-                passive: true,
+                capture: true,
             });
         } else {
             document.removeEventListener('click', this.globalClickListener);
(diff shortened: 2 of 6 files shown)


src/components/popover/popover.tsx
```diff
@@ -114,7 +114,7 @@ export class Popover {
     private setupClickHandler() {
         if (this.open) {
             document.addEventListener('click', this.globalClickListener, {
-                passive: true,
+                capture: true,
```
adrianschmidt, line 117: `capture: true` makes this listener activate in the "capture" phase of the event lifecycle. Quirksmode has a [good article](https://www.quirksmode.org/js/events_order.html) on the difference between event capturing and event bubbling.

In short, it means our global click lister will be called before any click handlers that are attached to any other elements. Because of that, we get the chance to detect if the click was outside the popover, close the popover and stop the event from propagating further, before the event even reaches any other components. This acts like a scrim, without requiring any extra element to catch the click.

adrianschmidt: The reason I removed `passive: true` is that I don't want to risk the event propagating before we get the chance to stop it. It did work even with `passive: true`, but it felt a bit paradoxical, so I removed it to not risk any race conditions. According to the [documentation](https://stenciljs.com/docs/events#passive), Stencil will make the listener passive if it can detect that it can do so, so this listener may be made passive by Stencil anyway. And, passive listeners are mostly important when listening to scroll-events anyway. A used won't _click_ hundreds of times a second 😅

jgroth: The passive stuff from the Stencil docs probably only applies if the `@Listen` decorator is used, not if we use `addEventListener` directly :slightly_smiling_face:

adrianschmidt: facepalm-sequence

Yeah, obviously 😂

src/components/popover/examples/popover.tsx
```diff
@@ -16,22 +16,27 @@ export class PopoverExample {
...
-                label="Click me!"
-                onClick={this.openPopover}
-            />,
             <limel-popover open={this.isOpen} onClose={this.onPopoverClose}>
+                <limel-button
+                    slot="trigger"
+                    primary={true}
+                    label="Click me!"
+                    onClick={this.openPopover}
+                />
```
adrianschmidt, line 25: Note that this needs doing in the table too. (This would be a breaking change, if not for the fact that we haven't published the docs for this component yet.)

maltja: It will still work with the old way of doing it in lime-crm-components so it is not a breaking change. However, portal will not get the correct dimension and therefore the portal's container will sometimes be misplaced.

adrianschmidt (approved):
Tested and approved by @maltja 😊

adrianschmidt (dismissed):
Tested and approved by @maltja 😊

adrianschmidt (approved):

src/components/popover/examples/popover.tsx
Kiarokh: sounds like we need to make a new issue for it, if you guys haven't already :)

adrianschmidt: It also screws up the alignment of the columns headers a little:
image

(The reason being that there's both an icon, and a space for the trigger, instead of just the space for the trigger with the icon _inside_ it 😄

adrianschmidt: Did either of you add an issue by the way? 😄

adrianschmidt: Fixed in Lundalogik/lime-crm-components#339

PR #1152, opened 2021-02-24
WIP trying to find the reason readonly doesn't work for limel-picker inside limel-form

adrianschmidt:
The `readonly` and `disabled` props for `limel-picker` works just fine when the component is used directly, but when used inside `limel-form`, those properties are not passed on to the component, regardless of whether the properties are set on the "root" of the field config, or in the `lime.component.props` section.

So far I've edited the limel-form examples to show only a single example with a single picker, and I've added loads of console.logs, to be able to see where the props "disappear".

As can be seen in the following screenshot, when `baseAdapter` renders the custom component, properties are available under `elementProps.schema`, but they have not been set at the root level of `elementProps`. I'm not quite sure were would be the best place to add them…? Maybe in the constructor on lines 60 to 76 in `base-adapter.ts`, before the call to `super` on line 67? 

image

fix: Lundalogik/crm-feature#1854

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/adapters/base-adapter.ts
+++ b/src/components/form/adapters/base-adapter.ts
@@ -66,6 +66,16 @@ export class LimeElementsAdapter extends React.Component<any, any> {
     ) {
         super(props);
 
+        if (props.elementProps) {
+            if (props.elementProps.formInfo?.schema?.disabled || props.elementProps.formInfo?.schema?.lime?.component?.props?.disabled) {
+                props.elementProps.disabled = true;
+            }
+
+            if (props.elementProps.formInfo?.schema?.readonly || props.elementProps.formInfo?.schema?.lime?.component?.props?.readonly) {
+                props.elementProps.readonly = true;
+            }
+        }
+
         this.state = {
             prevEvents: {},
         };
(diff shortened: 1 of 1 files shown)


adrianschmidt:
> Documentation has been published to [lundalogik.github.io/lime-elements/versions/PR-1152](https://lundalogik.github.io/lime-elements/versions/PR-1152/)

This is brilliant! Already a use for your great hack @jgroth!

jgroth:
> > Documentation has been published to [lundalogik.github.io/lime-elements/versions/PR-1152](https://lundalogik.github.io/lime-elements/versions/PR-1152/)
> 
> This is brilliant! Already a use for your great hack @jgroth!

Nice! :partying_face:

jensgustafsson:
This seems to be somewhat related to this https://github.com/Lundalogik/crm-feature/issues/1958

PR #1153, opened 2021-02-24
Feat file viewer

Kiarokh:
fix: https://github.com/Lundalogik/hack-tuesday/issues/169

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


Kiarokh:
image
it doesn’t seem safari understands the fullscreen buttons, or the `:fullscreen` selector

adrianschmidt:
Hey peeps! This looked amazing on the demo!

# 👏👏👏👏👏

How far away would you say this is from being something we could actually consider reviewing and merging? Like:

- [ ] Dude, it's so hackish, it'll be faster to start a clean slate and build it properly
- [ ] We wanna get there, but it still needs a _lot_ of work
- [ ] It's some work but not too bad, maybe by the end of next cooldown or something?
- [ ] Give me a day or two…
- [ ] Hang on, let me just squash these fixups!

src/components/file-viewer/file-viewer.tsx
```diff
@@ -0,0 +1,225 @@
...
+     * Title of the button that exists the fullscreen mode
+     */
+    @Prop({ reflect: true })
+    public titleExitFullscreen: string = 'Exit fullscreen';
+
+    /**
+     * Message which will be displayed when the file type is not supported
+     */
+    @Prop({ reflect: true })
+    public messageNoFileTypeSupport: string = 'Cannot display this file!';
```
jgroth, line 84: We need to take a look at how translations are handled in general in Lime Elements, and I'm not sure this is the best approach to have a lot of props on the components. We have some other components that require translations as well, e.g. the date pickers and some other component that I have forgotten about right now :sweat_smile: 
For the date picker we have translation files in the `src/translations` folder. We should probably do it the same way here for now until we decide how to do it in general :slightly_smiling_face:

src/components/file-viewer/file-viewer.tsx
```diff
@@ -0,0 +1,225 @@
...
+     * Alternative text for assistive technologies and screen readers, displayed for image files
+     */
+    @Prop({ reflect: true })
+    public alt: string;
+
+    /**
+     * The file type
+     */
+    @Prop({ reflect: true })
+    public type: string;
```
jgroth, line 54: This is the MIME type? We should clarify that in the documentation :slightly_smiling_face:

src/components/file-viewer/file-viewer.tsx
Kiarokh: That's a good point. We thought this was the way we handle them. But if there is a better way, let's do it ;)

src/components/file-viewer/file-viewer.tsx
Kiarokh: yes

Kiarokh:
> Hey peeps! This looked amazing on the demo!
> 
> # 👏👏👏👏👏

I think @anderssonjohan can answer your question. He has some additional plans for the first release.

src/components/file-viewer/file-viewer.tsx
anderssonjohan: @jgroth Either rename it to `contentType` (because https://github.com/Lundalogik/lime-elements/blob/ba612ac15dc4722b14880cd1a55e364009765996/src/components/file/file.tsx#L223) or keep the property name. We could steal this text that documents the property https://developer.mozilla.org/en-US/docs/Web/API/File/type#value or just use the same text as on FileInfo (https://github.com/Lundalogik/lime-elements/blob/fd23f9a530ed689f8c9509b84000db0159ab1971/src/components/file/file.types.ts#L17)

src/components/file-viewer/file-viewer.tsx
anderssonjohan: @jgroth resolved! [`9e61622` (#1153)](https://github.com/Lundalogik/lime-elements/pull/1153/commits/9e61622fb87c26630735d28758bcba43451b309c)

Kiarokh:
Last commit solved the problem with CSS selectors on Safari. However we still can't switch to fullscreen on Safari. Should be an easy fix.

There is something I don't really like when writing SCSS, and targeting `:host`. You cannot use commas to make rules like below. 

````scss
:host(:fullscreen),
:host(:-webkit-full-screen) { 
   /* stuff */
}
````
you have to write them separately, and repeat everything like this:
````scss
:host(:fullscreen) {
   /* stuff */
}
:host(:-webkit-full-screen) { 
   /* repeating THE SAME stuff */
}
````

Do you know a way to make the code more beautiful? Because that's how I have to do it in the last commit. (not a big deal though).

**Comment by anderssonjohan** [comment]
> Last commit solved the problem with CSS selectors on Safari. However we still can't switch to fullscreen on Safari. Should be an easy fix.
> 
> There is something I don't really like when writing SCSS, and targeting `:host`. You cannot use commas to make rules like below.
> 
> ```css-scss
> :host(:fullscreen),
> :host(:-webkit-full-screen) { 
>    /* stuff */
> }
> ```
> 
> you have to write them separately, and repeat everything like this:
> 
> ```css-scss
> :host(:fullscreen) {
>    /* stuff */
> }
> :host(:-webkit-full-screen) { 
>    /* repeating THE SAME stuff */
> }
> ```
> 
> Do you know a way to make the code more beautiful? Because that's how I have to do it in the last commit. (not a big deal though).

I also gave this a quick google and didn't have the time to complete it but I also came to the conclusion that the browser specific prefix should work.

**Comment by Kiarokh** [comment]
Hey @anderssonjohan @eketorp. 

Look what I found 🤩 We can use Google Docs' viewer and or Office 365 viewer, to see Word documents in read-only mode. This means we can add support for this, with no additional costs. (After all, this is a file viewer not editor :) )

**Live Examples:**
Google Docs Viewer: https://jsfiddle.net/7xr419yb/
Microsoft Office Viewer: https://jsfiddle.net/gcuzq343/

Microsoft's viewer seems to be better(?) since:
- it seems it can load larger files. 
- it can display open document formats such as `.odp`
 
I tried uploading different files with variety of formats on my own website, and linked it in the jsFiddle. Paste below examples in the jsfiddle and hit the `run` button.

## `.PPT`
````html
<iframe src='https://view.officeapps.live.com/op/embed.aspx?src=http://tinylittlemonsters.se/temp/test.ppt' width='100%' height='100%' frameborder='0'>This is an embedded <a target='_blank' href='http://office.com'>Microsoft Office</a> document, powered by <a target='_blank' href='http://office.com/webapps'>Office Online</a>.</iframe>
````


## `.PPTX` with animated gifs
Microsoft can load the large file
Animated gifs won't play, and that's a known Office viewer bug.
````html
<iframe src='https://view.officeapps.live.com/op/embed.aspx?src=http://tinylittlemonsters.se/temp/presentation.pptx' width='100%' height='100%' frameborder='0'>This is an embedded <a target='_blank' href='http://office.com'>Microsoft Office</a> document, powered by <a target='_blank' href='http://office.com/webapps'>Office Online</a>.</iframe>
````
But Google Drive cannot load the large file.
````html
<iframe src="https://docs.google.com/gview?url=http://tinylittlemonsters.se/temp/presentation.pptx&embedded=true" frameborder="0">
</iframe>
````
Google Drive cannot play animated gifs neither
````html
<iframe src="https://docs.google.com/gview?url=http://tinylittlemonsters.se/temp/test.pptx&embedded=true" frameborder="0">
</iframe>
````

## `.xlsx`
````html
<iframe src='https://view.officeapps.live.com/op/embed.aspx?src=http://tinylittlemonsters.se/temp/test.xlsx' width='100%' height='100%' frameborder='0'>This is an embedded <a target='_blank' href='http://office.com'>Microsoft Office</a> document, powered by <a target='_blank' href='http://office.com/webapps'>Office Online</a>.</iframe>
````

## `.odp`
````html
<iframe src='https://view.officeapps.live.com/op/embed.aspx?src=http://tinylittlemonsters.se/temp/test.odp' width='100%' height='100%' frameborder='0'>This is an embedded <a target='_blank' href='http://office.com'>Microsoft Office</a> document, powered by <a target='_blank' href='http://office.com/webapps'>Office Online</a>.</iframe>
````

Google drive cannot show ODPs:
````html
<iframe src="https://docs.google.com/gview?url=http://tinylittlemonsters.se/temp/test.odp&embedded=true" frameborder="0">
</iframe>
````

**Comment by Kiarokh** [comment]
I tested with these files too: `test.xml`, `test.zip`, `test.json` and `test.ics` just to see if these edge-case formats show up. But they didn't.

**Comment by anderssonjohan** [comment]
@Kiarokh Safari (not on ios, there's no support for that) fullscreen now works. Regarding the office/google docs viewers it only works with public documents. Does not work with blobs from the browser (preview before upload to a server) nor documents behind a protected server (lime crm, lime field etc, even though the latter serves a redirect to a temporary pre-signed url to an s3 bucket). So even though it could be useful for public documents when using lime-elements we can't use it for our CRM applications.

**Comment by eketorp** [comment]
> it only works with public documents.

This was sad to hear. Inability to view office files lowers the appetite for including this an application in general document handling for CRM. Nevertheless, I think it's a great feature to have within Lime Elements.

**Comment by anderssonjohan** [comment]
> > it only works with public documents.
> 
> This was sad to hear. Inability to view office files lowers the appetite for including this an application in general document handling for CRM. Nevertheless, I think it's a great feature to have within Lime Elements.

It's still usable for customizations that may provide a custom endpoint that allows the Microsoft viewer to read the file (preferably using some kind of pre-signed url/shared access signature). 

I think we should add a "embed service" property to allow the user to customize the URL used to embed different types of files. For example, if a customization provides an endpoint that is used to render a GPS track, it could be set up to render GPX/TCX files.
It could also allow for an internal endpoint if that is required in an on-prem solution.

One suggestion of how this could be defined:

```
<limel-file-viewer 
embed={{
  'google': 'pdf',
  'office': 'pptx,xlsx',
  'https://crm-server/limepkg-routeplotter/view/?url=': 'gpx,tcx',
  'https://crm-server/limepkg-document-renderer/view?url=': '*'
}}/>
```

If we use Office as the default it would be the same as using `embed={{ 'office': '*' }}` and to use Google Docs for everything (not handled by the component itself it would) you would have to specify `embed={{ 'google': '*' }}`. Using these names would just be an alias for using an URL prefix.

This still doesn't solve the use case for a file that only exists locally. This is the use-case when you have a file picker and want to show the local file before uploading it. This could be solved by uploading the picked file and exposing it temporarily and then either upload it again or move it into place. IMHO, this is outside the responsibility of a file viewer component. I think the file viewer should still support previewing local files (blobs) but with limited support for what file types it supports. Thoughts?

anderssonjohan:
Current status of this PR is that I'll add support for the MS Office File Viewer, then we can revisit it for review. I'll turn it into a draft until I have time to do that last part.

Kiarokh:
closing. a more up to date branch here: https://github.com/Lundalogik/lime-elements/pull/1430

PR #1155, opened 2021-02-25
fix(checkbox): use default material design style for checked state

Kiarokh:
this improves usability and makes it more
clear which box is checked

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -7,12 +7,11 @@
     display: flex;
 
     .mdc-checkbox {
-        @include mdc-checkbox-ink-color(primary);
         @include mdc-checkbox-container-colors(
             secondary,
             on-primary,
-            secondary,
-            on-primary
+            primary,
+            primary
         );
 
         &.mdc-checkbox--invalid {


src/components/checkbox/checkbox.scss
```diff
@@ -7,12 +7,11 @@
...
     .mdc-checkbox {
-        @include mdc-checkbox-ink-color(primary);
         @include mdc-checkbox-container-colors(
             secondary,
             on-primary,
-            secondary,
-            on-primary
+            primary,
+            primary
         );
```
adrianschmidt, line 15: If we're going back to using MDC's standard styling, I would assume that we can simply remove these includes completely… unless you already tried that?

Kiarokh: I tried to remove them all, but the checkbox turned to a dark grey (when checked). Not sure why :/

adrianschmidt: Hmm, interesting. But this one is needed, so it's 👌

adrianschmidt (approved):

PR #1157, opened 2021-02-25
fix(menu-surface): do not kill click events when clicking in an allowed element

### LINKED ISSUE #1062: limel-picker: Clicking the button to remove all chips when the dropdown is open breaks the picker

hannahu:
fix #1062

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu-surface/menu-surface.tsx
+++ b/src/components/menu-surface/menu-surface.tsx
@@ -32,6 +32,12 @@ export class MenuSurface {
     public open = false;
 
     /**
+     * Clicks in this element should not be prevented when the menu surface is open
+     */
+    @Prop()
+    public allowClicksElement: HTMLElement;
+
+    /**
      * Emitted when the menu surface is dismissed and should be closed
      */
     @Event()
@@ -106,10 +112,28 @@ export class MenuSurface {
     }
 
     private handleDocumentClick(event) {
-        if (this.open && !isDescendant(event.target, this.host)) {
-            this.dismiss.emit();
-            this.preventClickEventPropagation();
+        const elementPath = event.path || [];
+
+        if (!this.open) {
+            return;
         }
+
+        if (isDescendant(event.target, this.host)) {
+            return;
+        }
+
+        if (this.allowClicksElement) {
+            const clickedInAllowedElement = elementPath.includes(
+                this.allowClicksElement
+            );
+
+            if (clickedInAllowedElement) {
+                return;
+            }
+        }
+
+        this.dismiss.emit();
+        this.preventClickEventPropagation();
     }
 
     private handleResize() {
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -349,6 +349,7 @@ export class Picker {
             >
                 <limel-menu-surface
                     open={!!content}
+                    allowClicksElement={this.host}
                     style={{
                         '--menu-surface-width': '100%',
                         'max-height': 'inherit',


src/components/menu-surface/menu-surface.tsx
```diff
@@ -106,12 +112,32 @@ export class MenuSurface {
     }
 
     private handleDocumentClick(event) {
-        if (this.open && !isDescendant(event.target, this.host)) {
+        const elementPath = event.path || [];
+
+        const clickedInAllowedElement = this.isElementInPath(
+            elementPath,
+            this.allowClicksElement
```
adrianschmidt, line 119: Awesome! This was the once thing I had a comment about when I looked it over earlier, because a click in any _other_ picker would have also been allowed to propagate, and wouldn't close the picker 👍

adrianschmidt: Tiny question though: Rather than looking at the path, can't we use `isDescendant` for this check as well?

adrianschmidt:
```tsx
isDescendant(event.target, this.allowClicksElement);
```

hannahu: Then we can't know which element that was clicked since the event.target is the document?

adrianschmidt: Hmm… I'm pretty sure event.target should still be the element that was clicked, not the element on which the handler was registered…? Are you sure? 🤔

hannahu: Yes since it's shadow doms! :/

adrianschmidt: Whaa… really? Now I'm confused and curious! 😄

Imma fiddle around a bit, and see if I can understand what's happening… 🤔

adrianschmidt: Wow, yeah, you're absolutely right 👍

Good thing I found out about that little quirk before it bit me! 🙏

src/components/menu-surface/menu-surface.tsx
```diff
@@ -106,12 +112,32 @@ export class MenuSurface {
     }
 
     private handleDocumentClick(event) {
-        if (this.open && !isDescendant(event.target, this.host)) {
+        const elementPath = event.path || [];
+
+        const clickedInAllowedElement = this.isElementInPath(
```
adrianschmidt, line 117: We want to check `this.open` before we do this. At the moment, each menu-surface loops through the path for every single click, even when no surface-area is open. In the docs for `limel-select`, that's 10 times for each click 😅

We should also check if `this.allowClicksElement` is even truthy before we run this. If it's not, there's no reason to do the loop at all.

jgroth: I think we could also replace our own private method with something like t his :thinking: 
```ts
const clickedInAllowedElement = elementPath.includes(this.allowClicksElement);
```

src/components/menu-surface/menu-surface.tsx
```diff
@@ -31,6 +31,12 @@ export class MenuSurface {
     @Prop()
     public open = false;
 
+    /**
+     * Clicks in this element should not be prevented
```
jgroth, line 35:
```suggestion
     * Clicks in this element should not be prevented when the menu surface is open
```

src/components/menu-surface/menu-surface.tsx
hannahu: > I think we could also replace our own private method with something like this 
> 
> ```ts
> const clickedInAllowedElement = elementPath.includes(this.allowClicksElement);
> ```

ah yeah now we can do that!

hannahu: > We want to check `this.open` before we do this. At the moment, each menu-surface loops through the path for every single click, even when no surface-area is open. In the docs for `limel-select`, that's 10 times for each click 
> 
> We should also check if `this.allowClicksElement` is even truthy before we run this. If it's not, there's no reason to do the loop at all.

I love loops!!!

src/components/menu-surface/menu-surface.tsx
```diff
@@ -106,7 +112,23 @@ export class MenuSurface {
...
     private handleDocumentClick(event) {
-        if (this.open && !isDescendant(event.target, this.host)) {
+        const elementPath = event.path || [];
+
+        if (!this.open) {
+            return;
+        }
+
+        if (this.allowClicksElement) {
+            const clickedInAllowedElement = elementPath.includes(
```
hannahu, line 122: Fixup now done.
I know this row can be put in 121 as well but I think it's easier to follow what happens if we name the result here?

PR #1161, opened 2021-02-26
ci(pr-checks.yml): split publishing of docs into separate job, so failure does not fail the ci job

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -60,34 +60,34 @@ jobs:
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
 
-  docs:
-    name: Publish Docs
-    runs-on: ubuntu-latest
-    env:
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-      GH_USERNAME: lime-ci
-      GH_EMAIL: 617355+lime-ci@users.noreply.github.com
-      GH_TOKEN: ${{ secrets.CREATE_RELEASE }}
-      CI: true
-    steps:
-    - uses: actions/checkout@v2
-      with:
-        token: ${{ secrets.CREATE_RELEASE }}
-    - uses: actions/setup-node@v2
-      with:
-        node-version: '14'
-        check-latest: true
-    - run: npm i -g npm
-    - name: npm install
-      run: |
-        GH_TOKEN="${{ secrets.READ_PRIVATE_GITHUB_PACKAGES }}" ./generate_npmrc.sh
-        npm ci
-        rm .npmrc
-    - run: git config --global user.email "$GH_EMAIL"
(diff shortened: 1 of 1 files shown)


adrianschmidt:
@jgroth The publish docs script is often failing because another instance manages to push a new commit between the time the job checks out the branch and the time it pushes the commit it created. So I'm splitting the job out from the CI-job, so failing to publish the docs doesn't fail the CI-check.

Then I'll also comment out the PR docs publishing job for a little while, because I'm worried that it might cause the release of a new version to fail, and I'm not sure how messy that would be if it happens. Then I'll take a look at two things:

1. Can we reduce the time between checkout and push (or do anything else to make it less likely to fail)?
2. Can we give the publishing of released docs priority of some kind? I think we could probably just do a hard force-push when publishing from a release, but maybe there are better ways…?

jgroth:
> 1. Can we reduce the time between checkout and push (or do anything else to make it less likely to fail)?
> 2. Can we give the publishing of released docs priority of some kind? I think we could probably just do a hard force-push when publishing from a release, but maybe there are better ways…?

I guess we could do a `pull` right before we want to do a new commit to make sure everything is up to date. The docs generation probably takes a while, so we should either build it somewhere outside of the repository or stash all changes before we do the pull :thinking: 

This should at least work for all files that has not been modified, which I think should be all files apart from the one containing all the versions

adrianschmidt:
> > 1. Can we reduce the time between checkout and push (or do anything else to make it less likely to fail)?
> > 2. Can we give the publishing of released docs priority of some kind? I think we could probably just do a hard force-push when publishing from a release, but maybe there are better ways…?
> 
> I guess we could do a `pull` right before we want to do a new commit to make sure everything is up to date. The docs generation probably takes a while, so we should either build it somewhere outside of the repository or stash all changes before we do the pull 🤔
> 
> This should at least work for all files that has not been modified, which I think should be all files apart from the one containing all the versions

We are actually building outside the repo, in a sense, because what we do is that we create a separate clone of the repo, with the `gh-pages` branch checked out, in the `docsDist` folder. Then we build in the regular repo, and then copy the build output into `docsDist`. I've now added a pull that happens after we build, right before we start copying files. (Btw, that gives me an idea… I think we _copy_ the files at the moment, but maybe we could just _move_ them instead… that ought to be quicker… 🤔)

I've also added the option to use `--force` when pushing, and set semantic-release to use that flag when pushing, so that if there's a conflict when pushing while running the script as part of a release, we'll just override whatever was there before, losing what is likely a PR-version that had just been published.

PR #1163, opened 2021-02-26
feat(chip-set): make clear all chips button's label translatable

maltja:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -1,4 +1,4 @@
-import { Chip } from '@limetech/lime-elements';
+import { Chip, Languages } from '@limetech/lime-elements';
 import {
     MDCChipInteractionEvent,
     MDCChipSelectionEvent,
@@ -17,6 +17,7 @@ import {
     Watch,
 } from '@stencil/core';
 import { handleKeyboardEvent } from './chip-set-input-helpers';
+import translate from '../../global/translations';
 
 const SELECTED_CHIP_CLASS = 'mdc-chip--selected';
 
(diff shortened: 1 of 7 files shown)


src/components/chip-set/chip-set.tsx
```diff
@@ -107,6 +107,12 @@ export class ChipSet {
     @Prop({ reflect: true })
     public delimiter: string = null;
 
+    /**
+     * For chip-set of type `input`. The title and aria-label for the clear all chips button.
+     */
+    @Prop({ reflect: true })
+    public clearAllChipsLabel = 'Clear chips';
```
jgroth, line 114: I'm not sure we should add props for all translatable strings on the components, we need to take a look at the bigger picture regarding translations in lime elements :thinking: 

We have files with translations for the date picker right now, and we should probably use something similar here and then provide some way for the consumer to set their own translations if they want to override them.

maltja: That is true, Do you think I should do anything with this now or maybe we should decide how we want to do with translations in lime-elements in general. 🤔

anderssonjohan: @maltja I resolved the same comment from @jgroth about the file-viewer by using the translations we use in the date-picker. If we are going to use that for translations you could make the same changes here that I did in the following commit: [`449c03d` (#1153)](https://github.com/Lundalogik/lime-elements/pull/1153/commits/449c03d47b2ebaa95679522a61777b6531139a6a)

Note that this PR that adds Dutch to the game should probably be merged before this PR:
https://github.com/Lundalogik/lime-elements/pull/1164

anderssonjohan: @maltja Oh, I just saw your PR here as well and for a second I thought you had added the same change I did for lang nl. https://github.com/Lundalogik/lime-elements/pull/1162

jgroth (dismissed):

jgroth (approved):

PR #1165, opened 2021-03-01
feat(table): enable movable columns

jgroth:
fix Lundalogik/crm-feature#1863

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -18,6 +18,7 @@ const FIRST_PAGE = 1;
 /**
  * @exampleComponent limel-example-table
  * @exampleComponent limel-example-table-custom-components
+ * @exampleComponent limel-example-table-movable-columns
  * @exampleComponent limel-example-table-local
  * @exampleComponent limel-example-table-remote
  * @exampleComponent limel-example-table-activate-row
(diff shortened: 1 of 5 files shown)


src/components/table/table.tsx
```diff
@@ -182,14 +194,25 @@ export class Table {
...
         }
 
+        if (this.isReordered(newColumns, oldColumns)) {
+            return;
+        }
+
         this.tabulator.setColumns(this.getColumnDefinitions());
     }
 
+    private isReordered(newColumns: Column[], oldColumns: Column[]) {
```
Befkadu1, line 209: I am not sure if the function name is good name for what it does. This function returns a boolean whether all elements in the `newColumns`  exist in `oldColumns` . This doesn't show if there is a reorder or not.

This is how I check if there is reorder 

```
for (var i = 0; newColumns.length < i; i++) {
		if (newColumns[i] !== oldColumns[i]) return true;
	}
```
	
	My point is shortly, we probably give a better name for the function.

jgroth: I changed the name of the method

Befkadu1 (approved):

PR #1168, opened 2021-03-03
fix(table): prevent tall custom components break column headers UI

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/1859
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/partial-styles/header-component.scss
+++ b/src/components/table/partial-styles/header-component.scss
@@ -14,4 +14,5 @@
 
 .title-component-slot  {
     flex-shrink: 0;
+    max-height: pxToRem(16);
 }


src/components/table/partial-styles/header-component.scss
```diff
@@ -14,4 +14,5 @@
 
 .title-component-slot  {
     flex-shrink: 0;
+    max-height: pxToRem(16);
```
adrianschmidt, line 17: I'm a little conflicted about this… On the one hand, I think it's a good idea to have the component itself ensure the layout will be good. On the other hand, I think 16px might be a bit small for strict enforcement. The text element itself is 18 px tall for example.

So I think it may be better to leave to the consumer to get this right, which is the way it's solved in Lundalogik/lime-crm-components#339.

If you absolutely want this max-height anyway, then I'll go with your judgement, but if it's up to me, I say we don't add it.

adrianschmidt: (The end result with this PR vs Lundalogik/lime-crm-components#339 are pixel identical when the latter is used together with the latest `main` of lime-elements.)

Kiarokh: I see your point. Let's scrap this one for now and see if the need comes back in future ;)

PR #1170, opened 2021-03-03
fix(chip-set): update `--icon-color` default value to `rgb(var(--contrast-1100))`

Kiarokh:
Strong dominant colors make chips or buttons look like CTAs (Call To Action elements). This should be avoided and colors should be used to communicate meaningful messages or attract attention where needed.

Default styles should be neutral.

fix: https://github.com/Lundalogik/crm-feature/issues/1886

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -8,7 +8,7 @@
 
 /**
  * @prop --icon-background-color: Background color of the icon. Defaults to transparent.
- * @prop --icon-color: Color of the icon. Defaults to 54% black.
+ * @prop --icon-color: Color of the icon. Defaults to `rgb(var(--contrast-1100))`.
  * @prop --background-color: Background color of the field when type is set to input.
  * @prop --background-color-disabled: Background color of the field when type is set to input and the component is disabled or readonly.
  * @prop --input-chip-set-selected-chip-color: Color of the highlight around selected chips in input chip-sets.
@@ -69,7 +69,7 @@ $height-of-chip-set-input: pxToRem(36);
 limel-icon.mdc-chip__icon.mdc-chip__icon--leading {
     background-color: var(--icon-background-color, transparent);
     margin: 0 !important;
-    color: var(--icon-color, rgba(0, 0, 0, 0.54));
+    color: var(--icon-color, rgb(var(--contrast-1100)));
 }
 
 .mdc-chip-set {
--- a/src/components/chip-set/examples/chip-set-input.scss
+++ b/src/components/chip-set/examples/chip-set-input.scss
@@ -1,7 +1,4 @@
 limel-chip-set[type='input'] {
-    --icon-background-color: rgb(173, 173, 173);
-    --icon-color: white;
-
     flex-grow: 1;
     flex-shrink: 1;
     max-width: calc(100% - 7rem);
--- a/src/components/chip-set/examples/chip-set-input.tsx
+++ b/src/components/chip-set/examples/chip-set-input.tsx
@@ -49,11 +49,6 @@ export class ChipSetInputExample {
             this.createChip('Fish'),
         ];
 
-        this.value[0].iconColor = 'var(--lime-red)';
-        this.value[1].iconColor = 'var(--lime-orange)';
-        this.value[2].iconColor = 'var(--lime-green)';
-        this.value[3].iconColor = 'var(--lime-blue)';
-
         this.chipSetOnChange = this.chipSetOnChange.bind(this);
         this.onInteract = this.onInteract.bind(this);
         this.onInput = this.onInput.bind(this);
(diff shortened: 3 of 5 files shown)


Kiarokh:
image

src/components/chip-set/examples/chip-set-input.tsx
```diff
@@ -49,10 +49,11 @@ export class ChipSetInputExample {
...
 
-        this.value[0].iconColor = 'var(--lime-red)';
-        this.value[1].iconColor = 'var(--lime-orange)';
-        this.value[2].iconColor = 'var(--lime-green)';
-        this.value[3].iconColor = 'var(--lime-blue)';
+        this.value[0].iconFillColor = 'var(--lime-red)';
+        this.value[1].iconBackgroundColor = 'var(--lime-orange)';
+        this.value[2].iconBackgroundColor = 'var(--lime-green)';
+        this.value[3].iconFillColor = 'var(--lime-yellow)';
+        this.value[3].iconBackgroundColor = 'var(--lime-dark-blue)';
```
adrianschmidt, line 56: The example looks pretty bad with these styles, I think. Could we get it to look a little nicer? I think there needs to be a good balance between showing how to do things technically, and setting a good example for how things ought to look.

src/components/chip-set/examples/chip-set-input.scss
```diff
@@ -1,6 +1,5 @@
 limel-chip-set[type='input'] {
-    --icon-background-color: rgb(173, 173, 173);
-    --icon-color: white;
+    --icon-background-color: rgb(var(--contrast-300));
```
adrianschmidt, line 2: I think we should just remove the color from here, and let icons that doesn't have a style show the _actual_ default styling they would have otherwise.

src/components/chip-set/examples/chip-set-input.tsx
Kiarokh: 🤔 do you mean using CSS instead of writing things like this in the code?

src/components/chip-set/examples/chip-set-input.scss
Kiarokh: hmmm... the default is `transparent`. I thought we maybe want some more neutral icon color for the chip-set. But I don't have any strong argument to support why, except that it looked nicer (maybe because I'm so used to our current ui style?)

src/components/chip-set/examples/chip-set-input.tsx
adrianschmidt: No no, I just meant purely visually, I think this looks a bit all over the place:

image

Kiarokh: maybe we can reduce it to 1 example for each case? 
- only icon color
- only background color
- both icon and background color

Otherwise I don't have any problems with the example itself neither :)

src/components/chip-set/chip-set.scss
```diff
@@ -69,7 +69,7 @@ $height-of-chip-set-input: pxToRem(36);
 limel-icon.mdc-chip__icon.mdc-chip__icon--leading {
     background-color: var(--icon-background-color, transparent);
     margin: 0 !important;
-    color: var(--icon-color, rgba(0, 0, 0, 0.54));
+    color: var(--icon-color, rgb(var(--contrast-1100)));
```
adrianschmidt, line 72: This is the only change to the component itself, the rest is just changes to the examples. So I don't think "prevent default chips to appear visually so dominant" is a good description of the change here… I'll come up with something better, but I'm making this comment because I'll be logging of for the weekend now 🙂

Kiarokh: OK :)

Kiarokh (commented):
I think your changes are great @adrianschmidt  ;)

src/components/chip-set/chip-set.scss
adrianschmidt: Haha, I had _totally_ forgotten about this, so good thing I wrote that comment, and thanks for commenting back on it, so that I saw it again 😂 🙏

src/components/chip-set/examples/chip-set-input.scss
adrianschmidt: :zap:'ed by moving all colours to a separate example.

src/components/chip-set/examples/chip-set-input.tsx
adrianschmidt: Also ⚡'ed by moving all colours to a separate example.

src/components/chip-set/chip-set.scss
adrianschmidt: ⚡

adrianschmidt (approved):

PR #1172, opened 2021-03-04
fix(button): make buttons with has-reduced-presence really disappear

Kiarokh:
and do not occupy any visual width in the UI

fix: https://github.com/Lundalogik/crm-feature/issues/1890

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -85,18 +85,49 @@ button {
     }
 }
 
+$speed-of-reducing-presence: 0.3s;
+
+@keyframes reduced-presence {
+    0% {
+        opacity: 1;
+        transform: scale(1);
+    }
+    100% {
+        opacity: 0;
+        transform: scale(0.7);
+    }
+}
+
 :host(.has-reduced-presence) {
     button {
-        transition: opacity 300ms ease-in-out;
-
         &[disabled]:not(.loading):not(.just-loaded) {
-            opacity: 0;
+            animation: reduced-presence $speed-of-reducing-presence ease
+                forwards;
+            transition: padding $speed-of-reducing-presence ease,
+                min-width $speed-of-reducing-presence ease;
+            transition-delay: $speed-of-reducing-presence;
 
-            limel-icon,
+            padding: 0;
+            min-width: 0;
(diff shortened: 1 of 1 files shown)


src/components/button/button.scss
```diff
@@ -85,18 +85,43 @@ button {
...
-                // simultaneously fading out the button.
+            limel-spinner,
+            svg {
+                transition: all $speed-of-reducing-presence ease;
+                transition-delay: $speed-of-reducing-presence;
+
                 opacity: 0;
+                width: 0;
+                height: 0;
+                font-size: 0;
```
adrianschmidt, line 124: I think setting everything to zero width here is the reason for this ugly "jump":

https://user-images.githubusercontent.com/154725/109963923-42846b80-7ced-11eb-963f-6c7a3280c6f1.mov

Kiarokh: Oh really nice catch!

The ugly jump was because the font-size was animated too quickly or the CSS wasn't specific enough so the `.lable` didn't get the `transition` or something like that :) 
I pushed a fixup now.

https://user-images.githubusercontent.com/35954987/109974956-613d2f00-7cfa-11eb-86ef-8c05b851b997.mp4

adrianschmidt: Very nice!

Should we have a similar (but perhaps faster) animation when the button appears too? 🤔

adrianschmidt (approved):

src/components/button/button.scss
Kiarokh: Yes, it would be nice. I'll do it at some point. It's easier if the button gets a class as soon as it gets enabled in that case (I think) 🤔

adrianschmidt: > I'll do it at some point.

I'll take that to mean we'll merge this first?

Issue #1115, opened 2021-02-01
limel-input-field with autocompletions: dropdown can only be closed by selecting an autocompletion


adrianschmidt:
## Current behavior

https://user-images.githubusercontent.com/154725/106436343-474acb00-6474-11eb-9355-63ed42bc08ca.mov

1. Focus any limel-input-field with autocompletions.
2. Blur the element by clicking outside it, or tabbing away from it.

- The dropdown showing suggestions stays open.

## Expected behavior

The dropdown should close.

## Environment

- **lime-elements** version: v30.4.1
- Framework used: Can be reproduced in the _Input Field with Completions_ example in the docs
- Logs:

adrianschmidt:
This doesn't seem to be using limel-menu-surface or limel-portal or popper in any way, which it probably should 😄

eketorp:
Still there after today's release. Also expanding strangely in the viewport when doing the selection the second time

![ny person](https://user-images.githubusercontent.com/58422324/106945936-bcdec180-6728-11eb-89af-4fdd56ee2037.gif)

adrianschmidt:
Yeah, we haven't touched this at all.

adrianschmidt:
> expanding strangely in the viewport when doing the selection the second time

Not really. Since you already have "Dräng" in the field the second time, there are no other matching autocompletions, so the list is only one item long. If you empty the field, the dropdown should become taller again.

eketorp:
> [...] there are no other matching autocompletions

Ah! I always thought of this as a text field with options. Not as autocompletion.

PR #1173, opened 2021-03-05
fix(input-field): when using autocompletions, make it possible to close the dropdown

maltja:
fix Lundalogik/lime-elements#1115

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.scss
+++ b/src/components/input-field/input-field.scss
@@ -16,6 +16,7 @@
  */
 
 :host {
+    position: relative;
     display: block;
 }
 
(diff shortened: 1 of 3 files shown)


adrianschmidt (approved):
I found a few smaller issues, but since this version of the component was completely broken before, 1) this is a vast improvement, and 2) it doesn't seem like it was used by anyone, so fixing some small rough edges can probably wait.

The things I found:
- When tabbing out of the dropdown list, focus is reset (like we talked about on Slack).
- When arrowing down through the list, after a few items, you have to press the down-arrow twice to move to the next item. But once you've arrowed through the whole list, if you go back and forth, only one keypress is needed per step. It's pretty weird. This isn't new though, the same happens on main, so it doesn't have to be fixed in this PR.

adrianschmidt (approved):

Issue #1174, opened 2021-03-05
limel-file: Filter file types that are selectable


sampsonbryce:
## New feature motivation

In document templates, we only want users to be able to select docx files. 

## New feature description

A prop of file extensions/types that are selectable so that when the component opens your computers file picker it filters out all files not of the given type/extension

## New feature implementation

New prop `accept` that is passed directly to the underlying input element https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file

PR #1176, opened 2021-03-05
feat(file): add accept prop to filter selectable file types

sampsonbryce:
fix: #1174

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/file/examples/file-accepted-types.tsx
+++ b/src/components/file/examples/file-accepted-types.tsx
@@ -1,0 +1,40 @@
+import { FileInfo } from '@limetech/lime-elements';
+import { Component, h, State } from '@stencil/core';
+
+@Component({
+    tag: 'limel-example-file-accepted-types',
+    shadow: true,
+})
+export class FileAcceptedTypesExample {
+    @State()
+    private value: FileInfo = { filename: 'picture.jpg', id: 123 };
+
+    @State()
+    private required = false;
+
+    constructor() {
+        this.handleChange = this.handleChange.bind(this);
+        this.toggleRequired = this.toggleRequired.bind(this);
+    }
+
+    public render() {
+        return [
+            <limel-file
+                label="Attach only images (png, jpeg)"
+                onChange={this.handleChange}
+                required={this.required}
+                value={this.value}
+                accept="image/jpeg,image/png"
+            />,
+        ];
+    }
+
+    private handleChange(event: CustomEvent<FileInfo>) {
+        this.value = event.detail;
+        console.log('onChange', this.value);
+    }
+
+    private toggleRequired() {
+        this.required = !this.required;
+    }
+}
--- a/src/components/file/file.tsx
+++ b/src/components/file/file.tsx
@@ -49,6 +49,7 @@ const DEFAULT_ICON = 'note';
  *
  * @exampleComponent limel-example-file
  * @exampleComponent limel-example-file-custom-icon
+ * @exampleComponent limel-example-file-accepted-types
  */
 @Component({
     tag: 'limel-file',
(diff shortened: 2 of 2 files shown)


src/components/file/examples/file-accepted-types.tsx
```diff
@@ -0,0 +1,40 @@
+import { FileInfo } from '@limetech/lime-elements';
```
adrianschmidt, line 1: I added `fix: #1174` to the commit message and the PR description. That way, when the fix is released, the changelog will contain links to the commit and the issue, like [this](https://github.com/Lundalogik/lime-elements/releases/tag/v30.15.2).

adrianschmidt, line 1: I added `fix: #1174` to the commit message and the PR description. That way, when the fix is released, the changelog will contain links to the commit and the issue, like [this](https://github.com/Lundalogik/lime-elements/releases/tag/v30.15.2).

sampsonbryce: Thanks Adrian!

src/components/file/file.tsx
```diff
@@ -80,6 +81,14 @@ export class File {
     @Prop({ reflect: true })
     public disabled: boolean = false;
 
+    /**
+     * The accepted file types
+     *
+     * @see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#unique_file_type_specifiers
```
BregenzerK, line 87: Really nice and necessary to have this resource in the docs string. I noticed that I couldn't click it in the documentation. could that work if you replace `@see` with `@link`?

BregenzerK (commented):
I would approve the PR besides that one remark. @adrianschmidt is it ok if we merge it?

adrianschmidt:
> I would approve the PR besides that one remark. @adrianschmidt is it ok if we merge it?

Yeah, looks good to me! If `@link` doesn't work, maybe you can just add the link with the regular markdown syntax?

sampsonbryce:
@BregenzerK I changed to just a markdown link

BregenzerK (approved):

adrianschmidt:
Aside:

> > I would approve the PR besides that one remark. @adrianschmidt is it ok if we merge it?
> 
> Yeah, looks good to me! If `@link` doesn't work, maybe you can just add the link with the regular markdown syntax?

Ah, come to think of it, I think `@link` isn't for adding a url. I _think_ it's for adding a link to another component in the documentation.

Regardless, the markdown link works just fine 👌

src/components/file/file.tsx
```diff
@@ -80,6 +81,12 @@ export class File {
     @Prop({ reflect: true })
     public disabled: boolean = false;
 
+    /**
+     * The [accepted file types](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#unique_file_type_specifiers)
+     */
+    @Prop({ reflect: true })
+    public accept: string = '*';
```
adrianschmidt, line 88: (This isn't an issue I think should prevent merging this, just a small docs thing Johan and I might want to look into.)

Hmm… despite a default value being set here, it doesn't show up in the docs:

image

@jgroth Do you know off the top of your head what the reason might be? I've looked through some of our other components, but we don't have many string properties with default values. We do have `type` on limel-slider, which defaults to `''`, and that shows up as "Default value: `false`" in the docs… so I think the value is being handled incorrectly somewhere. I'm not sure if it's in Kompendium or in some lib we're using though…

image

image

PR #1180, opened 2021-03-09
fix(select): better visualize `focused` with keyboard

Kiarokh:
requires: https://github.com/Lundalogik/lime-elements/pull/1179

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/select/select.scss
+++ b/src/components/select/select.scss
@@ -58,12 +58,13 @@
 
         &:focus {
             outline: none;
-            .mdc-select__dropdown-icon {
-                box-shadow: var(--button-shadow-hovered);
-            }
+
             .mdc-floating-label {
                 color: var(--mdc-theme-primary);
             }
+        }
+        &:focus-visible {
+            box-shadow: var(--shadow-depth-8-focused) !important;
         }
 
         .mdc-floating-label {


src/components/select/select.scss
```diff
@@ -58,13 +58,14 @@
...
             outline: none;
-            .mdc-select__dropdown-icon {
-                box-shadow: var(--button-shadow-hovered);
-            }
+
             .mdc-floating-label {
                 color: var(--mdc-theme-primary);
             }
         }
+        &:focus-visible {
```
adrianschmidt, line 66: WOOT!

NICE!

Not supported in Safari yet, but that seems like something we can live without, right? And for Mobile Safari it doesn't seem like the lack of support should be much of an issue, since it's probably not _very_ common for people to connect keyboards and tab through fields on their iPads or iPhones.

Kiarokh: Tnx :) 

I think we can use this until the new IE (Safari) realizes that they should also support it 😄

src/components/select/select.scss
```diff
@@ -58,13 +58,14 @@
...
+            box-shadow: var(--shadow-depth-8-focused) !important;
```
adrianschmidt, line 67: I'm guessing the following isn't the intended look?

https://user-images.githubusercontent.com/154725/110787915-3872e780-826e-11eb-8269-81defc37a501.mov

adrianschmidt: (When there's no drop-shadow is when the element is focused using the keyboard.)

adrianschmidt: Ah‚ I now see that it requires #1179. I had missed that.

adrianschmidt (approved):

PR #1181, opened 2021-03-10
fix(tab-bar): prevent unintended clicks on the last tab

Kiarokh:
This happens when clicking fast on the `left` or `right` scroller buttons.
Now they stay in place while user hovers, but look kind of disabled.
They also disappear with a delay.
They're still clickable though, but that's because we can't access
the styles of `limel-icon-button` from outside.

fix: https://github.com/Lundalogik/crm-feature/issues/1862

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [x] Edge
- [x] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/tab-bar/partial-styles/tab-bar-scroller.scss
+++ b/src/components/tab-bar/partial-styles/tab-bar-scroller.scss
@@ -60,15 +60,15 @@
     position: relative;
     overflow: hidden;
 
-    &.scroll-left .scroll-fade.left,
-    &.scroll-left .scroll-button.left,
-    &.scroll-right .scroll-fade.right,
-    &.scroll-right .scroll-button.right {
+    &.can-scroll-left .scroll-fade.left,
+    &.can-scroll-left .scroll-button.left,
+    &.can-scroll-right .scroll-fade.right,
+    &.can-scroll-right .scroll-button.right {
         transform: translate3d(0, 0, 0);
     }
 
-    &.scroll-left:not(.scroll-right) .scroll-button.right,
-    &.scroll-right:not(.scroll-left) .scroll-button.left {
+    &.can-scroll-left:not(.can-scroll-right) .scroll-button.right,
+    &.can-scroll-right:not(.can-scroll-left) .scroll-button.left {
         opacity: 0.5;
         transition-delay: 0.5s;
     }
--- a/src/components/tab-bar/partial-styles/tab-bar-scroller.scss
+++ b/src/components/tab-bar/partial-styles/tab-bar-scroller.scss
@@ -19,9 +19,9 @@
         left: 0;
         background: linear-gradient(
             270deg,
-            rgba(246, 246, 247, 0) 0%,
-            rgba(246, 246, 247, 0.8) 40%,
-            rgba(246, 246, 247, 0.8) 100%
+            rgba($tab-background-color, 0) 0%,
+            rgba($tab-background-color, 0.8) 40%,
+            rgba($tab-background-color, 0.8) 100%
         );
     }
 
@@ -30,15 +30,15 @@
         right: 0;
         background: linear-gradient(
             90deg,
-            rgba(246, 246, 247, 0) 0%,
-            rgba(246, 246, 247, 0.8) 40%,
-            rgba(246, 246, 247, 0.8) 100%
+            rgba($tab-background-color, 0) 0%,
+            rgba($tab-background-color, 0.8) 40%,
+            rgba($tab-background-color, 0.8) 100%
         );
     }
 }
 
 .scroll-button {
-    --icon-background-color: #ffffff;
+    --icon-background-color: rgb(var(--contrast-100));
     top: pxToRem(4);
 
     &.left {
@@ -82,5 +82,5 @@
 .mdc-tab-scroller__scroll-content {
     padding: pxToRem(8) $tab-active-outer-edge-curve-size 0
         $tab-active-outer-edge-curve-size;
-    background-color: $tab-background-color;
+    background-color: rgb($tab-background-color);
 }
--- a/src/components/tab-bar/partial-styles/tab-bar-scroller.scss
+++ b/src/components/tab-bar/partial-styles/tab-bar-scroller.scss
@@ -1,0 +1,86 @@
+.scroll-fade,
+.scroll-button {
+    position: absolute;
+    transition: {
+        property: transform;
+        duration: 0.3s;
+        timing-function: ease-out;
+    }
+}
+
+.scroll-fade {
+    top: 0;
+    height: 100%;
+    width: pxToRem($tab-scroller-fade-width);
+    pointer-events: none;
+
+    &.left {
+        transform: translate3d(pxToRem(0 - $tab-scroller-fade-width), 0, 0);
+        left: 0;
+        background: linear-gradient(
+            270deg,
+            rgba(246, 246, 247, 0) 0%,
+            rgba(246, 246, 247, 0.8) 40%,
+            rgba(246, 246, 247, 0.8) 100%
+        );
+    }
+
+    &.right {
+        transform: translate3d(pxToRem($tab-scroller-fade-width), 0, 0);
+        right: 0;
+        background: linear-gradient(
+            90deg,
+            rgba(246, 246, 247, 0) 0%,
+            rgba(246, 246, 247, 0.8) 40%,
+            rgba(246, 246, 247, 0.8) 100%
+        );
+    }
+}
+
+.scroll-button {
+    --icon-background-color: #ffffff;
+    top: pxToRem(4);
+
+    &.left {
+        transform: translate3d(pxToRem(0 - $tab-scroller-fade-width), 0, 0);
+        left: pxToRem(4);
+    }
+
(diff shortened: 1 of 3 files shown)


src/components/tab-bar/partial-styles/tab-bar-scroller.scss
```diff
@@ -0,0 +1,86 @@
...
+
+    &.scroll-left:not(.scroll-right) .scroll-button.right,
+    &.scroll-right:not(.scroll-left) .scroll-button.left {
+        opacity: 0.5;
+    }
+}
+
+.mdc-tab-scroller__scroll-area--scroll {
+    scrollbar-width: none; // This hides the scrollbars appearing under the tab bar in Firefox
+    -ms-overflow-style: none; // Same as above for IE 11
```
jensgustafsson, line 79: have you actually tried this in IE? 🙈 😁

jensgustafsson (approved):
It seems to work as intended! I'm approving!

But it's hard for me to judge the css because there's so many fancy features I never used :P

src/components/tab-bar/partial-styles/tab-bar-scroller.scss
Kiarokh: I think I did actually on browser stack :)

adrianschmidt: We don't support IE, so there's no reason to add this. Also, our build-process should add any necessary prefixes automatically. I'm not sure that works if the properties aren't the same, but I think we should default to assuming manual adding of vendor-prefixed properties is not necessary, and only add them manually after confirming that it is through manual testing in the browser in question.

adrianschmidt (changes requested):
Imma override Jens's approval until I've had a look at this 😆

Kiarokh:
> It seems to work as intended! I'm approving!
> 
> But it's hard for me to judge the css because there's so many fancy features I never used :P

I've slightly minified the CSS, since it was too specific for an isolated web component. 

For the rest of it and how it works, you're not alone! I got some short circuits in my brain a few times, before I fully understood what to do and how it works 😄  It was like a mean IQ test (designed for CSS folk) I'm not saying your IQ is low @jensgustafsson , I just mean @adrianschmidt is smart

Kiarokh:
> Imma override Jens's approval until I've had a look at this 😆

I feel confident that this will be approved (if I don't have typos somewhere) 😄

adrianschmidt:
> > Imma override Jens's approval until I've had a look at this 😆
> 
> I feel confident that this will be approved (if I don't have typos somewhere) 😄

Yeah, so do I. Based on the description, this is something I want 👍 😄

I just want to take a look at the implementation, and that thing about not being able to change the styles of limel-icon-button from the outside made me interested 😉

Kiarokh:
> not being able to change the styles of limel-icon-button from the outside

Ah that'd be a good addition though. I tried making the button act/look as if it's a disabled button; or at least get a `cursor: not-allowed` style. But no success. I feel it's not a big deal. But it'd be nice if we could make it disabled for real, with some script.

src/components/tab-bar/partial-styles/tab-bar-scroller.scss
```diff
@@ -0,0 +1,86 @@
+.scroll-fade,
+.scroll-button {
+    position: absolute;
+    transition: {
+        property: transform;
+        duration: 0.3s;
+        delay: 0.5s;
```
adrianschmidt, line 7: I like the delay for hiding the button, but I don't like the delay for showing the button. It makes it so that instead of something happening as a direct response to something I did, it instead appears to be acting "on its own".

Kiarokh: I don't have the same experience. I think the delay is so short, so I still experience it as a reaction to my actions, but with a slightly less rush than before. In my head, it's better to first see that the bar scrolled, then to see that the button appeared (as a result), than seeing everything happening in the same time.

adrianschmidt: Really? I actually get slightly nauseous from it 😞

Kiarokh: Generally I don't think it's a good idea to animate many things in the UI simultaneously. Because users may miss what happened when too many things move around. (Unless they are gamers 😄)

It usually feels more peaceful and organic when there is slight delays and easing.
I hope this's not my personal opinion.


_However, what you say is possible to do in the CSS if the effect is too annoying._

adrianschmidt: I think _any_ animation that isn't a _direct_ and _immediate_ response to my interaction is extremely attention-grabbing, and makes the application an absolute nightmare to work in. To me, such animations tell me "drop everything you're doing, forget your train of thought, and look at me _NOW_".

I'm already really annoyed with the animated open/close button on collapsible-sections.

If you really want to go that way, we need to implement support for `prefers-reduced-motion`.

src/components/tab-bar/partial-styles/tab-bar-scroller.scss
Kiarokh: if you want me to change it or test it again, please make a new issue for it 👊😈
😁

src/components/tab-bar/partial-styles/tab-bar-scroller.scss
Kiarokh: We can certainly work with the `prefers-reduced-motion`. But I don't know anything about how it should be used. In other words, how much should we _reduce_ animations? What I've seen on the web is that they simply remove animations or reduce fram rate. For me that's totally bizarre. Everything in our surrounding world is animated, and they have delays and easing and so forth. UI animations are suposed to mimic the real world physics and that is why they are used to add an additional layer of meaning to the interface. I have a hard time understanding how completely removing some animation helps.

Anyhow, this is something I have to read more about. We can start by looking at iOS's **Reduce Motion** and **Prefer cross-fade transitions**.

But for this particular issue. It's not a big deal for me to remove the delay for when things appear. So let me know if you want a fixup for that. ;)

adrianschmidt: The reason for `prefers-reduced-motion` is that some people have disorders that make animations problematic. The [MDN page](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion) mentions "vestibular motion disorders", which, after a quick search, seems to include what we'd colloquially would call "motion sickness" or "sea sickness".

For me, the problem is probably related to my ADHD, which makes it extremely difficult to filter out things that move or blink in my visual field. This means that it's difficult, and very tiring, for me to for example read a text when there are animations on screen. If I'm already tired or stressed, focusing on anything while such an animation is present becomes completely impossible.

None of our animations are ongoing, thankfully, but all animations that happen somewhere other than where my focus is grabs my attention. Every. Single. Time.

To me, animations like the ones on the open/close button for collapsible-section are annoying and tiring the same way I'm guessing it would be for you if we flashed a large popover with some annoying ad in it every time you hovered over any part of a collapsible-section header.

> But for this particular issue. It's not a big deal for me to remove the delay for when things appear. So let me know if you want a fixup for that. ;)

I would very much appreciate that 🙂

Kiarokh: Awesome. Thanks for the description Adrian. I'll do a fixup soon then.

It's so fascinating how different we all are and it's so good to have you onboard too, to get things more accessible for the type of challenges you and people like you face.

I sometimes feel the same about my dyslexia. It helps simplify things and reduce amount of attention grabbing stuff on the screen.

Kiarokh:
pushed fixup

src/components/tab-bar/partial-styles/tab-bar-scroller.scss
adrianschmidt: Just tried it with the fixup, and I think it's great! Thank you!

adrianschmidt: Resolved, but will leave as unresolved, because I don't want to hide the good discussion 😊

src/components/tab-bar/partial-styles/tab-bar-scroller.scss
adrianschmidt: This was added in a previous PR, and just moved from one file to another in this one, so let's just leave it in there for now. I'm sure Kia will get annoyed with it and remove it in some future PR 😆

adrianschmidt:
@Kiarokh: On touch-screens, if the user taps the buttons instead of just swiping on the scrollbar, the invisible mouse-pointer is placed on the tapped button, and since the buttons won't disappear while they are being hovered, the button won't disappear until the user does something else that moves the invisible mouse-pointer again. I'll still approve and merge this, but we might want to take a look at whether we can tweak that later on.

adrianschmidt (approved):

PR #1184, opened 2021-03-11
feat(package): update @lundalogik/lime-icons8 to V2.4.0

Befkadu1:
fix Lundalogik/crm-feature#1889

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
   "devDependencies": {
     "@commitlint/config-conventional": "^12.0.1",
     "@lundalogik/cz-conventional-changelog": "^3.1.0",
-    "@lundalogik/lime-icons8": "^2.3.0",
+    "@lundalogik/lime-icons8": "^2.4.0",
     "@semantic-release/changelog": "5.0.1",
     "@semantic-release/exec": "5.0.0",
     "@semantic-release/git": "9.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -6,7 +6,7 @@
   "packages": {
     "": {
       "name": "@limetech/lime-elements",
-      "version": "30.18.2",
+      "version": "30.18.5",
       "license": "Apache-2.0",
       "dependencies": {
         "@limetech/material-components-web": "^4.0.1-p4.0.0.1",
@@ -27,7 +27,7 @@
       "devDependencies": {
         "@commitlint/config-conventional": "^12.0.1",
         "@lundalogik/cz-conventional-changelog": "^3.1.0",
-        "@lundalogik/lime-icons8": "^2.3.0",
+        "@lundalogik/lime-icons8": "^2.4.0",
         "@semantic-release/changelog": "5.0.1",
         "@semantic-release/exec": "5.0.0",
         "@semantic-release/git": "9.0.0",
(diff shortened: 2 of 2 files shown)


adrianschmidt (changes requested):
This isn't a feature, because it doesn't change the code that is published, only which icons are available in the documentation. It should be a `chore`, or possibly a `docs`, as it shouldn't trigger a new release.

Kiarokh:
on github's [preview build version](https://lundalogik.github.io/lime-elements/versions/PR-1184/#/component/limel-icon/) it's like this: 
image
I'm gonna test it locally now

Befkadu1:
> This isn't a feature, because it doesn't change the code that is published, only which icons are available in the documentation. It should be a `chore`, or possibly a `docs`, as it shouldn't trigger a new release.

Can `docs` trigger a new release.

Kiarokh:
hm... isn't it a new possibility / feature that can be used by consumers?
Or is it because of how we redistribute icons8 and such?

Kiarokh:
image
works fine in my local version

hannahu (approved):

hannahu (approved):

package.json
```diff
@@ -59,7 +59,7 @@
   "devDependencies": {
```
adrianschmidt, line 59: Commenting here so we can thread the discussion 🙂

> > This isn't a feature, because it doesn't change the code that is published, only which icons are available in the documentation. It should be a `chore`, or possibly a `docs`, as it shouldn't trigger a new release.
> 
> Can `docs` trigger a new release?

No, and it shouldn't.

> hm... isn't it a new possibility / feature that can be used by consumers?
> Or is it because of how we redistribute icons8 and such?

Exactly. We don't include the icons in the published package. The icons must be included separately, because we don't have the license to distribute the icons for free to other people.

Kiarokh: Hmmm... understood. tnx 4 the info 
I wonder whether we should have our own LimeIcons package parallell to this one? For icons which we make. Not sure what it requires. But before we make all these new filters icons available to everyone and people start using them, we should decide for that. 

We also discovered this issue yesterday: https://github.com/Lundalogik/crm-feature/issues/1915

Kiarokh: By the way, (the reason we wanted a new release was because we wanted to get the latest icons in CRM-components). But perhaps that works anyways, even if the version number doesn't increase? 🤔

adrianschmidt: > By the way, (the reason we wanted a new release was because we wanted to get the latest icons in CRM-components). But perhaps that works anyways, even if the version number doesn't increase? 🤔

Yeah, which icons are available (should) depend solely on which version of `lime-icons8` is used by the consumer, on this case `lime-webclient`. If you update the dependency there, you should be able to use the new icons.

> I wonder whether we should have our own LimeIcons package parallell to this one? For icons which we make. Not sure what it requires. But before we make all these new filters icons available to everyone and people start using them, we should decide for that.

I've considered that too… if we would want to make them publicly available, we would need to, yes. As long as we keep them private, it's easier to just keep it all in the same package I think.

> We also discovered this issue yesterday: [Lundalogik/crm-feature#1915](https://github.com/Lundalogik/crm-feature/issues/1915)

Yeah, I saw that… that's weird… Hopefully that's just something gone wrong when the docs were published. We'll be merging a PR that will create a new release soon, so I'll check then.

PR #1185, opened 2021-03-11
fix(input-field): fix don't render completions if completion list is …

maltja:
…empty

If the completion list is rendered and opened, but empty. The first click outside the list will
stillclose the list. This means that the user, for example, needs to click on a button two times to
click on the button. The completion list should therefore not be rendered if it is not necessary.

fix Lundalogik/crm-feature#1910

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -664,7 +664,7 @@ export class InputField {
     }
 
     private renderAutocompleteList() {
-        if (this.type === 'textarea') {
+        if (this.type === 'textarea' || !this.completions.length) {
             return;
         }
 


adrianschmidt (approved):
Great!

I renamed the commit to have the title (the first line) explain the change in a way a user might understand, like I mentioned in the call this morning.

maltja:
> Great!
> 
> I renamed the commit to have the title (the first line) explain the change in a way a user might understand, like I mentioned in the call this morning.

Thanks, I will think about this when naming commits in the future 🙂

PR #1188, opened 2021-03-22
docs(icon-button): clean up individual examples and add "composite" example for trying combinations

adrianschmidt:
fix: Lundalogik/crm-feature#1929

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/icon-button/examples/icon-button-composite.tsx
```diff
@@ -0,0 +1,78 @@
...
+ * Composite
+ *
+ * A place to try different combinations of states.
+ */
+@Component({
+    tag: 'limel-example-icon-button-composite',
+    shadow: true,
+    styleUrl: 'icon-button-composite.scss',
+})
+export class IconButtonCompositeExample {
```
adrianschmidt, line 13: The code in this file is taken from the limel-button example. I'd like to take a look at extracting it to some kind of helper file or something, so it can be imported and reused with ease. Then I want to make the docs for all simple components follow the structure seen here, with the most basic possible use case first, and then one separate example for each additional property, and at the end, a "composite" example where all properties can be toggled back and forth, so the user can get exactly the state they wish to see.

adrianschmidt: Oh, and if you can think of a better word than "composite", I'm very much open to suggestions 😄

jensgustafsson (approved):
Looks good to me! 👍

PR #1189, opened 2021-03-22
fix: allow snackbar message to be changed

anderssonjohan:
Please check that the modified example makes sense. 🙏 
The added code could also be moved to a separate example.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/snackbar/snackbar.tsx
```diff
@@ -118,9 +120,7 @@ export class Snackbar {
                         class="mdc-snackbar__label"
                         role="status"
                         aria-live="polite"
-                    >
-                        {this.message}
```
anderssonjohan, line 122: Note: This needed to be removed to avoid having the message rendered TWICE on the first call to show(). Instead of _rendering_ the message, we use MDCSnackbar properties to _set_ it before show, just like we do with timeout.

adrianschmidt:
image

The checkboxes are for the reviewer 😉

PR #1191, opened 2021-03-23
fix: pass z-index to datepicker portal

anderssonjohan:
Copy of fix from #1137 but for limel-date-picker.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/date-picker/date-picker.scss
+++ b/src/components/date-picker/date-picker.scss
@@ -1,5 +1,10 @@
 @import '../../style/internal/variables';
 
+// Note! The `--dropdown-z-index` property is used from `date-picker.tsx`.
+/**
+ * @prop --dropdown-z-index: z-index of the dropdown menu.
+ */
+
 div.container {
     position: relative;
 }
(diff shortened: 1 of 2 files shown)


src/components/date-picker/date-picker.scss
```diff
@@ -1,5 +1,10 @@
 @import '../../style/internal/variables';
 
+// Note! The `--dropdown-z-index` property is used from `date-picker.tsx`.
+/**
+ * @prop --dropdown-z-index: z-index of the dropdown menu.
```
adrianschmidt, line 5: I added the prop to the docs ☝️

adrianschmidt (approved):
I changed the commit message to read the same as the commits for #1137, because I think those described the change slightly better for an external dev reading the changelog, and added "fix" keyword links to the same (old) issues, to make it easier to find all related commits.

PR #1192, opened 2021-03-23
feat(popover): add consumer-overridable css property `--popover-z-index`

anderssonjohan:
Complimentary PR to #1137, adding CSS variable for the popover z-index so it renders correctly with other elements.

Before: Using popover with button as the trigger inside a limel-dialog shows the dialog on top of the popover
before picture popover in dialog

After: Using popover with icon button as trigger inside a limel-dialog, where a limel-select is used as the popover content

example popover in dialog

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/popover/popover.tsx
+++ b/src/components/popover/popover.tsx
@@ -111,11 +111,18 @@ export class Popover {
 
     public render() {
         const cssProperties = this.getCssProperties();
+        const popoverZIndex = getComputedStyle(this.host).getPropertyValue(
+            '--popover-z-index'
+        );
 
         return (
             <div class="trigger-anchor">
                 <slot name="trigger"></slot>
-                <limel-portal visible={this.open} containerId={this.portalId}>
+                <limel-portal
+                    visible={this.open}
+                    containerId={this.portalId}
+                    containerStyle={{ 'z-index': popoverZIndex }}
+                >
                     <limel-popover-surface
                         contentCollection={this.host.children}
                         style={cssProperties}
--- a/src/components/popover/popover.scss
+++ b/src/components/popover/popover.scss
@@ -2,6 +2,7 @@
  * @prop --popover-surface-width: Width of the popover surface. defaults to `auto`
  * @prop --popover-body-background-color: Background color of popover body, defaults to `--contrast-100`.
  * @prop --popover-border-radius: Border radius of popover, defaults to `0.75rem`.
+ * @prop --popover-z-index: z-index of the popover.
  */
 
 .trigger-anchor {
--- a/src/index.html
+++ b/src/index.html
@@ -11,6 +11,7 @@
     <style>
         :root {
             --dialog-z-index: 110;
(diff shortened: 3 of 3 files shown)


src/components/popover/examples/popover-in-dialog.tsx
```diff
@@ -0,0 +1,101 @@
...
+                            onClick={this.openPopover}
+                        />
+                        <limel-select
+                            value={this.value}
+                            options={this.options}
+                            onChange={this.onSelectChange}
+                            style={{ width: '30vw' }}
+                        />
+                    </limel-popover>
+                </limel-flex-container>
```
adrianschmidt, line 69: I think this example feels a bit too artificial to really be great, and it's also not working great, at least not on my machine (when opening the select, the page scrolls up to the top, but the menu stays where it was, making it "disappear" on my 13 inch laptop screen; I'm not sure if it's a problem with the example or with one of our components though). I prefer any example to be something we'd like people to copy and use in production code.

adrianschmidt (changes requested):
Since the portal component is still kept private, one could argue that the example matters less, and could be merged as-is, but I'd rather argue that no example is better than a "bad" example. I'm perfectly happy with the fix itself, so I'd be happy to merge just that, but leave the example out. Then we can consider whether it's worth putting in the effort to improve the example, while still getting the fix you need merged as soon as possible 🙂

src/components/popover/examples/popover-in-dialog.tsx
anderssonjohan: @adrianschmidt That sounds weird! Yeah, I agree that the example is a bit unrealistic and I don't think we really need an example to prove/show how the z-index works either. I'll remove that so we can merge this baby.

anderssonjohan:
@adrianschmidt the commit with the added example has now been removed

adrianschmidt (changes requested):
Nice work!

One tiny comment regarding adding the new css-variable to the docs, and I'd love it if you changed the commit message to the following, for consistency with the earlier commits that did similar things to other components:

```
feat(popover): add consumer-overridable css property `--popover-z-index`
```

src/components/popover/popover.tsx
```diff
@@ -111,11 +111,18 @@ export class Popover {
 
     public render() {
         const cssProperties = this.getCssProperties();
+        const popoverZIndex = getComputedStyle(this.host).getPropertyValue(
+            '--popover-z-index'
```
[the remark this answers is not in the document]
adrianschmidt: :zap:

adrianschmidt (approved):

Issue #1195, opened 2021-03-26
Make input components somewhat consistent regarding `disabled` and `readonly`


adrianschmidt:
## New feature motivation

Currently, some input components differentiates between `disabled` and `readonly`. For example `limel-picker`:

Enabled:
![image](https://user-images.githubusercontent.com/154725/112631889-c090e980-8e37-11eb-83da-4a91790dff47.png)

Readonly:
![image](https://user-images.githubusercontent.com/154725/112631911-c7b7f780-8e37-11eb-914d-88b93b148055.png)

Disabled:
![image](https://user-images.githubusercontent.com/154725/112631989-d9999a80-8e37-11eb-99f4-27e716c1f245.png)

In other words, when the component is `readonly`, the value cannot be changed, but can still be interacted with in other ways. When `disabled`, all interactions are disabled.

The problem is that we can't really use that in the Lime web client, because some other input components that have no need for the differentiation only have the `disabled` property.

That means that a consumer that wishes to prevent the user from changing the value of a component either has to check which component is being used, and know if that component supports `readonly`, or it has to just blanket use `disabled`, which is less than ideal.

## New feature description

It would be nice if all input components had both `readonly` and `disabled`. For components where the difference doesn't matter, setting either property to `true` should make the component readonly (or disabled, whichever you prefer to call it). That way, a consumer without intimate knowledge of the components can still prevent values to be changed, while still allowing "other" interactions, should the component happen to have any.

## New feature implementation

All components that only support either `readonly` or `disabled`, but not both, should support both. As a first step, the new property can just work exactly the same as the one already implemented. As a further step, any components that may benefit from differentiating between the two, but doesn't already, could be updated to behave differently between the two states.

Kiarokh:
This is an interesting subject and triggers some thoughts in my head. 

Things that are `disabled` are in other words potentially "interact-able". It usually means that if some conditions are (or would be) met, then they could become interactive again and let the users change their values or manipulate them. What I am trying to say is that the disabled state conveys this message that the element is disabled for a reason (and hopefully the user knows the reason too).

However, readonly is a totally different thing. An element that is readonly does not at all need to look like an "otherwise-interactable-element". It doesn't need to visualize any affordance for changeability or for receiving or accepting user input. Readonly could be even called Informative-only or illustrative-only. 

[...]

PR #1196, opened 2021-03-26
Add `readonly` property alongside `disabled` for all input components

adrianschmidt:
For most components, the properties do the same thing, and either can be used. In those cases, the addition of `readonly` is only to allow a more consistent interface between this and more complicated input components which do differentiate between the two.

For a few more complicated components, there is a difference between disabled and readonly, where readonly disallows changes to the value, but allows other interactions with the component, whereas disabled disallows all interaction.

This PR also includes a fix for a bug I found while making these changes to `limel-file`.

fix: #1195

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/checkbox/checkbox.tsx
+++ b/src/components/checkbox/checkbox.tsx
@@ -22,10 +22,18 @@ import { CheckboxTemplate } from './checkbox.template';
 })
 export class Checkbox {
     /**
-     * Disables the input field when `true`.
+     * Disables the checkbox when `true`. Works exactly the same as `readonly`.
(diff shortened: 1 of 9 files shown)


adrianschmidt:
This should be helpful in resolving Lundalogik/crm-feature#1854

jgroth (approved):

PR #1202, opened 2021-03-30
(re)enable keyboard navigation for button groups

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/1963

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/style/internal/z-index-variables.scss
+++ b/src/style/internal/z-index-variables.scss
@@ -9,3 +9,4 @@ $limel-slider--thumb-container--z-index: 1 !default;
 $limel-tab-bar--active-tab--z-index: 2 !default;
 $limel-table--has-interactive-rows--selectable-row--hover--z-index: 1 !default;
 $limel-popover-before--z-index: -1 !default;
+$limel-button-group-radio-button-keyboard-focused--z-index: 1 !default;
--- a/src/components/button-group/button-group.scss
+++ b/src/components/button-group/button-group.scss
@@ -37,20 +37,42 @@
         cursor: pointer;
     }
 
-    span[role='gridcell'] {
-        &:focus-within {
-            outline: none;
-            color: var(--mdc-theme-primary);
-        }
-    }
     input[type='radio'] {
         width: 0;
         position: absolute;
         opacity: 0; // This removes some visual defects on iOS and Firefox, created by "user agent styles"
+
+        &:focus-visible {
+            + label {
+                &:after {
+                    // visualizes keyboard navigation
+                    // only when non-pointer input is being used,
+                    // e.g. tabbed into using keyboard
+                    content: '';
+                    display: block;
+                    position: absolute;
+                    top: 0;
+                    right: 0;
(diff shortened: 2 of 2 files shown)


src/components/button-group/button-group.tsx
```diff
@@ -102,7 +99,8 @@ export class ButtonGroup {
                 <span role="gridcell">
                     <input
                         type="radio"
-                        name={this.radioGroupName}
+                        tabindex="0"
+                        name={button.id}
```
adrianschmidt, line 103: `name` is the _group_ that the radio-button belongs to. It should be the same for each radio-button in a group, because that's how the browser knows which radio-buttons belong together. Only one radio-button in each group can be selected at any one time.

The way radio-buttons are controlled with the keyboard is that tab focuses the currently selected button. If the user wishes to change the currently selected radio-button, they can do that using the arrow-keys. See docs here for more info: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio

If this is not the behaviour we want, we shouldn't be using radio-buttons. I think this _is_ the behaviour we want though, because it means that a user that doesn't wish to change the selected button can skip past the whole set with two tab-presses, regardless of the number of items in the set.

Kiarokh: Aha. Now I understand. Thanks for the clarification @adrianschmidt Then we can drop the first commit in this case. 

I can't answer whether we want this behaviour or not. I guess it's very hard to guess for keyboard users (but perhaps not for blind users). Not sure :/

src/style/internal/z-index-variables.scss
```diff
@@ -9,3 +9,4 @@ $limel-slider--thumb-container--z-index: 1 !default;
 $limel-tab-bar--active-tab--z-index: 2 !default;
 $limel-table--has-interactive-rows--selectable-row--hover--z-index: 1 !default;
 $limel-popover-before--z-index: -1 !default;
+$limel-button-group-radio-button-keyboard-focused: 1 !default;
```
adrianschmidt, line 12: When I created all these z-index variables, I added `--z-index` at the end of their names. I was thinking that it's nice to always know from the name of a variable that it's a z-index. But maybe that's unnecessary, since they'll probably be seen by the reader either in this file, or at the end of a line beginning with `z-index:`… 🤔

I guess if they're sent in to a mixin, it could be useful.

Anyway, I just wanted to check if leaving off the `--z-index` suffix was intentional or not? 😊

Kiarokh: Oh no I simply missed it. I have nothing agains having `--z-index` I'll do a fixup in a min

adrianschmidt (approved):

PR #1203, opened 2021-03-30
Input fields emit null instead of undefined

jensgustafsson:
fix: https://github.com/Lundalogik/crm-feature/issues/1960
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/widgets/input-field.ts
+++ b/src/components/form/widgets/input-field.ts
@@ -38,8 +38,9 @@ export class InputField extends React.Component {
     }
 
     private handleChange(event: CustomEvent<string>) {
-        const props = this.props;
         event.stopPropagation();
+        const props = this.props;
+
         if (!props.onChange) {
             return;
         }
--- a/src/components/form/widgets/input-field.ts
+++ b/src/components/form/widgets/input-field.ts
@@ -40,14 +40,19 @@ export class InputField extends React.Component {
     private handleChange(event: CustomEvent<string>) {
         event.stopPropagation();
         const props = this.props;
+        const type = getInputType(props.schema);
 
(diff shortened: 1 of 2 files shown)


jensgustafsson:
I like your suggestion @adrianschmidt! Let me know if you think I should squash it :)

adrianschmidt:
Cool! Yeah, you can squash it. I'm just going to test it in lime-webclient before approving 🙂

src/components/form/widgets/input-field.ts
```diff
@@ -44,10 +44,7 @@ export class InputField extends React.Component {
             return;
         }
 
-        let value;
-        if (event.detail !== '') {
-            value = event.detail;
-        }
+        const value = event.detail || null;
```
adrianschmidt, line 47: Hmm… using this, you can't empty any regular text-fields on the web client object card, or the save will fail, presumably because it doesn't expect null for string values… 😞

Errm… actually, testing a little further, _without_ this, the save won't fail, but it also won't save empty fields at all. It will just repopulate them with the old data…

I think what we need to do is output an empty string, unless the `type` is of a kind that doesn't return strings, in which case we should probably return `null` instead… 🤔

adrianschmidt:
> Nice! This looks like a better solution!

Great! I'll squash the fixup, and then go on to test datepicker 👍

adrianschmidt (approved):

PR #1205, opened 2021-04-01
fix(tab-bar): improve visualization of focus state

Kiarokh:
Focused tab does not require an extra click outside of it to lose
its styles which indicate that it is `focused`

fix: https://github.com/Lundalogik/crm-feature/issues/1969

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tab-bar/tab-bar.scss
+++ b/src/components/tab-bar/tab-bar.scss
@@ -35,6 +35,15 @@ $tab-scroller-fade-width: 65;
     }
 }
 
+.mdc-tab--active {
+    .mdc-ripple-upgraded--background-focused {
+        &:before {
+            background-color: transparent;
+            transition: background-color 1s ease;
+        }
+    }
+}
+
 .mdc-tab__ripple {
     box-sizing: border-box;
     border-radius: $tab-border-radius;
@@ -47,7 +56,7 @@ $tab-scroller-fade-width: 65;
 
     &:before,
     &:after {
-        transition: background-color 0.2s ease;
+        transition: background-color 0.5s ease;
     }
 }
 


src/components/tab-bar/tab-bar.scss
```diff
@@ -35,6 +35,15 @@ $tab-scroller-fade-width: 65;
     }
 }
 
+.mdc-tab--active {
+    .mdc-ripple-upgraded--background-focused {
+        &:before {
+            background-color: transparent;
+            transition: background-color 1s ease;
```
adrianschmidt, line 42: When the tab-bar is first loaded, hovering over the selected tab gives a hover effect. Once any tab has been clicked on, hovering over the selected tab no longer gives that hover effect.

https://user-images.githubusercontent.com/154725/114146639-22f9e780-9918-11eb-9517-cd06465f37d2.mov

That in itself isn't much of an issue I think, but when the currently selected tab receives focus by keyboard navigation, it highlights only to immediately fade out again. I'm not sure if that's going to be difficult to fix? I leave it to you to determine if we can merge this anyway, or if a fix is needed first.

adrianschmidt: https://user-images.githubusercontent.com/154725/114147492-0b6f2e80-9919-11eb-8a6f-4acbe8d3e101.mov

Kiarokh: Thanks. First one is fine I'd say. Displaying a hovering effect on an element which is already active and doesn't change even if you click on it is quite unnecessary (it's basically non-interactive in that state, so doesn't need a visual feedback on hover). But if it behaves in a certain way when the tab-bar is first loaded, hmm... is not even worth thinking about I'd say. 😄 it doesn't hurt to show that effect.

For the second point, I was fiddling around for a while to make a simple fix with minimal CSS. I think the effect is fine for keyboard navigation. Just to quickly "mention" that you just jumped to this tab (but it's already active).

In this video I'm playing with arrow keys back and forth. 


https://user-images.githubusercontent.com/35954987/114173580-ae36a580-9937-11eb-8f24-86474ad03e92.mp4

This effect is intentional and I personally like it. I don't think it requires any fix. Don't you like it?

adrianschmidt: Good enough for me 🙂

adrianschmidt (approved):

PR #1209, opened 2021-04-06
Add support for dark-mode to all components

Kiarokh:
fix: https://github.com/Lundalogik/hack-tuesday/issues/179

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/banner/banner.scss
+++ b/src/components/banner/banner.scss
@@ -1,13 +1,12 @@
-@import '@limetech/mdc-elevation/mixins';
 @import '../../style/internal/mdc-variables';
 @import '../../style/mixins';
 
 .lime-banner {
     min-height: $lime-banner-height;
-    @include mdc-elevation(1);
     display: none;
     background-color: rgba(var(--contrast-100), 0.7);
     backdrop-filter: blur(pxToRem(5));
+    box-shadow: var(--shadow-depth-16);
 
     &.lime-banner--open {
         display: block;
--- a/src/components/banner/banner.scss
+++ b/src/components/banner/banner.scss
@@ -1,12 +1,14 @@
 @import '@limetech/mdc-elevation/mixins';
-@import '@limetech/mdc-theme/mixins';
 @import '../../style/internal/mdc-variables';
+@import '../../style/mixins';
 
 .lime-banner {
     min-height: $lime-banner-height;
-    @include mdc-theme-prop(background-color, surface);
     @include mdc-elevation(1);
     display: none;
+    background-color: rgba(var(--contrast-100), 0.7);
+    backdrop-filter: blur(pxToRem(5));
+
     &.lime-banner--open {
         display: block;
         .lime-banner__surface {
@@ -28,6 +30,10 @@
         flex-wrap: wrap;
     }
 
+    div[class^='mdc-typography'] {
+        color: rgb(var(--contrast-1200));
+    }
+
     .lime-banner__icon {
         margin: 0 pxToRem(10);
         align-self: flex-start;
@@ -40,7 +46,10 @@
     }
 
     limel-icon {
-        color: white;
-        background-color: var(--icon-background-color, var(--lime-dark-grey));
+        color: rgb(var(--color-white));
+        background-color: var(
+            --icon-background-color,
+            rgb(var(--contrast-800))
+        );
     }
 }
--- a/src/components/banner/banner.scss
+++ b/src/components/banner/banner.scss
@@ -1,6 +1,12 @@
 @import '../../style/internal/mdc-variables';
 @import '../../style/mixins';
 
+/**
+ * @prop --banner-icon-fill-color: Color of the banner icon.
+ * @prop --banner-icon-background-color: Background color of the banner icon.
+ * @prop --icon-background-color: DEPRECATED. Use `--banner-icon-background-color` instead.
+ */
+
 .lime-banner {
     min-height: $lime-banner-height;
     display: none;
@@ -45,10 +51,10 @@
     }
 
     limel-icon {
-        color: rgb(var(--color-white));
+        color: var(--banner-icon-fill-color, rgb(var(--contrast-100)));
         background-color: var(
-            --icon-background-color,
-            rgb(var(--contrast-800))
+            --banner-icon-background-color,
+            var(--icon-background-color, rgb(var(--contrast-800)))
         );
     }
 }
--- a/src/components/banner/examples/banner.scss
+++ b/src/components/banner/examples/banner.scss
@@ -4,7 +4,7 @@ limel-banner {
     right: 0;
     top: 0;
     left: 0;
-    z-index: 10;
+    z-index: 112;
 
     @media (max-width: 720px) {
         position: absolute;
--- a/src/components/banner/examples/banner.scss
+++ b/src/components/banner/examples/banner.scss
@@ -1,5 +1,6 @@
 limel-banner {
-    --icon-background-color: rgb(var(--color-red-default));
+    --banner-icon-fill-color: rgb(var(--color-yellow-lighter));
+    --banner-icon-background-color: rgb(var(--color-red-default));
     position: fixed;
     right: 0;
     top: 0;
--- a/src/components/banner/examples/banner.scss
+++ b/src/components/banner/examples/banner.scss
@@ -1,5 +1,5 @@
 limel-banner {
-    --icon-background-color: var(--lime-deep-red);
+    --icon-background-color: rgb(var(--color-red-default));
     position: fixed;
     right: 0;
     top: 0;
--- a/src/components/collapsible-section/examples/collapsible-section.scss
+++ b/src/components/collapsible-section/examples/collapsible-section.scss
@@ -1,6 +1,7 @@
 :host(limel-example-collapsible-section-css-props) {
-    --closed-header-background-color: rgba(0, 0, 0, 0.05);
-    --open-header-background-color: rgba(0, 0, 0, 0.1);
-    --body-background-color: rgba(0, 0, 0, 0.05);
-    --body-padding: 0 1rem 1rem 1rem;
+    --closed-header-background-color: rgba(var(--contrast-1700), 0.05);
+    --open-header-background-color: rgba(var(--contrast-1700), 0.1);
+    --body-background-color: rgba(var(--contrast-1700), 0.05);
+    --body-padding: 1rem;
+    --header-stroke-color: rgb(var(--color-sky-default));
 }
--- a/src/components/grid/examples/grid.scss
+++ b/src/components/grid/examples/grid.scss
@@ -60,35 +60,35 @@ limel-grid {
 // when placing real components into a grid.
 
 my-deep-red-component {
-    background-color: var(--lime-deep-red);
+    background-color: rgb(var(--color-red-darker));
 }
 my-red-component {
-    background-color: var(--lime-red);
+    background-color: rgb(var(--color-red-default));
 }
 my-orange-component {
-    background-color: var(--lime-orange);
+    background-color: rgb(var(--color-orange-default));
 }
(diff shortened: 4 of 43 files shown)


src/components/banner/examples/banner.scss
```diff
@@ -1,10 +1,10 @@
 limel-banner {
-    --icon-background-color: var(--lime-deep-red);
+    --icon-background-color: rgb(var(--color-red-default));
     position: fixed;
     right: 0;
     top: 0;
     left: 0;
-    z-index: 10;
+    z-index: var(--dialog-z-index);
```
adrianschmidt, line 8: I don't like the use of `--dialog-z-index` here. It sets a bad example to use a variable that's clearly meant for use in a different component.

I think the banner should have the option to configure the z-index, similar to how that's possible for the dialog. I'll take a look at fixing that, it should be a simple thing to fix, and nice to add a clearly missing (albeit tiny) feature 😊

Kiarokh: Good point. 100% right!

src/style/internal/mdc-variables.scss
```diff
@@ -52,6 +52,14 @@ $mdc-theme-text-secondary-on-background: rgba(0, 0, 0, 0.54);
         --lime-text-disabled-on-background-color,
         rgba(var(--contrast-1700), 0.38)
     );
+    --mdc-theme-text-primary-on-background: var(
+        --lime-text-primary-on-background-color,
+        rgba(var(--contrast-1700), 0.87)
+    );
+    --mdc-theme-text-secondary-on-background: var(
```
adrianschmidt, line 65: Are these actually used by MDC, or are they variable names that we came up with? If they aren't defined by and used by MDC, they should not be prefixed with `--mdc-`.

If they are named the way they are because they are used by us to override the scss variable with the same name, then I think we could prefix these with `--override-mdc-` instead, so that `$mdc-theme-text-primary-on-background` becomes `--override-mdc-theme-text-primary-on-background` for example.

src/components/banner/examples/banner.scss
adrianschmidt: Eeeeh… wtf? _All_ the positioning of the element is actually just done in the example css!?

This is what the "banner" looks like when I remove all the styling from the example, and only the styling that is _actually_ delivered with the component is used:

https://user-images.githubusercontent.com/154725/114392051-8d20bf80-9b98-11eb-9ee6-613dc855f7a6.mov

## 🤦‍♂️

So… I'm not going to do anything major about that… I'll just change the hardcoded value in the example instead. Then we might want to consider reworking the banner component later.

src/components/list/examples/list-badge-icons.tsx
```diff
@@ -22,28 +22,28 @@ export class BadgeIconsListExample {
             secondaryText: '2-4 players',
             value: 2,
             icon: 'alien',
-            iconColor: 'var(--lime-green)',
+            iconColor: 'rgb(var(--color-lime-light))',
```
adrianschmidt, line 25: You had missed the ones in this file… maybe you only searched for colours in `.scss` files?

src/components/collapsible-section/examples/collapsible-section.scss
```diff
@@ -1,6 +1,7 @@
 :host(limel-example-collapsible-section-css-props) {
-    --closed-header-background-color: rgba(0, 0, 0, 0.05);
-    --open-header-background-color: rgba(0, 0, 0, 0.1);
-    --body-background-color: rgba(0, 0, 0, 0.05);
-    --body-padding: 0 1rem 1rem 1rem;
+    --closed-header-background-color: rgba(var(--contrast-1700), 0.05);
+    --open-header-background-color: rgba(var(--contrast-1700), 0.1);
+    --body-background-color: rgba(var(--contrast-1700), 0.05);
+    --body-padding: 1rem;
+    --header-stroke-color: rgb(var(--color-sky-default));
```
adrianschmidt, line 6: image

NICE! I like the blue 👍 🙂

src/style/internal/mdc-variables.scss
adrianschmidt: (We don't have to fix that in _this_ PR, but we should fix it. (Assuming the variables aren't "from" MDC of course.))

src/components/list/examples/list-badge-icons.tsx
Kiarokh: Hm.... I'm pretty sure I changed it :/ But probably missed it.

I never searched for any specific keyword. Went through everything line by line :)

src/components/banner/examples/banner.scss
Kiarokh: Yes, I wanted to tell you about that too. This is super weird and unpractical. But since the focus was on the colors I just updated what we had. Then I forgot about it. We need [a separate issue for it](https://github.com/Lundalogik/crm-feature/issues/1999).

src/style/internal/mdc-variables.scss
Kiarokh: Right now not 100% sure, but I think they're what you could see in the CSS code applied by MDC. That's why I reused them, and make a `--lime-` version of each. So we can some day talk about how consumers can "theme" their stuff if necessary, using these variables. Today they can only specify a few things like `--lime-primary-color`. We can talk about it later.

src/components/icon/icon.scss
```diff
@@ -1,11 +1,11 @@
 @import '../../style/internal/mdc-variables';
 
 /**
- * @prop --icon-background-color: Background color when attribute `badge` is set to `true`.
+ * @prop --icon-background-color: Background color when attribute `badge` is set to `true`; defaults to `transparent`
```
adrianschmidt, line 4:
```suggestion
 * @prop --icon-background-color: Background color when attribute `badge` is set to `true`. Defaults to `transparent`.
```

src/components/banner/banner.scss
```diff
@@ -1,12 +1,18 @@
-@import '@limetech/mdc-elevation/mixins';
-@import '@limetech/mdc-theme/mixins';
 @import '../../style/internal/mdc-variables';
+@import '../../style/mixins';
+
+/**
+ * @prop --banner-icon-fill-color: Color of the banner icon
+ * @prop --banner-icon-background-color: Background color of the banner icon
```
adrianschmidt, line 6:
```suggestion
 * @prop --banner-icon-fill-color: Color of the banner icon.
 * @prop --banner-icon-background-color: Background color of the banner icon.
 * @prop --icon-background-color: DEPRECATED. Use `--banner-icon-background-color` instead.
```

Fixup for `feat(banner): make icon fill color customizable & reveal color variables in the docs`.

While `--icon-background-color` was never properly documented, it was visible in the example, which means people might have used it. Better to make it clear that it has been deprecated and what to use instead.

src/components/icon/icon.scss
Kiarokh: me like

src/components/banner/banner.scss
Kiarokh: I have no specific opinion ;) Looks good.

src/components/grid/examples/grid.scss
```diff
@@ -60,35 +60,35 @@ limel-grid {
...
-    background-color: var(--lime-magenta);
+    background-color: rgb(var(--color-magenta-default));
 }
 my-light-grey-component {
-    background-color: var(--lime-light-grey);
+    background-color: rgb(var(--color-glaucous-default));
 }
 my-dark-grey-component {
-    background-color: var(--lime-dark-grey);
+    background-color: rgb(var(--color-gray-default));
```
adrianschmidt, line 93: Except for the two blue components, the new colors are easier to distinguish, which is better.

image

But, some of the colors no longer match their component names. In this example, it's fairly important that they do, for the reader to be able to follow the code and see how the code produces the end result.

image

Note that both the component name and the short identifier ('red', 'blu', 'mag', etc) needs to match. We could make the identifiers a little longer if we want to, but they should preferably all have the same number of characters, so they line up nicely in a grid matching the finished rendered grid…

(Either of us can fix it, but I'm leaving this as a note regardless to remember it should be done.)

Kiarokh: Good point. We let each other know if either of us starts fixing this 🌟

adrianschmidt: Yeah. I'm done with reviewing the code. Right now I'm going through all the examples, new and old side-by-side, for a visual inspection.

Kiarokh: coolz!

src/components/list/list.scss
```diff
@@ -93,9 +97,9 @@ $list-border-radius: 0.375rem; // 6px
         limel-icon {
             background-color: var(
                 --icon-background-color,
-                var(--lime-light-grey)
+                rgb(var(--contrast-700))
```
adrianschmidt, line 100: I assume this is what made these icons a little lighter when not specifying any colours…

image

I think they are a tad too light, making them look like they are disabled… when compared to the actually disabled item, it's pretty clear they're not, but still… and it's obviously worse if there are no disabled items, which is probably most often the case.

src/components/snackbar/snackbar.scss
```diff
@@ -10,7 +10,7 @@
 
 .mdc-snackbar__actions {
     .mdc-snackbar__dismiss {
-        color: $mdc-theme-on-secondary;
+        color: rgb(var(--mdc-theme-on-secondary));
```
adrianschmidt, line 13: This colour didn't turn out well.

image

adrianschmidt (changes requested):
I've reviewed the code, and done a visual inspection of all the pages of the docs (except those for _Types_) in light-mode. Everything I've found I've already commented about above this.

src/components/list/list.scss
Kiarokh: 💯 agree

src/components/snackbar/snackbar.scss
Kiarokh: Oh, good catch. I'll check.

src/style/internal/mdc-variables.scss
Kiarokh: if you look at this page: https://github.com/material-components/material-components-web/blob/master/docs/theming.md

you'll see a bunch of such variables. For example:
image

I think I saw them used somewhere in the original styles which are coming from MD. However, I don't remember which component it was. But that's why I decided to have them here too to ease switching to the dark mode.

But for me it's cool whichever we choose.

adrianschmidt: Ah! Ok, then it's a different matter completely 👍

src/components/grid/examples/grid.scss
Kiarokh: The reason colors messed up was that I picked whatever color name was there (which were lime brand color names) and used its equivalent variable name instead. But a bit inconsistently.

Anyhow, I pushed a fixup now, and they look like this 
image

src/components/list/list.scss
Kiarokh: last fixup looks like this
image

src/components/snackbar/snackbar.scss
Kiarokh: the reason was that RGB there 👉 `color: rgb(var(--mdc-theme-on-secondary));` 😄 
it should have been `color: var(--mdc-theme-on-secondary);` like the last PR

Kiarokh:
Great job! OK, I'm also done with the fixups now :)

adrianschmidt (approved):

Issue #1201, opened 2021-03-30
limel-menu: deprecate property `fixed`


adrianschmidt:
I think the fixed position version of the menu was created back before we used popper, as a "solution" to some problem we no longer have. I think we should deprecate it (on limel-menu that is). Thoughts? @jensgustafsson @jgroth

_Originally posted by @adrianschmidt in https://github.com/Lundalogik/lime-crm-components/pull/368#discussion_r601152176_

PR #1212, opened 2021-04-09
limel-menu: deprecate some old stuff

adrianschmidt:
#### feat(menu): deprecate use of limel-menu without a consumer provided trigger element

The default button was unstyled and we wouldn't want anyone to use it. Additionally, supplying a
custom trigger element is trivial, so there's no reason not to expect the consumer to do so.
Further, the default element seems to have been rendered every time the component re-rendered, even
though it was never seen by the user. Ergo: kill it with fire!

#### feat(menu): deprecate property `fixed`

Fixed position was used to get around a bug in the placement of the menu. This bug has since been
fixed, which makes this attribute obsolete.

fix: #1201

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu/menu.tsx
+++ b/src/components/menu/menu.tsx
@@ -61,6 +61,10 @@ export class Menu {
 
     /**
      * Defines whether the menu should have a fixed position on the screen.
+     *
+     * @deprecated Fixed position was used to get around a bug in the placement
+     * of the menu. This bug has since been fixed, which makes this attribute
+     * obsolete.
      */
     @Prop()
     public fixed = false;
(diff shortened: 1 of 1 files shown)


src/components/menu/menu.tsx
```diff
@@ -83,6 +90,12 @@ export class Menu {
         this.portalId = createRandomString();
     }
 
+    public componentDidLoad() {
+        if (!this.host.querySelector('[slot="trigger"]')) {
+            console.warn('Using limel-menu with the default trigger is deprecated. Please provide your own trigger element.');
```
adrianschmidt, line 95: Improved examples, not using the default trigger element, are coming in another branch.

jensgustafsson (approved):

PR #1213, opened 2021-04-09
docs(menu): add separate examples for each property

adrianschmidt:
Also add `limel-example-controls` component for wrapping controls for "composite" examples.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/examples/example-controls.tsx
```diff
@@ -0,0 +1,15 @@
+import { Component, h } from '@stencil/core';
+
+@Component({
+    tag: 'limel-example-controls',
+    styleUrl: 'example-controls.scss',
+})
+export class ExampleControls {
```
jgroth, line 7: I have had almost the exact same thought about a component like this. Nice to see that someone else saw the need for it! :smile: 

My idea was using `limel-form` though that took a simple schema for the controls. That would probably also remove a lot of event handlers and make setting the state of the example components a bit more simple :slightly_smiling_face:

jgroth: Using the form/schema approach would also make it possible to change stuff like labels dynamically without too much hassle :slightly_smiling_face:

jgroth (approved):
Looks great! Good job!
I will approve it but let you merge it in case you want to update the `ExampleControls` component. We can also  merge this now and add that during next cooldown if we think it's a good idea :slightly_smiling_face:

src/examples/example-controls.tsx
jgroth: I did some hacking in Kompendium over the weekend since I thought it would be quite easy to generate a schema for the components from there since all component metadata is available. The hacking resulted in a pre-release (since I did not have time to test it properly) where Kompendium will add an optional `schema` prop with the schema for the "owner" component to each example component :partying_face:

adrianschmidt: > My idea was using `limel-form` though that took a simple schema for the controls.

Oh… yeah… duh! 🤦‍♂️ 😂

adrianschmidt:
> Looks great! Good job!
> I will approve it but let you merge it in case you want to update the `ExampleControls` component. We can also merge this now and add that during next cooldown if we think it's a good idea 🙂

Yeah, let's merge this and update with a better version of `ExampleControls` later 👍

PR #1219, opened 2021-04-15
feat(chip-set): add badge to filter chip-set

maltja:
fix Lundalogik/crm-feature#2002

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -25,6 +25,7 @@ const SELECTED_CHIP_CLASS = 'mdc-chip--selected';
  * @exampleComponent limel-example-chip-set
  * @exampleComponent limel-example-chip-set-choice
  * @exampleComponent limel-example-chip-set-filter
+ * @exampleComponent limel-example-chip-set-filter-badge
  * @exampleComponent limel-example-chip-set-input
  * @exampleComponent limel-example-chip-icon-color
  */
@@ -577,6 +578,7 @@ export class ChipSet {
                         {chip.text}
                     </span>
                 </span>
+                {this.renderBadge(chip)}
             </div>
         );
     }
@@ -695,5 +697,13 @@ export class ChipSet {
         }
 
         return <div class="delimiter">{this.delimiter}</div>;
+    }
+
+    private renderBadge(chip: Chip) {
+        if (!chip.badge) {
+            return;
+        }
+
+        return <limel-badge label={chip.badge} />;
     }
 }
--- a/src/components/chip-set/chip.types.ts
+++ b/src/components/chip-set/chip.types.ts
@@ -44,4 +44,9 @@ export interface Chip<T = any> {
      * Value of the chip.
      */
     value?: T;
+
+    /**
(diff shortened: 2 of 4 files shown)


src/components/chip-set/chip.types.ts
```diff
@@ -44,4 +44,9 @@ export interface Chip<T = any> {
      * Value of the chip.
      */
     value?: T;
+
+    /**
+     * The value of the badge. Only valid for `filter`.
+     */
+    badgeValue?: number;
```
jgroth, line 51: If we rename this to just `badge`, we leave more room to add new features to this in the future. Right now we just allow a single value, but later we might want to add support for changing the color etc. e.g. by doing something like this

```ts
interface Chip {
    badge?: number | BadgeConfig;
}

const chip = {
    badge: {
        value: 4,
        backgroundColor: 'red'
    }
};
```

What do you think? :slightly_smiling_face:

src/components/chip-set/examples/chip-set-filter-badge.tsx
```diff
@@ -0,0 +1,101 @@
...
+                        onChange={this.setDisabled}
+                        checked={this.disabled}
+                    />
+                </limel-flex-container>
+            </p>,
+        ];
+    }
+
+    private chipSetOnChange(event: CustomEvent<Chip>) {
+        console.log(event.detail);
```
jgroth, line 69: Is this supposed to be here? :slightly_smiling_face:

src/components/chip-set/chip-set.tsx
```diff
@@ -577,6 +578,7 @@ export class ChipSet {
                         {chip.text}
                     </span>
                 </span>
+                {chip.badgeValue ? <limel-badge label={chip.badgeValue} /> : ''}
```
jgroth, line 581: I'm not a big fan of inline rendering conditions like this using the ternary operator since it can make the code harder to read. I suggest extracting this to a separate method instead, e.g. `renderBadge`

src/components/chip-set/examples/chip-set-filter-badge.tsx
maltja: This is just copied from the example with filter chip-set without the badge. I guess it was there to be able to test the method. 🤔

src/components/chip-set/chip.types.ts
maltja: ah, yes sounds good :)

src/components/chip-set/examples/chip-set-filter-badge.tsx
maltja: I will remove it, since we do something in the method anyway. 😛

jgroth (dismissed):

PR #1220, opened 2021-04-19
chore(list): move custom styles in a separate file

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/1212

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/list/examples/list-badge-icons-with-multiple-lines.tsx
+++ b/src/components/list/examples/list-badge-icons-with-multiple-lines.tsx
@@ -1,0 +1,65 @@
+import { ListItem } from '@limetech/lime-elements';
+import { Component, h } from '@stencil/core';
+
+/**
+ * Multi-line versus single-line layout
+ * By default, list items will always truncate the `text` line, which is displayed
+ * either alone, or as a primary heading (when there are both `text` and `secondaryText`)
+ * available. This means users will only see one line of text which is as wides as
+ * the list item, and no more. Thus, it is a good idea not to add long strings of
+ * text in the heading, as on mobile phones or small containers, they will be
+ * chopped off and truncated.
+ *
+ * However, the `secondaryText` which appears as a sub-heading is not truncated
+ * that quickly. By default, lists will display 3 lines of text, and then truncate
+ * the rest. Nevertheless, you can increase or decrease this number by specifying
+ * `maxLinesSecondaryText`.
+ * :::note
+ * Do not use `0`, negative numbers, decimal numbers, or very large numbers.
+ * :::
+ */
+@Component({
+    tag: 'limel-example-list-badge-icons-with-multiple-lines',
+    shadow: true,
+})
+export class BadgeIconsListExample {
+    private items: Array<ListItem<number>> = [
+        {
+            text:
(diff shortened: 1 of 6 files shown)


src/components/list/examples/list-badge-icons-with-multiple-lines.tsx
```diff
@@ -0,0 +1,65 @@
+import { ListItem } from '@limetech/lime-elements';
+import { Component, h } from '@stencil/core';
+
+/**
+ * Multi-line versus single-line layout
+ * By default, list items will always truncate the `text` line, which is displayed
+ * either alone, or as a primary heading (when there are both `text` and `secondaryText`)
+ * avialble. This means users will only see one line of text which is as wides as
```
maltja, line 8:
```suggestion
 * available. This means users will only see one line of text which is as wides as
```

src/components/list/list.tsx
```diff
@@ -122,10 +126,23 @@ export class List {
...
         const html = this.listRenderer.render(this.items, this.config);
 
         if (this.type !== 'menu') {
-            return html;
+            return (
+                <Host
+                    style={{
+                        '--maxLinesSecondaryText': `${maxLinesSecondaryText}`,
+                    }}
+                >
```
maltja, line 142: Could this style maybe be added in the listRenderer.render method instead? 🤔

Kiarokh: @Befkadu1

Befkadu1: > Could this style maybe be added in the listRenderer.render method instead? 🤔

Why? The objective is to style `limel-list` tag as you see in the pic below
If you can style the `limel-list` tag from listRenderer.render, you are welcome to do. But I wonder why we do it in listRenderer.render method

Screenshot 2021-04-20 at 18 11 16

maltja: I thought the objective was to override the style of the .mdc-list-item__secondary-text. Therefore, I thought the styling could be added to the ul tag instead, which is rendered in the listRenderer.render method. But I don't know what is the best practice when styling. 😛

Kiarokh: Please note that we are not really "styling" in that sense. We are applying a css variable (`--maxLinesSecondaryText`) to the parent (as a style which we have to), so that it becomes available for the children to use. So I think it's correct to style it at that level.

maltja: Okey then I approve it! 🙂

maltja (approved):

src/components/list/list.tsx
Kiarokh: Thanks Malin. So I'll squash the fixup now :)

Kiarokh: now somebody should rebase and merge :)

maltja (approved):

PR #1225, opened 2021-05-03
chore: add js version of color schema file

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/color-palette-extended.js
+++ b/color-palette-extended.js
@@ -1,0 +1,308 @@
+/* eslint-disable prettier/prettier */
+
+/*
+ * To be able to use colors with alpha, they are written in RGB.
+ * Comments with HEX codes in front of the values are used as reference only.
+ * Example of use:
+ * solid: `color: rgb(var(--contrast-100))`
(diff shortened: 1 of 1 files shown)


adrianschmidt:
@pontusn We discussed this one at standup, and we're not sure this is worth automating at the moment. The file added in this PR holds the "extended palette" we have currently, and I think I sent you a copy already, right? That's obviously going to slowly get out of sync as we update the values in lime-elements, but I don't think we'll actually be updating these values much at all, so I think the values you have will probably grow stale _very_ slowly. For the time being, I think it's better two 1) assume the values you have will work for quite some time, 2) update manually if and when the need arises, and 3) if we ever get to a point were we feel we're wasting too much time on updating manually, _then_ we automate.

Sounds ok?

Closing this for now.

PR #1236, opened 2021-05-12
fix(list): prevent flickering of scrollable menus

Kiarokh:
in Chrome on Windows, menus flicker when they
have a scroll bar and user hovers on them.

fix https://github.com/Lundalogik/crm-feature/issues/2090

## Review:

### Browsers tested:

Windows:
- [x] Chrome
- [x] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/list/list.scss
+++ b/src/components/list/list.scss
@@ -40,6 +40,8 @@ $list-border-radius: 0.375rem; // 6px
     border-radius: $list-border-radius;
 
     .mdc-list-item {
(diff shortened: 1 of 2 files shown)


adrianschmidt (approved):
I can't replicate the original problem, so I can't tell if the bug is fixed, but I have at least checked that this doesn't seem to break anything new 😅

PR #1237, opened 2021-05-12
fix(checkbox): better visualize keyboard navigation

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -58,3 +58,24 @@
         }
     }
 }
+
+.mdc-form-field {
+    .mdc-checkbox {
+        &.mdc-ripple-upgraded {
+            .mdc-checkbox__ripple {
+                &:before {
+                    animation: fade-out-focus-style 1s ease forwards;
+                }
+            }
+        }
+    }
+}
+
+@keyframes fade-out-focus-style {
+    0% {
+        opacity: 0.12; // What Material Design sets
+    }
+    100% {
+        opacity: 0; // What we like it to become, a moment after it gets focused
+    }
+}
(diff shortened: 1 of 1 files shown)


Kiarokh:
use `tab` key to test keyboard navigation

src/components/checkbox/checkbox.scss
```diff
@@ -43,6 +43,7 @@
 
     label {
         line-height: normal;
+        margin-left: -0.25rem;
```
adrianschmidt, line 46: There is already this `padding-left`:

image

I'm thinking maybe it's better to zero that out instead? 🤔

Kiarokh: Good catch. You're right!

Kiarokh: will you do or should I?

adrianschmidt: Did you check that your fixup works? Because I don't think it will. `.mdc-form-field label` has lower specificity than `.mdc-form-field > label`.

Kiarokh: yes I did :) it works

adrianschmidt: image

Indeed it does… but how the actual f… 🤨

I wonder if this is how physicists feel when they discover "new physics"… 😂

adrianschmidt (approved):

src/components/checkbox/checkbox.scss
Kiarokh: CSS weights man… I never understand them 😄

PR #1238, opened 2021-05-14
For input fields make readonly and disabled visually distinct

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/input-field/examples/input-field-text.tsx
```diff
@@ -39,7 +43,8 @@ export class InputFieldTextExample {
                 value={this.value}
                 required={this.required}
                 invalid={this.invalid}
-                disabled={this.disabled}
+                disabled={this.disabled || this.readonly}
```
adrianschmidt, line 46: The consumer shouldn't have to set both `disabled` and `readonly` to get the correct readonly styling.

Kiarokh: Well, I don't know how this exactly works. But probably what I did for the component itself is not enough then? 🤔 Because without `disabled={this.disabled || this.readonly}` this example doesn't make the input field both disabled and readonly in the same time. 👇

https://user-images.githubusercontent.com/35954987/118470101-aa502d00-b706-11eb-947f-4554b377f4a0.mp4


However, making other examples readonly by writing `readonly={true}` (and refreshing the page) will actually work as expected. They get both readonly and disabled in the same time (which I what we or at least I want.

adrianschmidt: What I mean is that all the correct styles for `readonly` should be applied by setting the `readonly` attribute. The consumer shouldn't have to set the `disabled` attribute _too_.

adrianschmidt: I can take a look at it 🙂

src/components/input-field/input-field.tsx
```diff
@@ -343,7 +345,8 @@ export class InputField {
         const classList = {
             'mdc-text-field': true,
             'mdc-text-field--invalid': this.isInvalid(),
-            'mdc-text-field--disabled': this.disabled || this.readonly,
+            'mdc-text-field--disabled': this.disabled,
+            'mdc-text-field--readonly': this.readonly,
```
adrianschmidt, line 349: As far as I can find, `mdc-text-field--readonly` doesn't exist in MDC, but is something we made up, right? If so, I don't think it should be called `mdc-text-field--readonly`. The `mdc-` prefix is used by MDC specifically to avoid conflicts with class names created by consumers (us). Apart from that consideration, using the `mdc-` prefix makes keeping the code clean more difficult. Imagine if we, somewhere down the line, remove all our own styling applied to `.mdc-text-field--readonly`. We'll probably keep applying the class name, under the assumption that it comes from MDC, and therefore MDC probably applies some styling for it.

Since we define this inside the shadow-DOM for limel-text-field, I don't think we need to specify that it's meant for use with and mdc-text-field, but if we absolutely want to indicate that, it should be done in a way that minimises the risk for confusion. Something like `lime-readonly-for-mdc-text-field` for example. If not, something like `readonly` is enough.

src/components/input-field/examples/input-field-text.tsx
adrianschmidt: I added two fixups to solve this.

src/components/input-field/examples/input-field-textarea.tsx
```diff
@@ -57,13 +67,15 @@ export class InputFieldTextareaExample {
...
+
+    private toggleReadonly() {
+        this.readonly = !this.readonly;
     }
 
-    private setRequired(event: CustomEvent<boolean>) {
-        event.stopPropagation();
-        this.required = event.detail;
+    private toggleRequired() {
+        this.required = !this.required;
```
adrianschmidt, line 79: I guess this was changed to make it all look the same when you copied `toggleReadonly` from the `input-field-text` example? I think the other pattern is better than the one this was changed to, so I'm going to change it back. I'll change the `input-field-text` example too though.

src/components/input-field/input-field.tsx
Kiarokh: > I don't think it should be called mdc-text-field--readonly.

Damn right. I forgot. We usually use `lime-` prefix for these.

src/components/input-field/examples/input-field-textarea.tsx
Kiarokh: yes I stole it from the other example 🙈

src/components/input-field/examples/input-field-text.tsx
Kiarokh: me trying to code 😁💩

so, all I wanted is that a `readonly` filed must be `disabled` as well, so that users don't tab into it or get hover effects and so forth. Also thought that this minimizes the amount of CSS I'd have to write.

adrianschmidt: I did a quick fix for it now, so that setting `readonly` also sets `disabled` "internally". But that isn't a great long-term solution, because that also disables the link-buttons for link-fields. However, for link-fields, I think the proper solution is to show the value as a link when the field is readonly, not just show a button after the value. I think we should do that separately though. It might be that we should do that in a separate component anyway, like Jens and Johan talked about at standup. I'm not sure…

src/components/input-field/examples/input-field-textarea.tsx
adrianschmidt: I've updated the examples a bunch now.

src/components/input-field/examples/input-field-text.tsx
Kiarokh: > that also disables the link-buttons for link-fields.
Hm.... I had not thought about that. Right. 

I am not sure if having two components and switching them on the go is easier. It's of course tempting to have static components that don't have a lot of logic with them and so forth. But I can't judge whether it's worth the efforts.

> link-fields the proper solution is to show the value as a link when the field is readonly, not just show a button after the value.

Yes, it feels in those cases, that entire value should be the clickable thing.

src/components/input-field/input-field.tsx
adrianschmidt: If you change this, and you're happy with my fixes, I think we can merge.

Kiarokh: i'll fixup

src/components/input-field/examples/input-field-text.tsx
Kiarokh: Check my last two commits. What do you think? It makes the trailing icons clickable, even if the field is readonly. They lose their custom colors (because I don't want to write a ton of extra CSS).

Kiarokh: I'm working a bit more on this @adrianschmidt. So we will soon get nice styles even for those link-buttons. Soon I'll push more fixups

Kiarokh: OK @adrianschmidt now I'm done. You can check the last 4 commits now, which together enable styling of link buttons for readonly fields ;)

Kiarokh: https://user-images.githubusercontent.com/35954987/118631661-05018b80-b7d0-11eb-816e-b044e5e875ae.mp4

adrianschmidt (approved):

PR #1239, opened 2021-05-17
fix(input-field): prevent labels getting cut off when input field is focused

Kiarokh:
fix: Lundalogik/crm-feature#1671

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.scss
+++ b/src/components/input-field/input-field.scss
@@ -10,6 +10,7 @@
 
 @import './partial-styles/trailing-icon.scss';
 @import './partial-styles/outlined-style-overrides.scss';
+@import './partial-styles/cropped-label-hack';
 
 /**
  * @prop --textarea-height: Height of the field when type is set to `textarea`
--- a/src/components/input-field/partial-styles/_cropped-label-hack.scss
+++ b/src/components/input-field/partial-styles/_cropped-label-hack.scss
@@ -1,0 +1,19 @@
+// Some font size applied to `label--float-above` causes the labels to get cut off
+// when an empty field gets focused
+
+.mdc-text-field--outlined.mdc-notched-outline--upgraded,
+.mdc-text-field--outlined .mdc-notched-outline--upgraded {
+    .mdc-floating-label--float-above {
+        //font-size: 1rem; This is what we get from MD now, which causes the miscalculations
+        font-size: 0.875rem;
+        // of course this is `14px` and the other one is `16px`.
+        // Unfortunately MD scales the floating label down, by applying a
+        // `transform` & `scale(0.75)` which is probably why they had to increase
+        // the font-size, to make it more readable.
+        // This is why I don't like this hack.
+    }
+}
+
+// NOTE: if we ever remove this hack, we must also update a mixin
+// called: `@mixin label--float-above()`
+// located in: style/internal/shared_input-select-picker.scss
--- a/src/style/internal/shared_input-select-picker.scss
+++ b/src/style/internal/shared_input-select-picker.scss
@@ -42,7 +42,8 @@ $height-of-helper-text-pseudo-before: 0.75rem; // There is strange a pseudo befo
 
 @mixin label--float-above() {
     transform: translateY(-2.38rem) scale(0.75);
-    font-size: pxToRem(16);
+    // font-size: pxToRem(16); // See comments in: /input-field/partial-styles/_cropped-label-hack.scss
+    font-size: 0.875rem;
 }
 
 @mixin looks-like-helper-line {


src/components/input-field/partial-styles/cropped-label-hack.scss
```diff
@@ -0,0 +1,19 @@
...
+.mdc-text-field--outlined.mdc-notched-outline--upgraded,
+.mdc-text-field--outlined .mdc-notched-outline--upgraded {
+    .mdc-floating-label--float-above {
+        //font-size: 1rem; This is what we get from MD now, which causes the miscalculations
+        font-size: 0.875rem;
+        // of course this is `14px` and the other one is `16px`.
+        // Unfortunately MD scales the floating label down, by applying a
+        // `transform` & `scale(0.75)` which is probably why they had to increase
+        // the font-size, to make it more readable.
+        // This is why I don't like this hack.
```
adrianschmidt, line 13: > which is probably why they had to increase the font-size

Who are "they"?

I think we increased the font-size, because I'm reasonably sure this bug doesn't appear in MDC's examples.

Kiarokh: I don't find anywhere in the code that we deliberately change the font size for floating labels of input fields. This must be Material Design's originally.

Screen Shot 2021-05-17 at 15 26 49 (2)

jensgustafsson: Could it not be due to our px to rem conversions?

Kiarokh: I don't know. Unless that conversion code is also rounding up numbers (which I don't think it does)?

Kiarokh: or is it JavaScript that is having mathematical problems (e.g. with decimals as Adrian was saying this morning?)

adrianschmidt: The pxToRem is done completely by Sass while compiling the scss to css. While I can't completely rule it out, I think it's unlikely to be caused by the conversion itself. I think it's more plausible that it could be caused by us using rem rather than px (that is, it wouldn't matter if we use the `pxToRem()` function or just defined values in `rem`, but it _would_ matter if we used `px` _instead_ of `rem`).

I still have this distinct memory that we somehow increased the font size for floating labels back in the day, but perhaps that was for Angular Material, not for MDC… 🤔

adrianschmidt: > or is it JavaScript that is having mathematical problems (e.g. with decimals as Adrian was saying this morning?) So while converting, it messes up this number?

I think that's very unlikely, because the error will only ever be fractional, so it would have to be multiplied by at least a factor of 10, but more likely at least 100, to even cause an error of a single pixel…

Kiarokh: I wasn't talking about our `pxToRem` function. I thought there is some script that automatically converts `px` values that are inside MD's original css files which we import and load. Because I don't find anything in our scss files.

adrianschmidt: Ah. Yes, there is a script that converts MDC's px-values to rem-values. There could be something going wrong there, true… 🤔

The script is here: https://github.com/Lundalogik/lime-material-scripts/blob/master/convert.js
Review is welcome! 😊

Kiarokh: @jensgustafsson, he's talking to you 😜

jensgustafsson: I guess we need to compare the lines of code after/before converting because that script doesn't tell me too much 😛🙈

jensgustafsson: But I'm fine with your change @Kiarokh. Not sure what @adrianschmidt thinks though

adrianschmidt: Yeah, I am too! 👍

src/components/input-field/input-field.scss
```diff
@@ -10,6 +10,7 @@
 
 @import './partial-styles/trailing-icon.scss';
 @import './partial-styles/outlined-style-overrides.scss';
+@import './partial-styles/cropped-label-hack.scss';
```
adrianschmidt, line 13: Should we use the underscore prefix since this is a partial?

Kiarokh: we should. iForgot
Will you please fix it up? Or you want me to do so?

adrianschmidt: I can do it 👍

adrianschmidt (approved):

PR #1244, opened 2021-05-18
package(feat): upgrade Kompendium from v0.7.3 to v0.8.7

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "jest": "^26.6.3",
     "jest-cli": "^26.6.3",
     "jsx-dom": "^6.4.21",
-    "kompendium": "^0.7.3",
+    "kompendium": "^0.8.7",
     "lodash-es": "^4.17.21",
     "moment": "^2.29.1",
     "number-abbreviate": "^2.0.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -45,7 +45,7 @@
         "jest": "^26.6.3",
         "jest-cli": "^26.6.3",
         "jsx-dom": "^6.4.21",
-        "kompendium": "^0.7.3",
+        "kompendium": "^0.8.7",
         "lodash-es": "^4.17.21",
         "moment": "^2.29.1",
         "number-abbreviate": "^2.0.0",
@@ -5282,6 +5282,15 @@
       "integrity": "sha1-GwqzvVU7Kg1jmdKcDj6gslIHgyc=",
       "dev": true
     },
+    "node_modules/fuse.js": {
+      "version": "6.4.6",
+      "resolved": "https://registry.npmjs.org/fuse.js/-/fuse.js-6.4.6.tgz",
+      "integrity": "sha512-/gYxR/0VpXmWSfZOIPS3rWwU8SHgsRTwWuXhyb2O6s7aRuVtHtxCkR33bNYu3wyLyNx/Wpv0vU7FZy8Vj53VNw==",
+      "dev": true,
+      "engines": {
+        "node": ">=10"
+      }
+    },
     "node_modules/gensync": {
       "version": "1.0.0-beta.2",
       "resolved": "https://registry.npmjs.org/gensync/-/gensync-1.0.0-beta.2.tgz",
@@ -7674,12 +7683,13 @@
       }
     },
     "node_modules/kompendium": {
-      "version": "0.7.3",
-      "resolved": "https://registry.npmjs.org/kompendium/-/kompendium-0.7.3.tgz",
-      "integrity": "sha512-G/Ouk8WutkcAla2OL66C0GZqBqTCJKAokOyjy3+r28C1Ap0qa7gAHsql0oiSce9riV/8Y9PpxO8asLF1DKxGQw==",
+      "version": "0.8.7",
+      "resolved": "https://registry.npmjs.org/kompendium/-/kompendium-0.8.7.tgz",
+      "integrity": "sha512-zfIj5PJq6MLNXBehHxKdZAN7eZIGsfd4WWLYr90ElJLNYgBEBZyL96N19W2dLm4lFTx3lpl4urOg44OeMFfuhQ==",
       "dev": true,
       "dependencies": {
         "chokidar": "^3.3.1",
+        "fuse.js": "^6.4.6",
         "glob": "^7.1.6",
         "lnk": "^1.1.0",
         "lodash": "^4.17.19",
@@ -7692,6 +7702,7 @@
         "remark-parse-yaml": "0.0.3",
         "remark-rehype": "^7.0.0",
         "typedoc": "^0.17.8",
+        "typescript": "^4.2.4",
         "unified": "^9.0.0",
         "unist-util-flatmap": "^1.0.0",
         "unist-util-map": "^2.0.1",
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -76,7 +76,7 @@
     "jest": "^26.6.3",
     "jest-cli": "^26.6.3",
     "jsx-dom": "^6.4.21",
-    "kompendium": "^0.7.3",
+    "kompendium": "^0.8.5",
```
adrianschmidt, line 79: `limel-collapsible-section` needs updating:

image
image

adrianschmidt, line 79: `limel-date-picker` looks ok for datetime, date, time, and week, but for month, quarter, and year, the options look disabled:

image

adrianschmidt (changes requested):
There are some issues when used in dark-mode.

I don't think I'll go through any more components right now. Should we disable dark-mode for Kompendium, or should we fix all these things before we update? 🙂

Kiarokh:
@adrianschmidt this is great! 

Let me tell you what the story is here.

When @jgroth and I made Kompendium, we just copy/pasted the same css variables from Lime Elements for colors and shadows. So they had the same name in both places.

But in Lime Elements, we turned off the support for dark mode, which means we didn't allow the CSS variables for colors to switch, when the browser is being in dark mode. This is why we never really got any darkmode in Lime Elements, for the documentation itself; but Kompendium Demo had it. 

However, the only css variable which was only used in Kompendium was the the box-shadow around the example showcases (which created that inflated look and feel). This one had a dark mode version as well, which was working properly, but resulted in a silly visual effect for people who were using dark mode. If you remember we talked about it.

In the latest release we renamed all those CSS variables in Kompendium, to differentiate between those and the ones from Lime Elements, and get the dark mode back.

Also, when you and I were trying to enable dark mode in Lime Elements, I noticed that many of the elements (`<span>`s, `<p>`s, etc.) just inherit their text color from Kompendium (from `body` or somewhere), and don't have any colors defined. I think that's the way it should be normally. 

Anyhow, this is the reason why text becomes very light grey in examples now. Because it's inheriting its colors, and Kompendium in dark mode has that color for text. The problem is that we intentionally make all example showcase boxes white in Kompendium, because Lime Elements doesn't have support for dark mode yet. Do you remember that we had to locally install this branch of Kompendium in Lime Elements: https://github.com/jgroth/kompendium/tree/dark-mode-for-examples to remove this white box examples from Kompendium, to be able to work with the dark mode during the hack Tuesday?


## Solution 1:
I can fix this in Kompendium and force the text color to stay the same for example showcase areas, even if the whole page is in dark mode. This solution will then fix problems like:

image

or

image

But it will NOT fix problems like the checkbox in this picture:
image

And that is because the checkbox is one of the elements that doesn't just inherit its text color. The color comes from `.mdc-form-field` specified in its CSS which is `color: var(--mdc-theme-text-primary-on-background, rgba(0, 0, 0, 0.87));`. Therefore, since in this case it's located outside the example showcase area with no white background, it will have a dark color even with that fix.

## Solution 2: I prefer that we do this solution as well!
That we add text colors to these components in Lime Elements instead. We can simply use `color: var(--mdc-theme-text-primary-on-background)` on elements that are missing text color. This way, they will be better prepared for dark mode as well, and won't get affected negatively like this.

******

Note: I should do solution one anyways, as long as we don't let example showcases in Kompendium to get dark in dark mode.

Kiarokh:
Also, we must give a background color to those examples which don use Kompendium's `show-case_component` to load (that white box).

Kiarokh:
PR for solution 1: https://github.com/jgroth/kompendium/pull/55 please approve and merge @jgroth (or @adrianschmidt)

fixes problems such as this one: 
image

Kiarokh:
another PR for Lime Elements: https://github.com/Lundalogik/lime-elements/pull/1245

fixes problems like:

### before:
image

### after:
image

jgroth:
> PR for solution 1: [jgroth/kompendium#55](https://github.com/jgroth/kompendium/pull/55) please approve and merge @jgroth
> 

`v0.8.6` speedboat :speedboat: released!

Kiarokh:
Thanks @jgroth 

OK, let's ignore this ☝️ PR and merge this one instead : https://github.com/Lundalogik/lime-elements/pull/1246

package.json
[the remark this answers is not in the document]
Kiarokh: yes, we can't fix this in any other way

adrianschmidt:
> Thanks @jgroth
> 
> OK, now that Johan has merged that, let's ignore this ☝️ PR and instead merge this one: 👉 #1246

No, let's ignore #1246 instead, since all conversiation is here already.

cc @jgroth

adrianschmidt:
> another PR for Lime Elements: #1245 @adrianschmidt please approve :)

I merged it into this PR. Just add fixes to this branch instead of opening new PRs please 🙂

Kiarokh:
sometimes I become so stupid 😄 

so let me know if anything else is left that you want me to do

adrianschmidt:
> sometimes I become so stupid 😄

Happens to the best of us 😊

image

package.json
```diff
@@ -76,7 +76,7 @@
     "jest": "^26.6.3",
```
adrianschmidt, line 76: The issue of having to place things on a light background is just temporary until we fully support dark-mode in lime-elements, right?

If so, then updating this last section of the _Contributing_ page probably isn't worth it, since we'll then probably just forget to change it back, right?

image

package.json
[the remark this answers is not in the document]
adrianschmidt: This checkbox still looks the same.

adrianschmidt: This still looks the same.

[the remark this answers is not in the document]
Kiarokh: [this commit](https://github.com/Lundalogik/lime-elements/commit/e04aecd9e66b6d1bd7982ad9e0351bc9ac27ead9) will fix it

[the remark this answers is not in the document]
Kiarokh: [this commit](fa295cfb0641f2eaaf35082d29688c9d3b3d2696) will fix it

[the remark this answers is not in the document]
Kiarokh: fixed [here](https://github.com/Lundalogik/lime-elements/commit/bff34961824b0697129bd69bfb5e0108b1c25673)

Kiarokh: seems to be fixed too

package.json
```diff
@@ -76,7 +76,7 @@
...
     "jest-cli": "^26.6.3",
     "jsx-dom": "^6.4.21",
-    "kompendium": "^0.7.3",
+    "kompendium": "^0.8.6",
```
adrianschmidt, line 79: I suggest this alternative PR instead: #1250

adrianschmidt: (Note that it requires jgroth/kompendium#57.)

package.json
adrianschmidt: This still looks the same as the screenshot above in my browser, but I don't think it has to stop us from merging.

adrianschmidt (approved):

PR #1249, opened 2021-05-20
build: update Stencil to 2.5.2

jgroth:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/global.d.ts
```diff
@@ -1,22 +0,0 @@
-// The ResizeObserver API is available in all major browsers. However,
-// TypeScript has yet to implement any type declarations for it. A basic
-// incomplete version of the API is declared here just to get the compiler
-// not to throw errors.
-//
-// https://github.com/microsoft/TypeScript/issues/37861
-
-interface IResizeObserver {
```
adrianschmidt, line 8: Unless this change is visible to consumers in any way, I think this is a `refactor` rather than `fix`.

Unless you specifically want the version bump, then I guess leave it 😄

adrianschmidt, line 8: What am I doing wrong here?

image

I've run `npm install`, and `npm list typescript` yields:

```
@limetech/lime-elements@31.9.0 /Users/adrian.schmidt/src/lime-elements
└── typescript@4.2.3
```

jgroth: I made it a `fix` because I got build errors I think :slightly_smiling_face:

jgroth: Hmm :thinking: 
It seems a bit odd... I get the same output
```
@limetech/lime-elements@31.8.0 /home/jth/workspace/lime-elements
└── typescript@4.2.3
```

Maybe another case of Mac vs Linux? :disappointed:

adrianschmidt: I've updated Stencil to 2.6.0 in #1274, and with your fix cherry-picked, that passes both on GitHub and locally on my machine, so I'll merge that one, and close this.

PR #1250, opened 2021-05-20
Enable darkmode support in Lime Elements

adrianschmidt:
~Works slightly better with https://github.com/jgroth/kompendium/pull/102~ Edit: PR branch updated to latest Kompendium!

## Test with the web client
1. **Web-client**, check out: https://github.com/Lundalogik/lime-webclient/pull/3169
2. **crm-components** check out https://github.com/Lundalogik/lime-crm-components/pull/1415 that has better support for dark mode
3. **crm-building-blocks** check out [link coming soon…] that has better support for dark mode

fix: https://github.com/Lundalogik/lime-elements/issues/1372
fix: https://github.com/Lundalogik/lime-elements/issues/1373
fix: https://github.com/Lundalogik/lime-elements/issues/1374

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -19,9 +19,6 @@
 }
 
 button {
-    @include mixins.is-elevated-clickable();
-    @include mixins.visualize-keyboard-focus;
-
     &.mdc-button {
         min-width: functions.pxToRem(36);
         padding: {
@@ -35,8 +32,10 @@ button {
         width: 100%;
 
         &:not(:disabled) {
-            color: var(--mdc-theme-primary, rgb(var(--color-teal-default)));
-            background-color: rgb(var(--contrast-100));
+            @include mixins.is-elevated-clickable(
+                $color: var(--mdc-theme-primary, rgb(var(--color-teal-default)))
+            );
+            @include mixins.visualize-keyboard-focus;
 
             &.mdc-button--raised,
             &.mdc-button--unelevated {
--- a/src/components/code-editor/code-editor.scss
+++ b/src/components/code-editor/code-editor.scss
@@ -9,134 +9,133 @@
  * @prop --code-editor-font-size: Defines the font size of the code, defaults to `0.875rem`.
  */
 
+@mixin light-mode-styles {
+    --code-editor-neutral-text-color: var(--contrast-1100);
+    --code-editor-background-color: var(--contrast-200);
+    --code-editor-gutter-background-color: var(--contrast-300);
+    --code-editor-line-number-text-color: var(--contrast-700);
+    --code-editor-cursor-color: var(--color-amber-dark);
+
+    --code-editor-selection-highlight-color: var(--color-cyan-default);
+    --code-editor-selection-lowlight-color: var(--contrast-600);
+
+    .cm-attribute,
+    .cm-property,
+    .cm-keyword {
+        color: rgb(var(--color-violet-light));
+    }
+    .cm-operator {
+        color: rgb(var(--color-sky-dark));
+    }
+    .cm-variable {
+        color: rgb(var(--color-pink-light));
+    }
+    .cm-variable-2 {
+        color: rgb(var(--contrast-1100));
+    }
+    .cm-variable-3,
+    .cm-type {
+        color: rgb(var(--color-amber-darker));
+    }
+    .cm-builtin,
+    .cm-meta {
+        color: rgb(var(--color-amber-default));
+    }
+    .cm-atom {
+        color: rgb(var(--color-coral-default));
+    }
+    .cm-tag,
+    .cm-number {
+        color: rgb(var(--color-magenta-default));
+    }
+    .cm-def {
+        color: rgb(var(--color-sky-default));
+    }
+    .cm-string {
+        color: rgb(var(--color-green-dark));
+    }
+    .cm-string-2 {
+        color: rgb(var(--color-pink-default));
+    }
+    .cm-qualifier {
+        color: rgb(var(--color-grass-default));
+    }
+    .cm-error {
+        color: rgb(var(--color-white));
+        background-color: rgb(var(--color-red-default));
+    }
+    .cm-comment {
+        color: rgba(var(--code-editor-neutral-text-color), 0.7);
+    }
+}
+
+@mixin dark-mode-styles {
+    --code-editor-neutral-text-color: var(--contrast-1600);
+    // --code-editor-background-color: { same as light };
+    // --code-editor-gutter-background-color: { same as light };
+    --code-editor-line-number-text-color: var(--contrast-800);
+    --code-editor-cursor-color: var(--color-amber-default);
+
+    --code-editor-selection-highlight-color: var(--color-cyan-light);
+    --code-editor-selection-lowlight-color: var(--contrast-100);
+
+    .cm-attribute,
+    .cm-property,
+    .cm-keyword {
+        color: rgb(var(--color-purple-lighter));
+    }
+    .cm-operator {
+        color: rgb(var(--color-sky-lighter));
+    }
+    // .cm-variable { same as light }
+    // .cm-variable-2 { same as light }
+    .cm-variable-3,
+    .cm-type {
+        color: rgb(var(--color-yellow-light));
+    }
+    // .cm-builtin,
+    // .cm-meta { same as light }
+    .cm-atom {
+        color: rgb(var(--color-coral-light));
+    }
+    .cm-tag,
+    .cm-number {
+        color: rgb(var(--color-magenta-light));
+    }
+    .cm-def {
+        color: rgb(var(--color-sky-light));
+    }
+    .cm-string {
+        color: rgb(var(--color-lime-light));
+    }
+    .cm-string-2 {
+        color: rgb(var(--color-pink-lighter));
+    }
+    .cm-qualifier {
+        color: rgb(var(--color-grass-light));
+    }
+    // .cm-error { same as light }
+    .cm-comment {
+        color: rgba(var(--code-editor-neutral-text-color), 0.4);
+    }
+}
+
 :host {
     display: flex;
     font-size: var(--code-editor-font-size, 0.875rem); // 14px
+    @include light-mode-styles;
+}
+
+@include mixins.in(dark-mode) {
+    :host {
+        @include dark-mode-styles;
+    }
 }
 
 .editor {
     display: flex;
     align-items: stretch;
     width: 100%;
-
-    &.is-light-mode {
-        --code-editor-neutral-text-color: var(--contrast-1100);
-        --code-editor-background-color: var(--contrast-200);
-        --code-editor-gutter-background-color: var(--contrast-300);
-        --code-editor-line-number-text-color: var(--contrast-700);
-        --code-editor-cursor-color: var(--color-amber-dark);
-
-        --code-editor-selection-highlight-color: var(--color-cyan-default);
-        --code-editor-selection-lowlight-color: var(--contrast-600);
-        .cm-attribute,
-        .cm-property,
-        .cm-keyword {
-            color: rgb(var(--color-violet-light));
-        }
-        .cm-operator {
-            color: rgb(var(--color-sky-dark));
-        }
-        .cm-variable {
-            color: rgb(var(--color-pink-light));
-        }
-        .cm-variable-2 {
-            color: rgb(var(--contrast-1100));
-        }
-        .cm-variable-3,
-        .cm-type {
-            color: rgb(var(--color-amber-darker));
-        }
-        .cm-builtin,
-        .cm-meta {
-            color: rgb(var(--color-amber-default));
-        }
-        .cm-atom {
-            color: rgb(var(--color-coral-default));
-        }
-        .cm-tag,
-        .cm-number {
-            color: rgb(var(--color-magenta-default));
-        }
-        .cm-def {
-            color: rgb(var(--color-sky-default));
-        }
-        .cm-string {
-            color: rgb(var(--color-green-dark));
-        }
-        .cm-string-2 {
-            color: rgb(var(--color-pink-default));
-        }
-        .cm-qualifier {
-            color: rgb(var(--color-grass-default));
-        }
-        .cm-error {
-            color: rgb(var(--color-white));
-            background-color: rgb(var(--color-red-default));
-        }
-        .cm-comment {
-            color: rgba(var(--code-editor-neutral-text-color), 0.7);
-        }
-    }
-
-    &.is-dark-mode {
-        --code-editor-neutral-text-color: var(--contrast-200);
-        --code-editor-background-color: var(--contrast-1600);
-        --code-editor-gutter-background-color: var(--contrast-1500);
-        --code-editor-line-number-text-color: var(--contrast-1000);
-        --code-editor-cursor-color: var(--color-amber-default);
-
-        --code-editor-selection-highlight-color: var(--color-cyan-light);
-        --code-editor-selection-lowlight-color: var(--contrast-1700);
-
-        .cm-attribute,
-        .cm-property,
-        .cm-keyword {
-            color: rgb(var(--color-purple-lighter));
-        }
-        .cm-operator {
-            color: rgb(var(--color-sky-lighter));
-        }
-        .cm-variable {
-            color: rgb(var(--color-pink-light));
-        }
-        .cm-variable-2 {
-            color: rgb(var(--contrast-300));
-        }
-        .cm-variable-3,
-        .cm-type {
-            color: rgb(var(--color-yellow-light));
-        }
-        .cm-builtin,
-        .cm-meta {
-            color: rgb(var(--color-amber-default));
-        }
-        .cm-atom {
-            color: rgb(var(--color-coral-light));
-        }
-        .cm-tag,
-        .cm-number {
-            color: rgb(var(--color-magenta-light));
-        }
-        .cm-def {
-            color: rgb(var(--color-sky-light));
-        }
-        .cm-string {
-            color: rgb(var(--color-lime-light));
-        }
-        .cm-string-2 {
-            color: rgb(var(--color-pink-lighter));
-        }
-        .cm-qualifier {
-            color: rgb(var(--color-grass-light));
-        }
-        .cm-error {
-            color: rgb(var(--color-white));
-            background-color: rgb(var(--color-red-default));
-        }
-        .cm-comment {
-            color: rgba(var(--code-editor-neutral-text-color), 0.4);
-        }
-    }
 
     &.readonly {
         .CodeMirror-focused {
@@ -232,22 +231,17 @@
 
     .CodeMirror-foldmarker {
         position: relative;
-        @include mixins.is-elevated-clickable();
-        color: transparent;
+        @include mixins.is-elevated-clickable($color: transparent);
         text-shadow: none;
 
         padding: 0 0.5rem;
         border-radius: 1rem;
 
         background: {
-            image: url("data:image/svg+xml; utf8, <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 32 32' xml:space='preserve'><circle cx='16' cy='16' r='2'/><circle cx='26' cy='16' r='2'/><circle cx='6' cy='16' r='2'/></svg>");
+            image: url("data:image/svg+xml; utf8, <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 32 32' xml:space='preserve'><circle fill='rgba(186,186,192,0.7)' cx='16' cy='16' r='2'/><circle fill='rgba(186,186,192,0.7)' cx='26' cy='16' r='2'/><circle fill='rgba(186,186,192,0.7)' cx='6' cy='16' r='2'/></svg>");
             size: contain;
             repeat: no-repeat;
             position: center;
-        }
-
-        .is-dark-mode & {
-            background-color: rgb(var(--contrast-1200));
         }
     }
 
--- a/src/components/dialog/dialog.scss
+++ b/src/components/dialog/dialog.scss
@@ -21,7 +21,7 @@
  */
 
 :host {
-    --dialog-background-color: rgb(var(--contrast-100));
+    --dialog-background-color: var(--lime-elevated-surface-background-color);
     --header-heading-color: var(--dialog-heading-title-color);
     --header-subheading-color: var(--dialog-heading-subtitle-color);
     --header-supporting-text-color: var(--dialog-heading-supporting-text-color);
@@ -50,6 +50,11 @@ $responsive-body-padding: 3vw; // 3% of viewport's width
             }
         }
     }
+
+    .mdc-dialog__scrim {
+        background-color: rgba(var(--color-black), 0.4);
+    }
+
     .mdc-dialog__container {
         height: 100%;
         width: var(--dialog-width, auto);
@@ -67,6 +72,7 @@ $responsive-body-padding: 3vw; // 3% of viewport's width
     }
 
     .mdc-dialog__content {
+        color: var(--mdc-theme-on-surface);
         padding: var(
                 --dialog-padding-top-bottom,
                 min(1.5rem, $responsive-body-padding)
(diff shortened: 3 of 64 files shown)


anderssonjohan:
@adrianschmidt I'm not sure if this is in the scope of this issue, but the lists seems off in their foreground color, both text + action menu icon. They are barely readable on my MBP retina screen.

Screen Shot 2021-09-29 at 19 20 57

adrianschmidt:
> I'm not sure if this is in the scope of this issue, but the lists seems off in their foreground color, both text + action menu icon. They are barely readable on my MBP retina screen.

Yeah, we don't support dark mode yet. We need to go over all components and make sure they work first…

But at least, with the switch in the docs, it's easy for us to switch to dark mode to work on the dark mode support, while it's still easy for people who just want to read the docs to switch to light mode.

It's a compromise at the moment…

adrianschmidt:
By the way, this PR is outdated, since we already have the option to switch between dark and light mode in the docs, but I've left it open because I think there was some tweaks we had done here that might be applicable still. I just haven't taken the time to get back to it.

anderssonjohan:
@adrianschmidt Yeah, that's why I wasn't sure if I understood this PR correctly. The button already exists and it's great!! :)

Should I create a separate issue specifically targeting dark mode support for the list component or do you think we should list whatever we find in one "dark mode support" issue? 

IMHO, I would suggest that we add:
* a darkmode label
* a dark mode milestone (to track our progress of completing the dark mode support)
* start off with creating one issue per component (that we find issues with)

Also, what do you think about adding a checkbox to the PR review checklist (like at the end of browsers tested) that simple says "Dark mode" ?

adrianschmidt:
> IMHO, I would suggest that we add:
> 
> * a darkmode label
> * a dark mode milestone (to track our progress of completing the dark mode support)
> * start off with creating one issue per component (that we find issues with)

❤️❗ I'll do all of these right now 😄 (Ok, maybe not going through every component and adding issues for them, but adding the label, the milestone, and one or two issues to get us started.)

> Also, what do you think about adding a checkbox to the PR review checklist (like at the end of browsers tested) that simple says "Dark mode" ?

Yeah, that's a good idea. It's hard to know if anyone actually uses the checklist, since no one but me actually checks any boxes, but perhaps people still use it as a reminder of what to check for… 🤷‍♂️

adrianschmidt:
https://github.com/Lundalogik/lime-elements/milestone/1

@anderssonjohan @Kiarokh @LucyChyzhova

LucyChyzhova:
Sorry for the late reply! @anderssonjohan do you need my help with the design of the "List with action menu"? It looks like there is some room for improvements and my guess is that there are other elements/components that need a redesign?

anderssonjohan:
> Sorry for the late reply! @anderssonjohan do you need my help with the design of the "List with action menu"? It looks like there is some room for improvements and my guess is that there are other elements/components that need a redesign?

If you mean fixing the CSS so the foreground color works better in dark mode - go crazy. I haven't started anything regarding that. For other redesigns of the list, let's put up a separate issue for that, depending on what you had in mind.

LucyChyzhova:
Yes, I will try to help with CSS first. It would be great if we could have a chat. Since Kia is on vacation I need more context regarding it

src/style/color-palette-extended.css
```diff
@@ -165,8 +165,148 @@
     --color-glaucous-darker: 37, 71, 88; /* #254758 */
 }
 
+:root[data-theme="force-dark"] {
```
Kiarokh, line 168: I'm not sure if this strategy is gonna work for us. The reason is if you write a specific darkmode style for a limel-component, using `@media (prefers-color-scheme: dark)`, the styles will not get applied by fiddling the (🌙 dark / ☀️ light) toggle of Kompendium. Which means it won't work in the webclinet neither, if we do the same strategy there too. You can see the styles only if you change the dark/light using system preferences.

I hope this video shows what I mean.

https://user-images.githubusercontent.com/35954987/210054586-c2b1605d-9804-442d-ac37-93d3e4544ce9.mov

Not sure how to solve this. Do we need to consider adding some `:root[data-theme="force-dark"]` style as well as `@media (prefers-color-scheme: dark)`, and basically duplicate whatever styles we are writing for the dark mode?

src/style/shadows.scss
```diff
@@ -109,3 +109,105 @@
         -1px -1px 0 1px rgba(var(--color-black), 0.07) inset,
         0 0 1px 1px rgba(var(--color-white), 0.98) inset;
 }
+
+@media (prefers-color-scheme: dark) {
```
Kiarokh, line 113: Due to what I explained in an earlier comment, I have not used any magic like 
````scss
:root[data-theme="force-dark"]` {
````
or 
````scss
@media (prefers-color-scheme: dark) {
    :root:not([data-theme="force-light"]) {
````

Not sure if that's the right way forward. 
If it is, then we need to use some `mixin` instead of copy/pasting the code into each block.

Comment (adrianschmidt) [reply src/style/shadows.scss:113]: I think we need a mixin then, because I really think we'll want to have support for a button to switch between light and dark mode… For one thing, I really want support for it in the docs, because I don't want to have to switch my system setting every time I want to test a component, but secondly, I think we'll want to have some setting like that in the Lime CRM web client too (but as I said, even if we wouldn't, I still want it in the docs).

Comment (Kiarokh) [reply src/style/shadows.scss:113]: We absolutely has to support the toggle button. But my question was, is this the only way to override `prefers-color-scheme: dark`? Can't we fake it and say to the browser that:
> No no... right now the user doesn't prefer dark color-scheme. Please ignore whatever they OS says, and run the light color-scheme.

If there was a way for this, we wouldn't have to make a duplicate copy of everything then.

I haven't googled it yet. But it's worth to give it a try.

Comment (adrianschmidt) [reply src/style/shadows.scss:113]: I don't think that's possible. I think I googled that a bit when we added dark mode support to Kompendium… but it's absolutely worth looking into 👍

By the way, we have a related issue in the _Color system_  design guideline document. The highlighting of the brand colours determines which swatches to highlight based only on the system setting for dark mode, and doesn't account for the user's setting:

https://user-images.githubusercontent.com/154725/210375826-12396b8e-d0bd-4064-8b4f-4dcd32c3e2d2.mov

Comment (Kiarokh) [reply src/style/shadows.scss:113]: Ok. cool. I think until we figure out whether there is another way or not, I'll copy the mixin that you created for Kompendium and use it here in Elements.

We will look into the issue you mention as well.

File: src/style/_theme-color-variables.scss [hunk 3]
```diff
@@ -36,4 +38,23 @@
...
+    // them with the darkest contrast swatch color, in dark mode.
+    // To avoid that, and instead of individually adding a
+    // `@media (prefers-color-scheme: dark)` query to CSS styles of
+    // each elevated surface, we use this CSS variable instead.
+    // Consumers can then tweak this, if they have used the component on a
+    // background color which interferes with this lightened surface,
+    // or provides a poor contrast.
+    --lime-elevated-surface-background-color: rgb(var(--contrast-100));
+    @include mixins.in(dark-mode) {
+        --lime-elevated-surface-background-color: rgb(var(--contrast-200));
```
Kiarokh, line 58: @adrianschmidt this mixin for dark mode will not work. It doesn't since this containing mixin (`theme-color-variables`) is used inside a `:root` in:
https://github.com/Lundalogik/lime-elements/blob/75aed282825b6384dedf616f7e46cc2a65efba9b/src/global/core-styles.scss#L8

and used inside a `:host` in:
https://github.com/Lundalogik/lime-elements/blob/75aed282825b6384dedf616f7e46cc2a65efba9b/src/style/internal/lime-theme.scss#L47

The solution would be that I move this new variable out to another file, for example `core-styles.scss`.

However, I don't think we need to import this mixin in `lime-theme.scss` file generally. The mixin only contains a bunch of CSS variables currently. Maybe hostorically it contained more SCSS variables. Isn't it enough that the mixin is used on `core-styles.scss`? Wouldn't they all be available everywhere via `:root`?

Kiarokh:
### Known issues

- checkboxes in limel-list don't get the styles correctly 
image

src/components/code-editor/code-editor.scss
```diff
@@ -9,135 +9,134 @@
...
     display: flex;
     font-size: var(--code-editor-font-size, 0.875rem); // 14px
+    @include light-mode-styles;
+}
+
+@include mixins.in(dark-mode) {
+    :host {
+        @include dark-mode-styles;
+    }
 }
```
Kiarokh, line 133: This one of course doesn't work, and I have no idea how to solve it.
The problem is we want certain styles for the dark-mode of the code editor.
We don't want to simply switch the colors to their dark-mode variants.
However, our `in(dark-mode)` mixin doesn't work, since it's targeting the `:root` and we can't do it from within the component.
Maybe we should add `data-theme='force-light'` and `data-theme='force-dark'` to the `<Host />` of limel-code-editor (instead of those classes of `.is-dark-mode` and `.is-light-mode` which we currently add); and then use some CSS like below to style the dark mode 🤔 I don't know. What do you say?

```scss
@media (prefers-color-scheme: dark) {
            :host(:not([data-theme='force-light'])) & {
                @content;
            }
        }
        :host([data-theme='force-dark']) & {
            @content;
        }
```

src/index.html
```diff
@@ -25,7 +25,7 @@
     <script type="module" src="/assets/kompendium/kompendium/kompendium.esm.js"></script>
     <script nomodule src="/assets/kompendium/kompendium/kompendium.js"></script>
 
-    <link href="/style/color-palette-extended-light-mode-only.css" rel="stylesheet" />
+    <link href="/style/color-palette-extended.css" rel="stylesheet" />
```
Kiarokh, line 28: Do we need to declare a BREAKING CHANGE in this commit message?

src/style/shadows.scss
Kiarokh: ### Other known issues

- lock icon is hard to see 
    image

src/index.html
adrianschmidt: Not for this specific change at least, since this only applies to the docs. We might need to declare a breaking change because of other changes though. Possibly because of the changes in `color-palette-extended.css`, but that shouldn't really be used by anyone yet, since it wasn't working before now. If anyone was, I guess this is to be considered a major bug fix 😆

adrianschmidt: I'll test this PR with `main` of lime-webclient. That should tell us if there are breaking changes…

Kiarokh: Thnx. I still have to do many things in the web client PR. But check it out the current. Link in the description.

src/design-guidelines/color-system/color-system.md
```diff
@@ -65,30 +105,30 @@ Colors should be used to communicate meaning, and attract attention to things th
...
 
-<limel-example-colors-in-components></limel-example-colors-in-components>
+What colors you can optionally change to communicate meaning:
 
+-   In buttons: text color, and also background color of buttons with certain functionalities (delete, restore, etc)
+-   In badges: color of the page and its text color
+-   In linear progress: color of the progressed section and what’s left (its background)
+-   etc…
 
+<limel-example-colors-in-components></limel-example-colors-in-components>
```
adrianschmidt, line 125: The colours for the buttons have broken here.

image

src/design-guidelines/shadows/shadows-and-depth.md
```diff
@@ -55,25 +59,29 @@ It's crucial that an interaction (like hovering or pressing) that changes a visu
...
 
 ### 2. Shadow variables for surfaces
+
 In a digital user interface, there are many floating UI elements like cards, modals, menus, or popovers that can benefit from a shadow effect. Here is a list of the various CSS variables for shadows that Lime Elements offers:
 
 <limel-example-surface-shadows></limel-example-surface-shadows>
 
 ### 3. Shadow variables for highlighting different states
+
 <limel-example-surface-shadows-states></limel-example-surface-shadows-states>
```
adrianschmidt, line 75: image

I never noticed it in light mode, but I just have to point out how god damn good it looks that the drop shadow is on top of the focus ring here! 😍 Chef's kiss! 👌

src/components/button/button.scss
```diff
@@ -19,9 +19,6 @@
 }
```
adrianschmidt, line 19: Colours in these examples in the Action button guidelines are broken:

This PR:
image
image


Latest `next`:
image
image

src/components/button/button.scss
```diff
@@ -19,9 +19,6 @@
...
 
 button {
-    @include mixins.is-elevated-clickable();
```
adrianschmidt, line 22: Shadows disappear when button is disabled. Maybe that's on purpose though?

image

src/components/button/button.scss
adrianschmidt: image

src/design-guidelines/shadows/shadows-and-depth.md
```diff
@@ -13,27 +14,30 @@ Despite skeuomorphism having already been replaced with more minimalistic UI des
...
+
 When correctly used, shadows can improve usability of UI elements. We can use shadows to:
-- create a sense of depth,
-- improve users' perception of clickable elements,
-- emphasize the hierarchy of sections of the user interface,
-- signal that a component is temporary, by placing it "on top of" a context,
-- clarify the direction of movement of an element, whether it is animated as a result of a user interaction, such as pressing a button, or to indicate transition from one state to another,
-- and even visualize distinct surfaces by clarifying their edges.
+
+-   create a sense of depth,
```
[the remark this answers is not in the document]
Kiarokh: 🤔 don't know. Did I add it now? I think it was there right? Lint was doing some fixes automatically. Maybe one of them?

src/design-guidelines/shadows/shadows-and-depth.md
Kiarokh: I'm glad you noticed it. Yes, the effect becomes really sexy. Specially the red one with the errors ;)

src/components/button/button.scss
Kiarokh: Ah good catch. Thanks! I'll fixup soon

adrianschmidt, line 19: Is it intentional that the text colour becomes black when a non-primary button is hovered?

image

adrianschmidt: image

adrianschmidt: (Note the text colour of the last button in this last screenshot.)

Kiarokh: This seems to be due to this commit 94663859f38863d4a88850154d8035b7b3cec74e

I'm not sure if I understand why we should import the lime-theme into each and every component. Currently too tired to figure out why. 

But adding the following to `button.scss` will fix this issue:
```scss
@use '../_theme-color-variables';
:host {
    @include theme-color-variables.theme-color-variables;
}
```

But I don't know if other components are affected by the same thing. So maybe we should simply droip this commit. 
I also realizes that I effed up a little bit in this commit and have some of the removal of old lime brand colors which should be on the next commit (a73dbb74ef222c6a1d99bbdcc0a6e524c47ff8d6) in here by mistake.

src/components/button/button.scss
Kiarokh: This was a side-effect but I think it's fine. A disabled button shouldn't have the "affordance" of being clicked. So no shadows would still work fine.

src/components/dialog/dialog.scss
```diff
@@ -21,7 +21,7 @@
  */
```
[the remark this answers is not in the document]
Kiarokh: yes, this is old. We need to do something about it

src/components/button/button.scss
Kiarokh: caused by the previous problem

src/components/dock/dock-button/dock-button.scss
```diff
@@ -6,7 +6,9 @@
     all: unset;
```
adrianschmidt, line 6: This is from the example _Using CSS color variables for theming the Dock_, and the bright button (_Calls_) is being hovered.

(BTW, in all of these screenshots, when there are two copies side-by-side like this, unless I say anything else, it's the published docs for this PR on the left, and the latest published `next` on the right. I also check mostly in light mode, since that's where we primarily care about not breaking stuff.)

I assume it's not really a bug that the background colour is bright on hover, because it doesn't look like the hover colour is changed by the example, but I wanted to flag it to be on the safe side.

image

src/components/code-editor/code-editor.scss
adrianschmidt: I'll try to remember to look into this.

src/design-guidelines/shadows/shadows-and-depth.md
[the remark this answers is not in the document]
adrianschmidt: It was added now… 🤷‍♂️

No biggie.

src/components/button/button.scss
adrianschmidt: > I'm not sure if I understand why we should import the lime-theme into each and every component. Currently too tired to figure out why.

Yeah, that's a bit wierd…

Maybe we aren't actually setting the variables on `:root` in the docs? 🤔

> But adding the following to button.scss will fix this issue:

That seems like a good solution, rather than doing it everywhere, like we did before.

> But I don't know if other components are affected by the same thing. So maybe we should simply droip this commit.

I'm looking through literally _every_ example in the docs, comparing side-by-side with the latest `next`, so if anything else is affected, I'll hopefully catch it 😊

src/components/button/button.scss
adrianschmidt: I agree, it's nice now that I've gotten used to it 👍

src/components/button/button.scss
adrianschmidt: > I also realizes that I effed up a little bit in this commit and have some of the removal of old lime brand colors which should be on the next commit (https://github.com/Lundalogik/lime-elements/commit/a73dbb74ef222c6a1d99bbdcc0a6e524c47ff8d6) in here by mistake.

If you haven't already fixed that, I can do it for you, so we don't get competing force-pushes again 😄

src/index.html
adrianschmidt: I will test it with the latest `main` of lime-webclient before this PR is merged, as that's what's going to tell us if there are breaking changes. Once this PR is merged, I'll go ahead and test what the dark mode actually looks like in your branch 👍

src/design-guidelines/color-system/color-system.md
adrianschmidt: ⚡

src/design-guidelines/shadows/shadows-and-depth.md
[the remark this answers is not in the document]
adrianschmidt: Ah, this was the commit message:

> chore(design-guidelines): lint markdown code in the documentations
>
> Seems to be some change on our lint settings.
> It happened automatically, when I was updating information here.

Just to satisfy my curiosity, I removed this commit, and then ran `npm run lint:fix`, and it did not make any changes to this file. So I'm guessing this happened due to some plugin you have in VSCode. I've actually wanted to include our markdown files in the linting and formatting for a long time, so it's nice in a way. It would be nice to have it controlled by our lint script though, so it doesn't just happen on some people's computers and not on other's. But let's not worry about that right now. Just be aware that you might end up with loads of formatting changes in a markdown file if you change anything in it. But it looks to me like you managed to isolate the formatting changes to its own commit, so that's great! 👍

src/components/button/button.scss
adrianschmidt: ⚡

adrianschmidt: I assume "the previous problem" refers to the problem with the button colours? I thought that was the case too, but that problem has been fixed now, while this issue remains.

But perhaps this wasn't fixed because I opted to only import those colour variables in `button.scss`, not remove the whole commit?

Regardless, this doesn't seem to be a problem on buttons where the colours have been set explicitly by the consumer, and with the default colours this certainly isn't a show-stopper, in my opinion.

src/components/icon-button/icon-button.scss
```diff
@@ -17,7 +17,9 @@ $height-of-limel-button: functions.pxToRem(36);
 }
```
adrianschmidt, line 17: image

There are some minor differences with limel-icon-button in this header example.

(I don't know why we've decided to set these weird colours on the _first_ example for limel-header? 😄 That's nothing new though, so not relevant for this PR.)

src/design-guidelines/color-system/examples/contrast-color-palette.scss
```diff
@@ -1,12 +1,6 @@
 @use '../../../style/functions';
```
adrianschmidt, line 1: Colours for linear-progress are not working:

image

adrianschmidt: ⚡

adrianschmidt, line 1: The default snackbar colour seems to have changed. At first I thought it was because I was running with the system in dark mode and Kompendium in forced light mode, and that the snackbar had always been white in true light mode (because I never use that, so I wouldn't know), but when I switched the system to light mode there was no difference.

image

Not a show-stopper in my opinion. I'm noting it here so we can fix it eventually. I would prefer if we _don't_ fix it now, because depending on the fix, I might have to review all the components again… 😅

adrianschmidt, line 1: The colour of the vertical ellipsis in split-button has changed.

image

(NOT a show-stopper. Fix _after_ merge please 🤗)

adrianschmidt, line 1: I've got to send some love for this new look on limel-switch too! 😍 👏

imageimage

(New on the left, if anyone happens to read this who doesn't already know 😊)

Kiarokh: Yes, this kind of became a logical result of all the changes. 
I write more in this issue: https://github.com/Lundalogik/lime-elements/issues/2027

src/components/dock/dock-button/dock-button.scss
Kiarokh: Oh nice catch. I'll fix this

src/style/color-palette-primary.scss.bak
```diff
@@ -104,6 +104,7 @@
     --color-glaucous-light: 135, 174, 193; // #87aec1
```
adrianschmidt, line 104: Issues when running in lime-webclient `main`:

There are some shadows that are markedly more pronounced when the system is in dark mode. That's no show-stopper though.

Worse is the navigation arrows that disappear from limel-table.

https://user-images.githubusercontent.com/154725/212094251-ad927775-3994-409a-a744-78d75c70269e.mov

I had missed the "missing" navigation arrows in the docs, but taking a closer look now, they are indeed white when the system is in dark mode, even when Kompendium is in forced light mode:

image

Kiarokh: Oh yes. We don't have a separate shadows file. Like `light-mode-shadows-only.css`, that's why if you are in dark mode, shadows will get super dark.

And also the white arrows is because your OS is in dark mode. Long story, but they are specifically made for table in dark mode. Because otherwise, in dark mode, the arrows would be not visible.

src/components/code-editor/code-editor.scss
adrianschmidt: Did this get solved? The code editor seemed to work just fine in dark and light mode? Or did I miss something? I primarily checked the light mode, since that's where there could be breaking changes…

Kiarokh: oh crap. no

Kiarokh: I have no idea how to solve it. 😢

src/style/color-palette-primary.scss.bak
adrianschmidt: I think we need to find _some kind_ of fix for the arrows, but we'll fix it after merging this. I'm thinking white arrows with a thin black border, and black arrows with a thin white border, would at least make sure the arrows are visible regardless of background colour.

Kiarokh (approved):
I approve on behalf of @adrianschmidt 
We still have a couple of issues, but will fix in upcoming PRs

PR #1253, opened 2021-05-21
the slider be able to use the configured step incase decimals/percentage

Befkadu1:
fix Lundalogik/crm-feature#2114

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/slider/slider.tsx
+++ b/src/components/slider/slider.tsx
@@ -40,8 +40,9 @@ export class Slider {
 
     /**
      * Default value: 1.
-     * The factor that the properties `value`, `valuemax` and `valuemin` are multiplied by.
-     * On `change` divides the value by the factor, so the original format stays the same
+     * The factor that the properties `value`, `valuemax`, `valuemin`, and
+     * `step` are multiplied by. On `change` divides the value by the factor,
+     * so the original format stays the same.
      */
     @Prop({ reflect: true })
     public factor: number = 1;
@@ -175,7 +176,7 @@ export class Slider {
                         aria-valuenow={this.multiplyByFactor(this.getValue())}
                         aria-label={this.label}
                         aria-disabled={this.disabled || this.readonly}
-                        data-step={this.step}
+                        data-step={this.multiplyByFactor(this.step)}
                     >
                         <div class="mdc-slider__track-container">
                             <div class="mdc-slider__track" />


src/components/slider/slider.tsx
```diff
@@ -175,7 +175,7 @@ export class Slider {
                         aria-valuenow={this.multiplyByFactor(this.getValue())}
                         aria-label={this.label}
                         aria-disabled={this.disabled || this.readonly}
-                        data-step={this.step}
+                        data-step={this.multiplyByFactor(this.step)}
```
adrianschmidt, line 178: Yeah, it probably makes more sense like this… 😊👍

This is a breaking change though, so we should mark it as such, to get a new major version when this is merged.

adrianschmidt (changes requested):

src/components/slider/slider.tsx
adrianschmidt: (I can fix it.)

src/components/slider/slider.tsx
```diff
@@ -175,7 +176,7 @@ export class Slider {
...
```
adrianschmidt, line 179: Before we merge a breaking change, I just want to get a second opinion from @jgroth: do you agree that it's reasonable that `step` should also be multiplied by `factor`?

Example:

We have a slider that goes from 0 to 100. The value is represented as a decimal number, from 0 to 1. Thus, we have a `factor` of `100`. If we want the slider to use steps of 5 in the UI, corresponding to stepping the value in multiples of 0.05, before this PR we'd set `step={5}`:

```tsx
this.value = 0.3 // Visualised by the slider as "30"

<limel-slider
	value={this.value}
	valuemin={0}
	valuemax={1}
	factor={100}
	step={5}
/>
```

After this PR, we'd instead set `step={0.05}`:

```tsx
this.value = 0.3 // Visualised by the slider as "30"

<limel-slider
	value={this.value}
	valuemin={0}
	valuemax={1}
	factor={100}
	step={0.05}
/>
```

Of course, how reasonable this looks may depend somewhat on where the `factor={100}` is placed 😄:

```tsx
<limel-slider
	value={this.value}
	valuemin={0}
	valuemax={1}
	step={5}
	factor={100}
/>
```

vs

```tsx
<limel-slider
	value={this.value}
	valuemin={0}
	valuemax={1}
	step={0.05}
	factor={100}
/>
```

jgroth: Yeah I agree, the old behavior was weird, having a step of `5` for a value between `0` and `1` does not make sense :smile: 
I'm not sure I agree on the breaking change though. Sure, there is a change in behavior but that can be said for almost any bug fix

adrianschmidt: I'll copy my reply to Bef when he asked about that:

> It's breaking for anyone already using the slider with multiplier and step if they already included the multiplier in the step themselves.
>
> `<limel-slider value={this.value} multiplier="100" step="5" />` for example. When this fix is released, they'd have to change that code to `<limel-slider value={this.value} multiplier="100" step="0.05" />`.

The [Semantic Versioning docs](https://semver.org/) says (my _cursive_):

> Given a version number MAJOR.MINOR.PATCH, increment the:
>
> - MAJOR version when you make incompatible API changes,
> - MINOR version when you add functionality in a backwards compatible manner, and
> - PATCH version when you make _backwards compatible bug fixes_.

I don't think this can be considered a backwards compatible bug fix 🤷‍♂️

adrianschmidt: > Yeah I agree, the old behavior was weird, having a step of 5 for a value between 0 and 1 does not make sense 😄

Oh, and thanks for the confirmation! I think that's the most important, because whether we do a major bump or a patch bump, I wouldn't want to merge it an only _then_ hear a very convincing argument from you about why this change doesn't make any sense 😅🙈

adrianschmidt, line 179: Just for the record:

I found this little visual bug when testing in the web client:

https://user-images.githubusercontent.com/154725/119490879-9dae8300-bd5d-11eb-94ec-e8dd01e6a3b3.mov

The underlying value is still just `0`, the value can be saved, and once the card is loaded again, it shows as `0`, as expected. The bug is there before this change too, but it seems it's easier to trigger when the `step` property is used. Without this fix (and the corresponding fix in lime-crm-components), the width of the area that triggers the bug is only 1px wide. With this fix, and the `step` set to 1/10 of the max value, the bug is triggered by an area approximately 10-15px wide.

adrianschmidt (approved):

PR #1261, opened 2021-05-31
fix(input): cursor jumps to end of input when rapidly typing

### LINKED ISSUE #1255: Limel input: Text cursor jumps to end of input when rapidly typing
## Current behavior

When adding text in the middle of an existing string in the input field the cursor sometimes jumps to the end of the input.
This seems to happen when rapidly typing.

This behaviour is reproducable in the live docs, but also locally.

1. Add some text to an input field
2. Place the cursor at the beginning or middle of the text
3. Type really fast
4. Cursor will randomly jump to end of text

## Expected behavior
The cursor should not move unexpectedly.

## Environment

- **lime-elements** version: 31.10.0
- Framework used: StencilJS

jensgustafsson:
When adding text in the middle of an existing string in the input field the cursor sometimes jumps
to the end of the input.

fix #1255

## Review:

### Browsers tested:

Windows:
- [x] Chrome
- [x] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -235,12 +235,8 @@ export class InputField {
         this.handleCloseMenu = this.handleCloseMenu.bind(this);
         this.setFocus = this.setFocus.bind(this);
 
-        const debounceTimeout = 100;
-        this.changeEmitter = debounce(this.changeEmitter, debounceTimeout, {
-            leading: false,
-            maxWait: 300,
-            trailing: true,
-        });
+        const debounceTimeout = 300;
+        this.changeEmitter = debounce(this.changeEmitter, debounceTimeout);
 
         this.portalId = createRandomString();
     }


src/components/input-field/input-field.tsx
```diff
@@ -235,12 +235,8 @@ export class InputField {
         this.handleCloseMenu = this.handleCloseMenu.bind(this);
         this.setFocus = this.setFocus.bind(this);
 
-        const debounceTimeout = 100;
-        this.changeEmitter = debounce(this.changeEmitter, debounceTimeout, {
-            leading: false,
-            maxWait: 300,
-            trailing: true,
-        });
```
jensgustafsson, line 243: I was able to "solve this" (not being able to repro it anymore) by removing the `maxWait` time and changing the debounce timeout to 300.

The default values for `leading` and `trailing` is already set to `false` and `true`, thus the whole config object was removed.

adrianschmidt: Nice!

I'll have my best go at "breaking" it while testing 😝

adrianschmidt (approved):

PR #1262, opened 2021-06-01
fix(input-field): prevent `formatted-input` of type `number` from rem…

Kiarokh:
…aining transparent in `readonly`
fix: https://github.com/Lundalogik/crm-feature/issues/2147

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/partial-styles/_readonly.scss
+++ b/src/components/input-field/partial-styles/_readonly.scss
@@ -1,10 +1,15 @@
 .mdc-text-field {
     &.lime-text-field--readonly {
-        .mdc-text-field__input {
+        input.mdc-text-field__input {
             background-color: transparent;
+            color: $input-text-color;
         }
 
-        .mdc-text-field__input,
+        .formatted-input {
+            color: $input-text-color;
+            opacity: 1;
+        }
+
         label.mdc-floating-label {
             color: $label-color;
         }
--- a/src/components/input-field/input-field.scss
+++ b/src/components/input-field/input-field.scss
@@ -49,10 +49,16 @@
         .mdc-floating-label {
             color: $label-color;
         }
+        .mdc-text-field__input {
+            color: $input-text-color;
+        }
     }
     &.mdc-text-field--disabled {
         .mdc-floating-label {
             color: $label-color-disabled;
+        }
+        .mdc-text-field__input {
+            color: $input-text-color-disabled;
         }
     }
 }
@@ -78,6 +84,7 @@
     position: absolute;
 
     display: none;
+    color: $input-text-color;
 }
 
 :not(.mdc-text-field--focused):not(.mdc-text-field--invalid) {
@@ -132,6 +139,18 @@ input.mdc-text-field__input {
         top: 0;
         bottom: 0;
         margin: auto;
+    }
+}
+
+.mdc-text-field {
+    &:not(.mdc-text-field--disabled) .mdc-text-field-character-counter,
+    &:not(.mdc-text-field--disabled)
+        + .mdc-text-field-helper-line
+        .mdc-text-field-character-counter,
+    &:not(.mdc-text-field--disabled)
+        + .mdc-text-field-helper-line
+        .mdc-text-field-helper-text {
+        color: $helper-text-color;
     }
 }
 
--- a/src/style/internal/shared_input-select-picker.scss
+++ b/src/style/internal/shared_input-select-picker.scss
@@ -8,8 +8,12 @@ $background-color-normal: rgba(var(--contrast-200), 0.5);
 $background-color-hovered: rgba(var(--contrast-200), 1);
 $background-color-focused: rgba(var(--contrast-100), 0.8);
 $background-color-disabled: transparent;
+
 $label-color: rgba(var(--contrast-1200), 1);
 $label-color-disabled: rgba(var(--contrast-1200), 0.5);
+$input-text-color: rgba(var(--contrast-1400), 1);
+$input-text-color-disabled: rgba(var(--contrast-1400), 0.5);
+$helper-text-color: $label-color;
 
 $height-of-mdc-text-field: 3.5rem; //This is written directly in rem, beucase the vaiable used to calculate things elsewhere
 $height-of-mdc-helper-text-block: 0.9375rem;


src/components/input-field/partial-styles/_readonly.scss
```diff
@@ -5,8 +5,10 @@
         }
 
         .mdc-text-field__input,
+        .formatted-input,
         label.mdc-floating-label {
             color: $label-color;
```
adrianschmidt, line 10: Why are we setting the color to `$label-color` for `.mdc-text-field__input` and `.formatted-input`? That doesn't seem right?

Kiarokh: For some type the color changes when disabled. For some the opacity. We want them all to be 100%  solid and same color. So I think the code covers cases that I might have missed, and also it explains what we want. I think it's OK. What do think?

adrianschmidt: Don't we have a `text-color` variable we could apply to the input, even if those happen to have the same value at the moment? If we use `label-color`, and some time in the future decide we want to have different colours for the label and the text, it will be quite unexpected that when readonly, the text gets the label colour instead of the text colour.

I just want us to take the extra minute or two now, to spare us a headache in the future. If a variable is called `label-color` it should set the colour of labels, preferably all labels, and nothing but labels.

Kiarokh: at the moment, this is the only thing we have.
image
And if one day we decide to have different colors for text and labels, that's a story for that day, and we have to take a deeper look on components which would need these different colors. 

So I'd just spare the headache for future. 😄 

(I don't think we will have different colors, since the floating label is so small and already hardly readable. Lower contrast may make it even harder to read. But if we decide to do so, it's a bigger story. For now, it's fine. We're just saying we have same color for all.)

adrianschmidt: It's very easy to create a tangled mess of spaghetti code by always thinking "we can always clean that up later". But once you've got that tangled mess on your hands, nothing is easy anymore, because whatever you change has unintended consequences that are impossible to foresee.

In my opinion, for a long-life product like this, there's only one supreme rule: we have to be able to change it. It may sound like that's self-evidently true for all software, but it's not. Unless care is taken, software — all software — will incredibly quickly devolve into an unmaintainable tangled mess that no one dares to change. The _least_ we can do is to do our future selves a service and not _actively_ trip ourselves up by using variable names that weren't even correct from the outset.

adrianschmidt (changes requested):

src/components/input-field/partial-styles/_readonly.scss
Kiarokh: I agree with your point, but I disagree that this is the case.
 
This is a very common way to use CSS variables. It semantically helps, when reading the code, the reader hopefully realizes that: _"OK, these all have the same color, which is the color of label."_ 

What's wrong with that? We do it all the time.

I might not understand what you're trying to say though… :/

adrianschmidt: The problem is when you come to it from the "other side", not when you read _this_ code, but when you want to change the color of all labels, go to `variables.scss` or wherever the variable might be defined, and change it to `$label-color: orange`. Then all labels throughout the app changes colour to orange, and everything is great. Except that when a number-input is readonly, then and only then, the _text_ also becomes orange.

That's obviously not the end of the world as an isolated case. The problem is when _every_ change made has that kind of unintended consequence somewhere in the app, because then we can't ever change _anything_ without introducing a host of bugs, that's going to piss off both our customers and us. And we'll have to spend 90 percent of our time chasing down a thousand obscure little bugs like that.

Kiarokh: I see I see. Gonna fix

adrianschmidt: Thank you! ❤️ Sorry for being so annoying about it 🙈

Kiarokh: No worries. I added a few changes. This made it bigger than what it had to be for this PR :) I think if you look at the fixes, you'll see why I was resisting to to this.

But it's fine, we had to do these things for dark mode anyway; and it's better that we're one step closer now. ;) 

p.s.
The person who just changes a variable value without checking where it's used should be suspended from school for 10 days! ☝️😑

adrianschmidt: This looks _much_ better!

> The person who just changes a variable value without checking where it's used should be suspended from school for 10 days! ☝️😑

Because knowing for sure that you've found every instance, and that every variable that has the same _name_ actually references the same _value_ is always super easy and straightforward, right?

Honestly, I don't understand why we need to keep arguing about this _every single time_? I spend _hours_, if not every day, then at least every week, arguing with you about why we need to have at least a semblance of a standard of quality in what we do. Is this _really_ actually productive? Because it seems to me like we never ever reach any kind of agreement, since it will just be the exact same argument for _hours on end_ the next time.

I think it's great that we can discuss important design decisions deeply, but I really can't see why it benefits either of us that I have to spend hours of my quite limited work time trying to explain to you exactly why, how, and in what way, things like this will cause problems in the future. Considering how often you swear over stupid shit in our old code bases, it's amazing to me that you can't seem to grasp that every single one of those seemingly idiotic choices were caused by layers upon layers of decisions like "let's just use the label-variable for this too, it'll probably be fine"…

adrianschmidt (approved):

src/components/input-field/partial-styles/_readonly.scss
Kiarokh: I'm up for quality.

I think we should just accept that we have different opinions about certain things. Using CSS variables seems to be one. 

For me a variable is there:
1. to be reused
2. to explain why things are in a certain way in the code (for the reader)
3. and to automate certain things when applicable

So for instance, if you see a button in a header, which has a height of `height: calc(var(--header-height) - 2rem);` you understand what the designer/coder expect the UI to be and behave, almost without looking at the UI.


Anyhow, I'm not against your argument generally. I'm just saying in this case, reusing the same variable wasn't considered "low quality" from my point of view. Given the fact that we didn't have any better usage of variables or our structural logic at the moment.

But now all is good, I hope.

cheers

PR #1263, opened 2021-06-01
fix(input): min and max property can be set to "0" on input field of …

### LINKED ISSUE #1259: min-property on input field
## Current behavior

I tried to put 0 as the minimum value of an input field of type number

## Expected behavior

Negative values should not be allowed

## Environment

- **lime-elements** version: @limetech/lime-elements: "31.1.3"

## Comment
If I instead put 1 as the minimum value it works as expected

jensgustafsson:
…type "number"

fix #1259

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -405,11 +405,11 @@ export class InputField {
             props.step = this.step;
         }
 
-        if (this.type === 'number' && this.min) {
+        if (this.type === 'number' && Number.isInteger(this.min)) {
             props.min = this.min;
         }
 
-        if (this.type === 'number' && this.max) {
+        if (this.type === 'number' && Number.isInteger(this.max)) {
             props.max = this.max;
         }
 


adrianschmidt (approved):

adrianschmidt:
I found a weird behaviour where if the field is emptied while it's invalid, it's behaviour "breaks" a little bit. I don't think it's anywhere near a critical bug, because as far as I could find, it only happens to number type inputs with a min- and/or max-value, and you can still change the value and everything in the field, it just behaves a little weirdly:

https://user-images.githubusercontent.com/154725/120320200-330fc100-c2e2-11eb-9bbb-577204a56dae.mov

Once triggered, when the input is emptied, it also loses focus. When focused, the notch in the border doesn't appear, even though the label floats correctly, and when a value is entered, the characters take a short while to appear.

PR #1248, opened 2021-05-20
docs: add dynamic examples with component schema

jgroth:
The new version of Kompendium will inject a schema of the component that an example belongs to. We can use this to render a form that will let the reader play around with the component without us having to write a complex example. 

This PR just me playing around a bit to try it out, we might not want it for all components :slightly_smiling_face: 

## Review:
- [ ] Commits are [atomic](https://seesparkbox.com/foundry/atomic_commits_with_git)
- [ ] Commits have the correct *type* for the changes made
- [ ] Commits with *breaking changes* are marked as such

### Browsers tested:
(Check any that applies, it's ok to leave boxes unchecked if testing something didn't seem relevant.)

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

PR #1269, opened 2021-06-02
 fix(select): better visualize invalid state

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/2166

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/select/examples/select.tsx
+++ b/src/components/select/examples/select.tsx
@@ -18,6 +18,9 @@ export class SelectExample {
     @State()
     public required = false;
 
+    @State()
+    public invalid = false;
+
     private options: Option[] = [
         { text: 'Luke Skywalker', value: 'luke' },
         { text: 'Han Solo', value: 'han', disabled: true },
@@ -35,6 +38,7 @@ export class SelectExample {
                     disabled={this.disabled}
                     readonly={this.readonly}
                     required={this.required}
+                    invalid={this.invalid}
                     onChange={this.changeHandler}
                 />
                 <p>
@@ -53,6 +57,11 @@ export class SelectExample {
                             checked={this.required}
                             label="Required"
                             onChange={this.setRequired}
+                        />
+                        <limel-checkbox
+                            checked={this.invalid}
+                            label="Invalid"
+                            onChange={this.setInvalid}
                         />
                     </limel-flex-container>
                 </p>
@@ -78,5 +87,10 @@ export class SelectExample {
     private setRequired = (event: CustomEvent<boolean>) => {
         event.stopPropagation();
         this.required = event.detail;
+    };
+
+    private setInvalid = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.invalid = event.detail;
     };
 }
--- a/src/components/select/select.scss
+++ b/src/components/select/select.scss
@@ -119,7 +119,6 @@
                 }
 
                 .mdc-select__dropdown-icon {
-                    transform: rotate(180deg);
                     color: var(--mdc-theme-primary);
                 }
             }
@@ -133,11 +132,23 @@
     }
 
     &.limel-select--invalid {
-        .limel-select-trigger .mdc-floating-label,
-        .mdc-select-helper-text {
+        .mdc-floating-label,
+        .limel-select__selected-option,
+        .mdc-select-helper-text,
+        .invalid-icon {
             color: var(--lime-error-text-color);
         }
+        .mdc-floating-label:not(.mdc-floating-label--float-above) {
+            max-width: calc(
+                100% - #{pxToRem(64)}
+            ); // leaves space for the invalid-icon
+        }
     }
+}
+
+.invalid-icon {
+    margin-right: pxToRem(8);
+    flex-shrink: 0;
 }
 
 .mdc-select-helper-line {
--- a/src/components/select/select.template.tsx
+++ b/src/components/select/select.template.tsx
@@ -55,7 +55,7 @@ export const SelectTemplate: FunctionalComponent<SelectTemplateProps> = (
 
     return (
         <div class={classList}>
-            <SelectValue {...props} hasValue={hasValue} />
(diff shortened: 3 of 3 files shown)


sampsonbryce (dismissed):

sampsonbryce:
LGTM @adrianschmidt can you double check the review pls?

adrianschmidt (changes requested):
Just "requesting changes" to prevent merge until I've reviewed 🙂

Kiarokh:
I was thinking… I wanna use `limel-icon` instead of embedding it in the CSS.

Befkadu1:
> I was thinking… I wanna use `limel-icon` instead of embedding it in the CSS.

yes it is possible to insert the icon above this line https://github.com/Lundalogik/lime-elements/blob/7f7b998d4e53274f6c816d37e956e48e660de678/src/components/select/select.template.tsx#L93

adrianschmidt:
> I was thinking… I wanna use `limel-icon` instead of embedding it in the CSS.

I think that icons that are defined by the components themselves, rather than by the consumer, are better to embed. That way, they are included for anyone else using lime-elements, even if they don't have access to lime-icons8…

(Even better if you would like to design icons @Kiarokh, then we don't have to worry about what exactly is permitted by icons8's license agreement…)

Kiarokh:
Interesting point @adrianschmidt . 
But then we already have a problems in our code, since we use icons for these purposes. 

For example, in `input-field.tsx` we have hard-coded icons that are from icons8. So consumers won't get them, right?

````tsx
	private getTrailingIcon() {
        if (this.isInvalid()) {
            return 'high_importance'; // 👈 This is the invalid icon that I'm embedding in the CSS in this PR
        }

        if (this.trailingIcon) {
            return this.trailingIcon;
        }

        if (this.showLink && this.type === 'email') {
            return 'filled_message';
        }

        if (this.showLink && this.type === 'tel') {
            return 'phone';
        }

        if (
            this.showLink &&
            (this.type === 'url' || this.type === 'urlAsText')
        ) {
            return 'external_link';
        }
    }
````

### Why embedded SVGs aren't good

Next problem is that the `currentColor` won't work in embedded SVGs. Because the embedded svg seems to become its own separate document. So it won't inherit the colors that we wanna apply with CSS. It means we have to hardcode an RGB value, and that will be on the way also when adding support to dark-mode. We can't use any of our own css variable colors neither.

## So, how to we go forward now?

For this specific case of visualizing invalid things, we don't need to use any SVG icons. We can simply add a pseudo element with `content: '!';` and some rounded corners and red color. But then I'll have to do this in all components that use the `high_importance` trailing icon or similar.

I don't know what other kinds of visualizations we have, which we use trailing icons for.

adrianschmidt:
> Interesting point @adrianschmidt .
> But then we already have a problems in our code, since we use icons for these purposes.
>
> For example, in `input-field.tsx` we have hard-coded icons that are from icons8.
> 
> ```tsx
> 	private getTrailingIcon() {
>         if (this.isInvalid()) {
>             return 'high_importance'; // 👈 This is the invalid icon that I'm embedding in the CSS in this PR
>         }
> ```

Yeah, ideally we should decide on what path to take, and normalise our code so it works the same everywhere.

> So consumers won't get them, right?

Just to clarify: any add-ons in Lime CRM Web Client, or any other product where we supply Lime Elements and lime-icons8, _will_ get the icons from lime-icons8.

But anyone using Lime Elements in an environment where lime-icons8 is not installed, will have to supply some other icons. They can do that of course, and to get any "hardcoded" icons to show up, they just need to make sure they add some icon in a file using the correct file name. For example, @Befkadu1 mentioned possibly using Lime Elements for his own site, and @anderssonjohan uses Lime Elements for a private project too, and they will have to provide their own icons for any that aren't embedded. Embedding in some instances an not others seem like the worst of both worlds, as every client that isn't using icon8 will have different icons for the same thing in different components.

> ### Why embedded SVGs aren't good
> Next problem is that the `currentColor` won't work in embedded SVGs. Because the embedded svg seems to become its own separate document. So it won't inherit the colors that we wanna apply with CSS. It means we have to hardcode an RGB value, and that will be on the way also when adding support to dark-mode. We can't use any of our own css variable colors neither.

That's only a problem when embedding in css instead of in the html. In fact, limel-icon loads the svg data from the file and embeds it in the html, precisely because that's the only way we can change the colours through css when the svg data is embedded in the html, not when just linking it in an `img` tag.

Kiarokh:
> They can do that of course, and to get any "hardcoded" icons to show up, they just need to make sure they add some icon in a file using the correct file name.

Sounds like we need to document which icons are hardcoded, and how they look, so they can do as you say.

So given the point discussed, it should be alright to add a conditional trailing icon, in the HTML. I'll do a fixup so we can see whether it's OK or not. ;)

Kiarokh:
I made an issue here for the icons:
https://github.com/Lundalogik/lime-elements/issues/1272

Added icons in the HTML instead, and pushed a fixup.

Kiarokh:
Not sure how we should go about this commit: 42547267dd0a126c4e526604e3c9646bf114f4d7

Where I just added a toggle to display invalid. Even though it's pedagogic, we don't have this toggle on other components. Should we add it there as well?

adrianschmidt:
> Not sure how we should go about this commit: [4254726](https://github.com/Lundalogik/lime-elements/commit/42547267dd0a126c4e526604e3c9646bf114f4d7)
> 
> Where I just added a toggle to display invalid. Even though it's pedagogic, we don't have this toggle on other component examples. Should we add it there as well?

I think it's nice, but we don't need to add it to a bunch of other examples right now. I think we'll get options for `invalid` with #1248 anyway.

adrianschmidt (approved):

PR #1270, opened 2021-06-08
fix(form-date-picker): check if value contains valid date string befo…

hannahu:
…re returning invalid date

fix Lundalogik/crm-feature#2196

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/widgets/date-picker.ts
+++ b/src/components/form/widgets/date-picker.ts
@@ -31,12 +31,30 @@ export class DatePicker extends React.Component {
         });
     }
 
-    private getValue() {
-        if (typeof this.props.value === 'string') {
-            return new Date(this.props.value);
-        } else {
+    private getValue(): Date {
+        if (typeof this.props.value !== 'string') {
             return this.props.value;
         }
+
+        if (!moment(this.props.value).isValid()) {
+            const dateString = this.getDateFromString(this.props.value);
+            if (dateString) {
+                return new Date(dateString);
+            }
+        }
+
+        return new Date(this.props.value);
+    }
+
+    private getDateFromString(value: string): string | null {
+        const dateRegexp = /([0-9][0-9][0-9][0-9])[-](0[1-9]|1[012])[-](0[1-9]|[12][0-9]|3[01])[T]([01][1-9]|[2][0-3])[:]([0-5][0-9])[:]([0-5][0-9])([+|-]([01][0-9]|[2][0-3])[:]([0-5][0-9])){0,1}/;
+        const dateString = value.match(dateRegexp);
+
+        if (Array.isArray(dateString) && !!dateString[0]) {
+            return dateString[0];
+        }
+
+        return null;
     }
 
     private handleChange(event: CustomEvent<Date>) {


src/components/form/widgets/date-picker.ts
```diff
@@ -31,12 +31,30 @@ export class DatePicker extends React.Component {
...
+                return new Date(dateString);
+            }
+        }
+
+        return new Date(this.props.value);
+    }
+
+    private getDateFromString(value: string): string | null {
+        const dateRegexp = /([0-9][0-9][0-9][0-9])[-](0[1-9]|1[012])[-](0[1-9]|[12][0-9]|3[01])[T]([01][1-9]|[2][0-3])[:]([0-5][0-9])[:]([0-5][0-9])([+|-]([01][0-9]|[2][0-3])[:]([0-5][0-9])){0,1}/;
+        const dateString = value.match(dateRegexp);
```
adrianschmidt, line 51: Do we really have to hardcode the date format like this? 😬

If I recall correctly, moment has some kind of function for checking if a string produces a valid date. Maybe we can use that instead of trying to build our own?

hannahu: Tried to find but might have missed, I'll check again tomorrow :P

adrianschmidt:
```ts
if (!moment(this.props.value).isValid()) {
```

🤦‍♂️😆 Seems you had already found it, I was just not looking carefully enough.

But… this is meant to find a valid date inside a string where the string as a whole _doesn't_ provide a valid date? 😬

I tried out the example date provided in the issue (`'1899-12-29T08:00:00+01:00:14'`), and it seems that the timezone segment of that type of string doesn't seem to allow for seconds to be included. `moment('2021-05-23T08:00:00+02:00:00').isValid()` also returns `false`, while `moment('2021-05-23T08:00:00+02:00').isValid()` returns `true`.

If we _really_ need to support this apparently non-standard date-format in Lime CRM, I don't think we should involve lime-elements if we can avoid it. I think we should detect the weird format in the web client (either in lime-webclient or lime-crm-components), and convert it to something standard before sending it off to lime-elements. If at all possible… 🙂

hannahu: Exactly! It felt better to take care of it when parsing the date values here in lime-elements and otherwise I need to do some special checks in crm-components like trying to parse values when getting form data from the limeobject and then this place felt better :P What do you think?

adrianschmidt: Hmm…

On the one hand, since we are checking for a valid date first, and only run this if the date string is invalid, I'm less opposed to it than when I first saw it and didn't release we did that check first.

On the other hand, I'm leaning towards thinking this shouldn't even be done in lime-webclient or lime-crm-components. This feels more like something that should be done by an add-on, that's only used for this one customer with broken data in their database… I hope it's just one customer? Has the "bug" been replicated by anyone other than Maria who reported it?

Even if it's something that's weird in Lime CRM, so that this appears in every database, I _really_ don't want us to spread that weirdness to our open sourced packages… 😣

adrianschmidt: @eketorp @jensgustafsson @jgroth Thoughts on this?

eketorp: I think I need this explained to me verbally to understand things.

adrianschmidt: Turns out this is a backend problem, so we shouldn't solve it here.
https://github.com/Lundalogik/crm-feature/issues/2196#issuecomment-861190239

I'll close this PR.

PR #1271, opened 2021-06-10
Add readonly and disabled to datepicker

jensgustafsson:
fix: https://github.com/Lundalogik/crm-feature/issues/2211
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/date-picker/date-picker.tsx
```diff
@@ -182,11 +182,12 @@ export class DatePicker {
             onAction: this.clearValue,
         };
 
-        if (this.useNative) {
+        if (this.disabled || this.readonly || this.useNative) {
```
adrianschmidt, line 185: Hmm… this doesn't work, it's not showing the existing value… 🤔

https://user-images.githubusercontent.com/154725/121667137-44a75480-caaa-11eb-8ba6-0ff6bd2e72d4.mov

image

Interestingly, it works on mobile (tested both Android and iOS) 😮

image

So, I guess it's something about using the native picker on desktop, or at least in Chrome on Mac.

src/components/date-picker/date-picker.tsx
```diff
@@ -250,6 +255,14 @@ export class DatePicker {
         this.nativeType = nativeTypeForConsumerType[this.type || 'default'];
         this.nativeFormat = nativeFormatForType[this.nativeType];
 
+        // Quarter is not a valid type.
+        // It's therefore set to type "text" such that we can set the
```
adrianschmidt, line 259:
```suggestion
        // It's therefore set to type "text" so that we can set the
```

#### The very short version:

**so that** is _much_ more common than **such that**—when in doubt, stick to **so that**.

#### Short version:

> _So that_ denotes the purpose or the effect of something or it carries the idea ‘in order that’ while _such that_ emphasizes on the manner or the way to which extent something is described.

> He checked carefully, so that the mistakes were caught.

The goal was to catch the mistakes, checking carefully was a means to an end.

> I spoke to my teacher such that she took me to the principal.

The goal wasn't to be taken to the principal, the "I" was just angry and used some impolite language. The (in this case unwanted) result was being sent to the principal.

#### Long version:

https://www.differencebetween.com/difference-between-so-that-and-vs-such-that/ 😄

src/components/input-field/input-field.tsx
```diff
@@ -371,6 +371,8 @@ export class InputField {
             <input
                 {...properties}
                 type={type}
+                readonly={this.readonly}
+                disabled={this.disabled}
```
adrianschmidt, line 375: I assume you've added these to avoid setting `disabled` when `readonly` is set, but setting `disabled` along with `readonly` is intentional, so this addition should be removed.

src/components/date-picker/date-picker.tsx
```diff
@@ -160,7 +160,11 @@ export class DatePicker {
     }
 
     public componentWillLoad() {
-        this.useNative = isIOSDevice() || isAndroidDevice();
+        this.useNative =
+            isIOSDevice() ||
+            isAndroidDevice() ||
+            this.disabled ||
+            this.readonly;
```
adrianschmidt, line 167: Hmm… I'm not sure this is the right way to go about it. This removes the formatting for a bunch of the types…

Isn't it possible to just prevent the flatpickr menu from opening instead? I can have a look at it if you'd like me to.

adrianschmidt: image

src/components/input-field/input-field.tsx
jensgustafsson: Why? 🤔  I thought we added support for both

src/components/date-picker/date-picker.tsx
jensgustafsson: > Hmm… I'm not sure this is the right way to go about it. This removes the formatting for a bunch of the types…
> 
> Isn't it possible to just prevent the flatpickr menu from opening instead? I can have a look at it if you'd like me to.

This was my quick fix for this. I agree that would be better but I couldn't find a good way to solve it. Feel free to look in to it if you want!

Aaah, the quarter picker with readonly only works in the web client then as that's were we're setting the format 🤦

jensgustafsson: Another quick fix could be to set type=text for all pickers and just render the date in a normal limel-input-field

src/components/input-field/input-field.tsx
adrianschmidt: We do have support for both. The behaviour is different if you set `disabled` or `readonly`, but if you set both, it's the same as when you only set `readonly`. The native elements have no concept of "readonly", only "disabled". So our "readonly" is basically a specific styling of a disabled field, used to indicate to the user that "the reason you can't edit this field is of a more permanent nature than if it was just disabled".

src/components/date-picker/date-picker.tsx
adrianschmidt: I'll have a look 👍

src/components/input-field/input-field.tsx
jensgustafsson: I thought we started to add support for that so I thought it was a good thing to set them

adrianschmidt: `limel-input-field` already has support for both 🙂

As far as the consumer is aware, disabled and readonly are different things. But _internally_, our `readonly` state sets the `disabled` state on the native element used. That makes sense, since we don't want readonly inputs to be enabled.

src/components/date-picker/date-picker.tsx
adrianschmidt: I've pushed an alternative version. Please have a look 🙂

adrianschmidt: Oh, btw, it's three separate commits. The first one is just some boy-scouting, the second is the fix, and the third is just adding the composite example to the docs. So I'd suggest focusing on the middle commit 🙂

jensgustafsson (commented):
I'm not allowed to "approve" since I opened the PR but I do. Great fix!

PR #1277, opened 2021-06-18
feat(file): add a visual dropzone

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/1800

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -389,6 +389,7 @@ export class ChipSet {
                             this.value.length || this.editMode
                         ),
                     }}
+                    dropzone-tip={this.dropZoneTip()}
                 >
                     <div class="mdc-notched-outline__leading"></div>
                     <div class="mdc-notched-outline__notch">
@@ -414,6 +415,10 @@ export class ChipSet {
             </div>
         );
     }
+
+    private dropZoneTip = (): string => {
+        return translate.get('file.drag-and-drop-tips', this.language);
+    };
 
     private isFull(): boolean {
         return !!this.maxItems && this.value.length >= this.maxItems;
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -169,8 +169,6 @@ export class ChipSet {
     private mdcChipSet: MDCChipSet;
     private mdcTextField: MDCTextField;
     private handleKeyDown = handleKeyboardEvent;
-    private clearAllChipsLabel: string;
-    private removeChipLabel: string;
 
     constructor() {
         this.renderChip = this.renderChip.bind(this);
@@ -229,17 +227,6 @@ export class ChipSet {
     @Method()
     public async emptyInput() {
         this.syncEmptyInput();
-    }
-
-    public componentWillLoad() {
-        this.clearAllChipsLabel = translate.get(
-            'chip-set.clear-all',
-            this.language
-        );
-        this.removeChipLabel = translate.get(
-            'chip-set.remove-chip',
-            this.language
-        );
     }
 
     public componentDidLoad() {
@@ -697,11 +684,20 @@ export class ChipSet {
                 class="mdc-text-field__icon clear-all-button"
                 tabindex="0"
                 role="button"
-                title={this.clearAllChipsLabel}
-                aria-label={this.clearAllChipsLabel}
+                title={this.clearAllChipsLabel()}
+                aria-label={this.clearAllChipsLabel()}
             />
         );
     }
+
+    private clearAllChipsLabel = (): string => {
+        return translate.get('chip-set.clear-all', this.language);
+    };
+
+    private removeChipLabel = (): string => {
+        return translate.get('chip-set.remove-chip', this.language);
+    };
+
     private handleDeleteAllIconClick(event: Event) {
         event.preventDefault();
         this.change.emit([]);
--- a/src/components/chip-set/partial-styles/_file-picker.scss
+++ b/src/components/chip-set/partial-styles/_file-picker.scss
@@ -1,0 +1,79 @@
+:host {
+    --dropzone-margin: 0.25rem;
+    --dropzone-color: rgb(var(--contrast-600));
+}
+
+:host(.highlight-dropzone) {
+    --dropzone-color: rgb(var(--color-lime-light));
+}
+
+:host(.is-file-picker.shows-dropzone) {
+    .mdc-text-field:not(.has-chips) {
+        .mdc-notched-outline {
+            &:before,
+            &:after {
+                box-sizing: border-box;
+                position: absolute;
+                margin: auto;
+                top: var(--dropzone-margin);
+                right: var(--dropzone-margin);
+                bottom: var(--dropzone-margin);
+            }
+
+            &:before {
+                // dropzone tip
+                content: attr(dropzone-tip);
+
+                display: inline-flex;
+                align-items: center;
+                justify-content: flex-end;
+
+                padding: 0 functions.pxToRem(12);
+
+                font-size: functions.pxToRem(13);
+                color: var(--mdc-theme-text-secondary-on-background);
+            }
+
+            &:after {
+                // dropzone visualization
+                content: '';
+
+                left: var(--dropzone-margin);
+
+                border-radius: functions.pxToRem(6);
+                border: 2px dashed var(--dropzone-color);
+
+                cursor: pointer;
+            }
+        }
+
+        &.mdc-text-field--disabled {
+            .mdc-notched-outline {
+                &:before,
+                &:after {
+                    // Drop zone visualization
+                    @include shared_input-select-picker.looks-disabled;
+                }
+            }
+        }
+    }
+}
+
+// below block prevents overlapping of `label` and dropzone tip text
+$max-width-of-dropzone-tip: 58%;
+$max-width-of-label-when-not-readonly: 40%;
+
+:host(.is-file-picker.shows-dropzone) {
+    .mdc-text-field:not(.has-chips):not(.lime-text-field--readonly) {
+        .mdc-notched-outline {
+            &:before {
+                width: $max-width-of-dropzone-tip;
+                text-align: right;
+            }
+
+            .mdc-notched-outline__notch {
+                max-width: $max-width-of-label-when-not-readonly;
+            }
+        }
+    }
+}
--- a/src/components/file/file.tsx
+++ b/src/components/file/file.tsx
@@ -1,4 +1,5 @@
-import { Chip, FileInfo } from '@limetech/lime-elements';
+import translate from '../../global/translations';
+import { Chip, FileInfo, Languages } from '@limetech/lime-elements';
 import { MDCTextField } from '@material/textfield';
 import {
     Component,
@@ -7,6 +8,7 @@ import {
     EventEmitter,
     h,
     Prop,
+    State,
 } from '@stencil/core';
 import { createRandomString } from '../../util/random-string';
 
@@ -96,6 +98,12 @@ export class File {
     public accept: string = '*';
 
     /**
+     * Defines the localisation for translations.
+     */
+    @Prop()
+    public language: Languages = 'en';
+
+    /**
      * Dispatched when a file is selected/deselected
      */
     @Event()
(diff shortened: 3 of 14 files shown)


src/components/file/examples/file-no-dropzone.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { Component, h, State } from '@stencil/core';
+/**
+ * No dropzone visualization
+ *
+ * You can hide the dropzone visualization by setting `visualize-Dropzone` property
+ * to `false`.
+ * :::note
+ * - Even if you hide the visualization, the component will still support
+ * drag and drop. But only the visualization of the dropzone will dissapear.
+ * - A `readonly` file picker does not show the dropzone.
```
Befkadu1, line 11: It is good to mention that `readonly` disables adding and removing files.

Befkadu1 (changes requested):

Befkadu1:
@adrianschmidt can you look this PR a bit if you have time especially the styling part?

src/translations/sv.ts
```diff
@@ -3,4 +3,6 @@ export default {
     'date-picker.quarter.heading': 'Kvartal',
     'date-picker.year.heading': 'År',
     'chip-set.clear-all': 'Rensa alla',
+    'file.drag-and-drop-tips':
+        'Drag and drop your file here, or click to browse.',
```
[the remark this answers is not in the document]
adrianschmidt: Good catch!

src/components/file/examples/file-no-dropzone.tsx
adrianschmidt: Hmm… I'm not sure, but I think I disagree. For one thing, I think it ought to be fairly self-evident that files can't be added and removed when the component is `readonly`. But, even if that _does_ need explaining, I think it should be explained in the docs for `readonly`, or in an example specifically for `readonly`. I don't think that it should be explained here, in the example for `dropzone`.

src/components/file/file.tsx
```diff
@@ -95,6 +97,18 @@ export class File {
     @Prop({ reflect: true })
     public accept: string = '*';
 
+    /**
+     * Set to `false` to hide the visual representation of the dropzone.
+     */
+    @Prop()
+    public visualizeDropzone: boolean = true;
```
adrianschmidt, line 104: Do we really need this option? Can't we at least wait to introduce it until someone asks for it?

src/translations/sv.ts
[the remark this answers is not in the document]
Kiarokh: indeed. Just made a fixup

src/components/chip-set/chip-set.tsx
```diff
@@ -232,6 +233,10 @@ export class ChipSet {
             'chip-set.clear-all',
             this.language
         );
+        this.dropZoneTip = translate.get(
```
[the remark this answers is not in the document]
Kiarokh: > I think it is possible to remove the constructor() by changing the binded functions to arrow function as well but it is not related to this PR.

That's right. We can perhaps do it later in another PR ;)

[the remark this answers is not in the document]
Kiarokh: > What if we refactor it like this, then we can remove componentWillLoad function.

I don't know. Do you want me to change it this way? Or should we also improve this in the PR that we change the bound functions to arrow functions?

src/components/file/file.tsx
Kiarokh: I think they already did when I demoed it. Example was when the component is used on a phone, they cannot drag and drop. So if someone wants to optionally hide this (like when they have a solution made for phones, or when god-knows-why) then they can do so programatically.

src/components/file/examples/file-no-dropzone.tsx
Kiarokh: Me neither. I think the concept comes with `readonly`. So I wouldn't explain it.

src/components/chip-set/chip-set.tsx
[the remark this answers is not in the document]
Befkadu1: Yes we can do that in this PR.

src/components/file/file.tsx
Kiarokh: it'll also prevent complains from people who don't like the change :P

Befkadu1 (dismissed):
It is fine for me. @adrianschmidt we will merge it if you are ok with the PR

src/components/file/file.tsx
adrianschmidt: Ah… ok… hmm… 🤔

I rather think it would be better to build in feature-detection for drag-and-drop and not show this if drag-and-drop isn't supported. But that's more complexity that I also think we shouldn't do unless there's explicit demand for it.

I still suggest that we remove this feature until it's explicitly asked for. It's easy to add later, but harder to remove if we realise there's a better way to do it.

Second opinion @jgroth?

adrianschmidt: Just "requesting changes" until we've decided this one (and I've looked over the css like Bef asked me to).

adrianschmidt (changes requested):

src/components/file/examples/file-accepted-types.tsx
```diff
@@ -1,8 +1,7 @@
 import { FileInfo } from '@limetech/lime-elements';
 import { Component, h, State } from '@stencil/core';
-
 /**
- * Limit accepted file types
+ * Limit accepted file types and extentions
```
adrianschmidt, line 4: We're not limiting the file extension, we're limiting the file type. It seems like at least macOS is deriving the file type from the file extension, so the effect on macOS is that the file extension is limited, but I'm not sure that's true for all systems.

src/components/file/file.tsx
jgroth: Yeah I think I agree, it should be quite easy to add this later if it is needed :slightly_smiling_face: I general we should try and be a bit restrictive of what we add to the public interface since it always easier to add stuff than change or remove them

src/components/chip-set/partial-styles/_file-picker.scss
```diff
@@ -0,0 +1,57 @@
...
+
+            &:before {
+                // dropzone tip
+                content: attr(dropzone-tip);
+
+                display: inline-flex;
+                align-items: center;
+                justify-content: flex-end;
+
+                left: 40%; // prevents overlapping of `label` and tip text
```
adrianschmidt, line 26: We might need to limit the max-width of the label too:

image

src/components/chip-set/partial-styles/_file-picker.scss
```diff
@@ -0,0 +1,57 @@
...
+                position: absolute;
+                margin: auto;
+                top: var(--dropzone-margin);
+                right: var(--dropzone-margin);
+                bottom: var(--dropzone-margin);
+            }
+
+            &:before {
+                // dropzone tip
+                content: attr(dropzone-tip);
```
adrianschmidt, line 20: Nice 😁

src/components/input-field/partial-styles/_readonly.scss
```diff
@@ -31,29 +31,32 @@
...
         }
 
-        .mdc-text-field__icon.trailing-icon {
-            display: flex;
-            justify-content: flex-end;
-            align-items: center;
-
-            pointer-events: auto;
-            cursor: pointer;
+        .mdc-text-field__icon {
```
adrianschmidt, line 34: > prevent leading icons becoming transparent in `readonly`

Hmm… this seems like something that ought to be fixed in chip-set rather than here? 🤔

src/components/chip-set/chip-set.tsx
```diff
@@ -698,11 +690,16 @@ export class ChipSet {
...
-                title={this.clearAllChipsLabel}
+                title={this.clearAllChipsLabel()}
                 aria-label={this.clearAllChipsLabel}
             />
         );
     }
+
+    private clearAllChipsLabel = (): string => {
+        return translate.get('chip-set.clear-all', this.language);
+    };
```
adrianschmidt, line 701: This change is very definitely not a `feat`. It's a `refactor`. There's no reason we should trigger the release of a new minor version because of this change, and the change should not be documented as a "new feature" in the changelog.

src/components/input-field/partial-styles/_readonly.scss
Kiarokh: I think it's fine here. Because we don't want the leading or trailing icons to become transparent and look disabled in `readonly` mode. It's also true in input field.

src/components/chip-set/chip-set.tsx
Kiarokh: That's right. This was a 🧠💨

src/components/file/examples/file-accepted-types.tsx
Kiarokh: OK I'll revert back

src/components/file/file.tsx
Kiarokh: OK. I'll remove it

src/components/input-field/partial-styles/_readonly.scss
adrianschmidt: Yeah, but if we fix it here, it doesn't apply to chip-set in readonly? If we fix it in chip-set, it should automatically apply in all components using chip-set.

src/components/chip-set/partial-styles/_file-picker.scss
Kiarokh: this commit f7ba8c7c207815b4b3843e5fc96ffb37b8ce9b61 (fix(chip-set): make long labels truncate correctly when there is leading icon) was needed to make this work. The fixup after takes care of label width

Kiarokh:
@adrianschmidt I'm done with your requested changes

src/components/file/file.scss
```diff
@@ -1,6 +1,7 @@
 /**
  * @prop --icon-background-color: Background color of the icon. Defaults to `--contrast-400`.
  * @prop --icon-color: Color of the icon. Defaults to `--contrast-1100`.
+ * @prop --file-picker-dropzone-distance: Distance between the file picker and the dropzone. Defaults to `1rem` which is the default gap in `limel-form`.
```
adrianschmidt, line 4: This doesn't seem to be used anywhere?

Kiarokh: Oh good catch. It's just trash, left from my first version of the dropzone, when dropzone was visually separated from the input field.

Kiarokh: i'll fixup

src/components/file/file.tsx
```diff
@@ -164,25 +175,48 @@ export class File {
                 disabled={this.disabled || this.readonly}
             />,
             <limel-chip-set
+                class={{
+                    'is-file-picker': true,
+                    'shows-dropzone': true,
+                    'highlight-dropzone': this.isDraggingOverDropZone,
```
adrianschmidt, line 181: I added some interactivity to the styles 🙂

https://user-images.githubusercontent.com/154725/126768512-9a88a5d2-71c3-4693-8625-36a1b74337ae.mov

adrianschmidt: There may be more cool things we can do with changing the pointer and stuff, but I don't have time before my vacation 😅

adrianschmidt (approved):
👏

src/components/file/file.tsx
Kiarokh: Lovely, so you basically wanna close this issue: https://github.com/Lundalogik/crm-feature/issues/2249 😄 in the same time

Kiarokh: Wh have [an issue](https://github.com/Lundalogik/crm-feature/issues/2249) for this. It can wait for perfection after the vacation ;)

adrianschmidt: Ah, yeah, I guess it would be really nice to have _all_ drop zones light up a bit when the user drags something anywhere on the page, and then still have each individual drop zone become even more highlighted when the user drags over it.

PR #1280, opened 2021-06-21
fix(flex-container): make action buttons slot work nicely in `limel-dialog`

Kiarokh:
buttons are located in the footer slot.
fix https://github.com/Lundalogik/crm-feature/issues/2232

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dialog/examples/dialog-closing-actions.tsx
+++ b/src/components/dialog/examples/dialog-closing-actions.tsx
@@ -49,12 +49,12 @@ export class DialogClosingActionsExample {
             <limel-button
                 primary={true}
                 label="Open"
-                onClick={this.triggerOnClick}
+                onClick={this.openDialog}
             />,
             <limel-dialog
                 open={this.isOpen}
                 closingActions={{ escapeKey: false, scrimClick: false }}
-                onClose={this.onClose}
+                onClose={this.closeDialog}
             >
                 <p>
                     This dialog doesn't close by clicking the scrim or pressing
@@ -62,22 +62,22 @@ export class DialogClosingActionsExample {
                 </p>
                 <limel-button
                     label="I understand"
-                    onClick={this.okOnClick}
+                    onClick={this.handleConfirmClick}
                     slot="button"
                 />
             </limel-dialog>,
         ];
     }
 
-    private triggerOnClick = () => {
+    private openDialog = () => {
         this.isOpen = true;
     };
 
-    private okOnClick = () => {
+    private handleConfirmClick = () => {
         this.isOpen = false;
     };
 
-    private onClose = () => {
+    private closeDialog = () => {
         this.isOpen = false;
     };
 }
--- a/src/components/dialog/examples/dialog-closing-actions.tsx
+++ b/src/components/dialog/examples/dialog-closing-actions.tsx
@@ -44,12 +44,6 @@ export class DialogClosingActionsExample {
     @State()
     private isOpen = false;
 
-    constructor() {
-        this.triggerOnClick = this.triggerOnClick.bind(this);
-        this.onClose = this.onClose.bind(this);
-        this.okOnClick = this.okOnClick.bind(this);
-    }
-
     public render() {
         return [
             <limel-button
@@ -75,15 +69,15 @@ export class DialogClosingActionsExample {
         ];
     }
 
-    private triggerOnClick() {
+    private triggerOnClick = () => {
         this.isOpen = true;
-    }
+    };
 
-    private okOnClick() {
+    private okOnClick = () => {
         this.isOpen = false;
-    }
+    };
 
-    private onClose() {
+    private onClose = () => {
         this.isOpen = false;
-    }
+    };
 }
--- a/src/components/dialog/examples/dialog-closing-actions.tsx
+++ b/src/components/dialog/examples/dialog-closing-actions.tsx
@@ -33,9 +33,11 @@ export class DialogClosingActionsExample {
                     This dialog doesn't close by clicking the scrim or pressing
                     the escape key. Only the Ok-button triggers a close event.
                 </p>
-                <limel-flex-container justify="end" slot="button">
-                    <limel-button label="Ok" onClick={this.okOnClick} />
-                </limel-flex-container>
+                <limel-button
+                    label="I understand"
+                    onClick={this.okOnClick}
+                    slot="button"
+                />
             </limel-dialog>,
         ];
     }
--- a/src/components/dialog/examples/dialog-closing-actions.tsx
+++ b/src/components/dialog/examples/dialog-closing-actions.tsx
@@ -2,7 +2,40 @@ import { Component, h, State } from '@stencil/core';
 
 /**
  * Custom closing actions
+ *
+ * Action buttons in dialogs can be used to add a clear visual indication for
+ * the sighted users to realize that the dialog can be closed by pressing
+ * a button as well.
+ * This may sometimes be considered an unnecessary usage of action buttons for
+ * sighted users. Because majority of them users know that clicking or tapping
+ * outside the dialog closes it.
+ *
+ * Such buttons are usually labeled ***OK***, ***Dismiss*** or ***Close***.
+ *
+ * :::tip
+ * When to use action buttons for simple "close" actions?
+ * - In fullscreen dialogs where clicking outside to close is hard.
+ * - When big dialogs are opened on phones, which make tapping outside hard for users.
+ * - When designing with accessibility in mind, and for those users who
+ * use screen readers to navigate the user interface.
+ * :::
+ *
+ * But sometimes, depending on the importance of the message which is displayed,
+ * you have to choose to display a close button, and disable other means of
+ * dismissing the dialog.
+ *
+ * :::tip
+ * When to use custom closing actions?
+ * - To make sure that the user really reads and understands the dialog's content.
+ * - To make sure that the user does not accidentally click outside and close the dialog.
+ * :::
+ *
+ * For such cases, avoid generic labels like ***OK***, or ***Close*** which unconsciously
+ * motivate users to dismiss the message; and instead use more purposeful labels
+ * such as ***I understand***, ***Looks good!***, ***Continue***, and similar;
+ * like in the example below.
  */
+
 @Component({
     tag: 'limel-example-dialog-closing-actions',
     shadow: true,
@@ -31,7 +64,7 @@ export class DialogClosingActionsExample {
             >
                 <p>
                     This dialog doesn't close by clicking the scrim or pressing
-                    the escape key. Only the Ok-button triggers a close event.
+                    the escape key. Only the button triggers a close event.
                 </p>
                 <limel-button
                     label="I understand"
--- a/src/components/dialog/examples/dialog-form.tsx
+++ b/src/components/dialog/examples/dialog-form.tsx
@@ -27,17 +27,6 @@ export class DialogFormExample {
     @State()
     private isConfirmationOpen = false;
 
-    constructor() {
-        this.openDialog = this.openDialog.bind(this);
-        this.closeDialog = this.closeDialog.bind(this);
-        this.onClosing = this.onClosing.bind(this);
-        this.nameOnChange = this.nameOnChange.bind(this);
-        this.ageOnChange = this.ageOnChange.bind(this);
-        this.closeConfirmation = this.closeConfirmation.bind(this);
-        this.onConfirmPositive = this.onConfirmPositive.bind(this);
-        this.onConfirmNegative = this.onConfirmNegative.bind(this);
-    }
-
     public render() {
         return [
             <limel-button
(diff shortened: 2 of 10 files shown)


src/components/dialog/examples/dialog-form.tsx
```diff
@@ -75,31 +75,27 @@ export class DialogFormExample {
...
-                <limel-flex-container slot="button" reverse={true}>
+                <limel-flex-container slot="button">
+                    <limel-button label="Cancel" onClick={this.closeDialog} />
                     <limel-button
                         primary={true}
                         label="Save"
                         disabled={!this.nameValid() || !this.ageValid()}
                         onClick={this.submitForm}
                     />
-                    <limel-button label="Cancel" onClick={this.closeDialog} />
```
adrianschmidt, line 86: > Having this was completely unnecessary from the beginning.
> Not sure why it was added in the example?
> Maybe to showcase what `flex-container` can do?!
> But consumers copy/paste this exact code
> and it potentially makes troubles for us.
> For example when trying to add fixes such as the previous commit.
> Which will cause the buttons to jump to the left side of the dialog,
> if consumer has copy/pasted `reverse={true}` as well.

The idea was that it would be easier for the users if a single tab-press brought them to the primary button, instead of having to tab past the cancel-button first.

It was probably misguided… 🤷‍♂️

src/components/dialog/examples/dialog-closing-actions.tsx
```diff
@@ -40,15 +34,15 @@ export class DialogClosingActionsExample {
         ];
     }
 
-    private triggerOnClick() {
+    private triggerOnClick = () => {
```
adrianschmidt, line 69: ❤️

I've been thinking for a while I should go through and make these changes. Thank you! 🙏

adrianschmidt: I'm not sure about the commit message though? "improve ui for config options in examples"?

Am I just being daft again? 😆

src/components/dialog/examples/dialog-form.tsx
Kiarokh: I see... I'm a bad example and I no clue about how people navigate with keyboard. But I think their expectation is to navigate from left to right. So I'd suggest we skip this, unless we have some usability/accessibility guidelines that says this is the expected behaviour.

src/components/dialog/examples/dialog-closing-actions.tsx
Kiarokh: 😄  I don't know dude. I just copied one of your own commit messages for an exact change in another component 😅 (don't remember  where it was, but I guess it was when we fixed `readonly` styles for the input field.

Kiarokh: but feel free to change. I have no clue what "improve ui for config options in examples" means 😄

src/components/flex-container/flex-container.scss
```diff
@@ -30,7 +28,6 @@
 :host(limel-flex-container[align='stretch']) {
     align-items: stretch;
 }
-
```
adrianschmidt, line 33: I've split this code formatting change into a separate commit from the functional change.

src/components/dialog/examples/dialog-form.tsx
adrianschmidt: Yeah, I agree. I don't have any guidelines for it, just an idea I had before I knew better 😆

Kill it with fire! 🙂

src/components/dialog/examples/dialog-closing-actions.tsx
adrianschmidt: Ah! Ok, I found the commit. The change that I was referring to there wasn't the change from `function` to `=>`, it was the fact that I had changed from using limel-switch and limel-button in a few different examples, to using limel-checkbox, for consistency…

I should have split that commit into two separate ones I guess 😄

adrianschmidt: I'll come up with some better commit message. Tomorrow. It's time to log off for today 😅

adrianschmidt: ⚡

src/components/flex-container/flex-container.scss
```diff
@@ -49,3 +46,14 @@
...
+// Since consumers often copy/paste from examples of `limel-dialog` from the
+// documentations, they often put their buttons in a `limel-flex-container`
+// which is used in the footer slot of those examples with buttons.
+// So to prevent unwanted layout problems of buttons having no distance, we use
+// below code to target `limel-flex-container`s in those slots.
+:host(limel-flex-container[slot='button']) {
+    gap: 0.5rem; // Makes sure buttons get some default distance
+    width: 100%;
+    justify-content: flex-end;
+}
```
adrianschmidt, line 59: Hmm… does this really have to be added in `flex-container.scss`? Since it's added on `:host`, I think it could be added by the dialog's css, and I wonder if that wouldn't make more sense?

Also, I've been thinking about whether we really need to specify that it should be limel-flex-container at all… 🤔 Maybe we could just say in the dialog css, that whatever element has the `slot="button"` attribute should have `display: flex` and these additional styles?

adrianschmidt, line 59: @jgroth I dug through the history to find where limel-flex-container was added, and it was added way back in [v17.0.0](https://github.com/Lundalogik/lime-elements/releases/tag/v17.0.0), without reference to any issue, or anything else really explaining what the purpose was. It was added in a commit by you, so I'm wondering if you remember any details? 🙂

src/components/dialog/examples/dialog-closing-actions.tsx
```diff
@@ -2,7 +2,36 @@ import { Component, h, State } from '@stencil/core';
...
+ * sighted users. Because majority of them users know that clicking or tapping
+ * outside the dialog closes it.
+ *
+ * Such buttons usually are labeled as `OK`, `Dismiss` or `Close`.
+ * :::tip
+ * When to use action buttons for simple "close" actions?
+ * - In fullscreen dialogs where clicking outside to close is hard.
+ * - When big dialogs are opened on phones, which make tapping outside hard for users.
+ * - When designing with accessibility in mind, and for those users who
+ * use screen readers to navigate the user interface.
```
adrianschmidt, line 19: For accessibility, it feels like there should always be a close button, but it could be visually hidden… but finding the best solution for that is perhaps more work than we're willing to put in right now?

src/components/flex-container/flex-container.scss
jgroth: Haha no sorry I don't remember :smile: I think we just wanted some element for the consultants that they could use if they did not know a lot of CSS

adrianschmidt: Yeah… and they probably like it! But I don't… 😆

Or, well, I'm just worried it will be causing us trouble… I feel like we should at least try not to build ourselves further into it by making it more or less required in dialog footers… 🤔

src/components/dialog/examples/dialog-closing-actions.tsx
```diff
@@ -2,7 +2,36 @@ import { Component, h, State } from '@stencil/core';
...
  * Custom closing actions
+ *
+ * Action buttons in dialogs can be used to add a clear visual indication for
+ * the sighted users to realize that the dialog can be closed by pressing
+ * a button as well.
+ * This may sometimes be considered an unnecessary usage of action buttons for
+ * sighted users. Because majority of them users know that clicking or tapping
+ * outside the dialog closes it.
+ *
+ * Such buttons usually are labeled as `OK`, `Dismiss` or `Close`.
```
adrianschmidt, line 13: Backticks formats the text as code. That _looks_ good in this case, but I don't think it's semantically good… I think quote marks, cursive text, and/or bold text would serve the purpose better.

```suggestion
 * Such buttons are usually labeled ***OK***, ***Dismiss*** or ***Close***.
```

src/components/dialog/examples/dialog-closing-actions.tsx
```diff
@@ -2,7 +2,36 @@ import { Component, h, State } from '@stencil/core';
...
+ * dismissing the dialog.
+ * :::tip
+ * When to use custom closing actions?
+ * - To make sure that the user really reads and understands the dialog's content.
+ * - To make sure that the user does not accidentally click outside and close the dialog.
+ * :::
+ * For such cases, avoid generic labels like `ok`, or `close` which unconciously
+ * motivate users to dismiss the message; and instead use more purposeful labels
+ * such as `I understand`, `Looks good!`, `Continue`, and similar;
+ * like in the example below.
```
adrianschmidt, line 32:
```suggestion
 * For such cases, avoid generic labels like ***OK***, or ***Close*** which unconsciously
 * motivate users to dismiss the message; and instead use more purposeful labels
 * such as ***I understand***, ***Looks good!***, ***Continue***, and similar;
 * like in the example below.
```

src/components/dialog/examples/dialog-closing-actions.tsx
```diff
@@ -34,21 +57,24 @@ export class DialogClosingActionsExample {
                     the escape key. Only the Ok-button triggers a close event.
```
adrianschmidt, line 57: This still refers to the "Ok-button", but the button has been relabelled.

src/components/dialog/examples/dialog-closing-actions.tsx
```diff
@@ -2,7 +2,36 @@ import { Component, h, State } from '@stencil/core';
 
 /**
  * Custom closing actions
+ *
+ * Action buttons in dialogs can be used to add a clear visual indication for
+ * the sighted users to realize that the dialog can be closed by pressing
+ * a button as well.
+ * This may sometimes be considered an unnecessary usage of action buttons for
+ * sighted users. Because majority of them users know that clicking or tapping
+ * outside the dialog closes it.
```
adrianschmidt, line 11: I think this section is a little incoherent

> … can be closed by pressing a button as well.

"As well" as what? It sounds like I'm meant to have read some missing section before this.

The whole paragraph seems to assume that the reader, and all users, see clicking outside a dialog as the _primary_ way to close it, and that having a close-button is some new invention, the purpose of which needs explanation. At least to my ears, that makes this paragraph sound quite weird.

src/components/flex-container/flex-container.scss
adrianschmidt: Hmm… no… it doesn't seem like it can be added from the dialog's styling… 😞

But I did find another solution, which I've added as a separate commit.

src/components/dialog/examples/dialog-closing-actions.tsx
adrianschmidt: ⚡

src/components/dialog/examples/dialog-closing-actions.tsx
adrianschmidt: @Kiarokh I'd like this paragraph to get a touch-up, but apart from that I'm ready to approve.

adrianschmidt (changes requested):

src/components/dialog/examples/dialog-closing-actions.tsx
Kiarokh: I think so… but that's a good idea to add an invisible button

src/components/dialog/examples/dialog-closing-actions.tsx
Kiarokh: Good point! What about this: 

to realize that the dialog can be closed by pressing a button as well as the conventional closing way, which is clicking outside the dialog.

adrianschmidt: Yeah, that's better at least. It's still a little weird to me to consider clicking outside the dialog the _main_ way of dismissing it, but sure 👍

adrianschmidt: Will you add a fixup @Kiarokh? I'd love to get this merged 🙂

adrianschmidt: I'm going to approve and merge this. You can always update the text later 🙂

adrianschmidt (dismissed):

src/components/flex-container/flex-container.scss
```diff
@@ -49,3 +46,17 @@
...
+:host(limel-flex-container[slot='button'][direction='horizontal'][reverse]) {
+    justify-content: flex-start;
```
adrianschmidt, line 61: I added this so that we don't "break" existing solutions that use `reverse={true}`.

Now:
image

Before:
image

adrianschmidt (approved):

src/components/flex-container/flex-container.scss
Kiarokh: smart

src/components/flex-container/flex-container.scss
Kiarokh: I really like your solution. Thanks for tweaking my commits!

PR #1281, opened 2021-06-21
progress-flow component

TommyLindh2:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/progress-flow/examples/progress-flow-alternative-ui.tsx
+++ b/src/components/progress-flow/examples/progress-flow-alternative-ui.tsx
@@ -1,0 +1,87 @@
+import { Component, h, State } from '@stencil/core';
+import { FlowItem } from '../progress-flow.types';
+
+/**
+ * Alternative UI
+ *
+ * You can render the component with an alternative layout which may be good for
+ * some use cases. To achive this, simply add the `has-detached-steps` to the
+ * component.
+ *
+ * Note that the `is-narrow` class does not affect this UI alternative.
+ *
+ */
+@Component({
+    tag: 'limel-example-progress-flow-alternative-ui',
+    shadow: true,
+})
+export class ProgressFlowAlternativeUiExample {
+    @State()
+    private flowItems: FlowItem[] = [
+        {
+            value: 'contact',
+            text: 'Customer contact',
+            activeColor: 'rgb(var(--color-orange-default))',
+            selected: true,
+            icon: 'meeting',
+            secondaryText: 'Via phone support',
+        },
+        {
+            value: 'requirement',
+            text: 'Demand analysis',
+            activeColor: 'rgb(var(--color-sky-default))',
+            icon: 'combo_chart',
+        },
+        {
+            value: 'tender',
+            text: 'Quote',
+            activeColor: 'rgb(var(--color-green-default))',
+            icon: 'paper_plane',
+        },
+        {
+            value: 'agreement',
+            text: 'Agreement',
+            activeColor: 'rgb(var(--color-teal-default))',
+            icon: 'agreement',
+        },
+        {
+            value: 'rejected',
+            text: 'Rejected',
+            isOffProgress: true,
+            activeColor: 'rgb(var(--color-red-dark))',
+            icon: 'do_not_disturb',
+            iconColor: 'rgb(var(--color-red-dark))',
+        },
+        {
+            value: 'onhold',
+            text: 'On hold',
+            isOffProgress: true,
+            activeColor: 'rgb(var(--color-coral-default))',
+            icon: 'circled_pause',
+            iconColor: 'rgb(var(--color-coral-default))',
+        },
+    ];
+
+    constructor() {
+        this.onChange = this.onChange.bind(this);
+    }
+
+    public render() {
+        return (
+            <limel-progress-flow
+                flowItems={this.flowItems}
+                onChange={this.onChange}
+                class="has-detached-steps"
+            />
+        );
+    }
+
+    private onChange(event: CustomEvent<FlowItem>) {
+        this.flowItems = this.flowItems.map((item) => {
+            return {
+                ...item,
+                selected: item.value === event.detail?.value,
+            };
+        });
+    }
+}
--- a/src/components/progress-flow/progress-flow-item/progress-flow-item.scss
+++ b/src/components/progress-flow/progress-flow-item/progress-flow-item.scss
@@ -1,0 +1,171 @@
+@import '../../../style/mixins';
+@import '../../../style/internal/variables';
+
+@import './partial-styles/_active-indicator';
+@import './partial-styles/_colors';
+@import './partial-styles/_has-detached-steps';
+
+.flow-item {
+    --step-background--inactive: var(
+        --progress-flow-step-background-color--inactive,
+        rgb(var(--contrast-600))
+    );
+    --step-background--active: var(
+        --progress-flow-step-background-color--active,
+        var(--mdc-theme-primary)
+    );
+    --step-background--passed: var(
+        --progress-flow-step-background-color--passed,
+        var(--step-background--active)
+    );
+
+    --step-text--inactive: var(
+        --progress-flow-step-text-color--inactive,
+        rgb(var(--contrast-1200))
+    );
+    --step-text--active: var(
+        --progress-flow-step-text-color--active,
+        var(--mdc-theme-on-primary)
+    );
+    --step-text--passed: var(
+        --progress-flow-step-text-color--passed,
+        var(--mdc-theme-on-primary)
+    );
+
+    --step-divider-color: var(
+        --progress-flow-step-divider-color,
+        rgb(var(--contrast-100))
+    );
+
+    position: relative;
+    width: 100%;
+
+    display: flex;
+    flex-direction: column;
+    align-items: stretch;
+
+    &:not(.off-progress-item, .last) {
+        .divider {
+            &:after {
+                content: '';
+            }
+        }
+    }
+}
+
+.step {
+    transition: background-color 0.2s ease, box-shadow 0.2s ease;
+
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    position: relative;
+    width: 100%;
+    height: var(--step-height);
+    border: none;
+    font-size: pxToRem(14);
+
+    &.disabled {
+        opacity: 0.75;
+        cursor: not-allowed;
+
+        &.readonly {
+            opacity: 1;
+            cursor: default;
+        }
+    }
+
+    &:focus {
+        outline: none;
+    }
+
+    &:focus-visible {
+        box-shadow: var(--shadow-depth-8-focused);
+    }
+
+    .flow-item:not(.off-progress-item, .first) & {
+        padding-left: calc(var(--step-height) / 2);
+    }
+
+    .flow-item:not(.active) & {
+        &:not(.disabled) {
+            cursor: pointer;
+            &:hover {
+                box-shadow: var(--button-shadow-normal);
+            }
+            &:active {
+                box-shadow: var(--button-shadow-pressed);
+            }
+        }
+    }
+
+    .off-progress-item & {
+        border-radius: pxToRem(4);
+        padding: 0 pxToRem(12);
+
+        .icon {
+            margin-left: 0;
+        }
+    }
+    .flow-item.first & {
+        border-top-left-radius: var(--step-height);
+        border-bottom-left-radius: var(--step-height);
+        padding-left: pxToRem(20);
+    }
+    .flow-item.last & {
+        border-top-right-radius: var(--step-height);
+        border-bottom-right-radius: var(--step-height);
+        padding-right: pxToRem(20);
+    }
+}
+
+.divider {
+    height: var(--step-height);
+    overflow-y: clip;
+    position: absolute;
+    right: 0;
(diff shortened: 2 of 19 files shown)


src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
```diff
@@ -0,0 +1,105 @@
...
+    Prop,
+} from '@stencil/core';
+import { FlowItem } from '../progress-flow.types';
+
+/**
+ * @private
+ */
+@Component({
+    tag: 'limel-progress-flow-item',
+    shadow: false,
```
jgroth, line 16: What is the reason for having this set to `false`? :slightly_smiling_face:

src/components/progress-flow/progress-flow.e2e.ts
```diff
@@ -0,0 +1,7 @@
+import { ProgressFlow } from './progress-flow';
+
+describe('limel-progress-flow', () => {
+    it('builds', () => {
+        expect(new ProgressFlow()).toBeTruthy();
```
jgroth, line 5: This test does not add much value, we should probably add some proper tests :smile: (same can be said for the flow-item component)

src/components/progress-flow/progress-flow.types.ts
```diff
@@ -0,0 +1,8 @@
+import { ListItem } from '@limetech/lime-elements';
+
+export interface FlowItem extends ListItem {
+    isOffProgress?: boolean;
+    activeColor?: string;
+    passedColor?: string;
+    iconColor?: string;
```
jgroth, line 7: These should be better documented since they are part of the public interface :slightly_smiling_face:

jgroth (commented):
Good work! I have not done a full review of this, just had a quick look and found a few issues/questions :slightly_smiling_face:

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
```diff
@@ -0,0 +1,105 @@
...
+            >
+                {this.renderIcon()}
+                <span class="text">{this.item.text}</span>
+                <div class="divider" />
+            </button>,
+            this.renderSecondaryText(),
+        ];
+    }
+
+    private handleClick() {
```
maltja, line 86: We could use an arrow function here instead like you did in handleFlowItemClick in the limel-progress-flow component. Then you don't have to bind it in the constructor. 🙂

maltja (changes requested):
Looks good, really nice component! 🎉  I just have some small comments and a question regarding the alternative UI.
When should this UI be used instead of the other standard UI, can you give an example?
I am also thinking about the visualisation of the off-progress steps. Why is there a line between rejected and on hold? Is rejected a progress step to on hold (then this is not visualised in the standard UI but maybe that is the main difference between them.)?
Screenshot 2021-06-30 at 14 00 18

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
Kiarokh: We didn't want to have the styles such as `active` and so forth, inside the child component. This made it easier to handle those UI design things in the same place, in the parent component. There was also a CSS issue regarding the order of steps. If shadow was true, then we had to use `reverse` in our flex container, which made a stupid tabbing effect. Also created some z-index rendering for the pseudo elements that create the arrows. So if this is not a critical issue, please don't change it.

src/components/progress-flow/progress-flow.e2e.ts
Kiarokh: @TommyLindh2 said he wasn't so fluent in writing such tests, and hoped that you guys can help with this part. So don't wait for him to fix this, :) either remove the tests, or make them nicely yourselves ;)

src/components/progress-flow/progress-flow.types.ts
Kiarokh: Sure, let's add some.

Kiarokh:
```suggestion
	/**
     * determines whether a step should be a part of the flow,
     * or should be an independent step visually located on the side. 
     */
    isOffProgress?: boolean;
    
    /**
     * Background color of active step.
     */
    activeColor?: string;
    
    /**
     * Background color of the step, when it is passed or done.
     */
    passedColor?: string;
    
    /**
     * Fill color of the icon on the step, when it is inactive.
     */
    iconColor?: string;
```

Kiarokh: How about these?

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
```diff
@@ -0,0 +1,105 @@
...
+})
+export class ProgressFlowItem {
+    @Element()
+    public element: HTMLLimelProgressFlowItemElement;
+
+    /**
+     * What flow item that should be rendered
+     */
+    @Prop()
+    public item: FlowItem = null;
```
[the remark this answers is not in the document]
Kiarokh: Good point. Can you suggest a fix, since Tommy isn't here to do anything about it.

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
```diff
@@ -0,0 +1,105 @@
...
+        this.handleClick = this.handleClick.bind(this);
+    }
+
+    public render() {
+        let secondaryText = this.item.secondaryText;
+        if (secondaryText) {
+            secondaryText = ' · ' + secondaryText;
+        }
+
+        const tooltip = this.item.text + secondaryText;
```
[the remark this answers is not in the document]
Kiarokh: The `secondaryText` is added to the tooltip, because both are used as a `title` attribute, which is displayed on `hover`. The reason that both are there is because the steps have a `max-width` to prevent them from growing too large. Also make better layouts on smaller containers. However, if text is too long, we can at least display it on hover, to enable users read it fully. If the `secondaryText` is optional, would it still say undefined? 

How do we solve this? Can you please suggest a code change?

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
Kiarokh: Awesome, please add a fixup

src/components/progress-flow/progress-flow.tsx
```diff
@@ -0,0 +1,170 @@
...
+    public readonly = false;
+
+    /**
+     * Fired when a new value has been selected from the progress flow
+     */
+    @Event()
+    public change: EventEmitter<FlowItem>;
+
+    constructor() {
+        this.handleFlowItemClick = this.handleFlowItemClick.bind(this);
```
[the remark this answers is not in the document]
Kiarokh: awesome, a fixup? :) shall I add or you will?

Kiarokh:
I also think the lint commit e63b6a7c23456d68955eb33fade3993512b779c4 should just be merged with the `feat` commit, since it's an entirely new feature.

Kiarokh:
fixed this problem here 068937ecc1fe7a603e89bddc9f5b4c8e9b4aed6b
image

Kiarokh:
> alternative UI. When should this UI be used instead of the other standard UI, can you give an example?

Good question. I've updated the example file and added further info about when to use the alternative UI

Kiarokh:
Safari's lack of support for `overflow-y: clip` renders the steps this way:
image

adrianschmidt:
> Safari's lack of support for `overflow-y: clip` renders the steps this way:
> image

While it looks a little bit odd, I don't think it looks horrible. I don't think it's a merge-blocker. (Unless I'm missing something? It's just the top and bottom corner of the squares that should be cut, but aren't, right?)

Kiarokh:
I am almost done with the css changes. It might be good to add a fixup. What do you think?

Kiarokh:
OK. Above is fixed and pushed. But there is a known issue which has always existed:

image


Depending on screen width and length of text, lines between steps might get rendered detached. As of today I haven't investigated why and how to fix. But we can do it later I feel

adrianschmidt:
> OK. Above is fixed and pushed. They are now rendered like this. For no apparent reason Safari adds some visual defects. But much better than before.
> image

Great!
 
> But there is a known issue which has always existed:
> 
> image
> 
> Depending on screen width and length of text, lines between steps might get rendered detached. As of today I haven't investigated why and how to fix. But we can do it later I feel

Yeah, I agree. I don't know if it becomes a bigger problem on very large screens, but in your screenshot I think it's still perfectly clear what the intention is.

Kiarokh:
Awesome. I suggest merging it since this component has been quite brutally reviewed several times now :) Also it enables great things to happen for consultants and packers

adrianschmidt:
> Looks good, really nice component! 🎉 I just have some small comments and a question regarding the alternative UI.
> When should this UI be used instead of the other standard UI, can you give an example?
> I am also thinking about the visualisation of the off-progress steps. Why is there a line between rejected and on hold? Is rejected a progress step to on hold (then this is not visualised in the standard UI but maybe that is the main difference between them.)?
> Screenshot 2021-06-30 at 14 00 18

Is this alternative UI actually used anywhere currently? I strongly suggest removing it if it's not absolutely necessary. I _promise_ keeping two separate stylings will cause us trouble in the future.

maltja:
This is how it looks today and it is also explained when it should be used. What do you think about it now @adrianschmidt? 🙂 
Screenshot 2021-08-17 at 10 55 04

src/components/progress-flow/examples/progress-flow-alternative-ui.tsx
```diff
@@ -0,0 +1,99 @@
+import { Component, h, State } from '@stencil/core';
+import { FlowItem } from '../progress-flow.types';
+
+/**
+ * Alternative UI
```
adrianschmidt, line 5: I'm bringing this into a thread 😄

> This is how it looks today and it is also explained when it should be used. What do you think about it now @adrianschmidt? 🙂
> Screenshot 2021-08-17 at 10 55 04

Ah! Ok, yeah, if it is functionally different, then I think including it is reasonable.

I don't understand why the helper class is called `has-detached-steps` though? "Detached steps" sounds like the same thing as "off-progress steps" to me.

Also, I think we should consider making this a property instead of a helper-class, and force the component to be read-only when that property is used. Otherwise this isn't an "alternative UI", but an "alternative look", and people will use it as that…

@Kiarokh

src/components/progress-flow/progress-flow-item/progress-flow-item.scss
```diff
@@ -0,0 +1,178 @@
...
+        var(--mdc-theme-on-primary)
+    );
+
+    --step-divider-color: var(
+        --progress-flow-step-divider-color,
+        rgb(var(--contrast-100))
+    );
+
+    position: relative;
+    width: 100%;
```
adrianschmidt, line 39: Perhaps not a merge-blocker, but this is annoying:

The component always scrolls a bit when the last step is selected 😞

https://user-images.githubusercontent.com/154725/129697868-75f9f5c8-be3c-4f0e-845f-553777de69cd.mov

src/components/progress-flow/examples/progress-flow-alternative-ui.tsx
Kiarokh: > I'm bringing this into a thread 😄
Awesome! 😄 

> helper class is called `has-detached-steps` though?

You're right. It's perhaps not the best name. However, something like "alternative-design" or "alternative-ui" isn't neither. In case we add a new alternative ui for this component in future, then we should call that "alternative-ui-2" or something 😅

Open for suggestions. I prefer some name that describes the way it looks visually. 


> we should consider making this a property instead of a helper-class,

I think it's a good idea to have properties actually, rather than helper classes. We can be more consistent about this for all components. Today, we sometimes use helper classes for things that change the visual look and feel, and sometimes use properties which I think is very inconsistent.

So I think it's up to us to make a convention and follow it in all components for simplicity. If you want we can start here and make both `is-narrow` and `has-detached-steps` (or whatever we call it) properties.

I think it may be more natural for consumers to use properties. What do you suggest?

src/components/progress-flow/progress-flow-item/progress-flow-item.scss
Kiarokh: Good catch. I'll have to fix this. I might have introduced this when I fixed the last issues with Safari. 

Just to explain why it happens.

The idea is that in small screen, users usually need to see previous, current, and next steps. Since the screen is small we don't have space for everything. Therefore, the component tries to center-align the selected step.

This also helps to see the "current" status, when a step is pre-selected and user loads the page. For example in ticket-handling scenarios. Always the selected/current step will be in center.

However, sometimes (like in your screen recording) there is an invisible edge, which is added by the `divider` on the last step, that causes this.
https://user-images.githubusercontent.com/35954987/129727810-7cf31f1d-2419-42ea-9cfc-8e7bc337a55d.mov

Kiarokh: ⚡️

PR #1283, opened 2021-06-22
fix(dialog): replace `heading` with `limel-header`

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/1908

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dialog/dialog.scss
+++ b/src/components/dialog/dialog.scss
@@ -20,6 +20,11 @@ slot[name='header'] {
 
 :host {
     --dialog-background-color: rgb(var(--contrast-100));
+    --header-heading-color: var(--dialog-heading-title-color);
+    --header-subheading-color: var(--dialog-heading-subtitle-color);
+    --header-supporting-text-color: var(--dialog-heading-supporting-text-color);
+    --header-icon-color: var(--dialog-heading-icon-color);
+    --header-icon-background-color: var(--dialog-heading-icon-background-color);
 }
 
 .mdc-dialog {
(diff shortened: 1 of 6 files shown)


maltja:
Looks good! However, I just tested it in the webclient and the create new dialog icon's color changed when using this version. Can we fix that? 
After:
Screenshot 2021-06-29 at 11 24 47
Before:
Screenshot 2021-06-29 at 11 33 13

Kiarokh:
thanks @maltja I'll check why it happens

Kiarokh:
I think this is because I removed the `badge`, and in the old dialog, if you had a `badge={true}` it'd get a white color as a fallbak (`$mdc-theme-on-secondary`)

’’’’scss
        limel-icon {
            margin-right: pxToRem(12);
            color: var(--dialog-heading-icon-color, #{$mdc-theme-secondary});

            &[badge] {
                background-color: var(
                    --dialog-heading-icon-background-color,
                    #{$mdc-theme-secondary}
                );
                color: var(
                    --dialog-heading-icon-color,
                    #{$mdc-theme-on-secondary}
                );
            }
        }
’’’’

I'll have to think whether we should solve this here, or tweak something in CRM components instead

src/components/dialog/dialog.scss
```diff
@@ -20,6 +20,11 @@ slot[name='header'] {
 
 :host {
     --dialog-background-color: rgb(var(--contrast-100));
+    --header-heading-color: var(--dialog-heading-title-color);
+    --header-subheading-color: var(--dialog-heading-subtitle-color);
+    --header-supporting-text-color: var(--dialog-heading-supporting-text-color);
+    --header-icon-color: var(--dialog-heading-icon-color);
+    --header-icon-background-color: var(--dialog-heading-icon-background-color);
```
adrianschmidt, line 27: Just copying the posts here, so we can have a threaded discussion 🙂

adrianschmidt: @maltja said:

> Looks good! However, I just tested it in the webclient and the create new dialog icon's color changed when using this version. Can we fix that? 
After:
Screenshot 2021-06-29 at 11 24 47
Before:
Screenshot 2021-06-29 at 11 33 13

adrianschmidt: @Kiarokh said:

> I think this is because I removed the `badge`, and in the old dialog, if you had a `badge={true}` it'd get a white color as a fallbak (`$mdc-theme-on-secondary`)
>
> ```scss
>        limel-icon {
>            margin-right: pxToRem(12);
>            color: var(--dialog-heading-icon-color, #{$mdc-theme-secondary});
>
>            &[badge] {
>                background-color: var(
>                    --dialog-heading-icon-background-color,
>                    #{$mdc-theme-secondary}
>                );
>                color: var(
>                    --dialog-heading-icon-color,
>                    #{$mdc-theme-on-secondary}
>                );
>            }
>        }
> ```
>
> I'll have to think whether we should solve this here, or tweak something in CRM components instead

adrianschmidt: I assume this is the default look of the header if no colours are set by the consumer?

image

I think we probably shouldn't change the icon colour here. We should set that in the same place as we set the background colour, which I assume is in crm-components?

Kiarokh: After consultation with Johan, I deprecated the `badge` in my PR here 9bf5aff2949bb29d3d4a59dd1dc8f2cce5f72d78 

My reason was because the absence of the badge (which also adds some distance around the icon), made the UI bad in the old header. The heading text would get too close to the icon. In the new `limel-header`, there is no such prop as badge, and instead the icon has a default transparent background. But the background can be specified with a color, and that doesn't change the layout and gaps between the elements in the header.

> I think we probably shouldn't change the icon colour here. We should set that in the same place as we set the background colour, which I assume is in crm-components?

I think it's a good idea to specify the icon color in `crm-components`, where we get colors in `limetype.ts`. However, I couldn't figure out how. If we don't do it, all icon colors will fall back to some dark grey, which is fine. I don't consider it as a "breaking change", but it may be visually disturbing in some cases (like company, which gives a poor contrast between the blue background and dark grey icon). 
image
But the proper visual-design decisions are expected to be made where the component is used. Which means we should fix them in crm-components. 

In the older `dialog`, if a customer has a custom object with a custom color which is very light, like the example below👇 they would certainly not just relied on the fall-back color of the icon which is white. They must have specified a color with a better contrast. So it's not a breaking change for them.
image

Summary is that I think we should merge this, and add the color in CRM components. How we add it there in the best way can be discussed.

maltja (dismissed):
Approved! 🙂

Kiarokh:
nice. I made an issue for the crm-components: https://github.com/Lundalogik/crm-feature/issues/2247

maltja (approved):

PR #1284, opened 2021-06-23
feat(file): render file chips with proper link to download file

adrianschmidt:
fix: Lundalogik/crm-feature#2231

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/chip-set/chip.types.ts
```diff
@@ -49,4 +49,10 @@ export interface Chip<T = any> {
      * The value of the badge. Only valid for `filter`.
      */
     badge?: number;
+
+    /**
+     * If supplied, the chip will render an `<a>` element, using the supplied
```
jensgustafsson, line 54: Isn't it always rendering an `<a>` element in this implementation?

adrianschmidt: Yes, it is actually… I wrote this before I implemented it, and it turned out that rendering an `<a>` without a `href` was much easier that switching between rendering an `<a>` or a `<div>`, and it had the same result…

I'll update the description 👍

adrianschmidt: ⚡

jensgustafsson (approved):
Nice!

I've tested it on iphone Safari and Chrome and on Mac OS Safari and Chrome

jensgustafsson (approved):

adrianschmidt:
Note to self: Note sure we tested this with keyboard-navigation, so do that and fix if necessary…

PR #1287, opened 2021-06-29
feat(file): display icon of most common filetypes on chip

Kiarokh:
fix: https://github.com/Lundalogik/hack-tuesday/issues/187
fix: https://github.com/Lundalogik/crm-feature/issues/2088

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip.types.ts
+++ b/src/components/chip-set/chip.types.ts
@@ -20,6 +20,11 @@ export interface Chip<T = any> {
     iconFillColor?: string;
 
     /**
+     * `title` attribute of the icon
+     */
+    iconTitle?: string;
+
+    /**
      * Background color of the icon. Overrides `--icon-background-color`.
      */
     iconBackgroundColor?: string;
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -637,6 +637,7 @@ export class ChipSet {
                 style={style}
                 size="small"
                 badge={true}
+                title={chip.iconTitle}
             />
         );
     }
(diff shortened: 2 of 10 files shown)


src/components/chip-set/chip.types.ts
```diff
@@ -25,6 +25,11 @@ export interface Chip<T = any> {
      */
     iconFillColor?: string;
 
+    /**
+     * `title` attribute of the icon
+     */
+    iconTitle?: string;
+
```
jgroth, line 32: It's probably time for us to add an `Icon` interface soon. A lot of the other components use the icon component and it creates a lot of duplications if all other components/interfaces has to add all these `iconXxx` properties. Instead we could have e.g. 
```ts
export interface Chip {
    icon?: string | Icon;
}

export interface Icon {
    name: string;
    color?: string;
    title?: string;
}
```
and reuse the `Icon` interface in the chip-set, menu, header component etc

It's nothing that has to be done in this PR, just making a general note about it :slightly_smiling_face:

Befkadu1: It is ok, can be fixed in this PR

jgroth: It's a bigger issue though and not related so it's probably better to separate it

Kiarokh: yes, let make an issue for icon interface so we don't forget; and fix it separately ;)

maltja (approved):
Nice work! 🎉

jensgustafsson (approved):

Issue #1247, opened 2021-05-19
Month picker doesn't change year when using <> buttons


anderssonjohan:
## Current behavior

Screen recording of: https://lundalogik.github.io/lime-elements/versions/31.7.2/#/component/limel-date-picker/

https://user-images.githubusercontent.com/435885/118808229-90028480-b8a9-11eb-832e-c8cfa8dc2d46.mov

1. Click january
2. Click ◀️  once
3. Picker changed year
4. Click february
5. Click ◀️  once
6. Picker doesn't indicate that anything have changed
7. Click ◀️  once more
8. Now picker changed the year because it takes two clicks to change year from Feb

## Expected behavior

* The year should change when clicking the ◀️ ▶️ buttons, like it does with the quarter picker

## Environment

- **lime-elements** version: 31.7.2 
- Framework used: 
- Logs:

adrianschmidt:
I agree that the issue belongs in lime-elements, but this doesn't describe the bug correctly. The bug is that the arrows are changing "displayed month" rather than "displayed year".

The selected month should _not_ change just because the "displayed year" is changed. When you select February 2021, and switch to 2020, February shouldn't be highlighted, because you didn't select February 2020, you selected February 2021. To select February 2020 instead, you need to click on it. This is the same as how the date-picker works when changing either month or year.

anderssonjohan:
@adrianschmidt That's what I thought these should do and I found it very strange that they changed the month. I can now see that the quarter picker actually works as you describe and I updated the issue to reflect that as the desired behavior.

What I find very strange is that the user may change year with up and down as well... ◀️ 🔽 🔼 ▶️ That made me unsure if it was the bug that I first described or not. The up/down buttons already works as with the quarter picker, but they're somewhat a hidden feature unless you hover them.

![Screen Shot 2021-05-19 at 18 57 56](https://user-images.githubusercontent.com/435885/118854140-d15c5980-b8d4-11eb-98de-a7ae6090ffdd.png)

adrianschmidt:
Oh, yeah, it's up/down-arrows, not a select. I've updated [my comment on Lundalogik/crm-feature#1964](https://github.com/Lundalogik/crm-feature/issues/1964#issuecomment-844062808) accordingly.

[...]

PR #1288, opened 2021-07-01
fix(date-picker): be able to change year with arrow buttons in month/quarter picker

maltja:
The problem was that handleOpen never was called which is where the event listeners are added. I guess this is because we only change the visibility of the portal which means that the picker never really opens and closes. However it closes if we select a date and therefore our custom event listeners are removed. 

Now we instead add the event listeners in init (called when the flatpicker is created) and remove them on destroy. 

fix Lundalogik/lime-elements#1247

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


hannahu (changes requested):
If you've chosen a date, it's not possible to change the year anymore.

maltja:
> If you've chosen a date, it's not possible to change the year anymore.

Nice catch! I copied those lines from the quarter picker where I thought the arrows worked. So we have the same bug for the quarter picker then. I will see if I can fix it for both the month and the quarter picker 🙂

maltja:
> If you've chosen a date, it's not possible to change the year anymore.

fixed! 🙂

hannahu (approved):
Looks good! We still have some weird thing that a date gets some orange mark in the picker when changing year but that's another issue and not something critical :P

PR #1289, opened 2021-07-01
chore(design-guidelines): move examples & `.md` files to a dedicated folder

Kiarokh:
Previously, they were inside the `components` folder,
which is not really where they should be.
So I made a new folder dedicated for design guidelines examples.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


adrianschmidt (approved):
I changed the type from `docs` to `chore`, because it's not of interest to a user that some files moved around, since the docs are still found in the same place.

PR #1290, opened 2021-07-01
docs(action-buttons): add guidelines regarding action button design conventions

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/2243

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/guides.ts
+++ b/guides.ts
@@ -11,6 +11,7 @@ export default [
             'src/design-guidelines/color-system/color-system.md',
             'src/design-guidelines/shadows/shadows-and-depth.md',
             'src/design-guidelines/size/size-rhythms.md',
+            'src/design-guidelines/action-buttons/action-buttons.md',
             'src/design-guidelines/decluttering.md',
         ],
     },
--- a/src/design-guidelines/action-buttons/action-buttons.md
+++ b/src/design-guidelines/action-buttons/action-buttons.md
@@ -1,0 +1,126 @@
+# Action buttons
+
+Action buttons are typically used in dialog footers, or in headers. These buttons indicate to users that they must deliberately choose one action from the set of actions presented, for the system to continue or finish a task.
+
+Action buttons must be presented together in one place, in a way that feels natural within the UI. When several action buttons are presented, their actions are often contradictory; for example **Discard** versus **Save**.
+
+When it comes to details such as placement of action buttons, choice of labels, and adding meaningful graphical details, it's important to follow a few design conventions, explained in this guide.
+
+<br/>
+<br/>
+
+*****
+
+## Placement
+On a user interface which is designed for a left-to-right script like English, action buttons of a footer (for example in a dialog) are typically located at the bottom-right corner.
+
+When there are two buttons placed together, the button that represents a positive action is always on the right side, and the one which presents a negative one is on the left.
+
+A positive action could be an action that:
+- takes the user forwards or to next step; e.g. *Next*, *Skip*, or *Continue*
+- is primary, and most probably the one that users is going to select, e.g. *Save*, *Update*, *Send* (or even *Delete* which is a negative notion, but can be an expected or "primary" action, depending on the context, like in a confirm deletion dialog)
+
+A negative action could be an action that:
+- takes the user back, dismisses a dialog, or cancels an ongoing process; e.g. *Back*, *Cancel*, *Discard*
+- is secondary, which means it is not likely that users clicks it in that context
+
+<limel-example-action-buttons-placement></limel-example-action-buttons-placement>
+
+:::tip
+In [`limel-dialog`](#/components/limel-dialog), there is a `slot` called `button`, which is a "flexbox". It automatically aligns its child elements to the right side. All you need to do is add `slot="button"` to the elements you wish to appear in the slot.
+
+```tsx
+<limel-button
+    label="OK"
+    slot="button"
+/>
+```
+:::
+
+<br/>
+
+*****
+
+## Primary versus secondary actions
+
+As you know, `limel-button` can have a `primary={true}` property, which makes it stand out by getting the primary color of your theme.
+
+One of the action buttons is normally the primary action. That is the action that you expect the users to take, or you that want them to take.
+
+<limel-example-action-buttons-primary-secondary></limel-example-action-buttons-primary-secondary>
+
+A primary button has a stronger visual weight, and appearing more distinct makes it easier for users to chose the correct, safe, or promoted action.
+
+You can use this visual distinction to promote the action that is safer, even if that action may not be the primary action. You can do so because sometimes:
+- it can be hard to identify the intention of the user
+- the action may be irreversible or risky
+
+<limel-example-action-buttons-primary-secondary-reversed></limel-example-action-buttons-primary-secondary-reversed>
+
+<br/>
+
+*****
+
+## Choosing good labels
+
+Labels you use should make sense together. You are usually dealing with contradictory actions. Therefore, what you put on each button should also reflect the contradiction.
+
+<limel-example-action-buttons-choosing-labels></limel-example-action-buttons-choosing-labels>
+
+Labels must make each option as clear as possible. An explicit label give users more confidence in selecting the correct action.
+
+:::note
+- Use a label that explains what the button does, rather than a generic label (such as “OK”, or "Yes").
+- Using verbs as labels instead of nouns is better, because a verb describes the action better.
+:::
+
+<limel-example-action-buttons-choosing-explicit-labels></limel-example-action-buttons-choosing-explicit-labels>
+
+<br/>
+
+*****
+
+## Providing a third alternative action
+
+Sometimes having two actions is not enough for the user to make a decision with peace of mind. A common scenario is when a user has unsaved changes and tries to navigate away from the view. In such cases, typically a prompt maybe be shown to alert the user and give them a few logical choices.
+
+This prompt can offer two main actions of "Save" and "Discard" (which means "save & continue navigating away from this page" and "discard & continue navigating away from this page"), along with a third choice that allows users to cancel (and go back / not continue navigating away!).
+
+<limel-example-action-buttons-third-alternative></limel-example-action-buttons-third-alternative>
+
+The action that takes users one step back in the process is intentionally placed on the left side. This makes it easier for users to mentally separate the other two. Additionally, the placement makes it easier to connect the action to the notion of going back. (For left-to-right scripts. Left and right will of course have reversed meaning in right-to-left script, which is precisely why the CSS properties for these things use values like `flex-start` and `flex-end`, rather than `left` and `right`. See below.)
+
+:::tip
+If buttons are placed in a full width flex container that has `justify-content: flex-end;`, these styles will align the targeted button to left side of the flex box. See "Example with three action buttons" in the [documentation for `limel-dialog`](#/component/limel-dialog/).
+
+```scss
+.back-button {
+    justify-self: flex-start;
+    margin-right: auto;
+}
+```
+:::
+
+<br/>
+
+*****
+
+## Adding more meaning with colors and icons
+
+To make it easer for users to understand the message and make a decision faster, you can use icons and colors on the buttons.
+
+Learn more about [usage of colors and color conventions in the UI](#/DesignGuidelines/color-system.md/), and [how to color limel-buttons](#/component/limel-button/).
+
+<limel-example-action-buttons-icon-color></limel-example-action-buttons-icon-color>
+
+:::note
+It might be tempting to use colors on action buttons to make the user interface more cheerful. But using colors in a user interface must follow certain UI design conventions, as colors convey meaning. Learn more about [usage of colors and color conventions in the UI](#/DesignGuidelines/color-system.md/).
+:::
+
+<limel-example-action-buttons-primary-secondary-reversed-colors></limel-example-action-buttons-primary-secondary-reversed-colors>
+
+:::note
+If an action is not irreversible or too risky, the design does not have to put a lot of visual emphasis on the secondary action, by means of icons and colors. In such cases, highlighting the primary action and making it more recognizable is more relevant, and more motivated.
+:::
+
+<limel-example-action-buttons-colors-do-dont></limel-example-action-buttons-colors-do-dont>
--- a/src/design-guidelines/action-buttons/examples/action-buttons.scss
+++ b/src/design-guidelines/action-buttons/examples/action-buttons.scss
@@ -1,0 +1,89 @@
+@import '../../_generic-example-styles/_do-dont';
+
+// We don't want the consumers to see these styles,
+// as they are just to visualize the fake examples, thus distracting.
+* {
+    box-sizing: border-box;
+}
+
+.example {
+    margin: 0.5rem 0;
+    background-color: rgb(var(--contrast-600));
+}
+
+.fake-dialog-container {
+    --header-top-right-left-border-radius: 0.25rem;
+    width: 100%;
+    height: 100%;
+    max-width: 40rem;
+    padding: 0 3rem 3rem 3rem;
+    margin: auto;
+    background-color: rgb(var(--contrast-600));
+
+    &.shows-full-dialog {
+        padding-top: 3rem;
+    }
+
+    p {
+        font-size: 0.875rem;
+        opacity: 0.7;
+        align-self: flex-start;
+        padding: 0 0.75rem;
+    }
+
+    h4 {
+        padding: 0 0.75rem;
+        margin-bottom: 0;
+    }
+}
+
+.action-bar {
+    display: flex;
+    gap: 0.5rem;
+    width: 100%;
+    justify-content: flex-end;
+    padding: 0.5rem;
+}
+
+.fake-dialog {
+    display: flex;
+    flex-direction: column;
+    min-height: 3rem;
+    background-color: rgb(var(--contrast-100));
+    box-shadow: var(--shadow-depth-16);
+    border-radius: 0 0 0.25rem 0.25rem;
+
+    .shows-full-dialog & {
+        border-radius: 0.25rem;
+    }
+}
+
+limel-header {
+    &.save {
+        --header-icon-color: rgb(var(--color-lime-default));
+    }
+    &.delete {
+        --header-icon-color: rgb(var(--color-red-default));
+    }
+}
+
+.button {
+    &.justify-left {
+        justify-self: flex-start;
+        margin-right: auto;
+    }
+    &.primary--neutral {
+        --lime-primary-color: rgb(var(--contrast-1100));
+    }
+    &.primary--caution {
+        --lime-on-primary-color: rgb(var(--color-gray-darker));
+        --lime-primary-color: rgb(var(--color-amber-default));
+    }
+    &.primary--danger {
+        --lime-on-primary-color: rgb(var(--color-red-default));
+        --lime-primary-color: rgb(var(--contrast-100));
+    }
+    &.primary--danger-highlighted {
+        --lime-primary-color: rgb(var(--color-red-default));
+    }
+}
--- a/src/design-guidelines/action-buttons/examples/icon-color.tsx
+++ b/src/design-guidelines/action-buttons/examples/icon-color.tsx
@@ -1,0 +1,83 @@
+import { Component, h } from '@stencil/core';
+
+@Component({
+    tag: 'limel-example-action-buttons-icon-color',
+    shadow: true,
+    styleUrl: 'action-buttons.scss',
+})
+export class ActionButtonsIconColorExample {
+    public render() {
+        return [
+            <div class="action-buttons-examples">
+                <div class="example">
+                    <div class="fake-dialog-container shows-full-dialog">
+                        <div class="fake-dialog">
+                            <limel-header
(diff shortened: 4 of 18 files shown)


src/design-guidelines/action-buttons/examples/action-buttons.scss
```diff
@@ -0,0 +1,89 @@
+@import '../../_generic-example-styles/_do-dont';
```
adrianschmidt, line 1: I'll swap the order of the commits, because this file doesn't exist yet at the time of this commit 😄

guides.ts
```diff
@@ -11,6 +11,7 @@ export default [
             'src/design-guidelines/color-system/color-system.md',
             'src/design-guidelines/shadows/shadows-and-depth.md',
             'src/design-guidelines/size/size-rhythms.md',
+            'src/design-guidelines/action-buttons/action-buttons.md',
```
adrianschmidt, line 14: Just out of curiosity, is there any particular order to these?

Kiarokh: Not a hardcore order. Just felt these less-visual and more usability-focused design guidelines can come further down.

src/design-guidelines/action-buttons/action-buttons.md
```diff
@@ -0,0 +1,126 @@
...
+## Choosing good labels
+
+Labels you use should make sense together. You are usually dealing with contradictory actions. Therefore, what you put on each button should also reflect the contradiction.
+
+<limel-example-action-buttons-choosing-labels></limel-example-action-buttons-choosing-labels>
+
+Labels must be making each option as clear as possible. An explicit label give users more confidence in selecting the correct action.
+
+:::note
+- Make sure the label explains what the button does, rather than a generic label (such as “OK”, or "Yes").
```
adrianschmidt, line 73: This sentence uses different grammatical forms before and after the comma.

```suggestion
- Use a label that explains what the button does, rather than a generic label (such as “OK”, or "Yes").
```

src/design-guidelines/action-buttons/examples/labels-explicit.tsx
```diff
@@ -0,0 +1,42 @@
...
+                        </div>
+                    </div>
+                </div>
+                <div class="do">
+                    <limel-header icon="ok" heading="Do"></limel-header>
+                    <div class="fake-dialog-container shows-full-dialog">
+                        <div class="fake-dialog">
+                            <p>Discard this draft?</p>
+                            <div class="action-bar">
+                                <limel-button label="Cancel" primary={true} />
```
adrianschmidt, line 33: "Cancel" isn't a great example here. How about "Stay on page" or "Back to editing"?

src/design-guidelines/action-buttons/action-buttons.md
```diff
@@ -0,0 +1,126 @@
...
+
+<limel-example-action-buttons-choosing-explicit-labels></limel-example-action-buttons-choosing-explicit-labels>
+
+<br/>
+
+*****
+
+## Providing a third alternative action
+
+Sometimes having two action is not enough for the user to take a decision with peace of mind. A common scenario is when a user has unsaved changes and tries to navigate away from the view. In such cases, typically a prompt maybe be shown to alert the user and give them a few logical choices.
```
adrianschmidt, line 85:
```suggestion
Sometimes having two actions is not enough for the user to make a decision with peace of mind. A common scenario is when a user has unsaved changes and tries to navigate away from the view. In such cases, typically a prompt maybe be shown to alert the user and give them a few logical choices.
```

src/design-guidelines/action-buttons/action-buttons.md
```diff
@@ -0,0 +1,126 @@
...
+
+This prompt can offer two main actions of "Save" and "Discard" (which means "save & continue navigating away from this page" and "discard & continue navigating away from this page"), along with a third choice that allows users to "Cancel" (and go back / not continue navigating away!).
```
adrianschmidt, line 87:
```suggestion
This prompt can offer two main actions of "Save" and "Discard" (which means "save & continue navigating away from this page" and "discard & continue navigating away from this page"), along with a third choice that allows users to cancel (and go back / not continue navigating away!).
```

The quote marks and uppercase 'C' suggests "Cancel" as the actual label, like "Save" and "Discard", but "Cancel" isn't a good label for all the reasons outlined in the previous section 😄

src/design-guidelines/action-buttons/action-buttons.md
```diff
@@ -0,0 +1,126 @@
...
+
+<limel-example-action-buttons-third-alternative></limel-example-action-buttons-third-alternative>
+
+The action that takes users one step back inte the process is intentionally placed on the left side. This firstly makes it easier for users to mentally separate the other two, and secondly the placement makes it easier to connect the action to the notion of going back, in a UI which is designed for left-to-right script.
```
adrianschmidt, line 91:
```suggestion
The action that takes users one step back in the process is intentionally placed on the left side. This makes it easier for users to mentally separate the other two. Additionally, the placement makes it easier to connect the action to the notion of going back. (For left-to-right languages. Left and right will of course have reversed meaning in right-to-left languages, which is precisely why the CSS properties for these things use values like `flex-start` and `flex-end`, rather than `left` and `right`. See below.)
```

src/design-guidelines/action-buttons/action-buttons.md
```diff
@@ -0,0 +1,126 @@
...
+```
+:::
+
+<br/>
+
+*****
+
+## Adding more meaning with colors and icons
+
+To make it easer for users to understand the message and take a decision faster, you can use icons and colors on the buttons.
```
adrianschmidt, line 110:
```suggestion
To make it easer for users to understand the message and make a decision faster, you can use icons and colors on the buttons.
```

src/design-guidelines/action-buttons/examples/icon-color.tsx
```diff
@@ -0,0 +1,83 @@
...
+                <div class="example">
+                    <div class="fake-dialog-container shows-full-dialog">
+                        <div class="fake-dialog">
+                            <limel-header
+                                icon="ask_question"
+                                heading="Save changes?"
+                                class="save"
+                            />
+                            <p>
+                                You have unsaved change. Do you want to save
```
adrianschmidt, line 21:
```suggestion
                                You have unsaved changes. Do you want to save
```

src/design-guidelines/action-buttons/examples/icon-color.tsx
```diff
@@ -0,0 +1,83 @@
...
+                                permanent action and cannot be undone!
+                            </p>
+                            <div class="action-bar">
+                                <limel-button
+                                    label="Back to selection"
+                                    class="button primary--neutral justify-left"
+                                    icon="left_arrow"
+                                />
+                                <limel-button
+                                    label="Don't delete"
```
adrianschmidt, line 65: This choice is non-obvious to me. With some thought, I understand it, but it's not optimal.

"Continue without deleting" is long, but at least explains the action.

src/design-guidelines/action-buttons/action-buttons.md
```diff
@@ -0,0 +1,126 @@
...
+
+Learn more about [usage of colors and color conventions in the UI](#/DesignGuidelines/color-system.md/), and [how to color limel-buttons](#/component/limel-button/).
+
+<limel-example-action-buttons-icon-color></limel-example-action-buttons-icon-color>
+
+:::note
+It might be tempting to use colors on action buttons, with the intention of making the user interface more cheerful. But using colors in a user interface must follow certain UI design conventions, as they convey meanings. Learn more about [usage of colors and color conventions in the UI](#/DesignGuidelines/color-system.md/).
```
adrianschmidt, line 117:
```suggestion
It might be tempting to use colors on action buttons to make the user interface more cheerful. But using colors in a user interface must follow certain UI design conventions, as colors convey meaning. Learn more about [usage of colors and color conventions in the UI](#/DesignGuidelines/color-system.md/).
```

src/design-guidelines/action-buttons/action-buttons.md
```diff
@@ -0,0 +1,126 @@
...
+:::
+
+<limel-example-action-buttons-primary-secondary-reversed-colors></limel-example-action-buttons-primary-secondary-reversed-colors>
```
adrianschmidt, line 120: It's not obvious how this examples relates to the surrounding text.

adrianschmidt (commented):
Great work! I have added some minor fixups, and a bunch of suggestions in comments 🙂

src/design-guidelines/action-buttons/action-buttons.md
```diff
@@ -0,0 +1,121 @@
+# Action buttons
+
+Action buttons are typically used in dialogs' footers, or in headers. These are buttons indicate to users that they must deliberately choose one action from a set of actions which are presented, for the system to continue or finish a task.
+
+Action buttons must naturally be presented together in one place; within the UI. When several action buttons are presented, their actions are often contradictory; for example **Discard** versus **Save**.
+
+When it comes to details such as placement of action buttons, choice of labels, and adding meaningful graphical details, it's important to follow a few design conventions, explained here:
```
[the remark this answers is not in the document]
Kiarokh: That's right. We can say "explained in this page" or "below"

src/design-guidelines/action-buttons/action-buttons.md
```diff
@@ -0,0 +1,121 @@
...
+
+<br/>
+<br/>
+
+*****
```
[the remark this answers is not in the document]
Kiarokh: Well, that's of course better. I thought we're usign `<hr>`s between sections of documentations as well, for instance between _Examples_ and _Properties_ sections. But they are just borders now when I check. So it should be fine to update Kompendiums styles for `<hr>`s

src/design-guidelines/action-buttons/examples/icon-color.tsx
Kiarokh: I think the length is the problem itself indeed. 
I have seen labels like this many times in Mac OS context. For instance:
image
So it should be fine.

src/design-guidelines/action-buttons/action-buttons.md
```diff
@@ -0,0 +1,121 @@
+# Action buttons
+
+Action buttons are typically used in dialogs' footers, or in headers. These are buttons indicate to users that they must deliberately choose one action from a set of actions which are presented, for the system to continue or finish a task.
```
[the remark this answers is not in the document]
Kiarokh: I don't think _dialog's footer_ (or "dialogs' footers" in this case is grammatically wrong 🙃. Maybe "dialogs' footers" sounds a bit silly though. 
image
[form here](https://developer.apple.com/design/human-interface-guidelines/macos/buttons/bevel-buttons/).

src/design-guidelines/action-buttons/action-buttons.md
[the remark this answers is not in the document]
adrianschmidt: I was thinking more like add a border and some top margin to `<h2>`, and remove the need to manually add a `<hr>` every time.

Anyway, it's a nitpick and we don't have to do anything about it right now. It's better to get this documentation out there before we go on vacation.

src/design-guidelines/action-buttons/action-buttons.md
Kiarokh: Perhaps I should have included a "Don't" example for example with an unconventional color (instead of red) on a negative action button.

src/design-guidelines/action-buttons/action-buttons.md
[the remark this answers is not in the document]
Kiarokh: yes, good point. we can come back to it. ;)

src/design-guidelines/action-buttons/action-buttons.md
adrianschmidt: Ah, so, is this related to the note about not adding colours just to be cheerful?

Kiarokh: Yes. I was just trying to say that the orange button or the red button have their colors for a reason. E.g. Warning, Danger, etc... so don't just pick a random color to make things look nice. :)

src/design-guidelines/action-buttons/action-buttons.md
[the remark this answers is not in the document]
adrianschmidt: No, it's not that it's grammatically wrong. It's just that "used in dialog footers" works just as well and is less clunky to say. The difference is easier to explain in Swedish:

- used in dialogs' footers -> används i dialogers sidfötter
- used in a dialog's footer -> används i en dialogs sidfot
- used in dialog footers -> används i dialogsidfötter

This is just to explain the relation between the different forms. In Swedish, I think the last form is the clunkiest, but in English, I think it's very neat.

"Used in a window's main content area" can't really be rewritten in the same way, because while 
(at least I think) you can call the footer of a dialog a "dialog footer", it doesn't really work to call the main content area of a window the "window main content area" (think of it as a single concatenated word in Swedish, like "windowmaincontentarea", vs "dialogfooter").

See the use of "dialog body" and "dialog footer" here:
image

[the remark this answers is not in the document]
adrianschmidt: (You're perfectly free to choose which form to use of course, it's just a suggestion 😊)

src/design-guidelines/action-buttons/action-buttons.md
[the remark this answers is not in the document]
adrianschmidt:
```suggestion
When it comes to details such as placement of action buttons, choice of labels, and adding meaningful graphical details, it's important to follow a few design conventions, explained in this guide.
```

src/design-guidelines/action-buttons/action-buttons.md
```diff
@@ -0,0 +1,126 @@
...
+- is primary, and most probably the one that users is going to select, e.g. *Save*, *Update*, *Send* (or even *Delete* which is a negative notion, but can be an expected or "primary" action, depending on the context, like in a confirm deletion dialog)
+
+A negative action could be an action that:
+- takes the user back, dismisses a dialog, or cancels an ongoing process; e.g. *Back*, *Cancel*, *Discard*
+- is secondary, which means it is not likely that users clicks it in that context
+
+<limel-example-action-buttons-placement></limel-example-action-buttons-placement>
+
+:::tip
+In [`limel-dialog`](#/components/limel-dialog), there is a `slot` called `button`, which is flex-box. It automatically aligns its child elements to the right side. All you need to do is to write `slot="button"` on your elements.
```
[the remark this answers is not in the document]
adrianschmidt: As in this example, some of these suggestions contain two unrelated changes, that just happen to appear on the same line of code. You're obviously free to accept or reject them independently  (although you'll have to add the change manually if you wish to just add one and not the other).

src/design-guidelines/action-buttons/examples/icon-color.tsx
adrianschmidt: Ah… ok, the "Save" vs "Don't Save" makes more sense to me, because I can imagine the context, and thus imagine the difference between the actions: I'm closing a modified document, *Save* will save and close, *Don't Save* will discard my changes and close, and *Cancel* will take me back to the document, not close it.

I'm having trouble imagining the context for your example though:

image

It seems I've probably been on some kind of list or table, and selected multiple items, then pressed some button to delete the selected items. If I press "Back to selection", that's obviously analogous to "Cancel" in your screenshot; I go back to the list or table I came from, and nothing is deleted. If I press "Delete", I obviously delete the items, but where do I end up after that? Wouldn't I normally be back at the list or table? But if so, then what's the difference between "Back to selection" and "Don't delete"? Both would _not delete the items_, obviously, but both would also have me back at the list or table, no?

I understand that this is just a made up example, and to many, the context won't matter, and sure, I understand what the example is trying to say anyway, but the fact that I can't imagine a context that makes sense gives me the feeling that maybe I've missed or misunderstood something. I don't get a sense of confidence, but start to doubt either myself or the reliability of the documentation (probably both, with somewhat alternating intensity). While I know I'm probably odd, I also assume I'm not unique 😆

src/design-guidelines/action-buttons/action-buttons.md
adrianschmidt: Yeah, I think adding a "Don't" would help then. Also, adding a caption to each example would really help in understanding them. Like they do on material.io.

adrianschmidt: Or, frankly, I'd maybe just remove the examples. Even with a "Don't" example, I think the note is perfectly clear, and the examples don't actually help in understanding the message in this case. It's up to you though.

src/design-guidelines/action-buttons/examples/icon-color.tsx
Kiarokh: I totally see your point! ;)

It's quite hard to illustrate the entire interaction flow chart in one shot though. 

_Step 1 > step 2 > modal with action buttons > different outcomes_


I hope the reader can imagine that there was a step before this modal where the imaginary user selected 23 items (while they could select more or less) like in a table view or list view. Then they did an action, for example clicked on [Delete & Close] or [Continue the cleanup process…] or whatever action that "leaves" this current page and continues, permanently removing their ability to go back and change their mind. Now they are asked if they're sure that this is the outcome they want. They maybe accidentally chose 23 items, maybe 20 items were selected in a page before, and 3 in the visible current page (imagine a table with pagination). 

I just had a hard time to find a better example. But it'd be nice to find ways of explaining and clarifying the situation that lead to this modal. If you can find a better scenario please suggest :) 

Otherwise, I think the example is good enough for the purpose of this guide. At the end of the day, these guidelines are for inspiration, and to sync conventions in our different products. Developers shouldn't be left alone to design as well as code. When and how a confirmation modal should be shown, and with how many buttons and which button labels is subject to a design decision, and hopefully there is a designer supporting them for each case.

src/design-guidelines/action-buttons/action-buttons.md
Kiarokh: Let's leave them there for now. I'll come up with a better clarification in future and update this guide.

src/design-guidelines/action-buttons/examples/icon-color.tsx
adrianschmidt: 👍

It's absolutely good enough. Let's get this baby merged! 🙂

src/design-guidelines/action-buttons/action-buttons.md
```diff
@@ -0,0 +1,126 @@
...
+
+When it comes to details such as placement of action buttons, choice of labels, and adding meaningful graphical details, it's important to follow a few design conventions, explained here:
+
+<br/>
+<br/>
+
+*****
+
+## Placement
+On a user interface which is designed for a left-to-right script like English, action buttons of a footer (for example in a dialog) are typically located at the bottom-right corner.
```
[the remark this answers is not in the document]
Kiarokh: a language cannpt be right-to-left. The way it is written can be :) I think that's why they chose to say script:

https://en.wikipedia.org/wiki/Right-to-left_script

But I don't know.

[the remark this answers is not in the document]
adrianschmidt: Ok 👍

src/design-guidelines/action-buttons/examples/labels-explicit.tsx
Kiarokh: Sound great! Specially back to editing? Will you suggest a change? :)

adrianschmidt:
```suggestion
                                <limel-button label="Back to editing" primary={true} />
```

src/design-guidelines/action-buttons/action-buttons.md
Kiarokh: right 😄

src/design-guidelines/action-buttons/action-buttons.md
Kiarokh: scripts, not languages?

src/design-guidelines/action-buttons/action-buttons.md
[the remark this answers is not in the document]
adrianschmidt: In case you liked this suggestion. I think "of" is ok to though.

```suggestion
On a user interface which is designed for a left-to-right script like English, action buttons in a footer (for example in a dialog) are typically located at the bottom-right corner.
```

src/design-guidelines/action-buttons/action-buttons.md
adrianschmidt:
```suggestion
The action that takes users one step back in the process is intentionally placed on the left side. This makes it easier for users to mentally separate the other two. Additionally, the placement makes it easier to connect the action to the notion of going back. (For left-to-right scripts. Left and right will of course have reversed meaning in right-to-left script, which is precisely why the CSS properties for these things use values like `flex-start` and `flex-end`, rather than `left` and `right`. See below.)
```

adrianschmidt (approved):

PR #1297, opened 2021-07-08
Update MDC from our own fork of v4.0.0 to the official v11.0.0

adrianschmidt:
fix: #1070

Issues yet to be fixed (merge-blockers are marked with ❌, any other issues are _not_ merge blockers):
- switch
  - [ ] disabled does not look disabled (does not seem to be a new bug, but not fixed with this upgrade neither)
- list
  - [x] ❌ multi-lined lists do not work anymore
- tab bar
  - [x] Not possible to change tabs using keyboard's arrow keys. Pressing them only jumps to the first or last tab.
- file
  - [x] disabled field can receive focus when clicked, which will float the label due to `lime-notched-outline--notched` and `lime-floating-label--float-above` (recent fixes in chipset). But no file browser window opens. So this looks only like a non-important visual bug
- picker
  - [x] ❌ Neither disabled nor readonly are actually disabled. You can still click on the `<input>` part to open the dropdown menu.
- select
  - [x] ❌ floating label doesn't float
  - [x] floating label becomes gray in `readonly`
  - [x] `mdc-select-helper-text` is visually misplaced due to some margin
  - [x] too much padding in `limel-select`, when it has `is-narrow` class
  - [ ] keyboard navigation doesn't properly work when an option is selected
- button-group:
  - a disabled button-group can be changed using the keyboard, but that's not a new bug
- chip-set:
  - [x] ❌ pressing the trailing icon on a chip does not remove the chip
  - [x] ❌ floating label doesn't float
  - [ ] missing translations for "Remove chip"
- input-field:
  - [x] the outline doesn't become green when the input is focused
  - [x] when invalid, the border turns from red back to grey when blurring the input
    (personally, I think this is actually better, as long as there is a warning icon, but that icon is missing for certain components, like input-field of type `textarea`, and some other components that look like input-field, like chip-set for example)
  - [x] `lime-formatted-input`'s placement in the DOM for `type="textarea"` causes the height of component to increase.
  - [x] Those full-width `readonly` trailing icons (when `showLink=true`):
      - [x] ❌ show wrong and confusing visual feedback on hover  in `readonly`
      - [x] do not get `box-shadow` on hover (generally, the way they get box-shadow must be improved, for example when invalid, they may show a red shadow or border or similar).
  - [x] ❌ the floating label for `textarea`s interfere with the text in the field (styling issue)
- menu:
  - [x] there is no hover highlight of items
- slider:
  - [x] ❌ the value of a disabled slider can still be changed
  - [x] most of our custom styling doesn't work
      - [x] ❌ layout is generally broken
      - [x] percentage colors don't work
      - [x] our special effects on slider thumb are gone
      - [x] special styling for `readonly` doesn't work
      - [x] ❌ the current value is only shown while the slider thumb is focused (it should always be shown)
- snackbar:
  - [ ] missing translations for `Dismiss`
- linear-progress:
  - [x] does not look good on colored backgrounds

- [ ] look over commented code sections and consider if they should be used or removed.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/button-group/button-group.scss
```diff
@@ -14,6 +14,8 @@
     grid-auto-flow: column;
     background-color: transparent;
     margin: functions.pxToRem(2) !important;
+    font-size: functions.pxToRem(14);
+    line-height: 1.25rem;
```
Kiarokh, line 18: Here, the line-height does not seem to have any visual effect (because of `align-items: center;`) on `.mdc-chip`, neither for iconed buttons, nor for text-only buttons. Thus can be removed.


https://user-images.githubusercontent.com/35954987/125753224-b1001440-e510-44e4-9969-93379644de3f.mp4

src/components/slider/partial-styles/_readonly.scss
```diff
@@ -11,16 +11,16 @@
 
         .mdc-slider__pin,
         .mdc-slider__track {
-            background-color: var(
-                --lime-primary-color,
-                mdc-variables.$mdc-theme-primary
-            );
+            // background-color: var(
```
Befkadu1, line 14: We have 2 issues here: 
1. The knob is not on the slider bar. 
2. We should remove the shadow from `.mdc-slider__thumb`. 

We don't see the value inside the knob, this might not be an issue but I should mention it here since it is a different feature. The slider value can be seen on top of the knob when draging the knob

Screenshot 2021-07-15 at 09 59 04

src/components/slider/slider.scss
```diff
@@ -21,16 +22,16 @@
...
-        var(--lime-primary-color, $mdc-theme-primary)
-    );
-    @include mdc-slider-thumb-color(
-        var(--lime-primary-color, $mdc-theme-primary)
-    );
-    @include mdc-slider-value-pin-fill-color(
-        var(--lime-primary-color, $mdc-theme-primary)
-    );
-    @include mdc-slider-value-pin-ink-color(on-primary);
+    // @include mdc-slider-highlight-color(
```
Befkadu1, line 25: We have 2 issues here:

1. The knob is not on the slider bar.
2. We should may be remove the shadow from .mdc-slider__thumb on line 207? .
We don't see the value inside the knob, this might not be an issue but I should mention it here since it is a different feature. The slider value can be seen on top of the knob when draging the knob

Screenshot 2021-07-15 at 09 59 04

src/components/button-group/button-group.scss
adrianschmidt: While we should certainly strive to clean stuff like this up, this is things I don't think we should spend _any_ time on before we get this merged.

src/components/input-field/input-field.scss
```diff
@@ -45,25 +45,33 @@
     }
 }
 
+@include textfield.core-styles;
```
Befkadu1, line 48: I don't know on which line I should give this comment but the floating label that should be above in case of textarea doesn't look good. 

Screenshot 2021-07-15 at 11 04 01

adrianschmidt: Oh yeah, good catch! I've seen this but had forgotten about it.

(Don't worry about what line to write a comment on. As long as it's on _some_ line, so answers can be threaded, I'm happy ☺️)

src/style/internal/shared_input-select-picker.scss
```diff
@@ -243,12 +240,12 @@ $height-of-helper-text-pseudo-before: 0.75rem; // There is strange a pseudo befo
...
                     }
                 }
-                .trailing-icon {
+                .mdc-text-field__icon--trailing {
                     color: var(--lime-error-text-color);
                 }
             }
 
-            .mdc-text-field__icon.trailing-icon {
+            .mdc-text-field__icon--trailing.trailing-icon {
```
Kiarokh, line 248: https://user-images.githubusercontent.com/35954987/125762427-e617b685-a5ec-4226-933c-1b96bca6990f.mov

There are a few visual problems here in the video. The biggest one which is the absence of full-width links in readonly mode is caused by `.mdc-text-field__icon--trailing.trailing-icon`, which should have been `.mdc-text-field__icon--trailing` only. I'll push one or a couple of fixups.

src/style/internal/shared_input-select-picker.scss
```diff
@@ -192,19 +190,18 @@ $height-of-helper-text-pseudo-before: 0.75rem; // There is strange a pseudo befo
...
     .mdc-text-field.mdc-text-field--focused.mdc-text-field--outlined {
         .mdc-notched-outline--notched {
             .mdc-notched-outline__notch {
                 padding-top: 0;
             }
         }
     }
+
+    .mdc-text-field__icon--leading,
+    .mdc-text-field__icon--trailing {
```
Kiarokh, line 202: Not a big deal, but we could write this for `mdc-text-field__icon`, since both leading and trailing icons share this same class, right?

adrianschmidt: Yeah, that makes sense 👍

Kiarokh: me do a fixup separately for this, so in case you have a reason for specifying these, we can drop that fixup

src/components/dialog/dialog.tsx
```diff
@@ -9,7 +10,7 @@ import {
     Prop,
     Watch,
 } from '@stencil/core';
-import * as focusTrap from 'focus-trap';
+// import * as focusTrap from 'focus-trap';
```
Befkadu1, line 13: Opening the dialog shows the error below in the dev tool. I can see that you re trying to import FocusTrap here. 

Screenshot 2021-07-15 at 11 29 33

src/components/button-group/button-group.scss
Kiarokh: Right, but it may be hard to forget it in future. Since we will never remember when/why it was added, and whether it can be removed or not. Usually these will remain in the code, and in some other situations we will have to override them elsewhere without knowing that they weren't needed. :) So I'll remove this with a fixup for now. If I find new things like this, I suggest we remove them too.

src/components/dialog/dialog.tsx
adrianschmidt: Ah, yeah, good find! I commented out the importing of focusTrap, because it caused a build error. I think the new dialog had some updated functionality for focus trapping…

src/components/button-group/button-group.scss
adrianschmidt: Ok. But since it's now reported in this comment, it at least won't be forgotten as long as this PR is open. So can you do the slider styling first? 😁

Kiarokh: 😄 for now the input-field is killing me. After that I'll prioritize the slider ;)

src/components/dialog/dialog.tsx
adrianschmidt: I'll have a look at this straight away.

adrianschmidt: Oh yeah! 🤘

image

adrianschmidt: ⚡

src/style/internal/shared_input-select-picker.scss
adrianschmidt: image

Wait… you made a video of a video? 🤨 😂

src/components/input-field/input-field.tsx
```diff
@@ -292,32 +295,35 @@ export class InputField {
...
+
         const labelClassList = {
             'mdc-floating-label': true,
             'mdc-floating-label--float-above':
                 !!this.value || this.isFocused || this.readonly,
         };
 
         return [
-            <div class={this.getContainerClassList()}>
+            <label class={this.getContainerClassList()}>
```
Kiarokh, line 310: Why is this `<label` instead of `<div`? On MD's site, it's rendered as a div.
image

Maybe this doesn't affect the styles. But it feels semantically wrong to call it label. What do you say? Can I change it to div @adrianschmidt?

src/components/input-field/input-field.tsx
```diff
@@ -292,32 +295,35 @@ export class InputField {
...
-                            htmlFor="tf-input-element"
-                        >
-                            {this.label}
-                        </label>
-                    </div>
-                    <div class="mdc-notched-outline__trailing"></div>
-                </div>
-                {this.renderIcons()}
-            </div>,
+                {this.renderTrailingIconNew()}
```
Kiarokh, line 325: leading and trailing icons should be placed and rendered before `mdc-notched-outline`, otherwise they create styling issues.

image

src/components/input-field/input-field.tsx
adrianschmidt: I don't think it's semantically wrong. It's semantically correct to write either
```html
<label for="blah">Blah blah</label>
<input id="blah" />
```
or
```html
<label>
    Blah blah
	<input />
</label>
```
and the code we use is equivalent to the latter example, just with a lot more gunk…

The code I've used comes straight from MDC's docs for v11.0.0: https://github.com/material-components/material-components-web/tree/v11.0.0/packages/mdc-textfield#outlined-text-example

src/components/input-field/partial-styles/trailing-icon.scss
```diff
@@ -1,11 +1,11 @@
 @use '../../../style/functions';
 
 .mdc-text-field--with-trailing-icon {
-    .mdc-text-field__icon.trailing-icon {
+    .mdc-text-field__icon--trailing {
         transition: background-color 0.2s ease, box-shadow 0.3s ease;
         border-radius: 50%;
         padding: functions.pxToRem(8);
-        right: functions.pxToRem(8);
+        margin-right: functions.pxToRem(8);
```
Kiarokh, line 8: This was fine actually to be `right` and not `margin-right`. The problem is that some CSS that should be coming from Material Design styles is missing.

The thing is that `mdc-text-field__icon` should have an absolute position, which comes from these styles:
```css
.mdc-text-field--with-leading-icon .mdc-text-field__icon, 
.mdc-text-field--with-trailing-icon .mdc-text-field__icon {
    position: absolute;
    top: 50%;
    transform: translateY(-50%);
    cursor: pointer;
}
```
image

But for some reason they don't appear here, which messes up the UI.

I'll revert this change in a fixup soon. But I can't really figure out why the MD styles are not coming. All class names seem to be fine :/

src/style/internal/shared_input-select-picker.scss
Kiarokh: hehe, yes 😸 because first I used GIPHY CAPTURE to screen record, but they seem to use a codec that makes HUUUUGE video files. So I couldn't upload that file. I recorded again using QuickTime Player. :P

src/components/input-field/input-field.tsx
adrianschmidt: I'm not entirely convinced that the code structure used on material.io should take precedence over the code structure used in the docs for the MDC version we are using. But if something isn't working as it should, and you can make it work by moving stuff around, feel free to do so 😄

src/components/input-field/partial-styles/trailing-icon.scss
adrianschmidt: Again, _please_ use the actual docs for the actual MDC version we're using as your primary guidelines. If something doesn't work the way it's supposed to when using those, then go ahead and change.

If there are styles missing, we're probably missing a `@include textfield.core-styles;` or other mixin somewhere. The way styles are included have changed completely between MDC v4 and v11.

adrianschmidt: https://github.com/material-components/material-components-web/tree/v11.0.0/packages/mdc-textfield

src/style/internal/shared_input-select-picker.scss
adrianschmidt: I just use the built-in tools available via `cmd-shift-5`. I think those work great 🙂

src/components/input-field/input-field.tsx
Kiarokh: Huh 🤔. OK. Strange that they have div on their documentations site.

src/components/input-field/input-field.tsx
Kiarokh: >  those trailing links (phone, email, website) cannot be clicked, since they will be covered by other stuff.
Me neither. But this is one of the problems with it. You can test it in `readonly` mode.

src/components/input-field/input-field.tsx
adrianschmidt: I think material.io is more about documenting the design ideas, rather than the specific implementation. They might not be keeping their MDC version up to date. They also don't specify which version they are using, so it's pretty clear to me that the intention isn't for those docs to be technical documentation about any one specific implementation.

src/components/input-field/partial-styles/trailing-icon.scss
Kiarokh: Awesome. I'll check there as well now. 

In this case, the `right: functions.pxToRem(8);` was added by us to override MD's default `right`. This is because we add some hover effects to those icons and render them like clickable buttons and so forth. Thus, we needed some certain amount of space on the right side.

However, it's hard to say, without getting those expected styles form MDC. `@include textfield.core-styles;` and `@include icon.icon-core-styles;` seem to be there in `input-field.scss`. Must we add it elsewhere? I don't know where and how 😞

src/components/input-field/input-field.tsx
Kiarokh: Well, let's go with the code structure of MDC. In this case, I can say that since those trailing and leading icons are missing styles (which gives them `position: absolute`), it doesn't matter if they are first in the DOM or not. They should be visually rendered on a layer below the value, and also below the `formatted-input` span. Otherwise, our styles in `readonly` for `showLink=true` example won't work correctly. I push a fixup now. Can you please see how we can get those missing styles that [I explained here](https://github.com/Lundalogik/lime-elements/pull/1297#discussion_r670399736)? I can't figure out how.

But I should say it's impossible to track down visual issue, without a rendered example. They link to material.io from MDC docs on github :/ to showcase the components.

src/components/input-field/partial-styles/trailing-icon.scss
adrianschmidt: I'm not entirely sure what is missing and where… I'll ping you in Slack, and maybe you can show me.

src/components/input-field/input-field.tsx
adrianschmidt: > They should be visually rendered on a layer below the value, and also below the `formatted-input` span.

I'm pretty sure they changed that. They're now using flex to handle the placing of the icons and the input element, which is why the dom structure has changed. We can override any styling we want in the readonly styles, but I don't love the idea of overriding the basic styles to allow for a completely different dom structure, as I think that would likely be a constant source of future suffering 😄

Can't we update our readonly styles to fit the new structure instead?

src/components/input-field/partial-styles/trailing-icon.scss
adrianschmidt: Ah, wait, I think this has become the same discussion as https://github.com/Lundalogik/lime-elements/pull/1297#discussion_r670450262, right? If so, let's just keep it in that thread instead 😊

src/components/input-field/input-field.tsx
Kiarokh: > I don't love the idea of overriding the basic styles to allow for a completely different dom structure

I hate this idea 😂

cd> Can't we update our readonly styles to fit the new structure instead?

We certainly can. I was not aware that material.io is different than MDC. Well, if we load all their styles, then we should just modify our own modifications. 


> They're now using flex to handle the placing of the icons and the input element, 

Hm.... OK. Let me check.

Kiarokh: Can't we view some vanilla MDCs somewhere?

src/style/internal/shared_input-select-picker.scss
adrianschmidt: Resolving, as this discussion is continued here instead: https://github.com/Lundalogik/lime-elements/pull/1297#discussion_r670372397

src/components/input-field/input-field.tsx
adrianschmidt: I can probably set something just using what they describe in the docs… I'll have a look at that tomorrow, if you remind me 😄

src/components/input-field/partial-styles/trailing-icon.scss
```diff
@@ -16,7 +16,7 @@
         }
     }
     &:hover {
-        .trailing-icon {
+        .mdc-text-field__icon--trailing {
             box-shadow: var(--button-shadow-normal);
```
Kiarokh, line 20: These full-width readonly trailing icons do not get box-shadow on hover. The result is that case of type `email`, there is no visual effect to communicate that the entire field is clickable.


https://user-images.githubusercontent.com/35954987/125903965-72ef0c8f-f335-4882-b04d-4b0746517bb9.mov

(generally, the way they get box-shadow must be improved, for example when invalid, they may show a red shadow or border or similar).

adrianschmidt: The email-button could get a light grey colour perhaps?

If you push what you have, I can have a look.

src/components/input-field/partial-styles/trailing-icon.scss
```diff
@@ -1,11 +1,11 @@
 @use '../../../style/functions';
 
 .mdc-text-field--with-trailing-icon {
-    .mdc-text-field__icon.trailing-icon {
+    .mdc-text-field__icon--trailing {
```
Kiarokh, line 4: I'm not sure whether this is completely related to CSS, but type `urlAsText` and `tel` don't do anything when clicked (on the original draft PR that Adrian made). See the attached video. 


https://user-images.githubusercontent.com/35954987/125908548-43b9f449-91f2-4f1c-94af-4e4f5dac0319.mov



But when I pushed the last input-filed fixups to the draft PR, none of them works. :/ Not sure how those two fixups could affect this functionality. They seem completely irrelevant.


On my local version where I fixed a few other styling problems for readonly (not pushed yet) none of them works when clicked. See the video above.

Kiarokh: I pushed my local changes now.

src/components/input-field/input-field.scss
adrianschmidt: I'll mark this as resolved, not because it's fixed, but because I've added it to the list in the main description.

src/components/input-field/partial-styles/trailing-icon.scss
adrianschmidt: ⚡

src/components/input-field/partial-styles/trailing-icon.scss
adrianschmidt: ⚡

src/components/input-field/input-field.tsx
Kiarokh: It'd be awesome. Just to be able to compare how their vanilla components work, without any of our customizations. It's easier to track missing styles, and more.

src/components/input-field/partial-styles/trailing-icon.scss
Kiarokh: in the old version, the links get a hover effect, styles are already there. But they won't show up here.

adrianschmidt: Did you try my latest changes?

src/components/input-field/input-field.tsx
adrianschmidt: Yeah, I'll do that once all components are "not-broken". We shouldn't be spending time on polishing before then anyway. See it as a carrot, the less time you spend on polishing now, the faster we can get everything to "not-broken", and then I'll look at getting you that vanilla version to use while polishing. Ok? 😊

src/components/input-field/partial-styles/trailing-icon.scss
Kiarokh: Ah awesome! I read your comment in the code as well 🤩 Great job!

This page wasn't refreshed automatically. So I missed that there are new changes, until I wanted to push mine on slider, and then I saw I must pull ;)

src/components/input-field/input-field.scss
```diff
@@ -65,7 +65,14 @@
 }
 
 .lime-looks-like-input-value {
-    @include shared_input-select-picker.looks-like-input-label;
+    @include shared_input-select-picker.looks-like-input-value;
```
adrianschmidt, line 68: We still had this connection to label which shouldn't be there.

src/style/internal/shared_input-select-picker.scss
```diff
@@ -50,6 +50,17 @@ $height-of-helper-text-pseudo-before: 0.75rem; // There is strange a pseudo befo
...
+@mixin looks-like-input-value() {
+    font-family: Roboto, sans-serif;
+    line-height: functions.pxToRem(28);
+    -moz-osx-font-smoothing: grayscale;
+    -webkit-font-smoothing: antialiased;
+
+    font-size: functions.pxToRem(14);
+    font-weight: 400;
+    letter-spacing: 0.009375em;
+}
```
adrianschmidt, line 62: This is identical to `looks-like-input-label()`, except for the text-colour, since we don't want to link those. I haven't tested this, so we might need to add a colour here too, but that should be the same variable we use for the input _value_ not the variable we use for the _label_ (even if the current _value_ of those variables happen to be the same at the moment).

src/components/input-field/input-field.scss
```diff
@@ -65,7 +65,14 @@
...
+
+    .mdc-text-field--disabled & {
+        @include shared_input-select-picker.looks-disabled;
+    }
+}
+
+.lime-formatted-input {
```
adrianschmidt, line 75: The disabled-styling felt like it belonged in `.lime-looks-like-input-value`, but the rest of the stuff here, `display: none` and stuff, doesn't belong there. That's specific to the formatted value use case.

Kiarokh: completely right!

src/components/chip-set/chip-set.tsx
```diff
@@ -375,9 +376,16 @@ export class ChipSet {
                     readonly={this.isFull()}
                 />
                 <div
+                    // Because MD removes some classes in chipset, we add custom
+                    // classes with similar names and expected behavior & styles.
+                    // These class names start with `lime-`, instead of `mdc-`.
```
adrianschmidt, line 382: I don't think we really need this comment both here and in the scss, and I prefer it in the scss… what do you think @Kiarokh?

Kiarokh: I don't have any problems if we remove it here

src/components/spinner/spinner.scss
```diff
@@ -45,7 +45,7 @@
 
 :host(.dropdown--spinner) {
     margin: functions.pxToRem(10) 0;
-    color: var(--lime-primary-color, #{mdc-variables.$mdc-theme-primary});
+    // color: var(--lime-primary-color, #{mdc-theme.$primary});
```
adrianschmidt, line 48: Just noticed this now. We don't need to fix it here, but this feels like something the consumer should add, rather than the spinner component itself. We only use the `dropdown--spinner` class in `limel-picker`, so this stuff could go in that stylesheet instead of being built into the spinner.

PR #1305, opened 2021-07-21
fix(switch): make the structure more like the latest MDC & make disabled look disabled

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/switch/partial-styles/_readonly.scss
+++ b/src/components/switch/partial-styles/_readonly.scss
@@ -26,32 +26,32 @@
             }
         }
 
-        .mdc-switch__thumb-underlay {
-            left: -1rem;
-        }
-
         &.mdc-switch--checked {
             .mdc-switch__track {
                 background-color: var(--mdc-theme-primary);
             }
             .mdc-switch__thumb {
                 &:after {
-                    background-image: url("data:image/svg+xml; utf8, <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 32 32'><path fill='none' stroke='rgb(255,255,255)' stroke-width='6' stroke-miterlimit='10' d='M29 7L11 25l-8-8'/></svg>");
+                    background-image: url("data:image/svg+xml; utf8, <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 32 32'><path fill='none' stroke='rgb(255,255,255)' stroke-width='7' stroke-miterlimit='10' d='M29 7L11 25l-8-8'/></svg>");
                 }
-            }
-            .mdc-switch__thumb-underlay {
-                transform: translateX(1rem);
             }
         }
 
         &:not(.mdc-switch--checked) {
             .mdc-switch__track {
-                background-color: var(--mdc-theme-text-disabled-on-background);
+                background-color: rgb(var(--color-red-default));
             }
             .mdc-switch__thumb {
                 &:after {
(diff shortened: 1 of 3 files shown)


src/components/switch/switch.scss
```diff
@@ -18,8 +18,7 @@
 }
 
 label {
-    line-height: normal;
-
+    // @include mdc-theme.mdc-typography(body2);
```
adrianschmidt, line 20: Forgot to remove, or purposefully left? 🙂

src/components/switch/switch.scss
```diff
@@ -18,8 +18,7 @@
 }
```
adrianschmidt, line 17: > fix(switch): make the structure more like the latest MDC

I'd say this is a refactor. Sure, it could theoretically affect a consumer who's relying on the internal html structure of the component to look a certain way, but they shouldn't be doing that anyway.

I can rename the commit, if you'd like me to?

src/components/switch/partial-styles/_readonly.scss
```diff
@@ -46,13 +46,20 @@
```
adrianschmidt, line 46: There are two additional changes in the readme mode, that we might want to fix while we're at it:
- the font and size of the label has changed
- the "track" has become slightly longer, making the positioning of the checkmark icon a little bit off when the switch is `selected`

v32.4.4
image

v33.0.0
image

adrianschmidt (changes requested):
As per comments above.

src/components/switch/switch.scss
Kiarokh: correct! I'll rename it

src/components/switch/switch.scss
Kiarokh: no just thought you needed to check these commented out stuff later :) 

But now when you asked about it, and reading your comments below, I notice that the label text is not at all similar to other labels, for example the checkbox. Font properties are being inherited from Kompendium instead.

I played around with `@use "@material/typography/mdc-typography";` but couldn't really take any advantage of it. Then I pushed a new fix instead.

Kiarokh: Not sure what's the most optimal way of making label be similar to others, and don't know why it already doesn't. But I think my fix is good enough.

src/components/switch/partial-styles/_readonly.scss
Kiarokh: good catch. I'll fix

src/components/switch/switch.scss
adrianschmidt: Hmm… so you didn't add this line then? I only asked because it looked like this:

image

That's weird… 🤔

src/components/switch/switch.scss
```diff
@@ -18,11 +18,12 @@
...
 
 label {
-    line-height: normal;
-
+    @include lime-typography.typography(body2);
```
adrianschmidt, line 21: I replaced your fixup with this.

`lime-typography` is just `mdc-typography`, but with our preferred defaults already configured.

adrianschmidt: Same goes for `lime-theme` by the way; it's `mdc-theme` with our config applied.

https://github.com/Lundalogik/lime-elements/blob/main/src/style/internal/lime-theme.scss
https://github.com/Lundalogik/lime-elements/blob/main/src/style/internal/lime-typography.scss

adrianschmidt (approved):

PR #1306, opened 2021-07-22
fix(chip-set): make label truncate correctly when there is a leading icon

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -216,6 +216,10 @@ limel-icon.mdc-chip__icon.mdc-chip__icon--leading {
 
 // used only in chipsets that do not have input
 .chip-set__label {
+    @include mixins.truncate-text;
+    width: 120%; // `120%` instead of `100%`,
+    // because this class is always together with `mdc-floating-label--float-above`,
+    // which scales the label down. So there is more horizontal space to display the label in.
     top: functions.pxToRem(13);
     padding-left: functions.pxToRem(20);
 }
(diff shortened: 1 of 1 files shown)


src/components/chip-set/chip-set.scss
```diff
@@ -272,12 +273,16 @@ limel-icon.mdc-chip__icon.mdc-chip__icon--leading {
...
+            padding-left: $leading-icon-space;
         }
 
         .mdc-floating-label {
-            &:not(.lime-floating-label--float-above) {
-                padding-left: functions.pxToRem(40);
+            padding-left: $leading-icon-space;
+            max-width: calc(
+                100% - #{$leading-icon-space}
+            ); // minus space taken by the leading-icon
```
adrianschmidt, line 287: This feels like a very superfluous comment, don't you think? The code literally says "one hundred percent minus leading icon space"… 😄

adrianschmidt (changes requested):
Works great, but I would like to get rid of that comment I pointed out, because I think it confuses more than helps. The code is plenty clear on its own.

src/components/chip-set/chip-set.scss
Kiarokh: 😄 I think I wrote it before I made the variable for it > then thought why not make a variable instead of commenting > then I forgot to remove the comment 

hahah

adrianschmidt: 😄 👌

adrianschmidt (approved):

PR #1308, opened 2021-07-23
fix(list): reduce the dead space around `badgeIcons` in avatar list

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/1213

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/list/partial-styles/_has-grid-layout.scss
```diff
@@ -0,0 +1,25 @@
+:host(.has-grid-layout) {
+    --gap: var(--list-grid-gap, 0.75rem);
+
+    .mdc-deprecated-list {
+        display: grid;
+        grid-gap: var(--gap);
+        grid-template-columns: repeat(
+            auto-fill,
+            minmax(min(100%, max(calc(var(--list-grid-item-max-width, 10rem) - var(--gap)), var(--list-grid-item-min-width, 7.5rem))), 1fr)
+        );
```
Kiarokh, line 10: @adrianschmidt lint messes around this section, and I'll get SASS errors when I build

adrianschmidt: I'll have a look 👍

adrianschmidt: I've pushed a fixup ⚡

Kiarokh:
running `npm run lint:fix` will throw this stupid error:
image

But going to `_has-grid-layout.scss` and pressing save will just rebuild and things work like a charm. I don't know how to get rid of this error. It's a false alarm.

src/components/list/examples/list-grid.tsx
```diff
@@ -0,0 +1,61 @@
+import { ListItem } from '@limetech/lime-elements';
+import { Component, h } from '@stencil/core';
+
+/**
+ * List with grid layout
+ * To display list items in a grid layout instead of a vertical column,
+ * simply add `has-grid-layout` class to
+ * the component.
+ *
+ * This layout can be customized, using a few CSS variables.
```
maltja, line 10: How can the layout be customised? 😛 The CSS variables are described in the menu component, but shouldn't they be described in the list component (here) instead? Maybe describe them here and then mention that they are described here in the menu component? 🤔

src/components/menu/examples/menu-grid.tsx
```diff
@@ -0,0 +1,82 @@
...
+ * two things are required:
+ * 1. Simply setting the `gridLayout` property to `true`, and
+ * 1. Specifying a proper width for the menu surface, using the
+ * `--menu-surface-width` variable.
+ *
+ * :::note
+ * Do not forget that menus should behave responsively, thus using a fixed `width`
+ * should be avoided. To make the width responsive, try using the `min()` function.
+ * This function selects the smallest value from a list of comma-separated expressions
+ * which are placed within the paranthesis.
```
maltja, line 16:
```suggestion
 * which are placed within the parentheses.
```

maltja (commented):
Looks good! Just fix the lint error (I don't either know how to fix it 😛), the commit lint error and take a look at my comment on the documentation.

src/components/list/examples/list-grid.tsx
Kiarokh: > How can the layout be customised?

What should I say instead? :) I think when you tweak the distances or max/min-width, it's "customizing" the layout. Maybe better to say "tweaking" or "adjusting" the layout? 

> but shouldn't they be described in the list component (here) instead?

You're right. But they are already explained _both_ in list and _menu_, because they can be tweaked in both components. But menu only lets users to customize some, coming from the list. For example, `--list-grid-max-width` is only available in list, because it doesn't make sense in menu.

List: 
image

Menu:
image

maltja: I kind of forgot that the styles are explained in the styles section 😅 So then I think it looks good! 🙂

jgroth:
I just had a quick look and I think it would be good if we could make the interface a bit more consistent. At the moment you enable it by setting a class for the list, but you set a prop if  you want to use it for the menu. I think it would be better if we pick one method and use it in both places.

src/components/menu/menu.tsx
```diff
@@ -133,6 +140,10 @@ export class Menu {
                         style={cssProperties}
                     >
                         <limel-list
+                            class={{
```
Kiarokh, line 143: @jgroth said:
I just had a quick look and I think it would be good if we could make the interface a bit more consistent. At the moment you enable it by setting a class for the list, but you set a prop if you want to use it for the menu. I think it would be better if we pick one method and use it in both places.
_____

Thanks Johan

We had a discussion about this subject (using props or classes) here: https://github.com/Lundalogik/lime-elements/pull/1281#discussion_r690332629

I personally prefer props. I feel it's easier for developers to deal with them. But I don't have strong opinion. 

However, here in this PR two things have happened.
One is that in the list, we add a class to render the items in a grid layout. This is to keep it consistent with the other styling classes in list component.

In the example in limel-list, there are two classes (`has-grid-layout` & `has-interactive-items`) used simultaneously but they can also be used individually. 
image

However, in the case of `limel-menu`, I actually need both of these classes to be applied in the same time to avoid the flat look and feel. This is why I used the `prop` instead. I didn't want to leave the responsibility to the consumer and also add extra notes about this that they have to have both classes. 

(note that `has-interactive-items` makes the items look more clickable and more interactive)

Also I don't think it is possible to add a class and affect a nested component in the menu. In the same time, I wanted to use the styling CSS variables of the new grid layout in limel-list, even in the menu, and they must be sent to portal and so forth... So it'd would be impossible to do it without the prop :)

adrianschmidt: @Kiarokh If you prefer props, then use props. You're the one who started using helper-classes, and no one else in the team has ever added any, as far as I know 😄

Kiarokh: 😄 that's right. Well, it was when I knew nothing about TypeScript and how things worked. Helper classes sounded like the easiest and least independent way of doing things. But now I started getting this feeling that props are perhaps easier for our consumers. Still somehow immature thoughts. So if others agree as well, for upcoming ones we can go for props.

src/components/menu/menu.tsx
```diff
@@ -65,6 +66,12 @@ export class Menu {
     @Prop({ reflect: true })
     public badgeIcons = false;
 
+    /**
+     * Renders list items in a grid layout, rather than a vertical list
+     */
+    @Prop({ reflect: true })
+    public gridLayout = false;
```
Kiarokh, line 73: By the way, there is one thing I still want to fix in the grid layout of menus.
It's about the separator lines (`<li class="mdc-deprecated-list-divider" role="separator"></li>`).

I want them to span through an entire row, when menu content is rendered in a grid. Currently, they will be just like others and take a cell.

Not a merge blocker, but if I manage to fix it before merge, I'll add a fixup here.

Kiarokh: done ⚡️ and pushed fixups

PR #1315, opened 2021-08-10
Tabbing in pickers

jensgustafsson:
fix Lundalogik/crm-feature#2268

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -21,6 +21,7 @@ import translate from '../../global/translations';
 import { getHref, getTarget } from '../../util/link-helper';
 
 const SELECTED_CHIP_CLASS = 'mdc-chip--selected';
+const INPUT_FIELD_TABINDEX = 1;
 
 /**
  * @exampleComponent limel-example-chip-set
@@ -356,6 +357,7 @@ export class ChipSet {
             >
                 {this.value.map(this.renderInputChip)}
                 <input
+                    tabIndex={INPUT_FIELD_TABINDEX}
                     type="text"
                     id="input-element"
                     disabled={this.readonly || this.disabled}
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -17,6 +17,8 @@ import {
     ARROW_UP_KEY_CODE,
     ENTER,
     ENTER_KEY_CODE,
+    ESCAPE,
+    ESCAPE_KEY_CODE,
     SPACE,
     SPACE_KEY_CODE,
     TAB,
@@ -236,6 +238,7 @@ export class InputField {
         this.getContainerClassList = this.getContainerClassList.bind(this);
         this.handleCloseMenu = this.handleCloseMenu.bind(this);
         this.setFocus = this.setFocus.bind(this);
+        this.handleKeyDownInDropdown = this.handleKeyDownInDropdown.bind(this);
 
         const debounceTimeout = 300;
         this.changeEmitter = debounce(this.changeEmitter, debounceTimeout);
@@ -308,7 +311,7 @@ export class InputField {
 
         return [
             <label class={this.getContainerClassList()}>
-                <span class="mdc-notched-outline">
+                <span class="mdc-notched-outline" tabindex="-1">
                     <span class="mdc-notched-outline__leading"></span>
                     <span class="mdc-notched-outline__notch">
                         <span class={labelClassList} id={labelId}>
@@ -447,7 +450,7 @@ export class InputField {
         }
 
         return (
-            <div class="mdc-text-field-helper-line">
+            <div tabIndex={-1} class="mdc-text-field-helper-line">
                 {this.renderHelperText()}
                 {this.renderCharacterCounter()}
             </div>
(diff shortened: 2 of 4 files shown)


src/components/picker/picker.tsx
```diff
@@ -195,28 +197,30 @@ export class Picker {
...
-                onStartEdit={this.handleInputFieldFocus}
-                onStopEdit={this.handleStopEditAndBlur}
-                emptyInputOnBlur={false}
-                clearAllButton={this.multiple}
-                {...props}
-            />,
-            this.renderDropdown(),
-        ];
+        return (
+            <Host tabIndex={0}>
```
jensgustafsson, line 201: This makes it possible to tab to this component

Kiarokh: This must be why we get the blue outlines around the component while tabbing, right?

https://user-images.githubusercontent.com/35954987/128882867-92024a0c-0c62-480a-a8fb-76bd7921697a.mov

I'm not sure if it's a good/intended behaviour. I think tabbing should be generally improved for the pickers. Even before your changes, user firs tabs into the chip(s), and then tabs to the input. 

I believe this is the correct order for tab index:
1. **input** should be the first thing users tab into. From here, users should be able to go through the search/suggestions list. But I think that is expected to be done via Enter, then arrow keys up and down. OR(?)
2. then *chips* individually (where they can also navigate left and right with keyboard, or press backspace to delete one.


And the main container component shouldn't receive focus individually (the blue outlines), because users cannot do anything there while the blue lines are shown anyway. They must tab once more to go deeper.

Kiarokh: The worst part is, when user actually has focused on the input field (when the green outlines are shown), they can't do anything by pressing the enter key. They cannot for example activate the search results or suggestions list.

jensgustafsson: > This must be why we get the blue outlines around the component while tabbing, right?

Exactly!

jensgustafsson: Solved ⚡

src/components/input-field/input-field.tsx
```diff
@@ -17,6 +17,7 @@ import {
     ARROW_UP_KEY_CODE,
     ENTER,
     ENTER_KEY_CODE,
+    ESCAPE,
```
Kiarokh, line 20: Don't you need to add this as well? `ESCAPE_KEY_CODE,`

jensgustafsson: It's deprecated so I don't think so but we're doing it everywhere else so perhaps I should do it for that reason... :)

src/components/picker/picker.tsx
jensgustafsson: Yes! And can't find a good solution to that. I'm leaning towards creating a new issue for that. I would argue its still better than before. Right? 🙈

Kiarokh: Yes, the whole is better than before I'd say. :) 
But it's important to make an issue from my comment above, so we don't forget it. 

They can still click outside to close that list, but I don't know how annoying it is for people who are dependent on keyboard navigation.

src/components/input-field/input-field.tsx
Kiarokh: 😄 I have no clue what these are, but I just saw that it's used everywhere else as you say. Let's maybe keep it consistent, and in another issue remove all deprecated stuff?

src/components/picker/picker.tsx
```diff
@@ -195,28 +196,30 @@ export class Picker {
...
-                onStartEdit={this.handleInputFieldFocus}
-                onStopEdit={this.handleStopEditAndBlur}
-                emptyInputOnBlur={false}
-                clearAllButton={this.multiple}
-                {...props}
-            />,
-            this.renderDropdown(),
-        ];
+        return (
+            <Host>
```
Kiarokh, line 202: Is this commit still necessary? Things seem to be working fine without it as well… But I don't know if this makes things better than before or what :)

Kiarokh (approved):
I think the component works nicely both in Elements and Webclient. Just don't know if the first commit is needed.

src/components/picker/picker.tsx
jensgustafsson: You're right! Commit now longer needed. I prefer this syntax but since it's not needed anymore I will just remove it.

Less is more :) Care to approve again? 😄

src/components/input-field/input-field.tsx
jensgustafsson: Added keycode as well ⚡

Kiarokh (dismissed):

Kiarokh:
I did approve again, didn't I? :) 

I do it again then

Kiarokh (approved):

src/components/picker/picker.tsx
adrianschmidt: Was this resolved? If not, should we create an issue, or just react if we hear complaints about it?

Kiarokh: yes it was fixed ;)

src/components/chip-set/chip-set.tsx
```diff
@@ -21,6 +21,7 @@ import translate from '../../global/translations';
 import { getHref, getTarget } from '../../util/link-helper';
 
 const SELECTED_CHIP_CLASS = 'mdc-chip--selected';
+const INPUT_FIELD_TABINDEX = 1;
```
adrianschmidt, line 24: @jensgustafsson @Kiarokh I came across this when doing other stuff in this file, and thought I'd take a moment to spread some knowledge 😊

From [MDN's docs for tabindex](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex):

image

However, I tested this in the web client, and it didn't seem to have that effect, so it's possible that the above doesn't apply between different shadow roots… 🤔

Anyway, it's probably for the best not to use `tabindex=1`… 🙂

PR #1316, opened 2021-08-11
feat(collapsible-section): enable keyboard navigation and interaction

Kiarokh:
Users now can tab and select section headers,
and also press the ENTER key to open or closes a section

fix: https://github.com/Lundalogik/crm-feature/issues/1972

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/collapsible-section/collapsible-section.tsx
+++ b/src/components/collapsible-section/collapsible-section.tsx
@@ -1,6 +1,14 @@
-import { Component, Event, EventEmitter, h, Prop } from '@stencil/core';
+import {
+    Component,
+    Event,
+    EventEmitter,
+    h,
+    Prop,
+    Element,
+} from '@stencil/core';
 import { dispatchResizeEvent } from '../../util/dispatch-resize-event';
 import { Action } from './action';
+import { ENTER, ENTER_KEY_CODE } from '../../util/keycodes';
 
 /**
  * @slot - Content to put inside the collapsible section
@@ -52,6 +60,9 @@ export class CollapsibleSection {
     @Event()
     private action: EventEmitter<Action>;
 
+    @Element()
+    private host: HTMLLimelCollapsibleSectionElement;
+
     constructor() {
         this.onClick = this.onClick.bind(this);
         this.renderActionButton = this.renderActionButton.bind(this);
@@ -60,7 +71,12 @@ export class CollapsibleSection {
     public render() {
         return (
             <section class={`${this.isOpen ? 'open' : ''}`}>
-                <header class="section__header" onClick={this.onClick}>
+                <header
+                    class="section__header"
+                    onClick={this.onClick}
+                    onKeyDown={this.handleKeyDown}
+                    tabindex="0"
+                >
                     <div class="section__header__expand-icon">
                         <div class="expand-icon__line"></div>
                         <div class="expand-icon__line"></div>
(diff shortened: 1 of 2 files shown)


src/components/collapsible-section/collapsible-section.tsx
```diff
@@ -92,6 +99,16 @@ export class CollapsibleSection {
         }
     }
 
+    private handleKeyDown(event: KeyboardEvent) {
```
maltja, line 102: Use this syntax instead and then remove the binding in the constructor 🙂 

```suggestion
    private handleKeyDown = (event: KeyboardEvent) => {
```

maltja (approved):
wow the experience is so much better in the web client when the collapsible sections on the card can be opened with enter 🎉 The only thing I notice is that if the collapsible list has actions then the actions can't be triggered with enter anymore but I think that can be fixed in another issue. What do you think? 🙂

src/components/collapsible-section/collapsible-section.tsx
Kiarokh: In this case I prefer to change all of the constructor bindings to this. Is it OK? Or we do it for all components in one big PR? :)

Kiarokh:
> if the collapsible list has actions then the actions can't be triggered with enter anymore but I think that can be fixed in another issue. What do you think?

Nice finding! 

I don't know. Maybe we should already fix it here, instead of introducing a new bug. How can we do it do you think?

src/components/collapsible-section/collapsible-section.tsx
maltja: I don't know how we should do it but in this file the "new" syntax is already used for one function `handleActionClick`. 🙂

maltja:
> > if the collapsible list has actions then the actions can't be triggered with enter anymore but I think that can be fixed in another issue. What do you think?
> 
> Nice finding!
> 
> I don't know. Maybe we should already fix it here, instead of introducing a new bug. How can we do it do you think?

Maybe check what kind of element is clicked, do you want me to try to fix it? 🙂

src/components/collapsible-section/collapsible-section.tsx
Kiarokh: Ah OK then. I'll fix, or you do a fixup if you want. :)

Kiarokh:
Yes please try to fix it. I have no clue how to do. :) I'm just copy/pasting stuff 😄

src/components/collapsible-section/collapsible-section.tsx
maltja: I can fix it :)

maltja (approved):

src/components/collapsible-section/collapsible-section.tsx
```diff
@@ -92,6 +108,18 @@ export class CollapsibleSection {
...
 
+    private handleKeyDown = (event: KeyboardEvent) => {
+        const isEnter = event.key === ENTER || event.keyCode === ENTER_KEY_CODE;
+
+        if (isEnter) {
+            event.stopPropagation();
+            event.preventDefault();
+            const element = (this.host.shadowRoot.activeElement ||
+                document.activeElement) as HTMLElement;
+            element.click();
```
adrianschmidt, line 119: @Kiarokh @jensgustafsson For next time, I think it would be better to refactor out the stuff that happens in `onClick` into a common function that could be called by both `onClick` and `handleKeyDown`, or even just outright call `onClick`, rather than synthesising a click-event through the DOM just to make the same thing happen.

I'm working on this component at the moment, so I might clean it up while I'm at it. Unless you tried that way already and there was some specific reason it didn't work?

Kiarokh: I just did what I was told 😄 I think you can improve it the way you believe is better.

adrianschmidt: Fixed in #1451.

PR #1319, opened 2021-08-19
feat(table): be able to set the page number in a table

maltja:
fix Lundalogik/crm-feature#2197

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/table/table.tsx
```diff
@@ -172,6 +173,19 @@ export class Table {
         this.updateMaxPage();
     }
 
+    @Watch('page')
+    public pageChanged() {
+        if (!this.tabulator) {
```
Befkadu1, line 178: Just a small question. When will `this.tabulator` becomes undefined?

maltja: I'm not sure, but since it is tested in the other watch methods or `this.tabulator?` was used, I thought it was best to test it 😛

adrianschmidt: I don't think it ever _becomes_ undefined, but it starts out as undefined, and is only defined after the component has loaded and rendered, in `initTabulatorComponent` on line 268. The `page` watcher will almost certainly have fired at least once by then, as I think these watchers always fire when the component loads. Without this check, line 182 below would cause an error.

Befkadu1: Ok, thanks for the explanation.  In the component lifecycle methods I know that the watcher is fired before the render method. For some reason without this check, line 182 didn't cause an error. The code in the watcher loads when `page` changes right?.

adrianschmidt: Hmm… then I guess the watcher didn't fire when the component loaded. Interesting. Did you give the `page` property a value from the beginning? If not, that could be the reason, but it might also be that the watcher simply doesn't fire until the initial value is changed.

Regardless, having the check is a very small cost for avoiding potential errors, so I think we should keep it.

Befkadu1 (approved):

src/components/table/table.tsx
Befkadu1: Can you merge this PR then. I approved it. It works perfect as it should

PR #1325, opened 2021-08-25
Align numerical data

Befkadu1:
fix: Lundalogik/crm-feature#2123
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/columns.ts
+++ b/src/components/table/columns.ts
@@ -19,6 +19,7 @@ export class ColumnDefinitionFactory {
         const definition: Tabulator.ColumnDefinition = {
             title: column.title,
             field: column.field,
+            hozAlign: column.horizontalAlign,
         };
 
         if (column.headerComponent) {
--- a/src/components/table/table.types.ts
+++ b/src/components/table/table.types.ts
@@ -28,6 +28,11 @@ export interface Column<T extends object = any> {
      * A component used to render inside the column header
      */
     headerComponent?: TableComponentDefinition;
+
+    /**
+     * Sets the horizontal text alignment for the column
+     */
+    horizontalAlign?: 'left' | 'center' | 'right';
 }
 
 export type TableFormatter = (value: any, data?: object) => string;
--- a/src/components/table/examples/table-activate-row.tsx
+++ b/src/components/table/examples/table-activate-row.tsx
@@ -19,9 +19,9 @@ export class TableExampleActivateRow {
     private tableData: Person[] = persons;
     private columns: Column[] = [
         { title: 'Name', field: 'name' },
-        { title: 'Age', field: 'age' },
+        { title: 'Age', field: 'age', horizontalAlign: 'right' },
         { title: 'Kind', field: 'kind' },
-        { title: 'Height', field: 'height' },
+        { title: 'Height', field: 'height', horizontalAlign: 'right' },
         { title: 'Stamina', field: 'stamina' },
         { title: 'Place of Birth', field: 'placeOfBirth' },
         { title: 'Sign', field: 'sign' },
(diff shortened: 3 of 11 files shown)


src/components/table/table.types.ts
```diff
@@ -28,6 +29,11 @@ export interface Column<T extends object = any> {
      * A component used to render inside the column header
      */
     headerComponent?: TableComponentDefinition;
+
+    /**
+     * Sets the horizontal text alignment for the column (left|center|right)
+     */
+    hozAlign?: Tabulator.ColumnDefinitionAlign;
```
adrianschmidt, line 36: Kind of nitpicky, but I don't really like the shortening `hozAlign`. (I understand it comes from Tabulator.) I think I'd like it to be either `horizontalAlign` or, if we want to go short, just `hAlign`… 🤔

@jgroth Thoughts?

src/components/table/columns.ts
```diff
@@ -19,6 +19,7 @@ export class ColumnDefinitionFactory {
         const definition: Tabulator.ColumnDefinition = {
             title: column.title,
             field: column.field,
+            hozAlign: column.hozAlign,
```
adrianschmidt, line 22: I think the first two commits can be squashed together.

src/components/table/table.types.ts
Befkadu1: I agree with you. I was about to write it like that at the beginning. When I updated the examples I had to use `hozAlign` . For example `{ title: 'Height', field: 'height', hozAlign: 'right' },`. I thought of consistency that is the reason I used same name everywhere

adrianschmidt: Hmm… ok. I'll have a quick look at it, and if I can't figure out how to make it work, we'll just go with `hozAlign` 🙂

jgroth: Yeah I don't like `hozAlign` either. I cannot see why we would not be able to use `horizontalAlign` :thinking:

jgroth: Also, we should not expose types from Tabulator like this. If they make changes to it we might make breaking changes in our own API without realizing it.

adrianschmidt: I pushed fixups for making it `hAlign`, but it's easy to go from there to `horizontalAlign`.

adrianschmidt: I changed it to `horizontalAlign`.

Befkadu1: > Also, we should not expose types from Tabulator like this. If they make changes to it we might make breaking changes in our own API without realizing it.

What do you think if we have our columnDefinationAlign then?
Like this `type ColumnDefinitionAlign = 'left' | 'center' | 'right';`

jgroth: I think we can just inline it for now. Or do you think it's useful to have a separate type for it?

adrianschmidt: Yeah, I tried with this:

```ts
    /**
     * Sets the horizontal text alignment for the column
     */
    horizontalAlign?: 'left' | 'center' | 'right';
```

and it looks like this:

image

👌

Befkadu1: We can squash those changes and merge.

adrianschmidt (approved):
I changed the scope of the commit from `column` to `table`, as that will make more sense to someone reading the changelog. I also updated the commit message slightly.

Good work 👍

PR #1328, opened 2021-08-30
fix(table): don't scroll table to the left after each reload

jensgustafsson:
fix: Lundalogik/crm-feature#2027

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -213,7 +213,7 @@ export class Table {
                 return;
             }
 
-            this.tabulator.setData(this.data);
+            this.tabulator.replaceData(this.data);
         });
     }
 
@@ -358,7 +358,7 @@ export class Table {
         if (abortRequest) {
             setTimeout(() => {
                 this.updateMaxPage();
-                this.tabulator.setData(this.data);
+                this.tabulator.replaceData(this.data);
             });
 
             return false;
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -351,6 +351,20 @@ export class Table {
         };
     }
 
+    /*
+     * The ajaxRequesting callback is triggered when ever an ajax request is made.
+     *
+     * Tabulator is requesting data with an AJAX request even though it has been
+     * given data when it was created.
+     *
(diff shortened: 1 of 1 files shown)


jgroth:
`setData` seems to be used here as well, should we replace it with `replaceData`?

https://github.com/Lundalogik/lime-elements/blob/5000fa40642bd8516aa6d04dd56fe510f92d7ba0/src/components/table/table.tsx#L354-L368

hannahu:
> `setData` seems to be used here as well, should we replace it with `replaceData`?
> 
> https://github.com/Lundalogik/lime-elements/blob/5000fa40642bd8516aa6d04dd56fe510f92d7ba0/src/components/table/table.tsx#L354-L368

Saw it as well but thought we could keep it as it was but maybe we should use the same on both places.

jensgustafsson:
> > `setData` seems to be used here as well, should we replace it with `replaceData`?
> > https://github.com/Lundalogik/lime-elements/blob/5000fa40642bd8516aa6d04dd56fe510f92d7ba0/src/components/table/table.tsx#L354-L368
> 
> Saw it as well but thought we could keep it as it was but maybe we should use the same on both places.


True. I thought it made sense to use `setData` here but I could look in to it once more and see if it makes sense to use it hear as well

jensgustafsson:
> > > `setData` seems to be used here as well, should we replace it with `replaceData`?
> > > https://github.com/Lundalogik/lime-elements/blob/5000fa40642bd8516aa6d04dd56fe510f92d7ba0/src/components/table/table.tsx#L354-L368
> > 
> > 
> > Saw it as well but thought we could keep it as it was but maybe we should use the same on both places.
> 
> True. I thought it made sense to use `setData` here but I could look in to it once more and see if it makes sense to use it hear as well

`replaceData` seems to work equally well so I changed to that

hannahu (approved):

PR #1334, opened 2021-09-01
Merge commits from `next` to `main`

adrianschmidt:


src/components/popover/popover.tsx
```diff
@@ -155,4 +159,14 @@ export class Popover {
 
         return zipObject(propertyNames, values);
     }
+
+    private handleGlobalKeyPress = (event: KeyboardEvent) => {
+        if (event.key !== ESCAPE) {
+            return;
+        }
+
+        event.stopPropagation();
```
adrianschmidt, line 168: Nitpick: Since we're registering this event on the `document`, and aren't registering it in the `capture` phase, I don't think there's any propagation left to be done after this triggers… @jgroth

jgroth: That is probably true :smile:

jgroth: Perhaps if there are several listeners on document it won't call those listeners after this one?

adrianschmidt: I think `stopImmediatePropagation` is needed to do that. Nothing we need to change anyway, just wanted to mention for knowledge sharing (and for the chance to learn myself, in case I was mistaken, of course!) 🙂

PR #1345, opened 2021-09-08
chore(style): remove unused mixin for `label--float-above`

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/2295

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -288,10 +288,11 @@ limel-icon.mdc-chip__icon.mdc-chip__icon--leading {
         }
 
         .mdc-floating-label {
-            padding-left: $leading-icon-space;
-            max-width: calc(100% - #{$leading-icon-space});
-            &.lime-floating-label--float-above {
-                padding-left: 0;
+            &:not(.lime-floating-label--float-above) {
+                left: $leading-icon-space;
+            }
+            &.mdc-floating-label--float-above {
+                left: functions.pxToRem(4);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


Kiarokh:
As mentioned here: https://github.com/Lundalogik/lime-elements/blob/4891e9bb4492286c4be260abe9572d14f3097eab/src/components/chip-set/chip-set.scss#L234

all of these UI problems (and a few more [see below 👇 ]) happen because MDC scripts remove their styling classes. That's why we made similar classes with duplicated styles to use in chip-set. A more sustainable approach would be finding a way to force these classes to stay. But not sure if it's possible.


As an example, see how the `mdc-notched-outline__notch` div changes its size when field is not focused anymore. That's because a dynamically calculated inline style of `width` gets applied to this div when the field is focused and the label floats above. Some script calculates how wide the floated and shrunk label should be and applies that as an inline style to the container div of the label which is `mdc-notched-outline__notch`.
https://user-images.githubusercontent.com/35954987/132499882-69484f9b-7c64-4fbd-8207-60e22cfd31fc.mov

Rest of our styles work because we put them on our own classes, but the inline `width` doesn't change, because we cannot control it.

jgroth (approved):

PR #1350, opened 2021-09-13
feat(picker): make delimiter of chip-set available in picker

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/2103
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -107,6 +107,12 @@ export class Picker {
     public multiple: boolean = false;
 
     /**
+     * Sets delimiters between chips.
+     */
+    @Prop({ reflect: true })
+    public delimiter: string = null;
+
+    /**
      * True if the dropdown list should be displayed without cutting the content
      *
      * @deprecated This was used for a workaround, and isn't needed any
@@ -237,6 +243,7 @@ export class Picker {
                 leadingIcon={this.leadingIcon}
                 value={this.chips}
                 disabled={this.disabled}
+                delimiter={this.delimiter}
                 readonly={this.readonly}
                 required={this.required}
                 searchLabel={this.searchLabel}


src/components/picker/picker.tsx
```diff
@@ -106,6 +106,12 @@ export class Picker {
     @Prop()
     public multiple: boolean = false;
 
+    /**
+     * Sets delimiters between chips.
+     */
+    @Prop({ reflect: true })
```
Befkadu1, line 112: I am wondering why `reflect` set true? When a `prop` is set to `reflect`, it means that their value will be rendered in the DOM as an HTML attribute. It will be like this 
Screenshot 2021-09-14 at 08 59 50

Kiarokh: it didn't work without setting it to true :/

Befkadu1: I think this should work. 
```suggestion
    @Prop()
```
I tested it and works as it should. I also saw same setting in `ChipSet` component

adrianschmidt: Simple props (strings, booleans, etc) should generally be reflected to the attribute, regardless of whether or not it's strictly needed for our own functionality to work.

https://developers.google.com/web/fundamentals/web-components/best-practices#aim-to-keep-primitive-data-attributes-and-properties-in-sync,-reflecting-from-property-to-attribute,-and-vice-versa.

> **Aim to keep primitive data attributes and properties in sync, reflecting from property to attribute, and vice versa.**
> Why? - You never know how a user will interact with your element. They might set a property in JavaScript, and then expect to read that value using an API like `getAttribute()`. If every attribute has a corresponding property, and both of them reflect, it will make it easier for users to work with your element. In other words, calling `setAttribute('foo', value)` should also set a corresponding foo property and vice versa. There are, of course, exceptions to this rule. You shouldn't reflect high frequency properties, e.g. currentTime in a video player. Use your best judgment. If it seems like a user will interact with a property or attribute, and it's not burdensome to reflect it, then do so.

Tangential: We should not reflect objects and arrays.

> **Aim to only accept rich data (objects, arrays) as properties.**
> Why? - Generally speaking, there are no examples of built-in HTML elements that accept rich data (plain JavaScript objects and arrays) through their attributes. Rich data is instead accepted either through method calls or properties. There are a couple obvious downsides to accepting rich data as attributes: it can be expensive to serialize a large object to a string, and any object references will be lost in this stringification process. For example, if you stringify an object which has a reference to another object, or perhaps a DOM node, those references will be lost.

With Stencil, we still write the jsx code as if we sent in objects and arrays via attributes, but I think Stencil (or jsx) deals with that so we don't really have to care. But reflecting them back to attributes would mean serialising the values every time their updated, which is completely unnecessary and just wastes processing power (a lot more than reflecting primitive values does).

Befkadu1 (approved):

adrianschmidt (changes requested):
Hold on

src/components/chip-set/chip-set.tsx
```diff
@@ -118,7 +118,7 @@ export class ChipSet {
     /**
      * For chip-set of type `input`. Sets delimiters between chips.
      */
-    @Prop({ reflect: true })
+    @Prop()
```
adrianschmidt, line 121: This is a breaking change. Please remove this commit.

src/components/picker/picker.tsx
```diff
@@ -106,6 +106,12 @@ export class Picker {
     @Prop()
     public multiple: boolean = false;
 
+    /**
+     * Sets delimiters between chips.
+     */
+    @Prop()
```
adrianschmidt, line 112:
```suggestion
    @Prop({ reflect: true })
```

`fixup! feat(picker): make delimiter of chip-set available in picker`

src/components/chip-set/chip-set.tsx
Befkadu1: ok, I will remove it

Befkadu1 (approved):

adrianschmidt (approved):

PR #1354, opened 2021-09-16
fix(table): don't show empty message when loading the table

maltja:
We could use tabulator's placeholder option and send in a DOM node instead of a string. However, I
think it is an easier solution not to use the placeholder option  and instead just render the empty
message in the render method.

fix Lundalogik/crm-feature#2326

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -315,7 +315,6 @@ export class Table {
             dataFiltered: this.updateMaxPage,
             nestedFieldSeparator: false,
             ...columnOptions,
-            placeholder: this.emptyMessage,
         };
     }
 
@@ -521,6 +520,7 @@ export class Table {
                 >
                     <limel-spinner size="large" />
                 </div>
+                {this.renderEmptyMessage()}
                 <div
                     id="tabulator-table"
                     class={{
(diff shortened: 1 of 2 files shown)


src/components/table/table.tsx
```diff
@@ -532,4 +532,18 @@ export class Table {
             </div>
         );
     }
+
+    private renderEmptyMessage() {
+        const showEmptyMessage =
```
maltja, line 537: When we render the spinner we toggle the style `display` instead of not rendering it. That is why I choose to do the same here. 

The comment above the rendering of the spinner: 
```
{/* Toggle style instead of removing the loader
    because removing the element will cause a rerender, breaking the
    tabulator table */}
```

Kiarokh (approved):
It looks good to me and works in the client when I test. 
I also added a couple of new commits to clean up the CSS. :)

I approve. However, a second opinion would be nice too; since I also added stuff.

Befkadu1 (approved):
It looks good

Issue #1348, opened 2021-09-09
Field loses focus when validation is performed


mjlime:
## Current behavior

![field-loose-focus-after-validation](https://user-images.githubusercontent.com/42939122/132688965-fff1f389-de1a-435e-8231-f9a5aca4e825.gif)

1. Create a schema with a required decimal field.
2. Start with the field empty.
3. Put focus on the field and start typing 1 digit.
4. The field now loses focus when the validation is performed and the error is removed.

## Expected behavior

The focus should still be on the field with the cursor at the same place so I can continue typing seamlessly.

## Environment

- **lime-elements** version: 33.11.0-next.1
- Framework used: 
- Logs:

adrianschmidt:
This is reproducible in the docs for v33.11.0 too. I'd say this is a pretty serious bug.

Thanks for the report!

jensgustafsson:
I could reproduce this with lime-elements v32 as well in the webclient.

adrianschmidt:
> I could reproduce this with lime-elements v32 as well in the webclient.

I've tested a little, and yes, the bug is similar, but not identical in v32.4.4 and the latest version. In v32.4.4, the focus is only lost when the field changes from valid to invalid, whereas in v33.12.0, focus is lost when the validity of the field changes, regardless of whether it's _valid → invalid_ or _invalid → valid_.

I did previously think that the focus problem also applied to non-numeric fields in v33+, but that doesn't seem to be the case. I'm not sure if something's changed, or if I was just confused to begin with 😅

However, the numeric input field in v32.4.4, when configured to format the input, shows the formatted input all the time, instead of showing the actual input when the field is focused. This creates a lag in the input, and other annoying bugs. So, while I think we want to fix the focus bug, because it's really annoying, I don't think it has to stop us from releasing v33+, even if we haven't fixed the bug yet.

@jensgustafsson

jensgustafsson:
Ping @blombergedvin

Some nice news :)

PR #1355, opened 2021-09-20
fix(input-field): retain focus when numeric field is validated

adrianschmidt:
fix: #1348 

Note that there are two refactors, and one bug fix. I would suggest reviewing commit by commit 😄

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


adrianschmidt:
This kinda did my head in for a few days, so when I finally found the problem and solution, I was very relieved and happy to open a PR…

…which means I might have missed stuff because I was too happy to really look deeply for issues 😬

So please assume I've missed something and test a little extra 😝

src/components/input-field/input-field.scss
```diff
@@ -85,9 +85,13 @@
     width: calc(100% - #{functions.pxToRem(20)});
     z-index: z-index.$input-field--formatted-value;
 
-    display: none;
+    position: absolute;
+    top: 0.9rem;
+    pointer-events: none;
+
+    opacity: 0;
```
adrianschmidt, line 92: Note to self (for after hackday): This should probably have a no-speak attribute too, because a screenreader should only need to see the actual input and the value present there.

jensgustafsson (approved):

PR #1361, opened 2021-09-27
fix(list): set correct selected index on mdc-list

maltja:
If an item is selected and the list is then updated or reordered, the new selected item would be the
previous selected item's index even if the selected property on an item was updated. Mdc-list has a
selextedIndex property to keep track of the selected items. This change updates the selectedIndex so
that the selected items are visualized correctly.

fix Lundalogik/crm-feature#2259

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


jgroth (approved):

jgroth:
> Will this fix fix [this issue](https://github.com/Lundalogik/crm-feature/issues/1954) (and similar ones)? :D

It might, you should try it! :sweat_smile:

PR #1366, opened 2021-09-29
feat(circular-progress): add new `limel-circular-progress` component

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/2338

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/circular-progress/circular-progress.scss
+++ b/src/components/circular-progress/circular-progress.scss
@@ -1,0 +1,86 @@
+@use '../../style/functions';
+@use '../../style/internal/lime-theme';
+@use '../../style/internal/z-index';
+
+/**
+ * @prop --circular-progress-size: Determines the visual size of the visualization. This does not override the `size` property if it is specified.
+ * @prop --circular-progress-track-color: Determines the color of the circular track. Defaults to `--contrast-400`.
+ * @prop --circular-progress-fill-color: Determines the color of the progressed section. Defaults to `--lime-primary-color`.
+ */
+
+:host {
+    display: block;
+    box-sizing: border-box;
+}
+
+:host([size='x-small']) {
+    --circular-progress-size: 1.5rem;
+    font-weight: bold;
+}
+
+:host([size='small']) {
+    --circular-progress-size: 2rem;
+    font-weight: bold;
+}
+
+:host([size='medium']) {
+    --circular-progress-size: 3rem;
+}
+
+:host([size='large']) {
+    --circular-progress-size: 4rem;
+}
+
+:host([size='x-large']) {
+    --circular-progress-size: 5rem;
+}
+
+.lime-circular-progress {
+    --size: var(--circular-progress-size, 3rem);
+    --fill-color: var(--circular-progress-fill-color, var(--mdc-theme-primary));
+    --track-color: var(
+        --circular-progress-track-color,
+        rgb(var(--contrast-400))
+    );
+
+    display: flex;
+    align-items: center;
+    justify-content: center;
+
+    width: var(--size);
+    height: var(--size);
+    border-radius: 50%;
+
+    box-shadow: 0 0 0 0.125rem rgba(var(--contrast-100), 0.8);
+    background: conic-gradient(
+        var(--fill-color) 0% var(--percentage),
+        var(--track-color) var(--percentage) 100%
+    );
+
+    &:before {
+        $size-of-middle-circle: calc(
+            var(--size) * 0.75 + 0.25rem
+        ); // this calculation ensures that the middle circle is responsive and always perfectly centered
+
+        content: '';
+        position: absolute;
+        width: $size-of-middle-circle;
+        height: $size-of-middle-circle;
+        border-radius: 50%;
+        background-color: rgb(var(--contrast-100));
+        box-shadow: var(--button-shadow-pressed);
+    }
+}
+
+.value {
+    font-size: calc(var(--size) * 0.25);
+    color: rgb(var(--contrast-1200));
+    z-index: z-index.$limel-circular-progress-value;
+    cursor: default;
+}
+
+.suffix {
+    opacity: 0.7;
+}
+
+@import './partial-styles/_percentage-colors';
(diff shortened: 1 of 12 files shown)


Kiarokh:
Feel free to correct the text, or improve the code if you review this while I'm gone! Don't wait for me to come back and fix it :)

src/components/circular-progress/examples/circular-progress-props.tsx
```diff
@@ -0,0 +1,48 @@
+import { Component, h } from '@stencil/core';
+/**
+ * Using the props
+ * This component is initially designed to visualize a percentage on a scale of
+ * zero to 100. However, you can easily visualize a progress in other scales,
+ * simply by setting `maxValue` and `suffix`.
+ *
+ * Look at this example to see how the component displays an angel in a
```
adrianschmidt, line 8:
```suggestion
 * Look at this example to see how the component displays an angle in a
```

src/components/circular-progress/circular-progress.tsx
```diff
@@ -0,0 +1,82 @@
...
+    public size: CircularProgressSize;
+
+    public render() {
+        const classList = {
+            'lime-circular-progress': true,
+            'displays-percentage-colors': this.displayPercentageColors,
+        };
+
+        const currentPercentage = (this.value * 100) / this.maxValue + '%'; // eslint-disable-line no-magic-numbers
+
```
Befkadu1, line 65: declare this 
`const value = (Math.round(this.value * 10)/10);` 

This will round it to a max of 1 decimal place. For example `6.28` becomes `6.3`

Then use this ` {abbreviate(value)}` on line 77 instead. This will shorten large number for example 
`1234` becomes `1k`

src/components/circular-progress/examples/circular-progress-percentage-colors.tsx
```diff
@@ -0,0 +1,40 @@
...
+export class CircularProgressPercentageColorsExample {
+    @State()
+    private value = 5;
+
+    public render() {
+        return [
+            <limel-input-field
+                label="Value"
+                type="number"
+                value={this.value.toFixed(0)}
```
Befkadu1, line 27:
```suggestion
                value={`${this.value}`}
```

or just declare a constant under the render method
```
const value=`${this.value}`

File: src/components/circular-progress/examples/circular-progress-props.tsx (hunk 1 above)
Comment (Befkadu1) [inline src/components/circular-progress/examples/circular-progress-props.tsx]:
```suggestion
 * Look at this example to see how the component displays an angle in a
```

src/components/circular-progress/circular-progress.tsx
```diff
@@ -0,0 +1,82 @@
+import { Component, h, Prop } from '@stencil/core';
+import { CircularProgressSize } from './circular-progress.types';
+// import { CircularProgressSize } from '@limetech/lime-elements'; // Kia: I'd like something like this, but it won't work :/
```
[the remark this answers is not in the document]
Kiarokh: yes sir

src/components/circular-progress/circular-progress.scss
```diff
@@ -0,0 +1,80 @@
...
+    color: rgb(var(--contrast-1200));
+    z-index: 1;
+    cursor: default;
+}
+
+.suffix {
+    opacity: 0.7;
+}
+
+@import './partial-styles/_percentage-colors';
```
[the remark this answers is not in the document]
Kiarokh: I think since these are `partial-styles` putting them at the bottom is more correct. Because usually these may override something which is before them. And generally they are about details, which should be further down when you have defined more universal styles. Also it helps when reusing sass variables which are defined before the import in the main file (if I remember correctly).

src/components/circular-progress/circular-progress.tsx
Kiarokh: awesome!

src/components/circular-progress/circular-progress.tsx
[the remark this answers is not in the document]
adrianschmidt: It's not an export, it's an import. It's the same import as the line above, but I think Kia means that he would like to use the alias `'@limetech/lime-elements'` instead of the path `'./circular-progress.types'`.

The alias way for importing works sometimes, but when the _Types_ section of the docs broke recently, it was because a type was being imported using `from '@limetech/lime-elements'`, so I think it might be best to avoid it, at least for types.

src/components/circular-progress/examples/circular-progress-props.tsx
adrianschmidt: @Kiarokh Didn't I make this exact comment? 😄

Maybe you shouldn't resolve comments before you've actually fixed them? 😝

adrianschmidt: …or maybe you just hadn't pushed the fix I guess… 🙂

Befkadu1: I did not see you commented before me. I started reviewing and it took me a while to publish my comment.

adrianschmidt: Ah! 🤦‍♂️ Of course, that makes sense 👍😄

src/components/circular-progress/circular-progress.tsx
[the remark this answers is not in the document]
Befkadu1: I know what he meant. If you want to use like this `'@limetech/lime-elements'`, you need to export `circular-progress.types` file `export * from './components/circular-progress/circular-progress.types';` in `interface.d.ts` file. That makes it available for the customer to use the types as well. I was just explaining how and when we use it.

Befkadu1 (dismissed):

Befkadu1 (approved):

src/components/circular-progress/circular-progress.tsx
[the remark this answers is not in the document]
adrianschmidt: Ah! Now I understand what you meant! 👍

PR #1367, opened 2021-09-29
perf(linear-progress): remove `buffer-dots` and thus their always…

Kiarokh:
… running animations

fix: https://github.com/Lundalogik/crm-feature/issues/2346

The `mdc-linear-progress__buffer-dots` is from a feature that
we do not use. However, their div was present, and their animation
was always running, affecting the performance.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


adrianschmidt (approved):
👌

PR #1371, opened 2021-09-29
fix(list): only apply badge effect to graphic

### LINKED ISSUE #1370: Action menu icon incorrectly gets badge icon
## Current behavior

Using the list component with an action menu and turning on badge icons also renders the action menu icon with a badge.

Screen Shot 2021-09-29 at 19 23 45

The margin to the list item text is also a bit messed up.
Screen Shot 2021-09-29 at 19 29 30

This selector seems to be the issue:
https://github.com/Lundalogik/lime-elements/blob/91998a83ed3ff47728f2ae02a3b734545372eaa4/src/components/list/list.scss#L70-L81

## Expected behavior

Using `badgeIcons` should only change the look of the list item icon, not the action menu icon.

Screen Shot 2021-09-29 at 19 23 58

## Environment

- **lime-elements** version: 33.12 
- Framework used: 
- Logs:

anderssonjohan:
fixes #1370, Lundalogik/limepkg-workorder-list#38

Tested with modified "list with actions" example to get the combination action menu + badge icons.

```typescript
    public render() {
        return (
            <limel-list
                badgeIcons={true}
                iconSize="large"
                items={this.items}
                onSelect={this.onSelectAction}
            />
        );
    }
```

**Before**

Screen Shot 2021-09-29 at 19 44 07

**After**

Screen Shot 2021-09-29 at 19 45 00

Also, margins of the action menu icon are unaffected even when using something different than the default icon size.

Screen Shot 2021-09-29 at 19 48 39
Screen Shot 2021-09-29 at 19 45 58

UPDATE:
Also makes sure that action menu is vertically centered (ty @Kiarokh @mjlime).

Before and after:
Screen Shot 2021-09-30 at 11 32 55

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/list/list.scss
+++ b/src/components/list/list.scss
@@ -70,7 +70,7 @@ $list-border-radius: 0.375rem; // 6px
     &.mdc-deprecated-list--avatar-list {
         position: relative;
 
-        limel-icon {
+        limel-icon.mdc-deprecated-list-item__graphic {
             background-color: var(
                 --icon-background-color,
                 rgb(var(--contrast-900))
--- a/src/components/list/list.scss
+++ b/src/components/list/list.scss
@@ -67,6 +67,12 @@ $list-border-radius: 0.375rem; // 6px
         cursor: pointer;
     }
 
+    .mdc-deprecated-list-item__meta {
+        // the action menu on the right side a list item
+        line-height: 100%;
+        margin-right: -0.5rem;
+    }
+
     &.mdc-deprecated-list--avatar-list {
         position: relative;
 


mjlime:
@anderssonjohan The vertical alignment of the action menu icon is a big off, is that possible to fix as well?

Kiarokh (changes requested):
Looks good, please just add the little detail that I requested in a separate commit ;) Message could be something like: `fix(list): correctly position action menu on list items`

regarding your second commit message, I'd write something like:
`fix(list): remove badge effect from action menu & apply to icon only`

Kiarokh (dismissed):
⭐️⭐️⭐️⭐️⭐️ great job

Kiarokh (approved):

PR #1391, opened 2021-10-20
fix(tabs): emit event for inactive tab before active

jgroth:
When switching between different tabs, the event for the inactive tab needs to be emitted before the
event for the active tab or else two tabs will be active at the same time

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ www/
 src/components.d.ts
 src/**/readme.md
 docsDist/
+coverage/
 
 .npmrc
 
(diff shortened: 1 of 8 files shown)


jgroth:
> heart_eyes This finally fixes a "bug" in the tab-bar examples in the docs, which has annoyed me for ages! heartheartheart

Unfortunately it did not seem to fix the bug I was helping @hannahu with :smile: :disappointed: but the old behavior was wrong anyway :upside_down_face:

PR #1409, opened 2021-10-29
fix(menu): make MenuItem generic

jgroth:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,15 @@
+## [33.14.0-next.10](https://github.com/Lundalogik/lime-elements/compare/v33.14.0-next.9...v33.14.0-next.10) (2021-10-29)
+
+
+### Features
+
+* **button-group:** apply a tooltip component ([43cc060](https://github.com/Lundalogik/lime-elements/commit/43cc06001a777c2d6a763f0359d929097e5f5251))
+
+
+### Bug Fixes
+
+* **button-group:** prefix id with letter, to avoid possible invalid id ([eeb2514](https://github.com/Lundalogik/lime-elements/commit/eeb25148d38d0df6327bebff4185560cff25f05b))
+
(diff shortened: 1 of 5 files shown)


src/components/menu/menu.types.ts
```diff
@@ -2,7 +2,7 @@ import { ListItem } from '../list/list-item.types';
 
 export type OpenDirection = 'left' | 'right';
 
-export interface MenuItem extends ListItem {
+export interface MenuItem<T = any> extends ListItem<T> {
```
adrianschmidt, line 5: Just curious here:

Huh… so, with the `export interface MenuItem extends ListItem`, `MenuItem` doesn't automatically inherit the `<T = any>` part from `ListItem`? Seems like it should, logically, so (assuming it's not inherited) is that just a flaw in Typescript, and this is basically a workaround, or does it actually make sense that `export interface MenuItem extends ListItem` and `export interface MenuItem<T = any> extends ListItem<T>` should mean different things and this is a feature of Typescript, rather than a flaw? 🙂

jgroth: I think it's by design, otherwise you would not be able to do stuff like this
```ts
export interface HorseItems extends ListItem<Horse> {}
export interface CowItems extends ListItem<Cow> {}
```

Maybe not the best example here, but I hope you get the idea :sweat_smile:

adrianschmidt: Ah! Yeah, of course! That makes sense!

(I think your example was very clear and helpful! 🙏)

PR #1410, opened 2021-11-01
feat(header): make header component public

maltja:
fix Lundalogik/crm-feature#2026

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -44,7 +44,6 @@ import { Component, h, Prop } from '@stencil/core';
  * @exampleComponent limel-example-header-responsive
  * @exampleComponent limel-example-header-narrow
  * @slot - Content (actions) to be put inside the far right surface of the header
- * @private
  */
 @Component({
     tag: 'limel-header',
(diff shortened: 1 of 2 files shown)


src/components/header/header.tsx
```diff
@@ -44,7 +44,6 @@ import { Component, h, Prop } from '@stencil/core';
  * @exampleComponent limel-example-header-responsive
  * @exampleComponent limel-example-header-narrow
  * @slot - Content (actions) to be put inside the far right surface of the header
- * @private
```
adrianschmidt, line 47: Same comment as for popover:

> Even though this technically only affects the docs, I think the commit should be a `feat` since, while we are _technically_ only making the docs public, and the feature was public all along, the idea is of course that people should not have been using the feature, but now they can, so we are "adding" a feature 🙂

maltja: fixed! 🙂

adrianschmidt (dismissed):

adrianschmidt:
I merged the popover PR into this one, so we can merge both together and get a single release from them. That way we only have to wait for this to build once, instead of having to wait for the `next` branch to first build the release and then rebase and wait for this to build again 😄

adrianschmidt (approved):

PR #1411, opened 2021-11-01
feat(popover): make popover public

maltja:
fix Lundalogik/crm-feature#2102

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/popover/popover.tsx
+++ b/src/components/popover/popover.tsx
@@ -62,7 +62,6 @@ import { ESCAPE } from '../../util/keycodes';
  *
  * @slot - Content to put inside the surface
  * @exampleComponent limel-example-popover
- * @private
  */
 @Component({
     tag: 'limel-popover',


src/components/popover/popover.tsx
```diff
@@ -62,7 +62,6 @@ import { ESCAPE } from '../../util/keycodes';
  *
  * @slot - Content to put inside the surface
  * @exampleComponent limel-example-popover
- * @private
```
adrianschmidt, line 65: Even though this technically only affects the docs, I think the commit should be a `feat` since, while we are _technically_ only making the docs public, and the feature was public all along, the idea is of course that people should not have been using the feature, but now they can, so we are "adding" a feature 🙂

maltja: I agree, fixed! 🙂

adrianschmidt (approved):

adrianschmidt (approved):

Issue #1408, opened 2021-10-28
Tooltip content is positioned in relation to `limel-tooltip` instead of the trigger element


adrianschmidt:
## Current behavior
![Screenshot 2022-08-04 at 12 35 39](https://user-images.githubusercontent.com/35954987/182827360-182df61f-257e-4969-9fb2-e0ee2e49043c.png)

The popover with the tooltip content is positioned in relation to the `<div class="trigger-anchor">` inside the `<limel-tooltip />`  instead of in relation to the trigger element. That's a problem if those elements are not rendered right next to each other, as may be the case when the target element is part of a grid or flex layout, for example.

There's also a separate, smaller bug, which is related, in a way: The main `limel-tooltip` element, while having a size of 0 by 0 pixels, is still part of the normal render flow (it currently has `display:inline`), which can cause issues with the alignment of the target element. If we solve the above bug, this one should be trivial to solve, since we could probably just give it `position:absolute` instead, and the consumer also no longer needs to place the `limel-tooltip` element right next to the target element.

## Expected behavior

The popover should be positioned in relation to the target element.

## Environment

- **lime-elements** version: v33.14.0-next.8
- Framework used: Found in AngularJS, but should be easily reproducible anywhere.
- Logs:

Kiarokh:
I changed the title, since the word "popover" made me think that we're talking about the combo use of `limel-popover` and `limel-tooltip` together on the same element.

adrianschmidt:
Good! It's much better now 👍

adrianschmidt:
This has not been fixed:

![image](https://user-images.githubusercontent.com/154725/141286813-762563ed-4af0-4fcf-afb7-de51c981c382.png)

Kiarokh:
oh I think we forgot to remove the wrong `fix:` link from the PR! Good that you noticed

adrianschmidt:
I think I understand what the problem is here. `limel-portal` is hardcoded to use its host element as the element to position the portal in relation to. Fixing this should be a simple matter of adding a property where limel-portal can be given a target element, and have limel-portal fall back to using its host only if no target has been supplied. Hopefully 😅

eketorp:
When this is fixed, look at https://github.com/Lundalogik/crm-feature/issues/2590

jgroth:
[...]

PR #1413, opened 2021-11-01
fix(tooltip): fix position of a tooltip

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/1408
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tooltip/tooltip-content.scss
+++ b/src/components/tooltip/tooltip-content.scss
@@ -1,4 +1,4 @@
-.tooltip-content {
+:host {
     animation: display-tooltip 0.2s ease;
     display: flex;
     flex-direction: row;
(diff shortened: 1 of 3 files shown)


Kiarokh (changes requested):
Can you please split this into two commits: 

## First one 
It can be called for example: `refactor(tooltip): simplify the HTML structure of tooltip-content`. It would include changes in `tooltip-content.scss` and `tooltip-content.tsx` files.

This is a `refactor` because we're not really fixing anything. We're just rewriting the code in another way, without changing the output.

## Second one
This one would be about the change in `tooltip.scss`. You can perhaps call it like `fix(tooltip): prevent it from breaking the UI in some situations".

Because we're not really fixing the position problems of the tooltip here. So this title is a little bit more descriptive, and more like the truth. 😄

Kiarokh (approved):
great job!

PR #1416, opened 2021-11-02
Create a duplicate of limel-list, called limel-menu-list, only used by limel-menu

adrianschmidt:
**EDIT:** _To clarify, the idea here is to initially make the smallest possible change that completely separates limel-menu from limel-list, so that no future changes made to limel-menu will cause changes to limel-list. This means a lot of duplicated code in this PR, and a lot of dead code too. That will be cleaned up in successive PRs._

This is the first step (step 2 is here: #1417) that will allow us to remove all menu-specific code from limel-list, and all code not necessary for use with limel-menu from limel-menu-list.

Later, it might be possible to make limel-menu-list part of the limel-menu component, but that's an issue for a later step.

To make the changes easier to follow, I've split the change into several steps:
1. The first commit only duplicates the whole folder `src/components/list` into `src/components/menu-list` (except for the examples, which won't be used), and renames the files, prefixing each of them with `menu-`.
2. The second commit changes the names of the components, interfaces, etc.
3. This commit is just changes the linter made because of differences in line lengths caused by step 2.
4. Made the docs page for the new component private.

I've marked commit 2 and 3 as fixups, because I think we really only need this split for the review. But if someone thinks it would be useful to keep the commits separate in the permanent history, that's ok with me too 🙂

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu-list/menu-list-item.types.ts
+++ b/src/components/menu-list/menu-list-item.types.ts
@@ -1,6 +1,4 @@
-import { ListSeparator, MenuListItem } from '@limetech/lime-elements';
-
-export interface ListItem<T = any> {
+export interface MenuListItem<T = any> {
     /**
      * Text to display in the list item.
      */
@@ -40,4 +38,8 @@ export interface ListItem<T = any> {
      * List of actions to display as a menu at the end of the item
      */
     actions?: Array<MenuListItem | ListSeparator>;
+}
+
+export interface ListSeparator {
+    separator: true;
 }
--- a/src/components/menu-list/menu-list-renderer-config.ts
+++ b/src/components/menu-list/menu-list-renderer-config.ts
@@ -1,0 +1,9 @@
+import { IconSize, MenuListType } from '@limetech/lime-elements';
+
+export interface MenuListRendererConfig {
+    isMenu?: boolean;
+    isOpen?: boolean;
+    badgeIcons?: boolean;
+    type?: MenuListType;
+    iconSize?: IconSize;
+}
--- a/src/components/menu-list/menu-list-renderer.tsx
+++ b/src/components/menu-list/menu-list-renderer.tsx
@@ -1,0 +1,349 @@
+import { MenuListItem, ListSeparator, MenuItem } from '@limetech/lime-elements';
+import { h } from '@stencil/core';
+import { CheckboxTemplate } from '../checkbox/checkbox.template';
+import { MenuListRendererConfig } from './menu-list-renderer-config';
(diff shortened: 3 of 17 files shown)


jgroth:
I'm not quite sure I understand this first step and the need for the `limel-menu-list` component :thinking: Could we not just render everything required by the menu and use the `MDCMenu` in the menu component instead of relying on `limel-list` or `limel-menu-list` ? It feels like we are adding a lot of duplicated code here that we will want to remove anyway. Maybe it's just me that wants to take a bigger first step :smile:

adrianschmidt:
> I'm not quite sure I understand this first step and the need for the `limel-menu-list` component 🤔 Could we not just render everything required by the menu and use the `MDCMenu` in the menu component instead of relying on `limel-list` or `limel-menu-list` ? It feels like we are adding a lot of duplicated code here that we will want to remove anyway. Maybe it's just me that wants to take a bigger first step 😄

I tried taking a bigger first step. Then it turned out there had been changes to limel-list related to the use with limel-menu on `next`, and when I tried to incorporate those changes I got dozens of failing tests and everything just felt like it was starting to fall apart. I want to take small steps, and the first one is to completely separate limel-menu from limel-list, so no new menu-related functionality is added to limel-list.

hannahu:
> > I'm not quite sure I understand this first step and the need for the `limel-menu-list` component thinking Could we not just render everything required by the menu and use the `MDCMenu` in the menu component instead of relying on `limel-list` or `limel-menu-list` ? It feels like we are adding a lot of duplicated code here that we will want to remove anyway. Maybe it's just me that wants to take a bigger first step smile
> 
> I tried taking a bigger first step. Then it turned out there had been changes to limel-list related to the use with limel-menu on `next`, and when I tried to incorporate those changes I got dozens of failing tests and everything just felt like it was starting to fall apart. I want to take small steps, and the first one is to completely separate limel-menu from limel-list, so no new menu-related functionality is added to limel-list.

Since you've investigated it and feel this is the best way forward then let's go with this!

hannahu (approved):

adrianschmidt:
Squashed!

adrianschmidt:
Since this was approved before I squashed, I'll use my admin powers to merge 😊

PR #1418, opened 2021-11-03
Stencil sourcemaps

jgroth:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@lundalogik/lime-icons8": "^2.6.0",
     "@popperjs/core": "^2.9.3",
     "@rjsf/core": "^2.5.1",
-    "@stencil/core": "^2.8.1",
+    "@stencil/core": "^2.10.0",
     "@stencil/eslint-plugin": "^0.3.1",
     "@stencil/sass": "^1.4.1",
     "@types/codemirror": "^5.60.2",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -46,7 +46,7 @@
     "@lundalogik/lime-icons8": "^2.6.0",
     "@popperjs/core": "^2.9.3",
     "@rjsf/core": "^2.5.1",
-    "@stencil/core": "^2.8.1",
+    "@stencil/core": "^2.10.0",
```
adrianschmidt, line 49: Phew! The lime-admin frontend build failed the first time I tried it with this branch, but after verifying that it works with the latest "next"-release, and then removing the old build output in lime-elements and doing a fresh build before linking and building in lime-webclient, it worked again! 😅

I'm just gonna test a little in the web client and then we should be able to merge this 👍

jgroth: Great!

I was also a bit worried because some tests were failing on the first run, but I guess they were just flaky since it's green now :slightly_smiling_face:

adrianschmidt: Ah, this doesn't seem to be a new issue. It only happens when linking in a dev-build of lime-elements in the admin frontend. So it shouldn't be a stopper for this PR, and I'm looking into fixing the problem in a separate branch.

adrianschmidt (approved):

PR #1420, opened 2021-11-03
fix: scrollbar showing in progress flow component

BjornOstberg:
The -y causes a scrollbar to appear under certain circumstances related to size of the window and length of the string in the "button".

image

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/progress-flow/progress-flow-item/progress-flow-item.scss
+++ b/src/components/progress-flow/progress-flow-item/progress-flow-item.scss
@@ -133,7 +133,7 @@
     z-index: z-index.$limel-progress-flow-divider;
     right: 0;
     transform: translateX(50%);
-    overflow-y: hidden;
+    overflow: hidden;
 
     &:after {
         position: absolute;


src/components/progress-flow/progress-flow-item/progress-flow-item.scss
```diff
@@ -133,7 +133,7 @@
     z-index: z-index.$limel-progress-flow-divider;
     right: 0;
     transform: translateX(50%);
-    overflow-y: hidden;
+    overflow: hidden;
```
adrianschmidt, line 136: It seems reasonable to me that the divider should have `overflow:hidden`, not just in the vertical direction, unless there's something I'm missing. The only thing that makes me the tiniest bit unconfident is, why was `overflow-y` used to begin with?

@Kiarokh You've been involved in this component before, right? Can you test this PR?

src/components/progress-flow/progress-flow-item/progress-flow-item.scss
```diff
@@ -133,7 +133,7 @@
     z-index: z-index.$limel-progress-flow-divider;
```
adrianschmidt, line 133: Regarding the failing tests: it seems the tests that depend on Org or Repo secrets are failing, because your code is coming from a fork. Which makes sense, since code from a fork could probably get hold of those secrets if they were supplied to the workflow; it definitely could if the workflow files used are those from the fork…

Since this is a trivial fix, I'll just run the tests locally to make sure they pass, but I'll have to look into if there's some better way at some point 🙂

adrianschmidt: Build, tests, and linting all pass 👍

src/components/progress-flow/progress-flow-item/progress-flow-item.scss
adrianschmidt: Oh, yeah, checking this branch out is a bit special, since it's a branch from a separate fork. Don't worry about it Kia. I've tested it in the docs, and I feel good enough about merging that I'll just approve it myself.

adrianschmidt (approved):

src/components/progress-flow/progress-flow-item/progress-flow-item.scss
adrianschmidt, line 133: I'd like the commit message to be formatted something like `fix(progress-flow): remove scrollbar that sometimes appears', but I'll just merge this into a new branch in this repo, and update it myself before merging to `next` 👍

src/components/progress-flow/progress-flow-item/progress-flow-item.scss
Kiarokh: Sorry for the late response @adrianschmidt . It's first now I could open this notification. 

Honestly I don't remember whether we or I had a specific reason this. Maybe it was because of the alternative UI which we pulled out for now? 🤔 Not sure, even that doesn't really make sense. To me too, it sounds logical to have `overflow`.

PR #1417, opened 2021-11-02
Clean up limel-list and limel-menu-list a bit

adrianschmidt:
This is the second step in splitting all menu related functionality out of limel-list.

The first step was #1416

## Review:
- [ ] Commits are [atomic](https://seesparkbox.com/foundry/atomic_commits_with_git)
- [ ] Commits have the correct *type* for the changes made
- [ ] Commits with *breaking changes* are marked as such

### Browsers tested:
(Check any that applies, it's ok to leave boxes unchecked if testing something didn't seem relevant.)

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

PR #1427, opened 2021-11-08
Split menu from list, step 3 - more cleanup

adrianschmidt:
Continuation from #1417.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/menu-list/menu-list.scss
```diff
@@ -237,4 +237,3 @@ $list-border-radius: 0.375rem; // 6px
 @import './partial-styles/custom-styles.scss';
 @import './partial-styles/enable-multiline-text.scss';
 @import './partial-styles/_has-grid-layout.scss';
-@import './partial-styles/_static-actions.scss';
```
Kiarokh, line 240: Generally we wanna avoid maintaining duplicate styles. Is there a reason not to import `list/list.scss` into this file, and remove most of the CSS here, keep only menu things?
I don't mean in this PR. But I can look into this as step 4

adrianschmidt: No particular reason other than not having gotten around to even looking at the css yet. I wanted to have a clean split to begin with, since the problem I was trying to address was that limel-menu specific things had crept into limel-list, which isn't good. If we can have a `list.scss` that knows nothing about limel-menu, but is still useful to import in limel-menu, then that's fine with me 🙂

adrianschmidt:
@LucyChyzhova helped me test this and didn't find any new bugs, so I'll merge this when the build is green.

src/components/menu-list/menu-list.scss
Kiarokh: Nice. I'll look into that as step 4 when you're ready here.

PR #1428, opened 2021-11-08
2446 better vertically center align table cell content

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/2446

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/table/table.scss
```diff
@@ -97,6 +97,15 @@ $tabulator-arrow-color-active: rgb($tabulator-arrow-color-active-raw-value);
     .tabulator-cell {
         border-right: transparent;
         padding: functions.pxToRem(8);
+
+        display: inline-flex;
```
Kiarokh, line 101: I am not sure whether we should mark this as a breaking change or not. I know this will affect for instance one of our own components as described here: https://github.com/Lundalogik/crm-feature/issues/2447

adrianschmidt: Hmm… yeah… that's a good question… 🤔

It's better to err on the side of caution, so I'd say yes, mark it as breaking, and describe the change and potential effect in the breaking change description 👍

Kiarokh: Thanks, I just don't know the rituals of introducing breaking changes? where do I write that this is a breaking change? How to mark it?

adrianschmidt: https://github.com/Lundalogik/lime-elements/blob/next/CONTRIBUTING.md#footer 🙂

src/components/table/partial-styles/tabulator-custom-styles.scss
```diff
@@ -67,8 +67,7 @@
     }
 
     .tabulator-cell {
-        height: functions.pxToRem(44);
-        line-height: functions.pxToRem(32);
+        height: functions.pxToRem(44) !important;
```
LucyChyzhova, line 70: is it possible to avoid using `!important` in this context?

LucyChyzhova (approved):

src/components/table/partial-styles/tabulator-custom-styles.scss
Kiarokh: I have tried to avoid using `!important` as much as it was possible. But because we're overriding Tabulator's original styles, sometimes it is not possible. Tabulator has a messy and bad-written CSS unfortunately. So it's hard to avoid these things.

This one is also a very specific case, so using !important is not a very big deal. To avoid it, we have to add a lot of more classes before `.tabulator-cell` to make the styles very specific. But that's not worth it. Because our own CSS becomes harder to read & maintain, and the output becomes larger. It's better to keep the CSS structure clean and do not nest classes inside each other.

PR #1431, opened 2021-11-09
HACKDAY - add localstorage memory of open state

adrianschmidt:
### Replaced by #1452 and #1457

When given a value for the new property `stateKey`, or automatically when used through limel-form,
the collapsible section will remember its state across page loads.

fix: Lundalogik/hack-tuesday#197

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/templates/array-field-collapsible-item.ts
+++ b/src/components/form/templates/array-field-collapsible-item.ts
@@ -70,6 +70,12 @@ export class CollapsibleItemTemplate extends React.Component {
                 className: 'limel-form-array-item--object',
                 ref: 'section',
                 'is-open': this.isOpen,
+                'state-key': encodeURIComponent(
+                    schema?.lime?.stateKey ||
+                        `${schema.items?.$ref}-${this.props.index}-${
+                            data.title || data.property
+                        }`
+                ),
             },
             this.props.item.children
         );
--- a/src/components/collapsible-section/collapsible-section.tsx
+++ b/src/components/collapsible-section/collapsible-section.tsx
@@ -5,10 +5,12 @@ import {
     h,
     Prop,
     Element,
+    Watch,
 } from '@stencil/core';
 import { dispatchResizeEvent } from '../../util/dispatch-resize-event';
 import { Action } from './action';
 import { ENTER, ENTER_KEY_CODE } from '../../util/keycodes';
+import { getState, setState } from 'src/util/state-service';
 
 /**
  * @slot - Content to put inside the collapsible section
@@ -43,6 +45,21 @@ export class CollapsibleSection {
     public actions: Action[];
 
     /**
+     * Used to identify the section for the purpose of remembering the
+     * open/closed state.
+     */
+    @Prop({ reflect: true })
+    public stateKey: string;
+
+    /**
+     * Set to `false` to disable state memory even when `stateKey` is set.
+     * Normally, `stateKey` can just be left without a value to achieve the
+     * same thing.
+     */
+    @Prop({ reflect: true })
+    public rememberState: boolean = true;
+
+    /**
      * Emitted when the section is expanded
      */
     @Event()
@@ -63,9 +80,34 @@ export class CollapsibleSection {
     @Element()
     private host: HTMLLimelCollapsibleSectionElement;
 
+    private stateId: string;
+
     constructor() {
         this.onClick = this.onClick.bind(this);
         this.renderActionButton = this.renderActionButton.bind(this);
+    }
+
+    public connectedCallback() {
+        if (this.rememberState && this.stateKey) {
+            this.stateId = `limel-collapsible-section.${this.stateKey}.isOpen`;
+
+            const isOpen = getState(this.stateId);
+
+            if (
+                (isOpen === true || isOpen === false) &&
+                this.isOpen !== isOpen
+            ) {
+                this.isOpen = isOpen;
+                this.emitOpenState();
+            }
+        }
+    }
+
+    @Watch('isOpen')
+    public openCloseHandler(value: boolean) {
+        if (this.rememberState && this.stateId) {
+            setState(this.stateId, value);
+        }
     }
 
     public render() {
@@ -98,7 +140,10 @@ export class CollapsibleSection {
 
     private onClick() {
         this.isOpen = !this.isOpen;
+        this.emitOpenState();
+    }
 
+    private emitOpenState() {
         if (this.isOpen) {
             this.open.emit();
             const waitForUiToRender = 100;
--- a/src/components/collapsible-section/collapsible-section.tsx
+++ b/src/components/collapsible-section/collapsible-section.tsx
@@ -54,7 +54,8 @@ export class CollapsibleSection {
     /**
      * Set to `false` to disable state memory even when `stateKey` is set.
      * Normally, `stateKey` can just be left without a value to achieve the
-     * same thing.
+     * same thing, but when limel-collapsible-section is used in limel-form,
+     * `stateKey` will always be set. Use this property to override.
      */
     @Prop({ reflect: true })
     public rememberState: boolean = true;
--- a/src/components/form/form.types.ts
+++ b/src/components/form/form.types.ts
@@ -148,6 +148,21 @@ export interface LimeSchemaOptions {
      * Mark the field as disabled
      */
     disabled?: boolean;
+
+    /**
+     * Collapsible sections will remember their open/closed state when used
+     * in a form. If different sections are not able to uniquely identify
+     * themselves, they will "share" the same state. To prevent this, a unique
+     * key can be assigned to each section here.
+     */
+    stateKey?: string;
+
+    /**
+     * To disable a collapsible section's "state memory", and always use the
+     * default state as set via the `collapsed` property, set this value to
+     * `false`.
+     */
+    rememberState?: boolean;
 }
 
 /**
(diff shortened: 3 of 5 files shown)


Kiarokh:
can this fix the following? https://github.com/Lundalogik/crm-feature/issues/1885

adrianschmidt:
> can this fix the following? [Lundalogik/crm-feature#1885](https://github.com/Lundalogik/crm-feature/issues/1885)

Indeed, this hack was spawned from that issue 😊

image

Kiarokh:
Nice! maybe you can add that issue as a fix link too

adrianschmidt:
> Nice! maybe you can add that issue as a fix link too

Currently, this isn't in shape to be a serious fix for that. But I might clean it up, and if so, I will 👍

src/components/form/templates/array-field-collapsible-item.ts
```diff
@@ -70,6 +70,12 @@ export class CollapsibleItemTemplate extends React.Component {
                 className: 'limel-form-array-item--object',
                 ref: 'section',
                 'is-open': this.isOpen,
+                'state-key': encodeURIComponent(
+                    schema?.lime?.stateKey ||
+                        `${schema.items?.$ref}-${this.props.index}-${
+                            data.title || data.property
+                        }`
```
adrianschmidt, line 77: We built this, and the corresponding bit in `src/components/form/templates/object-field.ts` below, by inspecting the data we had access to and just using what we thought looked useful.

I distinctly remember us having utility functions for parsing the schema, but I couldn't any in lime-elements… 😞

adrianschmidt: @jgroth The stuff we've added in the first and second commit in this PR, I feel are good enough to put up for review. But this stuff in the third commit doesn't feel quite as solid… 😅

We could just leave it out, but I think the value of the feature really is increased quite a lot if it's enabled by default whenever a limel-collapsible-section is used as part of a limel-form. So, what do you think? Is what we've done in the third commit something we could clean up and use, or is it rubbish that just happens to work for the specific circumstances in lime-webclient? 😬

jgroth: What is you concern here? That the key might not be unique enough if it has not been explicitly defined?

My concern is actually about the collapsible section itself :sweat_smile: Both because it makes the component more smart/stateful but maybe more so because it couples the component to local storage when the consumer might want to use another storage solution.

adrianschmidt: > What is you concern here? That the key might not be unique enough if it has not been explicitly defined?

Haha! The fact that you wonder what my concern is is very comforting, as I take that to mean you don't share it 😄

My concern was that, since we found properties to build a unique ID through simply looking at the data we had and taking what worked, without really knowing if those properties would be there in any case other than in Lime Admin and on the object card in the CRM web client, I wasn't sure this was applicable to the general case, especially since the data available was slightly different in the two cases (here in `array-field-collapsible-item.ts` and in `object-field.ts`).

> My concern is actually about the collapsible section itself 😅

Hopefully I can alleviate those concerns, since I feel quite good about that part! 😊

> Both because it makes the component more smart/stateful

True, but, I think the added complexity is relatively small, and the component is extremely simple to begin with, which means even with this added complexity it's still very simple. We didn't want to force the consumer to use this component as stateful, which is why we added the possibility to opt out, but now that I think about it, it seems much more sensible to create a stateful component that wraps the stateless existing version of `limel-collapsible-section`, so that you don't need to opt out, but can simply use `limel-collapsible-section` rather than this new stateful component. Thoughts? I think I can make a version like that quite quickly.

> but maybe more so because it couples the component to local storage when the consumer might want to use another storage solution.

That's why we made the state service, because @mjlime pointed out that 1) other components may need to store state, 2) the collapsible-section has no reason to know or care about how the state is stored, and 3) we might want to make it possible for the consumer to configure how storing is done or even supply their own storage functionality. Point 3 isn't implemented at the moment, as we didn't want to over-engineer the solution during the hackday. But if we feel this is something we should implement now, or that we need to change something in the current implementation to make room for adding such functionality later, I'm absolutely ok with doing that 👍

jgroth: > Haha! The fact that you wonder what my concern is is very comforting, as I take that to mean you don't share it

Hmm yeah well I did not look into the implementation in detail, but using the data to create the key might be weird as you point out :thinking:  Does this mean that changing the data in the form will create a new key? Will that not break stuff? :sweat_smile: 

> it seems much more sensible to create a stateful component that wraps the stateless existing version of limel-collapsible-section

Yeah maybe. Or maybe just make it opt-in instead of opt-out so that there is no change in behavior from before? 

I might be overly cautious here but state is almost always evil and keeping state always comes with a price. I guess I'm mostly questioning whether we are doing this at the right place or if it should be done closer to the consumer rather than in lime elements :thinking: Did you think about how the third point might be implemented?

adrianschmidt: > Does this mean that changing the data in the form will create a new key? Will that not break stuff? 😅

I guess it depends. If you put ever-changing information in it, then, yes, you'd have to give it a key instead. But that's what the `stateKey` property is there for.

This case (`array-field-collapsible-item.ts`) is the one used in Lime Admin, and the data we use is as follows:

image

- `schema.items.$ref` is `"#/definitions/CompanyCardSectionSchema"` - this is a reference to stuff in the schema, but it happens to nicely identify the parent section we're in.
- `this.props.index` is `0` - this is the index of the item we're about to render (the collapsible), among its siblings. So the _Addresses_ section has `1` here.
- `data.title` is `"untitled"` - The title for the section.

So the key for this section will be `'#/definitions/CompanyCardSectionSchema - 0 - untitled'`. This means that it won't collide with the section called `untitled`, with index `0` under _Deal Card_, because `$ref` will be different. It also won't remember its state if moved or renamed, and that's intentional. If the section is moved or renamed, for this purpose, we don't consider it to be the same section anymore. If the consumer knows it is, it has to explicitly tell us that by supplying its own `stateKey`.

If we look at the next level of collapsibles used in Lime Admin, it's the ones for properties, found inside the collapsible we just looked at. Let's look at the _Telephone_ field:

image

- `schema.items.$ref` is now `'#/definitions/CompanyEditPropertySchema'`
- `this.props.index` is `1`
- `data.title` is undefined, so instead, we use
- `data.property`, which is `phone`

So the key here is `'#/definitions/CompanyEditPropertySchema - 1 - phone'`.

In the case of `object-field.ts`, that's the one that's used on the object card in the web client. (I never really did figure out why the same code isn't used in both places. If you know or understand that, I'd love to get up to speed 🙂) The data available here was slightly different, but very similar:

image

- `props.formContext.schema.$ref` is `'#/definitions/CompanySchema'` - again, this lets us differentiate between sections that have the same name and are in the same place, but on different limetypes.
- `props.idSchema.$id` is `'root_section1'` - I guess the `root` part might change if we had nestled sections? Not sure. Anyway, the `1` works as the index, and lets us differentiate between two identically named sections on the same limetype.
- `props.title` is of course `'Addresses'`

While I mention "limetypes", Lime Elements obviously doesn't know what those are, but also doesn't have to. All that matters is that the string is different between "different places". The question is, can we assume that they will be, at least under "normal circumstances"?

If not, I guess we'll have to skip this part, and instead add something along these lines in `lime-crm-components` and in the code for Lime Admin (if we want the functionality there too).

adrianschmidt: Come to think of it, this stuff isn't actually even in #1452 (which replaces this PR). But I've added it in another separate PR (#1457). That one still needs work though, because of the changes in #1452, compared to this one.

jgroth: There is no guarantee that `schema.items.$ref` is unique. It probably is in lime-admin because of how we generate the schemas though, but we had to write custom code in the schema in order to get those unique. In a more general case it is more likely that the same sub-schema is used in several places and then `schema.items.$ref` will be the same for those.

I don't think we should use any data from the form at all (like the title) to generate the key, since that will create a new key whenever the data is changed. The way we use that at the moment to generate the section headers is already too hackish and I would prefer if we don't add any more to that logic :smile: 

Would it not be enough to use the property path from the root to the section to uniquely identify it? E.g. `card.sections[1].fields[1].layout`

jgroth: I guess we also need an ID for the whole form, so the property path might not be enough :thinking:

adrianschmidt:
Replaced by #1452 and #1457

src/components/form/templates/array-field-collapsible-item.ts
adrianschmidt: @civing I think a solution that is a bit of a hybrid between this and what you've done in your draft might be quite good.

- We could add an optional property for identifying the section, basically the same as what we called `stateKey` here, but it might have a different name.
- That property is set in the form in lime-crm-components, when the schema is first loaded (or maybe when it's updated, if we really need to, but then we need to make sure not to create a loop—I don't think we'll need to though…?).
- If that identifier set, the component includes that value in the `open` and `close` events somehow.
- Then we listen to those events in lime-crm-components, and take care of the data-storing stuff there.
- For the initial state when loading the form the first time, we can simply override the schema's value for whether the section is open with the value we got from the stored data.

Thoughts?

PR #1442, opened 2021-11-15
chore(dialog): remove deprecated property `badgeIcon` from `DialogHeading`

adrianschmidt:
The property has been deprecated for a long time, and setting it no longer has any
effect.

BREAKING CHANGE: The deprecated property `badgeIcon` on the interface `DialogHeading`
has been removed. Setting the property already had no effect, but since the property
has now been removed, any consumers that are importing types from
**@limetech/lime-elements** and are setting this property, will get an error when
building. The solution is to simply not set the property.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dialog/dialog.types.ts
+++ b/src/components/dialog/dialog.types.ts
@@ -3,8 +3,4 @@ export interface DialogHeading {
     subtitle?: string;
     supportingText?: string;
     icon: string;
-    /**
-     * @deprecated
-     */
-    badgeIcon?: boolean;
 }


jgroth (approved):

Befkadu1:
The property, `badgeIcon` should be removed from [here](https://github.com/Lundalogik/lime-crm-components/blob/fb59aa4002698cc24e250e834c7d01f5d1530c44/src/table-view/components/column-dialog/column-dialog.tsx#L99) as well.

adrianschmidt:
> @LucyChyzhova found build error. Shouldn't the property, `badgeIcon` be removed from [here](https://github.com/Lundalogik/lime-crm-components/blob/fb59aa4002698cc24e250e834c7d01f5d1530c44/src/table-view/components/column-dialog/column-dialog.tsx#L99) as well?

Yes it should. Which reminds me that we should rebuild the workflow for automatically updating lime-elements in crm-components and webclient, because currently I think the workflow just updates the dependency and pushes to `main`, which will break the build on `main` in cases like this, where lime-elements has had a major version update with breaking changes. Instead, the workflow should create a PR, so that build and tests are run. Then it should preferably merge the PR if everything is green. I think the add-ons team built something like that for their dependencies… 🤔

PR #1446, opened 2021-11-16
fix(list): fix broken import of `MenuItem` so docs builds correctly

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/list/list-item.types.ts
+++ b/src/components/list/list-item.types.ts
@@ -1,4 +1,4 @@
-import { MenuItem } from '@limetech/lime-elements';
+import { MenuItem } from '../menu/menu.types';
 
 export interface ListItem<T = any> {
     /**


src/components/list/list-item.types.ts
```diff
@@ -1,4 +1,4 @@
-import { MenuItem } from '@limetech/lime-elements';
+import { MenuItem } from '../menu/menu.types';
```
adrianschmidt, line 1: This was what was preventing the _Types_ section of the docs from working correctly.

I'm guessing importing from `@limetech/lime-elements` doesn't work in `.types.ts` files, because the types haven't finished generating and thus haven't been linked up to the `@limetech/lime-elements` "alias" yet…? Importing from `@limetech/lime-elements` works in other parts of the code, like in `menu.tsx`.

PR #1447, opened 2021-11-16
Improve form for admin

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/form.scss
+++ b/src/components/form/form.scss
@@ -136,3 +136,8 @@
     padding-left: 1rem;
     padding-top: 0.25rem;
 }
+
+.button-add-new {
+    margin-top: 0.5rem;
+    width: 100%;
+}
--- a/src/components/form/templates/array-field.ts
+++ b/src/components/form/templates/array-field.ts
@@ -32,6 +32,7 @@ export class ArrayFieldTemplate extends React.Component {
             label: 'Add',
             onClick: this.handleAddClick,
             icon: 'add',
+            class: 'button-add-new',
         });
     }
 
--- a/src/components/form/templates/array-field.ts
+++ b/src/components/form/templates/array-field.ts
@@ -29,7 +29,6 @@ export class ArrayFieldTemplate extends React.Component {
         }
 
         return React.createElement('limel-button', {
-            outlined: true,
             label: 'Add',
             onClick: this.handleAddClick,
             icon: 'add',


src/components/form/form.scss
```diff
@@ -136,3 +136,7 @@
     padding-left: 1rem;
     padding-top: 0.25rem;
 }
+
+.add-new {
```
adrianschmidt, line 140: lime-elements should know nothing about Lime Admin. We might still need to add this styling here, but it should probably be called something that's relevant and understandable to any consumer of lime-elements.

(It's a good hint that something is wrong when the title of a PR in lime-elements references a specific consumer, like this one does. The change should preferably be possible to explain and justify in more general terms. Then we can of course always mention that this will be used in Lime Admin.)

Kiarokh: Sorry I didn't understand the problem. 

Screenshot 2021-11-17 at 12 08 26
The button is already used in forms in the admin, also in the examples in Elements. I'm just changing the styles. 

Do you mean I should change the title of the branch? For the change log, isn't it enough that the commit titles are explanatory? Or what do you mean?

adrianschmidt: I mean that the class probably shouldn't be called `add-new`.

Also, come to think of it, I think this is solving only a very specific instance of a more general problem. I think the only reason that the text-fields higher up in that screenshot has nice margins is that that's where the helper-text would go if they had any. We've got an issue somewhere about selects needing some margin in Lime Admin, precisely because the helper-text overlaps the label of the next field.

In short, I think we simply need to set a `gap`…

adrianschmidt (changes requested):
As per previous comment 🙂

src/components/form/form.scss
Kiarokh: > I mean that the class probably shouldn't be called add-new.

I see. Do you have any suggestions? I don't really know what to call it to be generic. `form-field-button`? 

> In short, I think we simply need to set a gap…

Yes, we made an issue here: https://github.com/Lundalogik/crm-feature/issues/2466 and talked with @jgroth who points out that solving this for Admin seems to be a bit tricky, using a single CSS `@prop`. We would need to find solutions so that we can write it like the way we do for columns, 

e.g.
````py
…
lime {
	'layout' : {
        'type': 'grid',
        'column': 3,
        'gap': 2,
    }
}
````

File: src/components/form/templates/array-field.ts [hunk 2]
```diff
@@ -29,10 +29,10 @@ export class ArrayFieldTemplate extends React.Component {
         }
 
         return React.createElement('limel-button', {
-            outlined: true,
             label: 'Add',
             onClick: this.handleAddClick,
             icon: 'add',
+            class: 'button-add-new',
```
adrianschmidt, line 35: Ah! Sorry, I didn't realise the information that this was an add-button existed in lime-elements already. I did see this file, but I missed that the icon and label properties shows the form knows this is an _Add_ button. Since it knows the context, it knows that a margin is needed 👍

adrianschmidt (approved):

adrianschmidt (approved):

src/components/form/templates/array-field.ts
Kiarokh: no worries 😄 so we're all good?

adrianschmidt: Never anything but 😄 👍

PR #1451, opened 2021-11-17
Add tests and minor fixes for collapsible-section

adrianschmidt:
To test the new tests, I would recommend looking at the test description, then breaking the code, so that it no longer does what the test description says, then check that the test fails.

So, for something like `limel-collapsible-section > with a header and body > when clicking the header > opens`, you might want to comment out the code inside the `onClick` function. Then check that that specific test fails in the expected way. If other tests fail (which some probably will), check that it makes sense that they fail because of the change you made. Then undo the change, and repeat for the next test 🙂

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/collapsible-section/collapsible-section.e2e.ts
+++ b/src/components/collapsible-section/collapsible-section.e2e.ts
@@ -1,0 +1,317 @@
+import {
+    E2EElement,
+    E2EPage,
+    EventSpy,
+    newE2EPage,
+} from '@stencil/core/testing';
+import { Action } from './action';
+
+describe('limel-collapsible-section', () => {
+    let page: E2EPage;
+    let limelCollapsible: E2EElement;
+    let collapsibleHeader: E2EElement;
+    let collapsibleBody: E2EElement;
+
+    describe('with a header and body', () => {
+        beforeEach(async () => {
+            page = await createPage(`
+                <limel-collapsible-section header="Header text">
+                    Body text <button>Test</button>
+                </limel-collapsible-section>
+            `);
+            limelCollapsible = await page.find('limel-collapsible-section');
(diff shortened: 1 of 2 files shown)


src/util/e2eClickElement.ts
```diff
@@ -0,0 +1,33 @@
+import { E2EElement, E2EPage } from '@stencil/core/testing';
+
+/* eslint-disable no-underscore-dangle */
+export async function clickElement(page: E2EPage, elements: E2EElement[]) {
```
jgroth, line 4: What is `elements` here? Given the name of the function I would expect just a single element, what are the other elements used for? :slightly_smiling_face: :thinking:

jgroth (commented):
Great with some new tests! :partying_face: 
Looks good, just had a small question

src/util/e2eClickElement.ts
adrianschmidt: Yeah, I think I need to add a docblock here…

If you just send in a single element, it clicks that. But if your element is inside a shadowDOM, you need to give it the respective shadowRoot hosts too…

I'll see what I can come up with to make it clearer 👍

adrianschmidt: Ah… I think this was a completely redundant invention by me… 🤦‍♂️

image

jgroth (dismissed):

adrianschmidt:
Squashed 🙂

jgroth (approved):

PR #1452, opened 2021-11-17
feat(collapsible-section-stateful): add collapsible section with state memory

adrianschmidt:
fix: Lundalogik/hack-tuesday#197

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/collapsible-section-stateful/collapsible-section-stateful.tsx
+++ b/src/components/collapsible-section-stateful/collapsible-section-stateful.tsx
@@ -52,7 +52,7 @@ export class CollapsibleSectionStateful {
      * Emitted when an action is clicked inside the header
      */
     @Event()
-    private action: EventEmitter<Action>;
+    protected action: EventEmitter<Action>;
 
     private stateId: string;
 
@@ -80,7 +80,6 @@ export class CollapsibleSectionStateful {
                 actions={this.actions}
                 onOpen={this.handleOpenEvent}
                 onClose={this.handleCloseEvent}
-                onAction={this.handleActionEvent}
             >
                 <slot />
             </limel-collapsible-section>
@@ -112,10 +111,5 @@ export class CollapsibleSectionStateful {
         if (this.isOpen) {
             this.isOpen = false;
         }
-    };
-
-    private handleActionEvent = (event: CustomEvent) => {
-        event.stopPropagation();
-        this.action.emit(event.detail);
     };
 }
--- a/src/components/collapsible-section-stateful/collapsible-section-stateful.tsx
+++ b/src/components/collapsible-section-stateful/collapsible-section-stateful.tsx
@@ -1,0 +1,121 @@
+import { Component, Event, EventEmitter, h, Prop, Watch } from '@stencil/core';
+import { Action } from '../collapsible-section/action';
+import { getState, setState } from '../../util/state-service';
+
+/**
+ * @slot - Content to put inside the collapsible section
+ * @exampleComponent limel-example-collapsible-section-stateful
+ */
+@Component({
+    tag: 'limel-collapsible-section-stateful',
+    shadow: true,
+})
+export class CollapsibleSectionStateful {
+    /**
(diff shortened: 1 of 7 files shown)


src/components/collapsible-section-stateful/collapsible-section-stateful.tsx
```diff
@@ -0,0 +1,121 @@
...
+
+    private handleCloseEvent = () => {
+        if (this.isOpen) {
+            this.isOpen = false;
+        }
+    };
+
+    private handleActionEvent = (event: CustomEvent) => {
+        event.stopPropagation();
+        this.action.emit(event.detail);
```
adrianschmidt, line 119: I hate this, but I haven't found any way to get the `onAction` property to show up on the element's type, other than declaring an EventEmitter on the component, and then the compiler complains about the property being unused, so we end up with this. Any ideas?

jgroth: Yeah it's a bit annoying that you have to re-declare any custom events from inner components. I get why because of the shadow dom encapsulation, but it's also weird since all events bubble.

I've usually solved this in the past by declaring the event as `protected` just to get the linter to not complain about unused properties :sweat_smile:

adrianschmidt: Ah! Of course! I even introduced that trick myself for `@Watch()` handlers, but didn't think to use it here 🤦‍♂️ 😄

Thank you! 🙏

adrianschmidt: ⚡

jgroth:
How do you feel about a separate component for this vs just a property on the regular component? 

If we add other stateful components, should we create separate components for those as well? It feels like we might end up with duplicates for a lot components with one stateful and one regular component which might feel weird for the consumer when reading the documentation etc. 

Another problem with this approach is that it creates some duplication. If we add a new property to the regular component, we need to remember to add it to the stateful one as well and remember to map it to the inner component.

src/components/collapsible-section-stateful/collapsible-section-stateful.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+/**
+ * @slot - Content to put inside the collapsible section
+ * @exampleComponent limel-example-collapsible-section-stateful
+ */
+@Component({
+    tag: 'limel-collapsible-section-stateful',
+    shadow: true,
+})
+export class CollapsibleSectionStateful {
```
adrianschmidt, line 13: > How do you feel about a separate component for this vs just a property on the regular component?
> 
> If we add other stateful components, should we create separate components for those as well? It feels like we might end up with duplicates for a lot components with one stateful and one regular component which might feel weird for the consumer when reading the documentation etc.
> 
> Another problem with this approach is that it creates some duplication. If we add a new property to the regular component, we need to remember to add it to the stateful one as well and remember to map it to the inner component.

Oh, sorry, I thought you wanted it as a separate component, so it wouldn't increase complexity of the original component.

I don't know, I guess I can see reasonable arguments either way…

It feels like the idea for setting a stateKey automatically when using limel-form is looking less likely to happen, and I think that if we involve the consumer in that part, we're not really gaining much "for free" from the consumer's perspective by doing the state handling inside lime-elements, so I'm thinking we might scrap this in the end. So I think I'll convert this to a draft and put it on ice until we know more about how we want the whole thing to work. It might not be the best issue to spend a lot of time on in this cycle. We can probably get more value in less time by focusing on other issues.

The first commit, that just adds tests and stuff to the original collapsible, I'd still like to have reviewed and merged though 😊

PR #1457, opened 2021-11-19
Use stateful collapsible in form

adrianschmidt:

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/templates/array-field-collapsible-item.ts
+++ b/src/components/form/templates/array-field-collapsible-item.ts
@@ -70,6 +70,12 @@ export class CollapsibleItemTemplate extends React.Component {
                 className: 'limel-form-array-item--object',
                 ref: 'section',
                 'is-open': this.isOpen,
+                'state-key': encodeURIComponent(
+                    schema?.lime?.stateKey ||
+                        `${schema.items?.$ref}-${this.props.index}-${
+                            data.title || data.property
+                        }`
+                ),
             },
             this.props.item.children
         );
--- a/src/components/form/form.types.ts
+++ b/src/components/form/form.types.ts
@@ -148,6 +148,21 @@ export interface LimeSchemaOptions {
      * Mark the field as disabled
      */
     disabled?: boolean;
+
+    /**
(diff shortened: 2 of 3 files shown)


src/components/form/templates/array-field-collapsible-item.ts
```diff
@@ -70,6 +70,12 @@ export class CollapsibleItemTemplate extends React.Component {
                 className: 'limel-form-array-item--object',
                 ref: 'section',
                 'is-open': this.isOpen,
+                'state-key': encodeURIComponent(
```
adrianschmidt, line 73: Since we've changed the implementation from being built into `limel-collapsible-section`, to being built separately in `limel-collapsible-section-stateful`, this needs to be changed a bit, so that instead of just setting a property, we also switch which component we are using…

Thus the draft-PR…

src/components/form/templates/array-field-collapsible-item.ts
```diff
@@ -70,6 +70,12 @@ export class CollapsibleItemTemplate extends React.Component {
...
+                    schema?.lime?.stateKey ||
+                        `${schema.items?.$ref}-${this.props.index}-${
+                            data.title || data.property
+                        }`
```
adrianschmidt, line 77: Discussion continued from https://github.com/Lundalogik/lime-elements/pull/1431#discussion_r747617427

adrianschmidt: This was always the shakiest part, and I'm not surprised if it doesn't hold up. I think we probably need to just let the consumer identify the different collapsibles, and by then it's probably best to just let the consumer keep track of the state too. We already provide the possibility to set the state as open or closed from the outside, via the `isOpen` property, and it's starting to look more and more likely to me that that's the way to go.

PR #1460, opened 2021-11-23
fix(slider): ensure that the component doesn't crash when value % step !== 0

jensgustafsson:
fix: https://github.com/Lundalogik/crm-feature/issues/2419

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/slider/slider.tsx
+++ b/src/components/slider/slider.tsx
@@ -124,6 +124,9 @@ export class Slider {
             return;
         }
 
+        const inputElement: HTMLInputElement = element.querySelector('input');
+        const value = this.getValue();
+
         /*
         For some reason the input element's `value` attribute is removed
         (probably by Stencil) when the element is first rendered. But if the
@@ -131,9 +134,38 @@ export class Slider {
         MDCSlider crashes.
         So we add the attribute right before initializing MDCSlider. /Ads
         */
-        element
-            .querySelector('input')
-            .setAttribute('value', `${this.multiplyByFactor(this.getValue())}`);
+        inputElement.setAttribute('value', `${this.multiplyByFactor(value)}`);
+
+        /*
+        When creating the `mdcSlider` component, its important that the value set in
+        the input field obeys the range and the step size.
+
+        The MDCSlider will throw an exception unless the value in the input element
+        is dividible by the step value and is in the provided range.
+        If an exception occurs, this component will crash and it will be impossible to change
+        its value.
+        The logic below ensures that the component will render even though the
+        provided value is wrong.
+        This could be considered wrong, but it at least fixes so that it's possible
+        to change the value from the UI.
+        */
+        const greaterThanOrEqualMin = value >= this.valuemin;
+        const lessThanOrEqualMax = value <= this.valuemax;
+        const dividableByStep = value % this.step === 0;
+
+        if (!greaterThanOrEqualMin) {
+            const newMin = this.multiplyByFactor(value);
+            inputElement.setAttribute('min', `${newMin}`);
+        }
+
+        if (!lessThanOrEqualMax) {
+            const newMax = this.multiplyByFactor(value);
+            inputElement.setAttribute('max', `${newMax}`);
+        }
+
+        if (!dividableByStep) {
+            inputElement.removeAttribute('step');
+        }
 
         this.mdcSlider = new MDCSlider(element);
         this.mdcSlider.listen('MDCSlider:change', this.changeHandler);


Befkadu1 (changes requested):
This is what I get on the `Registration form` example when `this.value` is `undefined`.

Screenshot 2021-11-24 at 15 09 20

jensgustafsson:
> This is what I get on the `Registration form` example when `this.value` is `undefined`.
> 
> Screenshot 2021-11-24 at 15 09 20

Will look in to that! Thx!

src/components/slider/slider.tsx
```diff
@@ -124,16 +124,48 @@ export class Slider {
...
+        its value.
+        The logic below removes ensures that the component will render even though the
+        provided value is wrong.
+        This could be considered wrong, but it at least fixes such that it's possible
+        to change the value from the UI.
+        */
+        const greaterThanMin = value > this.valuemin;
+        const lessThanMax = value < this.valuemax;
+        const dividableByStep = value % this.step === 0;
+        const delta = 0.01;
```
Befkadu1, line 155: If the range is between 0-100 (`this.valuemin` - `this.valuemax` ), is this range 0.01?

jensgustafsson: delta is something made up by me just to ensure that we can initialize the slider with valid values. It could have been any value really.

Looking back at the problem, I think it's only needed when value is greaterThanMax.

src/components/slider/slider.tsx
```diff
@@ -124,16 +124,48 @@ export class Slider {
...
+        The MDCSlider will throw an exception unless the value in the input element
+        is dividible by the step value and is in the provided range.
+        If an exception occurs, this component will crash and it will be impossible to change
+        its value.
+        The logic below removes ensures that the component will render even though the
+        provided value is wrong.
+        This could be considered wrong, but it at least fixes such that it's possible
+        to change the value from the UI.
+        */
+        const greaterThanMin = value > this.valuemin;
```
Befkadu1, line 152:
```suggestion
        const greaterThanOrEqualMin = value >= this.valuemin;
```

src/components/slider/slider.tsx
```diff
@@ -124,16 +124,48 @@ export class Slider {
...
+        const lessThanMax = value < this.valuemax;
```
Befkadu1, line 153:
```suggestion
        const lessThanOrEqualMax = value <= this.valuemax;
```

src/components/slider/slider.tsx
Befkadu1: delta was needed in both `greaterThanMax` and `lessThanMax`. It crushes when you have a data for example between  1-100

src/components/slider/slider.tsx
```diff
@@ -124,16 +124,48 @@ export class Slider {
...
+
+        /*
+        When creating the `mdcSlider` component, its important that the value set in
+        the input field obeys the range and the step size.
+
+        The MDCSlider will throw an exception unless the value in the input element
+        is dividible by the step value and is in the provided range.
+        If an exception occurs, this component will crash and it will be impossible to change
+        its value.
+        The logic below removes ensures that the component will render even though the
```
adrianschmidt, line 147:
```suggestion
        The logic below ensures that the component will render even though the
```

src/components/slider/slider.tsx
```diff
@@ -124,16 +124,48 @@ export class Slider {
...
+        provided value is wrong.
+        This could be considered wrong, but it at least fixes such that it's possible
```
adrianschmidt, line 149:
```suggestion
        This could be considered wrong, but it at least fixes so that it's possible
```

src/components/slider/slider.tsx
jensgustafsson: Thanks for looking into this! I've created a fixup with your suggestions

src/components/slider/slider.tsx
```diff
@@ -124,16 +124,49 @@ export class Slider {
...
+        This could be considered wrong, but it at least fixes so that it's possible
+        to change the value from the UI.
+        */
+        const greaterThanOrEqualMin = value >= this.valuemin;
+        const lessThanOrEqualMax = value <= this.valuemax;
+        const dividableByStep = value % this.step === 0;
+        const delta = 0.01;
+
+        if (!greaterThanOrEqualMin) {
+            const newMin = this.multiplyByFactor(value) - delta;
```
Befkadu1, line 158:
```suggestion
            const newMin = this.multiplyByFactor(value);
```

Just remove delta, I don't see the use here

src/components/slider/slider.tsx
```diff
@@ -124,16 +124,49 @@ export class Slider {
...
+            inputElement.setAttribute('min', `${newMin}`);
+        }
+
+        if (!lessThanOrEqualMax) {
+            const newMax = this.multiplyByFactor(value) + delta;
```
Befkadu1, line 163:
```suggestion
            const newMax = this.multiplyByFactor(value);
```

src/components/slider/slider.tsx
```diff
@@ -124,16 +124,49 @@ export class Slider {
...
+        to change the value from the UI.
+        */
+        const greaterThanOrEqualMin = value >= this.valuemin;
+        const lessThanOrEqualMax = value <= this.valuemax;
+        const dividableByStep = value % this.step === 0;
+        const delta = 0.01;
+
+        if (!greaterThanOrEqualMin) {
+            const newMin = this.multiplyByFactor(value) - delta;
+            inputElement.setAttribute('min', `${newMin}`);
```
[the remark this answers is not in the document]
adrianschmidt: This seems like a case where we really would benefit from having some tests. If you want to avoid testing the whole component, breaking out just the calculations to its own function and testing that might be a good idea. I learnt a trick for testing "private" functions on classes recently, which might be useful:

```ts
const testMe = new SomeClass(); // I think we could probably either get hold of the object for an element, or perhaps simply import `Slider` and do `new Slider()`?
const result = (testMe as any).privateFunctionToTest();
```

If that doesn't work, we can also move the function to a separate file, and import it. Then it can be imported and tested from that file.

I'm trying to prioritise the inline-create review, but if you need help getting started with a test file, hit me up 😊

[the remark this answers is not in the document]
jensgustafsson: Aha, nice to know!

I think I would rather create a new file/module though where I could expose those functions as public functions instead and then import them in the slider component.

@Befkadu1 - I would say that we don't need to document that since we're documenting that they should set values that is in the range. Other scenarios should be considered as undefined behavior IMHO. We have just fixed such that the component no longer crashes if provided an empty value.

Befkadu1 (approved):

Befkadu1 (dismissed):
It works awesome now 🎊 .

PR #1462, opened 2021-11-23
fix(array-field-collapsible-item): add margin-bottom to the collapsible section 

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/2494
 (only for using in Form - lime Admin)

image
image

### DIFF
--- a/src/components/form/templates/array-field-collapsible-item.ts
+++ b/src/components/form/templates/array-field-collapsible-item.ts
@@ -67,7 +67,7 @@ export class CollapsibleItemTemplate extends React.Component {
             'limel-collapsible-section',
             {
                 header: findTitle(data, schema, formSchema) || 'New item',
-                className: 'limel-form-array-item--object',
+                class: 'limel-form-array-item--object',
                 ref: 'section',
                 'is-open': this.isOpen,
             },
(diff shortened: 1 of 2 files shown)


src/components/form/templates/array-field-collapsible-item.ts
```diff
@@ -67,7 +67,7 @@ export class CollapsibleItemTemplate extends React.Component {
             'limel-collapsible-section',
             {
                 header: findTitle(data, schema, formSchema) || 'New item',
-                className: 'limel-form-array-item--object',
+                style: { marginBottom: '0.2rem' },
```
Kiarokh, line 70: Looks good. But:

1. Fir sizes, we only use multiples of 4, as [explained here](https://lundalogik.github.io/lime-elements/versions/latest/#/DesignGuidelines/size-rhythms.md/). This will ensure that the UI is consistent and visually rhythmic. ;) So you should use `0.25rem` instead of `0.2rem`.
2. I'm not a big fan of inline styles. :) I believe we should handle visual styles in the CSS files, not in the component code. Therefore I think it is better to keep the class, and add this style in `form.scss` instead.

Kiarokh (changes requested):

Kiarokh (approved):

PR #1464, opened 2021-11-23
fix(form): use `title` as label on button-add-new

Kiarokh:
also use a simpler icon

fix: https://github.com/Lundalogik/crm-feature/issues/2495

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/templates/array-field.ts
+++ b/src/components/form/templates/array-field.ts
@@ -29,9 +29,9 @@ export class ArrayFieldTemplate extends React.Component {
         }
 
         return React.createElement('limel-button', {
-            label: 'Add',
+            label: this.props.title || 'Add',
             onClick: this.handleAddClick,
-            icon: 'add',
+            icon: 'plus_math',
             class: 'button-add-new',
         });
     }


src/components/form/templates/array-field.ts
```diff
@@ -29,9 +29,9 @@ export class ArrayFieldTemplate extends React.Component {
         }
 
         return React.createElement('limel-button', {
-            label: 'Add',
+            label: this.props.title,
```
Befkadu1, line 32: What if the title doesn't exist?
Screenshot 2021-11-24 at 14 21 56

Kiarokh: Then I think it's still clear that the button will add a field, like whatever is above. Right? :) I think it's fine. 

Previously we had `+ Add`, now we just have `+`, which are the same things.

Kiarokh: However, I wonder if this makes the button inaccessible for screen readers in these scenarios. But that's an edge case.

adrianschmidt: You could do this if you want to…

```suggestion
            label: this.props.title || 'Add',
```

Kiarokh: Good idea! 👍 💡

Befkadu1: Then it becomes like this similar to the previous, I think it is fine.
Screenshot 2021-11-24 at 14 46 45

Befkadu1: Fix the lint error !!

Kiarokh: yes sir. pushed

Befkadu1 (approved):

PR #1469, opened 2021-11-25
fix(progress-flow): fix rendering problem on iOS

Kiarokh:
in some bizarre way, (probably when the component is embedded
in another web-component) iOS could not render the dividers correctly
due to combination of `overflow` and `transform`.

fix: https://github.com/Lundalogik/crm-feature/issues/2453

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/progress-flow/progress-flow-item/progress-flow-item.scss
+++ b/src/components/progress-flow/progress-flow-item/progress-flow-item.scss
@@ -134,8 +134,7 @@
 
     position: absolute;
     z-index: z-index.$limel-progress-flow-divider;
-    right: 0;
-    transform: translateX(50%);
+    right: calc(var(--step-height) / 2 * -1);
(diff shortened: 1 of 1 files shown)


adrianschmidt (changes requested):
I can't test on iOS at the moment, because some of BrowserStack's iOS devices just don't want to load for me, and the ones that do won't boot the app after login. But this seems fine. It's a `fix`, not a `refactor` though, so please change that. Then I'll approve! 🙂

adrianschmidt (approved):

PR #1471, opened 2021-11-25
feat: add `limel-color-picker` component

Kiarokh:
Things left to fix:

- [x] make the swatches are navigatabel with keyboard tab
- [x] ~use translation keys~
- [x] make some things props
- [x] make sure selected color in the picker is highlighted
- [x] implement `FormComponent`
- [x] add `componentWillLoad` 
- [x] link to color guidelines.
- [x] implement required (maybe disabled) prop

fix https://github.com/Lundalogik/crm-feature/issues/2503

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/color-picker/color-picker-palette.tsx
+++ b/src/components/color-picker/color-picker-palette.tsx
@@ -1,0 +1,96 @@
+import { Component, h, Prop, Event, EventEmitter } from '@stencil/core';
+import { FormComponent } from '../form/form.types';
+import { brightnesses, colors, getColorName, getCssColor } from './swatches';
+
+/**
+ * @private
+ */
+@Component({
+    tag: 'limel-color-picker-palette',
+    shadow: true,
+    styleUrl: 'color-picker-palette.scss',
+})
+export class Palette implements FormComponent {
+    /**
+     * Color value that is manually typed by the user
+     */
+    @Prop({ reflect: true })
+    public value: string;
+
+    /**
+     * Label of the input field
+     */
+    @Prop({ reflect: true })
+    public label: string;
+
+    /**
+     * Helper text of the input field
+     */
+    @Prop({ reflect: true })
+    public helperText: string;
+
+    /**
+     * Set to `true` if a value is required
+     */
+    @Prop({ reflect: true })
+    public required: boolean;
+
+    /**
+     * Emits chosen value to the parent component
+     */
+    @Event()
+    public change: EventEmitter<string>;
+
+    public render() {
+        const background = this.value ? { '--background': this.value } : {};
+
+        return [
+            <div class="color-picker-palette">{this.renderSwatches()}</div>,
+            <div class="chosen-color-name">
+                <limel-input-field
+                    label={this.label}
+                    helperText={this.helperText}
+                    value={this.value}
+                    onChange={this.handleChange}
+                    required={this.required}
+                />
+                <div class="chosen-color-preview" style={background} />
+            </div>,
+        ];
+    }
+
+    private renderSwatches = () => {
+        return colors.map((color) => {
+            return brightnesses.map(this.renderSwatch(color));
+        });
+    };
+
+    private renderSwatch = (color: string) => (brightness: string) => {
+        const colorName = getColorName(color, brightness);
+        const classList = {
+            swatch: true,
+            [colorName]: true,
+            'swatch--selected': this.value === getCssColor(color, brightness),
+        };
+
+        return (
+            <div
+                class={classList}
+                onClick={this.handleClick(color, brightness)}
+                tabindex="0"
+            ></div>
+        );
+    };
+
+    private handleChange = (event: CustomEvent<string>) => {
+        event.stopPropagation();
+        this.change.emit(event.detail);
+    };
+
+    private handleClick =
+        (color: string, brightness: string) => (event: MouseEvent) => {
+            const value = getCssColor(color, brightness);
+            event.stopPropagation();
+            this.change.emit(value);
+        };
+}
--- a/src/components/color-picker/color-picker.tsx
+++ b/src/components/color-picker/color-picker.tsx
@@ -1,0 +1,148 @@
+/* eslint-disable multiline-ternary */
+import { Component, h, Prop, State, Event, EventEmitter } from '@stencil/core';
+import { FormComponent } from '../form/form.types';
+
+/**
+ * This component enables you to select a swatch from out color palette, simply
+ * by clicking on it. You can then copy the css variable name of the chosen color
+ * and use it where desired.
+ *
+ * The color picker can also show you a preview of any valid color name or color value.
+ *
+ * :::note
+ * Make sure to read our [guidelines about usage of colors](/#/DesignGuidelines/color-system.md/) from our palette.
+ * :::
+ *
+ * @exampleComponent limel-example-color-picker
+ * @exampleComponent limel-example-color-picker-readonly
+ */
+@Component({
+    tag: 'limel-color-picker',
+    shadow: true,
+    styleUrl: 'color-picker.scss',
+})
+export class ColorPicker implements FormComponent {
+    /**
+     * Name or code of the chosen color
+     */
+    @Prop({ reflect: true })
+    public value: string;
+
+    /**
+     * The label of the input field
+     */
+    @Prop({ reflect: true })
+    public label: string;
+
+    /**
+     * Helper text of the input field
+     */
+    @Prop({ reflect: true })
+    public helperText: string;
+
+    /**
+     * Displayed as tooltips when picker is hovered.
+     */
+    @Prop({ reflect: true })
+    public tooltipLabel: string;
+
+    /**
+     * Set to `true` if a value is required
+     */
+    @Prop({ reflect: true })
+    public required: boolean;
+
+    /**
+     * Set to `true` if a value is readonly. This makes the component un-interactive.
+     */
+    @Prop({ reflect: true })
+    public readonly: boolean;
+
+    /**
+     * Emits chosen value to the parent component
+     */
+    @Event()
+    public change: EventEmitter<string>;
+
+    @State()
+    private isOpen = false;
+
+    public render() {
+        return [
+            this.renderTooltip(),
+            <div class="color-picker">
+                {this.renderPickerPalette()}
+
+                <limel-input-field
+                    label={this.label}
+                    helperText={this.helperText}
+                    value={this.value}
+                    onChange={this.handleChange}
+                    required={this.required}
+                    readonly={this.readonly}
+                    class="chosen-color-input"
+                />
+            </div>,
+        ];
+    }
+    private renderTooltip = () => {
+        if (!this.readonly) {
+            return (
+                <limel-tooltip
+                    label={this.tooltipLabel}
+                    elementId="tooltip-button"
+                />
+            );
+        }
+    };
+
+    private renderPickerPalette = () => {
+        if (this.readonly) {
+            return this.renderPickerTrigger();
+        }
+
+        return (
+            <limel-popover open={this.isOpen} onClose={this.onPopoverClose}>
+                {this.renderPickerTrigger()}
+                <limel-color-picker-palette
+                    value={this.value}
+                    label={this.label}
+                    helperText={this.helperText}
+                    onChange={this.handleChange}
+                    required={this.required}
+                />
+            </limel-popover>
+        );
+    };
+
+    private renderPickerTrigger = () => {
+        const background = this.value ? { '--background': this.value } : {};
+
+        return (
+            <div
+                class="picker-trigger"
+                slot="trigger"
+                style={background}
+                role="button"
+                tabindex="0"
+                onClick={this.openPopover}
+                id="tooltip-button"
+            />
+        );
+    };
+
+    private openPopover = (event: MouseEvent) => {
+        event.stopPropagation();
+        this.isOpen = true;
+    };
+
+    private onPopoverClose = (event: CustomEvent) => {
+        event.stopPropagation();
+        this.isOpen = false;
+    };
+
+    private handleChange = (event: CustomEvent<string>) => {
+        event.stopPropagation();
+        this.change.emit(event.detail);
+    };
+}
--- a/src/components/color-picker/examples/color-picker.tsx
+++ b/src/components/color-picker/examples/color-picker.tsx
@@ -1,0 +1,25 @@
+import { Component, h, State } from '@stencil/core';
+@Component({
+    tag: 'limel-example-color-picker',
+    shadow: true,
+})
+export class ColorPickerExample {
+    @State()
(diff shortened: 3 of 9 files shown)


src/components/color-picker/color-picker-palette.tsx
```diff
@@ -0,0 +1,104 @@
...
+ */
+
+@Component({
+    tag: 'limel-color-picker-palette',
+    shadow: true,
+    styleUrl: 'color-picker-palette.scss',
+})
+export class PaletteExample {
+    /**
+     *
```
Befkadu1, line 13: Write appropriate doc here

src/components/color-picker/color-picker-palette.tsx
```diff
@@ -0,0 +1,104 @@
...
+    @State()
+    private value: string;
+
+    /**
+     * Emits chosen value to the parent component
+     */
+    @Event()
+    private change: EventEmitter<string>;
+
+    private colors = [
```
Befkadu1, line 27: May be move this to somewhere and can be reusable by other components if needed

src/components/color-picker/color-picker-palette.tsx
```diff
@@ -0,0 +1,104 @@
...
+        'amber',
+        'orange',
+        'coral',
+        'brown',
+        'gray',
+        'glaucous',
+    ];
+
+    public render() {
+        this.value = this.searchedColor;
```
Befkadu1, line 51: Good to avoid updating the state in the render method. We can create `componentWillLoad` method and assign the state there.

```
  public componentWillLoad() {
        this.value = this.searchedColor;
    }

**Review by Befkadu1 — COMMENTED** [review]
Great work!! It is amazing that you made this component.

File: src/components/color-picker/color-picker.tsx [hunk 4]
```diff
@@ -0,0 +1,75 @@
...
+                    />
+                    <limel-color-picker-palette
+                        searchedColor={this.color}
+                        onChange={this.onChosenColor}
+                    />
+                </limel-popover>
+
+                <limel-input-field
+                    label="Chosen color"
+                    helperText="You can also type a color name or value to preview it here"
```
jgroth, line 51: We should not hard code these, they should probably be props on the component iteself. For full compatibility with the `Form` component, we should implement as many as the props from the `FormComponent` interface as possible, otherwise we might have to create a wrapper component if we need any other of the other ones when we want to use it in a form

src/components/color-picker/color-picker.tsx
```diff
@@ -0,0 +1,75 @@
...
+            <div class="color-picker">
+                <limel-popover open={this.isOpen} onClose={this.onPopoverClose}>
+                    <div
+                        class="picker-trigger"
+                        slot="trigger"
+                        style={background}
+                        role="button"
+                        tabindex="0"
+                        onClick={this.openPopover}
+                        title="Click to pick a color"
```
jgroth, line 41: This should not be hard coded

src/components/color-picker/color-picker-palette.tsx
Kiarokh: What is `searchedColor` @LucyChyzhova ? :) I don't remember. Did you add it?

src/components/color-picker/color-picker-palette.tsx
Kiarokh: Sounds good. We have several examples in the Design Guidelines which use this list. But is the effort worth it? (also thinking if the PR gets too big).

src/components/color-picker/color-picker-palette.tsx
```diff
@@ -0,0 +1,104 @@
...
+        const background = this.value ? { '--background': this.value } : {};
+
+        return [
+            <div class="color-picker-palette">{this.renderSwatches()}</div>,
+            <div class="chosen-color-name">
+                <limel-input-field
+                    label="Chosen color"
+                    helperText="You can also type a HEX or RGB vale here to preview it"
```
jgroth, line 59: Should not be hard coded

jgroth (commented):
Looks good, but we need to work a bit on the public interface of the component. Right now there are not inputs or outputs at all from it :smile: 

Also, after clicking on a color in the picker I think it would be great if we could highlight that color somehow so I can see what color I clicked on :slightly_smiling_face:

Kiarokh:
Sorry... I updated the branch and force pushed.

src/components/color-picker/color-picker.tsx
Kiarokh: > Also, after clicking on a color in the picker I think it would be great if we could highlight that color somehow so I can see what color I clicked on 🙂

That's right. We had the same idea. Should we do it in this PR or later? We thought we could add it later.

Kiarokh: > We should not hard code these, they should probably be props on the component iteself. For full compatibility with the Form component, we should implement as many as the props from the FormComponent interface as possible, otherwise we might have to create a wrapper component if we need any other of the other ones when we want to use it in a form

Oh we forgot to put them in translations files (there is also a duplicate of them, one in the palette component and one in the picker. I think the text is not synced neither if I remember correctly). Do you mean we should instead make them string props, and let them rather be translated when implemented by consumer?

src/components/color-picker/color-picker.tsx
Kiarokh: also asked above :) should it be a prop or a translation fetched from within Elements translation fiels?

jgroth: It depends. Some of them should probably be set on the consumer side and then it should be a prop. If it's an internal label that should always stay the same, then it should be in a translation file

src/components/color-picker/color-picker.tsx
jgroth: > That's right. We had the same idea. Should we do it in this PR or later? We thought we could add it later.

I don't want to merge the component in this state. If you mark it as private then you can do whatever :smile:

jgroth: > Do you mean we should instead make them string props, and let them rather be translated when implemented by consumer?

I don't think all of these should be in the translation files. The consumer should be able to control some of these labels, just as a label for a regular input field. E.g. in Lime Admin I imagine that we want to change the label to something like "Limetype color" and the description to something like "Pick a color that will be used for this Limetype in icons and headers"

LucyChyzhova: "Also, after clicking on a color in the picker I think it would be great if we could highlight that color somehow so I can see what color I clicked on "

@jgroth Yes, I will commit this changes , just fixed it

Kiarokh: OK. Sorry. It's my bad. I moved it back to "in progress" again to fix these things first. ;)

src/components/color-picker/color-picker-palette.tsx
LucyChyzhova: Yes, I have added it last night 😂 
We need it to send the color from the input field at the parent component to the child - picker palette

src/components/color-picker/color-picker.tsx
Kiarokh: Tnx. But I assume you are not talking only about the `title` right? Since you marked a few lines… Or?

src/components/color-picker/color-picker.tsx
```diff
@@ -0,0 +1,91 @@
...
+            .shadowRoot.querySelector('limel-color-picker-palette');
+    }
+
+    private removeSelectedColor(
+        pickerPalette: HTMLLimelColorPickerPaletteElement
+    ) {
+        pickerPalette.shadowRoot
+            .querySelector('.selected-color-palette')
+            ?.classList.remove('selected-color-palette');
+    }
```
jgroth, line 90: Do we really need to manipulate the DOM manually like this? Can we not handle everything in the `render` method? :thinking:

LucyChyzhova: Maybe, but I'm not sure how.. , I'm manipulating with the DOM since it was necessary to dynamically add CSS on click

jgroth: We should use the current value to determine if one of the colors is selected or not :slightly_smiling_face:

src/components/color-picker/color-picker.tsx
```diff
@@ -0,0 +1,97 @@
...
+     */
+    @Prop({ reflect: true })
+    public helperText: string;
+
+    /**
+     * The title text of the input field.
+     */
+    @Prop({ reflect: true })
+    public tooltipLabel: string;
+
```
jgroth, line 42: We need to emit a `change` event from this component, otherwise the consumer will never know when the value has changed

src/components/color-picker/color-picker.tsx
```diff
@@ -0,0 +1,97 @@
...
+     */
+    @Prop({ reflect: true })
+    public label: string;
+
+    /**
+     * The helper text of the input field.
+     */
+    @Prop({ reflect: true })
+    public helperText: string;
+
```
jgroth, line 36: We might want to implement a few more props like `disabled`, `required` etc.
Have a look here and see what props make sense :slightly_smiling_face: 
https://lundalogik.github.io/lime-elements/versions/34.0.1/#/type/FormComponent/

src/components/color-picker/examples/color-picker.tsx
```diff
@@ -0,0 +1,20 @@
...
+    shadow: true,
+})
+export class ColorPickerExample {
+    public render() {
+        return (
+            <limel-color-picker
+                tooltipLabel="Click to pick a color"
+                helperText="You can also type a color name or value to preview it here"
+                label="Chosen color"
+            />
```
jgroth, line 17: We should listen for the `change` event in the example and show the user the selected value somehow

src/components/color-picker/color-picker-palette.tsx
```diff
@@ -0,0 +1,107 @@
...
+            </div>,
+        ];
+    }
+
+    private renderSwatches = () => {
+        return this.colors.map((color) => {
+            return this.brightness.map(this.renderBrightness(color));
+        });
+    };
+    private renderBrightness = (color) => (brightnes) => {
```
jgroth, line 83: I was a bit confused when I helped you, but I think a better name for this could be `renderSwatch` since it is called from `renderSwatches` and renders a single swatch :slightly_smiling_face:

src/components/color-picker/color-picker-palette.tsx
jgroth: Moving it will not make the PR any bigger :smile:

src/components/color-picker/color-picker.tsx
```diff
@@ -0,0 +1,106 @@
...
+        event.stopPropagation();
+        this.isOpen = false;
+    };
+
+    private onChosenColor = (event: any) => {
+        this.value = event.detail;
+    };
+
+    private handleChange = (event: CustomEvent<string>) => {
+        this.value = event.detail;
```
jgroth, line 104: Value is a prop so we should not update it ourselves, we should get the value from the consumer.
We need to emit a `change` event here instead and let the consumer know that the value has changed

src/components/color-picker/color-picker.tsx
```diff
@@ -0,0 +1,151 @@
...
+     * Set to `true` if a value is required
+     */
+    @Prop({ reflect: true })
+    public required: boolean;
+
+    /**
+     * Set to `true` if a value is readonly. This makes the component un-interactive.
+     */
+    @Prop({ reflect: true })
+    public readonly: boolean;
```
Kiarokh, line 59: One thing I want us to keep in mind is that we have made two distinct components here. One is `limel-color-picker` which is what we display to the public, and the 2nd one is `limel-color-picker-palette` which is private. But we can potentially use it in other places too. Below could be an example of such implementation:
image

Which is why we have `readonly` here, but not for `limel-color-picker-palette`. For that component, it doesn't make sense to have a `readonly` mode.

jgroth:
Code is looking good now, there is one issue remaining though that we should fix or there might be bugs in the future :slightly_smiling_face: 

Since this is a new component it would also be good to have some tests. There is not really a good reason to skip those any more. 

Also, do we need to have two input fields? It feels like the input field inside the popover does not really add that much value. Or what are your thoughts about it?

Kiarokh:
Thanks for the review ! 

> it would also be good to have some tests.

Agree. But we have no idea about how to write tests 😥 Can you or @Befkadu1 do it please? 

> do we need to have two input fields? 

Yes. As explained here: https://github.com/Lundalogik/lime-elements/pull/1471#discussion_r759597809 we want the `limel-color-picker-palette` to be independent, and reusable.  For example, I want to use it on top of the design guidelines page (without the picker). Also there is an example for future use, in the comment above. :)

jgroth:
> Yes. As explained here: [#1471 (comment)](https://github.com/Lundalogik/lime-elements/pull/1471#discussion_r759597809) we want the `limel-color-picker-palette` to be independent, and reusable. For example, I want to use it on top of the design guidelines page (without the picker). Also there is an example for future use, in the comment above. :)

Yeah but I did not see an input field in your image so that's another reason to why I asked :smile: Do you want the input field in the guidelines? :thinking:

jgroth:
I can have a look at some tests tomorrow

Kiarokh:
> I did not see an input field in your image 

that's an old sketch we made for the admin mode… 😅 so it's not exactly like what the design ended up to be. 

We have now a few reasons for having an input field
1. The palette component that opens in the popover now, can be used alone in other contexts too. It doesn’t have to be used together with color picker.
1. The purpose of the input filed is
    1. to show the name of the selected color, so users can copy it.
    1. to let users preview other color codes or names. We have some old variables such as --lime-red or lime-green and things like that. We still want to show them, even if they are not part of the palette.

jgroth:
Yeah I get why there is an input field outside of the popover, but I don't see a reason to why there needs to be an input field part of the palette component inside the popover :slightly_smiling_face: Anyway, I might not have the full picture so I will stop nagging about it :sweat_smile:

jgroth:
I made some minor fixes and added some basics tests. It it seems ok I will squash and approve :slightly_smiling_face:

jgroth:
I think you can do it! :slightly_smiling_face: 
You did it for the icon picker! :smile:

jgroth (approved):

LucyChyzhova:
> Looks wonderful! Thanks @jgroth !
> 
> Up next will be this one then: [Lundalogik/crm-feature#2477](https://github.com/Lundalogik/crm-feature/issues/2477) Do you think it's something we can fix @LucyChyzhova ? Or should we ask Johan 😄

yes, we can! ;)

PR #1475, opened 2021-12-01
fix(menu-list): import styles from `limel-list` to remove duplicate css

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/2460

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


LucyChyzhova:
I have checked out your branch and a got some bugs, take a look at screenshots 👇, please 
[(component/limel-list/)](http://localhost:3333/#/component/limel-list/)
Some layout changes:
image

When you hover an item:
image
image

Some layout changes and a hovering problem:
image

Kiarokh:
Thanks @LucyChyzhova Good catch. I think it's because the branch is out-dated. We fixed these issues somewhere else. I'll update it now and we can check again in a while :)

Kiarokh:
OK, I updated the branch and don't see those issues ;)

LucyChyzhova:
yes, now everything is correct! 🎉

LucyChyzhova (approved):

adrianschmidt:
@LucyChyzhova Good catch, even though it turned out to be a false alarm 🙂

Please make sure to always rebase on the latest version of the target branch (`next` in this case) before reviewing. (Imagine if this problem had not been there before rebasing, but appeared _after_ rebasing. Then reviewing without rebasing first would have missed the problem.)

PR #1485, opened 2021-12-06
feat(form): add transformErrors prop

johannesss:
Make it possible to provide the [transformErrors](https://react-jsonschema-form.readthedocs.io/en/latest/api-reference/form-props/#transformerrors) function to customize the error messages.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/examples/custom-error-message-form.tsx
+++ b/src/components/form/examples/custom-error-message-form.tsx
@@ -1,0 +1,67 @@
+import { ValidationStatus, FormError } from '@limetech/lime-elements';
+import { Component, h, State } from '@stencil/core';
+import { schema } from './custom-error-message-schema';
+
+/**
+ * Form with custom error message
+ *
+ * @link custom-error-message-schema.ts
+ */
+@Component({
+    tag: 'limel-example-custom-error-message',
+    shadow: true,
+})
+export class CustomErrorMessageFormExample {
+    @State()
+    private formData: object = {
+        personalIdentityNumber: '',
+    };
+
+    @State()
+    private valid = true;
+
+    public render() {
+        return [
+            <limel-form
+                onValidate={this.handleFormValidate}
+                onChange={this.handleFormChange}
+                value={this.formData}
+                schema={schema}
+                transformErrors={this.transformErrors}
+            />,
+            <br />,
+            <limel-button
+                label="Submit"
+                primary={true}
+                disabled={!this.valid}
+                onClick={this.handleSubmit}
+            />,
+        ];
+    }
+
+    private handleFormChange = (event) => {
+        this.formData = event.detail;
+    };
+
+    private handleSubmit = () => {
+        const json = JSON.stringify(this.formData, null, '    ');
+        alert(`Sending information to villains...\n\n${json}`);
+    };
+
+    private handleFormValidate = (event: CustomEvent<ValidationStatus>) => {
+        this.valid = event.detail.valid;
+    };
+
+    private transformErrors = (errors: FormError[]): FormError[] => {
+        return errors.map((error) => {
+            if (
+                error.name === 'pattern' &&
+                error.property === '.personalIdentityNumber'
+            ) {
+                error.message = 'Invalid format, use YYYYMMDD-NNNN';
+            }
+
+            return error;
+        });
+    };
+}
--- a/src/components/form/form.tsx
+++ b/src/components/form/form.tsx
@@ -9,7 +9,7 @@ import {
 } from '@stencil/core';
 import React from 'react';
 import { render, unmountComponentAtNode } from 'react-dom';
-import JSONSchemaForm from '@rjsf/core';
+import JSONSchemaForm, { AjvError } from '@rjsf/core';
 import retargetEvents from 'react-shadow-dom-retarget-events';
 import { FormError, ValidationStatus } from './form.types';
 import {
@@ -34,6 +34,7 @@ import { isInteger } from './validators';
  * @exampleComponent limel-example-props-factory-form
  * @exampleComponent limel-example-form-layout
  * @exampleComponent limel-example-form-span-fields
+ * @exampleComponent limel-example-custom-error-message
  */
 @Component({
     tag: 'limel-form',
@@ -75,6 +76,12 @@ export class Form {
     public propsFactory?: (schema: Record<string, any>) => Record<string, any>;
 
     /**
+     * Custom function to customize the default error messages
+     */
+    @Prop()
+    public transformErrors?: (errors: FormError[]) => FormError[];
+
+    /**
      * Emitted when a change is made within the form
      */
     @Event()
@@ -96,6 +103,7 @@ export class Form {
 
     public constructor() {
         this.handleChange = this.handleChange.bind(this);
+        this.getCustomErrorMessages = this.getCustomErrorMessages.bind(this);
     }
 
     public connectedCallback() {
@@ -154,6 +162,7 @@ export class Form {
                     ArrayFieldTemplate: ArrayFieldTemplate as any,
                     ObjectFieldTemplate: ObjectFieldTemplate,
                     disabled: this.disabled,
+                    transformErrors: this.getCustomErrorMessages,
                     formContext: {
                         schema: this.modifiedSchema,
                         rootValue: this.value,
@@ -233,5 +242,39 @@ export class Form {
                 schemaPath: error.schemaPath,
             };
         });
+    }
+
+    private getCustomErrorMessages(originalErrors: AjvError[]): AjvError[] {
+        if (!this.transformErrors) {
+            return originalErrors;
+        }
+
+        const errors: FormError[] = originalErrors.map((error: AjvError) => {
+            return {
+                name: error.name,
+                property: error.property,
+                message: error.message,
+                // For some reason 'schemaPath' is missing from the AjvError type definition:
+                // https://github.com/rjsf-team/react-jsonschema-form/issues/2140
+                // eslint-disable-next-line @typescript-eslint/dot-notation
+                schemaPath: error['schemaPath'],
+            };
+        });
+
+        // Use `.call({}, …)` here to bind `this` to an empty object to prevent
+        // the consumer submitted `transformErrors` from getting access to our
+        // component's internals. /Ads
+        return this.transformErrors
+            .call({}, errors)
+            .map((transformedError: FormError) => {
+                const originalError = originalErrors.find((error: AjvError) => {
+                    return transformedError.property === error.property;
+                });
+
+                return {
+                    ...originalError,
+                    message: transformedError.message,
+                };
+            });
     }
 }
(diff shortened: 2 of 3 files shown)


src/components/form/form.tsx
```diff
@@ -74,6 +74,12 @@ export class Form {
     @Prop()
     public propsFactory?: (schema: Record<string, any>) => Record<string, any>;
 
+    /**
+     * Custom function to customize the default error messages provided by JSON Schema
+     */
+    @Prop()
+    public transformErrors?: (errors: AjvError[]) => AjvError[];
```
adrianschmidt, line 81: We don't want to expose types from `rjsf` to our consumers, as that locks us into `rjsf`, it also means that updating `rjsf` could become more difficult if there are breaking changes made to the `AjvError` interface. Can we create a proxy interface, that exposes reasonably generic properties, likely to still be available (or made available through an adapter) if the underlying implementation changes or is swapped out?

johannesss: I've made some changes based on your comment

adrianschmidt: Nice! I'll have a look 👍

src/components/form/form.tsx
```diff
@@ -234,4 +244,35 @@ export class Form {
             };
         });
     }
+
+    private getCustomErrorMessages(originalErrors: AjvError[]): AjvError[] {
```
adrianschmidt, line 248: Oooh, so nice! Both with the hiding of AjvError and the example you added 😍 👏👏👏

src/components/form/examples/custom-error-message-form.tsx
```diff
@@ -0,0 +1,67 @@
+import { ValidationStatus, FormError } from '@limetech/lime-elements';
```
adrianschmidt, line 1: Small feedback on the commit messages:

### First commit:

> feat: add transformErrors prop

- add a "scope" when applicable: `feat(form): …`
This can sometimes be a little arbitrary, but here in Lime Elements it's almost always super easy, because we use the component's name, minus the `limel-` prefix. In other repos it's often a bit more complicated, since we tend do have deeper structures, and you might want to keep the "scope" pretty short.
- write the commit message subject so it makes sense for someone reading the changelog
The first line of the commit message (`type(scope): subject`) is used to create the changelog entry, so use a scope and a subject that makes sense in that context. Try to summarise what has been done from a user's perspective when possible. The first line can be up to 100 characters, including the type and scope, so if it fits you can also include something about what problem it fixes. In this case, you already have a solid summary in the commit message body, although I'd actually skip the "provided by JSON schema" part: `feat(form): make it possible to customize the default error messages` 👌
- include `fix: #123` at the bottom of the commit message (if the commit fixes an issue)
That way, the changelog item for this commit will include a link directly to the issue it fixes, like for the second item here: image

---

### The last three commits

The last three commits should just be fixup commits, since they fix things in a commit that hasn't been merged to `next. Fixup commits indicate that they should be squashed into an existing commit. We also have a PR check that fails if there are any fixup commits. That lets us make any changes based on review feedback with fixup commits, that remain un-squashed while the review is ingoing (so we can review only the changes each time something new is pushed), while the PR check prevents us from accidentally merging the PR before squashing the fixups. [Here's a good post](https://thoughtbot.com/blog/autosquashing-git-commits) explaining fixup commits better than I can.

---

I can help you out with editing the existing commits, if you want me to 👍

src/components/form/form.tsx
johannesss: Thanks a lot @adrianschmidt!

src/components/form/examples/custom-error-message-form.tsx
johannesss: Thank you for the feedback @adrianschmidt, it all makes sense! I will review the guide you linked and conform to these rules in future PRs!

johannesss: Yes, some help with editing the existing commits would be great!

adrianschmidt: Fixed! Since the commit history was the last thing to fix before I'd approve and merge this, I realised it was easiest to just squash the commits while I was at it.

(Just to clarify: we don't just blindly squash each PR to _one_ commit. We try to make [atomic commits](https://github.com/Lundalogik/lime-elements/blob/next/CONTRIBUTING.md#atomic-commits), and consider tests, docs, and examples for a new feature as the same "atom" as the feature itself. So in this case, a single commit 🙂)

adrianschmidt (approved):

adrianschmidt (changes requested):
Hang on 😄

src/components/form/form.tsx
```diff
@@ -96,6 +103,8 @@ export class Form {
 
     public constructor() {
         this.handleChange = this.handleChange.bind(this);
+        this.getCustomErrorMessages = this.getCustomErrorMessages.bind(this);
+        this.transformErrors = this.transformErrors.bind(this);
```
adrianschmidt, line 107: I don't think we should bind this function to `this`. `this.transformErrors` is an external function, and binding it to `this` will give it access to the internals of the form component. Although, I guess, since it's called as a method of `this` it might already have that access… 🤔

Regardless, I don't think we need this bind, and if we don't we should remove it.

johannesss: Hmm, I'll have a look at this!

src/components/form/form.tsx
```diff
@@ -96,6 +103,8 @@ export class Form {
 
     public constructor() {
         this.handleChange = this.handleChange.bind(this);
+        this.getCustomErrorMessages = this.getCustomErrorMessages.bind(this);
```
adrianschmidt, line 106: We can also get rid of this bind, by using an arrow function instead. We try to use arrow functions instead of binds when possible nowadays, but there's a bunch of old code still using bind, because we used that a lot a few years ago.

adrianschmidt: Although, you don't really have to change that now, because once we've merged this, I'll change the already existing functions in this file anyway, in a separate PR, so I can do this one then too.

johannesss: Ahh yes, you're right! I will commit some changes shortly @adrianschmidt

src/components/form/form.tsx
adrianschmidt: I had to look into whether internals are actually exposed, and indeed they are. I have a fix, that I can push if ok with you?

src/components/form/form.tsx
johannesss: Sounds good! @adrianschmidt

src/components/form/form.tsx
johannesss: Yes of course, please do!

src/components/form/form.tsx
```diff
@@ -234,4 +243,38 @@ export class Form {
...
+                // eslint-disable-next-line @typescript-eslint/dot-notation
+                schemaPath: error['schemaPath'],
+            };
+        });
+
+        // Use `.call({}, …)` here to bind `this` to an empty object to prevent
+        // the consumer submitted `transformErrors` from getting access to our
+        // component's internals. /Ads
+        return this.transformErrors
+            .call({}, errors)
```
adrianschmidt, line 268: @jgroth I think we probably have more places where we call a consumer supplied callback as a method on `this`, and we need to use this pattern on every single one of them, or we have a security problem. I don't think we need to go find every instance right now, but we should at least keep this in the back of our heads and make sure to fix existing instances when we come across them, and also not let new instances in.

I've tested what happens if the callback is created using an arrow function, or if the consumer uses `bind(this)` before sending in the callback to limel-form, and that still works as expected—the `.call({}, …)` doesn't destroy the existing binding.

cc @BregenzerK FYI

src/components/form/form.tsx
adrianschmidt: ⚡

adrianschmidt: Fixed. I'll squash and approve, if you're happy?

johannesss: Great, looks good to me!

johannesss: I tried to come up with a different prop name for `limel-form` than `transformErrors` to differentiate from rjsf, but failed. It feels clear concise to me, but feel free to change it.

adrianschmidt: Nah, I think `transformErrors` is great. No need to _differentiate_ from rjsf, just to _decouple_ 👍

adrianschmidt (approved):

src/components/form/form.tsx
jgroth: I don't think there are that many places actually :slightly_smiling_face: There's the `propFactory` that's also part of the form component and the `searcher` that's part of the picker. Besides those I don't think we have props that are functions. We might have some that are part of an inner structure of some prop, but I doubt we would bind it to the component in those cases :sweat_smile: 

Another way to get rid of the `this` binding is to simply assign the function to a constant/variable before invoking it, e.g.
```ts
const transformErrors = this.transformErrors;
transformErrors();
```

johannesss: Oh, that's a nifty solution. Nice!

BregenzerK: Sorry, I need some more context here. Is there anything we need to check and refactor in our add-ons?

adrianschmidt: @BregenzerK Sorry, I mentioned you since you're a Lime Elements maintainer. I should have mentioned the group instead, to make the intention clear. I don't think you need to worry about this in add-ons, no.

adrianschmidt: > I don't think there are that many places actually 🙂 There's the `propFactory` that's also part of the form component and the `searcher` that's part of the picker. Besides those I don't think we have props that are functions. We might have some that are part of an inner structure of some prop, but I doubt we would bind it to the component in those cases 😅

I agree we probably don't have many. We might have none actually, when I think about it. But we should at least keep in mind that this pattern could be dangerous, and that doing a `bind(this)` isn't necessary to open the vulnerability; it's enough to have a callback sent in as a property, and then calling it directly from `this` (like `this.consumerProvidedCallback()` for example).

> Another way to get rid of the `this` binding is to simply assign the function to a constant/variable before invoking it, e.g.
> 
> ```ts
> const transformErrors = this.transformErrors;
> transformErrors();
> ```

Nice 👍

adrianschmidt: > > Another way to get rid of the `this` binding is to simply assign the function to a constant/variable before invoking it, e.g.
> > ```ts
> > const transformErrors = this.transformErrors;
> > transformErrors();
> > ```
> 
> Nice 👍

This one might run a greater risk of being "refactored" into the vulnerable version though… 🤔

It's pretty easy to write a test to check that the provided callback doesn't have access to the inner scope though, and as long as we do that, I like your suggestion 👍

Issue #1477, opened 2021-12-03
limel-form: disabled prop not working for select inputs


johannesss:
## Current behavior

Unable to disable select inputs when used in limel-form

1. Create limel-form with select inputs and provide the disabled property via lime component props
2. See that the input is not disabled

## Expected behavior

The select input should be disabled

## Environment

- **lime-elements** version: 34.0.1
- Framework used: stenciljs, vanilla
- Logs: none

Codepen demonstrating the issue: https://codepen.io/johanneseriksson/pen/WNZvWoB

adrianschmidt:
You need to put the `disabled` property as a direct descendant of the `lime` property, like this:

```js
document.querySelector('limel-form').schema = {
   title: 'Registration form',
    description: 'Please enter your personal information',
    type: 'object',
    required: [],
    properties: {
        sidekicks: {
            type: 'array',
            title: 'Sidekicks',
            lime: {
              disabled: true
            },
            description: 'Please select your sidekicks',
            uniqueItems: true,
            minItems: 2,
            items: {
                title: 'Friends',
                type: 'string',
                anyOf: [
                    {
                        type: 'string',
                        const: 'super',
                        title: 'Superman',
                    },
                    {
                        type: 'string',
                        const: 'widow',
                        title: 'Black Widow',
                    },
                    {
                        type: 'string',
                        const: 'america',
                        title: 'Captain America',
                    },
                    {
                        type: 'string',
                        const: 'squirrel',
                        title: 'Squirrel Girl',
                    },
                    {
                        type: 'string',
                        const: 'devil',
                        title: 'Daredevil',
                    },
                    {
                        type: 'string',
                        const: 'marvel',
                        title: 'Captain Marvel',
                    },
                    {
                        type: 'string',
                        const: 'fantastic',
                        title: 'Mr. Fantastic',
                    },
                ],
            },
        },
    },  
}
```

Although, it certainly would be preferable if it was possible to set the value in either place. I'll have a look, if it's easy, we'll fix it, if not, I say we perhaps add something about it to the docs instead.

(I just had to check before I posted: I tried it out in the "Dynamic schema" example for limel-form in the docs, and it seems to work to put the value in either place when using limel-input, so I'm not sure why it wouldn't also work for limel-select…)

johannesss:
Ah, yes that works. Sounds good! Yeah I really figured they'd be consistent for every component.

PR #1486, opened 2021-12-07
honor `lime.component.props.disabled` for form components

adrianschmidt:
fix: #1477

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/widgets/select.ts
+++ b/src/components/form/widgets/select.ts
@@ -26,6 +26,8 @@ export class Select extends React.Component {
             value = findValue(props.value, options);
         }
 
+        const additionalProps = props.schema.lime?.component?.props || {};
+
         return React.createElement(LimeElementsWidgetAdapter, {
             name: 'limel-select',
             value: value,
@@ -36,6 +38,7 @@ export class Select extends React.Component {
             extraProps: {
                 multiple: props.multiple,
                 options: options,
+                ...additionalProps,
             },
         });
     }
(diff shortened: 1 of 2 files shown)


src/components/form/widgets/select.ts
```diff
@@ -73,3 +76,16 @@ function findValue(value: string, options: Option[]) {
 function findValues(value: string[], options: Option[]) {
     return options.filter((option: Option) => value.includes(option.value));
 }
+
+function getAdditionalProps(schema: any) {
+    let props: any = {};
+
+    if (schema.lime?.component?.props) {
+        props = {
+            ...props,
```
jgroth, line 85: Super minor, but this seems a bit redundant since `props` is always an empty object here :sweat_smile:

adrianschmidt: # 🤦‍♂️ 😂

I copied this from the input-field widget, and removed some code between the `let props` and the `if`, and obviously forgot to take a step back and look at the result.

Thanks for the catch!

adrianschmidt: Oh… hang on… actually, I'm gonna have a look at if I can write a regression test for this, before we merge 😊

adrianschmidt: Or… maybe I'm not… because there are no tests at all currently, and I don't really have time to write them from scratch right now… 😞

adrianschmidt: Anyway, this is ⚡

jgroth (approved):

adrianschmidt:
Previously approved. I've just squashed the fixup, so I'll merge.

PR #1496, opened 2021-12-16
Feature/#2556/enable slot for limel list

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/2556

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/list/list-item.types.ts
```diff
@@ -40,8 +40,25 @@ export interface ListItem<T = any> {
      * List of actions to display as a menu at the end of the item
      */
     actions?: Array<MenuItem | ListSeparator>;
+
+    /**
+     *Component used to render in the list
+     */
+    component?: ListComponent;
```
adrianschmidt, line 47: Since this doesn't replace the whole content of the list-item, I think we should name it something more specific. I wouldn't be surprised if we want to add more slots for custom components in the future, and possibly even an option to completely replace the whole content of the list-item with a custom component. If either of those happen, having this property called `component` is going to be problematic, and renaming it would be a very annoying breaking change 😕

src/components/list/list-item.types.ts
```diff
@@ -40,8 +40,25 @@ export interface ListItem<T = any> {
...
 }
+
+export interface ListComponent {
+    /**
+     * Name of the component
+     */
+    name?: string;
+
+    /**
+     * Extra properties to give the component in addition to the properties
```
adrianschmidt, line 61: In addition to what properties? This looks like a copy-paste error from Lime CRM Admin?

src/components/list/list-renderer.tsx
```diff
@@ -127,6 +127,7 @@ export class ListRenderer {
             'mdc-deprecated-list-item': true,
             'mdc-deprecated-list-item--disabled': item.disabled,
             'mdc-deprecated-list-item--selected': item.selected,
+            'has-custom-component': !!item.component,
```
adrianschmidt, line 130: On line 156, we use `item?.component`. If we need the `?.` there, don't we need it here too?

src/components/list/examples/list-custom-component.tsx
```diff
@@ -0,0 +1,86 @@
+import { ListItem } from '@limetech/lime-elements';
```
adrianschmidt, line 1: I think this commit could be part of the commit that adds the feature, but keeping it separate is ok too. _But_, if you keep it separate, the commit type should be `docs`, not `feat`:

`docs(list): create a list example with a custom component`

src/components/list/examples/list-action.tsx
```diff
@@ -20,6 +20,14 @@ export class ListActionExample {
...
             icon: 'gorilla',
             actions: this.actionItems,
+            component: {
+                name: 'limel-circular-progress',
+                props: {
+                    value: 5,
+                    maxValue: 10,
+                    suffix: '%',
+                },
+            },
```
jgroth, line 30: Do we really need this to show up in every example? I think it should be enough to have one specific example where we show how to use a custom component in the list

src/components/list/list-renderer.tsx
```diff
@@ -143,13 +144,25 @@ export class ListRenderer {
                 {...attributes}
             >
                 {item.icon ? this.renderIcon(this.config, item) : null}
+                {this.getCustomComponent(item)}
```
jgroth, line 147: I thought this was supposed to be used instead of a regular icon :thinking: Should you be able to specify _both_ an icon and a component?

src/components/list/list-item.types.ts
jgroth: My thought was that this was supposed to be used instead of a regular icon and not in addition to it. If it is instead of an icon we could also change `icon` to `string | ListComponent`

src/components/list/list-renderer.tsx
```diff
@@ -143,13 +144,25 @@ export class ListRenderer {
...
 
+    private getCustomComponent(item: ListItem): Element {
+        if (!item?.component) {
+            return;
+        }
+
+        const WebComponent = item?.component?.name;
+        const props = item?.component?.props;
+
+        return <WebComponent {...props} />;
```
jgroth, line 163: We probably need to set some restrictions here. Right now there's nothing stopping me from inserting a _huge_ component for an item that will break the whole list. I think we need a wrapper element here that sets a max size of the component

src/components/list/list-item.types.ts
LucyChyzhova: Thank you @adrianschmidt , if I understood you correctly,  It's should not be called a "component", could it be better if we changed it to "customComponent".  Since we don't know what exactly is used here in the future, right, so quite difficult to be more specific? Or what is your suggestion?

Kiarokh: good points!

🤔 I don't see a reason for the custom component not to exist with other features like the icons. It's my bad for not providing more use cases. Sorry.…

The intention for having a custom component is that we want to enrich the informativeness and interactivity of the list items. Previously we have managed to hack together lists that are supposed to be informative and interactive. For example the Work-order widget by @eketorp 's team 👇 
image
Even though there is a creative usage of colors and icons to reflect the **status** of an order, it's not always easy for the user to understand that meaning. You only go that far with available tools. But if they had a custom component there, that component could for instance show a string of text, in addition to colors. It could even be interactive and let users take the oder for instance by clicking on it. But it doesn't necessarily have to mean that the icon should be gone. It's then up to the consumer how they want to setup the list. They may just keep the icons, but add other things in the component, or move the icons in there.


Let me share some of our future dreams that we should take into account now.

One scenario is that the custom component should be optimally placed on the right side of the text. The **Update** button in front of **My pipe** in this example is one of them: 👇 
image

At the moment I'm not sure if we should let consumers have `leftCustomComponent` and `rightCostomComponent`, or whether we should let them choose to display their custom component on the right side or left side.

In some scenarios user may want to have two, one on left, and one on right. (Or is this one of those cases that Adrian says, that consumer needs to replace the whole content with their own custom component?)
145998368-d01ed796-fd0e-408e-8b86-0a5fdd8deb74

src/components/list/examples/list-action.tsx
Kiarokh: check the last commit message ;)
This was just to make testing easier. We will drop this commit.

src/components/list/list-renderer.tsx
Kiarokh: Answered in a previous thread

src/components/list/list-renderer.tsx
Kiarokh: I don't think so. It should be totally up to the consumer to make this as big as they want. Or even maybe let it take the entire width of the list item. We never know the use case

src/components/list/list-item.types.ts
adrianschmidt: > Thank you @adrianschmidt , if I understood you correctly, It's should not be called a "component", could it be better if we changed it to "customComponent". Since we don't know what exactly is used here in the future, right, so quite difficult to be more specific? Or what is your suggestion?

No, it's not that you need to add "custom", it's that the name is very generic, but the placement of the component is quite specific.

Imagine that we would like to add another slot on the right side of the listItem, I can't really come up with anything right now, but let's imagine that we didn't already have the option to render an action menu there, and instead of the solution we have now, where you supply an array of actions, imagine that we decided that we'd have a slot there, where the consumer can add a menu-component of their own.

Then we'd perhaps call that property menuComponent. But, now we have one property menuComponent, and one property component. That's a little odd, right?

Then perhaps we want to make it possible for the consumer to supply a component that replaces the whole standard list-item. What should we call that? We might call it replacementComponent perhaps, but now we have menuComponent for adding a menu, replacementComponent for replacing the whole list-item, and component for adding a little thing on the left side? That would make no sense :sweat_smile:

So we need to come up with a good name for that "little thing on the left side" :smile: And preferably something that doesn't include "left" :smile:

adrianschmidt: @Kiarokh Do you have any suggestions?

If we don't come up with anything better, I guess we could call it something like `widget`. Then _at least_ we keep `component` free for if we want to let the consumer replace the whole item one day 😅

adrianschmidt: `iconAdjacentWidget` 😂 🙈

adrianschmidt: > One scenario is that the custom component should be optimally placed on the right side of the text. The **Update** button in front of **My pipe** in this example is one of them: 👇 image

This feels like "action" to me.

> At the moment I'm not sure if we should let consumers have `leftCustomComponent` and `rightCostomComponent`, or whether we should let them choose to display their custom component on the right side or left side.
>
> In some scenarios user may want to have two, one on left, and one on right. (Or is this one of those cases that Adrian says, that consumer needs to replace the whole content with their own custom component?) 145998368-d01ed796-fd0e-408e-8b86-0a5fdd8deb74

This one could be "primary" and "secondary", respectively.

> Another example: Screenshot 2021-12-16 at 12 24 30

This one could perhaps also be "secondary".

So, we could use `primaryComponent` perhaps, leaving the door open for `secondaryComponent` to be placed on the right side, and, if we feel the need to differentiate between "secondary" and "action", we could add `actionComponent` one day.

Unless someone has a better idea 😊

Kiarokh: Or should we call them `leftWidget`, and `rightWidget`?

If someone decides to use the `leftWidget` or `rightWidget` as a 100%wide widget, then they can be my guest :D

adrianschmidt: I think `left` and `right` are less than optimal. RTL languages, we might want to move them (`.greenHeading { color: blue; }`), stuff like that…

jgroth: I really don't like the term `widget`, it feels like we are introducing another concept very similar to something we already have

Kiarokh: Right, I see your point. Nowadays, in CSS the use `block-start` and `block-end` to be compatible with different writing directions. For example instead of saying `margin-right`, they write `margin-block-end`. [Read more…](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-block-end) Not sure if it's a good idea to borrow this naming convention somehow?

Kiarokh: primary and secondary would of course work as well

src/components/list/list-renderer.tsx
```diff
@@ -127,6 +127,7 @@ export class ListRenderer {
             'mdc-deprecated-list-item': true,
             'mdc-deprecated-list-item--disabled': item.disabled,
             'mdc-deprecated-list-item--selected': item.selected,
+            'has-primary-component': !!item?.primaryComponent,
```
adrianschmidt, line 130: I suggest adding a function for checking if there is a primaryComponent, so that we're sure it's done the same way everywhere. I also suggest checking for `item?.primaryComponent?.name`.  I'll add a fixup commit with this suggestion though.

src/components/list/examples/list-primary-component.tsx
```diff
@@ -0,0 +1,90 @@
...
+        { text: 'Delete object', value: 11 },
+    ];
+
+    private items: Array<ListItem<number>> = [
+        {
+            text: 'King of Tokyo',
+            secondaryText: '2 players',
+            value: 1,
+            actions: this.actionItems,
+            primaryComponent: {
```
adrianschmidt, line 23: I think this example is better without the extra complications of having a menu on one of the items.

```suggestion
    private items: Array<ListItem<number>> = [
        {
            text: 'King of Tokyo',
            secondaryText: '2 players',
            value: 1,
            primaryComponent: {
```

`fixup! docs(list): create a list example with a custom component`

src/components/list/examples/list-primary-component.tsx
```diff
@@ -0,0 +1,90 @@
...
+                onSelect={this.onSelectAction}
+                class="has-striped-rows"
+            />
+        );
+    }
+
+    private onSelectAction(event: CustomEvent<ListItem>) {
+        console.log('Executing action: ', event.detail);
+    }
+}
```
adrianschmidt, line 90:
```suggestion
            <limel-list
                items={this.items}
                class="has-striped-rows"
            />
        );
    }
}
```

`fixup! docs(list): create a list example with a custom component`

adrianschmidt (changes requested):
Nice!

Just a few small wishes from me:

src/components/list/list-item.types.ts
```diff
@@ -40,8 +40,25 @@ export interface ListItem<T = any> {
...
 
 export interface ListSeparator {
     separator: true;
 }
+
+export interface ListComponent {
+    /**
+     * Name of the custom component
+     */
+    name?: string;
```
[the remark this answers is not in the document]
LucyChyzhova: I agree, in this case, we shouldn't have a question mark here

src/components/list/examples/list-action.tsx
```diff
@@ -20,6 +20,14 @@ export class ListActionExample {
             value: 1,
             icon: 'gorilla',
             actions: this.actionItems,
+            primaryComponent: {
+                name: 'limel-circular-progress',
+                props: {
+                    value: 5,
+                    maxValue: 10,
+                    suffix: '%',
+                },
```
adrianschmidt, line 29:
```suggestion
                props: {
                    value: 5,
                    suffix: '%',
                },
```

adrianschmidt (approved):

PR #1500, opened 2021-12-28
fix(list): fix layout of two-lined lists, when only some individual i…

Kiarokh:
…tems have `secondaryText`

fix https://github.com/Lundalogik/crm-feature/issues/2573

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/list/list.scss
+++ b/src/components/list/list.scss
@@ -123,7 +123,8 @@ $list-border-radius: 0.375rem; // 6px
 
     &.mdc-deprecated-list--two-line {
         .mdc-deprecated-list-item__text {
-            padding: functions.pxToRem(8) 0;
+            padding-top: functions.pxToRem(8);
+            padding-bottom: functions.pxToRem(8);
         }
         .mdc-deprecated-list-item__primary-text {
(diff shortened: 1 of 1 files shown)


src/components/list/list.scss
```diff
@@ -127,7 +127,8 @@ $list-border-radius: 0.375rem; // 6px
 
     &.mdc-deprecated-list--two-line {
         .mdc-deprecated-list-item__text {
-            padding: functions.pxToRem(8) 0;
+            padding-top: functions.pxToRem(8);
+            padding-bottom: functions.pxToRem(8);
```
adrianschmidt, line 131: Yeah, when we _do_ override properties, we should probably try to override as few properties as possible, and make sure to only override those properties we really mean to. So using the shorthand css property should basically be banned when overriding MDC styles.

adrianschmidt (approved):

PR #1502, opened 2022-01-03
2579 add helpertext support for chipset

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/2579

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -63,6 +63,17 @@ export class ChipSet {
     public label: string;
 
     /**
+     * Optional helper text to display below the chipset.
+     * When type is `input`, the helper text is displayed below the
+     * input field when it has focus.
+     * When type is not `input`, the helper text is always displayed
+     * if the device is touch screen; otherwise it is shown when chip-set
+     * is hovered or focused using keyboard navigation.
+     */
+    @Prop({ reflect: true })
+    public helperText: string;
+
+    /**
      * True if the chip set should be disabled
      */
     @Prop({ reflect: true })
@@ -289,6 +300,7 @@ export class ChipSet {
             <div class={classes} role="grid">
                 {chipSetLabel}
                 {this.value.map(this.renderChip)}
+                {this.renderHelperLine()}
             </div>
         );
     }
@@ -340,7 +352,7 @@ export class ChipSet {
     }
 
     private renderInputChips() {
-        return (
+        return [
             <div
                 class={{
                     'mdc-text-field mdc-text-field--outlined': true,
@@ -408,8 +420,9 @@ export class ChipSet {
                 </div>
                 {this.renderLeadingIcon()}
                 {this.renderClearAllChipsButton()}
-            </div>
-        );
+            </div>,
+            this.renderHelperLine(),
+        ];
     }
 
     private dropZoneTip = (): string => {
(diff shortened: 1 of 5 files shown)


src/components/chip-set/chip-set.tsx
```diff
@@ -62,6 +62,15 @@ export class ChipSet {
     @Prop({ reflect: true })
     public label: string;
 
+    /**
+     * Optional helper text to display below the chipset.
+     * When type is not `input`, the helper text is always displayed
+     * if the device is touch screen; otherwise it is shown when chip-set
+     * is hovered.
```
adrianschmidt, line 69: And when the type _is_ `input`?

Kiarokh: > otherwise it is shown when chip-set is hovered.

Maybe it wasn't so clear after all 😄 

It's actually a bit incomplete too. Because that'll be shown when chipset is hovered or focused with keyboard navigation. I'll add a fixup.

thanks for the feedback.

Kiarokh: Thanks for the feedback. It was unclear.. I'll add a fixup soon

adrianschmidt: I assume that when the type is input, the helper text just works the way it does for a limel-input-field, right? If so, I think it's ok to just write that. This comment doesn't have to describe that behaviour. It just needs to not seem like a part of the comment is missing, because that will confuse people 😄

Kiarokh: I was too quick in responding. 😄 changed my answer later, but I think you already saw it. 

I see what you mean. I've pushed a fixup now. Do you think I should simplify the comment and instead make an example? Or is it OK as it is?

src/components/chip-set/chip-set.tsx
```diff
@@ -62,6 +62,17 @@ export class ChipSet {
     @Prop({ reflect: true })
     public label: string;
 
+    /**
+     * Optional helper text to display below the chipset.
+     * When type is not `input`, the helper text is displayed below the
+     * input field when it has focus.
+     * When type is not `input`, the helper text is always displayed
```
adrianschmidt, line 69: Both of these say "When type is not \`input\`" now 😄

I commented about saying something like "works just like for limel-input-field", but that was because I assumed we simply passed the helper-text value on to limel-input-field when type is `input`, but that doesn't seem to be the case?

Is the mdc-chip-set component handling the showing and hiding of the helper text, or are we doing that?

Kiarokh: > Both of these say "When type is not `input`" now 😄

Oh lord 🙈 my memory…

> Is the mdc-chip-set component handling the showing and hiding of the helper text, or are we doing that?

yes, for the type=input, it is MDC which is doing it. I didn't do anything special. Just added the helper line. For other types, I did some CSS rules for touchscreen and focus and hover.

Kiarokh: for input field, we say:
````tsx
    /**
     * Optional helper text to display below the input field when it has focus
     */
````

Comment (adrianschmidt) [reply src/components/chip-set/chip-set.tsx]: Yeah, ok, then I think it's good as it is now. Nice work!

File: src/components/chip-set/chip-set.tsx [hunk 3]
```diff
@@ -276,6 +287,7 @@ export class ChipSet {
             disabled: this.disabled || this.readonly,
             'mdc-text-field--with-trailing-icon': true,
         };
+
```
adrianschmidt, line 290: Please avoid whitespace changes in commits with functional changes. I'm guessing this just happened, but if you meant to include it, please put it in a separate commit.

src/components/chip-set/chip-set.tsx
Kiarokh: Thanks :)

Kiarokh: Is it approved then? Should I squash and push?

adrianschmidt: I know you saw the comment about whitespace. But yes, you can squash it. I'll approve when you've pushed the squashed commits.

Kiarokh: Ah OK. I'll remove the whitespace in the same time then :) I thought it was fine to leave it

Kiarokh: Done ✅

adrianschmidt (approved):

PR #1504, opened 2022-01-03
fix(list): don't set selected index on mdc-list if selected item not found

FredrikWallstrom:
Note to the reviewer: There are two commits, one pure refactor and one that changes the functionality. Reviewing them one by one is recommended. I tried adding a regression test for the bug, but couldn't get it to work, and I don't think we should delay getting the bug fix because I'm currently having trouble understanding how to properly test for errors in async functionality involving user input 😅 (My failures at adding a test can be witnessed in #1602)
/ @adrianschmidt

Setting selectedIndex -1 on mdc-list will cause consequential errors.
This happens when none of the items are selected or if the selected prop is missing on the items.

Bug introduced in this commit: https://github.com/Lundalogik/lime-elements/commit/0a2fad6ed61d6c071ce364fb2fb255adbf9c357c

Not sure if this will cause other problems, please shout in that case.

**How to reproduce:**
Update the items in a list with items that doesn't have the `selected` property will result in this error:

Skärmavbild 2022-01-03 kl  16 27 34

fix: #1593 

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/list/list.tsx
+++ b/src/components/list/list.tsx
@@ -155,9 +155,15 @@ export class List {
                 .filter((item: ListItem) => item.selected)
                 .map((item: ListItem) => listItems.indexOf(item));
         } else {
-            this.mdcList.selectedIndex = listItems.findIndex(
+            const selectedIndex = listItems.findIndex(
                 (item: ListItem) => item.selected
             );
+
+            if (selectedIndex === -1) {
+                this.mdcList.initializeListType();
+            } else {
+                this.mdcList.selectedIndex = selectedIndex;
+            }
         }
     }
 
--- a/src/components/list/list.tsx
+++ b/src/components/list/list.tsx
@@ -150,17 +150,15 @@ export class List {
 
         const listItems = this.items.filter(this.isListItem);
 
-        if (!this.multiple) {
+        if (this.multiple) {
+            this.mdcList.selectedIndex = listItems
+                .filter((item: ListItem) => item.selected)
+                .map((item: ListItem) => listItems.indexOf(item));
+        } else {
             this.mdcList.selectedIndex = listItems.findIndex(
                 (item: ListItem) => item.selected
             );
-
-            return;
         }
-
-        this.mdcList.selectedIndex = listItems
-            .filter((item: ListItem) => item.selected)
-            .map((item: ListItem) => listItems.indexOf(item));
     }
 
     private setup = () => {


src/components/list/list.tsx
```diff
@@ -151,10 +151,14 @@ export class List {
         const listItems = this.items.filter(this.isListItem);
 
         if (!this.multiple) {
-            this.mdcList.selectedIndex = listItems.findIndex(
+            const selectedIndex = listItems.findIndex(
                 (item: ListItem) => item.selected
             );
 
+            if (selectedIndex !== -1) {
+                this.mdcList.selectedIndex = selectedIndex;
```
adrianschmidt, line 159: I'm thinking we might want to deselect or set the selectedIndex to 0 or something if `selectedIndex === -1`, but I'll have to test the PR to see what actually happens first… That'll have to be another day though, hopefully tomorrow.

FredrikWallstrom: Yes deselect sounds like an option, not sure how to do that though. I guess that if we set selectedIndex to 0, the first item would be selected (visually as well?), so that is probably not the right thing to do.
But I'm not 100% into this so yes its best if you test it out, thanks :)

src/components/list/list.tsx
```diff
@@ -151,10 +151,14 @@ export class List {
         const listItems = this.items.filter(this.isListItem);
```
adrianschmidt, line 151: > **How to reproduce:**
> Update the items in a list with items that doesn't have the `selected` property will result in this error:
>
> Skärmavbild 2022-01-03 kl  16 27 34

Ah… this only happens when the list has radio-buttons or checkboxes, right? I was having trouble reproducing this until I examined that screenshot in a little more detail…

FredrikWallstrom: Yea sorry, should have make that more clear. It only happens for list types that is not multiple, i.e. `radio` and `selectable`.

FredrikWallstrom: `checkbox` is multiple, so we will take another path for that. That path will result in that we are assigning `this.mdcList.selectedIndex` to an empty list, which seems to be valid (it doesn't throw any errors at least).

adrianschmidt:
Sorry for letting this sit for so long, I only now saw that you need this for a bug fix in e-sign. I'll try to get this merged as soon as I can.

src/components/list/list.tsx
adrianschmidt: Can you add an example to that reproduces what you are doing, so that the bug is reproduced if this code change is reverted? So I know that I'm testing the correct thing.

adrianschmidt: It's just for testing, so don't worry about it looking nice in the docs or anything.

FredrikWallstrom: Really stupid example but something like this:

```
    @Prop()
    public methods: ListItem[];

    public componentWillLoad() {
        this.methods = [{ text: 'foo', value: 1 }];
    }

    public render() {
        return (
            <limel-list
                onChange={() => {
                    this.methods = [{ text: 'foo', value: 1 }];
                }}
                items={this.methods}
                type="radio"
            />
        );
    }
```

We are doing something similar in eSign (reassigning methods list), but in a more complex situation.

adrianschmidt: Ok, I'll use that. Thanks 👍

Question though: Is the intention to deselect the selected item? If not, can't you make sure the selected item has the property `selected: true`? 🤔

FredrikWallstrom: If I recall correctly, yes that is the intention. We are clearing the selected item when you close the select signing method modal in eSign.
I don't know whether that is bad UX or not 😄

FredrikWallstrom: Or that might be wrong ⬆️ 
This happens when we are signing the document. When we do that, we will rerender everything because we will receive a new signing object. A consequence of that is that the methods will also be reseted because they are extracted from the signing object.
As you see, a little bit more complex situation. Would probably be possible to solve in eSign as well, but it feels like we want to support these situations?

adrianschmidt: Yes, ideally, if the items are updated, and there is no longer one that is selected, we would like to deselect all items. The problem is that this fix doesn't actually do that. It just circumvents the error by creating an inconsistent state, and I think that's worse.

Unfortunately, there doesn't seem to be a way to deselect all items in MDC List, so if that's what you wish to achieve, I would suggest using the `key` attribute on limel-list to force the creation of a new list component every time the selection should be reset. Alternatively, if the selection shouldn't be reset, check which item was selected before you replace the current items, then select the corresponding item in the new set. If there's a chance the new set doesn't include the same item, you would still need to handle recreating the list when no item should be selected.

The `key` attribute is described at the end of the [_Conditionals_ section](https://stenciljs.com/docs/templating-jsx#conditionals) of the Stencil docs.

adrianschmidt: Or, well… I guess an argument could be made that limel-list should take care of recreating the mdc-component if the items are reset and there is no longer a selected item… 🤔

I think that would be reasonable solution, but honestly, if you need a fix urgently, I suggest just fixing it in your add-on for now. You can always remove that once we've had the time to fix it in Lime Elements. This seems like an obscure enough use case that I don't think it's reasonable for us to give it enough prio to address it urgently.

FredrikWallstrom: Yes I agree with you. Putting it in an inconsistent/wrong state doesn't sound like something we want to do now when you are explaining it!

I will close this, thanks for you valuable input. 
I will take a closer look on our case in cool down, maybe we can restructure a bit. I guess the downside with the `key` attribute is performance, but that is maybe negligible.

FredrikWallstrom:
This is not how we want to solve this, see comments above for more information.

src/components/list/list.tsx
adrianschmidt: Please feel free to reach out to me for help, if you need it. A fix in the add-on should be faster to get to production, as you really only need to consider the possible use cases within the context of the add-on, as opposed to the much more general stance that must be taken in Lime Elements.

src/components/list/list.tsx
```diff
@@ -150,17 +150,15 @@ export class List {
...
             this.mdcList.selectedIndex = listItems.findIndex(
                 (item: ListItem) => item.selected
             );
-
-            return;
         }
-
-        this.mdcList.selectedIndex = listItems
-            .filter((item: ListItem) => item.selected)
-            .map((item: ListItem) => listItems.indexOf(item));
```
adrianschmidt, line 163: This refactor might not make much sense in isolation, but when I made the change in the next commit, I wanted to make this inversion, or "un-inversion", of the `if` statement. But then it was much cleaner to just do that refactor as a separate commit first, with no changes to the functionality.

jgroth:
I'm not able to reproduce this bug :thinking: First I thought it might have been fixed with the newer MDC, but I even tested on the version where the bug is supposed to have been introduced :thinking: 
The fix is small and seems reasonable anyway so we could just merge it. Just wanted to know if I'm doing something wrong when testing it :smile: 

I did the following to test it after having selected a `limel-list` with checkboxes in the DOM in the documentation

```js
$0.items = $0.items.map(i => ({text: i.text, value: i.value}))
```

This will remove all `selected` from the items so it should result in a `-1` to have been returned in the code

adrianschmidt:
Lists with checkboxes are handles differently. Try replicating by clicking the already selected radio-button in one of the radio-button list examples…

adrianschmidt:
Or do exactly what you did, but for a radio-button list 😄

jgroth:
> Lists with checkboxes are handles differently. Try replicating by clicking the already selected radio-button in one of the radio-button list examples…

Aha! Yeah, that was a bit easier :smile:

jgroth (approved):

PR #1513, opened 2022-01-12
limel-tooltip: add maxLength and allow text line-wrapping

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/2596

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tooltip/examples/tooltip-composite.tsx
+++ b/src/components/tooltip/examples/tooltip-composite.tsx
@@ -1,0 +1,58 @@
+import { Component, h, Prop, State } from '@stencil/core';
+
+/**
+ * Composite
+ *
+ * A place to try different combinations of states.
+ */
+@Component({
+    tag: 'limel-example-tooltip-composite',
+    shadow: true,
+})
+export class TooltipCompositeExample {
+    @Prop({ mutable: true })
+    public schema: any;
+
+    @State()
+    private props = {
+        label: 'Search',
+        helperLabel: 'alt + F',
+        maxlength: 50,
+    };
+
+    public componentWillLoad() {
+        this.schema = {
+            ...this.schema,
+            lime: {
+                layout: {
+                    type: 'grid',
+                },
+            },
+        };
+        delete this.schema.properties.elementId;
+    }
+
+    public render() {
+        return [
+            <limel-button icon="search" id="tooltip-example" />,
+            <limel-tooltip {...this.props} elementId="tooltip-example" />,
+            this.renderForm(),
+        ];
+    }
+
+    private renderForm() {
+        return (
+            <limel-collapsible-section header="Settings">
+                <limel-form
+                    schema={this.schema}
+                    value={this.props}
+                    onChange={this.handleChange}
+                />
+            </limel-collapsible-section>
+        );
+    }
+
+    private handleChange = (event: CustomEvent) => {
+        this.props = event.detail;
+    };
+}
--- a/src/components/tooltip/tooltip-content.tsx
+++ b/src/components/tooltip/tooltip-content.tsx
@@ -15,7 +15,7 @@ export class TooltipContent {
      * Short descriptive text of the owner element.
      */
     @Prop({ reflect: true })
-    label: string;
+    label!: string;
 
     /**
      * Additional helper text for the element.
@@ -23,12 +23,33 @@ export class TooltipContent {
      * owner element.
      */
     @Prop({ reflect: true })
-    helperLabel: string;
+    helperLabel?: string;
+
+    /**
+     * The maximum amount of characters before rendering 'label' and 'helperLabel' in two rows.
+     */
+    @Prop({ reflect: true })
+    maxlength?: number;
 
     public render() {
+        let isLabelsTextLong = false;
+        if (this.helperLabel && this.maxlength) {
+            isLabelsTextLong =
+                this.label.length + this.helperLabel.length > this.maxlength;
+        }
+
+        const props: any = {};
+        if (this.maxlength) {
+            props.style = {
+                '--tooltip-max-width-of-text': `${this.maxlength}` + 'ch',
+            };
+        }
+
         return [
-            <div class="label">{this.label}</div>,
-            <div class="helper-label">{this.helperLabel}</div>,
+            <text class={{ 'has-column-layout': isLabelsTextLong }} {...props}>
+                <div class="label">{this.label}</div>
+                <div class="helper-label">{this.helperLabel}</div>
+            </text>,
         ];
     }
 }
--- a/src/components/tooltip/tooltip-content.tsx
+++ b/src/components/tooltip/tooltip-content.tsx
@@ -12,21 +12,19 @@ import { Component, h, Prop } from '@stencil/core';
 })
 export class TooltipContent {
     /**
-     * Short descriptive text of the owner element.
+     * Read more in tooltip.tsx
      */
     @Prop({ reflect: true })
     label!: string;
 
     /**
-     * Additional helper text for the element.
-     * Example usage can be a keyboard shortcut to activate the function of the
-     * owner element.
+     * Read more in tooltip.tsx
      */
     @Prop({ reflect: true })
     helperLabel?: string;
 
     /**
-     * The maximum amount of characters before rendering 'label' and 'helperLabel' in two rows.
+     * Read more in tooltip.tsx
      */
     @Prop({ reflect: true })
     maxlength?: number;
--- a/src/components/tooltip/tooltip.tsx
+++ b/src/components/tooltip/tooltip.tsx
@@ -1,6 +1,8 @@
 import { Component, h, Prop, Element, State } from '@stencil/core';
 import { createRandomString } from '../../util/random-string';
 
+const DEFAULT_MAX_LENGTH = 50;
+
 /**
  * A tooltip can be used to display a descriptive text for any element.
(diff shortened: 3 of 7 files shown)


src/components/tooltip/tooltip.tsx
```diff
@@ -71,6 +72,13 @@ export class Tooltip {
     @Prop({ reflect: true })
     public elementId!: string;
 
+    /**
+     * The maximum amount of characters before rendering 'label' and 'helperLabel' in two rows.
+     * Defaults to 50 characters.
+     */
+    @Prop({ reflect: true })
+    public maxCharacterLength: number;
```
jgroth, line 80: Documenting the default like this can be prone to errors if we decide to change the value in the future. It can be very easy to forget to either update the documentation or the code itself. 

If we instead give the prop a default value, we can both remove the documentation about the default (since it will be documented automatically) and remove the code below in the render method.

```suggestion
     */
    @Prop({ reflect: true })
    public maxCharacterLength: number = DEFAULT_MAX_LENGTH;
```

(`DEFAULT_MAX_LENGTH` can be defined outside the component so we don't need to turn of the eslint rule about magic numbers)

src/components/tooltip/tooltip.tsx
```diff
@@ -113,6 +121,8 @@ export class Tooltip {
                     <limel-tooltip-content
                         label={this.label}
                         helperLabel={this.helperLabel}
+                        // eslint-disable-next-line no-magic-numbers
+                        maxCharacterLength={this.maxCharacterLength || 50}
```
jgroth, line 125:
```suggestion
                        maxCharacterLength={this.maxCharacterLength}
```

src/components/tooltip/tooltip.tsx
jgroth: We might also want to just name this `maxlength` to keep it aligned with the naming of `limel-input-field`

Kiarokh: Great feedback! I t actually happened to us too to forget updating it, when we changed the default.

Will fix :)

LucyChyzhova: Fixed ;)

src/components/tooltip/tooltip.tsx
```diff
@@ -71,6 +75,13 @@ export class Tooltip {
     @Prop({ reflect: true })
     public elementId!: string;
 
+    /**
+     * The maximum amount of characters before rendering 'label' and 'helperLabel' in two rows.
+     * Defaults to 50 characters.
```
Kiarokh, line 80: @LucyChyzhova we need to remove this `* Defaults to 50 characters.` too :)

src/components/tooltip/tooltip-content.tsx
```diff
@@ -25,10 +25,31 @@ export class TooltipContent {
     @Prop({ reflect: true })
     helperLabel: string;
 
+    /**
+     * The maximum amount of characters before rendering 'label' and 'helperLabel' in two rows.
+     * Defaults to 50 characters.
+     */
```
Kiarokh, line 31: @jgroth is it OK to repeat the same documentation in two places? We say this one time here, and one time in the main component.

src/components/tooltip/tooltip-content.tsx
```diff
@@ -25,10 +25,31 @@ export class TooltipContent {
     @Prop({ reflect: true })
     helperLabel: string;
 
+    /**
+     * The maximum amount of characters before rendering 'label' and 'helperLabel' in two rows.
+     * Defaults to 50 characters.
```
Kiarokh, line 30: @LucyChyzhova `* Defaults to 50 characters.` :)

src/components/tooltip/tooltip-content.tsx
jgroth: `@inheritdoc` won't do anything here since we are not inheriting from a parent. This component is marked as `@private`  however so it won't show up in the docs, I don't think documentation here is as important

I can't remember why we needed to have two components like this though :thinking:

Kiarokh: > why we needed to have two components like this though

because of the Portal i guess?

> I don't think documentation here is as important
I suggest to remove them and just say: `read more in looltip.tsx` or `refer to main component` or something like that. We don't want to do a lot of bookkeeping.

src/components/tooltip/tooltip.tsx
```diff
@@ -41,8 +43,10 @@ import { createRandomString } from '../../util/random-string';
  * effortlessly recognize can be hovered.
  *
  * @exampleComponent limel-example-tooltip
+ * @exampleComponent limel-example-tooltip-max-character
  * @private
  */
+
 @Component({
     tag: 'limel-tooltip',
```
adrianschmidt, line 51: I wonder if you have some incorrect setting in your editor, because I've seen a blank line added between the docblock and the `@Component` decorator, just like this, on loads of your PRs. Or is it just a habit that's hard to kill? 😄

src/components/tooltip/tooltip.tsx
```diff
@@ -41,8 +43,10 @@ import { createRandomString } from '../../util/random-string';
  * effortlessly recognize can be hovered.
  *
  * @exampleComponent limel-example-tooltip
+ * @exampleComponent limel-example-tooltip-max-character
  * @private
```
adrianschmidt, line 47: Hmm… @jgroth, do you know if there's any particular reason this is still private? Is it because we weren't sure about the structure of the component?

src/components/tooltip/tooltip-content.tsx
```diff
@@ -25,10 +25,30 @@ export class TooltipContent {
...
+     * The maximum amount of characters before rendering 'label' and 'helperLabel' in two rows.
+     */
+    @Prop({ reflect: true })
+    maxlength: number;
+
+    private isLabelsTextLong = false;
+
+    public componentWillLoad() {
+        this.isLabelsTextLong =
+            this.label.length + this.helperLabel.length > this.maxlength;
```
adrianschmidt, line 38: `helperLabel` is optional, so `this.helperLabel` might be undefined, in which case this line will throw an error, and the application will crash.

src/components/tooltip/tooltip.tsx
jgroth: I cannot remember. I think we just forgot to remove it :slightly_smiling_face:

src/components/tooltip/tooltip-content.tsx
adrianschmidt: The easiest and best solution is probably just to make `helperLabel` default to an empty string.

src/components/tooltip/tooltip.tsx
Kiarokh: this might have come from my bad habits 😄

LucyChyzhova: yeah, lint fix didn't correct this

src/components/tooltip/tooltip-content.tsx
```diff
@@ -15,34 +15,43 @@ export class TooltipContent {
...
     @Prop({ reflect: true })
-    maxlength: number;
+    maxlength?: number;
 
     private isLabelsTextLong = false;
 
     public componentWillLoad() {
-        this.isLabelsTextLong =
-            this.label.length + this.helperLabel?.length > this.maxlength;
+        if (this.helperLabel && this.maxlength) {
```
adrianschmidt, line 34: I added this fixup to make `helperLabel` and `maxLength` properly optional.

src/components/tooltip/tooltip-content.tsx
```diff
@@ -29,18 +29,13 @@ export class TooltipContent {
     @Prop({ reflect: true })
     maxlength?: number;
 
-    private isLabelsTextLong = false;
-
-    public componentWillLoad() {
+    public render() {
+        let isLabelsTextLong = false;
```
adrianschmidt, line 33: I then added this second fixup where I moved the checks into the render method, so that they would apply correctly if the values changed after the component was first created. This was necessary to make the composite example work.

adrianschmidt: It's possible that it would be more performant to have these checks run in a watcher, rather than for every render, but I'm not so sure, because these checks ought to be quite fast, and a watcher would (I assume) need to set up listeners and stuff.

src/components/tooltip/examples/tooltip-composite.tsx
```diff
@@ -0,0 +1,58 @@
+import { Component, h, Prop, State } from '@stencil/core';
```
adrianschmidt, line 1: @LucyChyzhova I squashed your fixups, and added a few commits. If you're happy with my changes and additions, I'm happy to squash and merge 🙂

src/components/tooltip/tooltip-content.tsx
jgroth: Yeah we don't need to think about performance here I think :slightly_smiling_face: 
Things that are typically slow are network requests and manipulating the DOM and we do neither here. A watcher would require us to introduce some kind of state and state is always evil :smile:

adrianschmidt (approved):

PR #1514, opened 2022-01-12
feat: table row selection

anderssonjohan:
Spike on using row selection features in Tabulator.

Features:

- Enable checkbox column for row selection using `selectable={true}` prop
- Multi-select rows by clicking checkboxes in column
- Shift-click to extend selection without emitting row activation
- Clicking anywhere inside the row selector cell toggles selection without activating the row
- Activating the row (teal highlight) does not toggle its row selector (the checkbox)
- Pre-selection of rows using `selection={ [ .... ] }`
- Event handler `onSelect` to get row selection changes (name chosen without `Row` suffix to be consistent with `onActivate`)
- Event handler `onSelectAll` (emitted after `select`) to get a boolean value when the user interacts with the column header "Select All" checkbox

⚠️ Unresolved questions/need feedback: ⚠️

-

Resolved questions:

- When activating a row and then clicking its checkbox, should that clear the activeRow immediately or when a second item is checked? If not clearing this immediately it will always open up the limeobject preview dialog when making the first selection. See screenrec for example of selecting the activated item and deactivating it while it remains visually active. One solution could be to use one visual style for activation and another for selection? **Solution: let's keep activerow and selection separete and don't clear activerow on selection changes**
- Rows seems to expand to the height of the checkbox ripple sometimes (consistently in the webclient with many rows) **This was caused by the MDC ripple size that made the checkbox 40x40, now set to 18px**
- Row shadow on hover is not applied to frozen columns **Fixed**
- Active row styling on frozen column is inconsistent with non-frozen columns **Fixed**

Suggested todos:

- [x] built-in Tabulator formatter `rowSelection` captures the checkbox click event and prevents the expected shift-click extension of the selection when clicking the checkboxes. Shift-click to extend only works when clicking _outside_ the checkbox! This should be replaced by our own "rowSelection column", which also opens up the possibility to use a limel-checkbox for improved styling
- [x] tests!

No-gos:

- Disabling selection of specific rows using a callback function. Perhaps usable in the future.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [x] Firefox
- [x] Safari

Mobile:
- [x] Chrome on Android
- [ ] iOS


src/components/table/examples/table-selectable-rows.tsx
```diff
@@ -0,0 +1,61 @@
...
+ * @link persons.ts
+ */
+@Component({
+    tag: 'limel-example-table-selectable-rows',
+    styleUrl: 'table.scss',
+    shadow: true,
+})
+export class TableExampleSelectableRows {
+    @State()
+    private activeRow: Person;
```
Kiarokh, line 17: @anderssonjohan asked:
> When activating a row and then clicking its checkbox, should that clear the activeRow immediately or when a second item is checked? If not clearing this immediately it will always open up the limeobject preview dialog when making the first selection. See screenrec for example of selecting the activated item and deactivating it while it remains visually active. 

We're not sure what activeRow exactly refers to. If by "should that clear the activeRow immediately" you mean removing the green background highlight? The answer would be YES. 

But if you mean "closing the related view", we believe currently believe that the related view should remain open. The green background highlight should only show which item is open in related view. This interaction should remain independent, if possible. You can get a feel for this interaction if you play around in Lime Go.

> One solution could be to use one visual style for activation and another for selection?

We think a nice and visually noticeable checkbox (not Tabulator's default checkboxes) is enough as a distinct and different visual style.

src/components/table/table.tsx
```diff
@@ -328,7 +387,24 @@ export class Table {
...
+                ...columnDefinitions,
+            ];
+        }
+
+        return columnDefinitions;
+    }
+    private getRowSelectorColumnDefinition(): TabulatorTable.ColumnDefinition {
+        return {
+            title: '',
+            formatter: 'rowSelection',
```
Kiarokh, line 404: @anderssonjohan asked:
> built-in Tabulator formatter rowSelection captures the checkbox click event and prevents the expected shift-click extension of the selection when clicking the checkboxes. Shift-click to extend only works when clicking outside the checkbox! This should be replaced by our own "rowSelection column", which also opens up the possibility to use a limel-checkbox for improved styling

The default tabulator behavior of shift+click to create selections is really bad. For example, you cannot manually select row 1, then 2, and then row 8 and shift+click to select from row 8 to row 20. This interaction will result in selecting everything from row 1 to 20. This sucks.

But do we really want to correct them? I'd say it will become a deep rabbit hole, and even if we fix it, upgrading Tabulator to next versions will kill us.

> which also opens up the possibility to use a limel-checkbox for improved styling

Don't worry for styling, and we don't _have to_ use limel-checkbox if it's hard. We can always make their checkbox look like ours.

src/components/table/examples/table-selectable-rows.tsx
adrianschmidt: > If by "should that clear the activeRow immediately" you mean removing the green background highlight? The answer would be YES.
>
> But if you mean "closing the related view", we believe currently believe that the related view should remain open. The green background highlight should only show which item is open in related view.

Sounds to me like you're contradicting yourself here, so either I misunderstood what I read, or you didn't write what you meant 😄

1. You say that the green background highlight should be removed
2. You say that the related view should not be closed
3. You say that a line should be green when (and only when), that item is open in the related view

---

Personally, I think that clicking a checkbox shouldn't change the status of the related-view at all. Whether checking or unchecking a checkbox, that interaction should neither open nor close the related view, and it also shouldn't change what's shown in the related view, if it's already open and showing something else.

src/components/table/examples/table-selectable-rows.tsx
```diff
@@ -0,0 +1,61 @@
...
+        { title: 'Sign', field: 'sign' },
+        { title: 'Date of Birth', field: 'dateOfBirth' },
+        { title: 'Role', field: 'role' },
+    ];
+
+    public render() {
+        return [
+            <limel-table
+                data={this.tableData}
+                activeRow={this.activeRow}
```
Kiarokh, line 38: Currently in the demo, pressing a row (creating an `activeRow`) will deselect all selected rows. @LucyChyzhova and I think it shouldn't be like that, as explained in the comment above. Looking at Lime Go's interaction, it feels better to keep the active row active and keep the selections. These two interactions shouldn't interfere. 


https://user-images.githubusercontent.com/35954987/149308629-905fcdf6-24fc-40e7-a924-80fe519b33dd.mov

src/components/table/table.tsx
Kiarokh: Look at the correct shift+click behavior 

https://user-images.githubusercontent.com/35954987/149308834-543c4b44-1a36-4655-a5fe-2fac2b4ecf37.mov

src/components/table/examples/table-selectable-rows.tsx
adrianschmidt: Came here to say exactly this 😄

This becomes extra annoying when you just accidentally miss a checkbox 😣

Uploading Screen Recording 2022-01-13 at 11.11.51.mov…

anderssonjohan:
An example screen recording of using this branch within the web client to clarify how the related view / active row features now works without reacting to changes of the row selection (keeping the related view open ignoring the fact that activerow is "cleared" upon selection):


https://user-images.githubusercontent.com/435885/149310952-51af3474-08c4-4223-8d03-2cbc2743b9d4.mov

src/components/table/examples/table-selectable-rows.tsx
adrianschmidt: Is there any practical reason `activeRow` has to be touched at all by this functionality? Or is that just because you wanted to highlight the selected rows in green, and then `activeRow` became a conflicting thing?

It looks to me like we could just leave `activeRow` alone if we don't make the selected rows green, and we don't want to make them green anyway 🙂

adrianschmidt: Sorry that I'm butting in so much, but I'm just setting out on adding keyboard navigation for setting `activeRow`, so if the checkbox functionality didn't need to touch the `activeRow` functionality at all, that would be pretty great for me, I think 😄

adrianschmidt: > Yes, I forgot to mention this regarding the suggestion of making our own row selection column/ formatter.
It's not only for being able to use a the same style as the limel-checkbox but also to make sure that if you click anywhere it should just use the checkbox select behavior.

Hmm… 🤔 So, you're saying that if the table has checkboxes, clicking anywhere on a row would check the checkbox for that row? Or do you just mean that clicking anywhere in the _cell_ that holds the checkbox should select the checkbox?

src/components/table/table.tsx
anderssonjohan: > @anderssonjohan asked:
> 
> > built-in Tabulator formatter rowSelection captures the checkbox click event and prevents the expected shift-click extension of the selection when clicking the checkboxes. Shift-click to extend only works when clicking outside the checkbox! This should be replaced by our own "rowSelection column", which also opens up the possibility to use a limel-checkbox for improved styling
> 
> The default tabulator behavior of shift+click to create selections is really bad. For example, you cannot manually select row 1, then 2, and then row 8 and shift+click to select from row 8 to row 20. This interaction will result in selecting everything from row 1 to 20. This sucks.
> 
> But do we really want to correct them? I'd say it will become a deep rabbit hole, and even if we fix it, upgrading Tabulator to next versions will kill us.

Nah, it won't be that bad. The built-in rowSelection formatter does actually keep track of the last clicked row, but the resulting behavior is like you wrote; it selects everything between the first selected row and the one you shift-click.

I'm probably going to make a Tabulator add-on similar to the Knockout custom binding I did (9 years ago! 🤭) in Lime Field: 
https://github.com/Lundalogik/field-web-commons/commit/90eeaf21c0762909868d8d197f2359dcc3602b8b (internal link)

It provides the behavior you describe and we want:

https://user-images.githubusercontent.com/435885/149316074-997b2ff4-cbe6-4312-855e-d192f48ace8b.mov

src/components/table/examples/table-selectable-rows.tsx
anderssonjohan: Roger roger!

1. Then I will remove the part emitting null for activeRow in the selection handler.
2. It is up to the consumer of limel-table to do what they want by using `onActivate` and `onSelect`
3. The data represented by `activeRow` may or may not be in `selection`
4. I'll update the `formatRow` to use the highlight css class `active` for the _selected rows_ only when there's a selection. Thus, it will only highlight `activeRow` when there's no selection; `activeRow` is not always represented visually by having the `active` css class applied to it.

adrianschmidt: > 1. Then I will remove the part emitting null for activeRow in the selection handler.
> 2. It is up to the consumer of limel-table to do what they want by using `onActivate` and `onSelect`
> 3. The data represented by `activeRow` may or may not be in `selection`
> 4. I'll update the `formatRow` to use the highlight css class `active` for the _selected rows_ only when there's a selection. Thus, it will only highlight `activeRow` when there's no selection; `activeRow` is not always represented visually by having the `active` css class applied to it.

I agree with all of those, except point 4. `activeRow` should always be highlighted if there is one, the rows in `selection` should never affect the highlighting. The `selection` status is communicated by the checkboxes.

adrianschmidt: Either the `active` css class shouldn't be set on selected rows, or, if that's set automatically by Tabulator, we should add a separate css class for applying the highlighting for the active row.

adrianschmidt: Aaah… crap… the `activeRow` uses Tabulator's row selection feature… now I understand the depth of the problem… 😞

Maybe we should rethink the whole "select the whole row to open the item it represents" thing? It feels like we're abusing Tabulator by trying to make it work like a list, rather than a table… 🤔

Or perhaps we could at least decouple the "visually highlight to indicate a row is opened in the related view" from the row selection stuff… 🤔

LucyChyzhova: > Personally, I think that clicking a checkbox shouldn't change the status of the related-view at all. Whether checking or unchecking a checkbox, that interaction should neither open nor close the related view, and it also shouldn't change what's shown in the related view, if it's already open and showing something else.

We meant exactly the same. :) Selecting / unselecting checkbox should not affect a related view. The Related view should be changed only if you press on an object (doesn't matter it's checked/selected or not)

Kiarokh: > Sounds to me like you're contradicting yourself here, so either I misunderstood what I read, or you didn't write what you meant 😄

Well, what I wrote is certainly not clear 😄 sorry for that. But if you look at the picture below (from the demo screen recording), it seems when a row is selected, it also has the activeRow style, no matter if it is active or not. Two rows cannot be active in the same time, right? 

image

So yes 👇 
> You say that a line should be green when (and only when), that item is open in the related view

So what I mean is like what you mean Adrian.

src/components/table/table.tsx
Kiarokh: Well. I'm not a developer. So do what's right 😊

src/components/table/examples/table-selectable-rows.tsx
adrianschmidt: Oh! Sorry, no, I'm confused. I thought `tabulator.selectRow` was involved in `activeRow` too, but I was wrong.

Then there shouldn't be too much of a problem I think…? 😅

src/components/table/table.tsx
adrianschmidt: Just don't do _round_ checkboxes! 😝

src/components/table/examples/table-selectable-rows.tsx
anderssonjohan: > Oh! Sorry, no, I'm confused. I thought `tabulator.selectRow` was involved in `activeRow` too, but I was wrong.
> 
> Then there shouldn't be too much of a problem keeping them separate in behaviour, I think…? 😅

I removed all use of Tabulator's own row selection because it does not allow us to keep the selection when the user activates a row.
The reason is the behavior added by this else-block in the SelectRow row click handler that keeps activeRow + selection in sync:
https://github.com/olifolkerd/tabulator/blob/4fc283da83dbf6764a88076c8d808d2bcbc37aee/src/js/modules/SelectRow/SelectRow.js#L133-L134

Doing our own makes things much simpler to keep things separate and also avoid any of the unwanted behaviors from the builtin row selection.

src/components/table/table.tsx
anderssonjohan: > Just don't do _round_ checkboxes! 😝

You're banging on open doors, I'm not the designer of this ;)

src/components/table/examples/table-selectable-rows.tsx
anderssonjohan: > Is there any practical reason `activeRow` has to be touched at all by this functionality? Or is that just because you wanted to highlight the selected rows in green, and then `activeRow` became a conflicting thing?
> 
> It looks to me like we could just leave `activeRow` alone if we don't make the selected rows green, and we don't want to make them green anyway 🙂

Sorry for the late replies. This is fixed and the row selection won't touch the activeRow nor active css class since we decided to keep them separate.

anderssonjohan: > > Yes, I forgot to mention this regarding the suggestion of making our own row selection column/ formatter.
> > It's not only for being able to use a the same style as the limel-checkbox but also to make sure that if you click anywhere it should just use the checkbox select behavior.
> 
> Hmm… 🤔 So, you're saying that if the table has checkboxes, clicking anywhere on a row would check the checkbox for that row? Or do you just mean that clicking anywhere in the _cell_ that holds the checkbox should select the checkbox?

Sorry for the vague reply there. I was referring to the cell. That's also why I removed the user interaction with resizing the checkbox cell. 

The current implementation means that using `<limel-table selectable={true}/>` will add a frozen, non-resizable, checkbox column that will have a click handler on the cell that toggles the selection of the row the cell belongs to. Clicking anywhere in that cell, on the checkbox rendered inside it or in the space between the cell border and the checkbox inside it, will toggle the row's selection.

Using Tabulator's default provided shift-click in any cell to extend the selection. We don't want that because it would interfere with shift-click (to open a new window) in cells providing a link (to a relation for instance).
There are many small details in behavior of Tabulator's own row selection that are unwanted.

src/components/table/partial-styles/_row-selection.scss
```diff
@@ -0,0 +1,81 @@
+#tabulator-table {
+    &.has-rowselector {
+        .tabulator-frozen.tabulator-frozen-left {
```
Kiarokh, line 3: @anderssonjohan I'm not a big fan of targeting the cells with checkboxes, using this CSS. We're simply saying that if a cell is frozen and it's frozen to left, in a table with row-selectors, then apply these styles for that cell. This means if a consumer adds another frozen cell, these styles will affect them too.

Can we apply a specific and unique class to cells that contain the checkboxes? That'd be a better solution.

Kiarokh: for example, we may add a custom frozen cell to limel-table in future which enables keyboard navigation with arrow keys.

anderssonjohan: For sure, this was used to match and override the styles provided by tabulator's default CSS.
Btw, all frozen columns will have `.tabulator-frozen` but only the one on the edge will have `tabulator-frozen-left`. See the frozen column module for reference on this: https://github.com/olifolkerd/tabulator/blob/c3c6bac2d739ff637897a7d51a79115f1c52d5e4/src/js/modules/FrozenColumns/FrozenColumns.js#L186-L188

I'm aware of the possibility of a future problem here, but I don't usually try to make my code aware of future challenges (not the same as writing future proof code!).

If you're saying we are about to use more frozen columns right now/in a near future I think we need to discuss more things, such as how and where these are injected to get them in the right order. 
Right now the table selection feature prepends its checkbox column to the existing set of columns (https://github.com/Lundalogik/lime-elements/pull/1514/files#diff-bab4217d5a11d0c96931f9e9f63bb15ed9564ee64b74eebd6927d68643ea6b49R42)

anderssonjohan: Also, I don't think we are able to add custom css classes to the columns. My opinion is that the left-most (for default left to right text direction, which we only support for now) frozen column should be the row selector checkbox, thus this selector should be fine for now. If we decide to add more frozen columns we may need to skip using the tabulator-frozen-left class, which will select the right-most frozen-to-the-left column. Then we could select the first column with this frozen class instead. 

I'm happy to change this if we have more plans for frozen columns but for now I think we're good.

Kiarokh: Well, I hope we will do the example that I explained. But I don't know when. 

However, if we don't use frozen cells, someone internal or external who is using `limel-table` will do. (now I don't know who, but they may). Our baseline for Elements is that it is a public library and should not eff-up for others. If we were doing this in crm-components, I'd just write this as a comment in the CSS. But since it's elements, I think we should try to mitigate the potential problem now and make it as future-proof as we humanly can.

So if you can inject a class on those cells that contain the checkbox, we don't even have to use Tabulators STUPID frozen things that lag like heck due to their `absolute` positions, and dynamic calculations for `left`, as user scrolls left and right. We will just use a `sticky` position and we will be in control of future too.

anderssonjohan: > Also, I don't think we are able to add custom css classes to the columns.

This was a false statement. Now I found the `cssClass` property of the column definitions which also adds the given class to the row cells (not only the header) in that column. Feel free to change the class name I used! :)

(Nice work on the sticky fix. Now it looks really good when scrolling!)

About frozen columns, no one CAN use it because we don't allow them too. So it's not true that a consumer of limel-table may use frozen columns until we also add support for it in limel-table's column specification. We only allow a subset of all the options that Tabulator provides.

Kiarokh: > no one CAN use it because we don't allow them

Gotcha! But we may allow that too. So better not make it hard for our future-selves :) 

Great job with adding the class! Now I can sleep well 😇

src/components/table/partial-styles/_row-selection.scss
```diff
@@ -0,0 +1,81 @@
...
+                        opacity: 1;
+                    }
+                }
+            }
+        }
+    }
+}
+
+#tabulator-table {
+    // Below code overrides stupid-smartness of Tabulator. /Kia
```
Kiarokh, line 65: We can now remove these comments as well 😄

anderssonjohan: Please do! :)

maltja (changes requested):
Awesome work! 🥳 I just have some small comments. I also discovered a really small visual thing when I tested it (nothing that stops us from merging it). When scrolling sideways to the right the top checkbox (select all, deselect all) loses the 'background' shadow styling.

After scrolling some columns to the right:
Screenshot 2022-02-02 at 03 31 52

After just scrolling a little bit to the right:
Screenshot 2022-02-02 at 03 31 44

Kiarokh:
> Awesome work! 🥳 I just have some small comments. I also discovered a really small visual thing when I tested it (nothing that stops us from merging it). When scrolling sideways to the right the top checkbox (select all, deselect all) loses the 'background' shadow styling.

Thanks 😊 
I'm pretty sure that we had fixed these issues. Have you perhaps thrown away some CSS or changed any class names? 🙂
I'll check soon, but it'd be good to know in advance if you remember something

maltja:
@Kiarokh 
I tested it again, I haven't changed anything and have the latest on the table-selection branch. It is still the same.

src/components/table/table.tsx
```diff
@@ -280,17 +330,33 @@ export class Table {
...
 
         const observer = new ResizeObserver(() => {
             this.tabulator = new TabulatorTable(table, options);
+            this.initSelection();
             observer.unobserve(table);
         });
         observer.observe(table);
     }
 
+    private initSelection() {
```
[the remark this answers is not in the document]
anderssonjohan: Thanks!! initSelection did more things in the beginning and the name was chosen to follow the name of initTabulatorComponent to set up the selection when the tabulator instance became ready.

I'll clean this up a bit.

src/components/table/table-selection.ts
```diff
@@ -0,0 +1,143 @@
...
+                this.selection.toggleSelectionFromLastIndex(rowPosition)
+            );
+        } else {
+            this.updateRowSelectors(
+                this.selection.toggleSelection(rowPosition)
+            );
+        }
+
+        ev.stopPropagation();
+        ev.preventDefault();
```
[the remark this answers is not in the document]
anderssonjohan: Awesome, didn't know we had a convention like that!

src/components/table/table.tsx
```diff
@@ -519,18 +627,45 @@ export class Table {
...
                         'has-movable-columns': this.movableColumns,
+                        'has-rowselector': this.selectable,
+                        'has-selection': this.tableSelection?.hasSelection,
                     }}
                 />
             </div>
         );
     }
 
+    private renderSelectAll() {
```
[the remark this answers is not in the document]
anderssonjohan: Yes, IIRC you were one telling me that we should introduce a specPage/table.spec.ts in a separate PR since we didn't have any tests at all for table.tsx. 
I'm all in on adding such tests (not only for the selection features) separately, but if want feel we need to do that now I can do that too.

src/components/table/selection.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    public has(data: any) {
+        return this.selectedItems.has(data);
+    }
+
+    public toggleSelection(index: number): SelectionChangeSet {
+        return this.toggleRange(index, index);
+    }
+
+    public toggleSelectionFromLastIndex(index: number): SelectionChangeSet {
```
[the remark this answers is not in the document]
adrianschmidt: I don't know the context here (I just saw the comments because I'm subscribed to the whole repo), but you might also want to consider decorating these with `@Method()` too: https://stenciljs.com/docs/methods#method-decorator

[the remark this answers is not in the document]
anderssonjohan: But this is not a component so I won't add `@Method()` to these. :)

maltja (approved):
Approved! 🙂

mjlime (approved):

Kiarokh:
Sorry for the delay guys. I've been in meetings all day long. Great job! Happy you finally merged this! Weehoooo! 🎉 🎊 

I don't see that problem you mentioned @maltja though :/ 

I just see this one 👇 

https://user-images.githubusercontent.com/35954987/152130720-55c87bf1-3fd9-4b2d-b002-b6857f7765dd.mov

which I fixed now here: https://github.com/Lundalogik/lime-elements/pull/1539


However, @anderssonjohan said that this problem is created when the viewport is smaller than ~50% of the scroll width. I haven't been able to reproduce it yet. It'd be great if you guys call me tomorrow and show me. I we inspect it on your machines.

src/components/table/selection.ts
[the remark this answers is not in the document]
adrianschmidt: # 🤦‍♂️😆✅

PR #1516, opened 2022-01-14
feat(checkbox): indeterminate state support

anderssonjohan:
https://user-images.githubusercontent.com/435885/149546729-b2edc3c7-8b0d-4cda-a547-dea43a0028e5.mov

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/checkbox/checkbox.template.tsx
+++ b/src/components/checkbox/checkbox.template.tsx
@@ -4,6 +4,7 @@ interface CheckboxTemplateProps {
     disabled?: boolean;
     id: string;
     checked?: boolean;
+    indeterminate?: boolean;
     required?: boolean;
     invalid?: boolean;
     onChange?: (event: Event) => void;
@@ -14,6 +15,11 @@ interface CheckboxTemplateProps {
 export const CheckboxTemplate: FunctionalComponent<CheckboxTemplateProps> = (
     props
 ) => {
+    const inputProps = {};
+    if (props.indeterminate) {
+        inputProps['data-indeterminate'] = 'true';
+    }
+
     return (
         <div class="mdc-form-field ">
             <div
@@ -22,6 +28,7 @@ export const CheckboxTemplate: FunctionalComponent<CheckboxTemplateProps> = (
                     'mdc-checkbox--invalid': props.invalid,
                     'mdc-checkbox--disabled': props.disabled,
                     'mdc-checkbox--required': props.required,
+                    'mdc-checkbox--indeterminate': props.indeterminate,
                 }}
             >
                 <input
@@ -32,6 +39,7 @@ export const CheckboxTemplate: FunctionalComponent<CheckboxTemplateProps> = (
                     disabled={props.disabled}
                     required={props.required}
                     onChange={props.onChange}
+                    {...inputProps}
                 />
                 <div class="mdc-checkbox__background">
                     <svg class="mdc-checkbox__checkmark" viewBox="0 0 24 24">
@@ -49,6 +57,7 @@ export const CheckboxTemplate: FunctionalComponent<CheckboxTemplateProps> = (
                     'mdc-checkbox--invalid': props.invalid,
                     'mdc-checkbox--disabled': props.disabled,
                     'mdc-checkbox--required': props.required,
+                    'mdc-checkbox--indeterminate': props.indeterminate,
                 }}
                 htmlFor={props.id}
             >
--- a/src/components/checkbox/checkbox.tsx
+++ b/src/components/checkbox/checkbox.tsx
@@ -94,10 +94,6 @@ export class Checkbox {
         this.mdcCheckbox.indeterminate = newValue;
     }
 
-    constructor() {
-        this.onChange = this.onChange.bind(this);
-    }
-
     public connectedCallback() {
         this.initialize();
     }
@@ -141,9 +137,9 @@ export class Checkbox {
         );
     }
 
-    private onChange(event: Event) {
+    private onChange = (event: Event) => {
         event.stopPropagation();
         this.change.emit(this.mdcCheckbox.checked);
         this.modified = true;
-    }
+    };
 }
--- a/src/components/checkbox/checkbox.tsx
+++ b/src/components/checkbox/checkbox.tsx
@@ -8,6 +8,7 @@ import {
     h,
     Prop,
     State,
+    Watch,
 } from '@stencil/core';
 import { createRandomString } from '../../util/random-string';
 import { CheckboxTemplate } from './checkbox.template';
@@ -55,6 +56,12 @@ export class Checkbox {
     public checked = false;
 
     /**
+     * Enables indeterminate state. Set to `true` to signal indeterminate check.
+     */
+    @Prop({ reflect: true })
+    public indeterminate = false;
+
+    /**
      * Set to `true` to indicate that the checkbox must be checked.
      */
     @Prop({ reflect: true })
@@ -75,6 +82,17 @@ export class Checkbox {
     private formField: MDCFormField;
     private mdcCheckbox: MDCCheckbox;
     private id: string = createRandomString();
+
+    @Watch('checked')
+    protected handleCheckedChange(newValue: boolean) {
+        this.mdcCheckbox.checked = newValue;
+    }
+
+    @Watch('indeterminate')
+    protected handleIndeterminateChange(newValue: boolean) {
+        this.mdcCheckbox.checked = this.checked;
+        this.mdcCheckbox.indeterminate = newValue;
+    }
 
     constructor() {
         this.onChange = this.onChange.bind(this);
@@ -113,7 +131,8 @@ export class Checkbox {
                 disabled={this.disabled || this.readonly}
                 label={this.label}
                 helperText={this.helperText}
-                checked={this.checked}
+                checked={this.checked || this.indeterminate}
+                indeterminate={this.indeterminate}
                 required={this.required}
                 invalid={this.required && this.modified && !this.checked}
                 onChange={this.onChange}
--- a/src/components/checkbox/examples/checkbox.tsx
+++ b/src/components/checkbox/examples/checkbox.tsx
@@ -42,8 +42,8 @@ export class CheckboxExample {
                         />
                         <limel-checkbox
                             checked={this.value}
-                            label="Toggle checked"
-                            onChange={this.toggleChecked}
+                            label="Checked"
+                            onChange={this.setChecked}
                         />
                     </limel-flex-container>
                 </p>
@@ -66,7 +66,7 @@ export class CheckboxExample {
         this.required = event.detail;
     };
 
-    private toggleChecked = (event: CustomEvent<boolean>) => {
+    private setChecked = (event: CustomEvent<boolean>) => {
         event.stopPropagation();
         this.value = !this.value;
     };
--- a/src/components/checkbox/examples/checkbox.tsx
+++ b/src/components/checkbox/examples/checkbox.tsx
@@ -15,6 +15,11 @@ export class CheckboxExample {
     @State()
     private value: boolean = false;
 
+    @State()
+    private indeterminate: boolean = false;
+
+    private eventPrinter: HTMLLimelExampleEventPrinterElement;
+
     public render() {
         return (
             <section>
@@ -24,6 +29,7 @@ export class CheckboxExample {
                         label="My fab checkbox"
                         id="fab"
                         checked={this.value}
+                        indeterminate={this.indeterminate}
                         required={this.required}
                         onChange={this.handleChange}
                     />
@@ -45,15 +51,35 @@ export class CheckboxExample {
                             label="Checked"
                             onChange={this.setChecked}
                         />
+                        <limel-checkbox
+                            checked={this.indeterminate}
+                            label="Indeterminate"
+                            onChange={this.setIndeterminate}
+                        />
                     </limel-flex-container>
                 </p>
-                <limel-example-value value={this.value} />
+                <limel-example-value label="Checked" value={this.value} />
+                <limel-example-value
+                    label="Indeterminate"
+                    value={this.indeterminate}
+                />
+                <limel-example-event-printer
+                    ref={(el) => (this.eventPrinter = el)}
+                />
             </section>
         );
     }
 
     private handleChange = (event: CustomEvent<boolean>) => {
         this.value = event.detail;
+
+        // The only way a user can interact with the checkbox is to check it or
+        // uncheck it. The indeterminate state can only be set programmatically
+        // and will always be unset when the user interacts with the checkbox.
+        // Therefore, we must set indeterminate to `false` here.
+        this.indeterminate = false;
+
+        this.eventPrinter.writeEvent(event);
     };
 
     private setDisabled = (event: CustomEvent<boolean>) => {
@@ -68,6 +94,11 @@ export class CheckboxExample {
 
     private setChecked = (event: CustomEvent<boolean>) => {
         event.stopPropagation();
-        this.value = !this.value;
+        this.value = event.detail;
+    };
+
+    private setIndeterminate = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.indeterminate = event.detail;
     };
 }


anderssonjohan:
The minus part with this change lies in the icon. #sealapproved

anderssonjohan:
Sorry about pinging for reviews. Converted to draft now because I noticed that the div is lacking the equivalent css class we have for disabled/required + it's not mutable after render (can't toggle a rendered limel-checkbox by `querySelector('limel-checkbox').indeterminate = true`).

src/components/checkbox/checkbox.template.tsx
```diff
@@ -4,6 +4,7 @@ interface CheckboxTemplateProps {
     disabled?: boolean;
```
adrianschmidt, line 4: Replying here so comments can be threaded:

> Sorry about pinging for reviews. Converted to draft now because I noticed that the div is lacking the equivalent css class we have for disabled/required + it's not mutable after render (can't toggle a rendered limel-checkbox by `querySelector('limel-checkbox').indeterminate = true`).

No worries! 🙂

You _can_ use `querySelector('limel-checkbox').indeterminate = true` to change the state of the checkbox. I just did in the docs for this PR. My guess is that you're not finding the checkbox because `document.querySelector('limel-checkbox')` returns `null`, because `querySelector` doesn't penetrate shadow-roots. _But_, I'm having trouble seeing when this might be a legit way to control state? Surely this state should be held somewhere it can be controlled without finding the checkbox through the DOM?

src/components/checkbox/examples/checkbox.tsx
```diff
@@ -70,4 +83,9 @@ export class CheckboxExample {
         event.stopPropagation();
         this.value = !this.value;
     };
+
+    private toggleIndeterminate = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.indeterminate = !this.indeterminate;
```
adrianschmidt, line 89:
```suggestion
        this.indeterminate = event.detail;
```

src/components/checkbox/examples/checkbox.tsx
```diff
@@ -70,4 +83,9 @@ export class CheckboxExample {
         event.stopPropagation();
         this.value = !this.value;
     };
+
+    private toggleIndeterminate = (event: CustomEvent<boolean>) => {
```
adrianschmidt, line 87:
```suggestion
    private setIndeterminate = (event: CustomEvent<boolean>) => {
```

src/components/checkbox/examples/checkbox.tsx
```diff
@@ -45,9 +49,18 @@ export class CheckboxExample {
                             label="Toggle checked"
                             onChange={this.toggleChecked}
                         />
+                        <limel-checkbox
+                            checked={this.indeterminate}
+                            label="Toggle indeterminate"
+                            onChange={this.toggleIndeterminate}
```
adrianschmidt, line 55:
```suggestion
                            onChange={this.setIndeterminate}
```

src/components/checkbox/examples/checkbox.tsx
```diff
@@ -45,9 +49,18 @@ export class CheckboxExample {
                             label="Toggle checked"
                             onChange={this.toggleChecked}
                         />
+                        <limel-checkbox
+                            checked={this.indeterminate}
+                            label="Toggle indeterminate"
```
adrianschmidt, line 54: Oh, I guess the other checkboxes are also called "toggle…". I guess these suggestions are better applied when a little boyscouting has been done 😄

Anyway:

```suggestion
                            label="Indeterminate"
```

(I think the names comes from us originally using buttons, rather than checkboxes. For checkboxes, it makes more sense to skip the "toggle" bit.)

src/components/checkbox/checkbox.template.tsx
anderssonjohan: @adrianschmidt I was testing this as well when you wrote this comment and I can also properly update the state by changing the example to use `@Element` and `this.element.shadowRoot.querySelector('limel-checkbox').indeterminate`. I didn't actually come across that error in this branch but when using a limel-checkbox for the "select all" checkbox inside tabulator. I know we had some similar issues when manipulating Lime Elements components from FullCalendar (read DOM manipulation and not stencil re-render after state/prop changes), so I thought it was the same thing going here. I'll need to investigate my changes to limel-table where the checkbox won't update properly. Thanks for testing and confirming it's working!

adrianschmidt: But wait, why are you updating the state of a component this way in the first place? There really seems to be something wrong if you need to perform DOM manipulation rather than setting the value of a `@Prop`? Can you send me a link to the code you're talking about? (The table stuff preferably, I'll try to keep out of FullCalendar for now, if possible 😅)

src/components/checkbox/examples/checkbox.tsx
anderssonjohan: Does it make sense to you if I also change toggleChecked to setChecked and remove the word toggle?

In my head, the tristate (not checked, checked, indeterminate) can be represented in many ways. 
Three ways are: 
1. enum 
2. a nullable boolean (null = indeterminate) 
3. using two boolean values (like MDC)

To me, indeterminate is part of the checked state and not part of other features such as disabled/required. Setting indeterminate will make the component to ignore the checked value altogether and I think we should use the same verb here (toggle, set or nothing) for both these properties.

src/components/checkbox/checkbox.template.tsx
anderssonjohan: I'm not sure if this is the case with my callback from Tabulator, I think it's something else, so never mind for now.  :)
In the table column header I need to return a DOM element, so I use `document.createElement('limel-checkbox')` to add the select all checkbox.

However, for the sake of FullCalendar it has no Stencil connector and uses preact internally so it's more similar to the case in the limel-table where we want to avoid re-render and toggle CSS instead (the table loader element for instance). 

What's important here is my concern about setting props manually (https://stenciljs.com/docs/javascript#setting-the-prop-manually), which both you and I have confirmed working.

src/components/checkbox/examples/checkbox.tsx
anderssonjohan: I'm having the exact same thoughts and then when I found out that MDC already got support using the "two boolean" variant I was really happy because adding it doesn't break anything.

I think the props and events can co-exist as separate, even though the visual effect of setting `indeterminate` will hide the visual effect of the state of `checked`. I didn't see the need of it at this moment, but an event for indeterminate could be added later on.

Consider this test case:

1. User clicks checkbox, sets `checked`, emits `changed` => `true`, visually shows `[v]`
2. User clicks something that causes the consumer code to set `indeterminate`, emits `indeterminateChanged` => `true`, visually shows `[-]`
3. User clicks checkbox, unsets `checked` and `indeterminate`, emits `changed` => `false` and `indeterminateChanged` => `false`, visually shows `[ ]`

Given this test case I need to make it untick the checkbox when clicking it. Now it remains in indeterminate state, which I find unexpected.

So, shortly, when interacting (keyboard spacebar/mouse click) with a checkbox with indeterminate state twice will make it transition: indeterminate -> not checked -> checked.
Please note that it only applies for user interaction. Modifying the `checked` prop `true` => `false` programmatically is not the same. 

The correct user interaction event handler would therefore ought to be: 

* with indeterminate set: unset checked and indeterminate
* with indeterminate unset: toggle checked

anderssonjohan: PS. I don't mean to add an event for indeterminateChanged now. Only implement "unset indeterminate on user interaction".

anderssonjohan: Also, what regards the events, while the event detail of the `change` event represents the current `checked` state, it will still be emitted for any change (disabled, required, indeterminate).

So without adding a separate event for indeterminate (we don't have any for disabled/required), I need to update the checkbox example indeterminate checkbox from that event handler to keep it in sync.
This have not been required for disabled/required because they can't be modified by user interaction while indeterminate can change from `true` => `false`, assuming the rules I stated above are correct.

I still think the `change` event is enough, it'll just make the example code a bit wonky not having a separate event + I need to make the indeterminate prop mutable.

adrianschmidt: > I'm having the exact same thoughts and then when I found out that MDC already got support using the "two boolean" variant I was really happy because adding it doesn't break anything.

👍

> I think the props and events can co-exist as separate, even though the visual effect of setting `indeterminate` will hide the visual effect of the state of `checked`.

Yeah, I see no problem with that. We have the same thing where we have several components that can be either disabled or readonly. In many cases, a readonly component is _also_ disabled, just styled differently, so whether or not you explicitly set `disabled` doesn't matter for the appearance of the component.

> I didn't see the need of it at this moment, but an event for indeterminate could be added later on.
>
> Consider this test case:
> 
> 1. User clicks checkbox, sets `checked`, emits `changed` => `true`, visually shows `[v]`
> 2. User clicks something that causes the consumer code to set `indeterminate`, emits `indeterminateChanged` => `true`, visually shows `[-]`
> 3. User clicks checkbox, unsets `checked` and `indeterminate`, emits `changed` => `false` and `indeterminateChanged` => `false`, visually shows `[ ]`

I know this isn't your main point here, but I still want to explain why the latter half of step two would never happen:

The component should never emit events for state changes _initiated by the consumer_. If the user clicks the unchecked checkbox, we need to emit an event to tell the consumer that the state should change (probably, that's up to the consumer, really). But if the consumer sets `checked = true` programatically, the consumer already knows the new state, and so we don't emit an event about that. Since there's nothing the user can do to a checkbox, to set _that same_ checkbox to `indeterminate`, the case where we need to emit an event never appears.

Sure, if we perform DOM-manipulation, and argument could be made for emitting an event, but honestly, if DOM-manipulation is performed for any legitimate reason, I would expect that it's the consumer that's performing it, which means the consumer already knows about the new state, which in turn means we shouldn't emit an event for it.
 
> Given this test case I need to make it untick the checkbox when clicking it. Now it remains in indeterminate state, which I find unexpected.

Yeah, no, we definitely need to set indeterminate to false when the checkbox is clicked 👍
 
> So, shortly, when interacting (keyboard spacebar/mouse click) with a checkbox with indeterminate state twice will make it transition: indeterminate -> not checked -> checked.

I agree 👍

> Please note that it only applies for user interaction. Modifying the `checked` prop `true` => `false` programmatically is not the same.

Yeah, it's probably best to only change the value of the property actually changed when changed programatically, and not change the value of `indeterminate` just because `checked` was changed.

> The correct user interaction event handler would therefore ought to be:
> 
> * with indeterminate set: unset checked and indeterminate
> * with indeterminate unset: toggle checked

Or to be more verbose:

* checked: true, indeterminate: true -> checked: false, indeterminate: false - emit change event with value `false`
* checked: false, indeterminate: true -> checked: false, indeterminate: false - emit change event with value `false`
* checked: true, indeterminate: false -> checked: false, indeterminate: false - emit change event with value `false`
* checked: false, indeterminate: false -> checked: true, indeterminate: false - emit change event with value `true`

👍

adrianschmidt: > Also, what regards the events, while the event detail of the `change` event represents the current `checked` state, it will still be emitted for any change (disabled, required, indeterminate).
> 
> So without adding a separate event for indeterminate (we don't have any for disabled/required), I need to update the checkbox example indeterminate checkbox from that event handler to keep it in sync. This have not been required for disabled/required because they can't be modified by user interaction while indeterminate can change from `true` => `false`, assuming the rules I stated above are correct.
> 
> I still think the `change` event is enough, it'll just make the example code a bit wonky not having a separate event + I need to make the indeterminate prop mutable.

The change event should not be emitted when the checkbox becomes `indeterminate`. I think my last answer probably explains why, and why the indeterminate prop shouldn't have to be mutable either (I think, again, _possibly_ with the exception for supporting DOM-manipulation, but let's cross that bridge when we determine we really do need to). If you disagree when you've read it, please comment about it again, and we'll discuss from there 👍

anderssonjohan: How about required/disabled? (we emit the change event when those are modified even though the user can't interact with those props)

adrianschmidt: 🤨

I think (or hope) that you're confusing the change-handlers in the example controls for the events actually sent by limel-checkbox. limel-checkbox only send one type of events, according to the docs, and that's `change`.

The example component has checkboxes for triggering programmatic changes of the state. They have event handlers called things like `setRequired` and `setDisabled`, but they listen to the `change` event from their respective checkbox, neither of which are the checkbox "being demoed". It becomes a little confusing in this case, because both the component we are demoing and the components we use to control the demo, are limel-checkbox'es 😬

anderssonjohan: Oh, yeah! Thanks for the heads up. I was totally having my breakpoint in the onChange handler of the checkbox class and not the change handler in the example file. I thought I was going crazy there for a moment. I got the change events for the checkboxes used in the example.

src/components/checkbox/checkbox.tsx
```diff
@@ -113,7 +119,8 @@ export class Checkbox {
                 disabled={this.disabled || this.readonly}
                 label={this.label}
                 helperText={this.helperText}
-                checked={this.checked}
+                checked={this.checked || this.indeterminate}
```
anderssonjohan, line 122: We want the MDC checkbox checked when indeterminate to force the transition indeterminate->unchecked.
If not, the MDC checkbox will both remove the indeterminate state and toggle checked upon interaction.
For an example of this, see the example at https://kradio3.github.io/react-mdc-web/components/checkbox/ and:

1. check the checkbox
2. click "make indeterminate"
3. click the checkbox to remove indeterminate state
4. checkbox gets the unchecked state
5. click "make indeterminate"
6. click the checkbox to remove the indeterminate state
7. checkbox gets the checked state

By forcing `checked={true}` will force the wanted unchecked state when clicking an indeterminate checkbox.

src/components/checkbox/checkbox.scss
```diff
@@ -39,6 +39,20 @@
...
+            don't show the checkmark when transitioning from mixedmark (indeterminate)
+            to unchecked. Expected behavior is to show indeterminate -> unchecked
+            without showing the checkmark during the transition.
+            The checked->unchecked selector is used because the checkbox transitions
+            from indeterminate->checked->unchecked when the user clicks a
+            checkbox when it is in the indeterminate state.
+            note: transition indeterminate->checked is still animated on purpose
+            */
+            animation-name: none;
+        }
```
anderssonjohan, line 55: A clip of the annoying effect when this CSS override is left out:
https://user-images.githubusercontent.com/435885/149826966-7484174b-6230-422f-9876-0bde6431aba7.mov

src/components/checkbox/examples/checkbox.tsx
anderssonjohan: After some further research I came to the conclusion that the MDC checkbox:
1. always unsets indeterminate on interaction
2. always toggles checked on interaction

Now the code is free from mutable props etc and the change event is emitted as expected. I also added an event printer to the example to make it super clear how the change event works with checked+indeterminate.

src/components/checkbox/checkbox.tsx
anderssonjohan: After a second thought, maybe we don't want to limit the checkbox like this? 🤔  If I remove the change on this line a consumer can do as they want, just like one may do with the MDC checkbox. Do you agree about removing this change is better @adrianschmidt ? (less opinionated)

src/components/checkbox/checkbox.scss
anderssonjohan: @Kiarokh maybe setting animation-name to none like this can work for your loading animation issue too?

Kiarokh: @anderssonjohan not sure what the discussion is about here. I didn't read the comments, as I feel I'm not the right person to have opinions here. But I suggest that you leave all animation problems out of this PR. When this is merged, I'll go through the checkbox component and fix/improve/disable the animations in a separate PR. :)

src/components/checkbox/checkbox.tsx
adrianschmidt: I've got to look at the effects of either choice, and I'm holding a trainee workshop today so might not get to that until later, but my gut instinct is to go with _more_ opinionated 😄

After all, we want not just looks, but also behaviour, to be streamlined and consistent when using Lime Elements.

anderssonjohan: In that case I think this is ready for review again. Let me know if I need to move some things in my commits. The boyscout commit is there so I'm hoping it's like you asked for.

src/components/checkbox/checkbox.scss
anderssonjohan: @Kiarokh it was just a tip for you if your change of keyframes didn't work out. Removing the animation name may be another solution for you, that was all :)

src/components/checkbox/checkbox.scss
```diff
@@ -39,6 +39,20 @@
                 }
```
adrianschmidt, line 39: I'll assign myself and put this on the Feature team's cycle board too, so it's visible to my team that I'm reviewing this, and so that I don't forget 😄

src/components/checkbox/checkbox.template.tsx
```diff
@@ -14,6 +15,11 @@ interface CheckboxTemplateProps {
 export const CheckboxTemplate: FunctionalComponent<CheckboxTemplateProps> = (
     props
 ) => {
+    const indeterminateProps = {};
```
adrianschmidt, line 18: I think it makes more sense that this would be called `inputProps` 🤔

src/components/checkbox/examples/checkbox.tsx
```diff
@@ -15,6 +15,11 @@ export class CheckboxExample {
     @State()
     private value: boolean = false;
 
+    @State()
+    private indeterminate: boolean = false;
```
adrianschmidt, line 19: Hmm… 🤔

If I set this to `true`, I can't programmatically set the checkbox to _not_ indeterminate again 😮

Going for lunch now, will look into it again when I'm back if you haven't already found something out by then 😄

anderssonjohan: That was a bug I had for two days and then I haven't been able to reproduce it after restarting the stencil dev server. Please please show me :)

src/components/checkbox/checkbox.template.tsx
anderssonjohan: For sure! Good suggestion, I'll rename it.

src/components/checkbox/examples/checkbox.tsx
anderssonjohan: Ok, so my idea worked. The MDCCheckbox sets the property (not attribute) indeterminate=true on the native input and that will keep the indeterminate style because it still matches the selector "nativeinput:indeterminate". Only updating the CSS classes mdc-checkbox--indeterminate was not enough when using indeterminate={true} on the initial render. 

The fix I applied isn't pretty but it works. Instead of manually resetting the indeterminate state of the native element I added a watcher that unsets indeterminate on the MDCCheckbox instance instead (letting it do the job of toggling the state, which also unsets things like `aria-checked="mixed"`).

src/components/checkbox/checkbox.tsx
```diff
@@ -134,4 +135,14 @@ export class Checkbox {
         this.change.emit(this.mdcCheckbox.checked);
         this.modified = true;
     }
+
+    @Watch('indeterminate')
+    private onIndeterminateChanged(newValue) {
```
adrianschmidt, line 140: If you set this to protected, eslint will stop complaining.

src/components/checkbox/checkbox.tsx
```diff
@@ -134,4 +135,14 @@ export class Checkbox {
...
+        // To properly remove the indeterminate state of the rendered
+        // input native control by the CheckboxTemplate, it's required
+        // to set indeterminate = false on the MDCCheckbox instance.
+        if (!newValue) {
```
adrianschmidt, line 144: I think I'd prefer that we always set `this.mdcCheckbox.indeterminate`, rather than only set it when it becomes disabled. I know it means we'll have to do something about the horrendous animation when going from `indeterminate` to `checked`, but I don't much like the idea of having inconsistent state throughout the component.

Or am I being too pedantic? 🤔

adrianschmidt: I mean… I _was_ ok with not touching `this.mdcCheckbox.indeterminate` at all, so adding this shouldn't really make a difference to me…

😖

src/components/checkbox/checkbox.tsx
anderssonjohan: I noticed we used protected everywhere else but didn't know it was because of eslint. Good catch, I'll update it!

src/components/checkbox/checkbox.tsx
anderssonjohan: I only added the if-statement to avoid extra calls/renders but it also adds complexity to the code. I removed it now because it works just fine without it.

Kiarokh:
I unsubscribed from this PR. But let me know on Slack if anything concerns me here :)

src/components/checkbox/checkbox.tsx
adrianschmidt: Quoted from resolved thread:

> The fix I applied isn't pretty but it works. Instead of manually resetting the indeterminate state of the native element I added a watcher that unsets indeterminate on the MDCCheckbox instance instead _**(letting it do the job of toggling the state, which also unsets things like `aria-checked="mixed"`).**_

I just checked, and since we now update `this.mdcCheckbox.indeterminate` both when true and when false, MDC sets `aria-checked="mixed"` for us when indeterminate is true, not just when it's true from first render. So that's very good 👍

adrianschmidt: I actually removed the comment, because it doesn't feel like we need to explain why we update the state of the MDC component. That's kind of the default, and there's more reason to comment on why if we _don't_ update it than if we _do_ 🙂

I also moved the watcher. We're a little inconsistent about where we place watchers. Originally I used to place all public functions first, then protected functions (which is only ever watchers), and last private functions. But Stencil's own guidelines say to place watchers immediately after the property they are watching. I don't like mixing the watchers and the props, as that makes it harder to see what props exist, but I have begun to place the watchers above the other functions, so they're at least close to the props they watch… I don't really know what's best to be honest. Consistency I guess, and consistent is the only thing we're not at the moment… 😣 well well…

src/components/checkbox/checkbox.tsx
```diff
@@ -83,8 +83,14 @@ export class Checkbox {
...
 
+    @Watch('checked')
+    protected handleCheckedChange(newValue: boolean) {
+        this.mdcCheckbox.checked = newValue;
+    }
+
     @Watch('indeterminate')
-    protected onIndeterminateChanged(newValue: boolean) {
+    protected handleIndeterminateChange(newValue: boolean) {
+        this.mdcCheckbox.checked = this.checked;
```
anderssonjohan, line 93:
```suggestion
        this.mdcCheckbox.checked = this.checked || newValue;
```

Shouldn't we force checked here if we're only allowing the interaction indeterminate => unchecked (and never indeterminate => checked)?

adrianschmidt: We do allow indeterminate -> checked, if indeterminate is unset programmatically.

If the consumer changes the value of `indeterminate` back and forth, without changing the value of `checked`, I didn't want to put `this.mdcCheckbox.checked` in a state that doesn't match `this.checked`, for fear of causing incorrect transitions somewhere else.

adrianschmidt (approved):

PR #1520, opened 2022-01-20
fix(checkbox): remove ripple effect

anderssonjohan:
Removes the ripple effect of the checkbox.

It becomes a problem sometimes, for example when the checkbox is used in a table and all the checkboxes ripple animations triggers at the same time:

https://user-images.githubusercontent.com/435885/150353673-84de1575-e8cd-48b2-afb6-d19c36164e8f.mov

This PR removes the ripple while still maintaining the keyboard focus indicator.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -27,11 +27,8 @@
                         content: '';
                         display: block;
                         position: absolute;
-                        top: -0.75rem;
-                        right: -0.75rem;
-                        bottom: -0.75rem;
-                        left: -0.75rem;
-                        border-radius: 50%;
+                        inset: -0.25rem;
+                        border-radius: 0.25rem;
                         box-shadow: var(--shadow-depth-8-focused);
                     }
                 }


Kiarokh:
related to: https://github.com/Lundalogik/crm-feature/issues/1425

src/components/checkbox/checkbox.scss
```diff
@@ -20,21 +20,19 @@
...
+                + .mdc-checkbox__background {
+                    &:after {
+                        content: '';
+                        display: block;
+                        position: absolute;
+                        top: -0.75rem;
+                        right: -0.75rem;
+                        bottom: -0.75rem;
+                        left: -0.75rem;
+                        border-radius: 50%;
```
Kiarokh, line 34:
```suggestion
                        inset: -0.25rem
                        border-radius: 0.25rem;
```

Kiarokh: The PR looks great. So happy we're removing these ripples finally! 🎉

We should also do this in a separate commit, where we decrease the amount of space waste for that visual effect of keyboard focus.

Kiarokh (approved):
great job! rip ripple

anderssonjohan:
Final result:
https://user-images.githubusercontent.com/435885/150385890-da98d3e3-2e10-469b-b886-3f414e77fdd0.mov

src/components/checkbox/checkbox.scss
adrianschmidt: I think that the "space waste" isn't because of the visual effect, but in order to provide a big enough size as a touch target.

Kiarokh: You're right about that Adrian. But we didn't change the touch area or anything. We just manipulated our custom `:after` element that is only used for visualization of the keyboard `focus`.

PR #1521, opened 2022-01-21
fix(checkbox): make a solid background color

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/2612
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -14,6 +14,17 @@
     flex-wrap: wrap;
 
     .mdc-checkbox {
+        .mdc-checkbox__native-control {
+            &[disabled],
+            &:enabled {
+                &:not(:checked):not(:indeterminate) {
+                    ~ .mdc-checkbox__background {
+                        background-color: rgb(var(--contrast-300));
+                    }
+                }
+            }
+        }
+
         &.mdc-checkbox--invalid {
             .mdc-checkbox__native-control:enabled:not(:checked):not(:indeterminate)
                 ~ .mdc-checkbox__background {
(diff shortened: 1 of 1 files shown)


src/components/checkbox/checkbox.scss
```diff
@@ -14,10 +14,21 @@
...
+                    }
+                }
+            }
+        }
+
         &.mdc-checkbox--invalid {
             .mdc-checkbox__native-control:enabled:not(:checked):not(:indeterminate)
                 ~ .mdc-checkbox__background {
-                border-color: var(--lime-error-text-color);
+                border-color: var(--lime-red-text-color);
```
adrianschmidt, line 31: Hmm… isn't "error" the one we should use for invalid stuff? Or rather, maybe we should create a separate variable that's called something with "invalid". Generally, we want to avoid using the names of colours in variables like this, since, if we one day decide that we want to use a different colour for invalid, say purple for example, it would make no sense for it to be called `--lime-red-text-color`.

adrianschmidt, line 31: Also, this change doesn't seem to be related to the background color change? Should it perhaps be in a separate commit?

Kiarokh: this must have happened by accident  :)

LucyChyzhova: yes, you are right, I have changed it accidentally by testing something... fixed now

Kiarokh (approved):

src/components/checkbox/checkbox.scss
```diff
@@ -35,16 +46,15 @@
...
+    }
 
     label {
         cursor: pointer;
         line-height: normal;
         padding-left: 0;
 
-        &.mdc-checkbox--disabled {
-            opacity: 0.37;
-        }
```
adrianschmidt, line 46: Was this change intentional? 😊

LucyChyzhova: yes :) 0.37 was too transparent

Issue #1522, opened 2022-01-24
limel-table does not adjust row height to height of custom component


anderssonjohan:
## New feature motivation

When using custom components, Tabulator normalizes the row height before the component renders, not taking the height of the custom component into account.

## New feature description

- Extend the existing resizeobserver that sets the column width based on the custom component width to also check the row height

anderssonjohan:
Tried to fix this in https://github.com/Lundalogik/lime-elements/compare/1522-check-row-height-after-custom-component-render but calling `row.normalizeHeight()` or `cell.checkHeight()` during the observed resize or after all rows are rendered will totally mess up the scrolling. The row height is rendered correctly, but as soon as the table reaches the bottom (when scrolling), the table is scrolled to top once per row that is adjusted (all rows in the limel-table example here).

justFredster:
Would also be beneficial for the text snippets feature in the integration of shared email accounts.

![image](https://github.com/user-attachments/assets/804b62dd-2fb0-42bb-bfdd-2bdc3691b667)

PR #1523, opened 2022-01-25
feat(table): adjust row height to custom component

anderssonjohan:
#1522

tell tabulator to check the row height when the custom component have rendered.

Note: this is just a draft because it seems to interfere with Tabulator's vertical VDom when scrolling. 
The solution works for the initial rows but as soon as more rows are added on scroll (row virtualization) changing the height of these rows while also the VDom code sets the scrollTop makes the table scroll to top over and over again while trying to scroll down to the bottom of the table. 

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/columns.ts
+++ b/src/components/table/columns.ts
@@ -180,7 +180,7 @@ export function createCustomComponent(
     element.style.display = 'inline-block';
     setElementProperties(element, props);
 
-    createResizeObserver(element, cell.getColumn());
+    createResizeObserver(element, cell);
 
     return element;
 }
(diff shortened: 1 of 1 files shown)


anderssonjohan:
A screen recording from https://lundalogik.github.io/lime-elements/versions/PR-1523/#/debug/limel-example-table-custom-components showing the scrolling error.

https://user-images.githubusercontent.com/435885/151012957-791f9f7b-8d83-4cb5-8617-5996a94e23eb.mov

The reason for this behavior seems to be because of the "large scroll" that triggers a table redraw that also resets the scroll top to zero.

I read somewhere that you should set the vdom buffer to 5 times your largest row height. So when using `virtualDomBuffer: 310,` in the docs example (and using large icons) I'm not able to get the issue. It's however impractical to set a fixed buffer size because we never know what the row heights may be.

There are numerous changes in Tabulator v5 and we're using v4.9, so I think I will try an upgrade to v5 and then see if the problem persist.

Kiarokh:
closing due to this history: https://github.com/Lundalogik/lime-elements/pull/3401

PR #1524, opened 2022-01-26
fix(input-field): make sure label has our default label color

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/2624

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/select/select.scss
+++ b/src/components/select/select.scss
@@ -42,7 +42,9 @@
 }
 
 .limel-select__selected-option__text {
-    @include shared_input-select-picker.looks-like-input-label;
+    .mdc-select:not(.mdc-select--disabled) & {
+        @include shared_input-select-picker.looks-like-input-label;
+    }
     @include mixins.truncate-text;
 }
 
@@ -72,6 +74,7 @@
         }
 
         .mdc-floating-label {
+            color: shared_input-select-picker.$label-color;
             width: calc(
                 100% - #{functions.pxToRem(16)}
             ); //This forces the label to truncate when container is too little.
--- a/src/components/slider/slider.scss
+++ b/src/components/slider/slider.scss
@@ -7,11 +7,18 @@
 @use '@material/slider/styles';
 @use '@material/floating-label/mdc-floating-label';
 
+:host([disabled]:not([readonly])) {
+    .slider__label {
+        color: shared_input-select-picker.$label-color-disabled;
+    }
+}
+
 .slider {
     position: relative;
 }
 
 .slider__label {
+    color: shared_input-select-picker.$label-color;
     padding-left: functions.pxToRem(20);
     top: functions.pxToRem(
         9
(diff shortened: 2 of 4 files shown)


anderssonjohan (dismissed):
Using the dialog example here (enter a name + age = 30), https://lundalogik.github.io/lime-elements/versions/PR-1524/#/debug/limel-example-dialog-form, the number input value is still in another grey color than the text input (name field).

Screen Shot 2022-01-27 at 11 08 12

src/components/slider/slider.scss
```diff
@@ -7,11 +7,18 @@
 @use '@material/slider/styles';
 @use '@material/floating-label/mdc-floating-label';
 
+:host([disabled]) {
```
adrianschmidt, line 10: With the above, there's a difference if the component is `readonly` _and_ `disabled`, or _just_ `readonly`:

https://user-images.githubusercontent.com/154725/151988254-56f37344-e5ae-4813-bb3d-5079bc0d11e1.mov

It's preferable if the readonly state looks the same whether or not the component also has the `disabled` attribute set:

```suggestion
:host([disabled]:not([readonly])) {
```

src/components/select/select.scss
```diff
@@ -79,6 +79,7 @@
                 100% - #{functions.pxToRem(16)}
             ); //This forces the label to truncate when container is too little.
             &.mdc-floating-label--float-above {
+                font-size: shared_input-select-picker.$cropped-label-hack--font-size;
```
adrianschmidt, line 82: I'm not sure if the font just happened to be the correct size already, and this was just to avoid a potential future bug, if we or MDC changes the font size, or if I'm not looking in the right place, but I'm not seeing any difference in the floating label size for `limel-select` in this branch vs `next`? 🤔

src/style/internal/shared_input-select-picker.scss
```diff
@@ -53,7 +53,7 @@ $cropped-label-hack--font-size: 0.875rem; //14px
     line-height: functions.pxToRem(28);
     -moz-osx-font-smoothing: grayscale;
     -webkit-font-smoothing: antialiased;
-
+    color: $input-text-color;
```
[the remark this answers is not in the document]
adrianschmidt: I've updated the commit message for one.

src/components/select/select.scss
LucyChyzhova: Yes, the font-size should not be touched in this PR, seems to be done by mistake

src/style/internal/shared_input-select-picker.scss
```diff
@@ -54,6 +54,7 @@ $cropped-label-hack--font-size: 0.875rem; //14px
     -moz-osx-font-smoothing: grayscale;
     -webkit-font-smoothing: antialiased;
 
+    color: $input-text-color;
```
adrianschmidt, line 57: I squashed this into the fix for `limel-input-field`, as that's the component that was affected by this change too. They're effectively both "making sure label has our default label color" in limel-input-field 🙂

adrianschmidt (approved):

PR #1527, opened 2022-01-27
bug repro test: example showing mdc checkbox disconnect flicker issue

anderssonjohan:
example showing how the mdc checkbox anim css classes causes
flickering when checkbox is disconnected on click redrawing the DOM.

## Review:
- [ ] Commits are [atomic](https://seesparkbox.com/foundry/atomic_commits_with_git)
- [ ] Commits have the correct *type* for the changes made
- [ ] Commits with *breaking changes* are marked as such

### Browsers tested:
(Check any that applies, it's ok to leave boxes unchecked if testing something didn't seem relevant.)

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

PR #1528, opened 2022-01-28
fix(checkbox): clean up animation classes before reconnected

anderssonjohan:
fix #1527  (that PR contains a visual repro that should not be added to the examples, hence this separate PR with the fix)

when the checkbox is disconnected and the mdc checkbox
instance is destroyed, it will prematurely remove the
animation end handler that removes all the animation
classes mdc adds when interacting with the checkbox.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/checkbox/checkbox.tsx
+++ b/src/components/checkbox/checkbox.tsx
@@ -102,20 +102,6 @@ export class Checkbox {
         this.initialize();
     }
 
-    private initialize() {
-        const element =
-            this.limelCheckbox.shadowRoot.querySelector('.mdc-form-field');
-        if (!element) {
-            return;
-        }
-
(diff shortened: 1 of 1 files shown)


adrianschmidt (approved):
Works like a charm!

I added two tiny refactor commits, because I want the public function up top ([newspaper code structure](https://kentcdodds.com/blog/newspaper-code-structure)).

PR #1529, opened 2022-01-28
Specify a default max-width for limel-menu

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/2623

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu-surface/menu-surface.scss
+++ b/src/components/menu-surface/menu-surface.scss
@@ -13,4 +13,9 @@
     width: var(--menu-surface-width, auto);
     max-height: 100%;
     position: relative;
+:host(limel-menu-surface.has-grid-layout) {
+    .mdc-menu-surface {
+        width: var(--menu-surface-width, min(calc(100vw - 2rem), 40rem));
+        max-width: unset;
+    }
 }
--- a/src/components/menu-surface/menu-surface.scss
+++ b/src/components/menu-surface/menu-surface.scss
@@ -10,9 +10,14 @@
 @include menu.core-styles;
 
 .mdc-menu-surface {
-    width: var(--menu-surface-width, auto);
     max-height: 100%;
     position: relative;
+    --mdc-menu-max-width: var(
+        --menu-surface-width,
+        min(calc(100vw - 2rem), 20rem)
+    );
+}
+
 :host(limel-menu-surface.has-grid-layout) {
     .mdc-menu-surface {
         width: var(--menu-surface-width, min(calc(100vw - 2rem), 40rem));
--- a/src/components/menu/examples/menu-grid.tsx
+++ b/src/components/menu/examples/menu-grid.tsx
@@ -3,15 +3,16 @@ import { Component, h } from '@stencil/core';
 
 /**
  * With grid layout
- * To render items of a menu in a gird layout instead of a vertical list,
- * two things are required:
- * 1. Simply setting the `gridLayout` property to `true`, and
- * 1. Specifying a proper width for the menu surface, using the
- * `--menu-surface-width` variable.
+ * To render items of a menu in a grid layout instead of a vertical list,
+ * simply setting the `gridLayout` property to `true`.
  *
  * :::note
- * Do not forget that menus should behave responsively, thus using a fixed `width`
- * should be avoided. To make the width responsive, try using the `min()` function.
+ * Menus with the grid layout has a responsive width by default,
+ * which will not grow wider than a certain size. However, if the default size is not
+ * wide enough for your use case, you can try setting another responsive width, using
+ * the `--menu-surface-width` variable.
+ *
+ * To achieve a responsive width, try using the `min()` function.
  * This function selects the smallest value from a list of comma-separated expressions
  * which are placed within the parentheses.
  *
--- a/src/components/menu/examples/menu-basic-secondary-text.tsx
+++ b/src/components/menu/examples/menu-basic-secondary-text.tsx
@@ -8,11 +8,11 @@ import { Component, h, State } from '@stencil/core';
  * will be displayed in two lines, and then get truncated.
  *
  * :::important
- * Keep in mind that a menu's drop-down will stretch as much as it can to display
- * its content. You must therefore specify a maximum width for menus with
- * long secondary text, otherwise they will fill out the entire screen width.
+ * Keep in mind that a menu's drop-down surface will stretch as much as its default
+ * maximum width values allow. However, if this default maximum width does not suit
+ * your use case, you can override it using the `--menu-surface-width` variable.
  *
- * But do not forget that menus should behave responsively, thus using a fixed `width`
+ * But do not forget that menus should still behave responsively, thus assigning a fixed value
  * should be avoided. To make the width responsive, try using the `min()` function.
  * This function selects the smallest value from a list of comma-separated expressions
  * which are placed within the parentheses.
--- a/src/components/menu/menu.scss
+++ b/src/components/menu/menu.scss
@@ -4,7 +4,7 @@
 
 /**
  * @prop --dropdown-z-index: `z-index` of the dropdown menu.
- * @prop --menu-surface-width: Width of the menu surface. Defaults to `auto`.
+ * @prop --menu-surface-width: Width of the menu surface.
  * @prop --list-grid-item-max-width: Maximum width of items in the menu list when `gridLayout={true}`. Defaults to `10rem`.
  * @prop --list-grid-item-min-width: Minimum width of items in the menu list when `gridLayout={true}`. Defaults to `7.5rem`.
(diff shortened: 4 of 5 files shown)


src/components/menu-surface/menu-surface.scss
```diff
@@ -10,9 +10,14 @@
 @include menu.core-styles;
 
 .mdc-menu-surface {
-    width: var(--menu-surface-width, auto);
     max-height: 100%;
     position: relative;
+    --mdc-menu-max-width: var(
```
Kiarokh, line 15: I was thinking maybe we should mark this commit as a breaking change. This is probably no problem at all, but there might be someone who have had huge menus (due to long names on menu items) and they were happy with the size because their users could read the text.

Now after this their menu will render smaller, and grow max to 20 rem. (unless they override it). 

I don't know how much of a big deal this is. But it wouldn't hurt to add this as a new paragraph to the commit message:

````md
BREAKING CHANGE: menus' dropdown lists will not render wider than `20rem`, and their text will be truncated. If you have menus in which the whole menu item label has to be displayed for the end user, make sure to increase the menu width and read our docs to learn how to do it in a responsive way.
````

File: src/components/menu-surface/menu-surface.scss [hunk 2]
```diff
@@ -10,7 +10,17 @@
 @include menu.core-styles;
```
adrianschmidt, line 10: Please use `menu` as the scope for these commits, not `menu-surface`. `limel-menu-surface` is not a public component.

adrianschmidt (changes requested):

src/components/menu/examples/menu-grid.tsx
```diff
@@ -3,15 +3,16 @@ import { Component, h } from '@stencil/core';
...
- * 1. Simply setting the `gridLayout` property to `true`, and
- * 1. Specifying a proper width for the menu surface, using the
- * `--menu-surface-width` variable.
+ * To render items of a menu in a grid layout instead of a vertical list,
+ * simply setting the `gridLayout` property to `true`.
  *
  * :::note
- * Do not forget that menus should behave responsively, thus using a fixed `width`
- * should be avoided. To make the width responsive, try using the `min()` function.
+ * The menu surface of menus with the grid layout has a responsive width by default,
```
adrianschmidt, line 10: "The menu surface of menus" can be shortened to just "Menus":

```suggestion
 * Menus with the grid layout has a responsive width by default,
```

Although, that's true for non-grid menus too, it's just the maximum width that differs…

src/components/menu/examples/menu-grid.tsx
```diff
@@ -3,15 +3,16 @@ import { Component, h } from '@stencil/core';
...
+ * which will not grow wider than a certain size. However, if the default size is not
+ * wide enough for your use case, you can try setting another responsive width, using
+ * the `--menu-surface-width` variable.
```
adrianschmidt, line 13: I think it's a pity that we've gone with `--menu-surface-width`. I feel like `--menu-max-width` had been more concise and to the point. Perhaps we should deprecate `--menu-surface-width` in favour of `--menu-max-width`, but we can just as well do that in a separate PR…

When naming things that is seen by the consumer, please try to take one step back, and look at what the consumer, or the developer of that consumer, sees. If in doubt, please do reach out for a second opinion 😊

adrianschmidt: (I know the name was already used before this PR, just mentioning it here, since I commented about not using "the surface of menus" above.)

src/components/menu-surface/menu-surface.scss
adrianschmidt: Yeah, the breaking change might be a good idea. It feels like a case right on the line of what's usually considered breaking and not, but I think it's better to err on the side of caution. It's better that a change marked as breaking doesn't break anything for anyone, than that a change _not_ marked as breaking _does_ break something for someone 😅

adrianschmidt: "If you have menus in which the whole menu item label has to be displayed for the end user"

We can't really guarantee that the whole item will be shown anyway, as this is just setting the max-width, right? So this is a little misleading. But Monday morning Adrian doesn't quite have the energy to come up with a better formulation. This is good _enough_, I think.

src/components/menu-surface/menu-surface.scss
```diff
@@ -10,7 +10,17 @@
...
     position: relative;
+    --mdc-menu-max-width: var(
+        --menu-surface-width,
+        min(calc(100vw - 2rem), 20rem)
+    );
+}
+
+:host(limel-menu-surface.has-grid-layout) {
+    .mdc-menu-surface {
+        width: var(--menu-surface-width, min(calc(100vw - 2rem), 40rem));
```
adrianschmidt, line 23: Hmm… above we use `--mdc-menu-max-width`, but here we actually set the `width`, not the `max-width`. How come?

src/components/menu-surface/menu-surface.scss
```diff
@@ -10,7 +10,17 @@
...
+        max-width: unset;
```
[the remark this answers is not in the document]
Kiarokh: Regarding previous comment and this one, you're right! A name like `--menu-max-width` would have been more semantically correct and definitely a better name. But we were afraid someone may have used this variable (`--menu-surface-width`) in their code. Because menus are so widely used in all solutions. 

Of course one way would be deprecating the previous variable before going forward. But that comes with its pains, and would be a big breaking change.

Additionally, the usage of this custom variable in the previous code was not so nice neither. Because it was put on the `width` property. It would mean that any value that consumers would put there, would give the menus a hard-coded and static width. 

So even if their menu item labels were too short (lets say in their English solution versus Finish), the menu component would show them a lot of dead space. Basically a wide menu with short text labels.

Then we thought we could actually use the custom variable (ignoring the lack of perfection in its name) and put it in the `--mdc-menu-max-width` instead, and combine it with a `min` function. This way, we would make sure menus will remain responsive, even if the consumer has put a static value for the `--menu-surface-width`. Because now, we are applying it to the `max-width` property, instead of the `width` property.


I hope the story of our decisions is more clear now. Otherwise just ask or let's have a live talk :)

What do you suggest? Should we depreciate `--menu-surface-width` and rename it to `--menu-max-surface-width`? Or is this one OK?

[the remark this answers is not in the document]
Kiarokh: > Edit: Oh, and we even unset the max-width… 🤔

This is because we are saying `width: auto;` and `max-width: the fancy variabel stuff…` on normal menus. 

While in grid menus, we'd need a static (but still responsive) width, otherwise they will not render in a grid mode. The static `width` will need to use the `min` function to become responsive. And if we inherit the `max-width` from the normal menus, any value used for  `width` that is larger than that inherited `max-width` will not work. 

It's a bit of a mess conceptually, but I think we have handled it well. And believe me, we ourselves were totally 🧠 💨ing the whole way 😄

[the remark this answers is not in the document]
adrianschmidt: > > Edit: Oh, and we even unset the max-width… 🤔
> 
> This is because we are saying `width: auto;` and `max-width: the fancy variabel stuff…` on normal menus.
> 
> While in grid menus, we'd need a static (but still responsive) width, otherwise they will not render in a grid mode. The static `width` will need to use the `min` function to become responsive. And if we inherit the `max-width` from the normal menus, any value used for `width` that is larger than that inherited `max-width` will not work.
> 
> It's a bit of a mess conceptually, but I think we have handled it well. And believe me, we ourselves were totally 🧠 💨ing the whole way 😄

Ah! Ok, then I understand 👍

> Regarding previous comment and this one, you're right! A name like `--menu-max-width` would have been more semantically correct and definitely a better name. But we were afraid someone may have used this variable (`--menu-surface-width`) in their code. Because menus are so widely used in all solutions.
> 
> Of course one way would be deprecating the previous variable before going forward. But that comes with its pains, and would be a big breaking change.
> 
> […snip…]
> 
> I hope the story of our decisions is more clear now. Otherwise just ask or let's have a live talk :)
> 
> What do you suggest? Should we depreciate `--menu-surface-width` and rename it to `--menu-max-surface-width`? Or is this one OK?

I understand and agree with the reasoning 👍

Deprecating the variable is not the same as removing it. When deprecating, we make both the old and the new variable name work, with the new taking precedence if both are specified. We also mark the old as deprecated in documentation, and use the new one in any examples. Then the deprecated variable can eventually be removed in some future version, but _that_ would indeed be a breaking change.

I don't think we should complicate _this_ PR further by adding the changes necessary for deprecating the variable. We can do that in a separate PR if we want to.

adrianschmidt (approved):

Issue #1525, opened 2022-01-27
Add padding in the footer of the dialog (mobile view)


Befkadu1:
Currently it looks like the below pic:
<img width="295" alt="Screenshot 2022-01-27 at 13 47 30" src="https://user-images.githubusercontent.com/22968497/151364471-83c969f9-9c91-49dc-8b52-45afd1bfad0e.png">

Suggestion(possible to add 24px in the `<footer/>`):

<img width="288" alt="Screenshot 2022-01-27 at 13 51 03" src="https://user-images.githubusercontent.com/22968497/151364554-e8baf268-c66a-4416-aa8e-331c2d8bedf3.png">

PR #1531, opened 2022-01-30
fix(dialog): add padding to the footer in mobile view

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/1525

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dialog/dialog.scss
+++ b/src/components/dialog/dialog.scss
@@ -120,4 +120,7 @@ slot[name='button'] {
     slot[name='button'] {
         flex-direction: column-reverse;
     }
+    .mdc-dialog__actions {
+        padding: 1rem 1.5rem 1.5rem 1.5rem;
+    }
 }


src/components/dialog/dialog.scss
```diff
@@ -120,4 +120,7 @@ slot[name='button'] {
     slot[name='button'] {
         flex-direction: column-reverse;
     }
+    .mdc-dialog__actions {
+        padding: 0.5rem 1.5rem;
```
adrianschmidt, line 124: Shouldn't there also be increasing padding below the buttons?

image

Or does that eat too much valuable vertical screen space?

LucyChyzhova: yes 👍, let's have the same padding below the last button

src/components/dialog/dialog.scss
```diff
@@ -120,4 +120,7 @@ slot[name='button'] {
     slot[name='button'] {
         flex-direction: column-reverse;
     }
+    .mdc-dialog__actions {
+        padding: 1rem 1.5rem 1.5rem;
```
adrianschmidt, line 124: I think this shorthand is horrible to read. I think the 4 and 2 value one is bad enough, but with this one, I never have _any_ clue what value will be applied where 😭

```suggestion
        padding: 1rem 1.5rem 1.5rem 1.5rem;
```

adrianschmidt (approved):

PR #1533, opened 2022-01-31
fix(list): add margin to the "limel-list" for proper keyboard visualization

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/2629

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/list/list.scss
+++ b/src/components/list/list.scss
@@ -33,6 +33,9 @@ $list-border-radius: 0.375rem; // 6px
         --icon-color,
         rgb(var(--contrast-900))
     );
+    margin: functions.pxToRem(
+        4
+    ); // added space to visualize keyboard-focused items
     padding: 0;
     border-radius: $list-border-radius;
 


adrianschmidt (approved):

src/components/list/list.scss
```diff
@@ -33,6 +33,8 @@ $list-border-radius: 0.375rem; // 6px
         --icon-color,
         rgb(var(--contrast-900))
     );
+    margin: functions.pxToRem(4);
+    // added space to visualize keyboard-focused items
```
adrianschmidt, line 37: We usually put comments either _above_ the code the comment concerns, or at the end of the line it concerns (Kia's preferred way). But putting the comment _after_ the line it concerns is a little odd, I think…

Also, I generally prefer not to have comments that explain what code does. If that's necessary, the code should preferably be refactored to be more self-explanatory instead, for example by extracting the complicated code to a named function. We don't use that much in scss, probably mostly out of habit from not having functions in css. But since scss has both mixins and functions, I guess there really is no reason why we couldn't just as well extract one or more css properties to a function or mixin, with good descriptive name, like `addSpaceForKeyboardFocusVisualization`… 🤷‍♂️

adrianschmidt: ⚡

adrianschmidt (approved):

src/components/list/list.scss
LucyChyzhova: yeah, I agree with you, but the comment was moved by ` run lint:fix` and it even looked worst (👇). so I put a comment to the next line and run the lint again and it has passed so I decided to keep this layout. But you are right 'above' or with the 'descriptive name' would be better. 
image

adrianschmidt: I agree it's not beautiful, but "looks worse" is better than "harder to understand" 😊

I placed the comment on the same line and let the linter do its job, so what was merged is exactly as what your screenshot shows. But that's ok 👍

PR #1538, opened 2022-02-02
feat(table): add aggregates property

mjlime:
## Problem

Updating aggregate values only is not currently possible without changing the columns and re-initialize the table.

## Solution

By adding a new `@Prop` called `aggregates` the user of `<limel-table>` can let the table re-render with new aggregate values. A watcher of the added prop can call Tabulator's `recalc` function to update the values in its calculation row (`bottomCalc`).

Note: The aggregator given by the column definition should have higher priority than the value in the `aggregates` prop. 

## Use case

This feature is added to support updating the aggregate value as the user selects rows using the `selection` prop added in #1514.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -8,7 +8,12 @@ import {
     Event,
 } from '@stencil/core';
 import TabulatorTable from 'tabulator-tables';
-import { Column, TableParams, ColumnSorter } from './table.types';
+import {
+    Column,
+    TableParams,
+    ColumnSorter,
+    ColumnAggregate,
+} from './table.types';
 import { ColumnDefinitionFactory, createColumnSorter } from './columns';
 import { isEqual, has } from 'lodash-es';
 import { ElementPool } from './element-pool';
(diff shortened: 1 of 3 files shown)


src/components/table/table.tsx
```diff
@@ -279,6 +291,24 @@ export class Table {
         this.init();
     }
 
+    @Watch('aggregates')
+    protected updateAggregates(
+        newAggregates: ColumnAggregate[],
+        oldAggregates: ColumnAggregate[]
+    ) {
+        if (isEqual(newAggregates, oldAggregates)) {
```
anderssonjohan, line 299: I think it's good to add a check for this.tabulator here, to be consistent with other functions

mjlime: @anderssonjohan Fixed!

maltja (approved):

anderssonjohan (approved):

PR #1544, opened 2022-02-03
fix(table): never emit internal row selector column

anderssonjohan:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -266,11 +266,15 @@ export class Table {
             return;
         }
 
-        const existingColumns = this.tabulator
+        const columnsInTable = this.tabulator
             .getColumns()
-            .map(this.findColumn);
+            .filter((c) => c.getField());
 
-        if (this.areSameColumns(newColumns, existingColumns)) {
+        const oldColumnsInTable = columnsInTable.map((c) =>
+            oldColumns.find((old) => old.field === c.getField())
+        );
+
+        if (this.areSameColumns(newColumns, oldColumnsInTable)) {
             return;
         }
 
@@ -593,7 +597,7 @@ export class Table {
     };
 
     private handleMoveColumn = (_, components: Tabulator.ColumnComponent[]) => {
-        const columns = components.map(this.findColumn);
+        const columns = components.map(this.findColumn).filter((c) => c);
         this.changeColumns.emit(columns);
     };
 


src/components/table/table.tsx
```diff
@@ -593,10 +591,16 @@ export class Table {
     };
 
     private handleMoveColumn = (_, components: Tabulator.ColumnComponent[]) => {
-        const columns = components.map(this.findColumn);
+        const columns = this.findColumns(components);
         this.changeColumns.emit(columns);
     };
 
+    private findColumns = (
```
maltja, line 598: I think that if a function starts with find in its name it involves some kind of search. In this case the function should just return all columns that is not undefined. Therefore, I think getColumns would be a better name 🤔 What do you think?

src/components/table/examples/table-selectable-rows.tsx
```diff
@@ -60,19 +65,25 @@ export class TableExampleSelectableRows {
                 onClick={this.clearSelection}
             />,
             <limel-button label="Reset" onClick={this.resetSelection} />,
+            <limel-switch
```
maltja, line 68: This is really good for testing purposes but is it really relevant for showing the select feature? 🤔

maltja (changes requested):
nice, I just have some small comments 🙂

src/components/table/examples/table-selectable-rows.tsx
anderssonjohan: I removed it now but the reason it was relevant is because we add a column behind the scenes and the column change event doesn't state that you won't get a column for the row selector. But it may not be necessary to explain that the event will only emit the contents of the columns prop (it may be in another order but never have other items that what is put in the prop)

adrianschmidt: If the row selector is a column we add behind the scenes, I think it's (probably) something that should stay behind the scenes, and thus be filtered out from change-events and the like. (General thoughts; I haven't looked at the specifics in this case.)

maltja (approved):

maltja (approved):

PR #1547, opened 2022-02-04
fix(icon-button): apply the default background color if it's not specified

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/2630

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/icon-button/icon-button.scss
+++ b/src/components/icon-button/icon-button.scss
@@ -9,10 +9,12 @@
 }
 
 /**
- * @prop --icon-background-color: Background color when attribute `elevated` is set to `true`.
+ * @prop --icon-button-background-color: Background color when attribute `elevated` is set to `true`.
  */
 
 :host {
+    --specified-background-color: var(--icon-button-background-color),
+        (var(--icon-background-color));
     display: inline-flex;
     align-items: center;
     justify-content: center;
@@ -33,6 +35,10 @@
 
     limel-icon {
         @include mixins.is-elevated-clickable;
+        background-color: var(
+            --specified-background-color,
+            rgb(var(--contrast-100), 0.8)
+        );
     }
 }
 


src/components/icon-button/icon-button.scss
```diff
@@ -33,6 +37,10 @@
 
     limel-icon {
         @include mixins.is-elevated-clickable;
+        background-color: var(
+            --specified-background-color,
+            rgb(var(--contrast-100))
+        );
```
Kiarokh, line 41: I think a better default value is `rgb(var(--contrast-100), 0.8)` or something like that. This way, if the icon-button is used on a colored background (let's say a header), it'll still blend in quite well. Because some of the background color will leak through the button, and the consumer doesn't have to necessarily change the icon-button's background color.

src/components/icon-button/icon-button.scss
```diff
@@ -9,10 +9,12 @@
 }
 
 /**
- * @prop --icon-background-color: Background color when attribute `elevated` is set to `true`.
+ * @prop --icon-button-background-color: Background color when attribute `elevated` is set to `true`.
  */
 
 :host {
+    --specified-background-color: var(--icon-button-background-color),
+        (var(--icon-background-color));
```
adrianschmidt, line 17: Does this work? Shouldn't it be:

```suggestion
    --specified-background-color: var(--icon-button-background-color,
        var(--icon-background-color));
```

Kiarokh: When did we make this PR? Totally forgot about it. Anyhow, today, we change this `var(--lime-elevated-surface-background-color)` in any component that needs another color for their elevated surface.

PR #1565, opened 2022-02-07
update @stencil/sass from v1.4.1 to v1.5.2 (plus replacing some now deprecated syntax in two scss files)

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "@rjsf/core": "^2.5.1",
         "@stencil/core": "^2.10.0",
         "@stencil/eslint-plugin": "^0.4.0",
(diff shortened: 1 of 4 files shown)


adrianschmidt:
Nope. It seems we can't update this without also updating MDC. So all-at-once it is 😅

PR #1566, opened 2022-02-07
Update to MDC 13

adrianschmidt:
This also updates `@stencil/sass`, since updating either package alone breaks the build (MDC 11 uses some old scss syntax that is no longer supported in the newer Sass version, and MDC 13 uses some new scss feature that isn't supported by the older Sass version 🙄😄)

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/switch/switch.tsx
```diff
@@ -86,28 +88,44 @@ export class Switch {
 
     public render() {
         return [
-            <div
+            <button
+                id={this.fieldId}
                 class={{
                     'mdc-switch': true,
-                    'mdc-switch--disabled': this.disabled || this.readonly,
                     'lime-switch--readonly': this.readonly,
```
Kiarokh, line 95: The design of Switch has changed a lot in their new version, which basically cancels our customizations. Specially fr the `readonly` mode. We will have to go through this

adrianschmidt: I'd love it if either 1) we think it's ok if the looks change for a little while, so we can merge this and fix the styling later, or 2) we can prioritise fixing the styling so we can get this merged relatively soon.

The above assuming that we don't find any _other_ major stoppers of course. So please continue testing this for changes and bugs, before jumping on fixing the styling for limel-switch 😊

jgroth (commented):
Nice! :slightly_smiling_face: 
I've tested it and it seems to work as expected. I just have one small question about a comment that maybe should be removed

jgroth (approved):

src/components/switch/switch.tsx
adrianschmidt: @Kiarokh Ping on ☝️

Kiarokh: @adrianschmidt I've gone through most of the components, and nothing catches my attention as a bug yet. So we can merge this if you want. I'm working on the switch now. If I manage to fix the switch in this PR, I'd push it as a fixup. If you merge it first, I'll add it as a new PR ;)

src/components/switch/switch.tsx
```diff
@@ -86,28 +88,44 @@ export class Switch {
...
+                                viewBox="0 0 24 24"
+                            >
+                                <path d="M19.69,5.23L8.96,15.96l-4.23-4.23L2.96,13.5l6,6L21.46,7L19.69,5.23z" />
+                            </svg>
+                            <svg
+                                class="mdc-switch__icon mdc-switch__icon--off"
+                                viewBox="0 0 24 24"
+                            >
+                                <path d="M20 13H4v-2h16v2z" />
+                            </svg>
```
Kiarokh, line 124: image

Even though I understand where they come from, I dislike the visual mess these icons create. I think the direction and color are enough for users to realize that they are ON or OFF. 

- We don't care that much about RTL layout neither, do we? 
- And I don't think we want anyone to use a grey color as their accent color (to make the switch confusing), right?

Kiarokh:
> I've tested it and it seems to work as expected. I just have one small question about a comment that maybe should be removed

Which one @jgroth ? 


A side note…
Maybe a stupid question, but I myself wondered if `limel-tab-bar` should be a private component, and we move its examples to tab panel? Will anyone use tab bar alone? Or will people only use `limel-tab-panel`? 

Additionally, I wish we would make `limel-flex-container` private.

jgroth:
> Which one @jgroth ?

It's been fixed by @adrianschmidt :slightly_smiling_face:

jgroth:
> Additionally, I wish we would make `limel-flex-container` private.

Could we not just remove it then? Do we even use it internally? :smile:

Kiarokh:
> > Additionally, I wish we would make `limel-flex-container` private.

> Could we not just remove it then? Do we even use it internally? :smile: 

Oh yes, both we, packers, hackers, and many consultants and everyone is using it sadly. I always remove it when I see it somewhere and tell people not to use it. But it's widely used.

src/components/switch/switch.tsx
adrianschmidt: I'd guess the icons are a matter of being colour-blind friendly?

I actually kind of like the icons. Since we're using green as the "on" colour, and I am fully capable of distinguishing green from grey, I find our switches _fairly_ clear anyway. At least when _on_. When off, if there are no "on" switches to compare with, the state is probably less obvious. But I've seen switches that are light grey (dull) when "off", and white (bright and shiny) when "on". I think my Android phone has that colour scheme. And even though I know "left" typically means "off", and "right" typically means "on", _every single bloody time_ I come across one of those switches, I have to flip it once to check if it looks "more on" or "more off" in the other position, just to be sure 🤮

And while it's easy to say "those switches are obviously using a much worse colour scheme than ours are", keep in mind that _someone_ must have thought those switches are clear enough, just like we think _our_ switches are clear enough. We see our switches every day, and know by heart what the "opposite" state looks like. While we can make an effort to imagine our components from the view of a first time user of them, it's impossible for any of us to truly rid ourselves of that contextual knowledge.

src/components/switch/switch.tsx
adrianschmidt: Good stuff! I decided to squash them into `chore: fixes for MDC 13`. Since we don't really announce to the user of lime-elements what version of MDC we're using, I didn't think that fixes that are meant mainly to make the v13 switch look like the v11 switch was something that needed listing in the changelog…

src/components/switch/switch.tsx
jgroth: > I have to flip it once to check if it looks "more on" or "more off" in the other position, just to be sure

I also do this a lot :smile:

Kiarokh:
https://github.com/Lundalogik/crm-feature/issues/2659

https://github.com/Lundalogik/crm-feature/issues/2660

src/components/switch/switch.tsx
Kiarokh: > This opinion probably betrays my age, right?

😸  certainly not. I think you're young enough to have experienced both. 

But it's super interesting that you find this toggle switch confusing. 🤔  I really can't explain why, but I haven't used android for at least 5 years now. I just know that I always disliked details of material design, in comparison with iOS. I also always strongly disliked their switches and can't really say why. (maybe that's why I tried to make the switch slightly different this time). 

But this switch you showed is more like the iOS style. The only problem I'd say is that the thumb is greyed-out which to me looks **disabled-and-off**. 

Maybe that's why it makes you confused too? 🤔 

I guess to be able to explain why you are confused by the switch, we need to check how the switches look like and behave across their platform and apps. Probably there is an inconsistency there that creates this issue.

Kiarokh: I remember Android settings has checkboxes, radios and switches all over the place and at leas the time I worked with that at Sony Mobile, it was a very confusing and inconsistent thing. There was an Android guideline when to use checkboxes vs. switches though. They said whenever there is a SAVE button, you can use a checkbox, and if the thing auto-applies or auto-saves, use switches. But do you think anyone in Sony Mobile (who customized the hell of Android, and had their own Sony apps) knew or followed that guideline?

iOS settings on the other hand don't have these variety of things. It's mostly switches and a very specific way of having radios (which don't look like radios 😄 ). 👇 

image

Anyhow...

The only legitimate explanation for me is that switches probably confuse the hell out of RTL users, since some of the apps don't care to adjust their UI elements and some do and the direction in which the switch is ON or OFF (along with those inconsistencies) make them very confused. So a checkmark is probably a good addition.

PR #1590, opened 2022-02-10
feat(form): add support for extra errors

jgroth:
This adds a new `errors` prop that can be used to display extra errors in the form, e.g. errors
coming from an API endpoint etc.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [x] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/form/form.types.ts
```diff
@@ -36,6 +36,13 @@ export interface FormError {
     message: string;
 }
 
+export type ValidationError = {
+    /**
+     * Name of the field the error belongs to
+     */
+    [key: string]: string[] | ValidationError;
+};
```
jgroth, line 44: I was thinking a bit about both the name of this and what the best structure would be, `FormError` is already taken and has a structure that is not optimal for this use case. It would have been nice if we could have the same format both here and where `FormError` is used right now, but it will probably cause more work for both consumers and ourselves since we have to convert the errors between different formats.

The format here is very close to what Rjsf wants and is also the same format that Marshmallow will return when there is a validation error so I think this will be a good middle ground.

civing (approved):
Looks good!

Just one small comment is that the limel example for villains should have more padding. Right now the error text are overlapping the name of the next limel-input-field. This is a problem inherited from the already existing example though.

jgroth:
> Looks good!
> 
> Just one small comment is that the limel example for villains should have more padding. Right now the error text are overlapping the name of the next limel-input-field. This is a problem inherited from the already existing example though.

Thanks!

Yes, this is an issue that is already present in the form even without any errors. We have discussed it before and there might be an issue about it already :slightly_smiling_face:

PR #1619, opened 2022-02-15
feat(checkbox): add support for readonly state

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/2672

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/checkbox/checkbox.template.tsx
+++ b/src/components/checkbox/checkbox.template.tsx
@@ -4,6 +4,7 @@ interface CheckboxTemplateProps {
     disabled?: boolean;
     id: string;
     checked?: boolean;
+    readonly?: boolean;
     indeterminate?: boolean;
     required?: boolean;
     invalid?: boolean;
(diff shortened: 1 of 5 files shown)


src/components/checkbox/checkbox.template.tsx
```diff
@@ -38,6 +40,7 @@ export const CheckboxTemplate: FunctionalComponent<CheckboxTemplateProps> = (
                     checked={props.checked}
                     disabled={props.disabled}
                     required={props.required}
+                    readonly={props.readonly}
```
Kiarokh, line 43:
```suggestion
                    disabled={props.disabled || props.readonly}
                    required={props.required}
```

Kiarokh: I think this is better to do. We don't need to make the `input` readonly. I don't think it makes sense. But we can make it disabled, in case readonly = true

adrianschmidt: Yes, I don't think `input type="checkbox"` understands the attribute `readonly`, so `disabled` should be used. The difference is only visual (and perhaps if there's any difference to what aria attributes that should be used (I'm not sure)).

Kiarokh (approved):
Awesome job. Good use of CSS variables! 👏

PR #1625, opened 2022-02-17
docs(dialog): add note about opening dialogs in Lime CRM

adrianschmidt:
Prompted by a request from @DavidAkkerman:

> Almost a year ago, you helped me solve a problem with the rendering of a dialog in the web client, by advising me to use the dialogservice for creating the dialog. I am now getting questions from several other appers that are running into the same problem, because the first thing everyone does is going to the lime-elements documentation page and copy-paste some code from the examples, which most of the time works really well. However, if you go to the documentation of the dialog, the example just shows that you can render a button and a dialog in the same web component, and that you only have to set the dialog as open when you click the button. I'm guessing it's not possible to use the dialogservice in the example because it's not available, but would it be possible to add some note/warning at the top of the page that people should 'always' use the dialogservice to create a dialog, so that they don't face rendering issues? I'm hoping that would help people prevent this problem before they run into it 🙂

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dialog/dialog.tsx
+++ b/src/components/dialog/dialog.tsx
@@ -22,6 +22,13 @@ import { createRandomString } from '../../util/random-string';
  * See the example _Nested `close` events_.
  * :::
  *
+ * :::important
+ * Are you developing for
+ * [Lime CRM](https://www.lime-technologies.com/en/lime-crm/)? Please note that
+ * you should use the [DialogService](https://lundalogik.github.io/lime-web-components/versions/latest/#/api/dialog-service)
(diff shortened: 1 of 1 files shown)


adrianschmidt:
I'm just going to merge this now. Anyone who has an opinion has until the PR checks have run to pipe up.

PR #1632, opened 2022-02-21
fix(table): make 'selectAll' is disabled if the table is empty

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/2673

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -747,6 +747,7 @@ export class Table {
             >
                 <limel-checkbox
                     onChange={this.selectAllOnChange}
+                    disabled={!this.totalRows}
                     checked={this.tableSelection?.hasSelection}
                     indeterminate={
                         this.tableSelection?.hasSelection &&


src/components/table/table.tsx
```diff
@@ -734,6 +734,7 @@ export class Table {
             >
                 <limel-checkbox
                     onChange={this.selectAllOnChange}
+                    disabled={!this.totalRows ? true : false}
```
maltja, line 737: You can use double exclamation point here. That will convert a truthy or falsy value (here that is totalRows) to “true” or “false”. 
```suggestion
                    disabled={!!this.totalRows}
```

maltja (changes requested):

src/components/table/table.tsx
maltja: Thats true 🙂

LucyChyzhova: thank you! ❤Only, we need to have disabled in case, when there is no row, so ` disabled={!this.totalRows} `

maltja (approved):

src/components/table/table.tsx
adrianschmidt: > Only, we need to have disabled in case, when there is no row, so `disabled={!this.totalRows}`

Isn't that what I wrote in the suggested change? (I'm asking because I _think_ you missed my suggested change, but it's Monday and I'm tired, so it's completely possible that I'm staring at some difference that just isn't registering in my brain… 😅)

LucyChyzhova: oh👀😃 now I see your comment @adrianschmidt . I guess my thread wasn't updated/ or delayed, while I added my comment. At that time I saw only Malin comment, but it's great that we had the same solution 😉

Issue #1668, opened 2022-03-11
limel-file only recognizes lowercase file extensions


dorisenthecat:
## Current behavior

1. Go to https://lundalogik.github.io/lime-elements/versions/NEXT-35.0.0-next.31/#/component/limel-file/
2. In the basic example remove the attached file
3. Attach a word file with capital letters in the file extension (`.DOCX`)

![image](https://user-images.githubusercontent.com/1199682/157819410-bd8c7163-45a0-4477-8b16-9182e48bfd9e.png)

## Expected behavior

The word icon should also be shown for files with extension `.DOCX`.

## Environment

- **lime-elements** version: 35.0.0-next.31
- Framework used: 
- Logs:

adrianschmidt:
This seems like it should be exceedingly easy to fix. PRs very much welcome!

adrianschmidt:
In fact, something like this ought to solve it (not checked for possible errors if `extension` has a weird value):

```diff
- return filetypeIconTable[extension] || DEFAULT_ICON;
+ return filetypeIconTable[extension?.toLowerCase()] || DEFAULT_ICON;
```

https://github.com/Lundalogik/lime-elements/blob/cfb1ea6bcb27a65a048389fbe71aeadc57200990/src/components/file/icons.ts#L128

PR #1676, opened 2022-03-17
Uppercase filenames does not resolve to the correct file icon

anderssonjohan:
Issue: Files with names such as DOC1.DOCX aren't showing the word icon in the web client.

fix: #1668

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/file/file-metadata.spec.ts
+++ b/src/components/file/file-metadata.spec.ts
@@ -1,0 +1,80 @@
+import {
+    getFileIcon,
+    getFileColor,
+    getFileBackgroundColor,
+} from './file-metadata';
+
+describe('file metadata', () => {
+    [
+        {
+            description: 'with a known file extension',
+            input: {
+                id: '',
+                filename: 'file.docx',
+            },
+            expectedOutput: {
+                iconName: 'ms_word_copyrighted',
+                iconColor: 'rgb(var(--color-sky-dark))',
+                iconBackgroundColor: 'rgba(var(--color-sky-lighter), 0.4)',
+            },
+        },
+        {
+            description: 'with an unknown file extension',
+            input: {
+                id: '',
+                filename: 'file.justsomethingrandom',
+            },
+            expectedOutput: {
+                iconName: 'file',
+                iconColor: 'rgb(var(--color-gray-dark))',
+                iconBackgroundColor: 'rgba(var(--color-gray-lighter), 0.4)',
+            },
+        },
+        {
+            description: 'with no file extension',
+            input: {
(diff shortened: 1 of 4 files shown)


src/components/file/icons.ts
```diff
@@ -125,5 +125,5 @@ const filetypeIconTable: Record<string, string> = {
 };
 
 export function getIconForFile(extension: string): string {
-    return filetypeIconTable[extension] || DEFAULT_ICON;
+    return filetypeIconTable[extension.toLowerCase()] || DEFAULT_ICON;
```
adrianschmidt, line 128: I would suggest that we test this with a file that has no filename extension, just to make sure that doesn't result in an error.

src/components/file/file-metadata.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+            const iconName = getFileIcon({
+                id: '',
+                filename: 'README',
+            });
+            expect(iconName).toEqual('file');
+        });
+        it('is case insensitive', () => {
+            const iconA = getFileIcon({ id: '', filename: 'doc.docx' });
+            const iconB = getFileIcon({ id: '', filename: 'DOC.DOCX' });
+            expect(iconA).toEqual(iconB);
```
anderssonjohan, line 26: @adrianschmidt what do you think about this? something more that would be good to add while i'm at it?

src/components/file/icons.ts
adrianschmidt: Ah, I found it. I'll push a fix for the colours 👍

adrianschmidt (changes requested):

src/components/file/icons.ts
adrianschmidt: ⚡

adrianschmidt (dismissed):

src/components/file/icons.ts
adrianschmidt: @anderssonjohan If you're happy with my update, I'll squash and merge this 👍

anderssonjohan: I thought you were expecting me to squash so I've already done that. Just need your accept on the review again since it became stale when I pushed.

adrianschmidt (approved):

PR #1687, opened 2022-03-24
feat(button): add loadingFailed property

vicgeralds:
fix: Lundalogik/crm-feature#2733

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -83,12 +83,16 @@ button {
     }
 
     svg {
-        fill: currentColor;
         height: functions.pxToRem(30);
         opacity: 0;
         pointer-events: none;
         position: absolute;
         width: functions.pxToRem(30);
+
+        line {
+            stroke: rgb(var(--color-white));
+            stroke-width: 2;
+        }
     }
 
     limel-icon,
@@ -109,7 +113,8 @@ button {
         }
     }
 
-    &.just-loaded {
+    &.just-loaded,
+    &.just-failed {
         limel-icon,
         .label {
             opacity: 0;
@@ -119,12 +124,52 @@ button {
         }
         limel-spinner {
             display: block;
+            animation: fade-out-spinner 0.3s ease;
         }
+    }
+
+    &.just-failed {
+        background-color: var(--lime-error-text-color) !important;
+        animation: shake 0.82s cubic-bezier(0.36, 0.07, 0.19, 0.97) both;
     }
 }
 
 .mdc-button--outlined {
     @include button.outline-color(primary);
+}
+
+@keyframes fade-out-spinner {
+    0% {
+        opacity: 1;
+        transform: scale(1);
+    }
+    100% {
+        opacity: 0;
+        transform: scale(1.5);
+    }
+}
+
+@keyframes shake {
+    10%,
+    90% {
+        transform: translate3d(#{functions.pxToRem(-1)}, 0, 0);
+    }
+
+    20%,
+    80% {
+        transform: translate3d(#{functions.pxToRem(2)}, 0, 0);
+    }
+
+    30%,
+    50%,
+    70% {
+        transform: translate3d(#{functions.pxToRem(-3)}, 0, 0);
+    }
+
+    40%,
+    60% {
+        transform: translate3d(#{functions.pxToRem(3)}, 0, 0);
+    }
 }
 
 $speed-of-reducing-presence: 0.3s;
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -1,4 +1,4 @@
-import { Component, Element, h, Prop, Watch } from '@stencil/core';
+import { Component, h, Prop, State, Watch } from '@stencil/core';
 
 /**
  * @exampleComponent limel-example-button-basic
@@ -8,7 +8,8 @@ import { Component, Element, h, Prop, Watch } from '@stencil/core';
  * @exampleComponent limel-example-button-outlined
  * @exampleComponent limel-example-button-primary
  * @exampleComponent limel-example-button-composite
- * @exampleComponent limel-example-button-click
+ * @exampleComponent limel-example-button-click-success
+ * @exampleComponent limel-example-button-click-fail
  * @exampleComponent limel-example-button-reduce-presence
  * @exampleComponent limel-example-button-colors
  */
@@ -57,48 +58,66 @@ export class Button {
     @Prop({ reflect: true })
     public loading = false;
 
-    @Element()
-    private limelButton: HTMLLimelButtonElement;
+    /**
+     * Set to `true` to indicate failure instead of success when the button is
+     * no longer in the `loading` state.
+     */
+    @Prop({ reflect: true })
+    public loadingFailed = false;
 
-    private justLoadedTimeout: NodeJS.Timeout;
+    @State()
+    private justLoaded = false;
+
+    private justLoadedTimeout?: number;
 
     public render() {
         return (
             <button
-                class={`
-                    mdc-button
-                    ${this.primary ? 'mdc-button--unelevated' : ''}
-                    ${this.loading ? 'loading' : ''}
-                    ${this.outlined ? 'mdc-button--outlined' : ''}
-                `}
+                class={{
+                    'mdc-button': true,
+                    'mdc-button--unelevated': this.primary,
+                    loading: this.loading,
+                    'just-loaded': this.justLoaded && !this.loadingFailed,
+                    'just-failed': this.justLoaded && this.loadingFailed,
+                    'mdc-button--outlined': this.outlined,
+                }}
                 disabled={this.disabled}
             >
                 <div class="mdc-button__ripple" />
                 {this.renderIcon()}
                 <span class="label mdc-button__label">{this.label}</span>
                 <limel-spinner limeBranded={false} />
-                <svg viewBox="0 0 30 30">
-                    <path d="M20.659 10l-6.885 6.884-3.89-3.89-1.342 1.341 5.053 5.052.182.176L22 11.341z" />
-                </svg>
+                <svg viewBox="0 0 30 30">{this.renderLoadingIcons()}</svg>
             </button>
         );
     }
 
     @Watch('loading')
     protected loadingWatcher(newValue: boolean, oldValue: boolean) {
-        const JUST_LOADED = 'just-loaded';
-        const button = this.limelButton.shadowRoot.querySelector('button');
         if (oldValue && !newValue) {
-            button.classList.remove('loading');
-            button.classList.add(JUST_LOADED);
+            this.justLoaded = true;
             const TIMEOUT = 2000;
-            this.justLoadedTimeout = setTimeout(() => {
-                button.classList.remove(JUST_LOADED);
+            this.justLoadedTimeout = window.setTimeout(() => {
+                this.justLoaded = false;
             }, TIMEOUT);
         } else if (newValue) {
-            button.classList.remove(JUST_LOADED);
-            clearTimeout(this.justLoadedTimeout);
+            this.justLoaded = false;
+            window.clearTimeout(this.justLoadedTimeout);
         }
+    }
+
+    private renderLoadingIcons() {
+        if (this.loadingFailed) {
+            return [
+                <line x1="9" y1="9" x2="21" y2="21"></line>,
+                <line x1="21" y1="9" x2="9" y2="21"></line>,
+            ];
+        }
+
+        return [
+            <line x1="8" y1="14" x2="15" y2="20"></line>,
+            <line x1="23" y1="9" x2="14" y2="20"></line>,
+        ];
     }
 
     private renderIcon(): HTMLElement {
--- a/src/components/button/button.e2e.ts
+++ b/src/components/button/button.e2e.ts
@@ -192,54 +192,34 @@ describe('limel-button', () => {
                 limelButton = await page.find('limel-button');
                 innerButton = await page.find('limel-button>>>button');
             });
-            it('the state is "not loading"', () => {
-                expect(innerButton).not.toHaveClass('loading');
-                expect(innerButton).not.toHaveClass('just-loaded');
-            });
-            it('the property is falsy', async () => {
-                const propValue = await limelButton.getProperty('loading');
-                expect(propValue).toBeFalsy();
-            });
+            testNotLoading();
+            testPropertyIsFalsy();
 
             describe('when setting the property to `true`', () => {
                 beforeEach(async () => {
                     limelButton.setProperty('loading', true);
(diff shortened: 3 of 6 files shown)


src/components/button/button.scss
```diff
@@ -109,6 +109,20 @@ button {
...
+            opacity: 0;
+        }
+        svg {
+            opacity: 1;
+            color: red;
+        }
+        limel-spinner {
+            display: block;
+        }
+    }
```
vicgeralds, line 124: Todo for @Kiarokh and @LucyChyzhova
We also need to render another svg for the loading failed state.

src/components/button/button.tsx
```diff
@@ -57,18 +57,24 @@ export class Button {
...
     public loading = false;
 
-    @Element()
-    private limelButton: HTMLLimelButtonElement;
+    /**
+     * Set to `true` to indicate failure instead of success when the button is
+     * no longer in the `loading` state.
+     */
+    @Prop()
+    public loadingFailed = false;
```
adrianschmidt, line 65: Always use `reflect: true` for `@Prop`s that have a primitive value. [Why?](https://developers.google.com/web/fundamentals/web-components/best-practices#aim-to-keep-primitive-data-attributes-and-properties-in-sync,-reflecting-from-property-to-attribute,-and-vice-versa.)

```suggestion
    @Prop({ reflect: true })
    public loadingFailed = false;
```

adrianschmidt, line 65: Hmm… I'm wondering if perhaps this functionality is starting to get a little convoluted. We already kind of override the default behaviour when reduced presence is used, and now we need to add another property, and then that will have to be unset again by sending out an event and stuff…

Maybe, instead, we should just let the consumer set an "loadingEndedIcon" or something…? Of course, if we want the icon colour to also change for that, and don't want the consumer to have to worry about timings, we'd have to let them set a colour for the loadingEndedIcon too… 🤔

Urgh… I'm not sure. But I think this current implementation is leading us down a rabbit hole of complexity (started by me when I added the built-in end-of-loading animation, so I'm not blaming anyone but myself here), and we should take a step back and try to simplify instead.

src/components/button/button.tsx
```diff
@@ -57,18 +57,24 @@ export class Button {
...
+
+    private justLoadedTimeout?: number;
 
-    private justLoadedTimeout: NodeJS.Timeout;
+    private justLoaded = false;
```
vicgeralds, line 69: I forgot to add the `@State()` decorator here. When testing manually I saw the `just-loaded` class wan't removed, so it means I made some mistake when refactoring the tests

src/components/button/button.tsx
vicgeralds: I understand your concern. I think it's fine still that it's this complicated as long as the loading functionality is useful and easy to use.

One other way to communicate the loading state to the component could be via an async callback. Then you would only pass the callback and no `loading` prop (maybe even set `disabled` because it wouldn't really make sense to keep track of and await more than one callback).

Not sure it makes sense to return a promise from a click handler though, because you expect it to be just a regular click handler.

I kind of like the idea of passing a promise via the loading prop... :smile:

src/components/button/button.tsx
vicgeralds: fixed now! (and updated the tests to fail when it should)

src/components/button/button.tsx
vicgeralds: Maybe we want to have a separate LoadingButton? MUI has that https://mui.com/api/loading-button/

adrianschmidt: @jgroth Can you give me a quick sanity check on my thoughts here ☝️? 🙏

adrianschmidt: AFK sum up: @jgroth agreed that it sounds like a good way forward 👍

vicgeralds: OK, sounds good! I'll do that

Kiarokh: Sorry I'm not the right person to have opinions about this. But I think doing a new component (e.g. `limel-loading-button`) is a bit unnecessary, and here are my reasons:

1. Consumers shouldn't override our colors, icons, or animations in this case. We decide these things. This is our design language.
2. I am completely fine if we figure out a solution for applying time-outs and delays ourselves as well and not let them be override-able. This is also a pat of our design language. In other words, the guide here (and my new additions to it: d9e2527cfc5078ad0c2e4ae3dbe6cb6b7f1a5ca6) should not be something that the consumer should care about. These should be things which we have thought through and decided. We should just have a doc-example that shows how it works if loading fails or succeeds. 
3. Deprecation is hard and time consuming, and we also have to update our implementation everywhere. We ourselves don't even tend to update our own implementations when we deprecate something, for a very long while. (e.g. `iconColor`) and they remain deprecated for a very long time.
4. We have to introduce a breaking change at some point of time. Can't deprecate the `loading` prop and do this "promise based approach" thingy already here in this component? _(I don't know what it is, but I guess what you mean). So if the the deprecated `loading` prop is used, we don't do this dynamic thingy with all our timeouts and so on. But if it's not used, we do and all goes smoothly and the consumer doesn't have to care. (Maybe I don't know at all what you mean by it. Please explain if possible.)_
5. Compared to other components we have, this one is already very small. Making another one will require kind of the exact same amount of code and styles and tests anyway. Or how is that other one supposed to work? Import `limel-button` in itself and reuse it?

adrianschmidt: Detailed responses below, but I think I'll use today's hackday to experiment with some solutions, so tomorrow I might be able to show you instead 😊

---

> * Consumers shouldn't override our colors, icons, or animations in this case. We decide these things. This is our design language.

I agree, but this isn't the reason I want to separate the components (I'll get to that reason below).

> * I am completely fine if we figure out a solution for applying time-outs and delays ourselves as well and not let them be override-able. This is also a pat of our design language. In other words, the guide here (and my new additions to it: [d9e2527](https://github.com/Lundalogik/lime-elements/commit/d9e2527cfc5078ad0c2e4ae3dbe6cb6b7f1a5ca6)) should not be something that the consumer should care about. These should be things which we have thought through and decided. We should just have a doc-example that shows how it works if loading fails or succeeds.

Again, I agree, but not part of the reason I want to separate the components.

> * Deprecation is hard and time consuming, and we also have to update our implementation everywhere. We ourselves don't even tend to update our own implementations when we deprecate something, for a very long while. (e.g. `iconColor`) and they remain deprecated for a very long time.

Deprecation isn't hard, nor time consuming. Actually _removing_ the old functionality can be, but by marking that old functionality as deprecated and referring people to the new functionality, we at least do what we can to lead people away from using the old stuff, so removing it at some point in the future doesn't become even harder than it has to be.

> * We have to introduce a breaking change at some point of time. Can't deprecate the `loading` prop and do this "promise based approach" thingy already here in this component? _(I don't know what it is, but I guess what you mean). So if the the deprecated `loading` prop is used, we don't do this dynamic thingy with all our timeouts and so on. But if it's not used, we do and all goes smoothly and the consumer doesn't have to care. (Maybe I don't know at all what you mean by it. Please explain if possible.)_

We can't use the `loading` property to determine if we should use the new or the old behaviour, because we can't know if a value of `false` (or the non-presence of the attribute on the html element) means "use the new behaviour" or "use the old behaviour, but the state is not `loading` at the moment". There just is no way to tell. And we need to know at the time the user clicks the button, which is _before_ loading starts 😕

> * Compared to other components we have, this one is already very small.

Following this argument, it's ok to make any component worse, so long as there's some other component that's worse still…

> Making another one will require kind of the exact same amount of code and styles and tests anyway.

The amount of code and tests are more or less irrelevant. What's relevant is complexity, because higher complexity code requires higher cognitive load to work on, and understand, which in turn leads to longer development time, mistakes that create bugs, problems where otherwise simple features become very difficult to implement because they are conflicting with existing code, etc, etc. Basically, every single time you've sworn over Lime CRM being complicated or stupid in some way, that's because at some point, someone made a decision, consciously or unconsciously, that led us down a path of increased complexity.

To draw a parallell, the normal button, that doesn't have any loading behaviour, could be compared to a flashlight, while the loading-button could be compared to a laser pointer. Sure, you could definitely build a product that has both a flashlight and a laser pointer built in, but that product would not be simpler than the sum of the two separate products. Sure, you might only need one battery, and the combined product would probably use less plastic for the housing than the sum of the two separate products. But you'd need an interface for the user to select which functionality to use, with all the complications of communicating the difference. And the internals obviously would be much more complicated, because a lot of stuff would be duplicated. You'd have two light-sources, two switches (presumably), double the amount of wires, etc. Now if there's a problem with the product in the future, which one do you think would be easier to service, the one that's just a flashlight (or laser pointer), or the one that has both functions combined?

> Or how is that other one supposed to work? Import `limel-button` in itself and reuse it?

It wouldn't import and reuse `limel-button`, but it could potentially share parts. It could probably share all of its styling. It might even be the case that `limel-button` could just wrap the new `limel-loading-button`, and use that to replicate the old deprecated behaviour 🤔

adrianschmidt: I've made a suggestion for a new limel-loading-button in #1698. Please have a look!

src/components/button/examples/button-click.tsx
```diff
@@ -1,18 +1,25 @@
...
- * automatically displays a checkmark icon for 2 seconds. Note that our click
- * handler isn't actually involved in this.
+ * automatically displays a checkmark icon for 2 seconds, to visualize that the
+ * previously ongoing process just succeeded.
+ * Note that our click handler isn't actually involved in this.
+ *
+ * :::important
+ * The button should remain disabled as long as it is loading, to reflect that
+ * a process is ongoing and prevent users to retrigger the same process.
+ * As soon as the loading is completed, the button should be enabled again.
```
adrianschmidt, line 18: @Kiarokh I think you've misunderstood the reason for the timeout that's setting the button to enabled again. But that's a good thing, because if you've misunderstood it, I'm sure that many of our consultants have too. So it needs clarification.

Funnily enough, I actually added som clarification regarding this to the new examples I made in my suggestion for a `limel-loading-button`, before I had seen this change of yours 😄

image

But I had only added it as a comment in the example code, so someone only looking at the UI wouldn't see it. I'll add an explanation to the description of the example too.

Also, I've been thinking of perhaps adding an example with a small form, where the whole behaviour is a little more like it would be in reality. I think that might help.

Kiarokh: Awesome! 👏

src/components/button/button.tsx
Kiarokh: Thanks for the detailed answers and examples. As I said, I'm not the right person to have opinions about what's better to do. My worry was that we want a flash-light and a laser pointer. We don't want to add the laser pointer to the existing flash light, to keep it simple. Instead, we create a new flash light that is exactly the same as before, and add the laser pointer to that new one instead.

Whether the new component is like that or not is completely out of my expertise and I leave the decision to you guys :)

src/components/button/examples/button-click.tsx
vicgeralds: I really don't see why it should be possible to have a button that's both loading and clickable. What should happen if I click a loading button? Should it cancel the loading? 😄 

I think we can get rid of confusion (and those clarifications) by rendering
```jsx
<button disabled={loading || disabled} />
```

Kiarokh: you might be right @vicgeralds 🤔 I can't think about a scenario where a user needs to re-trigger or re-click the same button. I can only think that the button will re-trigger the same process if clicked. But would it? Should it?

src/components/button/button.tsx
Kiarokh: > But I still think functionality like that could live in its own component that wraps the dumber regular button.

Isn't it the reverse of this in Adrian's PR? It seems `limel-button` wraps `limel-loading-button` or?

src/components/button/button.tsx
```diff
@@ -57,18 +58,25 @@ export class Button {
...
+    private justLoadedTimeout?: number;
 
     public render() {
         return (
             <button
                 class={`
                     mdc-button
                     ${this.primary ? 'mdc-button--unelevated' : ''}
-                    ${this.loading ? 'loading' : ''}
+                    ${this.renderLoadingState()}
```
civing, line 79: This function name does not really reflect what it does. It's providing a class and not actually rendering anything

civing (commented):
Just a small comment, feel free to ignore it.

src/components/button/button.tsx
adrianschmidt: > > But I still think functionality like that could live in its own component that wraps the dumber regular button.
> 
> Isn't it the reverse of this in Adrian's PR? It seems `limel-button` wraps `limel-loading-button` or?

Yes, that's true. I was thinking to let limel-loading-button use limel-button too to begin with, but I couldn't see a good way to do that, since we want special animations and stuff for the loading. It turned out much simpler to use the button that supports loading animations and promises, and wrap it in a simpler interface that doesn't make use of the animation and promise. (Or won't, once we remove the deprecated `loading` property.)

jgroth: > I was thinking to let limel-loading-button use limel-button too to begin with, but I couldn't see a good way to do that, since we want special animations and stuff for the loading.

It might be that the button component is not flexible enough to handle that right now. That could probably be solved one way or another if we want to

adrianschmidt: Ah, wait… wasn't there some way to "tunnel" in styles for web components that we talked about a while back? Not just css variables, something more capable? 🤔

src/components/button/button.tsx
vicgeralds: What are your atoms of rendering? 😄 Rendering a class name (using the component's state as input) is what that method does.

But I guess you expect a render method to return a rendered element.

src/components/button/button.tsx
vicgeralds: I just found out about a `::slotted` pseudo selector...
https://developer.mozilla.org/en-US/docs/Web/CSS/::slotted

That only means the button can style what's in its slot, so I guess it's difficult to move all the styling to the loading button that wraps the button. Maybe that's something we can live with still.

src/components/button/examples/button-click.tsx
adrianschmidt: I'm breaking this whole discussion out by removing the commit updating the docs from this PR. Let's do that in a separate PR instead, to keep things as simple as we can.

adrianschmidt (dismissed):

adrianschmidt (approved):

PR #1688, opened 2022-03-25
Make the index page for docs versions get published from the latest `next`, also update some stuff on that page

adrianschmidt:
It's a pita to update the index page, since you have to check out the gh-pages branch to make changes to it. This PR adds the file to the normal branches. When publishing, we always grab the file from the latest published `next`-version.

See the respective commit messages for the small changes I made to the index file.

Update: I added functionality for removing outdated patch versions (we only keep the latest patch version for each minor version—so if we have 1.2.3 and 1.2.4, we remove the 1.2.3 version), and outdated NEXT versions (when we release v35.0.0, all `NEXT-35.0.0-next.*` are removed).

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/docs-index.html
+++ b/docs-index.html
@@ -35,8 +35,8 @@
                 return b[2].substring(0, 1) > a[2].substring(0, 1) ? 1 : -1;
             }
         });
-        createLink(`versions/latest/`, 'Latest');
-        createLink(`versions/next/`, 'Next');
+        createLink(`versions/latest/`, 'Latest Official Release');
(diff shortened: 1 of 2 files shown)


adrianschmidt:
I don't think anyone is particularly interested in reviewing this, and I've tested it locally and on the live gh-pages branch too.

PR #1703, opened 2022-03-30
remove indentation when having required error in textarea

Befkadu1:
fix Lundalogik/crm-feature#2658

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -341,7 +341,6 @@ export class InputField {
             'mdc-text-field--disabled': this.disabled || this.readonly,
             'lime-text-field--readonly': this.readonly,
             'mdc-text-field--required': this.required,
-            'mdc-text-field--with-trailing-icon': !!this.getTrailingIcon(),
             'lime-text-field--empty': !this.value,
         };
 
@@ -351,6 +350,8 @@ export class InputField {
                 !!this.helperText || !!this.maxlength;
         } else {
             classList['mdc-text-field--with-leading-icon'] = !!this.leadingIcon;
+            classList['mdc-text-field--with-trailing-icon'] =
+                !!this.getTrailingIcon();
         }
 
         return classList;


src/components/input-field/input-field.tsx
```diff
@@ -351,6 +350,8 @@ export class InputField {
                 !!this.helperText || !!this.maxlength;
         } else {
             classList['mdc-text-field--with-leading-icon'] = !!this.leadingIcon;
+            classList['mdc-text-field--with-trailing-icon'] =
```
Kiarokh, line 353: Right now you’re saying that the `textarea` should never show any icons, leading or trailing. What if the consumer sends an icon to it? Have you tested if it will show up?

I think the problem was that you were getting the class for leading icon `mdc-text-field--with-leading-icon`, when the `textarea` was invalid. While you should have probably gotten the class for trailing icon instead (which opens space for the ( ! ) icon on the right side. 

I think that’s what they (MD) wants to put in there probably, which is why you were getting that class, but with a slightly wrong name (`leading` instead of `trailing`).


image

I don't know why the `high_importance` icon is not showing up for invalid `textarea`s right now. I'll check.

Kiarokh (changes requested):

src/components/input-field/input-field.tsx
Kiarokh: OK, I see what happens if I remove this 👇 
image


the textarea is inside the `label` element, and fixing this requires a lot of change.
image

Not worth it. So I'd say your solution is good for now

Kiarokh (approved):

Kiarokh:
Sorry @civing I jumped in here. I think things look good and approved. But you can also have a second look.

src/components/input-field/input-field.tsx
adrianschmidt: Yeah, I don't think we've ever supported trailing icons for textareas, and I'm not even sure MDC really does… (I haven't looked it up though) 🤔

PR #1705, opened 2022-03-31
fix(input-field): avoid unnecessary code error messages at webclient

LucyChyzhova:
## Review:
fix: https://github.com/Lundalogik/crm-feature/issues/2657

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/form.types.ts
+++ b/src/components/form/form.types.ts
@@ -21,6 +21,11 @@ export interface FormError {
     name: string;
 
     /**
+     * Params of the error
+     */
+    params?: unknown;
+
+    /**
      * Name of the invalid property
      */
     property: string;
--- a/src/components/form/form.tsx
+++ b/src/components/form/form.tsx
@@ -276,6 +276,7 @@ export class Form {
         const errors: FormError[] = originalErrors.map((error: AjvError) => {
             return {
                 name: error.name,
+                params: error.params,
                 property: error.property,
                 message: error.message,
                 // For some reason 'schemaPath' is missing from the AjvError type definition:


maltja (dismissed):
looks good!

src/components/form/widgets/input-field.ts
```diff
@@ -50,9 +50,9 @@ export class InputField extends React.Component {
         if (event.detail || typeof event.detail === 'number') {
             value = event.detail;
         } else if (type === 'number') {
-            value = null;
+            value = undefined;
```
jgroth, line 53: It seems that setting it to `undefined` may have side effects 0259623

LucyChyzhova: Yes, we need to investigate more `null/undefined` issue. But when the `error.name === 'required' ` we could merge I guess. or?

jgroth: If we set it to `undefined` here, I think it will be impossible to remove a value (judging from the description in 0259623) 
It might also be that the issue is in the webclient and we need a better way to handle that issue there. Did you test what will happen on the card in the webclient with this set to `undefined` What happens when you remove a value and save the card? Will it be saved with the empty value or will the old value stick?

LucyChyzhova: yes, I have tested it in the webclient. It was possible to remove data but since these input fields are required it was not possible to save them as empty fields.

jgroth: Ok, then we need to test it with a field that is not required :slightly_smiling_face:

LucyChyzhova: yes, it's still possible to save the empty field (when it's not required)

LucyChyzhova: notes for the future: it was not possible to save an empty field when the `"type": "integer"`  (in case we changed null to undefined)

src/components/form/form.types.ts
```diff
@@ -20,6 +20,11 @@ export interface FormError {
      */
     name: string;
 
+    /**
+     * Type of the current input field
+     */
+    type?: string;
```
jgroth, line 26: I don't think this is correct. The `error.params` field can contain pretty much anything it seems and depends on what the value of `keyword` is according to the docs https://ajv.js.org/api.html#error-parameters

It probably make more sense to expose `params`, but without making any claims on what it is e.g. `Record<string, any>`

jgroth (dismissed):

PR #1710, opened 2022-04-01
fix(limel-form): suppress ajv $ref: keywords warning

### LINKED ISSUE #1709: Lime Admin: console log spamming of "$ref: keywords ignored at ..."
## Current behavior

Config pages that contains nested schemas generates excessive warning messages _including a stack trace_ in the browser console. This is especially bad with nested schemas that are used in conjunction with dynamic fields. Some add-on configurations even has several levels of nested schemas which can produce warnings in the number of thousands. This significantly slows down page rendering and can even cause the browser the crash.

The problem can be largely mitigated by closing the dev-console, so this is firstly a problem for developers and appers but could potentially affect customers and admins.

1. Open chrome dev-panel and switch to the console tab
2. In Lime Admin, open an Add-on (for instance gdpr, kpi, esign)

## Expected behavior

There should be no warnings logged.

## Environment

- **lime-elements** version: 34.1, 35.0
- Framework used: web client
- Logs: 
image

joelricci:
Fixes #1709

The main reason for these warnings (described in issue above) seems to be a discrepancy between the json-schema specification of the schema generated by the backend (lime-serializer), and the specification of the schema library (ajv) used on the frontend, notably by limel-form. 

Limel-form currently uses an older version of [ajv](https://github.com/Lundalogik/lime-elements/blob/633eba877525a73e2cfbfb79e79d1ebc5a92d675/package.json#L61) which only supports draft version 6 of JSON-Schema. This version of the specification treats nested schemas differently and [doesn't allow any extra keywords to be specified](https://json-schema.org/understanding-json-schema/structuring.html#ref) with the `$ref` keyword. However, lime-serializer ignore this fact and seems to adhere to a later draft of the specification which does indeed allow keywords together with $ref. The default for this version of ajv is to [log a warning](https://github.com/ajv-validator/ajv/tree/v6.7.0#referenced-schema-options) about this:

> when $ref is used other keywords are ignored (as per [JSON Reference](https://tools.ietf.org/html/draft-pbryan-zyp-json-ref-03#section-3) standard). A warning will be logged during the schema compilation.

One way of fixing this problem would obviously be to make the backend adhere to the older draft json-schema specification, but that would just be counterproductive looking forward.

Luckily there's an option to force ajv to validate _all_ keywords, including the 'extra keywords' (default behaviour in [newer versions of the draft](https://ajv.js.org/v6-to-v8-migration.html#removed-options)) instead of logging a warning. This would effectively quench the warning logs. However, ajv is used in two places in the limel-form component. It's used as a stand-alone validator but also integrated in the react-json-schema-form component which limel-form is essentially a wrapper to. 

At one point validation was moved out from react-json-schema-form to using the 'stand-alone' ajv library instead, but some validation is still being done by rsjf; about 1/3 of the warnings are produced by rjsf.

This PR does two things:

- completely disables validation by react-json-schema-form; this should be safe since [validation is already done outside rsjf](https://github.com/Lundalogik/lime-elements/pull/855).
- treat and validate $ref siblings ('keywords') as normal keywords [as is consistent with later specs](https://ajv.js.org/v6-to-v8-migration.html#removed-options).

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/form.tsx
+++ b/src/components/form/form.tsx
@@ -161,6 +161,7 @@ export class Form {
             React.createElement(
                 JSONSchemaForm,
                 {
+                    noValidate: true,
                     schema: this.modifiedSchema,
                     formData: this.value,
                     onChange: this.handleChange,
--- a/src/components/form/form.tsx
+++ b/src/components/form/form.tsx
@@ -233,6 +233,7 @@ export class Form {
             unknownFormats: 'ignore',
             allErrors: true,
             multipleOfPrecision: 2,
+            extendRefs: true,
         }).addFormat('integer', isInteger);
         this.validator = validator.compile(this.schema);
     }


adrianschmidt:
OMG, yes! 😍

image

src/components/form/form.tsx
```diff
@@ -161,6 +161,7 @@ export class Form {
             React.createElement(
```
adrianschmidt, line 161: This looks simple enough. What remains to be done?

I assume we'll want to test it thoroughly, to make sure we haven't lost any validation functionality we didn't want to lose.

joelricci: Yes, that's probably my greatest concern as well. Perhaps we should also assess whether we still need extra validation outside of rjsf (just making sure that requirement still stands today).

PR #1728, opened 2022-04-13
fix(dialog): make it possible to update closing actions

maltja:
fix Lundalogik/crm-feature#1122

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dialog/dialog.tsx
+++ b/src/components/dialog/dialog.tsx
@@ -1,4 +1,4 @@
-import { DialogHeading } from '@limetech/lime-elements';
+import { DialogHeading, ClosingActions } from '@limetech/lime-elements';
 import { MDCDialog } from '@material/dialog';
 import {
     Component,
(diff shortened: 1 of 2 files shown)


src/components/dialog/dialog.tsx
```diff
@@ -202,6 +195,15 @@ export class Dialog {
         }
     }
 
+    @Watch('closingActions')
+    protected closingActionsChanged(newValue: boolean, oldValue: boolean) {
+        if (isEqual(newValue, oldValue)) {
```
jgroth, line 200: I don't think we need to use `lodash` just to compare two `boolean`s :smile: 
```suggestion
        if (newValue === oldValue) {
```

adrianschmidt: I don't think the typings of `newValue` and `oldValue` are correct here:

Screenshot of the diff showing that the property closingActions is an object with two boolean properties, not a boolean itself

maltja: You are right, I will do a fixup! 🙂

jgroth (dismissed):

jgroth (approved):

Issue #1718, opened 2022-04-08
In Chip-set, the `<input>` `type` should be `search`


Kiarokh:
## Current behavior
Currently the `limel-chip-set `used also in `limel-picker` uses a `limel-input` with a type of `text`.
<img width="1301" alt="image" src="https://user-images.githubusercontent.com/35954987/162475023-f889f490-df87-495a-865b-5bf3709bac17.png">

This creates usability problems in certain cases, because there is a difference between `search` and `text` types for browsers. You can [read about some of these differences here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#using_search_inputs). But here are a couple of reasons why we should have used type `search` for the Chip-set at least.

## Accessibility 
Screenreaders do not know that users can search in the input box, just because we put a magnifying glass there. Thus they cannot announce that the form is a search form or this is a search box. [More…](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#search_form_labels_and_accessibility)

## Usability on smart devices
Smart devices ([at least on iOS](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#autocorrect)) know what language the user is writing in (due to the keyboard layout that's chosen). Normally, their auto-correction kicks in when user is typing in input fields. Because it could be a form they are filling in or a message that they're typing. This is a good thing. But not user is searching. When searching, the keywords could be really anything, and it should be up to the search engine (like Elastic, or Google) to guess suggest what may be a relevant keyword. This happens a lot when searching for Persons Names, Company names, since the auto-correction dictionary does not have all of those in the chosen language. 

Here is an example of how auto-correction switches user's input to something else, as soon as the <kbd>Enter</kbd>  key is pressed in a relation picker in Lime CRM. If user is not paying 100% attention, they will wonder why aren't they getting the results they expect. Since corrected words are very similar to the assumed typos, it'll be very hard to distinguish. 

https://user-images.githubusercontent.com/35954987/162478333-eb89053b-02db-49d1-ab06-398538178b76.MP4

## ⚠️ One thing to be careful about
Sometimes browsers put a ( X ) icon for clearing the typed input, or put a 🔍 icon by default to `<input type="search">`. When fixing this, we have to check that we won't get the same behavior as below:

https://user-images.githubusercontent.com/35954987/162479245-ae264f7a-a32a-42b3-b35c-669d76762a21.MP4

adrianschmidt:
[...]

PR #1736, opened 2022-04-19
feat(chip-set): make it possible to set `input` type of `search` or `text`

LucyChyzhova:
fix: #1718

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -29,6 +29,8 @@ const INPUT_FIELD_TABINDEX = 1;
  * @exampleComponent limel-example-chip-set-filter
  * @exampleComponent limel-example-chip-set-filter-badge
  * @exampleComponent limel-example-chip-set-input
+ * @exampleComponent limel-example-chip-set-input-type-text
+ * @exampleComponent limel-example-chip-set-input-type-search
  * @exampleComponent limel-example-chip-icon-color
  * @exampleComponent limel-example-chip-set-composite
  */
@@ -87,6 +89,13 @@ export class ChipSet {
      */
     @Prop({ reflect: true })
     public readonly: boolean = false;
+
+    /**
+     * For chip-sets of type `input`. Value to use for the `type` attribute on the
+     * input field inside the chip-set.
+     */
+    @Prop({ reflect: true })
+    public inputType: 'search' | 'text' = 'text';
 
     /**
      * For chip-sets of type `input`. Limits the maximum number of chips.
@@ -370,7 +379,7 @@ export class ChipSet {
                 {this.value.map(this.renderInputChip)}
                 <input
                     tabIndex={INPUT_FIELD_TABINDEX}
-                    type="text"
+                    type={this.inputType}
                     id="input-element"
                     disabled={this.readonly || this.disabled}
                     class={{
--- a/src/components/chip-set/examples/chip-set-input-type-search.tsx
+++ b/src/components/chip-set/examples/chip-set-input-type-search.tsx
@@ -1,0 +1,81 @@
+import { Chip } from '@limetech/lime-elements';
+import { Component, h, State } from '@stencil/core';
+import { ENTER, ENTER_KEY_CODE } from '../../../util/keycodes';
+
+/**
+ * Input chip set with `inputType` of `search`
+ *
+ * When autocorrection is potentially harmful for the user experience and for
+ * your intended result, use `search` as `inputType`. For instance, for a
+ * question like "Please suggest unique names for our newly founded company",
+ * you probably don't want autocorrection, because you would expect many
+ * valid suggestions to not exist in the autocorrection dictionary. Therefore,
+ * you do not want the respondent's input to be regarded as a typo and to be
+ * changed when they press <kbd>Enter</kbd> or <kbd>Space</kbd>.
+ */
+@Component({
+    tag: 'limel-example-chip-set-input-type-search',
+    shadow: true,
+})
+export class ChipSetInputExample {
+    @State()
+    private value: Chip[];
+
+    @State()
+    private textValue = '';
+
+    @State()
+    private maxItems = 3;
+
+    @State()
+    private emptyInputOnBlur: boolean = true;
+
+    constructor() {
+        this.value = [this.createChip('Lundalogik'), this.createChip('Lime')];
+    }
+
+    public render() {
+        return [
+            <limel-chip-set
+                type="input"
+                inputType="search"
+                label="Suggest three unique names for our newly founded company"
+                maxItems={this.maxItems}
+                value={this.value}
+                onChange={this.handleChange}
+                onInput={this.handleInput}
+                onKeyUp={this.onKeyUp}
+                emptyInputOnBlur={this.emptyInputOnBlur}
+            />,
+        ];
+    }
+
+    private handleInput = (event: CustomEvent<string>) => {
+        this.textValue = event.detail;
+    };
+
+    private onKeyUp = (event: KeyboardEvent) => {
+        if (
+            (event.key === ENTER || event.keyCode === ENTER_KEY_CODE) &&
+            this.textValue.trim()
+        ) {
+            this.value = [
+                ...this.value,
+                this.createChip(this.textValue.trim()),
+            ];
+            this.textValue = '';
+        }
+    };
+
+    private handleChange = (event: CustomEvent<Chip[]>) => {
+        this.value = event.detail;
+    };
+
+    private createChip = (name: string): Chip => {
+        return {
+            id: name,
+            text: name,
+            removable: true,
+        };
+    };
+}
--- a/src/components/chip-set/examples/chip-set-input-type-text.tsx
+++ b/src/components/chip-set/examples/chip-set-input-type-text.tsx
@@ -1,0 +1,106 @@
+import { Chip } from '@limetech/lime-elements';
+import { Component, h, State } from '@stencil/core';
+import { ENTER, ENTER_KEY_CODE } from '../../../util/keycodes';
+
+/**
+ * Input chip set with `inputType` of `text`
+ *
+ * There is a slight difference in the way browsers treat `input` field
+ * with `type="text"` and `type="search"`. You can read more about this
+ * difference in [Mozilla's documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#using_search_inputs),
+ * but the most important difference in this case is activation of the
+ * autocorrection feature on most smart devices.
+ *
+ * When a user makes a spelling mistake while typing in an input field with
+ * `type="text"`, the mistake will be corrected automatically, right after they
+ * press <kbd>Enter</kbd> or <kbd>Space</kbd>. Input fields with `type="search"`
+ * do not auto correct the user's input.
+ *
+ * If you want to use limel-chip-set in a form context, where autocorrection is
+ * a good thing, use `text` as `inputType`. It is important to know that the
+ * chip-set component creates a chip from the autocorrected value, after the
+ * user has pressed the <kbd>Enter</kbd> key and the auto correction has fixed
+ * existing typos! For example, for a question like "Please type five of your
+ * favorite fruits", you would want to avoid misspellings, to collect higher
+ * quality data.
+ */
+@Component({
+    tag: 'limel-example-chip-set-input-type-text',
+    shadow: true,
+})
+export class ChipSetInputExample {
+    @State()
+    private value: Chip[];
+
+    @State()
+    private companyValue: Chip[];
+
+    @State()
+    private textValue = '';
+
+    @State()
+    private maxItems = 5;
+
+    @State()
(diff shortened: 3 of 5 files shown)


src/components/chip-set/chip-set.tsx
```diff
@@ -370,7 +394,7 @@ export class ChipSet {
                 {this.value.map(this.renderInputChip)}
                 <input
                     tabIndex={INPUT_FIELD_TABINDEX}
-                    type="text"
+                    type={this.inputType}
```
Kiarokh, line 397: Image from iOS

## ⚠️ head up

We shouldn't merge this yet, until we know how to get rid of the extra 🔍 icon that iOS adds.

Kiarokh: https://stackoverflow.com/questions/23296072/html5-search-input-magnifying-glass-override/23296152

LucyChyzhova: The last changes fixed an extra X button and the background colour 
 Image from iOS (1)
Image from iOS

src/components/chip-set/chip-set.tsx
```diff
@@ -88,6 +88,30 @@ export class ChipSet {
     @Prop({ reflect: true })
     public readonly: boolean = false;
 
+    /**
```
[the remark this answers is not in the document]
LucyChyzhova: Sounds good! I will create a new example and put this long description there.

src/components/chip-set/chip-set.tsx
```diff
@@ -370,7 +394,7 @@ export class ChipSet {
                 {this.value.map(this.renderInputChip)}
                 <input
                     tabIndex={INPUT_FIELD_TABINDEX}
```
adrianschmidt, line 396: Sorry for being off topic to the PR, but are we still setting tab-index to 1 here? tab-index should only ever be set to `-1` or `0`. I pointed this out before, but can't remember if there was some reason we couldn't just set it to `0`? @Kiarokh

Kiarokh: I don't know really. But I made this issue now so we don't forget to test this: https://github.com/Lundalogik/lime-elements/issues/1741

src/components/chip-set/examples/chip-set-input-type-search.tsx
```diff
@@ -0,0 +1,81 @@
+import { Chip } from '@limetech/lime-elements';
```
adrianschmidt, line 1: I squashed the commit with these examples into the feature commit, as I don't think they need to be separate, since it's docs for the feature you added.

I also added a fixup where I reworded some of the descriptions a little.

I've only tested this quickly on an iPhone on Browserstack, but didn't get any autocorrection there regardless of input type… 🤔

I'd like to test a little bit more, but I need to log off for today now.

src/components/chip-set/chip-set.tsx
adrianschmidt: So, what's the status here @Kiarokh? OK to merge after review, or not?

src/components/chip-set/examples/chip-set-input-type-search.tsx
adrianschmidt: On Browserstack, I'm not getting autocorrection on either iOS or Android, regardless of whether I test the `text` or the `search` example.

I am getting the magnifying glass icon on the enter key on Android though. For the picker I guess that's good. For the chip-set example it seems quite misplaced though…

Using `search`:
A screenshot of the limel-example-chip-set-input-type-search component with Android's software keyboard open, showing a magnifying glass icon on the Enter key

Using `text`:
A screenshot of the limel-example-chip-set-input-type-text component with Android's software keyboard open, showing an arrow-pointing-right icon on the Enter key

I do think we should allow setting the type regardless, so we can use `search` for the picker, but I think we might also want to allow some more granular control, like that described [here](https://davidwalsh.name/disable-autocorrect). That would be something we'd first enable in limel-input though…

adrianschmidt: Ah, we don't use limel-input in the chips-set, right? So, no, we don't need to implement anything on limel-input before implementing it here 🙂

LucyChyzhova: > we don't use limel-input in the chips-set, right?

No, we don't use limel-input here.

src/components/chip-set/examples/chip-set-input-type-text.tsx
```diff
@@ -0,0 +1,106 @@
...
+ * press <kbd>Enter</kbd> or <kbd>Space</kbd>. Input fields with `type="search"`
+ * do not auto correct the user's input.
+ *
+ * If you want to use limel-chip-set in a form context, where autocorrection is
+ * a good thing, use `text` as `inputType`. It is important to know that the
+ * chip-set component creates a chip from the autocorrected value, after the
+ * user has pressed the <kbd>Enter</kbd> key and the auto correction has fixed
+ * existing typos! For example, for a question like "Please type five of your
+ * favorite fruits", you would want to avoid misspellings, to collect higher
+ * quality data.
```
adrianschmidt, line 25: Bias declaration: I don't use automatic autocorrection, because I prefer to just get a red underline for potential typos, rather than having my phone change stuff automatically, because I just can't stand when it "autocorrects" something away from what I actually meant to something else. Given that, I have a hard time imagining what a user who likes autocorrection might prefer here. But:

I assume that, if you have autocorrection enabled, if you actually meant what you wrote, you can go back and change it again (I guess you can, right? It won't then autocorrect _again_?). If so, what about when it's used in the chip-set? If you write something that gets incorrectly "autocorrected", remove the incorrect chip, and then write the same thing again, will your phone understand not to change it this time, or will it "autocorrect" it again, since you did "other stuff" in between?

If it would "autocorrect" again, having autocorrection enabled on a chip-set would make it very hard to write anything the autocorrection feature doesn't "allow"…

I guess you could write your word, press space to have it autocorrect, then backspace and correct what you wrote, and only _then_ press enter. But how many users are actually going to figure that out?

As I said, I'm clearly biased, and I'm not even going to be subjected to any of this, since I have autocorrection disabled, but I feel like having autocorrection on a chip-set is exactly the kind of thing that could make users hate our app… 😞

src/components/chip-set/examples/chip-set-input-type-search.tsx
Kiarokh: sorry for the delayed response. Yes the granular controls is a great idea. Actually both in here and in limel-input. :) But setting type to search sounds also like these granular controls. no?

> image

interesting effect here. I didn't know it works like that in Android.

> On Browserstack, I'm not getting autocorrection on either iOS or Android,

I think they have turned these features off on their keyboard settings. Probably they were problematic, given the service is used world-wide.

src/components/chip-set/chip-set.tsx
Kiarokh: I think it's alright.

src/components/chip-set/examples/chip-set-input-type-text.tsx
Kiarokh: > you can go back and change it again (I guess you can, right? It won't then autocorrect again?) […] 

It depends on the device, and timing of when you correct it, where you are typing, and some other factors that I don't really figure out. But it will in many cases. 

>  […] what about chipset […]

I think this also depends on device, timing, etc. On iOS when I try, the browser stops auto-correcting (for a while[?] while the `focus` is still in that field[?]) when I remove the wrong chip or word and try again. But the primary problem here is that user must realize first that what's in the "chip" is not what they wrote. For my dyslexic brain for instance, this may take a very long time sometimes. 

Note that chips don't show that red underline that indicates spelling errors.

> But how many users are actually going to figure that out?

If you realize that this autocorrection happens, you can find a way to hack it. But first you should consciously realize it.

src/components/chip-set/examples/chip-set-input-type-search.tsx
adrianschmidt: > sorry for the delayed response. Yes the granular controls is a great idea. Actually both in here and in limel-input. :) But setting type to search sounds also like these granular controls. no?

> I do think we should allow setting the type regardless, so we can use `search` for the picker

😄

> I think they have turned these features off on their keyboard settings. Probably they were problematic, given the service is used world-wide.

Makes sense I guess… it's just a pity if it can't be turned on and off so it can be tested 😞 But I haven't looked, maybe there _is_ a setting for it 🤷‍♂️

src/components/chip-set/chip-set.tsx
```diff
@@ -88,6 +90,13 @@ export class ChipSet {
     @Prop({ reflect: true })
     public readonly: boolean = false;
 
+    /**
+     * Defines how the web browser should treat the input field and what kind of
+     * interaction it should have.
```
adrianschmidt, line 95:
```suggestion
     * For chip-sets of type `input`. Value to use for the `type` attribute on the
     * input field inside the chip-set.
```

src/components/chip-set/examples/chip-set-input-type-text.tsx
adrianschmidt: I agree with your assessment above. It's unlikely the user will notice any autocorrection will happen, and it's even less likely they will figure out how to get around it, in case the autocorrect didn't do what they wanted. So I think we should just disable autocorrect for all limel-chip-sets, regardless of whether the input type is text or search…

I think we can do that either in this PR or separately, but if we want to do that, it seems quite unnecessary to now add a long explanation of why enabling autocorrection is good in some cases. That's the only reason I'm bringing this up here, rather than just merging 🙂

src/components/chip-set/examples/chip-set-input-type-search.tsx
Kiarokh: yes. When we tested on a real iOS device, the behavior changed as follows:

- The auto-correction would not automatically kick in before pressing space or enter. 
- the keyboard's predictive and spelling suggestions would still show word suggestions on keyboard's top bar. So user would notice that they could pick another word, or some spelling might be wrong. Like this image 👇 
image

I don't know how Android behaves though

src/components/chip-set/examples/chip-set-input-type-text.tsx
Kiarokh: yes makes sense. If we want to introduce `auto-correction= off` then we should not explain these stuff for input type. And at least I don't know a if we will have any more legit reasons to keep input type (we're overwriting the browser styles anyways). Or do you think we should keep both as props?

adrianschmidt: We could keep both, but if the main point is to disable autocorrection, it would be a smaller change to just add that in limel-chip-set, without changing the interface at all. Then we can always add the inputType option if we feel an actual need for it.

Kiarokh: sounds good to me

Kiarokh: what do you think @LucyChyzhova ?

Kiarokh: Oh actually when I read [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search#search_form_labels_and_accessibility) again, I change my mind. Autocorrection is one of the things we want to remove, when the user action is about searching. But that doesn't solve the accessibility problems.

So in this search field 
image
a screen reader wouldn't give enough information to users. Specially because our label ("Begins with") is also confusing. It becomes more clear for a sighted user, when we show the magnifying glass icon there. But not-sighted users won't see that. They would need more clues to realize they can actually perform a search.

adrianschmidt: The way you describe that autocorrection works for limel-chip-set, it _really_ doesn't sound to me like we would want autocorrection there, since it would be pretty much impossible for a user to write something that their phone thinks it ought to autocorrect, in case the user doesn't actually want it "corrected".

But I'll leave that up to you, since you are our interaction designer, after all 😊

Kiarokh (approved):
I think it's good to go with this solution, and in another PR add new features to `limel-input` (or other plain _non-MDC_ input fields which we might have in some components), to enable things like:

````html
<input 
	autocomplete="off" 
	autocorrect="off" 
	autocapitalize="off" 
	spellcheck="false" 
/>
````
Maybe we get these when we upgrade MDC, or maybe they're just there already for us to activate as new features.

And I'm not sure how big of a demand is out there for such features. I don't think at least us in the CRM would need them any time soon? Or maybe we do, but never have thought about them. Maybe it's good for instance to activate `autocorrect="off"` for the company name input field. 🤷‍♂️

PR #1774, opened 2022-06-08
feat(dialog): add CSS properties for max size

jgroth:
When a dialog is displayed fullscreen inside a popup window, the current backdrop creates an ugly
grey border around the dialog element. There should not be a need for a backdrop like that when a
dialog is displayed fullscreen in a popup window.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dialog/dialog.scss
+++ b/src/components/dialog/dialog.scss
@@ -13,6 +13,9 @@
  * @prop --dialog-heading-supporting-text-color: Color of the supporting text.
  * @prop --dialog-heading-icon-color: Color of the icon.
  * @prop --dialog-heading-icon-background-color: Background color of the icon when displayed as a badge.
+ * @prop --dialog-max-width: Max width of the dialog.
+ * @prop --dialog-max-height: Max height of the dialog.
+ * @prop --dialog-border-radius: Border radius of the dialog corners
  */
 
 :host {
@@ -45,6 +48,7 @@
     }
     .mdc-dialog__container {
         height: 100%;
+        width: var(--dialog-width, auto);
     }
 
     .mdc-dialog__surface {
@@ -52,6 +56,10 @@
         height: var(--dialog-height, auto);
         background-color: var(--dialog-background-color);
         box-shadow: var(--shadow-depth-64);
+
+        max-width: var(--dialog-max-width, calc(100vw - 2rem));
+        max-height: var(--dialog-max-height, calc(100% - 2rem));
+        border-radius: var(--dialog-border-radius, 0.25rem);
     }
 }
 


adrianschmidt:
Can you post a screenshot? I don't quite understand, and am very curious! 😊

jgroth:
@FredrikWallstrom can you help out with a screenshot? :smile:

jgroth:
Never mind, I cheated a bit and created something that should explain the use case. Dialogs in the desktop client will be displayed in a popup window, right now it will look something like this (ignore the content, it's broken :smile: )
Screenshot from 2022-06-08 16-16-58

With this fix, it will hopefully look something like this
Screenshot from 2022-06-08 16-17-34

FredrikWallstrom:
This is how it looks like in the desktop client:

Skärmavbild 2022-06-09 kl  08 32 25

adrianschmidt:
Ah, so, this is for opening a dialog in _new browser window_? I did notice you wrote "popup" but I somewhat ignored it because I just assumed that wasn't what you actually meant…

Why would we even want to do that? 🤨

But, ok, yeah, as long as it's only applied when opening the dialog in a new window (or new tab), so there is "nothing behind the dialog", this change makes sense 👍

src/components/dialog/dialog.tsx
```diff
@@ -145,6 +145,7 @@ export class Dialog {
                 class={{
                     'mdc-dialog': true,
                     'full-screen': !!this.fullscreen,
+                    popup: !!window.opener,
```
adrianschmidt, line 148: Hmm… [the MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/opener) for `window.opener` says:

> The [Window](https://developer.mozilla.org/en-US/docs/Web/API/Window) interface's `opener` property returns a reference to the window that opened the window, either with [open()](https://developer.mozilla.org/en-US/docs/Web/API/Window/open), or by navigating a link with a [target](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target) attribute.
>
> In other words, if window `A` opens window `B`, `B.opener` returns `A`.

I think that means that if anyone uses the "open in a new window" button (see screenshot), any full screen dialogs in the new tab will get this behaviour… 🤔

adrianschmidt: > > Why would we even want to do that?
> 
> It's for the desktop client. Rendering a dialog inside the action pad does not make much sense 😄

Ah! Of course! Yeah, then we really need this 👍

That also shifts my opinion on the value of this vs the downside that I mentioned in my comment on the code…

But, I don't think we should apply it so blindly here in lime-elements. I think this needs to be controlled by the consumer somehow… property, css-variable, css-class 🤷‍♂️

jgroth: Yeah, sure, we can maybe add a CSS property or something

jgroth: > I think that means that if anyone uses the "open in a new window" button (see screenshot), any full screen dialogs in the new tab will get this behaviour… 

I tried this and the `opener` was not set, but I agree that the solution I did not feel 100% fool proof

adrianschmidt: Huh, interesting… 🤔

Still, if possible within reason, it would be nice to have the consumer control it 👍

Kiarokh: > But, I don't think we should apply it so blindly here in lime-elements. I think this needs to be controlled by the consumer somehow… property, css-variable, css-class 🤷‍♂️

I agree. I think a css variable would be nice, safe, and wise

FredrikWallstrom:
Can we find out a smooth way to avoid this media query when we are in popup/desktop client mode? A little bit annoying that the buttons have the column direction in the desktop client even though there are plenty of space.

```
@media screen and (max-width: 760px) {
    slot[name='button'] {
        flex-direction: column-reverse;
    }
    .mdc-dialog__actions {
        padding: 1rem 1.5rem 1.5rem 1.5rem;
    }
}
```

Skärmavbild 2022-06-09 kl  12 21 00

src/components/dialog/dialog.scss
```diff
@@ -45,13 +45,18 @@
...
 
     .mdc-dialog__surface {
         width: var(--dialog-width, auto);
         height: var(--dialog-height, auto);
         background-color: var(--dialog-background-color);
         box-shadow: var(--shadow-depth-64);
+
+        max-width: var(--dialog-max-width, calc(100% - 32px));
+        max-height: var(--dialog-max-height, calc(100% - 32px));
+        border-radius: var(--dialog-border-radius, 4px);
```
Kiarokh, line 59: `px` 😱

Kiarokh, line 59: If you wanna show a dialog in 100% wide/tall (fully-fullscreen-mode), then we don't need a variable for the backdrop layer's background color. There won't be any backdrop layer anyways. 

But the fully-fullscreen-mode for a dialog doesn't make sense. A dialog **MUST** look like a dialog (a temporary layer on top of the current context). Therefore, it should be avoided, unless you want to open the dialog in a new browser tab (I guess what you refer to as popup here). In that condition (which I don't know much about and don't understand why you would need that) it's OK to show the dialog without any margins or border-radius. 

So I disagree with this change, unless you limit it somehow to popup windows.

jgroth: I copied it from the Material styles :upside_down_face:  You are more than welcome to improve it :smile: 

Is it possible to set it to something else as a default? I wanted it to use the value from Material if it has not been set but did not get it to work.

jgroth: You can see the before and after in my images above, we want to get rid of the grey area, margins, and rounded corners. In the desktop client, all dialogs will open in new browser windows, since it will be horrible to render a dialog in the action pad. There will not be any other context behind the dialog in the same window, just an empty body

Kiarokh: Yes I understood that. But we already have fullscreen dialogs in the web-client since before, right? 🤔 Or do I remember incorrectly? The `.fullscreen` isn't something new, or? https://github.com/Lundalogik/lime-elements/blob/f96edc5be2cec4d0c36f1c10b04cfe3a7b73e7b9/src/components/dialog/dialog.scss#L37

My worry is that we make all those so called fullscreen dialogs suddenly look like a fullscreen app.

jgroth: Yes, but even if the dialog is set to fullscreen it will add the grey border and margin around it, something that does not make sense when the dialog is displayed alone inside its own popup window. That's why these properties are introduced

jgroth: In my first image above, the dialog is set to fullscreen without these changes. The second one is with the changes and the HTML page has set the new CSS properties

jgroth: > My worry is that we make all those so called fullscreen dialogs suddenly look like a fullscreen app.

Nothing will change unless you manually set the new properties

FredrikWallstrom:
Something has happened with the dialog in the web client now with the new `max-*` properties. The `max-width: calc(....)` is wrapping the last word to a new line in all dialogs. 

Skärmavbild 2022-06-10 kl  09 20 22

Skärmavbild 2022-06-10 kl  09 22 47

Skärmavbild 2022-06-10 kl  09 23 07

src/components/dialog/dialog.scss
```diff
@@ -45,13 +48,18 @@
...
```
Kiarokh, line 62:
```suggestion
        max-width: var(--dialog-max-width, calc(100% - 2rem));
        max-height: var(--dialog-max-height, calc(100% - 2rem));
        border-radius: var(--dialog-border-radius, 0.25rem);
```

src/components/dialog/dialog.scss
Kiarokh: I just meant we should use `rem` not `px` 😄 added a suggestion 😉

Kiarokh:
> Can we find out a smooth way to avoid this media query when we are in popup/desktop client mode? A little bit annoying that the buttons have the column direction in the desktop client even though there are plenty of space.

@FredrikWallstrom I don't think we can do anything about it at the moment. The only thing possible now is that you set a size on the window in future. In hopefully-a-close-future, I guess we will have the possibility of using [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries) instead of media queries.

FredrikWallstrom:
>@FredrikWallstrom I don't think we can do anything about it at the moment. The only thing possible now is that you set a size on the window in future. In hopefully-a-close-future, I guess we will have the possibility of using [container queries](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Container_Queries) instead of media queries.

Ok! Lets leave it as it is then, shouldn't be too big of a deal.

src/components/dialog/dialog.scss
jgroth: > Aha.
So you mean in your example, the consumer has set all of these on 100%: `--dialog-width: 100%`, `--dialog-max-width: 100%`, `--dialog-max-height: 100%`.

> Sorry I didn't understand what had happened in the images and din't really connect the points. It'd be nice if you explained a bit more :sweat_smile: Using variables by consumers in this way is a nice solution. :ok_hand:

It sounds like you understand everything now, so I don't think I need to explain it any further :smile: This will just allow the consumer to set these properties if they want, to be able to display the dialog in a popup window

src/components/dialog/dialog.scss
jgroth: Fixed cbeb776
Also changed it to `100vw` instead of `100%` for the default max width since that was what Material was using. It might have been this that caused  the issues @FredrikWallstrom mentioned above

Kiarokh (approved):

Kiarokh:
@adrianschmidt I think all is good, so I approved. What do you think?

adrianschmidt (dismissed):
👌

adrianschmidt:
I don't have time to test this, but it looks good to me. Please just run it through the web client while you're squashing it, to make sure it's not completely bananas 😄

jgroth:
> I don't have time to test this, but it looks good to me. Since you'll be squashing it anyway, while you're at it, please just run it through the web client, to make sure it's not completely bananas smile

Great! Thanks! I've tested it in the web client and did not see any bananas :banana:

src/components/dialog/dialog.scss
Kiarokh: Oh nice! that sounds better. I was thinking in terms of Portal and thought that 100% would be 100% of the viewport anyways when dialogs are opened in the page. But that's probably another story when it comes to this popup scenario.

Kiarokh: but I wonder if height should also be `100vh` in that case? 🤔

jgroth: It was not `100vh` in the Material defaults and I did not want to screw anything up :smile:

Kiarokh: I think it depends on the html structure of the popup. I don't know how to test this. But I when I think about it, it feels 100% should be alright and actually better, due to mobile phones.
[They say](https://developer.chrome.com/blog/url-bar-resizing/):
>  `vh` units will be sized to the viewport height as if the URL bar is always hidden. That is, `vh` units will be sized to the "largest possible viewport". This means `100vh` will be larger than the visible height when the URL bar is shown.

So the proper way to set the height to fill the visible viewport is with height: 100%, either on the <html> element or on a position: fixed element. and `.mdc-dialog` has `position: fixed`.

Kiarokh (approved):

PR #1779, opened 2022-06-29
fix(progress-flow): make it possible to set an icon color

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/2841

Screenshot 2022-06-29 at 11 58 27
Screenshot 2022-06-29 at 11 59 38

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/progress-flow/examples/progress-flow-colors-css.scss
+++ b/src/components/progress-flow/examples/progress-flow-colors-css.scss
@@ -14,5 +14,5 @@
     --progress-flow-step-text-color: rgb(var(--contrast-200));
     --progress-flow-step-text-color-passed: rgb(var(--contrast-1100));
 
-    --progress-flow-icon-color: rgb(var(--color-lime-default));
+    --progress-flow-icon-color--inactive: rgb(var(--color-lime-default));
 }
--- a/src/components/progress-flow/examples/progress-flow-colors-css.scss
+++ b/src/components/progress-flow/examples/progress-flow-colors-css.scss
@@ -9,7 +9,7 @@
 
     --progress-flow-step-background-color: rgb(var(--contrast-1200));
     --progress-flow-step-background-color--selected: rgb(var(--color-sky-dark));
-    --progress-flow-passed-step-background-color: rgb(var(--contrast-700));
+    --progress-flow-step-background-color--passed: rgb(var(--contrast-700));
 
     --progress-flow-step-text-color: rgb(var(--contrast-200));
     --progress-flow-step-text-color--passed: rgb(var(--contrast-1100));
--- a/src/components/progress-flow/examples/progress-flow-colors-css.scss
+++ b/src/components/progress-flow/examples/progress-flow-colors-css.scss
@@ -12,7 +12,7 @@
     --progress-flow-passed-step-background-color: rgb(var(--contrast-700));
 
     --progress-flow-step-text-color: rgb(var(--contrast-200));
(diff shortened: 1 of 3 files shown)


Kiarokh (changes requested):
Good catch @LucyChyzhova ! 
I guess the history behind this is that just before putting this component for review, we decided to change the variable name from `--progress-flow-icon-color` to `--progress-flow-icon-color--inactive`, to keep it in sync with the rest of variable colors, which also reflect the state `--selected` or `--passed`, etc…

Then we forgot to update the variable name in 3 places: 
1. in the documentation (declaration of those variables) here https://github.com/Lundalogik/lime-elements/blob/5283bd7741dd7bd8938815e6bb708384a87faecb/src/components/progress-flow/progress-flow.scss#L12 
2. in the CSS code here: https://github.com/Lundalogik/lime-elements/blob/5283bd7741dd7bd8938815e6bb708384a87faecb/src/components/progress-flow/progress-flow-item/partial-styles/_colors.scss#L6
3. as well as the variable name in the example here: https://github.com/Lundalogik/lime-elements/blob/5283bd7741dd7bd8938815e6bb708384a87faecb/src/components/progress-flow/examples/progress-flow-colors-css.scss#L17 . 

Normally, this would be a breaking change if we fix those two problems now. But because: 
a. this variable name has never worked from the beginning when this component was added
b. nobody has complained about it or reported it as a bug, 
 
I would argue that it is safe to assume that we can instead fix those 3 problems (documentation, CSS code, & the CSS-coloring example).

What do you think @adrianschmidt ?

Kiarokh:
I also noticed that we have another typo in the documentations, 
which should separately be fixed from `-passed` to `--passed`
1. here in the documentations: https://github.com/Lundalogik/lime-elements/blob/5283bd7741dd7bd8938815e6bb708384a87faecb/src/components/progress-flow/progress-flow.scss#L10 
2. and also here example: https://github.com/Lundalogik/lime-elements/blob/5283bd7741dd7bd8938815e6bb708384a87faecb/src/components/progress-flow/examples/progress-flow-colors-css.scss#L15

@LucyChyzhova would you please add a new `fix` commit for this as well?

LucyChyzhova:
> @LucyChyzhova would you please add a new `fix` commit for this as well?

Sure! will be added

adrianschmidt:
> I also noticed that we have another typo in the documentations, which should separately be fixed from `-passed` to `--passed`
> 
> 1. here in the documentations: https://github.com/Lundalogik/lime-elements/blob/5283bd7741dd7bd8938815e6bb708384a87faecb/src/components/progress-flow/progress-flow.scss#L10
> 2. and also here example: https://github.com/Lundalogik/lime-elements/blob/5283bd7741dd7bd8938815e6bb708384a87faecb/src/components/progress-flow/examples/progress-flow-colors-css.scss#L15
> 
> @LucyChyzhova would you please add a new `fix` commit for this as well?

If the property is correctly named in the code, and this is just a matter of changing the docs, the commit type should be `docs`, not `fix` 😊

Kiarokh (dismissed):
great job Lucy! 👏

Kiarokh (approved):

PR #1781, opened 2022-07-05
fix(common): avoid crashes when no name/title and required prop given

Befkadu1:
When an object has a title or a name, the function returns the title or a name. If no title or
name given, it will return the required property value as a title. In our case the property is
neither a title/name nor a required property which caused the crash. This PR solves the issue

fix Lundalogik/crm-feature#2857

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/templates/common.spec.ts
+++ b/src/components/form/templates/common.spec.ts
@@ -49,6 +49,7 @@ const schema = {
                 },
             ],
         },
+        props: { type: 'object' },
     },
 };
 
@@ -182,6 +183,10 @@ describe('findTitle()', () => {
                 schema,
             ],
             output: 'test',
+        },
+        {
+            input: [{ props: { limetype: 'deal' } }, schema, schema],
+            output: null,
         },
     ].forEach(({ input, output }) => {
         const inputJson = JSON.stringify(input[1]);
--- a/src/components/form/templates/common.ts
+++ b/src/components/form/templates/common.ts
@@ -60,6 +60,10 @@ export function findTitle(data: any, fieldSchema: any, formSchema: any) {
         return null;
     }
 
+    if (!subSchema.properties) {
+        return null;
+    }
+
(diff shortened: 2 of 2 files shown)


src/components/form/templates/common.spec.ts
```diff
@@ -183,9 +184,13 @@ describe('findTitle()', () => {
...
             output: 'test',
         },
+        {
+            input: [{ props: { limetype: 'deal' } }, schema, schema],
+            output: null,
+        },
     ].forEach(({ input, output }) => {
         const inputJson = JSON.stringify(input[1]);
-        it(`returns ${output} when called with ${inputJson}`, () => {
+        it.only(`returns ${output} when called with ${inputJson}`, () => {
```
adrianschmidt, line 193:
```suggestion
        it(`returns ${output} when called with ${inputJson}`, () => {
```

Befkadu1: omg

Befkadu1: ⚡

src/components/form/templates/common.ts
```diff
@@ -56,10 +56,15 @@ export function findTitle(data: any, fieldSchema: any, formSchema: any) {
     data = sortDataByProperties(data, subSchema.properties);
 
     const firstEntry = findFirstEntry(data, subSchema);
+
```
adrianschmidt: I removed this. I think it's unnecessary, but if you really want it, the change belongs in a separate commit 🙂

adrianschmidt (approved):

PR #1782, opened 2022-07-06
 feat(portal): add more open direction possibilities for elements that…

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/1779

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/menu/menu.types.ts
```diff
@@ -1,4 +1,14 @@
...
+    | 'left'
+    | 'right'
+    | 'bottom-center'
+    | 'top-center'
+    | 'left-bottom'
+    | 'left-center'
+    | 'left-top'
+    | 'right-bottom'
+    | 'right-center'
+    | 'right-top';
```
Kiarokh, line 11: Great job. A few notes on the namings:

### 1. Naming 
I have some problems with these naming conventions. Maybe that's only me... what do you think?
I think what the popperjs does is quite OK. They simply use `start´ and `end` prefixes. 

Example:
| consistent naming       | visual naming     |  |
|--------------|-----------|------------|
| left-start | let-top      |        |
| left      | left-center  |       |
| left-end      | left-bottom |       |
| top-start | top-left      |        |
| top      | top-center  |       |
| top-end      | top-left  |       |

an so on…

### 2. Missing locations
`top` and `bottom` do not have `left` and `right` locations.

### 3. Duplicates
`left` and `left-center` are the same 
`right` and `right-center` too :)

adrianschmidt: > ### 1. Naming
> I have some problems with these naming conventions. Maybe that's only me... what do you think? I think what the popperjs does is quite OK. They simply use `start` and `end` prefixes. Not a big deal if we don't change them.
> 
> Example:
> 
> consistent naming	visual naming	
> left-start	let-top	
> left	left-center	
> left-end	left-bottom	
> top-start	top-left	
> top	top-center	
> top-end	top-left	
> an so on…

I don't quite follow. Could you be more explicit with what it is you think should be changed, and to what? I don't understand what "They simply use `start` and `end` prefixes" means 🤔

LucyChyzhova: > They simply use `start` and `end` prefixes.

Thanks, I see your point. Maybe you are right, I just want to explain why I named it with the left and right instead. We are all different and someone `starts` from the left - so `it's LEFT` others from the `right` and `it's their start point` like reading a sentence. I would say it could be even okay with the horizontal direction - as most of the users are in Europe and probably they get used to `starting as a left` and `end as the right`. BUT vertical is a bit confusing since I `wouldn’t say the top is a start` usually the `bottom is when we start`. but again it's the opposite for the reading. 🤔  I hope I didn't confuse you more 😁

LucyChyzhova: > ### 2. Missing locations
> `top` and `bottom` do not have `left` and `right` locations.

We use it for the `right` and `left`
   ```
        let placement: Placement = 'bottom-start';   // default right position
        let flipPlacement: Placement = 'top-start';

        if (this.openDirection === 'left') {.  // left position
            placement = 'bottom-end';
            flipPlacement = 'top-end';
        }
```

Comment (LucyChyzhova) [reply src/components/menu/menu.types.ts:11]: > ### 3. Duplicates
> `left` and `left-center` are the same `right` and `right-center` too :)

You are right, but, Unfortunately again because of the previous naming we will lose left center placement and right center placement since they are taken for the : 
     ` let placement: Placement = 'bottom-start';
        let flipPlacement: Placement = 'top-start';

        if (this.openDirection === 'left') {
            placement = 'bottom-end';
            flipPlacement = 'top-end';
        }`

But we wanted to have a center position as an option SO to distinguish them a bit I named `left-center` and `right-center`

Comment (adrianschmidt) [reply src/components/menu/menu.types.ts:11]: I think it's a good idea to stick with what popper is using, if it's not too late already.

I think that `left` should be used if the it stays left regardless of whether the chosen locale is `ltr` or `rtl`. (And conversely for `right` of course.) But if the menu is on the left side when Swedish is selected, and moves to the right side when Arabic is selected (if we supported Arabic), then `left` is a very bad name to describe that behaviour. Then `start` is much better (and a widely used convention nowadays).

> I wouldn’t say the top is a start usually the bottom is when we start

😮 🤨 Maybe we're thinking of different examples, but I definitely associate the _top_ to be the starting point, not the _bottom_. Unless building a house, sure. But for any examples I can think of that feels relevant to this context. What examples are you thinking of? 🙂

Comment (LucyChyzhova) [reply src/components/menu/menu.types.ts:11]: > I don't quite understand what's going on here. But that sounds like a mistake in the code, and something we should fix regardless. Because `bottom-end` and `top-end` are definitely located on the right side of the element in popper, not on the left side. 🤔 Or?

left has 
    `      if (this.openDirection === 'left') {
            placement = 'bottom-end';
            flipPlacement = 'top-end';
        }`

and it looks: Screenshot 2022-07-06 at 15 47 26

So it's up to us keep or change - but it's how it was before

Comment (Kiarokh) [reply src/components/menu/menu.types.ts:11]: 🤔 Not sure why that decision was made, and why they call it `left`. Maybe they thought this would mostly be used for menus, which are typically located on the top-right side of the screen? 

I have no idea. 🤷‍♂️ But I feel we have to fix this, and default the location either to something else. Or not call that default thing `left`, and instead correctly call it `bottom-end`. 

I'm super confused, and don't know if I'm thinking straight now. 

Changing the default opening directions may become a breaking change, but not a super horrible one. Because the portal will still do its best to place things in the viewport as I've understood it. However, that would mean we need to adjust the open direction in a few of our components, like the `limec-property-filter`. 

Does anyone have a good suggestion? @adrianschmidt @jgroth ?

Comment (adrianschmidt) [reply src/components/menu/menu.types.ts:11]: @eketorp Not because we need your input, but because you should read the thread from the start up to this point. It might give you some answers as to why so often want to take some time to consider the bigger picture even if the task at hand is just to create a minimum viable version of something 😄 (I'm _not_ saying this was brought about by a Product Owner, not at all. But it's a great example of how even trivial things can cause significant headache further down the road.)

File: src/components/portal/portal.tsx [hunk 2]
```diff
@@ -276,6 +276,46 @@ export class Portal {
...
+        if (this.openDirection === 'bottom-center') {
+            placement = 'bottom';
+            flipPlacement = 'top';
+        }
+
+        if (this.openDirection === 'top-center') {
+            placement = 'top';
+            flipPlacement = 'bottom';
+        }
+
```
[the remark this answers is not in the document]
Befkadu1: I can help you with it tomorrow

src/components/menu/menu.types.ts
LucyChyzhova: I just updated the naming in the correct way.🤓🙈😁 Let's talk about it tomorrow. (I will commit after it)

LucyChyzhova: > The default placement has been called `left`

No, 😀 the default one is the 'right' direction
Screenshot 2022-07-06 at 18 54 11

BUT you are right we can say **`bottom-start`** is a default one and we will get the same UI:
Screenshot 2022-07-06 at 18 55 28

LucyChyzhova: The problem will be only if the `left` position was specified. in this case they will get:
Screenshot 2022-07-07 at 07 48 43
 
But expected behaviour is: 
Screenshot 2022-07-07 at 10 36 52

adrianschmidt: > The problem will be only if the `left` position was specified.

Or if someone explicitly specified `right`. So we should mark the change as breaking, but with Kia's excellent research (❤️), we can be confident this change will not cause us any significant problem 👍

adrianschmidt (changes requested):
```
BREAKING CHANGE
We have done renaming of the `openDirection`s, which can affect
how things on rendered on the user interface now.
So, if you had previously specified an `openDirection` of an element
(for instance a menu) to be `right`, you can expect its portal
to be opened in `bottom-start` location, in relation to its
trigger element. If you had specified the `openDirection` to be `left`,
the portal would previously open in the `bottom-end` location,
but now it will open on `left` side of the trigger and aligned to the
center point.
```

This does not seem to trigger a breaking change. I think there has to be a colon after `BREAKING CHANGE`. I'm trying it out on my machine.

src/components/portal/portal.tsx
```diff
@@ -45,7 +45,7 @@ export class Portal {
      * Decides which direction the portal content should open.
      */
     @Prop()
-    public openDirection: OpenDirection = 'right';
+    public openDirection: OpenDirection = 'bottom';
```
adrianschmidt, line 48: Shouldn't this be either `bottom-end` or `bottom-start`?

src/components/menu/menu.tsx
```diff
@@ -20,6 +20,8 @@ import {
  * @exampleComponent limel-example-menu-basic
  * @exampleComponent limel-example-menu-disabled
  * @exampleComponent limel-example-menu-open-left
+ * @exampleComponent limel-example-menu-open-bottom
+ * @exampleComponent limel-example-menu-open-right-start
```
Kiarokh, line 24: I wonder if we need new examples? I think it's enough with one example about open direction, and we also have the composite example. So I'd suggest removing this, and instead writing some docs in the original example about possibility of opening the menu in 12 different locations (if you think we have to).

src/components/portal/portal.tsx
Kiarokh: we decided to default the location to bottom and align to center. This is also what popper.js does as default. Things look better this way if you as us :)

src/components/menu/examples/menu-open-bottom.tsx
```diff
@@ -0,0 +1,26 @@
+import { MenuItem, ListSeparator } from '@limetech/lime-elements';
+import { Component, h } from '@stencil/core';
+
```
Kiarokh, line 3: @adrianschmidt do you like the commit title, message, and breaking change?

src/components/menu/examples/menu-open-bottom.tsx
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Opening to the bottom
```
adrianschmidt, line 5: Rather than adding several different examples, I think it makes more sense to have a single example where the placement can be updated, similar to the example on https://popper.js.org/. We don't need to have the fancy dots though, I think it's good enough to just have the supported values in a select.

The same can of course also be seen in the composite example, but it's a little more hidden among all the other options there, so I think it's probably useful to have a separate example for just the positioning.

jgroth:
Disclaimer: I have not read all the comments in this issue :smile: 
Do we have to make this a breaking change? Can we deprecate the old values and make the component backward compatible before removing them?

src/components/menu/examples/menu-open-bottom.tsx
adrianschmidt: Yeah, I think I could whip something up quite quickly 👍

src/components/portal/portal.tsx
adrianschmidt: Yeah, ok, I guess `bottom` is common to use on mobile devices. (Which is a little weird, since it means you will cover the newly opened menu with your hand until you move your finger away 🤦‍♂️)

Funny how I was thinking "No one uses `bottom` for menus!" just because I happen to be seated in front of a laptop right now 😂

src/components/tooltip/tooltip.tsx
```diff
@@ -114,6 +114,7 @@ export class Tooltip {
         return (
             <div class="trigger-anchor">
                 <limel-portal
+                    openDirection="bottom-start"
```
Kiarokh, line 117: First we wanted to make this an editable `prop` and introduce a new feature, and default it to `top`. But due to [this bug](https://github.com/Lundalogik/lime-elements/issues/1408) we cannot this this yet. So we will have to do that later.

src/components/menu/examples/menu-open-bottom.tsx
adrianschmidt: I've updated the breaking change message, but I think I'll have some more changes to make still.

src/components/tooltip/tooltip.tsx
adrianschmidt: Oh, _there_ is the issue I was talking about in our call @LucyChyzhova ☝️ 😄

So, not directly related to limel-portal, but to the specific case where we use it in limel-tooltip. That's why I couldn't quite remember what the problem was, or find the issue 😄

src/components/menu/examples/menu-open-bottom.tsx
adrianschmidt: I guess, tell me when you're done with changing the code, and I can have a look at the commits @LucyChyzhova @Kiarokh

Kiarokh: That's super strange, because we pulled your changes before force pushing :/ No idea what happened. Sorry for that. We're done messing around now. You can review/update again.

adrianschmidt: No worries. I guess it's because it was just a change in the commit message, not in the code, and git really only tries to ensure that changes to the _code_ aren't lost… 🙂

adrianschmidt: And, great work! 👏

LucyChyzhova: > But it's a little worrying that someone seems to force push things without noticing what gets lost… 😐

yeah, we were trying to always pull first 🤞.. but it could be my fault 🙈

>I guess, tell me when you're done with changing the code, and I can have a look at the commits @LucyChyzhova @Kiarokh

no more changes 😁 in the lime-elements repo

src/components/portal/portal.tsx
adrianschmidt: Eh, sorry, I was confused. No, popover isn't private, but I meant _portal_ (those two are so damn confusing… 😖).

Since `limel-portal` is private, it doesn't make sense to document breaking changes on it. The changes need to be documented on the components that people can use 😄

I'm working on it now, and I'll fix some nice example while I'm at it, so should be ready for this to be merged fairly soon. Hopefully today…

src/components/color-picker/color-picker.tsx
```diff
@@ -102,7 +102,11 @@ export class ColorPicker implements FormComponent {
         }
 
         return (
-            <limel-popover open={this.isOpen} onClose={this.onPopoverClose}>
+            <limel-popover
+                open={this.isOpen}
+                openDirection="bottom-start"
```
adrianschmidt, line 107: Since this can't be set by the consumer, and there is no change to the what the user sees, it's "part of" the main change to limel-portal. So I've squashed this commit into that one.

src/components/list/list-renderer.tsx
```diff
@@ -247,7 +247,7 @@ export class ListRenderer {
             <limel-menu
                 class="mdc-deprecated-list-item__meta"
                 items={actions}
-                openDirection="left"
+                openDirection="bottom-end"
```
adrianschmidt: Same thing here.

> Since this can't be set by the consumer, and there is no change to the what the user sees, it's "part of" the main change to limel-portal. So I've squashed this commit into that one.

src/components/menu/examples/menu-open-bottom.tsx
adrianschmidt: ⚡

src/components/list/list-renderer.tsx
adrianschmidt: Oh, well, not quite the same here actually, because I added this so that we "keep the same default" here, but I've also kept the separate commit, which changes the default to a new position 😊

adrianschmidt (changes requested):
Hmm… the flip positions do not work in the new example I made 🤔

adrianschmidt (approved):
Ah! Never mind, yes they do 😅

adrianschmidt:
@Kiarokh @LucyChyzhova I'm happy with this. Please have a look, and if you're happy with it, I can squash and merge it 👍

src/components/popover/popover.tsx
```diff
@@ -11,6 +11,7 @@ import { createRandomString } from '../../util/random-string';
 import { zipObject } from 'lodash-es';
 import { portalContains } from '../portal/contains';
 import { ESCAPE } from '../../util/keycodes';
+import { OpenDirection } from '../menu/menu.types';
```
Kiarokh, line 14: should this be imported from `menu.types';`? or from portal?

src/components/portal/portal.tsx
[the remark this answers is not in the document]
Kiarokh: sounds correct! :)

LucyChyzhova:
> @Kiarokh @LucyChyzhova I'm happy with this. Please have a look, and if you're happy with it, I can squash and merge it 👍
Looks like we missed  the description above the example  
Screenshot 2022-07-08 at 16 01 05

src/components/popover/popover.tsx
adrianschmidt: The types are currently specified in `src/components/menu/menu.types.ts`. That does feel a little bit weird, but it doesn't matter for an external consumer, as they'll just import it with `from '@limetech/lime-elements';` anyway. So we can change it later if we want to.

src/components/menu/examples/menu-open-direction.tsx
```diff
@@ -0,0 +1,77 @@
...
+    MenuItem,
+    ListSeparator,
+    OpenDirection,
+    Option,
+} from '@limetech/lime-elements';
+import { Component, h, State } from '@stencil/core';
+
+/**
+ *
+ */
```
LucyChyzhova, line 11: Something like "Here you can try different positions of the menu"

adrianschmidt:
> Looks like we missed the description above the example

Ah! Good catch! I didn't know what to write at first, and then apparently totally forgot about it 😂

src/components/menu/examples/menu-open-direction.tsx
```diff
@@ -0,0 +1,82 @@
...
+        return [
+            <p>
+                <limel-menu
+                    items={this.items}
+                    openDirection={this.selectedOpenDirection?.value}
+                >
+                    <limel-button label="Menu" slot="trigger" />
+                </limel-menu>
+            </p>,
+            <limel-select
```
Kiarokh, line 68: I'll just make the UI a bit nicer here, and push soon 
image

Kiarokh: ⚡️ Done
image

src/components/menu/examples/menu-open-direction.tsx
adrianschmidt: (I copied your first comment about this into your comment above, and hid the original. Just so that it's all in this thread that we can resolve later 🙂)

src/components/menu/examples/menu-open-direction.tsx
```diff
@@ -9,8 +9,8 @@ import { Component, h, State } from '@stencil/core';
 /**
  * openDirection
  *
- * The value of the `openDirection` property defines how the menu is aligned,
- * and in which direction it extends.
+ * The value of the `openDirection` property defines how the menu content
+ * is aligned with its trigger element, and in which direction it opens.
```
adrianschmidt, line 13: 👌 🙏

src/components/menu/examples/menu-open-direction.tsx
adrianschmidt: Very nice!

I'm adding a small fixup, which sets a min-width, to avoid this problem:

image

adrianschmidt: Here's a screenshot of the fixup diff:

Screenshot showing that `limel-select { min-width: 8.75rem; }` has been added to `src/components/menu/examples/menu-open-direction.scss`.

I'll go ahead and squash it, since it's so small and I'm about to squash the older fixups now anyway.

adrianschmidt (approved):

Issue #1790, opened 2022-07-12
remove the ripple from limel-icon-button


Kiarokh:
We have removed the ripple effect from many components.
The icon-button is one of them which still has it. For such components, we have used another effect, applied by a mixin called `is-flat-clickable()`. This is the style we normally use for such elements, which show interactivity on hover and click.

On the other hand, there are many lines of code ([for example](https://github.com/Lundalogik/lime-elements/commit/a9fbffeef2525803d59f404a830be619f4a1e6eb) or [this](https://github.com/Lundalogik/lime-elements/issues/1065)) which are done in the component for handling the ripple effect. We can simplify it by removing them. 

Additionally, we cannot scale the component with CSS, which cases issues like this: https://github.com/Lundalogik/crm-feature/issues/2895

Also, there is a buggy behavior with the ripple effect. Sometimes, user needs to click an element, before actually getting a hover effect (from the ripple div).

https://user-images.githubusercontent.com/35954987/178450579-a6049359-b42d-4246-890a-c04773aff457.mov

Another issue with icon-buttons is their size `48px`, which is bigger than buttons `36px`, and button groups `36px`
![image](https://user-images.githubusercontent.com/35954987/178485228-867a76c9-71f0-43f8-9476-4658f83efb36.png)
it makes it hard to place these elements on an action bar for instance, and make everything work visually fine and balanced 
![image](https://user-images.githubusercontent.com/35954987/178489370-4b7056e0-a315-409b-b84d-0b723bcf88d2.png)

## What to do
1. remove the ripple effect (ripple div and all its belongings)
2. then, add our own styles for interactive elements
3. make the height of the icon button smaller

Kiarokh:
Oh this went fast! 😄  Thanks @adrianschmidt

PR #1791, opened 2022-07-12
fix(icon-button): get rid of ripple effect

Kiarokh:
fix https://github.com/Lundalogik/lime-elements/issues/1790

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/icon-button/icon-button.tsx
```diff
@@ -99,12 +62,6 @@ export class IconButton {
             buttonAttributes.tabindex = this.host.getAttribute('tabindex');
         }
 
-        const iconAttributes: { badge?: boolean; size?: IconSize } = {};
-        if (this.elevated) {
-            iconAttributes.badge = true;
-            iconAttributes.size = 'small';
```
Kiarokh, line 105: Screenshot 2022-07-12 at 14 58 32

icon size is now handled in the styles instead, and badge is true always. So background color can be applied to the entire buttons (not just the icon part), using the same old CSS var of `--icon-background-color`.

Kiarokh:
Maybe both commits should be 1 instead. But it made it easier to explain the changes in two commits

src/components/icon-button/icon-button.tsx
adrianschmidt: So, the button is visually slightly larger, but actually takes less space in the page flow? That's nice 👍

adrianschmidt:
> Maybe both commits should be 1 instead. But it made it easier to explain the changes in two commits

Nah, I think it's nice to keep them separate. Much clearer that way 👍

PR #1792, opened 2022-07-13
Feature/#1803/dock

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/1803

### Todo:

- [x] make it possible to shrink and expand the dock (get rid of labels and show tooltips instead of labels)
- [x] make possible to align several dock items to end
- [x] responsive design: for mobile view
- [x] allow consumer to set their own **mobile** breakpoint (component should automatically switch to mobile view after a default breakpoint, but also react to consumer's defined breakpoint. 
- [x] is popover part good?
- [x] add docs
- [x] split and clean up css
- [x] add css vars for max-width of dock
- [x] add docs to `types`
- [ ] emit and tell the outside world that the dock is expanded or not
- [ ] Add labels optionally on mobile devices
- [ ] unit test

#### Todo outside Lime Elements:
- [ ] Remember user's choice for expanded / shrunk
- [x] add a feature switch and use the new dock
- [x] hotkeys? :/ 

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dock/dock-button/dock-button.scss
+++ b/src/components/dock/dock-button/dock-button.scss
@@ -1,0 +1,71 @@
+@use '../../../style/mixins';
+@use '../../../style/functions';
+@use '../../../style/internal/variables';
+
+.button {
+    all: unset;
+
+    box-sizing: border-box;
+    display: flex;
+    align-items: center;
+    position: relative;
+
+    width: 100%;
+    height: var(--dock-item-height);
+
+    border-radius: functions.pxToRem(6);
+    font-size: functions.pxToRem(14);
+    padding: 0 0.5rem;
+    min-width: var(--dock-item-height);
+
+    color: var(--limel-dock-item-text-color);
+    background-color: var(--dock-background-color);
+
+    &:focus-visible {
+        box-shadow: var(--shadow-depth-8-focused);
+    }
+
+    .dock-item:not(.selected) & {
+        &:not(.disabled) {
+            cursor: pointer;
+            @include mixins.is-flat-clickable();
+        }
+    }
+
+    &:hover {
+        z-index: 1;
+    }
+
+    &.selected {
+        color: var(--limel-dock-item-text-color--selected);
+        background-color: var(
+            --dock-item-background-color--selected,
+            rgb(var(--contrast-200))
+        );
+        box-shadow: var(--button-shadow-inset);
+
+        .icon {
+            color: var(--limel-dock-item-text--selected);
+        }
+    }
+}
+
+limel-popover {
+    display: grid; // an ugly hack to make buttons that are wrapped in a popover become fullwidth
+}
+
+.text {
+    @include mixins.truncate-text();
+    padding-left: 0.5rem;
+    padding-right: 0.75rem;
+}
+
+.icon {
+    flex-shrink: 0;
+    width: calc(var(--dock-item-height) - 1rem);
+    height: calc(var(--dock-item-height) - 1rem);
+    color: var(
+        --dock-item-icon-color--deselected,
+        var(--limel-dock-item-text-color)
+    );
+}
--- a/src/components/dock/dock.scss
+++ b/src/components/dock/dock.scss
@@ -1,0 +1,80 @@
+@use '../../style/mixins';
+@use '../../style/functions';
+@use '../../style/internal/variables';
+
+/**
+* @prop --dock-expanded-max-width: The maximum width of the Dock when it is expanded. Defaults to `max-content` which means the Dock will adjust its width to the widest dock item.
+* @prop --dock-background-color: Background color of the whole component, defaults to `--contrast-100`.
+* @prop --dock-item-background-color--selected: Background color of selected dock item, defaults to `--contrast-200`.
+* @prop --dock-item-text-color--deselected: Text of dock items, defaults to `--contrast-1100`.
+* @prop --dock-item-text-color--selected: Text color of selected dock item, defaults to `--contrast-1300`.
+* @prop --dock-item-icon-color: Color of the optional icons used in each dock item. Only affects inactive dock items, defaults to text colors for default or selected states.
+* @prop --popover-surface-width: Defines the width of the popover that is opened for dock items with custom components. Defaults to `auto`.
+*/
+
+:host(limel-dock) {
+    --dock-item-height: 2.75rem;
+    --dock-padding: 0.25rem;
+    --dock-expand-shrink-button-height: 1rem;
+
+    --limel-dock-item-text-color: var(
+        --dock-item-text-color--deselected,
+        rgb(var(--contrast-1100))
+    );
+    --limel-dock-item-text-color--selected: var(
+        --dock-item-text-color--selected,
+        rgb(var(--contrast-1300))
+    );
+
+    isolation: isolate;
+    position: relative;
+
+    display: inline-flex;
+    flex-direction: column;
+
+    background-color: var(--dock-background-color, rgb(var(--contrast-100)));
+    box-shadow: #{functions.pxToRem(3)} 0 #{functions.pxToRem(6)} #{functions.pxToRem(
+                -2
+            )} rgba(var(--color-black), 0.15),
+        #{functions.pxToRem(3)} 0 #{functions.pxToRem(10)} #{functions.pxToRem(
+                -2
+            )} rgba(var(--color-black), 0.05);
+}
+
+:host(limel-dock:not(.has-mobile-layout)) {
+    height: 100%;
+    width: calc((var(--dock-padding) * 2) + var(--dock-item-height));
+}
+
+:host(limel-dock.expanded) {
+    width: var(--dock-expanded-max-width, max-content);
+}
+
+.footer-separator {
+    margin-top: auto;
+    justify-self: flex-end;
+}
+
+nav {
+    box-sizing: border-box;
+    display: inline-flex;
+    flex-direction: column;
+    gap: functions.pxToRem(6);
+    flex-grow: 1;
+
+    padding: var(--dock-padding); // needed for the focus effect
+
+    overflow-y: auto;
+    scrollbar-width: none; // This hides the scrollbars appearing under in Firefox
+    -ms-overflow-style: none; // Same as above for IE 11
+    &::-webkit-scrollbar {
+        display: none; // This hides the scrollbars appearing under in Chrome
+    }
+
+    :host(limel-dock.has-mobile-layout) & {
+        justify-content: space-between;
+        flex-direction: row;
+    }
+}
+
+@import './partial-styles/shrink-expand-button';
--- a/src/components/dock/dock.tsx
+++ b/src/components/dock/dock.tsx
@@ -1,0 +1,178 @@
+import {
+    Component,
+    Event,
+    EventEmitter,
+    h,
+    Host,
+    Prop,
+    State,
+} from '@stencil/core';
+import { DockItem } from './dock.types';
+
+const DEFAULT_MOBILE_BREAKPOINT = 700;
+
+/**
+ * @exampleComponent limel-example-dock-basic
+ * @exampleComponent limel-example-dock-custom-component
+ * @exampleComponent limel-example-dock-mobile
+ * @exampleComponent limel-example-dock-expanded
+ * @exampleComponent limel-example-dock-colors-css
+ * @private
+ */
+@Component({
+    tag: 'limel-dock',
+    shadow: true,
+    styleUrl: 'dock.scss',
+})
+export class Dock {
+    /**
+     * Items that are placed in the dock.
+     */
+    @Prop()
+    public dockItems: DockItem[] = [];
+
+    /**
+     * Items that are placed at the bottom of the dock. (Or at the end in mobile
+     * layout.)
+     */
+    @Prop()
+    public dockFooterItems?: DockItem[] = [];
+
+    /**
+     * A label used to describe the purpose of the navigation element to users
+     * of assistive technologies, like screen readers. Especially useful when
+     * there are multiple navigation elements in the user interface.
+     * Example value: "Primary navigation"
+     */
+    @Prop({ reflect: true })
+    public accessibleLabel?: string;
+
+    /**
+     * Defines the width of the component, when it loads.
+     * - `true`: shows both icons and labels of the Dock items.
+     * - `false`: only shows icons of the doc items, and displays
+     * their labels as tooltip.
+     *
+     * Note: when `useMobileLayout` is `true`, labels will always
+     * be shown as tooltips. Read more below…
+     */
+    @Prop({ reflect: true })
+    public expanded? = false;
+
+    /**
+     * Set to `false` if you do not want to allow end-users
+     * to exapnd or shrink the Dock. This will hide the
+     * expand/shrink button, and the only things that defines
+     * the layout will be the `expanded` property, and
+     * the `mobileBreakPoint`.
+     */
+    @Prop({ reflect: true })
+    public allowResize? = true;
+
+    /**
+     * Defines the breakpoint in pixles, at which the component will be rendered
+     * in a hoizontal layout. Default breakpoint is `700` pixels, which means
+     * when the screen size is smaller than `700px`, the component will automatically
+     * switch to a horizontal layout.
+     */
+    @Prop({ reflect: true })
+    public mobileBreakPoint?: number = DEFAULT_MOBILE_BREAKPOINT;
+
+    /**
+     * Fired when a Dock item has been selected from the dock.
+     */
+    @Event()
+    private itemSelected: EventEmitter<DockItem>;
+
+    /**
+     * Is used to render the component horizontally, and place
+     * the Dock items in a row.
+     */
+    @State()
+    private useMobileLayout = false;
+
+    private resizeObserver: ResizeObserver;
+
+    public componentDidLoad() {
+        this.resizeObserver = new ResizeObserver(this.handleResize);
+        this.resizeObserver.observe(document.body);
+    }
+
+    disconnectedCallback() {
+        this.resizeObserver.disconnect();
+    }
+
+    public render() {
+        return (
+            <Host
+                class={{
+                    dock: true,
+                    expanded: this.expanded,
+                    'has-mobile-layout': this.useMobileLayout,
+                }}
+            >
+                <nav aria-label={this.accessibleLabel}>
+                    {this.dockItems.map(this.renderDockItem)}
+                    {this.renderSeparator()}
+                    {this.dockFooterItems.map(this.renderDockItem)}
+                </nav>
+                {this.renderExpandShrinkToggle()}
+            </Host>
+        );
+    }
+
+    private renderSeparator = () => {
+        return this.useMobileLayout ? null : <span class="footer-separator" />;
+    };
+
+    private renderDockItem = (item: DockItem) => {
+        return (
+            <limel-dock-button
+                class={{
+                    'dock-item': true,
+                    selected: item.selected,
+                }}
+                item={item}
+                expanded={this.expanded && !this.useMobileLayout}
+                useMobileLayout={this.useMobileLayout}
+                onInteract={this.handleDockItemClick}
+            />
+        );
+    };
+
+    private handleDockItemClick = (event: CustomEvent<DockItem>) => {
+        if (!event.detail.selected) {
+            this.itemSelected.emit(event.detail);
+        }
+    };
+
+    private handleResize = () => {
+        if (window.innerWidth <= this.mobileBreakPoint) {
+            this.useMobileLayout = true;
+        } else {
+            this.useMobileLayout = false;
+        }
+    };
+
+    private renderExpandShrinkToggle() {
+        if (this.useMobileLayout || !this.allowResize) {
+            return;
+        }
+
+        return (
+            <button
+                class={{
+                    'expand-shrink': true,
+                    expanded: this.expanded,
+                }}
+                onClick={this.toggleDockWidth}
+            >
+                <limel-icon name="angle_right" />
+            </button>
+        );
+    }
+
+    private toggleDockWidth = () => {
+        this.expanded = !this.expanded;
+    };
+}
--- a/src/components/dock/dock.types.ts
+++ b/src/components/dock/dock.types.ts
@@ -1,0 +1,63 @@
+export interface DockItem {
+    /**
+     * A non-changing value to uniquely identify each item.
+     */
+    id: string;
+
+    /**
+     * Text to display for the item.
+     */
+    label: string;
+
+    /**
+     * Name of the icon to use.
+     */
+    icon: string;
+
+    /**
+     * Additional helper text for the dock item.
+     * Example usage can be a keyboard shortcut to activate the dock item.
+     */
+    helperLabel?: string;
+
+    /* eslint-disable no-irregular-whitespace */
+    /**
+     * Whether the dock item should indicate it is selected.
+     * These dock items normally take the user to a top-level location within
+     * the navigation tree; for example "Home", "Search" or "My Account".
+     * Set `selected` to `true`, when:
+     * - the user interface is showing the same top-level location as the dock
+     * item is pointing at, or
+     * - the user interface is showing a page which is a sub-location of the
+     * top-level location. For example, when user is at
+     * _My Account > Notification Settings_, the dock item of _My Account_
+     * should have the `selected` state.
+     */
+    selected?: boolean;
+    /* eslint-enable no-irregular-whitespace */
+
+    /**
+     * Used to specify a custom component to render as a menu for the dock item.
+     */
+    dockMenu?: DockMenu;
+}
+
+export interface DockMenu {
+    /**
+     * The tag name of a custom component to be displayed in a popover when
+     * clicking on the dock item this menu belongs to.
+     */
+    componentName: string;
+
+    /**
+     * Whether the menu is open.
+     */
+    menuOpen?: boolean;
+
+    /**
+     * Any properties that should be set on the custom component.
+     */
+    props?: {
+        [key: string]: any;
+    };
+}
--- a/src/components/dock/examples/dock-colors-css.tsx
+++ b/src/components/dock/examples/dock-colors-css.tsx
@@ -1,0 +1,84 @@
+import { Component, h, State } from '@stencil/core';
+import { DockItem } from '../dock.types';
+
+/**
+ * Using CSS color variables for theming the Dock
+ *
+ * A few CSS variables can be used to customize the look and feel of the steps.
+ *
+ * :::note
+ * Using CSS variables to tweak the colors, applies the colors globally to the
+ * component, not to individual Dock items!
+ * :::
+ * :::important
+ * Make sure that:
+ * - text has enough contrast with its background and is readable.
+ * :::
+ */
+
+@Component({
+    tag: 'limel-example-dock-colors-css',
+    shadow: true,
+    styleUrl: 'dock-colors-css.scss',
+})
+export class DockColorsCssExample {
+    @State()
+    private dockItems: DockItem[] = [
+        {
+            id: '1',
+            label: 'Home',
+            selected: true,
+            icon: 'home',
+        },
+        {
+            id: '2',
+            label: 'Search',
+            icon: 'search',
+        },
+        {
+            id: '3',
+            label: 'Calls',
+            icon: 'phone',
+        },
+        {
+            id: '4',
+            label: 'Chats',
+            icon: 'chat',
+        },
+    ];
+
+    @State()
+    private dockFooterItems: DockItem[] = [
+        {
+            id: '5',
+            label: 'Settings',
+            icon: 'settings',
+        },
+    ];
+
+    public render() {
+        return (
+            <div class="application">
+                <limel-dock
+                    accessibleLabel="Dock Example: CSS color variables"
+                    dockItems={this.dockItems}
+                    dockFooterItems={this.dockFooterItems}
+                    onItemSelected={this.handleSelected}
+                    expanded={true}
+                />
+            </div>
+        );
+    }
+
+    private handleSelected = (event: CustomEvent<DockItem>) => {
+        const setSelection = (item: DockItem) => {
+            return {
+                ...item,
+                selected: item.id === event.detail.id,
+            };
+        };
+
+        this.dockItems = this.dockItems.map(setSelection);
+        this.dockFooterItems = this.dockFooterItems.map(setSelection);
+    };
+}
--- a/src/components/dock/dock-button/dock-button.tsx
+++ b/src/components/dock/dock-button/dock-button.tsx
@@ -1,0 +1,148 @@
+import { Component, Event, EventEmitter, h, Prop, State } from '@stencil/core';
+import { DockItem } from '../dock.types';
+import { createRandomString } from '../../../util/random-string';
+
+/**
+ * @private
+ */
+@Component({
+    tag: 'limel-dock-button',
+    shadow: false,
+    styleUrl: 'dock-button.scss',
+})
+export class DockButton {
+    /**
+     * @inheritdoc
+     */
+    @Prop()
+    public item!: DockItem;
+
+    /**
+     * @inheritdoc
+     */
+    @Prop({ reflect: true })
+    public expanded? = false;
+
+    /**
+     * @inheritdoc
+     */
+    @Prop({ reflect: true })
+    public useMobileLayout? = false;
+
+    /**
+     * Fired when clicking on the flow item.
+     */
+    @Event()
+    private interact: EventEmitter<DockItem>;
+
+    @State()
+    private isOpen = false;
+
+    @Event()
+    public close: EventEmitter<void>;
+
+    private tooltipId: string;
+
+    constructor() {
+        this.tooltipId = createRandomString();
+    }
+
+    public render() {
+        if (this.item?.dockMenu?.componentName) {
+            return this.renderPopover();
+        }
+
+        return this.renderButton(this.handleClick);
+    }
+
+    private renderPopover() {
+        const CustomComponent = this.item?.dockMenu.componentName;
+
+        return (
+            <limel-popover
+                openDirection={this.useMobileLayout ? 'top' : 'right'}
+                open={this.isOpen || this.item.dockMenu.menuOpen}
+                onClose={this.onPopoverClose}
+            >
+                {this.renderButton(this.openPopover, 'trigger')}
+                <CustomComponent
+                    {...(this.item.dockMenu.props || {})}
+                    onClose={this.onPopoverClose}
+                />
+            </limel-popover>
+        );
+    }
+
+    private renderButton(
+        handleClick: (event: MouseEvent) => void,
+        slot?: string
+    ) {
+        return (
+            <button
+                slot={slot}
+                tabindex="0"
+                id={this.tooltipId}
+                type="button"
+                class={{
+                    button: true,
+                    selected: this.item?.selected,
+                }}
+                onClick={handleClick}
+            >
+                {this.renderIcon()}
+                {this.renderLabel()}
+                {this.renderTooltip()}
+            </button>
+        );
+    }
+
+    private openPopover = (event: MouseEvent) => {
+        event.stopPropagation();
+        this.isOpen = true;
+    };
+
+    private onPopoverClose = () => {
+        this.isOpen = false;
+        this.close.emit();
+    };
+
+    private handleClick = (event: MouseEvent) => {
+        event.stopPropagation();
+        this.interact.emit(this.item);
+    };
+
+    private renderIcon() {
+        if (!this.item.icon) {
+            return;
+        }
+
+        return <limel-icon name={this.item.icon} class="icon" />;
+    }
+
+    private renderLabel() {
+        if (this.expanded) {
+            return <span class="text">{this.item.label}</span>;
+        }
+    }
+
+    private renderTooltip() {
+        if (!this.expanded && this.item.label) {
+            return (
+                <limel-tooltip
+                    elementId={this.tooltipId}
+                    label={this.item.label}
+                    helperLabel={this.item.helperLabel}
+                />
+            );
+        }
+
+        if (this.expanded && this.item.helperLabel) {
+            return (
+                <limel-tooltip
+                    elementId={this.tooltipId}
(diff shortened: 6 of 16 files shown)


src/components/dock/dock.tsx
```diff
@@ -0,0 +1,131 @@
...
+    @Prop({ reflect: true })
+    public mobileBreakPoint: number = DEFAULT_MOBILE_BREAKPOINT;
+
+    /**
+     * xxxxxx
+     */
+    private hasMobileLayout = false;
+
+    public componentWillLoad() {
+        if (window.innerWidth <= this.mobileBreakPoint) {
```
Kiarokh, line 59: **⚠️ for the reviewer:** the reason we couldn't add CSS variables for consumers to define their mobile breakpoint is this: https://bholmes.dev/blog/alternative-to-css-variable-media-queries/

src/components/dock/dock-item/dock-item.tsx
```diff
@@ -0,0 +1,165 @@
...
+ * @private
+ */
+@Component({
+    tag: 'limel-dock-item',
+    shadow: false,
+    styleUrl: 'dock-item.scss',
+})
+
+// buttons in a menu component
+export class DockItemMenu {
```
adrianschmidt, line 23: Hmm… I'm having some trouble following your thoughts here 🙂

The tag for this component is `limel-dock-item`, but the class name is `DockItemMenu`. I didn't pay that much mind at first, as I just assumed that you had renamed the component and forgotten the class name or something, which I do myself all the time.

But then I realised that there's an interface called `DockItem` and this component does _not_ implement that interface, but has a property called `item` that expects an object of type `DockItem` 😮

I feel very confused right now… 😕

vicgeralds: I was also very confused here, expecting a menu that should be passed items as a prop, not a menu item passed a single item along with an optional component prop that could implement a menu...

Befkadu1: The interface should be `DockItemType`

Kiarokh: ⚡️

adrianschmidt: > It's been 10 months since we started playing with this component.

😮

src/components/dock/dock.tsx
```diff
@@ -0,0 +1,187 @@
...
+     * Items that are placed within the `<nav />` section of the dock.
+     */
+    @Prop()
+    public dockItems: DockItem[] = [];
+
+    /**
+     * Fired when a Dock item has been selected from the dock.
+     */
+    @Event()
+    public change: EventEmitter<DockItem>;
```
vicgeralds, line 42: Stencil warns about using any event name that conflicts with some native DOM event name. Don't we want to another name?

vicgeralds: I'm also thinking that we might want to emit an event with detail `DockItem['value']` instead?

src/components/dock/dock-item/dock-item.tsx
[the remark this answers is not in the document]
adrianschmidt: I'll be working a bit on the interface stuff after helping Lucy with the popover arrow, if that's ok with everyone. So no need for anyone else to change this right now (in fact, I'd prefer it if there wasn't too many changes made at the moment, so I can get a chance to really wrap my head around things, and maybe do some changes without getting loads of merge conflicts later on 😊).

src/components/dock/dock.tsx
adrianschmidt: My thoughts on this has originally been that we are (or were, at the time), mostly making replacements for standard components. Like, instead of just using `<button>`, you'd use `<limel-button>` and get our styling and perhaps some nice extra functionality. As such, it would make sense for the consumer to expect the same kind of events from `<limel-x>` as from the native `<x>`. I still think that to some extent, but with the warnings not being possible to disable, and with me seeing the following warning for the first time today, a warning that has clearly been hiding in plain sight and missed because of all the warnings about overriding native event names, I'm really starting to change my mind quite rapidly… 😣

Build Warn: ./src/components/input-field/input-field.tsx:104:12 The @Prop() name &quot;prefix&quot; is a reserved public name. Please rename the &quot;prefix&quot; prop so it does not conflict with an existing standardized prototype member. Reusing prop names that are already defined on the element's prototype may cause unexpected runtime errors or user-interface issues on various browsers, so it's best to avoid them entirely.

adrianschmidt: Should we prefix these with `limel` too? So `limelChange` in this case?

Since changing existing ones will be a breaking change, perhaps we should at least start prefixing any new ones we add?

adrianschmidt: Ah, I honestly didn't look at what the event in question was in this specific instance, so I totally missed that the current name wasn't very fitting. I agree that this event should not be called `change`, regardless of any other event-naming concerns we have more generally 😄

`itemSelected` sounds good to me 👍

adrianschmidt: I guess we need to check if using capital letters in an event name might cause issues with jsx though. I'm thinking if `onItemSelected={…}` might be interpreted to mean the event name is `item-selected`?

src/components/dock/dock.types.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * Background color of selected Dock item.
+     */
+    selectedBackgroundColor?: string;
+
+    /**
+     * Value of the Dock item.
+     */
+    value?: T;
```
Kiarokh, line 35: should we call this `id` instead?

vicgeralds: If an `id` could be useful (for example for aria labelling purposes), we can make it work without a generic value and have `id: string` instead.

But I think a generic value is more useful, with an id you have to lookup the value that you need in your event handler.

src/components/dock/dock.tsx
vicgeralds: Camel cased events are normal in JSX, and that's also what Stencil uses in their examples: https://stenciljs.com/docs/events

I'm not sure now if they're transformed to lowercase though. Native events are always lowercase, even if React JSX uses camel casing for them...

vicgeralds: Update: Camel cased custom events are not transformed to lowercase, so `onItemSelected` is same as listening to `itemSelected`.

Native event names also need to be camel cased in JSX, but are transformed to lowercase (with Stencil, you will get a type error if you try to use `onTransitionend` instead of `onTransitionEnd`.

vicgeralds: > If we avoid single words for our own events, we will be fine.

Appears I was wrong here, at least when using JSX in Stencil. If I define a custom event with

```ts
@Event()
    public transitionEnd: EventEmitter;
```

I will get a build warning about event name conflict (if I also use that event emitter). But the emitted event will still be `transitionEnd` which is not really conflicting with the native DOM event when using `element.addEventListener`, etc. But there is a conflict when using JSX, so you will end up listening to the native event instead!

Still, we're less likely to run into event name conflicts when using camel cased event names because most native DOM events are single words. But then we really _have to_ avoid those name conflicts, otherwise it may be impossible to bind an event listener using JSX.

src/components/dock/dock.types.ts
```diff
@@ -0,0 +1,88 @@
+export interface DockItem {
```
adrianschmidt, line 1: @Kiarokh @LucyChyzhova I removed some things, like colors for example, from the interface (temporarily), because I wanted to make it as clean as possible to start out from, both to make things easier for myself and to make it a little easier to show and tell about my changes.

There's no problem adding them back in 😊

adrianschmidt: (Although, do we really need to set stuff like button background colour? Do we have an existing use case for that? 🤔)

src/components/dock/dock.tsx
```diff
@@ -0,0 +1,186 @@
...
+                class={{
+                    dock: true,
+                    expanded: this.expanded,
+                    'has-mobile-layout': this.useMobileLayout,
+                }}
+            >
+                <nav aria-labelledby={this.ariaLabelledBy}>
+                    {this.dockItems.map(this.renderDockItem)}
+                    <span class="footer-separator" />
+                    {this.footerItems.map(this.renderDockItem)}
```
adrianschmidt, line 124: The `<span class="footer-separator" />` is hidden, but is used to target the first item that comes after it, applying the extra margin to it.

src/components/dock/dock.tsx
```diff
@@ -0,0 +1,186 @@
...
+    }
+
+    private renderDockItem = (item: DockItem) => {
+        let ButtonComponent = 'limel-default-dock-button';
+        if (item.dockButton?.componentName) {
+            ButtonComponent = item.dockButton.componentName;
+        }
+
+        return (
+            <ButtonComponent
```
adrianschmidt, line 138: `limel-default-dock-button`, formerly known as `limel-dock-item`, is no longer involved at all if someone defines a custom component to be used to render the button. Instead, if there's a custom dock-button configured, we render that directly, if not, we default to using `limel-default-dock-button`.

src/components/dock/dock.types.ts
```diff
@@ -1,75 +1,88 @@
...
 
     /**
-     * Background color of selected Dock item.
+     * Used to specify a custom component to render as a menu for the button.
      */
-    selectedBackgroundColor?: string;
+    dockMenu?: DockMenu;
+}
 
+export interface DockButton {
```
Befkadu1, line 40: The interface `DockButton` and `DockMenu` is exactly same. The other thing I would like to say here is we might not need `dockButton?: DockButton;` for now at least. It is better to add the property when we need it. 
Having and not using it confuses the consumer. What do you think?

src/components/dock/dock.types.ts
```diff
@@ -0,0 +1,88 @@
...
+     */
+    componentName: string;
+
+    /**
+     * Any properties that should be set on the custom component.
+     */
+    props?: {
+        [key: string]: any;
+    };
+}
```
adrianschmidt, line 68: @Befkadu1 pointed out that these two interfaces are identical, and yes, there's no reason we need two of them, I'd say. They started out different from each other, but then ended up the same, so I think we can just remove one and come up with a name for the remaining one that works for both cases.

src/components/dock/dock.types.ts
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Whether the button should indicate it is selected.
+     * The button should be selected when … (@Kiarokh: add description here please!)
+     */
+    selected?: boolean;
+
+    /**
+     * Used to specify a custom component to render as the dock button.
+     */
+    dockButton?: DockButton;
```
adrianschmidt, line 32: This is for defining a custom component to render a button in the dock, instead of the default one we'd otherwise use. (I imagine it will be quite rare that anyone would use this…)

src/components/dock/dock.types.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Used to specify a custom component to render as a menu for the button.
+     */
+    dockMenu?: DockMenu;
```
adrianschmidt, line 37: This is for the content of the popover.

src/components/dock/dock.types.ts
adrianschmidt: Although, even thought the signatures are identical, perhaps it's valuable to keep the separate documentation? It might be less confusing when looking at the docs, rather than the code… 🤔

src/components/dock/dock.types.ts
```diff
@@ -0,0 +1,88 @@
...
+    expanded?: boolean;
+
+    /**
+     * Tells the individual items whether the dock has a horizontal or vertical
+     * layout. When `true`, the dock is horizontal, and placed at the bottom of
+     * the viewport. Otherwise, the dock is vertical, and placed on the left
+     * edge of the viewport (for RTL languages).
+     */
+    useMobileLayout?: boolean;
+}
```
adrianschmidt, line 88: This is the interface that you should implement in order to make a custom component that can be sent in to `DockItem.dockButton`.

src/components/dock/default-dock-button/default-dock-button.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+/**
+ * @private
+ */
+@Component({
+    tag: 'limel-default-dock-button',
+    shadow: false,
+    styleUrl: 'default-dock-button.scss',
+})
+export class DefaultDockButton implements CustomDockButton {
```
adrianschmidt, line 13: It may look odd that `limel-default-dock-button` implements `CustomDockButton`, but if you see it from the point of view of the consumer, `CustomDockButton` is the interface you should implement if you wish to make a custom dock button!

src/components/dock/examples/dock-colors.tsx
```diff
@@ -0,0 +1,101 @@
+import { Component, h, State } from '@stencil/core';
+import { DockItem } from '../dock.types';
+
+/**
+ * Using color properties for individual Dock items
```
adrianschmidt, line 5: This one doesn't work at the moment, since I removed the colour variables from the interface. But if there really is a current use case for them, we should of course add them back in. If there isn't, I would argue that we should add the support when we have a use case. But I'm not ardent about it 😄

adrianschmidt (commented):
Here are some comments about the changes I made:

src/components/dock/dock.tsx
```diff
@@ -0,0 +1,186 @@
...
+    public render() {
+        return (
+            <Host
+                class={{
+                    dock: true,
+                    expanded: this.expanded,
+                    'has-mobile-layout': this.useMobileLayout,
+                }}
+            >
+                <nav aria-labelledby={this.ariaLabelledBy}>
```
adrianschmidt, line 121: `aria-labelledby` is not the correct attribute to use:

> The [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) attribute identifies the element (or elements) that labels the element it is applied to.

We could use `aria-label`, or we could put the text into an element, give it an `id`, and then refer to that `id` with `aria-labelledby` 🙂

If it's more of a description than a label, we could use `aria-describedby`. `aria-description` also exists, but apparently isn't ready for live use yet.

> The [aria-description](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description) attribute is similar to [aria-label](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) in that both provide a text string to associate with the element, but a label should be short and concise, while the description can be longer as it is intended to provide more context and information.

> Note: [aria-description](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-description) is still in W3C Editor's Draft for ARIA 1.3. For the time being, continue to use [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-describedby), which has been supported since ARIA 1.1.

src/components/dock/examples/dock-colors-css.tsx
```diff
@@ -0,0 +1,84 @@
+import { Component, h, State } from '@stencil/core';
+import { DockItem } from '../dock.types';
+
+/**
+ * Using CSS color variables for themeing the Dock
```
adrianschmidt, line 5:
```suggestion
 * Using CSS color variables for theming the Dock
```

src/components/dock/dock.types.ts
adrianschmidt: Yeah, true. I think removing the option to have a custom dock-button is better than keeping it and merging the two interfaces, because I think the latter would be even more confusing.

adrianschmidt: I'll copy this to the thread I made about it, and hide this thread, so we can keep it in one place 👍

src/components/dock/dock.types.ts
adrianschmidt: @Befkadu1 wrote:

> The interface `DockButton` and `DockMenu` is exactly same. The other thing I would like to say here is we might not need `dockButton?: DockButton;` for now at least. It is better to add the property when we need it. 
Having and not using it confuses the consumer. What do you think?

adrianschmidt: Yeah, true. I think removing the option to have a custom dock-button is better than keeping it and merging the two interfaces, because I think the latter would be even more confusing.

src/components/dock/dock.types.ts
Kiarokh: would these examples suffice 😄 
image
image

I just feel colors are so basic things that most consumers would probably wanna tweak. 
So I'd rather have them back. We also have an example that showcased usage of colors.

Maybe having button background color sounds a bit overkill. But icon color, not :)

src/components/dock/dock.scss
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+.footer-separator {
+    display: none;
+}
+.footer-separator + .dock-item {
+    margin-top: auto;
+    justify-self: flex-end;
+}
```
Kiarokh, line 77:
```suggestion
.footer-separator {
    margin-top: auto;
    justify-self: flex-end;
}
```

Kiarokh: I argue this would be less confusing

src/components/dock/dock.types.ts
adrianschmidt: Ah, ok, yeah, that's definitely example enough! 👍 😄

src/components/dock/dock.tsx
```diff
@@ -0,0 +1,182 @@
...
+     */
+    @Prop()
+    public dockItems: DockItem[] = [];
+
+    /**
+     * Items that are placed at the bottom of the dock. (Or at the end in mobile
+     * layout.)
+     */
+    @Prop()
+    public footerItems: DockItem[] = [];
```
Kiarokh, line 40:
```suggestion
    public dockFooterItems: DockItem[] = [];
```

Kiarokh: footer might be misleading. The footer items are still dock items, so maybe this is a better name

src/components/dock/dock.scss
adrianschmidt: Yeah, no argument from me. I just wanted to keep the styling the same as what you had already done, in case adding in a new element would cause any problems for you. If it doesn't, then let's go with this simpler version 👍

Kiarokh: let's go ;)

src/components/dock/dock.tsx
Kiarokh: Aha ok.
They said:
> A document may have several [<nav>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav) elements, for example, one for site navigation and one for intra-page navigation. [aria-labelledby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby) can be used in such case to promote accessibility, see [example](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#labeling_section_content).

But then I didn't check their example 🙈 👍 

```html
<header>
  <nav aria-labelledby="primary-navigation">
    <h2 id="primary-navigation">Primary navigation</h2>
    <!-- navigation items -->
  </nav>
</header>

<!-- page content -->

<footer>
  <nav aria-labelledby="footer-navigation">
    <h2 id="footer-navigation">Footer navigation</h2>
    <!-- navigation items -->
  </nav>
</footer>
```

Shall we go for `aria-label` instead?

src/components/dock/dock.scss
adrianschmidt: No, it doesn't work in mobile view:

My suggestion:
Screenshot showing no extra space between the last &quot;header&quot; button and the first &quot;footer&quot; button in mobile view

Your suggestion:
Screenshot showing extra space between the last &quot;header&quot; button and the first &quot;footer&quot; button in mobile view

Kiarokh: Ah. good catch. In there we can either: 
```scss
:host(.has-mobile-layout) {
   .footer-separator {
        display: none;
    }
}
````

or not render it at all. 
renderFooterSeparator if !useMobileLayout

On mobile layout, we don't have any need for dividing items. They should rather be equally spread through the available width, according to the existing conventions.

File: src/components/dock/dock.tsx (hunk 15 above)
Comment (adrianschmidt) [reply src/components/dock/dock.tsx]: ⚡

I've changed to `aria-label`, and also made the property required. Is it too much to make it required? We want to strongly encourage accessibility, but maybe a label isn't always necessary to make the functionality clear? Also, it may be better to have no label than have some really bad label that someone just added because the property was required…?

Hmm… I think I'm changing my mind on this required thing… 😅

File: src/components/dock/dock.scss (hunk 17 above)
Comment (adrianschmidt) [reply src/components/dock/dock.scss]: ⚡

File: src/components/dock/dock.tsx (hunk 7 above)
Comment (adrianschmidt) [reply src/components/dock/dock.tsx]: I've changed this, removing the support for a custom button. We can add that if and when we actually need it instead.

File: src/components/dock/dock.types.ts (hunk 9 above)
Comment (adrianschmidt) [reply src/components/dock/dock.types.ts]: ⚡

File: src/components/dock/dock.types.ts (hunk 10 above)
Comment (adrianschmidt) [reply src/components/dock/dock.types.ts]: Removed.

File: src/components/dock/dock.types.ts (hunk 12 above)
Comment (adrianschmidt) [reply src/components/dock/dock.types.ts]: This interface has been removed, since the support for custom buttons has been removed.

File: src/components/dock/dock.tsx (hunk 15 above)
Comment (adrianschmidt) [reply src/components/dock/dock.tsx]: @Kiarokh Thoughts on this prop being required or not?

File: src/components/dock/dock.types.ts (hunk 5 above)
Comment (adrianschmidt) [reply src/components/dock/dock.types.ts]: Hmm… I checked out the commit just before my first commit, and the example _ Using color properties for individual Dock items_ only seems to support `iconColor`. `selectedBackgroundColor` and `selectedTextColor` doesn't seem to be applied at all? 🤨

I think we should get a first version functional and merged, and add further support for colours and other extras once we have a working MVP.

File: src/components/dock/dock.tsx (hunk 3 above)
Comment (adrianschmidt) [reply src/components/dock/dock.tsx]: ⚡

Event name changed from `selected` to `itemSelected`.

File: src/components/dock/dock.types.ts (hunk 4 above)
Comment (adrianschmidt) [reply src/components/dock/dock.types.ts]: I changed this to `id: string`, but Bef wanted to keep the option to use a custom type, so he's added `value: T` again. I'm not fully convinced Lime Elements should have to care about what data a consumer wants to "attach" to the dock item, so I'm not sure we should keep the `value` property. I think it might make more sense for the consumer to use the `id` to look up the additional data needed.

Regardless, I'm not a fan of the property name `value` in this instance. It makes sense for things like fields, but a dockItem doesn't really have a "value", in my opinion. This is basically just a way for a consumer to attach some data to the item, that they would like to get back along with the item when the `itemSelected` event is emitted. As such, I think it makes more sense either to have a property like `eventPayload: any`, or simply allow any extra properties on the object by adding `[key: string]: any`.

The problem with specifying an "attachment" like this is that we basically take on responsibility for it. I don't like that at all. It feels very contrary to the SOLID principles to me… 😞

File: src/components/dock/dock.tsx (hunk 15 above)
Comment (adrianschmidt) [reply src/components/dock/dock.tsx]: I made it optional again.

Comment (Kiarokh) [reply src/components/dock/dock.tsx]: Well, I think we can't really play the Moral Police here. 😄 There are many things that are "required", when it comes to accessibility; but people skip them and we can't do much. 

For this specific case of `limel-dock`, the accessibility-damage of not providing a required aria-label is not so big. The element is already using `nav` which makes it way more accessible. Providing `aria-label` is good specially when there are multiple navigation methods. Since this Dock thingy is most probably used within an "app UI" (not a website UI), there is good reasons to believe that there is only this primary navigation way. (unless folks are using something like a Breadcrumbs elements or similar. 

Anyhow, I think we can easily leave the responsibility to the user here. Our responsibility was providing this prop.

Nevertheless, if we wanna play the Moral Police and provide something like this as "required" –which we think that people may not provide a proper value for– I'd say we can already be a good police and fallbacks in place (in translation files) ourselves.

We would need these anyways for our own CRM, so I already have collected them via [Microsoft Language portal](https://www.microsoft.com/en-us/language/Search?&searchTerm=primary%20navigation&langID=750&Source=true&productid=0).

<table>
<tr>
	<td>🇬🇧 en: 
	<td>Primary navigation
<tr>
	<td>🇸🇪 sv: 
	<td>Primär navigering
<tr>
	<td>🇳🇴 no: 
	<td>Primærnavigasjon
<tr>
	<td>🇩🇰 da: 
	<td>Primaire navigatie
<tr>
	<td>🇫🇮 fi: 
	<td>Ensisijainen siirtyminen
<tr>
	<td>🇩🇪 de: 
	<td>Primäre Navigation
<tr>
	<td>🇫🇷 fr: 
	<td>Navigation principale
<tr>
	<td>🇳🇱 nl: 
	<td>Primaire navigatie
</table>

File: src/components/dock/dock.tsx [hunk 19]
```diff
@@ -0,0 +1,177 @@
...
+     */
+    @Prop()
+    public dockItems: DockItem[] = [];
+
+    /**
+     * Items that are placed at the bottom of the dock. (Or at the end in mobile
+     * layout.)
+     */
+    @Prop()
+    public dockFooterItems?: DockItem[] = [];
```
Kiarokh, line 39: this may not be a big deal, but this is not the way we do such a thing in `limel-progress-flow`. In there, we put `isOffProgress` on each item which has to go to the "end group": https://github.com/Lundalogik/lime-elements/blob/ebc7b54ee90f43df483f067cd0241e685df5b276/src/components/progress-flow/examples/progress-flow-off-progress-steps.tsx#L49

So we change out convention here a bit. 

However, we will most probably have a 3rd group of Dock items in future, which may make this a good decision (?). The concept which is not fully designed in details is that user may pin some custom things into the dock. For example, they may pin a shortcut to an object or to a place (app, visualization, etc…) in the CRM. They may wanna do so to make their shortcut accessible from anywhere within the app, and remove the need of going back to for example the dashboard, to access a shortcut.

With this in mind, it may already be a good plan to divide the Dock items into 2 groups now, and add a 3rd group later when we are ready with the design. We know we will have something like this, but don't know how it will work in details yet.

src/components/dock/dock.types.ts
Kiarokh: If we start with `iconColor`, we need to add tweaks in the CSS files as well. I need to see the code to be able to answer why you don't get colors.

However, I think we can drop `selectedBackgroundColor` and `selectedTextColor` for now. If you push, I will tweak the CSS accordingly

src/components/dock/dock.tsx
adrianschmidt: I'm a little divided on this myself. Mostly I just didn't like that you were supposed to set a certain property only on the first of the "special" items. If you'd set the property on all items that should go in the bottom, I'd be much more ok with it.

src/components/dock/dock.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+    /**
+     * A label used to describe the purpose of the navigation element inside the
+     * dock to users of assistive technology, like screen readers.
+     * Example value: "Main Site Navigation"
```
Kiarokh, line 44:
```suggestion
     * A label used to describe the purpose of the navigation element
     * to users of assistive technologies, like screen readers.
     * Specially useful when there are multiple navigation elements in the user interface.
     * Example value: "Primary navigation"
```

adrianschmidt: Done. But corrected "Specially" to "Especially" 🙂

src/components/dock/dock.types.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Additional helper text for the button.
+     * Example usage can be a keyboard shortcut to activate the button.
+     */
+    helperLabel?: string;
+
+    /**
+     * Whether the button should indicate it is selected.
+     * The button should be selected when … (@Kiarokh: add description here please!)
```
adrianschmidt, line 25: @Kiarokh Can you improve this description for me? I didn't really know what to write… 😄

adrianschmidt:
> #### Todo outside Lime Elements:

I split out the things I think belong outside Lime Elements, into its own todo-list.

> * [ ]  Remember user's choice for expanded / shrunk

I think this one should be done by the consumer, as we don't know how the consumer might prefer storing it (or if they even prefer not to store it). In the case of the Lime CRM web client, for example, we might want to store it either in localStorage or in userData…

src/components/dock/dock-button/dock-button.scss
```diff
@@ -0,0 +1,65 @@
+@use '../../../style/mixins';
```
adrianschmidt, line 1: > * [ ]  Add labels optionally on mobile devices

@Kiarokh, can you elaborate on what this one means? 🙂

Kiarokh: Yes. This was a discussion we had with Lucy and decided to leave it for future. 

From accessibility perspective, it's good to have labels together with icons. Which is why many apps have labels on their mobile nav-bars.

**App Store**
IMG_7966B35DAAC0-1
**Linkedin**
IMG_02AC416698D7-1

However, many other also chose not to, like 
**Instagram**
IMG_6E5A81968945-1

So we thought we could enable this as an optional feature for the component in future. So if consumer or end-user wants to show labels, a setting can enable them to do so.

src/components/dock/dock.scss
```diff
@@ -0,0 +1,93 @@
...
+@use '../../style/internal/variables';
+
+/**
+* @prop --dock-expanded-max-width: The maximum width of the Dock when it is expanded. Defaults to `max-content` which means the Dock will ajdust its width to the widest dock item.
+* @prop --dock-background-color: Background color of the whole compoentn, defaults to `--contrast-100`.
+* @prop --dock-item-background-color--deselected: Background color of buttons, defaults to `--dock-background-color`.
+* @prop --dock-item-background-color--selected: Background color of selected button, defaults to `--contrast-200`.
+* @prop --dock-item-text-color--deselected: Text of buttons, defaults to `--contrast-1100`.
+* @prop --dock-item-text-color--selected: Text color of selected button, defaults to `--contrast-1300`.
+* @prop --dock-item-icon-color: Color of the optional icons used in each button. Only affects inactive buttons, defaults to text colors for default or selected states.
```
Kiarokh, line 12: is there a need to change all `--dock-item-…` variable names to `--dock-button-…`?

src/components/dock/dock.scss
```diff
@@ -0,0 +1,93 @@
...
+* @prop --popover-surface-width: Defines the width of the popover that is opened for dock items with custom components. Defaults to `auto`.
+*/
+
+:host {
```
Kiarokh, line 16:
```suggestion
:host(limel-dock) {
```

src/components/dock/dock.scss
```diff
@@ -0,0 +1,93 @@
...
+                -2
+            )} rgba(var(--color-black), 0.05);
+}
+
+:host(:not(.has-mobile-layout)) {
+    height: 100%;
+    width: calc((var(--dock-padding) * 2) + var(--dock-item-height));
+}
+
+:host(.expanded) {
```
Kiarokh, line 58:
```suggestion
:host(limel-dock.expanded) {
```

src/components/dock/dock.scss
```diff
@@ -0,0 +1,93 @@
...
+}
+
+:host(.has-mobile-layout) {
+    width: 100%;
+    height: calc((var(--dock-padding) * 2) + var(--dock-item-height)); // FIXME
+    nav {
+        justify-content: space-between;
+        flex-direction: row;
+    }
+}
```
Kiarokh, line 69:
```suggestion
```

src/components/dock/dock-button/dock-button.scss
adrianschmidt: Ah! Yeah, that makes sense! (Both having it, and not adding it right now 😄)

src/components/dock/dock.scss
Kiarokh: The width and height don't do anything here not needed. 

For the `nav` styles, check below. I moved them to the `nav` itself. 👇

src/components/dock/dock.scss
```diff
@@ -0,0 +1,93 @@
...
+
+    padding: var(--dock-padding); // needed for the focus effect
+
+    overflow-y: auto;
+    scrollbar-width: none; // This hides the scrollbars appearing under in Firefox
+    -ms-overflow-style: none; // Same as above for IE 11
+    &::-webkit-scrollbar {
+        display: none; // This hides the scrollbars appearing under in Chrome
+    }
+}
```
Kiarokh, line 91:
```suggestion
    }
    :host(limel-dock.has-mobile-layout) & {
    	justify-content: space-between;
    	flex-direction: row;
	}
}
```

src/components/dock/dock.scss
adrianschmidt: Hmm… maybe…

When we supported the option to use a custom component for it, I called the interface `CustomDockButton`. If we support it again in the future, I'd guess that might be the name we'd use. But at the moment we don't support that, so I don't think the consumer will actually see any references to any `dock-button` or `dockButton`. (Unless I'm forgetting something?)

So then it seems like it would make more sense to a consumer that we prefix the css variables with `--dock-item` rather than `--dock-button`. I think? Thoughts? 😅

src/components/dock/dock.tsx
adrianschmidt: Sum up of discussion from standup:

I was considering merging these two arrays, and adding a property that should be set on all items that should go in the footer. But during standup Kia showed some ideas for the future, where we'd have a kind of "history" of previous visited locations in the app, or maybe user-pinned locations, between the "head" and the "footer" items, and then I felt that having them split is probably quite a good idea. Merging the arrays would just mean we need to loop through them and pick out what to render where, every time we updated the array.

src/components/dock/dock.scss
Kiarokh: yes you're right. We still send the array as `private dockItems: DockItem[] = [` so it makes sense to call them `--dock-item-`. Then based on what you said regarding future plans, we don't have to change any filename or tag name here, right?
image
or should we do it for consistency anyways?

adrianschmidt:
There's an unsquashed copy of this branch, called `feature/#1803/dock-unsquashed`. That can be used to find things that has been removed in this PR, if and when we want to add some of those things again.

src/components/dock/dock.scss
adrianschmidt: No, I think `--dock-item` is the better option. And we shouldn't have to change that even if we do allow custom dock buttons in the future. Nothing will really force a custom dock button so use those variables anyway (although a good custom dock button probably should 😄).

adrianschmidt: (I've edited the above. I had written "Nothing will really force a custom dock button **so** use those variables anyway" but meant "Nothing will really force a custom dock button **to** use those variables anyway". That seemed like an important typo to fix 😅)

src/components/dock/dock.scss
```diff
@@ -0,0 +1,81 @@
...
+* @prop --dock-item-icon-color: Color of the optional icons used in each button. Only affects inactive buttons, defaults to text colors for default or selected states.
+* @prop --popover-surface-width: Defines the width of the popover that is opened for dock items with custom components. Defaults to `auto`.
+*/
+
+:host(limel-dock) {
+    --dock-item-height: 2.75rem;
+    --dock-padding: 0.25rem;
+    --dock-expand-shrink-button-height: 1rem;
+
+    --limel-dock-item-text-color: var(
```
Kiarokh, line 21: I renamed these internal variables to `--limel-…`. I believe this could be a good convention to follow from now on. Private variables that are used for making things work, can always start with a `--limel-` prefix. We have many of these in other components. Would be good to change them too in some beautiful future. :)

src/components/dock/dock.scss
Kiarokh: I can tell you that I was contemplating on the comment for a while, and then gave up 😄 

Also didn't really understand if you will rename the files back to original or not. E.g. `dock-button.tsx` -> `dock-item.tsx` . But I guess you won't. Either way I'm fine. :)

adrianschmidt: I hadn't considered renaming the files… 🤔

Anyway, I haven't even gotten started on emitting that event for expanding and collapsing the dock, so I need to get working on that after lunch. All these discussions are distracting me 😅

src/components/dock/dock-button/dock-button.scss
```diff
@@ -18,8 +18,11 @@
     padding: 0 0.5rem;
     min-width: var(--dock-item-height);
 
-    color: var(--button-text);
-    background-color: var(--button-background);
+    color: var(--limel-dock-item-text-color);
+    background-color: var(
+        --dock-item-background-color--deselected,
```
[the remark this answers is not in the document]
Kiarokh: I agree. Lucy and I also had this discussion. But since "unselected" didn't seem to be a "word" according to [Longman Dictionary](https://www.ldoceonline.com/spellcheck/english/?q=unselected) at least, we decided to go with [deselected](https://www.ldoceonline.com/dictionary/deselect).

Anyhow, maybe a better suffix would be `…--default` instead of deselected or unselected?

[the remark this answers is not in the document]
adrianschmidt: Or, we don't need a suffix for the default value, and just have suffixes for the special cases, like `--selected` and `--recently-deselected`.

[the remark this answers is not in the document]
Kiarokh: OK, then let's remove it from here as well. We have never had anything for default or unselected. Can I quickly do it before you merge this?

src/components/dock/dock.types.ts
adrianschmidt: `value` has been removed again, and Bef found a way to work around it in crm-components 👌

adrianschmidt (approved):
🎉

src/components/dock/dock-button/dock-button.scss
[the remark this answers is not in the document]
adrianschmidt: Ah, sorry, I missed this. Good thing it's still a private component, so we don't need to make it a breaking change! 😅

PR #1797, opened 2022-07-20
fix(button): remove the `ripple` div & add our own keyboard focus styles

Kiarokh:
We have removed the ripple effect from almost all elements.
This will unify our styles more.
In this component, the original Material Design ripple was not
showing at all when clicking. Probably due to lack of some `import`s.
The div was only used to add a not-so-visible highlight
when the button was focused using keyboard.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -20,6 +20,9 @@
 
 button {
     @include mixins.is-elevated-clickable();
+    &:focus-visible {
+        box-shadow: var(--shadow-depth-8-focused);
+    }
 
     &.mdc-button {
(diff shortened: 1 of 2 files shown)


adrianschmidt (approved):
_YES!_ Love it! So much win for so little code! 😍

PR #1798, opened 2022-07-20
1827 arrow popper js popover

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/1827
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/portal/portal.tsx
```diff
@@ -161,7 +162,7 @@ export class Portal {
     }
 
     public render() {
-        return <slot />;
+        return [<slot />, <div data-popper-arrow id="arrow"></div>];
```
Kiarokh:
```suggestion
        return [<slot />, <div data-popper-arrow id="arrow" />];
```

Kiarokh: would this work?

src/components/portal/portal.scss
```diff
@@ -11,11 +11,49 @@
...
+
+#arrow,
+#arrow::before {
     position: absolute;
-    inset: 0px auto auto 0px;
-    margin: 0px;
-    transform: translate3d(175px, 370px, 0px);
-    background-color: aqua;
+    width: 8px;
+    height: 8px;
```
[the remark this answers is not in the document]
adrianschmidt: I think this is still very much WIP, so I'm not sure you need to review it yet… 🤨 😄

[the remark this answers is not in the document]
LucyChyzhova: yes, it's in progress 😉

src/components/portal/portal.tsx
```diff
@@ -111,6 +112,8 @@ export class Portal {
 
     private container: HTMLElement;
 
+    private arrow: HTMLElement;
```
Kiarokh, line 115: I wonder if we should make this optional. So if the consumer wants to show an arrow for the popover, they can choose that. In that case, we render the arrow (and add some proper margins around the surface so that the arrow doesn't cover the trigger).

However, we would probably don't want to have this as an optional thing for the tooltip component 🤔 I guess tooltips should be always with arrows. What do you think?

adrianschmidt: I agree 👍

LucyChyzhova: done :)

Kiarokh:
super nice work! Now it's really the time to fix this bug: https://github.com/Lundalogik/lime-elements/issues/1408 😄

Kiarokh:
Oh, perhaps we can have a **Composite** example for the popover and also talk about the arrow feature a little bit.

src/components/portal/portal.tsx
```diff
@@ -78,6 +79,18 @@ export class Portal {
...
     public inheritParentWidth = false;
 
+    /**
+     * Set `showArrow` to `true` if you need to show sourse of `limel-popover` or `limel-tooltip`
+     */
+    @Prop()
+    public showArrow = false;
+
+    /**
+     * ..popover class
```
adrianschmidt, line 89: I don't understand this description… 🤔

src/components/portal/portal.tsx
```diff
@@ -78,6 +79,18 @@ export class Portal {
     @Prop()
     public inheritParentWidth = false;
 
+    /**
+     * Set `showArrow` to `true` if you need to show sourse of `limel-popover` or `limel-tooltip`
+     */
+    @Prop()
+    public showArrow = false;
```
adrianschmidt, line 86: There seems to be some kind of problem with the Basic Example:

Screenshot showing the tooltip with a misaligned arrow

src/components/portal/portal-arrow.scss
```diff
@@ -0,0 +1,94 @@
...
+        var(--contrast-1300)
+    ); // the same color as the tooltip background-color
+    opacity: 0;
+    animation: display-tooltip 0.2s ease forwards;
+
+    animation-delay: 0.2s;
+}
+
+.arrow-style-popover::before {
+    --portal-arrow-color: var(--popover-body-background-color);
```
adrianschmidt, line 84: Popover is a consumer of portal, so portal should not have to know anything about popover. Popover is the one that should make this "conversion" from `--popover-body-background-color` to some other variable prefixed with `--portal-`.

src/components/portal/portal-arrow.scss
```diff
@@ -0,0 +1,94 @@
...
+        &[data-popper-placement^='top'] {
+            margin-bottom: $size-of-arrow !important;
+        }
+        &[data-popper-placement^='bottom'] {
+            margin-top: $size-of-arrow !important;
+        }
+    }
+}
+
+.arrow-style-tooltip::before {
```
adrianschmidt, line 73: Same here, portal should know that tooltip is a consumer. We should inject these styles somehow instead.

src/components/portal/portal.tsx
```diff
@@ -78,6 +79,18 @@ export class Portal {
     @Prop()
     public inheritParentWidth = false;
 
+    /**
+     * Set `showArrow` to `true` if you need to show sourse of `limel-popover` or `limel-tooltip`
```
adrianschmidt, line 83:
```suggestion
     * Set `showArrow` to `true` if you need to show source of `limel-popover` or `limel-tooltip`
```

I think this description can be improved a bit apart from the typo. But I'm a little to tired to come up with something better myself right now. Maybe tomorrow… 😅

src/components/portal/portal.tsx
Kiarokh: Yes as I said, that tooltip attachment bug must be fixed. I think this is due to that bug.
Otherwise, we cannot have arrows for tooltips yet.

src/components/portal/portal-arrow.scss
Kiarokh: Sounds OK for me. Something like `--portal-surface-background-color`

src/components/portal/portal.tsx
adrianschmidt: Hmm… I don't think those are related… if they are, I think we've got something wrong in the tooltip styling anyway, because the `<limel-tooltip>` element is just another element. It's not the element we want to position the tooltip by, so the _position_ of the tooltip may be wrong, but regardless of what the element we position the tooltip by, the tooltip should never look like it does in the screenshot above. If the attachment bug was the only problem, I would expect the tooltip to perhaps look like this:

image

src/components/portal/portal-arrow.scss
adrianschmidt: No, what I mean is that limel-portal shouldn't have to be aware of css variables that are "owned" by other components. If we have a `--popover-z-index`, a `--tooltip-z-index`, and a `--menu-z-index`, limel-portal should not need to be aware of any of those. Instead, limel-popover, limel-tooltip, and limel-menu, respectively, should take care of setting a variable called something like `--portal-z-index`, using the value from their own respective variable. (Also, we shouldn't just cut a corner tell the consumer to use `--portal-z-index` directly, because the consumer shouldn't have to know or care that limel-popover, limel-tooltip, or limel-menu is using limel-portal internally.)

src/components/portal/portal.tsx
adrianschmidt: ⚡

src/components/portal/portal.tsx
Kiarokh: I'm not sure if you are still waiting for my response regarding this one or not.
But a short answer:

what popper does with the "arrow" is that it will place it in relation to the trigger element, and based on all those different `openDirection`s. In case of tooltip, since we have that bug, the trigger element is ignore, and the tooltip is attached to `<limel-tooltip />` itself (not the trigger which is a `2.25rem` by `2.25rem` `<limel-button />`. In the picture above, `<limel-tooltip />` is placed just below the button, has no `height` but gets a `width` of 100%  of its parent container. Therefore the arrow is pointing to the bottom-center of that 100%-wide element (as it's considered as the trigger), while the tooltip has the `openDirection` of `bottom-start` (left), and is attaching itself to the left side of the 100%-wide-trigger.

So without that bug, everything would be fine.

src/components/portal/portal.tsx
adrianschmidt: ⚡

src/components/portal/portal.tsx
adrianschmidt: Ok, nice! I think that bug should be pretty easy to fix actually, now that I (hopefully) understand what's causing it 😊

src/components/portal/portal-arrow.scss
adrianschmidt: This was easy to fix for popover, but…

src/components/portal/portal-arrow.scss
adrianschmidt: … while the z-index is easy for tooltip too, the fact that we use a keyframed animation is making this case pretty complicated for me. I added some properties to allow injecting the necessary styles, but then released that because neither the container nor the arrow are using shadow-dom, the styles "leak" into the rest of the document. So I set about trying to fix that, and I've managed to fix it so it's working for popover and tooltip, but now limel-menu is completely broken. It's not even possible to click on a menu item 😭

(If you ever wonder why I keep nagging about keeping complexity as low as possible, this is one of the reasons; in complex code, changes often have complicated side-effects 😞)

I don't really know where to go with this at the moment. The portal component is very messy already before this arrow addition, and the code is very hard to follow. This PR doesn't really make it _worse_, but it adds code to the mess, which just makes the mess _bigger_. I hate it when people have put in a lot of hard work and then I have to come and say that "no we can't merge this, because blah blah", none which is the fault of the people working on the thing I'm reviewing 😭, but still, I think this needs some more work before it should be merged, work I unfortunately don't have time to do before going on vacation again 😞

Since there isn't really anyone to hand this over to right now, and I'm off on vacation from tomorrow, I'm ok with having a call with someone on one of my vacation days, to explain and hand this over. Just send me a direct message in Slack!

Issue #1785, opened 2022-07-08
Upgrade MDC


Kiarokh:
In Lime Elements, we're using MDC from 13 to 14

# MDC's Change Log, from 13.0.0 to 14.0.0
### Bug Fixes
- [x] **button:** update HCM shim to use the existing focus-ring ([a657abb](https://github.com/material-components/material-components-web/commit/a657abb61a2c7d23c2bd92c3cbd54ed404d0bafe))
> @Kiarokh: Does not seem relevant for us now, as we do our own focus ring, using our own mixins. However, it seems they have some strategies for handling high-contrast mode, which is something we should also consider doing in future.
- [x] * **checkbox:** Add explicit system color for checkmark in HCM. ([8c4da22](https://github.com/material-components/material-components-web/commit/8c4da223a7d35c4ca0a4f27534ecdb13087d0f1b))
- [x] * **checkbox:** move forced-colors theme out of static styles ([bbd1126](https://github.com/material-components/material-components-web/commit/bbd11268f0ea4fd45205a642f1d26a4c4ff34e05))
- [x] * **checkbox:** Update checkbox theme styles mixin to accept css vars ([c14e977](https://github.com/material-components/material-components-web/commit/c14e977ee36c26074b04dea5ce37e2300e9d3735))
> @Kiarokh: it seems this is all about the new `focus-ring` which we already have, and some theming like the previous one. We need to check these separately and have a good theming strategy.

- [x] * **chips:** Fix typography selector in GMDC-Wiz chips theming ([43c7d87](https://github.com/material-components/material-components-web/commit/43c7d87dc0c928d74652ceaebebca51c69b6eaaf))
> @Kiarokh: we are still using `@use '@material/chips/deprecated/mdc-chips';` I think because we need the chips to act as links (`<a />`) in the CRM. So this doesn't concern us right now.

- [x] * **datatable:** Adjust data table last row border-radius to support setting row background-color. ([ba78e87](https://github.com/material-components/material-components-web/commit/ba78e87246af31e1fe7dd03241a7565f2892fd64))
> @Kiarokh: we don't have it

- [ ] * **dialog:** Render dividers in Firefox 94 on Windows HCM ([fae6c65](https://github.com/material-components/material-components-web/commit/fae6c652d7debc3e2875ab9049d806513ddc2421))
[...]

PR #1799, opened 2022-07-26
chore(package): upgrade `material-components-web` from v.13.0.0 to v.…

Kiarokh:
…14.0.0

fix https://github.com/Lundalogik/lime-elements/issues/1785

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/icon-button/icon-button.scss
```diff
@@ -16,10 +16,6 @@ $height-of-limel-button: functions.pxToRem(36);
     display: none;
 }
 
-:host([disabled]) {
-    pointer-events: none;
-}
-
```
Kiarokh, line 18: First I thought we had this because onClick actions could be set on host (`limel-action-button`) not on the `<button` element within the component. But when testing it seems everything works fine.

https://user-images.githubusercontent.com/35954987/181221405-4e89278e-8305-4ef8-8b77-308e238ff349.mov

src/components/select/select.scss
```diff
@@ -3,9 +3,6 @@
 @use '../../style/mixins';
 
 @use '@material/select/styles';
-@use '@material/select/helper-text/mdc-select-helper-text';
-@use '@material/floating-label';
-@use '@material/floating-label/mdc-floating-label';
```
Kiarokh, line 8: First of all, MDC wants us to use another set of styles:
https://github.com/material-components/material-components-web/tree/69fd619d0cf28dc6588650282bc825858b104a74/packages/mdc-select
which are:
```scss
@use "@material/list/mdc-list";
@use "@material/menu-surface/mdc-menu-surface";
@use "@material/menu/mdc-menu";
@use "@material/select/styles";
```
But neither those we used to have, nor the ones MDC suggests
have any effect. I guess this is due to us having a custom implementation
of their `select` component. So I decided to remove them.

adrianschmidt:
Since you argued we shouldn't update MDC anymore, I'll close this, @Kiarokh 👍

Kiarokh:
it's a bold claim, but I think it'll help us in the long run 😅 (or maybe we regret it :P )

PR #1802, opened 2022-07-27
fix(select): set `aria-required` and `aria-disabled` on trigger eleme…

Kiarokh:
…nts to improve accessibility

Both for default and native trigger elements.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/select/select.template.tsx
+++ b/src/components/select/select.template.tsx
@@ -91,6 +91,8 @@ const SelectValue: FunctionalComponent<
             aria-haspopup="listbox"
             aria-expanded="false"
             aria-labelledby="s-label s-selected-text"
+            aria-required={props.required}
+            aria-disabled={props.disabled}
         >
             <span id="s-label" class={labelClassList}>
                 {props.label}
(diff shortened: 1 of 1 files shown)


src/components/select/select.template.tsx
```diff
@@ -91,6 +91,8 @@ const SelectValue: FunctionalComponent<
             aria-haspopup="listbox"
             aria-expanded="false"
             aria-labelledby="s-label s-selected-text"
+            aria-required={props.required}
+            aria-disabled={props.disabled}
```
Kiarokh, line 95: This is according to MDC guidelines: https://github.com/material-components/material-components-web/tree/69fd619d0cf28dc6588650282bc825858b104a74/packages/mdc-select

Befkadu1 (approved):

PR #1808, opened 2022-08-03
feat: enable adding placeholder text for input-field & date-picker

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/date-picker/date-picker.tsx
+++ b/src/components/date-picker/date-picker.tsx
@@ -86,6 +86,12 @@ export class DatePicker {
     public label: string;
 
     /**
+     * The placeholder text shown inside the input field, when the field is focused and empty
+     */
+    @Prop({ reflect: true })
+    public placeholder: string;
+
+    /**
      * Optional helper text to display below the input field when it has focus
      */
     @Prop({ reflect: true })
@@ -211,6 +217,7 @@ export class DatePicker {
                 readonly={this.readonly}
                 invalid={this.invalid}
                 label={this.label}
+                placeholder={this.placeholder}
                 helperText={this.helperText}
                 required={this.required}
                 value={this.formattedValue}
--- a/src/components/input-field/examples/input-field-placeholder.tsx
+++ b/src/components/input-field/examples/input-field-placeholder.tsx
@@ -1,0 +1,55 @@
+import { Component, h, State } from '@stencil/core';
+
+/**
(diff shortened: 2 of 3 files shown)


src/components/date-picker/date-picker.tsx
```diff
@@ -85,6 +85,12 @@ export class DatePicker {
     @Prop({ reflect: true })
     public label: string;
 
+    /**
+     * The placeholder text shown inside the input field, when the field is focused and empty
```
Kiarokh, line 89: Not sure about this solution here for limel-chip-set

https://github.com/Lundalogik/lime-elements/blob/0611bf4a80c989795f3c4620c2a04eabafe7e1d1/src/components/chip-set/chip-set.tsx#L397

I think we should have called `searchLabel`, `placeholder` instead. Should we deprecate it and do so? Or fall it back to placeholder?

Kiarokh: > normally a placeholder would always be visible as long as the field is empty,

True that! _However_, in material design, that's not the case. Labels are visible and inside the filed, when the field is empty. Therefore, even if you specify a placeholder and **no label,** the placeholder won't be shown, unless you click on the input field.

So they are working exactly the same.

Kiarokh: >  even if you specify a placeholder and no label, the placeholder won't be shown, unless you click on the input field.

💩 I just realized that due to this, I can't use `placeholder` as shown in my prototype here: https://github.com/Lundalogik/crm-feature/issues/2916

So to show "calculated date and time", I have to use `label` instead of `placeholder` when the filed is empty and not focused. And when it's focused, use the placeholder, and switch the label to something else :/  

Blä…

This is a good feature to have nevertheless

adrianschmidt: Then I guess it would make sense to deprecate `searchLabel` and add `placeholder` 👍

Kiarokh: OK. Let's do it in another PR

Befkadu1 (approved):

PR #1811, opened 2022-08-09
fix(date-picker): remove warning "trailingIcon" changed during rendering

Befkadu1:
This warning occurs when clearing the value from the date picker. When clearing the value from the
date picker the trailingIcon becomes undefined and then changes to null while rendering. See the
attached video for more details.

https://user-images.githubusercontent.com/22968497/183623735-77115138-1843-44ea-9566-883afba0e24d.mov

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/date-picker/date-picker.tsx
+++ b/src/components/date-picker/date-picker.tsx
@@ -185,6 +185,7 @@ export class DatePicker {
     public render() {
         const inputProps: any = {
             onAction: this.clearValue,
+            trailingIcon: null,
         };
 
         if (this.value && !this.readonly) {


Befkadu1:
> Since the problem is in limel-select

I have no idea if the problem is in limel-select

adrianschmidt:
> This fixes the problem but... Do you know why or where the tailingIcon property is changed from undefined to null in limel-select? Since the problem is in limel-select, shouldn't it be fixed there? or maybe I don't really understand it 🤔

I think I know why, at least in basic terms.

Without this change, the render-loop when clicking the clear-button looks something like this:

1. Detect click on clear-button
2. Clear the value, and don't set any value for `trailingIcon`
3. Render updates
4. I can't see anywhere in the code where we explicitly change the value of `trailingIcon` from `undefined` to `null`, but it happens somewhere. It maybe Stencil doing it somehow, or I might just not have found where we do it. But it happens approximately at this point.
5. `trailingIcon` is a `@Prop`, so when the value is changed, a new render-loop is kicked off.
6. Stencil detects that a new render-loop was triggered from inside a running render-loop, and warns that this might cause infinite loops.

With this change, the render-loop instead looks something like this:

1. Detect click on clear-button
2. Clear the value, and set `trailingIcon` to `null`
3. Render updates
4. Done

src/components/date-picker/date-picker.tsx
```diff
@@ -185,6 +185,7 @@ export class DatePicker {
     public render() {
         const inputProps: any = {
             onAction: this.clearValue,
+            trailingIcon: null,
```
adrianschmidt, line 188: Hmm… I can't reproduce the bug on `next`… 🤔

I wonder if it has been fixed some other way since this PR was opened?

adrianschmidt: Huh!? Seems like we haven't had a single release of Lime Elements for 16 days 😮 😄

Befkadu1:
> 

Yes. It is difficult to know where the value of `trailingIcon` from `undefined` to `null`. How do we go forward from here?

PR #1813, opened 2022-08-22
feat(dock-button): emit event when dock menu is opened

Befkadu1:
fix Lundalogik/crm-feature#2931

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dock/dock-button/dock-button.spec.tsx
+++ b/src/components/dock/dock-button/dock-button.spec.tsx
@@ -1,0 +1,33 @@
+import { h } from '@stencil/core';
+import { newSpecPage, SpecPage } from '@stencil/core/testing';
+import { DockItem } from '../dock.types';
+import { DockButton } from './dock-button';
+
+let page: SpecPage;
+let item: DockItem;
+
+beforeEach(async () => {
+    item = {
+        id: 'tables',
+        label: 'Tables',
+        icon: 'insert_table',
+        dockMenu: { componentName: 'my-custom-menu' },
+    };
+    page = await newSpecPage({
+        components: [DockButton],
+        template: () => <limel-dock-button item={item} />,
+    });
+    await page.waitForChanges();
+});
+
+test('emits an event when dock menu is opened', async () => {
+    const eventSpy = jest.fn();
+    page.body.addEventListener('menuOpen', eventSpy);
+
+    const button = page.root.querySelector('.button') as HTMLButtonElement;
+    button.click();
+    await page.waitForChanges();
+
+    expect(eventSpy).toHaveBeenCalled();
+    expect(eventSpy.mock.calls[0][0].detail).toEqual(item);
+});
--- a/src/components/dock/dock-button/dock-button.tsx
+++ b/src/components/dock/dock-button/dock-button.tsx
@@ -38,6 +38,12 @@ export class DockButton {
     private itemSelected: EventEmitter<DockItem>;
 
     /**
+     * Fired when a dock menu is opened.
+     */
+    @Event()
+    private menuOpen: EventEmitter<DockItem>;
+
+    /**
      * Indicated whether the popover that renders a component is open.
      */
     @State()
@@ -107,6 +113,7 @@ export class DockButton {
     private openPopover = (event: MouseEvent) => {
         event.stopPropagation();
         this.isOpen = true;
+        this.menuOpen.emit(this.item);
     };
 
     private onPopoverClose = () => {


Befkadu1:
> You reminded me of this last week @Befkadu1 How do I know if it works? 😄

come on 😆 . You are right. We can add the test in all components may be in the cool down period?

jgroth:
Adding a test for this should not take more than a few minutes

src/components/dock/dock-button/dock-button.tsx
```diff
@@ -107,6 +107,7 @@ export class DockButton {
     private openPopover = (event: MouseEvent) => {
         event.stopPropagation();
         this.isOpen = true;
+        this.itemSelected.emit(this.item);
```
vicgeralds, line 110: This seems to work, in the sense that it fixes https://github.com/Lundalogik/crm-feature/issues/2931, because we happen to have implemented our `itemSelected` event handler in a certain way.

But I don't think it's the right event to emit here. The item is not selected (we don't expect `this.item.selected` to be set to `true`).

vicgeralds (changes requested):
I think we need to introduce a new event for this

src/components/dock/dock-button/dock-button.tsx
```diff
@@ -7,7 +7,7 @@ import { createRandomString } from '../../../util/random-string';
  */
 @Component({
     tag: 'limel-dock-button',
-    shadow: false,
+    shadow: true,
```
adrianschmidt, line 10: It's good that this change is in a separate commit. But the other two changes in this commit should be in a `fixup! fix(dock-button): emit event when dock item is clicked for custom components` commit instead.

vicgeralds: This change also breaks styling in `dock-button.scss` where we use the selector `.dock-item:not(.selected) .button:not(.disabled)`.

adrianschmidt: Ah, Victor's comment made me look at this a little closer, and if I recall correctly, I think I was involved in creating this component (not entirely sure, it was before my vacation), and I think we made it non-shadowed on purpose, precisely because we needed that styling to work, and the button is a "sub-component" of `limel-dock`, only intended for use inside the dock, and so it's still isolated from the rest of the consuming application by `limel-dock`'s shadow DOM.

@Befkadu1 Was there a specific reason for making this component shadowed here, or was it just an effort to align it with our usual conventions? If the latter, while definitely a laudable effort, I think we can keep this component non-shadowed.

Befkadu1: I thought we disabled the encapsulation by mistake. If it was made on purpose, I would revert back the change. I was encapsulation. I was trying to shield the component’s style.

adrianschmidt: Cool! As I said, that's a good intention! 👍 ❤️ But yes, in this specific case, it was done on purpose 😊

Befkadu1: Ok then if there is no other comments, I am squashing the commits.

src/components/dock/dock-button/dock-button.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+        label: 'Tables',
+        icon: 'insert_table',
+        dockMenu: { componentName: 'my-custom-menu' },
+    };
+    page = await newSpecPage({
+        components: [DockButton],
+        template: () => <limel-dock-button item={item} />,
+    });
+    eventSpy = jest.fn();
+    page.body.addEventListener('dockMenuOpened', eventSpy);
```
vicgeralds, line 22: I would move this into the test. Creating a globally scoped `eventSpy` makes it less clear what event is spied on, and other tests (when added) might not need to spy on any event.

src/components/dock/dock-button/dock-button.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+    await page.waitForChanges();
+});
+
+test('emits an event with the menu item when the menu item in the dock is clicked', async () => {
+    const button = (await page.root.shadowRoot.querySelector(
```
vicgeralds, line 27:
```suggestion
    const button = (await page.root.querySelector(
```

Befkadu1: Yea that was implemented when the shadow DOM was true.

src/components/dock/dock-button/dock-button.tsx
```diff
@@ -37,6 +37,12 @@ export class DockButton {
     @Event()
     private itemSelected: EventEmitter<DockItem>;
 
+    /**
+     * Fired when a dock menu has been clicked from the dock.
+     */
+    @Event()
+    private dockMenuOpened: EventEmitter<DockItem>;
```
vicgeralds, line 44: `dockMenuOpened` or `menuOpened`?

I'm not sure why we named it `item.dockMenu` and not just `item.menu`. So `dockMenuOpened` is consistent with that, but feels inconsistent alongside `itemSelected` (why not `dockItemSelected`). 

Not sure what to suggest here, that's why I've stayed away from discussing the event name before 🙂

Befkadu1:
> feat(dock-button): emit event when dock menu is opened

You know sometimes you can do a fixup if it is a small thing, that would make the process a bit faster. I will do as you suggested though!

src/components/dock/dock-button/dock-button.spec.tsx
```diff
@@ -0,0 +1,34 @@
+import { h } from '@stencil/core';
+import { newSpecPage, SpecPage } from '@stencil/core/testing';
+import { DockItem } from '../dock.types';
+import { DockButton } from './dock-button';
+
+let page: SpecPage;
+let item: DockItem;
+let eventSpy: jest.Mock;
```
vicgeralds, line 8: No need to have this as a global (that was implicit in my suggestion to move it into the test)

vicgeralds:
> You know sometimes you can do a fixup if it is a small thing, that would make the process a bit faster. I will do as you suggested though!

Yes, but reviewing is not about getting things merged fast. It's also about spreading knowledge. I wanted to explain why the commit message was unclear.

src/components/dock/dock-button/dock-button.tsx
Befkadu1: I chose **_dockMenuOpened_** to make it consistent with _item.**dockMenu**._  Don't you think the event **_itemSelected_** is consistent with _item.**selected**_ ? 

But we can have another PR to refactor if we need to change the properties in the Interface

vicgeralds: Yes, I agree that's consistent in a way, but I don't understand why the `dock` prefix is used here and not on `itemSelected`.

Renaming the event is not refactoring, it would be a breaking change. That's why I don't want to rush this.

> Don't you think the event **_itemSelected_** is consistent with _item.**selected**_

Yes, but it doesn't help if you can't apply the same rule to name the new event that's reflected in `item.dockMenu.menuOpen`.

I can see naming the event `menuOpen` is also an option.

```jsx
<limel-dock
  dockItems={this.dockItems}
  onItemSelected={this.handleSelected} 
  onMenuOpen={this.handleMenuOpen}
/>
```

What do you think @adrianschmidt ?

vicgeralds: We also have the `close` event that we could have named `menuClose` or `dockMenuClosed` I guess 🤔

adrianschmidt: Yeah, I'm with you on this being inconsistent. I feel like that was maybe discussed on the [original PR](https://github.com/Lundalogik/lime-elements/pull/1792), but there was a lot of discussion there, and I tired of searching before finding anything. I'm not sure if it was ever discussed and there was a good reason or if it was just something that was missed.

> Renaming the event is not refactoring, it would be a breaking change. That's why I don't want to rush this.

Yes, and I agree with this sentiment, but I think limel-dock is still marked as private, so the docs for it are hidden, and we are free to break it as we wish, until we "publish it" 👍 Whether you use that to justify merging this PR without deciding on the naming first, or use it to justify changing the names of existing properties and events, that's up to you 😊

adrianschmidt: I think `menuOpen` is a good name for the event, but I'm not deeply invested either way 😅

Befkadu1: The event name `menuOpen` can be a good name actually. It actually go with `item.dockMenu.menuOpen`. I agree with this.

vicgeralds (approved):
Looks good!

Sorry if I was over-reviewing a bit and not taking into account that limel-dock is still marked as private 🙂

PR #1818, opened 2022-08-29
feat(code-editor): lint and fold json

anderssonjohan:
Adds linting and folding support for JSON! 🎉 

(needs tooltip CSS from @Kiarokh or @LucyChyzhova !)

Screen Shot 2022-08-29 at 17 00 19

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "html-escaper": "^3.0.3",
     "jest": "^26.6.3",
     "jest-cli": "^28.0.2",
+    "jsonlint-mod": "^1.7.6",
     "jsx-dom": "^7.0.4",
     "kompendium": "^0.11.3",
     "lodash-es": "^4.17.21",
--- a/src/components/code-editor/code-editor.scss
+++ b/src/components/code-editor/code-editor.scss
@@ -1,4 +1,8 @@
+@use '../../style/mixins';
+
 @import '../../../node_modules/codemirror/lib/codemirror.css';
+@import '../../../node_modules/codemirror/addon/lint/lint.css';
+@import '../../../node_modules/codemirror/addon/fold/foldgutter.css';
 
 /**
  * @prop --code-editor-max-height: Defines how tall the code editor can get before content becomes scrollable, defaults to `10rem`.
@@ -215,5 +219,49 @@
                 right: 0;
             }
         }
+    }
+    .CodeMirror-lint-marker-error,
+    .CodeMirror-lint-message-error {
+        background: {
+            image: url("data:image/svg+xml; utf8, <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 32 32'><defs/><path fill='rgb(255,255,255)' d='M7.219 5.781L5.78 7.22 14.563 16 5.78 24.781 7.22 26.22 16 17.437l8.781 8.782 1.438-1.438L17.437 16l8.782-8.781L24.78 5.78 16 14.563z'/></svg>");
+            color: rgb(var(--color-red-default));
+            size: 0.75rem;
+        }
+        border-radius: 50%;
+    }
+
+    .CodeMirror-foldmarker {
+        position: relative;
+        @include mixins.is-elevated-clickable();
+        color: transparent;
+        text-shadow: none;
+
+        padding: 0 0.5rem;
+        border-radius: 1rem;
+
+        background: {
+            image: url("data:image/svg+xml; utf8, <svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 32 32' xml:space='preserve'><circle cx='16' cy='16' r='2'/><circle cx='26' cy='16' r='2'/><circle cx='6' cy='16' r='2'/></svg>");
+            size: contain;
+            repeat: no-repeat;
+            position: center;
+        }
+
+        .is-dark-mode & {
+            background-color: rgb(var(--contrast-1200));
+        }
+    }
+
+    [class^='CodeMirror-foldgutter'] {
+        color: var(--code-editor-neutral-text-color);
+        transition: opacity 0.2s ease;
+        opacity: 0.4;
+
+        &:hover {
+            opacity: 1;
+        }
+    }
+
+    .CodeMirror-foldgutter-folded {
+        opacity: 0.7;
     }
 }
--- a/src/components/code-editor/code-editor.tsx
+++ b/src/components/code-editor/code-editor.tsx
@@ -13,6 +13,11 @@ import CodeMirror from 'codemirror';
 import 'codemirror/mode/javascript/javascript';
 import 'codemirror/addon/selection/active-line';
 import 'codemirror/addon/edit/matchbrackets';
+import 'codemirror/addon/lint/lint';
+import 'codemirror/addon/lint/json-lint';
+import 'codemirror/addon/fold/foldgutter';
+import 'codemirror/addon/fold/brace-fold';
+import jslint from 'jsonlint-mod';
 
 /**
  * Currently this component support syntax highlighting for `javascript`,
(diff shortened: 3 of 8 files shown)


mjlime (approved):

src/components/code-editor/code-editor.tsx
```diff
@@ -168,13 +188,32 @@ export class CodeEditor {
                 name: 'application/json',
                 json: true,
             };
+            if (this.lint) {
```
mjlime, line 191: Should we limit linting for use when `language` is `json` since I don't know what happens if we list when language is something else?

package.json
```diff
@@ -72,6 +72,7 @@
     "html-escaper": "^3.0.3",
     "jest": "^26.6.3",
     "jest-cli": "^28.0.2",
+    "jsonlint-mod": "^1.7.6",
```
dorisenthecat, line 75: No commits on this package since 2020, its a branch of jsonlint which has has commits from 2021 but that is only for updating an abandoned dependency (nomnom).

src/components/code-editor/code-editor.tsx
anderssonjohan: This does not limit anything but adds the necessary linter when using json as the language.
If we add other linters I guess we add the lint-markers gutter in that case?
I'm not sure if I understand your comment correctly.

I added the lint prop as an opt-in to avoid breaking any current use. To me it seems reasonable to always have linting enabled for json content, but I guess it's better to let the consumer be in control of that behavior.

Note: This PR only add support for json linting

mjlime: What I meant is that if you enable linting it will be enabled regardless of what `this.language` is. And as far as I can understand the linter you add in this PR does only support `json`? So what happens if you combine `language = typescript` with `lint = true`?

anderssonjohan: I see what you mean but nothing will happen. Even though `lint` is `true` in the editor options there will be no support since no lint gutter is added + the codemirror lint addon for the language is not imported. In addition to that, a linter implementation is also needed. All this was added for json but not for other languages. 
If we would import the lint addon for typescript it would also log an error in the console if we don't provide a linter in the window object (required by the lint addon).

dorisenthecat:
Since we have gotten questions from users that didn't know that double vs single quotes matters in json I tried this

image

This linter is technically correct and shows a parser error, for none developers it highlights the line before the error and says something incomprehensible? It took me several minutes just to understand `--------------------^` wasn't a line break but an arrow 😄 But getting this in a screenshot from a user asking for help is much much better than screenshot of a wall of json.

Compared to vscode

image

dorisenthecat (dismissed):

adrianschmidt:
Prettier is able to autofix single-quotes in json:

Screenshot from prettier.io/playground showing that Prettier replaces single-quotes in json with double-quotes.

[Link to Prettier playground for json](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEwA6ICGAbbB5KDJAAgwGUBLKAc2zgwBpSQAZK+5ZuAD0wFsADnUZpCIKBDIDMYDkhgAnAK5wGo4swCOSiPCLF0UDRoDkAZyq04ARR3wTJEz35C4JtUeNfv3jFQAmCPD++ooqHhoAvhHMAgAW0HAAckp8AEZwCmb6ANrqxuggMACeAnLMCXwcMV4YUKkZCvoYAEwAjC3EAKw9ALQdAMwALBjR+RqFJWXNIBAAZnMUsiKetUUKmBTYljOQfHyYZsRpxcQyYBSBsKMxALoekSAMswIwFNBmyKCYCgoQAO4ABR+CE+KBw-0wxU+zzSGzAAGs4DAyPw4GwoHBkHMcGZVCA4TIkSjpBcaMgwvi4Ok4P5Av4WJgaEpMNQ4AAxCAKA4wN7k8FKGAQJ4gOIwPjYADqcQo8DMpLgZFBsooADdZcVkOAzDCQFQ8QoYICNtQDtjcfiAFZmbiUGh0Wy6LFIHHYPHPaRZTJa63QEUCBRUGCSy4wOLIAAcAAYPX88ZKNgItQG4AbVVjntoncaIAIwVgzL1MbTaSKFHBtBRy8bWWaXRbnni+BQKcp8RZ7TY7M7Xe6ipg0iH-GHkC1nopNtsaABhCD7TBa1NdEVKPEAFQHYN7+NVKgAklAriiwIHXgBBQ9kEp0c1uuCRSJAA)

Is there a Prettier-plugin for the code editor? Or could auto-fix be enabled for the linter that's added here?

anderssonjohan:
> Prettier is able to autofix single-quotes in json:
> 
> Screenshot from prettier.io/playground showing that Prettier replaces single-quotes in json with double-quotes.
> 
> [Link to Prettier playground for json](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEwA6ICGAbbB5KDJAAgwGUBLKAc2zgwBpSQAZK+5ZuAD0wFsADnUZpCIKBDIDMYDkhgAnAK5wGo4swCOSiPCLF0UDRoDkAZyq04ARR3wTJEz35C4JtUeNfv3jFQAmCPD++ooqHhoAvhHMAgAW0HAAckp8AEZwCmb6ANrqxuggMACeAnLMCXwcMV4YUKkZCvoYAEwAjC3EAKw9ALQdAMwALBjR+RqFJWXNIBAAZnMUsiKetUUKmBTYljOQfHyYZsRpxcQyYBSBsKMxALoekSAMswIwFNBmyKCYCgoQAO4ABR+CE+KBw-0wxU+zzSGzAAGs4DAyPw4GwoHBkHMcGZVCA4TIkSjpBcaMgwvi4Ok4P5Av4WJgaEpMNQ4AAxCAKA4wN7k8FKGAQJ4gOIwPjYADqcQo8DMpLgZFBsooADdZcVkOAzDCQFQ8QoYICNtQDtjcfiAFZmbiUGh0Wy6LFIHHYPHPaRZTJa63QEUCBRUGCSy4wOLIAAcAAYPX88ZKNgItQG4AbVVjntoncaIAIwVgzL1MbTaSKFHBtBRy8bWWaXRbnni+BQKcp8RZ7TY7M7Xe6ipg0iH-GHkC1nopNtsaABhCD7TBa1NdEVKPEAFQHYN7+NVKgAklAriiwIHXgBBQ9kEp0c1uuCRSJAA)
> 
> Is there a Prettier-plugin for the code editor? Or could auto-fix be enabled for the linter that's added here?

There's a formatting addon to codemirror but I haven't yet found a codemirror + prettier example but it should be doable: https://discuss.codemirror.net/t/formatting-the-code/2442/5

The question is how much all this will add to the lime-elements dist size; Will it be ok or too much?

mjlime (approved):

src/components/code-editor/code-editor.scss
```diff
@@ -227,4 +229,21 @@
         }
         border-radius: 50%;
     }
+
+    .CodeMirror-foldmarker {
```
anderssonjohan, line 233: Screen Shot 2022-08-31 at 15 56 08

This is barely visible on my machine. (I only get the dark mode)

Also, can we use something else for the arrows in the gutter instead of the default (ugly) ones?
I found that they are given by this style:

```css
.CodeMirror-foldgutter-open::after {
  content: "▾";
}
```

dorisenthecat (approved):

dorisenthecat:
This so cool @anderssonjohan 🥳

anderssonjohan:
@adrianschmidt @dorisenthecat The error caught by JSON.parse is actually better than the one from json-lint :/
See the message below the code editor in this screenshot:

Screen Shot 2022-09-02 at 12 53 04

adrianschmidt:
I don't want _any_ error message, I want it to be fixed automatically 😄

But yeah, in lieu of that, use the one that gives better error messages, of course! 👍

PR #1819, opened 2022-08-30
2941 table layouts

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/2941

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/examples/invoices.ts
+++ b/src/components/table/examples/invoices.ts
@@ -1,0 +1,66 @@
+export interface Invoice {
+    id: number;
+    invoiceNumber: number;
+    invoiceDate: string;
+    referencePerson: string;
+    amount: number;
+}
+
+export const invoices: Invoice[] = [
+    {
+        id: 1,
+        invoiceNumber: 15467,
+        invoiceDate: '02/10/2020',
+        referencePerson: 'Fauji Sheladia',
+        amount: 21.328,
+    },
+    {
+        id: 2,
+        invoiceNumber: 15433,
+        invoiceDate: '03/06/2020',
+        referencePerson: 'Mäkelä Jehkinen',
+        amount: 9.169,
+    },
+    {
+        id: 3,
+        invoiceNumber: 34326,
+        invoiceDate: '30/03/2021',
+        referencePerson: 'Helrover Shgrover',
+        amount: 1.295,
+    },
+    {
+        id: 4,
+        invoiceNumber: 12357,
+        invoiceDate: '14/09/2020',
+        referencePerson: 'Goliham Gigantlle',
+        amount: 11.468,
+    },
+    {
+        id: 5,
+        invoiceNumber: 12467,
+        invoiceDate: '07/08/2020',
+        referencePerson: 'Eloise Sullivan',
+        amount: 2.583,
+    },
+    {
+        id: 6,
+        invoiceNumber: 12653,
+        invoiceDate: '01/01/2021',
+        referencePerson: 'Gladys Fyffe',
+        amount: 5.201,
+    },
+    {
+        id: 7,
+        invoiceNumber: 12655,
+        invoiceDate: '01/01/2021',
+        referencePerson: 'Cunninghamarms',
+        amount: 5.181,
+    },
+    {
+        id: 8,
+        invoiceNumber: 12622,
+        invoiceDate: '23/12/2020',
+        referencePerson: 'Agalle Cunningpower',
+        amount: 18.257,
+    },
+];
--- a/src/components/table/examples/table-layout-default.tsx
+++ b/src/components/table/examples/table-layout-default.tsx
@@ -1,0 +1,73 @@
+import { Component, h, State } from '@stencil/core';
+import { Column } from '../table.types';
+import { invoices, Invoice } from './invoices';
+
+/**
+ * Layout
+ * Columns and their content can be decisive factors in how a table is
+ * preferred to rendered in the user interface. To set your preferred
+ * rendering, choose one of the available `layout` properties.
+ *
+ *
+ * ````tsx
+ * layout="default"
+ * ````
+ * The default layout resizes the table's columns,
+ * in a way that each column becomes as wide as the data it holds.
+ *
+ * :::important
+ * Note that be default, table columns have a maximum width of `40rem`.
+ * This means, they can never grow wider than that, unless you specify
+ * another size using the `--table-max-column-width` CSS variable.
+ *
+ * This applies to all other layouts presented further down as well!
+ * :::
+ *
+ * If there is additional space available on the right side of the last column,
+ * rows will stretch to fill the space and look visually as wide as the table.
+ *
+ * :::note
+ * While scrolling, new rows get lazy-loaded. Since the new data may have wider
+ * length, it might affect rendering of the layout in real-time.
+ * This means columns can get resized while user is scrolling down.
+ * :::
+ *
+ * :::tip
+ * It is also possible to affect internal layout of each column, by specifying
+ * `horizontalAlign` on the column headers, to `left` (default), `center`,
+ * or `right`. This basically defines the text-alignment for all the cells in that column.
+ * :::
+ *
+ * @link invoices.ts
+ */
+@Component({
+    tag: 'limel-example-table-layout-default',
+    styleUrl: 'table.scss',
+    shadow: true,
+})
+export class TableExampleLayoutDefault {
+    @State()
+    private tableData: Invoice[] = invoices;
+
+    @State()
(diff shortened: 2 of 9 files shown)


src/components/table/examples/table-layout-fitDataStretch.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ *
+ * ````tsx
+ * layout="fitDataStretch"
+ * ````
+ *
+ * Works just like `fitDataFill`, but unlike the default layout
+ * which resulted in having an empty last column, in this layout
+ * the last existing column will stretch out to fill up the remaining table width.
```
LucyChyzhova, line 13: In the full-screen view, the last column has only the header:
Screenshot 2022-09-06 at 06 40 01

src/components/table/examples/table-layout-fitColumns.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { Column } from '../table.types';
+import { invoices, Invoice } from './invoices';
+
+/**
+ *
+ * ````tsx
+ * layout="fitColumns"
+ * ````
+ *
+ * With this layout, the table stretches columns so that all
```
LucyChyzhova, line 11: A similar (to layout="fitDataStretch") problem appears if you resize the columns  

https://user-images.githubusercontent.com/50618208/188550499-c1d839ef-08a2-48c2-809f-5f7bcad4118d.mov

src/components/table/examples/table-layout-fitData.tsx
```diff
@@ -0,0 +1,57 @@
+import { Component, h, State } from '@stencil/core';
+import { Column } from '../table.types';
+import { invoices, Invoice } from './invoices';
+
+/**
+ *
+ * ````tsx
+ * layout="fitData"
```
LucyChyzhova, line 8: Examples `table-layout-fit-data`  and `table-low-density` have quite similar UI if I'm not mistaken the only difference is a hover effect - for the `table-layout-fit-data` it's gray hover and for the `table-low-density` white hover.  I would suggest renaming the `table-low-density` to  `table-layout-fit-data-interactive-rows` to reduce confusions 



https://user-images.githubusercontent.com/50618208/188556682-badac3a2-473f-4bae-abfc-1f8d27bb3bc6.mov

Kiarokh: No. This this is probably slightly confusing documentations. In the last example, both the class and the layout are applied.

image

src/components/table/examples/table-layout-fitColumns.tsx
Kiarokh: Good finding! I didn't notice this. But this is what we get from Tabulator and cannot do anything about it.

I guess we can make resizability as a customizable prop

LucyChyzhova: The last column works fine in all other cases except these two `limel-example-table-layout-fit-data-stretch`  `limel-example-table-layout-fit-columns`
I like the column behavior in `limel-example-table-layout-fit-data` 👇
https://user-images.githubusercontent.com/50618208/188583001-7884a90d-4fab-4e3a-b077-e3f7be6d635e.mov

Would it be possible to have the same approach for that two?

src/components/table/examples/table-layout-fitDataStretch.tsx
LucyChyzhova, line 13: > The last existing column will stretch out to fill up the remaining table width. 

- do you mean the last filled column? or this empty column should not be here?

src/components/table/examples/table-layout-fitColumns.tsx
adrianschmidt: 1. While column resizability being optional may perhaps be a feature we want anyway, I don't think we should consider something like that to be a fix for a bug. Taking user adjustability away just to force something to line up—not just before the user changes anything, but always—is a bad design choice, in my opinion.
2. Unless I'm missing something here, I don't think it's a bug that the columns don't fill the whole width _after_ the user changes the width of one or more columns. Quite the opposite, I find it very annoying when a set of columns are locked to a certain width, so I can't change the width of one column without some other column also changing to "cancel out" the width difference 😤
It should be enough that the columns spread to take the full width by default. If the user wants to change that, let them change it. Power to the people! ✊

src/components/table/examples/table-layout-fitDataFill.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Layout
+ * Columns and their content can be decisive factors in how a table is
+ * preferred to rendered in the user interface. To set your preferred
+ * rendering, choose one of the available `layout` properties.
+ *
+ *
+ * ````tsx
+ * layout="fitDataFill"
```
LucyChyzhova, line 13: > * layout="fitDataFill"

I found the naming of the layouts not that self-explanatory 
What do you think about these suggestions: 
Screenshot 2022-09-06 at 11 02 32
Screenshot 2022-09-06 at 11 02 48
Screenshot 2022-09-06 at 11 03 13

Screenshot 2022-09-06 at 11 03 23

src/components/table/examples/table-layout-fitData.tsx
LucyChyzhova: I think the problem is that not only in this example the rows are interactive. Just the interaction is different. 

I like that you added this `note`
Screenshot 2022-09-06 at 11 09 47

src/components/table/examples/table-layout-fitColumns.tsx
Kiarokh: I agree with both of you and your preferences. As I said, the way these columns are rendered is coming from Tabulator. It's not my choice. Otherwise in most cases I'd also prefer to visually stretch out the row to the end of the table (just like the default). So if you resize the table columns manually, you don't get empty space on the right side.

Anyhow, this is what they give us. We're now only enabling the features that they have for our consumers.

## resizability of columns
If I don't get this wrong, this is already an optional feature available on individual columns: http://tabulator.info/docs/5.3/layout#resize-column we're just not documenting it here in the Docs, and don't have examples for it. I haven't tested it though, but I assume it works.

However, the consumer/designer can and should have this choice. Sometimes, for instance when there are very few columns and one of these layouts has been chosen. In some scenarios, the design choice might be no letting the end-user manipulate the column widths. Specially if the layout is "fitting the intended need". However, the consumer can use this feature in the wrong way, like any other feature. And we can't do anything about it.

src/components/table/examples/table-layout-fitData.tsx
Kiarokh: it's true. that's also bugging me. how non-interactive rows are today highlighted on hover and also selectable (highlighted on click) by default could probably be tweaked a little bit. we can talk about it one day. that also comes out of the box.

highlighting on hover actually has usability benefits, so I don't want to remove it completely. Because it helps the users focus and read on the content of one row in a messy table, and don't mix rows up

src/components/table/examples/table-layout-fitDataFill.tsx
Kiarokh: Yes, naming that Tabulator has chosen are a bit weird. I think they try to explain the table behavior, while we try to explain the column behavior. :D

The question is whether we can we do so or not: 
if user choses `responsiveColumns` we convert it to `fitColumns` and pass it to the table? 
if the choose `lowDensity` we convert it to `fitData` and pass it to the table.

I think it should be easy. But I don't know how to do it.

Kiarokh: Yes, naming that Tabulator has chosen are a bit weird. I think they try to explain the table behavior, while we try to explain the column behavior. :D

The question is whether we can we do so or not: 
if user choses `responsiveColumns` we convert it to `fitColumns` and pass it to the table? 
if the choose `lowDensity` we convert it to `fitData` and pass it to the table.

I think it should be easy. But I don't know how to do it. Do you know how to do it?

src/components/table/examples/table-layout-fitDataStretch.tsx
Kiarokh: yes, last existing = last filled, or last column with content
consumers send an array of columns to tables, so the last one of them.

LucyChyzhova (approved):
💪🥳

PR #1820, opened 2022-08-31
2944 keyboard navigation

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/2944#issuecomment-1231614412
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dock/dock-button/dock-button.tsx
+++ b/src/components/dock/dock-button/dock-button.tsx
@@ -1,4 +1,12 @@
-import { Component, Event, EventEmitter, h, Prop, State } from '@stencil/core';
+import {
+    Component,
+    Event,
+    EventEmitter,
+    h,
+    Prop,
+    State,
+    Watch,
+} from '@stencil/core';
 import { DockItem } from '../dock.types';
 import { createRandomString } from '../../../util/random-string';
 
@@ -56,6 +64,7 @@ export class DockButton {
     public close: EventEmitter<void>;
 
     private tooltipId: string;
+    private customComponentElement: HTMLElement;
 
     constructor() {
         this.tooltipId = createRandomString();
@@ -69,6 +78,18 @@ export class DockButton {
         return this.renderButton(this.handleClick);
     }
 
+    @Watch('isOpen')
+    protected openWatcher() {
+        if (!this.isOpen) {
+            return;
+        }
+
+        const observer = new IntersectionObserver(
+            this.focusCustomComponentElement
+        );
+        observer.observe(this.customComponentElement);
+    }
+
     private renderPopover() {
         const CustomComponent = this.item?.dockMenu.componentName;
 
@@ -80,6 +101,7 @@ export class DockButton {
             >
                 {this.renderButton(this.openPopover, 'trigger')}
                 <CustomComponent
+                    ref={this.setCustomComponentElement}
                     {...(this.item.dockMenu.props || {})}
                     onClose={this.onPopoverClose}
                 />
@@ -114,6 +136,10 @@ export class DockButton {
         event.stopPropagation();
         this.isOpen = true;
         this.menuOpen.emit(this.item);
+    };
+
+    private setCustomComponentElement = (element: HTMLElement) => {
+        this.customComponentElement = element;
     };
 
     private onPopoverClose = () => {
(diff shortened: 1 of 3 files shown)


adrianschmidt:
Copied from Slack conversation, for reference:

> I realised we don't need to bother with portalId, because we have the custom component right there in limel-dock anyway, so we can get a reference straight to it. But then I also realised that limel-dock doesn't know anything about the component, and what it might or might not contain, so we can't go looking for a limel-list that might not even be there. It has to be the consumer that sets the focus I think… :thinking_face:

> Also, we need to make sure that focus is put back on the dock-button when the menu is closed, but that ought to be fairly easy I think.

> Or, no, maybe the consumer shouldn't have to set focus. Maybe limel-dock can find the first focusable element inside the custom component, and just focus that… :thinking_face:

> I think the best way might be to try to use https://github.com/focus-trap/focus-trap. It seems to have some support for shadow-DOM nowadays.

vicgeralds:
> Or, no, maybe the consumer shouldn't have to set focus. Maybe limel-dock can find the first focusable element inside the custom component, and just focus that… :thinking_face:
 
> I think the best way might be to try to use https://github.com/focus-trap/focus-trap. It seems to have some support for shadow-DOM nowadays.

What happened to this discussion? Did you explore this? Why did you decide to let the consumer set the focus?

vicgeralds:
> It seems to have some support for shadow-DOM nowadays.

might still be pitfalls
https://github.com/focus-trap/focus-trap#shadow-dom

> Maybe limel-dock can find the first focusable element inside the custom component

If we can't do this, because the custom component uses shadow dom, I think it's better to avoid shadow dom than forcing the consumer to set focus. It should just be a matter of querying elements having `tabindex` I think.

I haven't quite read up on this topic yet, but I noted that focus behaviour has been "under-specified" until recently
https://blog.whatwg.org/focusing-on-focus

vicgeralds:
We might want to try using [delegatesFocus](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot/delegatesFocus) 

```ts
@Component({
  shadow: { delegatesFocus: true }
})

**Comment by vicgeralds** [comment]
> From the description, it's not entirely clear to me what that does

I agree. But it's quite clear from the name, actually. If it receives focus, it delegates focus to first focusable child. So If I just focus the list, it will work. No need for focus-trap.

image

File: src/components/dock/dock-button/dock-button.tsx [hunk 1]
```diff
@@ -153,4 +172,25 @@ export class DockButton {
...
+        }
+
+        setTimeout(() => {
+            console.log('id', this.portalId);
+            const list: HTMLElement = document.querySelector(
+                `#${this.portalId} limel-menu-surface limel-list`
+            );
+            console.log('LIST', list);
+
+            const firstItem: HTMLElement =
```
vicgeralds, line 188: If we have a reference to the custom component element here, we can just do this:

```typescript
if (element.delegatesFocus) {
    element.focus();
    return;
}
```

vicgeralds (commented):
If we assume the component passed in `item.dockMenu.componentName` is decorated using `shadow: { delegatesFocus: true }` we can just give it focus instead of trying to find the first item.

If the item element that should receive focus has `tabindex`, no more work is needed.,

src/components/dock/dock-button/dock-button.tsx
Befkadu1: Yes I confirm that `delegatesFocus` works as we want. I will push the fix shortly.

src/components/dock/dock-button/dock-button.tsx
```diff
@@ -69,6 +78,18 @@ export class DockButton {
         return this.renderButton(this.handleClick);
     }
 
+    @Watch('isOpen')
+    protected openWatcher() {
+        if (!this.isOpen) {
```
Kiarokh, line 83: it works like a charm! Great job!

My only comment is about the interaction design. After the menu is opened, if user press <kbd>esc</kbd>, the focus will not go back to the trigger element. 

Not sure if it's possible to fix, but it'd be great if user could close the menu using the keyboard, and continue their navigation pattern. This could however be fixed in another PR too.

src/components/dock/dock-button/dock-button.tsx
adrianschmidt: I added `shadow: { delegatesFocus: true }` to the menu component in the example.

adrianschmidt (approved):

Issue #1770, opened 2022-06-03
visualize disabled steps of progress flow better


Kiarokh:
![image](https://user-images.githubusercontent.com/35954987/171864597-4e1e8009-cc6b-4b4e-9135-937525c3e316.png)

Using transparency is not a good way to visualize a disabled step.

adrianschmidt:
If we want to have something that works in most cases without having to either supply a secondary colour value for disabled, or calculating such a value from the given value, adding a semitransparent layer on top of the "buttons" might work. But you might already have some better idea in mind 😄

Kiarokh:
Updated the post with a picture. I had copied in from Microsoft Edge,
![image](https://user-images.githubusercontent.com/35954987/171864904-9b35a3f6-14da-4e66-aeac-0af93f641a94.png)
😠 and to no surprise the freaking thing did not copy the image of course. So I had pasted the linek which was in the clipboard, instead of the image. 😄 

Thanks for the feedback Adrian. I have some ideas. But I'll sketch a bit first.

Kiarokh:
I've had a look. This is quite hard to fix, without applying too many changes unfortunately. I might take it up again during the summer.

Kiarokh:
The easiest fix would be using css filters, however when you apply a filter to something that has no z-index specified, it will mess around with the stacking context. So for example a disabled step with a filter of `filter: saturate(15%);` will go below the surrounding steps and its arrow will be covered by them. This would require us to apply z-index inline on each step, and it should be the reverse of their `index` in the dom. So the first step should get the highest z-index, and so forth.

Another solution might be us finding a better way to visualize a disabled step.

or adding background divs which can get the background color, and the filter on them. But these would require many changes in test files and styles

Kiarokh:
I also tried using `clip-path` for instance `clip-path: polygon(0% 0%, calc(100% - 16px) 0%, 100% 50%, calc(100% - 16px) 100%, 0% 100%);` on the step or the divider. But that also fails us, since we need the divider lines between the steps, and that's hard to get when the elements are masked.

## About the linked PR:

I'd argue that the best way to address this is to avoid transparency on disabled steps for now. It's enough that the disabled steps have a `cursor: not-allowed` and get no shadow effect. 
To give a more clear signal to users, I also added a lock icon for disabled steps.

Selected and disabled 👇 
[...]

PR #1836, opened 2022-09-19
fix(progress-flow): visualize disabled steps better

Kiarokh:
Disabled but passed steps will not look bad.
There is no transparency, and instead there is a
lock icon to visualize disabled state.

fix https://github.com/Lundalogik/lime-elements/issues/1770

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/progress-flow/progress-flow-item/partial-styles/_selected-indicator.scss
+++ b/src/components/progress-flow/progress-flow-item/partial-styles/_selected-indicator.scss
@@ -6,11 +6,7 @@
         position: absolute;
         right: var(--selected-indicator-right);
 
-        background-color: var(--step-divider-color);
-        width: functions.pxToRem(6);
-        height: functions.pxToRem(6);
         border-radius: 50%;
-        opacity: 0.7;
     }
 
     .last & {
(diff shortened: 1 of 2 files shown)


LucyChyzhova:
I love your solution, Kia!🤩😍 Simple and clear! 💪 

I found one edge case which we could improve 😁 . please take a look at the video
(if these disabled steps are between active steps and they are passed, the locks  🔒  icons are hard to see => not clear why we can't press those disabled steps 

https://user-images.githubusercontent.com/50618208/191026407-a94c735e-b933-4cbb-8c7c-9b2284cb23fa.mov

Kiarokh:
Thanks a lot Lucy! Good catch! 
I'll see what I can do. 😊 and will add a fixup!

LucyChyzhova:
I guess the background layer -grey circle  ⭕️  is gone
Screenshot 2022-09-19 at 15 30 39

LucyChyzhova (approved):
great job! 👏  🤩 💪

Kiarokh:
> I guess the background layer -grey circle ⭕️ is gone Screenshot 2022-09-19 at 15 30 39

Depending on the background color of the button, some color will be applied due to the `mix-blend-mode: difference;` filter. This makes sure that the color of the lock icon is always "visible"-ish. However, the background circle will also be affected. Since it's too small, it's perhaps hard to notice. But it's there.

LucyChyzhova (approved):

PR #1838, opened 2022-09-21
Split button

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/2953
close https://github.com/Lundalogik/hack-tuesday/issues/265

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -24,7 +24,12 @@ button {
 
     &.mdc-button {
         min-width: functions.pxToRem(36);
-        padding: 0 functions.pxToRem(12);
+        padding: {
+            top: 0;
+            right: var(--button-padding-right, #{functions.pxToRem(12)});
+            bottom: 0;
+            left: var(--button-padding-left, #{functions.pxToRem(12)});
+        }
         height: 100%;
         min-height: functions.pxToRem(36);
         width: 100%;
--- a/src/components/split-button/split-button.tsx
+++ b/src/components/split-button/split-button.tsx
@@ -1,0 +1,90 @@
+import { Component, Event, EventEmitter, Host, h, Prop } from '@stencil/core';
+import { ListSeparator, MenuItem } from '@limetech/lime-elements';
+
+/**
+ * A split button is a button with two components:
+ * a button and a side-menu attached to it.
+ *
+ * Clicking on the button runs a default action,
+ * and clicking on the arrow opens up a list of other possible actions.
+ *
+ * :::warning
+ * - Never use a split button for navigation purposes, such as going to next page.
+ * The button should only be used for performing commands!
+ * - Never use this component instead of a Select or Menu component!
+ * :::
+ *
+ * @exampleComponent limel-example-split-button-basic
+ * @exampleComponent limel-example-split-button-repeat-default-command
+ */
+@Component({
+    tag: 'limel-split-button',
+    styleUrl: 'split-button.scss',
+    shadow: true,
+})
+export class SplitButton {
+    /**
+     * The text to show on the default action part of the button.
+     */
+    @Prop({ reflect: true })
+    public label: string;
+
+    /**
+     * Set to `true` to make the button primary.
+     */
+    @Prop({ reflect: true })
+    public primary = false;
+
+    /**
+     * Set icon for the button
+     */
+    @Prop({ reflect: true })
+    public icon: string;
+
+    /**
+     * Set to `true` to disable the button.
+     */
+    @Prop({ reflect: true })
+    public disabled = false;
+
+    /**
+     * A list of items and separators to show in the menu.
+     */
+    @Prop()
+    public items: Array<MenuItem | ListSeparator> = [];
+
+    /**
+     * Is emitted when a menu item is selected.
+     */
+    @Event()
+    public select: EventEmitter<MenuItem>;
+
+    render() {
+        return (
+            <Host>
+                <limel-button
+                    label={this.label}
+                    primary={this.primary}
+                    icon={this.icon}
+                    disabled={this.disabled}
+                />
+                <limel-menu
+                    class={{
+                        primary: this.primary,
+                    }}
+                    disabled={this.disabled}
+                    items={this.items}
+                    openDirection="bottom"
+                >
+                    <button
+                        class="menu-trigger"
+                        slot="trigger"
+                        disabled={this.disabled}
+                    >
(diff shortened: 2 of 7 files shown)


Befkadu1:
One of the tests fails here but passes locally. The issue is the icon could not be found here on gh. I am not sure how to handle it here on gh. The icon `http://localhost:3333/assets/icons/cat.svg` could not be found. @adrianschmidt anything you want to say here?

Screenshot 2022-11-01 at 2 25 23 PM

adrianschmidt (commented):
I've removed the dev-dependency on `lime-icons8` so that external contributors should not encounter problems because they cannot install it. This also speeds up the dev build a lot, since it doesn't have to copy 5000 icons. To permit the tests to still function, I've added a public domain svg called `unit-test.svg`. So in tests, just use `"unit-test"` as the name for the icon.

There are also a few other files in the icons folder, but these are all the same image, just with different filenames, because the name of the icon was hard-coded in certain places:

image

src/components/split-button/test/split-button.e2e.ts
```diff
@@ -0,0 +1,195 @@
...
+
+                splitButton = await page.find('limel-split-button');
+                limelButton = await page.find(
+                    'limel-split-button>>>limel-button'
+                );
+            });
+
+            it('inner limel-button renders the icon "cat"', async () => {
+                const propValue = await limelButton.getProperty('icon');
+                expect(propValue).toBe('cat');
```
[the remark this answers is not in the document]
Befkadu1: Ok great! Can you review only the unit tests? We did the whole thing together I am not sure if we need another reviewer though

src/components/split-button/test/split-button.e2e.ts
```diff
@@ -0,0 +1,195 @@
+import { E2EElement, E2EPage, newE2EPage } from '@stencil/core/testing';
```
adrianschmidt, line 1: Is there any particular reason for why these test files have been places in a sub-folder called `test`? We haven't done that anywhere else in this repo as far as I know, and I prefer not to separate the tests from the code being tested like that 🙂

src/components/split-button/test/split-button.e2e.ts
```diff
@@ -0,0 +1,195 @@
...
+                    <limel-split-button></limel-split-button>
+                `);
+
+                splitButton = await page.find('limel-split-button');
+                limelButton = await page.find(
+                    'limel-split-button>>>limel-button'
+                );
+            });
+
+            it('the property is falsy', async () => {
```
adrianschmidt, line 49:
```suggestion
            it('the inner button is not "primary"', async () => {
```

src/components/split-button/test/split-button.e2e.ts
```diff
@@ -0,0 +1,195 @@
...
+                     <limel-split-button></limel-split-button>
+                `);
+
+                splitButton = await page.find('limel-split-button');
+                limelButton = await page.find(
+                    'limel-split-button>>>limel-button'
+                );
+            });
+
+            it('the property is falsy', async () => {
```
adrianschmidt, line 87:
```suggestion
            it('the inner button is not disabled', async () => {
```

adrianschmidt (changes requested):
I'd like some of the test descriptions to be updated. The descriptions should describe what our "goal" is, not what we technically do in the test to try to determine if that "goal" has been reached. The code itself describes that just fine 🙂

I considered for a while if we should make the tests actually look at, for example, whether the native button actually used by `limel-button` is actually disabled or not, but I feel like that would be testing too much implementation. We can't really get around the fact that we have to test the implementation of the `limel-split-button` component, since we can't just write a test like "check that there is a button of some kind", but at least we can avoid to test the implementation of _another_ component (limel-button). Testing that if the `disabled` prop is set on the limel-button, it actually renders a disabled button, belongs in the tests for limel-button, so in _these_ tests I think it's enough to check that the prop is set on the limel-button. Just like you've done! 😄 👍

src/components/button/button.scss
```diff
@@ -24,7 +24,12 @@ button {
 
     &.mdc-button {
         min-width: functions.pxToRem(36);
-        padding: 0 functions.pxToRem(12);
+        padding: {
+            top: 0;
+            right: var(--button-padding-right, #{functions.pxToRem(12)});
+            bottom: 0;
+            left: var(--button-padding-left, #{functions.pxToRem(12)});
+        }
```
adrianschmidt, line 32: This commit should come before the other one 🙂

src/components/split-button/test/split-button.e2e.ts
Befkadu1: > Is there any particular reason for why these test files have been places in a sub-folder called test?

I think you are the one who did it. Maybe can you answer it 😆 ? You created the test file running the command `npm run generate`

Kiarokh: Never trust robots! 😅

adrianschmidt (approved):

src/components/split-button/split-button.tsx
```diff
@@ -0,0 +1,90 @@
...
+ * a button and a side-menu attached to it.
+ *
+ * Clicking on the button runs a default action,
+ * and clicking on the arrow opens up a list of other possible actions.
+ *
+ * :::warning
+ * - Never use a split button for navigation purposes, such as going to next page.
+ * The button should only be used for performing commands!
+ * - Never use this component instead of a Select or Menu component!
+ * :::
```
adrianschmidt, line 15: No _must_, and can be polished later too, but it might be nice to have some example that's a little easier to understand for these "warnings". Like the DOs and DON'Ts that the Material Design docs use, or something.

Also, it might be worth adding something about how it's good to collect "similar" actions under a split button (like _Save_, and _Save As_), but not to collect "dissimilar" actions (like _Save_ and _Delete_).

adrianschmidt:
(Don't mind the SonarCloud failure. It's some check for code duplication, which honestly just shouldn't be run for test files, in my opinion. Then there's some check that says the test coverage of the newly added code is 0 %, which is just blatantly untrue 🤦‍♂️)

PR #1840, opened 2022-09-22
use mixin to visualize keyboard focus

Kiarokh:
We use the same mixin in crm-components already.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dock/dock-button/dock-button.scss
+++ b/src/components/dock/dock-button/dock-button.scss
@@ -5,6 +5,10 @@
 .button {
     all: unset;
 
+    cursor: pointer;
+    @include mixins.is-flat-clickable();
+    @include mixins.visualize-keyboard-focus;
+
     box-sizing: border-box;
     display: flex;
     align-items: center;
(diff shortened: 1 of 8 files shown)


src/components/dock/dock-button/dock-button.scss
```diff
@@ -44,6 +41,11 @@
         );
         box-shadow: var(--button-shadow-inset);
 
+        &:focus-visible {
+            outline: none;
+            box-shadow: var(--shadow-depth-8-focused);
+        }
+
```
LucyChyzhova, line 48: could we replace it with @mixins as well?

Kiarokh: ah it's further up. this is a repetition of the same code 😄 good find

Kiarokh: Also I noticed that we didn't have `disabled` on dock-button, so I removed the styles for it

Kiarokh: Good catch anyways!

LucyChyzhova (dismissed):
Great optimization! 🤩

LucyChyzhova (approved):

PR #1843, opened 2022-09-28
fix(dock-button): give back focus when menu is closed

vicgeralds:
Almost fixes https://github.com/Lundalogik/lime-elements/pull/1820#discussion_r978358996

fix: https://github.com/Lundalogik/crm-feature/issues/2989

The tricky part is we don't know here how the menu was closed. If the user navigated to an item in the menu, we don't expect focus back.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dock/dock-button/dock-button.tsx
+++ b/src/components/dock/dock-button/dock-button.tsx
@@ -66,6 +66,10 @@ export class DockButton {
     private tooltipId: string;
     private customComponentElement: HTMLElement;
 
+    private buttonElement: HTMLButtonElement;
+
+    private hadVisibleFocus = false;
+
     constructor() {
         this.tooltipId = createRandomString();
     }
@@ -123,7 +127,10 @@ export class DockButton {
                     button: true,
                     selected: this.item?.selected,
                 }}
+                ref={this.setButtonElement}
                 onClick={handleClick}
+                onFocus={this.handleFocus}
+                onBlur={this.handleBlur}
             >
                 {this.renderIcon()}
                 {this.renderLabel()}
@@ -142,7 +149,15 @@ export class DockButton {
         this.customComponentElement = element;
     };
 
+    private setButtonElement = (element: HTMLButtonElement) => {
+        this.buttonElement = element;
+    };
+
     private onPopoverClose = () => {
+        if (this.hadVisibleFocus) {
+            this.buttonElement.focus();
+        }
+
         this.isOpen = false;
         this.close.emit();
     };
@@ -150,6 +165,16 @@ export class DockButton {
     private handleClick = (event: MouseEvent) => {
         event.stopPropagation();
         this.itemSelected.emit(this.item);
+    };
+
+    private handleFocus = () => {
+        this.hadVisibleFocus = this.buttonElement.matches(':focus-visible');
+    };
+
+    private handleBlur = () => {
+        if (!this.isOpen) {
+            this.hadVisibleFocus = false;
+        }
     };
 
     private renderIcon() {


src/components/dock/dock-button/dock-button.tsx
```diff
@@ -142,7 +149,15 @@ export class DockButton {
         this.customComponentElement = element;
     };
 
+    private setButtonElement = (element: HTMLButtonElement) => {
+        this.buttonElement = element;
+    };
+
     private onPopoverClose = () => {
+        if (this.hadVisibleFocus) {
```
vicgeralds, line 157: Maybe we can check that the popover wasn't closed due to click. I guess we could just set `this.hadVisibleFocus = false` in a click event handler.

adrianschmidt: I think we use a scrim for popover, right? So it shouldn't be possible to focus on something outside the open menu with a single click anyway. You first need to click—either on something inside the menu which closes it, or outside the menu—to close the menu, then click on the thing you want to focus. And if you don't do that second click, but tab instead, we should, ideally, be tabbing to the next this after the trigger.

So I think the _best_ thing to do, would be to always set focus on the button, but not make the focus visible if closing was triggered from a pointer-event. The next best thing would be to always set focus, and either just keep it visible or invisible, as it was. The third best would probably be to always set focus but make it always visible. And the fourth best to only set focus if closing was triggered by a non-pointer event. (@Kiarokh might disagree about the order of these last two…)

I haven't looked at the rest of the code yet, but I'm guessing `this.hadVisibleFocus` is true if the trigger had visible focus when the menu was opened? Using that is not perfect, but descent enough, because it should guarantee that we set focus for any user that uses _only_ non-pointer input methods (users who use _only_ the keyboard, for example). Those are the users for whom it's most important to handle focus correctly.

Kiarokh: It's a bit hard for me to follow everything and I'm not sure how everything works exactly. So sorry if my answer is off.

But if I could wish, my wish would be limiting the focus visible to keyboard interactions only. Clicking to close and open stuff is an entirely different story, and very hard to find a good solution for. [To see why, follow the discussion here…](https://github.com/Lundalogik/lime-elements/pull/1197).

So one of the things I dislike with setting the `focus` the way we do it today is the green outline ring that we will get from `focus-visible` styles, even if the menus (or popovers) are opened using a click/tap interaction. The `focus-visible` styles are for keyboard navigation only; but we get them now for mouse interactions, which is not good. I guess this is what you mean when you say:
> but not make the focus visible if closing was triggered from a pointer-event. 

https://user-images.githubusercontent.com/35954987/192983019-ee807354-ea8b-4283-96ee-a99b88c79eb5.mov

It is possible to keep the focus "invisible"? And make it only visible if navigation is done using keyboard?



Anyhow, I guess what Lucy wanted to achieve was setting focus back on the trigger, if we close the popover by pressing <kbd>esc</kbd>, after having opened the popover by pressing <kbd>enter</kbd> on its trigger (all keyboard navigation). Or?

adrianschmidt: I think I have a solution where we can _always_ set focus freely, and not worry about what type of event triggered the focus-setting—and control the _visibility_ of the current focus completely separately.

I'll keep you in suspense for a little while though, while I fix up some other things 😜

Kiarokh: this is torture! 😄

adrianschmidt: If we use a css variable for the colour of the highlight, and define that on the page's html or body element, we can control the look of the highlight globally throughout the page.

Then we set up some event listeners on the `body` element, which listen any relevant events, and sets the highlight colour to make it either visible or transparent, as fits the situation. Crucially, we listen to the event in the _capture_ phase, which will let us catch the event before any event listeners further down in the DOM catches it and potentially runs `stopPropagation()` on it…

(Note: it does _not_ work to attach the listener to `document`, as I discovered while testing a bit just now.)

Kiarokh: I don't 100% understand the technical aspect of what you try to explain. 

But we currently apply this focus style using `box-shadow: var(--shadow-depth-8-focused);` via https://github.com/Lundalogik/lime-elements/blob/d77d37ed7bcf6e6fda7519f898882b7c8b60bf11/src/style/mixins.scss#L10
The var applies this style:
```scss
--shadow-depth-8-focused: 0 0.2rem 0.45rem 0 rgba(var(--color-black), 0.132), 0 0.0375rem 0.1125rem 0 rgba(var(--color-black), 0.108), 0 0 0 0.125rem var(--mdc-theme-primary);
```
which means what gives it the color is `var(--mdc-theme-primary)`. In Lime Go, this would turn to orange for instance. 

We use this var in all repos to visualize the same thing. But I don't think we could override `var(--mdc-theme-primary)` or `var(--shadow-depth-8-focused)` with something else. Or?

adrianschmidt: No, exactly, `--shadow-depth-8-focused` would not be a good thing to override, as other components may also use this. (Why are we using such a weirdly named variable? "Shadow depth 8" sounds like it would have something to do with shadows?)

We should use a new variable, maybe something with a slightly more intuitive name, like `--lime-focus-border` perhaps? (Prefixed since it's global and it's important that other libraries doesn't overwrite it by accident.) Then it would be no problem to just change the value of that variable, as we see fit.

Kiarokh: the reason why is in the examples in this guide: https://lundalogik.github.io/lime-elements/versions/next/#/DesignGuidelines/shadows-and-depth.md/

So I think naming is good enough. :)

adrianschmidt: Ok, but then I guess it's not supposed to be used for anything other than conveying where the focus is? So then I guess we could override that to make the focus outline invisible?

(Oh, I _knew_ "highlight" wasn't the word I was looking for. "Outline" is what it's usually called! 😅)

Kiarokh: yes it's used for indicating where the focus is. So if the element already has a shadow with another depth (e.g. e popover), we must set the correct focus style for that depth.

I guess focus can only be at one place at a time. So it should be perhaps safe to override it. But I still don't think I understand what you meant about how it's done technically.

****

oh yes… all these naming things (or the Nomenclature 😄 ).  The problem is even `outline` is a thing for itself in CSS, with certain limitations.. I think the latest material design is using a term called "Focus ring". No idea how it's implemented though.

adrianschmidt: Yeah, I know "outline" is technically its own thing, but that's what we're recreating basically. But sure, "focus ring" is perhaps a good name for it. That's implementation agnostic 🙂

adrianschmidt: > But I still don't think I understand what you meant about how it's done technically.

I was tempted to implement it today, but I need to log off. Perhaps tomorrow… 😊

adrianschmidt: https://github.com/Lundalogik/lime-elements/pull/1845
https://github.com/Lundalogik/lime-crm-components/pull/1268

It's not perfect yet, because we use `--shadow-depth-8-focused` for everything, including things that aren't otherwise `--shadow-depth-8`, so things still get the shadow when focused.

Kiarokh: > so things still get the shadow when focused. 

That's the whole intention. No matter if an element is elevated and has a visual depth, it should still be able to get focused (and keep its elevation level). So This is completely fine, and the main reason we did it. That's why we have a focused variant for each shadow.

vicgeralds: Good to see lots of discussion (and work) here. 🙂 

> So I think the best thing to do, would be to always set focus on the button, but not make the focus visible if closing was triggered from a pointer-event. 

I'm not sure about that. The main use case is that the user clicks (or keyboard press enter) to navigate something in the menu. In that case I think we want to focus whatever should be focused on that page on "page load". Of course, that's not the concern of our dock button, but isn't there a risk that we interfere with that if we set the focus on click?

I took the approach to try to deal with `:focus-visible`/keyboard interaction as a special case, which narrows down the problem when closing the menu, but I can see that focus-setting in general is a big topic where we might want to avoid treating visible focus differently.

vicgeralds: > Crucially, we listen to the event in the capture phase, which will let us catch the event before any event listeners further down in the DOM catches it

Nice! That's a new trick that I just learned about.

> I don't 100% understand the technical aspect of what you try to explain.

It's ok 😄

vicgeralds: > and sets the highlight colour to make it either visible or transparent, as fits the situation.

Isn't this the same thing as using the `:focus-visible` selector? Except we may override the user agent choice?

adrianschmidt: > Good to see lots of discussion (and work) here. 🙂
> 
> > So I think the best thing to do, would be to always set focus on the button, but not make the focus visible if closing was triggered from a pointer-event.
> 
> I'm not sure about that. The main use case is that the user clicks (or keyboard press enter) to navigate something in the menu. In that case I think we want to focus whatever should be focused on that page on "page load". Of course, that's not the concern of our dock button, but isn't there a risk that we interfere with that if we set the focus on click?
>
> I took the approach to try to deal with `:focus-visible`/keyboard interaction as a special case, which narrows down the problem when closing the menu, but I can see that focus-setting in general is a big topic where we might want to avoid treating visible focus differently.

Sorry, I didn't mean that we should always set focus on the button regardless of why the menu was closed (or, if that was what I meant, I recant). What I mean is that whether the user interacted using a keyboard or a mouse, should not determine whether or where we set focus. It should only determine whether we make the focus visible.

> > and sets the highlight colour to make it either visible or transparent, as fits the situation.
> 
> Isn't this the same thing as using the `:focus-visible` selector? Except we may override the user agent choice?

Ehh… yes, probably? 😬 Sorry, I haven't kept up to date with new CSS features for quite a few years, so, while I recognise `:focus-visible` now, I had completely forgotten about it. If that does what we want, why aren't we using it in the first place then? 🤷‍♂️

vicgeralds: > whether the user interacted using a keyboard or a mouse, should not determine whether or where we set focus. It should only determine whether we make the focus visible.

Problem is we can only do that by overriding browser behaviour. The browser will make focus visible or not, but we can only set focus, and when we do that with `element.focus()` it seems browsers make focus visible.

> why aren't we using it in the first place then?

I think we should, but we can't really rely on `:focus-visible` if we want to use `element.focus()` to also set "hidden" focus.

adrianschmidt: > > whether the user interacted using a keyboard or a mouse, should not determine whether or where we set focus. It should only determine whether we make the focus visible.
> 
> Problem is we can only do that by overriding browser behaviour. The browser will make focus visible or not, but we can only set focus, and when we do that with `element.focus()` it seems browsers make focus visible.

OK, but then what I suggest clearly isn't the same as just using `:focus-visible`!? 😅

vicgeralds: I think it is!? It's only the focus-setting part that's different? 😄

adrianschmidt: Final remark: Why I think we should _set_ focus the same, regardless of whether focus is _visible_, is that, if I for example click the scrim to close a menu, or press escape to close it, and then press tab, I want the same thing to happen regardless.

But, sure, it's less critical that focusing is handled correctly when using a mix of mouse and keyboard than when using _only_ keyboard 😊 👍

vicgeralds: You too! 🤗 

Sorry, my comment didn't make things any clearer. I didn't mean that we should ignore that part. What I meant with same thing as using `:focus-visible` was specifically styling focus as visible or not based on input method. My point is trying to solve that ourselves may not be the best idea as there's a web standard solution available.
I think they're doing a good job with [focus-visible](https://github.com/WICG/focus-visible). It might be that they solve some issue that we haven't thought about. It might also be that it's not ready yet for something we want.

About what needs to change in this PR, Adrian convinced me that we shouldn't set focus conditionally only when we want it visible. 👍 

I saw that `element.focus({ focusVisible })` was added very recently to the html standard: https://github.com/whatwg/html/pull/8087, so there's some hope other browsers will implement it soon (enough?). We could start using that, and continue the discussion in the other PR about how we should style focus as visible or not.

PR #1845, opened 2022-09-30
feat: add CSS variable `--lime-focus-ring-color`

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/style/shadows.scss
```diff
@@ -1,9 +1,11 @@
 @use './functions';
 
-$visualise-focused-state: 0 0 0 #{functions.pxToRem(2)} var(--mdc-theme-primary);
+$visualise-focused-state: 0 0 0 #{functions.pxToRem(2)} var(--lime-focus-ring-color);
 $visualise-error-state: 0 0 0 #{functions.pxToRem(2)} rgb(var(--color-red-default));
 
 :root {
+    --lime-focus-ring-color: var(--mdc-theme-primary);
```
adrianschmidt, line 7: This isn't perfect yet, because we use `--shadow-depth-8-focused` for everything, including things that aren't otherwise `--shadow-depth-8`, so things still get the shadow when focused.

Kiarokh: That's intentional. 

1. No matter if an element is elevated and has a visual depth, it should still be able to get focused (and keep its elevation level). So This is completely fine, and the main reason we did it. That's why we have a focused variant for each shadow.

2. If something is flat (has no shadow/depth)  but interactive, it gets a shadow/depth  when hovered. When pressed (using `:active` styles), it goes down. This is the type of skeuomorphic design that we want to have, as described [here](https://lundalogik.github.io/lime-elements/versions/next/#/DesignGuidelines/shadows-and-depth.md/). This gives them "affordance". So focusing on something with keyboard shouldn't be any different. If something is flat and gets focused, it can also get elevated. 

Sadly, there is no CSS selector like `:active` to visualize things that are "activated" by pressing a key on the keyboard. So that part of the interaction can not be visualized with CSS only. But we can still elevate and lift the element visually, when it's focused by keyboard.

PR #1852, opened 2022-10-04
feat: allow string labels in limel-badges

### LINKED ISSUE #1851: limel-badge with words
## New feature motivation
`limel-badge` would be amazing to not just show number indicators but also short words. In our use case we want to point out that something is new.

## New feature description

Change the type specification of label from number to number or string. Make sure to shorten strings to a max of 5 characters to avoid long badges.

alexwero:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

Fix #1851 
The "Styles" specification (https://lundalogik.github.io/lime-elements/versions/35.0.0/#/component/limel-badge/styles/) doesn't appear locally, but I'm guessing it has something to do with how docs are generated?


src/components/badge/badge.tsx
```diff
@@ -11,12 +11,17 @@ import { abbreviate } from './format';
...
     @Prop({ reflect: true })
-    public label: number;
+    public label: number | string;
 
     render() {
-        return abbreviate(this.label);
+        if (typeof this.label === 'number') {
+            return abbreviate(this.label);
+        } else if (typeof this.label === 'string') {
+            return truncate(this.label);
```
Kiarokh, line 24: From what I see in the preview docs, this doesn't really "truncate" the label, which is not good

We want this 👇 :
image

but we are getting this 👇 
image

Also, this creates accessibility issues, because:
1. the word is in fact turned to `Webco` in the DOM. So even screen readers will read it out loud like that
2. there is no way for a sighted user to know what the word is, for instance by hovering on the component and seeing it as a tooltip. So putting this as `title` attribute on the badge would be a great solution.

To avoid these, you can leave the truncation job to CSS. Thus we don't need a function for it at all.

Kiarokh (changes requested):

Kiarokh:
Also don't forget to add the scope in the commit message. Like this `feat(badge)`

So this. would be fine:
`feat(badge): allow string labels`
or 
`feat(badge): enable having string labels`

src/components/badge/examples/badge-number.tsx
```diff
@@ -1,8 +1,14 @@
...
+/**
+ * Number badges
+ *
+ * Numbers containing more than 3 digits get abbreviated
+ * using number units (K, M, B, and T).
+ */
 @Component({
-    tag: 'limel-example-badge',
-    styleUrl: 'badge.scss',
+    tag: 'limel-example-badge-number',
```
Kiarokh, line 10: image
The reason for this is that you renamed the component tag here. But the CSS is still saying `:host(limel-example-badge) {` So you should just add a -badge in there too 😊

src/components/badge/badge.tsx
```diff
@@ -1,8 +1,9 @@
-import { Component, Prop } from '@stencil/core';
+import { Component, Prop, h } from '@stencil/core';
 import { abbreviate } from './format';
 
 /**
- * @exampleComponent limel-example-badge
+ * @exampleComponent limel-example-badge-number
```
Kiarokh, line 5:
```suggestion
 * The Badge component can display both `number` and `string` as `label`.
 * 
 * @exampleComponent limel-example-badge-number
```

src/components/badge/badge.tsx
```diff
@@ -11,12 +12,25 @@ import { abbreviate } from './format';
 })
 export class Badge {
     /**
-     * Label to display in the badge
+     * Label to display in the badge.
+     * String labels get truncated if longer than 6ch
```
Kiarokh, line 16:
```suggestion
     * String labels get truncated if their visual length is longer than
     * six characters (six `0`s to be exact).
```

src/components/badge/badge.tsx
```diff
@@ -11,12 +12,25 @@ import { abbreviate } from './format';
...
      */
     @Prop({ reflect: true })
-    public label: number;
+    public label: number | string;
 
-    render() {
-        return abbreviate(this.label);
+    public render() {
+        if (typeof this.label === 'number') {
+            return abbreviate(this.label);
```
Kiarokh, line 23: I also found out that we don't abbreviate decimals. So badges can become super long. It'd be nice to add this later as a separate fix. Perhaps another PR.
image

src/components/badge/examples/badge-number.tsx
```diff
@@ -1,8 +1,14 @@
 import { Component, h } from '@stencil/core';
 
+/**
+ * Number badges
+ *
+ * Numbers containing more than 3 digits get abbreviated
+ * using number units (K, M, B, and T).
```
Kiarokh, line 7:
```suggestion
 * Numeric labels larger than 999 will get both rounded and abbreviated.
 * For example, if the label is `1090` the badge will display `1.1K`.
 * Abbreviation units used are `K` (Kilo) that stands for Thousands, 
 * `M` for Millions, `B` for Billions, and `T` for Trillions.
```

src/components/badge/examples/badge-string.tsx
```diff
@@ -0,0 +1,27 @@
+import { Component, h } from '@stencil/core';
+
+/**
+ * String badges
+ *
+ * Strings get truncated if they take up too much space
+ * (defined as ...). Hovering the cursor on the truncated
+ * badges shows the full text.
```
Kiarokh, line 8:
```suggestion
 * String labels get truncated if their visual length is longer than
 * six characters placed side by side (six `0`s to be exact).
```

Kiarokh, line 8: > Hovering the cursor on the truncated badges shows the full text.


This is not how it works yet. So we have to add this with the next commit, when we add this functionality.

Kiarokh (changes requested):

src/components/badge/examples/badge-number.tsx
```diff
@@ -1,8 +1,16 @@
 import { Component, h } from '@stencil/core';
 
+/**
+ * Number badges
+ *
+ * Numeric labels larger than 999 will get both rounded and abbreviated.
+ * For example, if the label is `1090` the badge will display `1.1K`.
+ * Abbreviation units used are `K` (Kilo) that stands for Thousands,
```
Kiarokh, line 8:
```suggestion
 *
 * Abbreviations are indicated by `K` (Kilo) that stands for Thousands,
```

src/components/badge/badge.tsx
Kiarokh:
```suggestion
            return <span>{abbreviate(this.label)}</span>;
```

Kiarokh (dismissed):

src/components/badge/examples/badge-number.tsx
jgroth: Not sure how picky we should be or if this is just confusing... but if we are using SI units, uppercase `K` actually stands for kelvin and measures degrees :smile: Lowercase `k` stands for `kilo`
https://en.wikipedia.org/wiki/International_System_of_Units

Kiarokh: Wow...! I had no clue. But doesn't `MM` actually stand for millions? So I don't know neither. Maybe it's just a UI matter here. What do you think @LucyChyzhova ?

I myself prefer lowercase for everything to be honest. 
Uppercase is kind of both: 
- aggressive & ugly; 
- and also makes it harder to see where the number ends and the letter starts (for my dyslexic brain at least). 
 
Maybe we should convert all to lowercase instead?

Kiarokh: I'm pretty sure nobody except you and they guy who write the Wiki knows this 😄 Maybe we should just do whatever we want

jgroth: Haha, most people that have studied math or physics probably know it :smile:

LucyChyzhova: I think it's should be uppercase M. agree with Johan, this convention comes from the math lessons 🤓

src/components/badge/badge.tsx
Kiarokh: OK, since you are not available right now @alexwero I added the fixup myself. I'll also add an additional fix commit for the next thing, and we will ask others to review this PR.

Kiarokh (commented):
since I added a fix myself, it's better that someone else reviews this as well. But from my point of view this is fine.

LucyChyzhova (changes requested):
If the label is not truncated or abbreviated there is no sense in showing the helper cursor and repeating the value.  
The same is true for the numbers: only if the value is more than 1000 we show a helper cursor with a precise value.

Kiarokh:
Makes great sense Lucy. I'll add a fixup in sec!

LucyChyzhova:
could you please change the commit message to:
fix(badge): use unabbreviated & un-truncated label as `title` to improve accessibility

Kiarokh:
⚡ done, and squashed and forcepushed

src/components/badge/examples/badge-number.tsx
jgroth: I'm intrigued :eyes: :smile:

Kiarokh: It's OK. 😄 
I think in the contexts where badges are used, nobody would interpret their values as Kelvin. Even those who know the difference. I mean you normally won't use a badge to visualize the temperature or things like that. They're always counters.
image

But this is an unimportant thing really, the lowercase works too.

I also like the idea of adding a space. Tried it in the image above ☝️

adrianschmidt: It's not that people might misunderstand. It's that it's very annoying and makes us look ignorant.

Or to put it the Günther Steiner way: https://youtu.be/W5WVVls74mI?t=25 😂

PR #1853, opened 2022-10-06
feat(progress-flow): update some internals to enable tracking using Heap 

TommyLindh2:
## Description

We want to add `value` to each `progress-flow-item` element just like a `select` element have a `value` on each `option` element.
This is for two reasons.

1. I think it's expected behavior.
2. We need it to track a specific flow in Heap (Product analytics).

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/progress-flow/progress-flow.tsx
+++ b/src/components/progress-flow/progress-flow.tsx
@@ -99,6 +99,7 @@ export class ProgressFlow {
                 readonly={this.readonly}
                 item={item}
                 onInteract={this.handleFlowItemClick(item)}
+                data-tracking-value={item.value}
             />
         );
     };
@@ -122,6 +123,7 @@ export class ProgressFlow {
                 readonly={this.readonly}
                 item={item}
                 onInteract={this.handleFlowItemClick(item)}
+                data-tracking-value={item.value}
             />
         );
     };


src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
```diff
@@ -49,26 +50,28 @@ export class ProgressFlowItem {
...
-                disabled={this.isDisabled()}
-            >
-                {this.renderIcon()}
-                <span class="text">{this.item.text}</span>
-                {this.renderDivider()}
-            </button>,
-            this.renderSecondaryText(),
-        ];
+        return (
+            <Host value={this.item?.value}>
```
Kiarokh, line 54: 1. I don't think I'm the right person to review this :)
2. But can you please explain a little bit in the PR or the issue, about what this does or why you need it. It helps the reviewer.
3. Shouldn't the commit message be `feat`?

jgroth: I'm a bit confused I think :smile: 
This should probably be set by the parent component instead, which means you should have a `@Prop` named `value` in here. However, I'm not sure it's needed since you already have `item`, which in turn contains `value` :slightly_smiling_face: 
I might be missing something though :smile:

jgroth: Yeah, I guessed you needed it in the DOM :slightly_smiling_face: 
But can you not just use `element.item.value` then? :thinking:

vicgeralds: I guess you want to be able to use a selector like `limel-progress-flow-item[value=something]`?

Reflecting `item.value` to attribute `value` would be strange and require a hacky solution in a Stencil component I think.

Changing this component to accept `value` prop in addition to `item` (for just this component) would also be strange. 

If you need this, it is an issue with any component using props based on `ListItem` interface, so it's a bigger issue that needs more discussion.

jgroth: On another note, this component is marked with `@private` which hints that consumers should not be aware of it and just interact with the parent component instead

TommyLindh2: Yeah, I was actually a little surprised that the `value` wasn't set on each item using for example `ListItem` when I looked. So I agree that we maybe need to have a bigger discussion.

So the reason we need it in Heap is that it's a tool that tracks all events and stuff for example click events and it brings the selector of the clicked item (including the shadow root parts). So for us to example measure how long a user is between two different stages in our lead flow for example, we need to be able to see the value on it.

But as I said before, I really think that the value should be on the value elements even on a limel-list, select and so on.
And there's still a lot in lime-elements that is not consistent already, so I thought that maybe a few steps in the right direction at the time is ok. Do you think this will affect something in elements?

TommyLindh2: @vicgeralds @jgroth  anything more to add, would be nice to know if you want us to change something or if you just want to close this without merging.

blombergedvin:
Our (and probably addons teams) tracking of user workflows in Heap is broken without this or something similiar. We can't track what a user presses on in a progress flow component, which is rather important. 
Do you guys have any input to Tommys questions?

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
Kiarokh: I'll put this in review column. I think everyone forgot about it

vicgeralds: No, without giving this higher priority over other things, I can only think of hacky solutions that we don't want.

vicgeralds:
> Our (and probably addons teams) tracking of user workflows in Heap is broken without this or something similiar. We can't track what a user presses on in a progress flow component, which is rather important.

How can this be broken? It was never a working feature. You expect to able to track anything (and Heap promises that you can), but it's only possible (if at all) by relying on internal implementation details. I don't think we have discussed the scope of what you can expect to be able to track in any detail.

blombergedvin:
> How can this be broken? It was never a working feature. You expect to able to track anything (and Heap promises that you can), but it's only possible (if at all) by relying on internal implementation details. I don't think we have discussed the scope of what you can expect to be able to track in any detail.

Broken might have been a poor word to use I realize. Maybe it should have been seen as with the shadow DOM issue, where big parts of the functionality in heap couldn't be utilized. The scope of what we would be able to track might not been completely defined since we expected most things to work and that we are learning about the possibilities as we go. But at the same time we expected that our platform might require tweaks and changes to work correctly with heap.

I was not saying that you were to drop everything to work on this issue, I was just referring to the discussion happening above, and that we don't leave it at that but rather have plan forward. Paying a license and investing a lot of resources (mainly yours in the feature teams time and effort) to get heap up and running should not go to waste on the finish line.

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
vicgeralds: Sorry that I didn't reply to your questions earlier. I thought I made my points clear, and I didn't think I had anything to add to help you solving this issue quickly.

> But as I said before, I really think that the value should be on the value elements even on a limel-list, limel-select and so on.

You expect a different interface, that to me seems inconsistent with what we have. Don't get me wrong, I'm not really a fan of the current interface, I see problems/limitations with it, and to me it's quite unusual compared to popular React component libraries, for example. But being able to track things based on values in forms reflected to the DOM is a feature that shouldn't be taken for granted. Tracking could be implemented in other ways as well. We could add tracking to form change events, for example.

> And there's still a lot in lime-elements that is not consistent already, so I thought that maybe a few steps in the right direction at the time is ok.

We haven't identified the right direction 🙂

vicgeralds:
> The scope of what we would be able to track might not been completely defined since we expected most things to work and that we are learning about the possibilities as we go.

I do understand that expectation. Unfortunately, anything that just works can just break if we don't have a common understanding of what's a public interface that can be used for tracking purposes.

I think developers tend to have the opposite expectation, that anything that isn't clearly meant as a public API or explicitly tracking related shouldn't be tracked. I've had previous experience where the feature team only added tracking to specific parts on request or on our own initiative.

It would be good to have a discussion about typical tracking selectors, and maybe we can agree on some way to quickly mark up the DOM for tracking purposes. For example, by adding attributes like `data-tracking="state"` or `data-tracking-value="123"` to elements.

adrianschmidt:
I'm going to have a look at what's been discussed so far, and if there's any non-hacky solution that I can find 🙂

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
adrianschmidt: > Don't get me wrong, I'm not really a fan of the current interface, I see problems/limitations with it, and to me it's quite unusual compared to popular React component libraries, for example.

The current interface of limel-progress-flow, limel-list, limel-select, all of them, or something else? Can you elaborate a little on what you don't like. If you don't feel like writing, we can huddle in Slack sometime maybe? I just want to get an idea of what you don't like and where you would want it to go instead, just to get an impression of whether I agree 😊

> We haven't identified the right direction 🙂

Thanks for pointing this out. I think this is often what's being missed when someone wonders why we don't "just add a quick fix" 👍

adrianschmidt: > 3\. Shouldn't the commit message be `feat`?

> 3\. I think it's a `fix` because it's just a small fix, I think we missed when we implemented it before, and it doesn't affect anything else

It's small, but it's still adding functionality, rather than fixing functionality, so I'll make it a `feat`.

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
```diff
@@ -49,26 +50,28 @@ export class ProgressFlowItem {
...
-                disabled={this.isDisabled()}
-            >
-                {this.renderIcon()}
-                <span class="text">{this.item.text}</span>
-                {this.renderDivider()}
-            </button>,
-            this.renderSecondaryText(),
-        ];
+        return (
+            <Host data-tracking-value={this.item?.value}>
```
adrianschmidt, line 54: > So using data-tracking, and data-tracking-*, I think is a very acceptable way forward 😊

I've updated this.

Unless @jgroth or @vicgeralds wishes to veto this, I will approve an merge it.

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
vicgeralds: I was mostly think about limel-list and the likes. The thing I don't like (or perhaps rather what I see as limiting, for good or bad), is that we often require items etc to be passed as data rather than as components. For example:

```tsx
<limel-list items={this.items} />
```

vs something like

```tsx
<limel-list>
    {this.items.map(item => <limel-list-item {...item} />)}
</limel-list>
```

It's much more flexible if you can render the item components yourself and pass them in slots. You can have your own (functional) component for rendering the item, adding any props or event handlers, and you can use state or data however you want. Adding some attribute like `data-tracking` to an item would also be easy here.

Imagine you want to add a click handler (or handle some other native event) to items and need to identify which item was clicked. You would just do:

```tsx
<limel-list-item onClick={() => this.handleClick(item)} {...item} />
```

but when passing items as data, that's not possible I think, without relying on some undocumented implementation details:

```tsx
<limel-list items={this.items} onClick={
    (clickEvent) => this.handleClick(this.items[/* figure out index of clickEvent.target */])
} />
```

Also, while there is a `change` event, it's not obvious how and if you can set a value on those components. I see that other components like limel-select and limel-input-field look a lot like [controlled components]
(https://reactjs.org/docs/forms.html#controlled-components) while a few others (limel-tab-panel probably best example) have a mutable "value" prop that I've found very confusing to work with.

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
vicgeralds: I would prefer if we did this in the parent component instead. Adding an attribute to `Host` is nearly as bad as having mutable props. Otherwise I think this ok. 🙂

jgroth: Yes I agree, and I think I mentioned that as well in an earlier discussion :smile:

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
adrianschmidt: > ```tsx
> <limel-list items={this.items} />
> ```
> vs something like
> ```tsx
> <limel-list>
>     {this.items.map(item => <limel-list-item {...item} />)}
> </limel-list>
> ```

Ah, yeah, that makes sense 👍

> ```tsx
> <limel-list-item onClick={() => this.handleClick(item)} {...item} />
> ```

While that's easy to write, it's not so nice for performance, since it's 1) setting up a new listener for each item, instead of just one listener for the whole list, and 2) creating a new function every time the list item is redrawn, because of the arrow function inside the template (as far as I understand).

For our custom events we always return the affected value, but you're right that it becomes difficult when it's vanilla click events… 🤔

jgroth: I think both approaches has its pros and cons. While something like this
> ```tsx
> <limel-list>
>     {this.items.map(item => <limel-list-item {...item} />)}
> </limel-list>
> ```
is super flexible, it also puts more effort on the consumer to get it right and can create more work for us to make sure stuff does not break or behave in a weird way. Something like the following does not make a lot of sense, but would be possible from the consumer point of view with a fully flexible API

```tsx
<limel-list>
  <limel-tab-bar>
    <limel-info-tile />
  </limel-tab-bar>
</limel-list>
```

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
adrianschmidt: ⚡

adrianschmidt (approved):

src/components/progress-flow/progress-flow-item/progress-flow-item.tsx
adrianschmidt: I've approved it myself now 🙈, but please have a look!

adrianschmidt: Ah, what the heck, it's a "private" change, so we can change or remove it without breaking anything other than the tracking, in which case we're not making anything worse than it already was without this. So I'm merging it now 😊

PR #1867, opened 2022-10-12
feat(datepicker): highlight today in datepickers

### LINKED ISSUE #1810: Highlight `today` in datepickers
## New feature motivation

In order to more easily navigate in a date picker, I consider it very helpful to highlight `today` in the picker. That makes it much easier to find which date is three days or a week from now, for example. I would say that this is a pretty standard feature in date pickers.

## New feature implementation

Flatpickr seems to support this:
image

I'm guessing that it's only a property that needs to be set. But I can't find it.

Aksa12:
Now the today date is highlighed in the date-picker component

Closes #1810


eketorp:
I took the liberty of setting @adrianschmidt as reviewer. Please change if you think someone else should review.

adrianschmidt (dismissed):

adrianschmidt:
Looks great to me 👍 Good work, and thank you!

image

@Kiarokh Approved? 5 is selected, 13 is today, and 23 is being hovered by the mouse.

Kiarokh:
Great job.! 👏 

I think we should have two distinct ways to visualize today and selected. In other words, user should be able to see that they have selected "today". Does it make sense?

So for example, a ring which is slightly larger than the green circle and has a distinctly different color would be nice.

This could be achieved by adding this code for example to `today` on line `185` instead:

```scss
        &.today {
            background: transparent;
            &:after {
                // opacity: 0.25;
                box-shadow: 0 0 0 0.125rem rgb(var(--contrast-100)),
                    0 0 0 0.25rem rgb(var(--color-orange-default));
            }
```
Note that I've removed the `opacity` too

preview 👇 
image

In this image, 13 is today, and it is selected.

Kiarokh:
I believe for now, "today" can have its own constant style all the time, no matter if it is selected or not. ;) 

We might add some more interactivity and animations in future. E.g. when today is selected, the orange ring can get larger. But we can take these details later. This PR would be perfect for merge, if this little tweak is applied.

Aksa12:
Hey @Kiarokh @adrianschmidt , should I work on this change? Also, is the inner circle (in light orange color) supposed to be added in the unselected mode for `today`? It's getting added through `box-shadow: 0 0 0 0.125rem rgb(var(--contrast-100)),
                    0 0 0 0.25rem rgb(var(--color-orange-default));`


Screenshot 2022-10-13 at 7 31 05 PM

Kiarokh:
Since this is your PR, yes @Aksa12 you can update the PR.

I suggest using `box-shadow`, as that does not risk layout-shifts or other visual defects which we may get using `border` for example. Box-shadow is drawn outside of the box model and it visually isn't a part of it.

Aksa12:
Updated @Kiarokh @eketorp. Please review. 
Screenshot 2022-10-13 at 8 09 47 PM

Screenshot 2022-10-13 at 8 10 15 PM

Kiarokh (approved):
looks great. thanks a lot!

adrianschmidt:
I squashed the second commit into the first, since we want atomic commits. This will be merged as soon as the checks pass 👍

PR #1875, opened 2022-10-13
feat(tab): allow string labels in badge

FredrikWallstrom:
We added the possibility to have string labels in badges in this release: https://github.com/Lundalogik/lime-elements/releases/tag/v36.1.0-next.1
However, it is not possible to use it in tabs because we didn't update the Tab interface.

Use case: 
Skärmavbild 2022-10-13 kl  15 51 41

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tab-bar/examples/tab-bar.tsx
+++ b/src/components/tab-bar/examples/tab-bar.tsx
@@ -60,6 +60,11 @@ export class TabBarExample {
             iconColor: 'var(--lime-deep-red)',
             badge: 4,
         },
+        {
+            id: 8,
+            text: 'Inception',
+            badge: 'NEW',
+        },
     ];
 
     public render() {
--- a/src/components/tab-bar/tab.types.ts
+++ b/src/components/tab-bar/tab.types.ts
@@ -27,5 +27,5 @@ export interface Tab {
     /**
      * Shows a badge within the tab with a specified label
      */
-    badge?: number;
+    badge?: number | string;
 }


anderssonjohan:
I don't want to have the final saying about this, but I think the change in itself is ok but the usage is something that we don't recommend. In the docs we want the user to consider:

> Avoid having long labels for tabs. 

...which easily will be violated by using string badges on tabs.

It could of course work for shorter things like "Messages (new)", which also depends on the display language in use.
There are also special occasions when I think this could work. Like when the number of open tabs in Chrome Mobile gets larger than 99, it instead renders ":-)". To do that you would need this change.

FredrikWallstrom:
Yea we were discussing the use case as well, not sure. We want to use it with the `new` text to visualize newly created tables (see image in description)

I think we are truncating the badge text after 5-6 chars, so the whole tab cant grow that much just because of the badge text.

src/components/tab-bar/examples/tab-bar.tsx
```diff
@@ -60,6 +60,11 @@ export class TabBarExample {
             iconColor: 'var(--lime-deep-red)',
             badge: 4,
         },
+        {
+            id: 8,
+            text: 'Inception',
+            badge: 'Movie',
```
Kiarokh, line 66: I think in a generic component library like Lime Elements, there is no point in limiting the consumers to only send numbers into the tabs. 

That’s rather a design question for the consumer. They need to decide if showing words in the badges is a good idea or not. We standing here shouldn’t care about that. 

But we should perhaps have a "recommendation" note about it. I just don't know currently what to say, to tell users when to use this or not.

So from my point of view, this is PR could be approved already.

Can you just change the word `Movie` to `NEW`. Then the coexistence of those digits and this word would make more sense. It will otherwise be a strange example.

FredrikWallstrom: ⚡

Kiarokh (approved):

PR #1876, opened 2022-10-17
Scan code with SonarCloud on push

anderssonjohan:
Initial attempt to integrate sonarcloud scanner with the PRs + next branch on Github with project on SonarCloud.io:
https://sonarcloud.io/project/configuration?id=Lundalogik_lime-elements

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


adrianschmidt:
Is it possible to disable SonarCloud's posting of comments on every single PR? I'm already getting an annoying amount of emails from GitHub, and these don't really add anything, _especially_ not when everything is green… 😖

To me, it's enough if the check fails if there's a problem. It's ok, but unnecessary imo, if it also posts a comment when something fails.

.github/workflows/sonarcloud.yml
```diff
@@ -0,0 +1,20 @@
+name: SonarCloud Scan
+on:
+  push:
+    branches:
+      - next
+  pull_request:
+    types: [opened, synchronize, reopened]
```
adrianschmidt, line 7: This seems to be the default if `types` are not specified. Is there any particular reason for setting this explicitly anyway?

image

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

.github/workflows/sonarcloud.yml
```diff
@@ -0,0 +1,20 @@
+name: SonarCloud Scan
+on:
+  push:
+    branches:
+      - next
```
adrianschmidt, line 5: This means that it only runs on PRs targeting `next`, right? Seems reasonable 👍

.github/workflows/sonarcloud.yml
anderssonjohan: Good to know: The whole workflow file is generated by Sonarcloud for you when setting up the CI mode. I've just copy pasted it. It doesn't hurt to be explicit with the event types!

.github/workflows/sonarcloud.yml
anderssonjohan: Yes, `next` is the main branch in sonarcloud for lime-elements: https://sonarcloud.io/project/branches_list?id=Lundalogik_lime-elements

I configured the pattern for long-lived branches to next and main. I suspect main should appear in the list above after the first scanned commit is merged to main, but I'm not entirely sure how it will work. Maybe we need to add main to the list here to issue another scan, maybe not.

adrianschmidt: I'm not sure we need it to scan `main` too. So far, `main` has exactly the same commits as `next`, it's just trailing in time. Or, rather, it gets updated to the current state of `next` at semi-regular intervals. I have thought a bit about decoupling them more, but that's the current state anyway. So right now, any warnings on `main` would just be warnings we've already had on `next`, and we wouldn't fix them on `main` by any other means than fixing them on `next`, and then merging `next` to `main`.

adrianschmidt (approved):

.github/workflows/sonarcloud.yml
adrianschmidt: Oh, no, this isn't running on PRs targeting `next`, it's running on `next` once a PR has been merged.

That's perhaps a little late…? 🤔

It does limit the number of runs it has to do though, since if we run it on PRs, it will run every time a PR is updated with new code changes. Let's leave it like it is for now, and see if we feel any need to run it on PRs.

PR #1900, opened 2022-10-21
feat(dashboard-shortcut): add `limel-dashboard-shortcut` component

LucyChyzhova:
Fix: https://github.com/Lundalogik/crm-feature/issues/3025

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/shortcut/shortcut.e2e.ts
+++ b/src/components/shortcut/shortcut.e2e.ts
@@ -1,0 +1,145 @@
+import { E2EElement, E2EPage, newE2EPage } from '@stencil/core/testing';
+
+describe('limel-shortcut', () => {
+    let page: E2EPage;
+    let limelShortcut: E2EElement;
+    let link: E2EElement;
+    let label: E2EElement;
+    let badge: E2EElement;
+
+    describe('with a label', () => {
+        beforeEach(async () => {
+            page = await createPage(`
+                <limel-shortcut label="iSpiffy"></limel-shortcut>
+            `);
+            limelShortcut = await page.find('limel-shortcut');
+            link = await page.find('limel-shortcut >>> a');
+            label = await page.find('limel-shortcut >>> span');
+        });
+        it('displays the correct label', () => {
+            expect(label).toEqualText('iSpiffy');
+        });
+        it('includes the label in `aria-label`', () => {
+            expect(link).toEqualAttribute('aria-label', 'iSpiffy');
+        });
+
+        describe('when changing the label', () => {
+            beforeEach(async () => {
+                limelShortcut.setProperty('label', 'imTired');
+                await page.waitForChanges();
+            });
+            it('displays the new label', async () => {
+                expect(label).toEqualText('imTired');
+            });
+            it('updates `aria-label`', () => {
+                expect(link).toEqualAttribute('aria-label', 'imTired');
+            });
+        });
+    });
+
+    describe('with `linkTitle`', () => {
+        beforeEach(async () => {
+            page = await createPage(`
+                <limel-shortcut link-title="Open the iSpiffy app"></limel-shortcut>
+            `);
+            limelShortcut = await page.find('limel-shortcut');
+            link = await page.find('limel-shortcut >>> a');
+            label = await page.find('limel-shortcut >>> span');
+        });
+        it('sets the `title` attribute on the link', () => {
+            expect(link).toEqualAttribute('title', 'Open the iSpiffy app');
+        });
+        it('includes the label in `aria-label`', () => {
+            expect(link).toEqualAttribute('aria-label', 'Open the iSpiffy app');
+        });
+
+        describe('when changing the `linkTitle`', () => {
+            beforeEach(async () => {
+                limelShortcut.setProperty('linkTitle', 'I need a cup of tea');
+                await page.waitForChanges();
+            });
+            it('uses the new title', async () => {
+                expect(link).toEqualAttribute('title', 'I need a cup of tea');
+            });
+            it('updates `aria-label`', () => {
+                expect(link).toEqualAttribute(
+                    'aria-label',
+                    'I need a cup of tea'
+                );
+            });
+        });
+    });
+
+    describe('with a label and a linkTitle', () => {
+        beforeEach(async () => {
+            page = await createPage(`
+                <limel-shortcut label="iSpiffy" link-title="Open the iSpiffy app"></limel-shortcut>
+            `);
+            limelShortcut = await page.find('limel-shortcut');
+            link = await page.find('limel-shortcut >>> a');
+            label = await page.find('limel-shortcut >>> span');
+        });
+        it('displays the correct label', () => {
+            expect(label).toEqualText('iSpiffy');
+        });
+        it('uses both the label and the linkTitle to construct the `aria-label`', () => {
+            expect(link).toEqualAttribute(
+                'aria-label',
+                'iSpiffy. Open the iSpiffy app'
+            );
+        });
+
+        describe('when changing the label', () => {
+            beforeEach(async () => {
+                limelShortcut.setProperty('label', 'imTired');
+                await page.waitForChanges();
+            });
+            it('updates `aria-label`', () => {
+                expect(link).toEqualAttribute(
+                    'aria-label',
+                    'imTired. Open the iSpiffy app'
+                );
+            });
+        });
+
+        describe('when changing the linkTitle', () => {
+            beforeEach(async () => {
+                limelShortcut.setProperty('linkTitle', 'I need a cup of tea');
+                await page.waitForChanges();
+            });
+            it('updates `aria-label`', () => {
+                expect(link).toEqualAttribute(
+                    'aria-label',
+                    'iSpiffy. I need a cup of tea'
+                );
+            });
+        });
+    });
+
+    describe('with a value for badge', () => {
+        beforeEach(async () => {
+            page = await createPage(`
+                <limel-shortcut badge="NEW"></limel-shortcut>
+            `);
+            limelShortcut = await page.find('limel-shortcut');
+            badge = await page.find('limel-shortcut >>> limel-badge');
+        });
+        it('displays a badge with the correct value', () => {
+            expect(badge).toEqualAttribute('label', 'NEW');
+        });
+
+        describe('when changing the badge value', () => {
+            beforeEach(async () => {
+                limelShortcut.setProperty('badge', '3');
+                await page.waitForChanges();
+            });
+            it('displays the new badge value', async () => {
+                expect(badge).toEqualAttribute('label', '3');
+            });
+        });
+    });
+});
+
+async function createPage(content) {
+    return newE2EPage({ html: content });
+}
--- a/src/components/shortcut/shortcut.tsx
+++ b/src/components/shortcut/shortcut.tsx
@@ -1,0 +1,119 @@
+import { Component, Prop, h } from '@stencil/core';
+
+/**
+ * This component can be used on places such as a start page or a dashboard.
+ * Clicking on the component should navigate the user to a new screen,
+ * to which you need to provide a URL, using the `href` property.
+ *
+ * By default, this navigation will happen within the same browser tab.
+ * However, it is possible to override that behavior, using the `target` property.
+ *
+ * @exampleComponent limel-example-shortcut
+ * @exampleComponent limel-example-shortcut-notification
+ * @exampleComponent limel-example-shortcut-styling
+ * @exampleComponent limel-example-shortcut-with-click-handler
+ */
+
+@Component({
+    tag: 'limel-shortcut',
+    shadow: true,
+    styleUrl: 'shortcut.scss',
+})
+export class Shortcut {
+    /**
+     * Name of icon for the shortcut.
+     */
+    @Prop({ reflect: true })
+    public icon: string;
+
+    /**
+     * The text to show below the shortcut. Long label will be truncated.
+     */
+    @Prop({ reflect: true })
+    public label?: string = null;
+
+    /**
+     * The `title` tag of the hyperlink, which can be used to
+     * provide additional information about the link.
+     * It improves accessibility both for sighted users
+     * and users with assistive technologies.
+     */
+    @Prop({ reflect: true })
+    public linkTitle?: string = null;
+
+    /**
+     * Set to `true` if shortcut is disabled.
+     */
+    @Prop({ reflect: true })
+    public disabled?: boolean = false;
+
+    /**
+     * The url that the shortcut leads to.
+     */
+    @Prop({ reflect: true })
+    public href?: string = null;
+
+    /**
+     * Where to load the linked URL, as the name for a browsing context:
+     * - `_self`: in the current browsing context. (Default)
+     * - `_blank`: in a new tab.
+     * - `_parent`: in the parent browsing context of the current one.
+     * If no parent, behaves as `_self`.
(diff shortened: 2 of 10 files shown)


src/components/dashboard-shortcut/dashboard-shortcut.tsx
```diff
@@ -0,0 +1,55 @@
...
+                tabindex="0"
+            >
+                <limel-icon name={this.icon} />
+            </a>,
+            <span class="label">{this.label}</span>,
+        ];
+    }
+
+    private handleClick = (event: MouseEvent) => {
+        event.stopPropagation();
```
adrianschmidt, line 52: We should not stop this click.

adrianschmidt: The easiest way to handle it is probably just to let the native click through. If there's an `href`, the browser will automatically activate that, unless the consumer handles the click event differently, and if there isn't an `href`, the consumer has to handle the click event somehow, or nothing at all will happen 🙂

adrianschmidt: We should document the event though, so the consumer knows to expect it.

Kiarokh:
Sorry this is still a draft PR.

src/components/dashboard-shortcut/dashboard-shortcut.tsx
```diff
@@ -0,0 +1,68 @@
...
+     * If no parent, behaves as `_self`.
+     * - `_top`: the topmost browsing context (the "highest" context
+     * that's an ancestor of the current one). If no ancestors, behaves as `_self`.
+     */
+    @Prop({ reflect: true })
+    public target: '_self' | '_blank' | '_parent' | '_top' = '_self';
+
+    public render() {
+        return [
+            <a
```
Kiarokh, line 52: First I thought we should have  `rel="noopener"`: But maybe we don't? 
https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noopener

src/components/dashboard-shortcut/dashboard-shortcut.tsx
Kiarokh: Thanks Adrian. How do we make sure that the onClick triggers when the `a` is clicked, not when the whole component is clicked?

adrianschmidt: In this specific case, since we're talking about "dashboard shortcut", I'll start by answering your question with another question: shouldn't it trigger regardless of where on the component you click? 🤔

But assuming you have a good answer for that, or just in the more general case: that's a good point! 👍

An alternative would be to have an `interact` event on this component, catch the click event from the link, and emit our custom interact event instead. But we'd have to make sure we included the original click event in the details for the interact event, because the consumer might provide a href for the link, and still wish to preventDefault on the click and handle the navigation themselves. For example, we would like to always provide a url to all our links, so the user can right-click or cmd-click to open the link in a new window, but if we detect a straight left-click with no modifier keys active, we might want to preventDefault on it, and trigger the navigation through the router, because if we just let the link trigger the navigation, it makes the app load again before you get to the link's destination 🙁

adrianschmidt: I'm happy to "consult" on this component, as always 😊 But now I'll try to log off to have me a spot of weekend 😄

# 👋 ❤️

src/components/dashboard-shortcut/dashboard-shortcut.tsx
adrianschmidt: I guess we could… I didn't know about that attribute before, so I haven't yet formed much of an opinion…

Kiarokh: it was a security thing before, but they say this:
> Note: Setting target="_blank" on <a> elements now implicitly provides the same rel behavior as setting rel="noopener" which does not set window.opener. See [browser compatibility](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#browser_compatibility) for support status.

I don't know how "new" this is. Can we take it for granted that everyone's browsers will have this security feature? Or should we add it to be safe?

src/components/dashboard-shortcut/dashboard-shortcut.tsx
Kiarokh: Thanks a lot! I think I'll need your help next week. 
But quick answer: The interactive part of this component is the link part (the `<a>`) so I definitely think we should make keyboard or pointer actions only work on that, not the entire component's surface.

src/components/dashboard-shortcut/dashboard-shortcut.tsx
adrianschmidt: Ah, I've read up on it on [the page MDN linked to](https://mathiasbynens.github.io/rel-noopener/).

Modern browsers have this patched already, and it already wasn't a problem on cross-domain links, only on links to the same domain. It's possible there could be some situation where a user could make some sort of attack using this, but it seems fairly unlikely at first glance at least. At the very least, I think it could only be used by a user that has some kind of access to the system already, for example if an attacker has compromised the account of a normal user, and is trying to craft an attack on an admin user to gain access to that account. But the user would need to be able to upload or generate some content that would execute a script the user has written, and we really shouldn't be allowing that anywhere. If we do, that's an XSS vulnerability already, and the lack of `rel="noopener"` would "only" be making the existing vulnerability a bit worse. _If I understand things correctly._

src/components/dashboard-shortcut/dashboard-shortcut.tsx
adrianschmidt: But it's basically meant to be an app icon on the start page, like on the home screen of Android or iOS, right?

image

You wouldn't want to have to click or tap on the text to trigger the link, right?

Although, my own example shows that there could be situations where a click should not trigger the link, like if you've long-pressed an icon on iOS and entered the edit mode where you can tap the x-button to remove a shortcut… So yes, I agree that we need to be able to not trigger the link from just _any_ click on the component. But still, in the _normal_ case, surely the whole thing should be a click target and trigger the corresponding "action", whatever it may be?

src/components/dashboard-shortcut/dashboard-shortcut.scss
```diff
@@ -0,0 +1,80 @@
+/**
+* @prop --dashboard-shortcut-border-radius: defines the radius of corners of the shortcut. Defaults to `1rem`
+* @prop --dashboard-shortcut-icon-color: defines the fill color of the shortcut icon. Defaults to `--contrast-1000`
+* @prop --dashboard-shortcut-label-color: defines the color of the shortcut label. Defaults to `--contrast-11000`
+* @prop --dashboard-shortcut-background-color: defines the backgrounds color of the shortcut icon. Defaults to `--contrast-100`
+* @prop --dashboard-shortcut-badge-text-color: Text color of the notification badge. Defaults to `--color-white`
+* @prop --dashboard-shortcut-badge-background-color: Background color of the notification badge. Defaults to `--color-red-default`
```
Kiarokh, line 7: Is this overkill?
I mean, it would already be still possible to style the badge, using `limel-badge`'s own CSS properties. But the consumer should be "one of us" to know that they are available. Or have a deep understanding of how css variables can penetrate the shadow doms. 

On the other hand, I have never seen the notification badges, in any color in other than white text on red background. Do we want people to change their colors?

Maybe we want, if the color is trying to attract attention for some reason. I can't really decide.

adrianschmidt: > Do we want people to change their colors?

I can see arguments for either option, but I prefer not restricting possibilities unless necessary, and I prefer not adding code unless necessary. When considering adding a restriction that would require extra code, there better be a _very_ good reason, imho 😝

I was about to suggest that rather than adding this extra layer of CSS variables, we could just document the variables available from `limel-badge` here too, but then I realised that's exactly the kind of "leaking" if information about the internal implementation we want to avoid, as it becomes a problem if and when we want to change the internal implementation. So I say we keep these CSS variables! 👍

src/components/dashboard-shortcut/dashboard-shortcut.tsx
Kiarokh: @adrianschmidt I think we're ready for review, except the part for the `onClick`. What do you think is best to do here?

adrianschmidt: Nice! I'll have a look at it when I'm done with some stuff I'm working on at the moment 😊

Kiarokh: Cool. Meanwhile, I'll use this component in crm-components instead of what we have there in `action-button` and connect it the other PR to this issue.

Kiarokh: ping me before you start, because until then I'll be forcepushing my fixups

src/components/dashboard-shortcut/dashboard-shortcut.tsx
```diff
@@ -0,0 +1,110 @@
...
+ * By default, this navigation will happen within the same browser tab.
+ * However, it is possible to override that behavior, using the `target` property.
+ *
+ * @exampleComponent limel-example-dashboard-shortcut
+ * @exampleComponent limel-example-dashboard-shortcut-notification
+ * @exampleComponent limel-example-dashboard-shortcut-styling
+ */
+
+@Component({
+    tag: 'limel-dashboard-shortcut',
```
Kiarokh, line 17: We decided to call this "shortcut" not "dashboard shortcut"!

If we wanna keep the word "dashboard", the other component will also need to be called "dashboard info tile".

Dashboard is where this component is supposed to be used, not what the component is. Right?

adrianschmidt: I agree with dropping the `dashboard`. But I'm not so sure I like `limel-shortcut`. That seems a bit too generic to describe this component. How about something like `limel-app-shortcut`, `limel-app-button`…?

src/components/dashboard-shortcut/dashboard-shortcut.tsx
adrianschmidt: I'm starting on it now 👍

src/components/shortcut/shortcut.tsx
```diff
@@ -0,0 +1,105 @@
...
+    public href?: string;
+
+    /**
+     * Where to load the linked URL, as the name for a browsing context:
+     * - `_self`: in the current browsing context. (Default)
+     * - `_blank`: in a new tab.
+     * - `_parent`: in the parent browsing context of the current one.
+     * If no parent, behaves as `_self`.
+     * - `_top`: the topmost browsing context (the "highest" context
+     * that's an ancestor of the current one). If no ancestors, behaves as `_self`.
```
adrianschmidt, line 63: We could just link to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target instead of copying the explanation.

adrianschmidt:
```suggestion
     * See [documentation for target](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target) for explanation of possible values.
```

src/components/dashboard-shortcut/dashboard-shortcut.tsx
Kiarokh: Well, my choice of name comes from this very old memory 😄 👇 
image

app-shortcut 🤔 this could be really shortcut to anything, app, website, "ugly modal", a pane/view/bookmark in the app. I don't know. I feel shortcut is OK :)

adrianschmidt: Yeah, I feel like `limel-shortcut` has grown on me 👍

src/components/shortcut/shortcut.tsx
adrianschmidt: (I'm _requesting_ a change here, I just wanted to add a suggestion to indicate what I think we could write, if we do want to change it. Feel free to keep it as is too…)

adrianschmidt: Sorry, I had missed a pretty important "not" in my last comment. It should read "I'm not requesting a change" 🤦‍♂️

src/components/shortcut/shortcut.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * If specified, will display a notification badge
+     * on the shortcut.
+     */
+    @Prop({ reflect: true })
+    public badge?: number | string;
+
+    public render() {
+        const extendedAriaLabel = this.label + '. ' + this.linkTitle;
```
adrianschmidt, line 76: On line 95, we have an `if (this.label)`, indicating that we expect it could be empty. But here, we just assume that it's set, otherwise the label will begin with `. `.

I've added a few tests for how I would expect this to behave. I haven't implemented it, so the tests are currently failing. I was thinking perhaps you could do the implementation @LucyChyzhova? 😊

LucyChyzhova: Thank Adrian! Sure I will add this 🤗

src/components/shortcut/shortcut.tsx
Kiarokh: I think it's easier for the consumer to read a brief thing here. But we can certainly link too.

src/components/shortcut/shortcut.e2e.ts
```diff
@@ -0,0 +1,145 @@
...
+            it('updates `aria-label`', () => {
+                expect(link).toEqualAttribute('aria-label', 'imTired');
+            });
+        });
+    });
+
+    describe('with `linkTitle`', () => {
+        beforeEach(async () => {
+            page = await createPage(`
+                <limel-shortcut link-title="Open the iSpiffy app"></limel-shortcut>
```
adrianschmidt, line 43: @LucyChyzhova I pushed a fix for the tests. I had written `linkTitle="…"` here, instead of `link-title="…"` 😄 That's why the tests where failing. Sorry about that!

src/components/shortcut/shortcut.e2e.ts
```diff
@@ -0,0 +1,145 @@
...
+                    'I need a cup of tea'
+                );
+            });
+        });
+    });
+
+    describe('with a label and a linkTitle', () => {
+        beforeEach(async () => {
+            page = await createPage(`
+                <limel-shortcut label="iSpiffy" link-title="Open the iSpiffy app"></limel-shortcut>
```
adrianschmidt, line 76: (I had made the same error here.)

src/components/shortcut/shortcut.tsx
```diff
@@ -0,0 +1,125 @@
...
+    public target: '_self' | '_blank' | '_parent' | '_top' = '_self';
+
+    /**
+     * If specified, will display a notification badge
+     * on the shortcut.
+     */
+    @Prop({ reflect: true })
+    public badge?: number | string;
+
+    public getAreaLabel(label, linkTitle) {
```
adrianschmidt, line 75: This should be private.

adrianschmidt (approved):

Kiarokh:
yeay! great job guys!

PR #1911, opened 2022-10-27
feat(info-tile): add `limel-info-tile` component

Kiarokh:
fix: https://github.com/Lundalogik/crm-feature/issues/3025

## testing in the web client
Add `limel-info-tile` on the start page as a widget, with the following config, and try different sizes:

```json
{
    "icon": "full_of_shit",
    "label": "Money we earned through some crazy deals this year",
    "value": "20,000",
    "prefix": "earned",
    "suffix": "SEK",
    "href": "https://www.google.com",
    "target": "_blank",
    "progressValue": "24"
}
````
````json
{
    "icon": "poo",
    "label": "Such a wonderful year",
    "value": "4 432",
    "prefix": "money",
    "suffix": "SEK",
    "href": "https://www.google.com",
    "target": "_blank",
    "badge": "2345",
}
````

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/info-tile/examples/info-tile-loading.tsx
+++ b/src/components/info-tile/examples/info-tile-loading.tsx
@@ -1,0 +1,58 @@
+import { Component, h, State } from '@stencil/core';
+
+/**
+ * Loading state
+ *
+ * Sometimes the value needs to be calculated, updated, or fetched
+ * through a process that requires some time. In such cases, it is
+ * a great idea to let the users know that the data is being updated.
+ *
+ * To do so, set the `loading` property to `true`. The component will then
+ * show an indeterminate progressbar indicating the data is being updated,
+ * while the older value is still being displayed.
+ *
+ * :::note
+ * Note that this does _not_ disable the link, and most probably you
+ * do not need it to be disabled either.
+ * If the link should be disabled while loading, the
+ * `disabled` property should be set to `true` as well.
+ * :::
+ */
+@Component({
+    tag: 'limel-example-info-tile-loading',
+    shadow: true,
+    styleUrl: 'info-tile-loading.scss',
+})
+export class InfoTileLoadingExample {
+    @State()
+    public loading = false;
+
+    public render() {
+        const link = {
+            href: 'https://duckduckgo.com/?q=weather',
+            title: 'Click to see real-time weather forecast',
+        };
+
+        return [
+            <limel-info-tile
+                icon="partly_cloudy_rain"
+                label="Partly cloudy with a risk of rain"
+                prefix="temp"
+                value="23"
+                suffix="°C"
+                link={link}
+                loading={this.loading}
+            />,
+            <limel-checkbox
+                label="Loading"
+                checked={this.loading}
+                onChange={this.setLoading}
+            />,
+        ];
+    }
+
+    private setLoading = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.loading = event.detail;
+    };
+}
--- a/src/components/info-tile/examples/info-tile-progress.tsx
+++ b/src/components/info-tile/examples/info-tile-progress.tsx
@@ -1,0 +1,62 @@
+import { Component, h, State } from '@stencil/core';
+import { InfoTileProgress } from '../info-tile.types';
+
+/**
+ * Displaying a progress bar
+ *
+ * By defining a numeric `progressValue`, you can display
+ * a circular progress bar to visualize more data on the component.
+ * This can for instance help illustrate how much of a
+ * set goal has been reached, which together with the `value` will help users
+ * get a better overview of the provided data.
+ *
+ * When the circular progress is shown, that would become the primary
+ * illustrative element on the component,
+ * which means the icon will be rendered smaller, only as a supportive
+ * contextual visual element.
+ *
+ * :::tip
+ * It is possible to customize the progress bar's suffix, but it is
+ * set to display the percentage sign (**%**) by default.
+ * :::
+ *
+ */
+@Component({
+    tag: 'limel-example-info-tile-progress',
+    shadow: true,
+    styleUrl: 'info-tile-progress.scss',
+})
+export class InfoTileProgressExample {
+    @State()
+    private progress: InfoTileProgress = {
+        value: 76,
+        prefix: '↑',
+    };
+
+    public render() {
+        return [
+            <limel-info-tile
+                label="Won deals this month"
+                icon="money"
+                prefix="Total value"
+                value="70,659"
+                suffix="EUR"
+                link={{ href: '#' }}
+                progress={this.progress}
+            />,
+            <limel-input-field
+                label="Progress value"
+                type="number"
+                value={`${this.progress.value}`}
+                onChange={this.handleChange}
+            />,
+        ];
+    }
+
+    private handleChange = (event) => {
+        this.progress = {
+            ...this.progress,
+            value: +event.detail,
+        };
+    };
+}
--- a/src/components/info-tile/info-tile.scss
+++ b/src/components/info-tile/info-tile.scss
@@ -1,0 +1,344 @@
+/**
+* @prop --info-tile-border-radius: defines the radius of corners of the info-tile. Defaults to `1rem`
+* @prop --info-tile-icon-color: defines the fill color of the info-tile icon. Defaults to `--contrast-1000`
+* @prop --info-tile-text-color: defines the color of the info-tile label. Defaults to `--contrast-1100`
+* @prop --info-tile-background-color: defines the backgrounds color of the info-tile icon. Defaults to `--contrast-100`
+* @prop --info-tile-badge-text-color: Text color of the notification badge. Defaults to `--color-white`
+* @prop --info-tile-badge-background-color: Background color of the notification badge. Defaults to `--color-red-default`
+* @prop --info-tile-progress-fill-color: Determines the color of the progressed section. Defaults to `--lime-primary-color`.
+* @prop --info-tile-progress-background-color: Determines the background color of the central section of the progress bar. Defaults to `--info-tile-background-color`.
+* @prop --info-tile-progress-suffix-color: Determines the color of the progress prefix. Defaults to `--contrast-1000`.
+ * @prop --info-tile-progress-text-color: Determines the color of the progress value. Defaults to `--info-tile-text-color`.
+ * @prop --info-tile-progress-prefix-color: Determines the color of the progress suffix. Defaults to `--contrast-1000`.
+*/
+
+@use '../../style/mixins';
+
+:host(limel-info-tile) {
+    --badge-text-color: var(
+        --info-tile-badge-text-color,
+        rgb(var(--color-white))
+    );
+    --badge-background-color: var(
+        --info-tile-badge-background-color,
+        rgb(var(--color-red-default))
+    );
+
+    --circular-progress-text-color: var(
+        --info-tile-progress-text-color,
+        var(--info-tile-text-color)
+    );
+    --circular-progress-suffix-color: var(--info-tile-progress-suffix-color);
+    --circular-progress-prefix-color: var(--info-tile-progress-prefix-color);
+    --circular-progress-track-color: rgb(var(--contrast-800), 0.3);
+    --circular-progress-fill-color: var(--info-tile-progress-fill-color);
+    --circular-progress-background-color: var(
+        --info-tile-progress-background-color,
+        var(--info-tile-background-color)
+    );
+
+    --label-min-size: 0.75rem;
+    --label-preferred-size: 6cqw;
+    --label-max-size: 1rem;
+
+    --value-min-size: 1rem;
+    --value-preferred-size: 20cqw;
+    --value-max-size: 4rem;
+
+    --suffix-prefix-min-size: 0.75rem;
+    --suffix-prefix-preferred-size: 8cqw;
+    --suffix-prefix-max-size: 1.5rem;
+
+    --icon-min-size: 2rem;
+    --icon-preferred-size: 60cqh;
+    --icon-max-size: calc(100cqw - 0.5rem);
+
+    isolation: isolate;
+    container-type: size;
+    position: relative;
+    display: flex;
+
+    width: 100%;
+    height: 100%;
+
+    * {
+        box-sizing: border-box;
+    }
+}
+
+:host(limel-info-tile[disabled]) {
+    a {
+        opacity: 0.5;
+        cursor: not-allowed;
+    }
+}
+
+a {
+    all: unset;
+    overflow: hidden;
+
+    display: flex;
+    flex-direction: column;
+    justify-content: flex-end;
+    align-items: flex-start;
+
+    height: 100%;
+    width: 100%;
+    flex-grow: 1;
+
+    padding: 0.25rem 1rem 0.5rem 1rem;
+    border-radius: var(--info-tile-border-radius, 1rem);
+    background-color: var(
+        --info-tile-background-color,
+        rgb(var(--contrast-100))
+    );
+
+    &.is-clickable {
+        @include mixins.is-elevated-clickable;
+        @include mixins.visualize-keyboard-focus;
+        cursor: pointer;
+    }
+}
+
+.icon {
+    position: absolute;
+    top: 0.5rem;
+    right: 0.75rem;
+    padding: 0.25rem;
+
+    display: flex;
+    justify-content: center;
+
+    color: var(--info-tile-icon-color, rgb(var(--contrast-1000)));
+
+    border-radius: 0;
+    height: clamp(
+        var(--icon-min-size),
+        var(--icon-preferred-size),
+        var(--icon-max-size)
+    );
+
+    @supports not (container-type: size) {
+        height: min(60%, var(--icon-max-size));
+    }
+
+    .has-circular-progress & {
+        top: unset;
+        bottom: 0.5rem;
+        --icon-min-size: 1.5rem;
+        --icon-preferred-size: 20cqh;
+    }
+}
+
+.progress {
+    position: absolute;
+    top: 0.75rem;
+    right: 0.75rem;
+
+    --circular-progress-size: min(
+        var(--icon-preferred-size),
+        var(--icon-max-size)
+    );
+}
+
+.label {
+    z-index: 1;
+    color: var(--info-tile-text-color, rgb(var(--contrast-1100)));
+
+    line-height: normal;
+    // white-space: normal;
+    // overflow: hidden;
+    // display: -webkit-box;
+    // -webkit-line-clamp: 2; // max number of lines before truncation
+    // -webkit-box-orient: vertical;
+    font-size: clamp(
+        var(--label-min-size),
+        var(--label-preferred-size),
+        var(--label-max-size)
+    );
+    @supports not (container-type: size) {
+        font-size: 0.875rem;
+    }
+}
+
+limel-badge {
+    position: absolute;
+    top: -0.5rem;
+    right: -0.5rem;
+}
+
+limel-linear-progress {
+    --lime-primary-color: var(--info-tile-text-color);
+    position: absolute;
+    inset: auto 0 0 0;
+    border-radius: 1rem;
+    overflow: hidden;
+}
+
+.value-group {
+    position: relative;
+    z-index: 1;
+    display: flex;
+    flex-direction: column;
+    color: var(--info-tile-text-color, rgb(var(--contrast-1100)));
+}
+
+.value-and-suffix,
+.label {
+    text-shadow: 0 0 0.5rem
+            var(--info-tile-background-color, rgb(var(--contrast-100))),
+        0 0 0.25rem var(--info-tile-background-color, rgb(var(--contrast-100)));
+}
+
+.value-and-suffix {
+    display: flex;
+}
+
+.prefix,
+.suffix {
+    font-size: clamp(
+        var(--suffix-prefix-min-size),
+        var(--suffix-prefix-preferred-size),
+        var(--suffix-prefix-max-size)
+    );
+    opacity: 0.7;
+
+    @supports not (container-type: size) {
+        font-size: 0.75rem;
+    }
+}
+
+.prefix {
+    align-self: flex-start;
+    line-height: normal;
+    transform: translateY(40%);
+}
+
+.value {
+    transition: opacity 0.2s ease,
+        transform 0.2s cubic-bezier(0.25, 0.46, 0.45, 0.95);
+    transform-origin: left;
+    transform: translate3d(0, 0, 0) scale(1);
+
+    @include mixins.truncate-text;
+    font-weight: bold;
+    line-height: normal;
+
+    font-size: clamp(
+        var(--value-min-size),
+        var(--value-preferred-size),
+        var(--value-max-size)
+    );
+    @supports not (container-type: size) {
+        font-size: 2.5rem;
+    }
+
+    :host(limel-info-tile[loading]) & {
+        opacity: 0.3;
+        transform: translate3d(0, 0, 0) scale(0.9);
+    }
+}
+
+.suffix {
+    transform: translateY(10%);
+}
+
+$xs: 8rem; //128px:
+$s: 18.75rem; //300px
+$m: 40.5rem; //648px
+$l: 62.5rem; //1000px
+
+@container (width < #{$xs}) {
+    .progress {
+        top: 0.25rem;
+        right: 0.25rem;
+    }
+    a {
+        padding: 0.375rem;
+        gap: 0.125rem;
+    }
+}
+
+@container (width < #{$s}) {
+    .progress {
+        top: 0.5rem;
+        right: 0.5rem;
+    }
+    .icon {
+        top: 0.25rem;
+        right: 0.5rem;
+        .has-circular-progress & {
+            right: 0.25rem;
+            bottom: 0.25rem;
+        }
+    }
+}
+
+@container (width < #{$m}) {
+    .value {
+        &.ch-1,
+        &.ch-2,
+        &.ch-3,
+        &.ch-4 {
+            --value-preferred-size: 20cqw;
+        }
+        &.ch-5 {
+            --value-preferred-size: 18cqw;
+        }
+        &.ch-6 {
+            --value-preferred-size: 17cqw;
+        }
+        &.ch-7 {
+            --value-preferred-size: 16cqw;
+        }
+        &.ch-8 {
+            --value-preferred-size: 15cqw;
+        }
+        &.ch-9 {
+            --value-preferred-size: 14cqw;
+        }
+        --value-preferred-size: 13cqw;
+    }
+}
+
+@container (height > #{$xs}) {
+    a {
+        padding-top: 0.75rem;
+        padding-bottom: 1rem;
+    }
+}
+
+@container (height < #{$xs}) and (width > #{$xs}) {
+    .value {
+        --value-preferred-size: 32cqh !important;
+    }
+    .suffix,
+    .prefix {
+        --suffix-prefix-preferred-size: 16cqh !important;
+    }
+}
+
+@container (height > #{$s}) {
+    .progress,
+    .icon {
+        position: relative;
+        top: unset;
+        right: unset;
+    }
+    a {
+        align-items: center;
+        justify-content: center;
+    }
+    .label {
+        text-align: center;
+        // -webkit-line-clamp: 3;
+    }
+    .has-circular-progress {
+        .icon {
+            position: absolute;
+            top: 0.5rem;
+            right: 0.5rem;
+            --icon-max-size: 3rem;
+        }
+    }
+}
--- a/src/components/info-tile/info-tile.tsx
+++ b/src/components/info-tile/info-tile.tsx
@@ -1,0 +1,222 @@
+import { Component, Prop, h } from '@stencil/core';
+import { InfoTileProgress } from './info-tile.types';
+import { Link } from '@limetech/lime-elements';
+
+/**
+ * This component can be used on places such as a start page or a dashboard.
+ * It offers features for visualizing aggregated data along with supplementary
+ * information.
+ *
+ * If clicking on the component should navigate the user to
+ * a new screen or web page, you need to provide a URL,
+ * using the `link` property.
+ *
+ * @exampleComponent limel-example-info-tile
+ * @exampleComponent limel-example-info-tile-badge
+ * @exampleComponent limel-example-info-tile-progress
+ * @exampleComponent limel-example-info-tile-loading
+ * @exampleComponent limel-example-info-tile-styling
+ */
+@Component({
+    tag: 'limel-info-tile',
+    shadow: true,
+    styleUrl: 'info-tile.scss',
+})
+export class InfoTile {
+    /**
+     * A piece of text or number that is the main piece of information
+     * which the component is intended to visualize.
+     */
+    @Prop({ reflect: true })
+    public value: number | string;
+
+    /**
+     * Name of icon for the info tile.
+     */
+    @Prop()
+    public icon?: string;
+
+    /**
+     * The text to show below the info tile. Long labels will be truncated.
+     */
+    @Prop({ reflect: true })
+    public label?: string = null;
+
+    /**
+     * A string of text that is visually placed before the value.
+     */
+    @Prop({ reflect: true })
+    public prefix?: string;
+
+    /**
+     * A string of text that is visually placed after the value.
+     */
+    @Prop({ reflect: true })
+    public suffix?: string;
+
+    /**
+     * Set to `true` if info tile is disabled.
+     */
+    @Prop({ reflect: true })
+    public disabled? = false;
+
+    /**
+     * If supplied, the info tile will display a notification badge.
+     */
+    @Prop({ reflect: true })
+    public badge?: number | string;
+
+    /**
+     * Set to `true` to put the component in the `loading` state.
+     * This does _not_ disable the link. To do so, the
+     * `disabled` property should be set to `true` as well.
+     */
+    @Prop({ reflect: true })
+    public loading? = false;
+
+    /**
+     * If supplied, the info tile will be a clickable link.
+     *
+     * Supplying a value also adds an elevated effect using a shadow,
+     * as well as `cursor: pointer`, which appears on hover.
+     * While we strongly recommend supplying a link whenever the
+     * component should act as a link, if this is not possible, and
+     * you need to provide interaction through a click handler,
+     * you can still get the correct styling by supplying a `Link`
+     * object with the `href` property set to `'#'`.
+     */
+    @Prop()
+    public link?: Link;
+
+    /**
+     * Properties of the optional circular progress bar.
+     *
+     * Defaults:
+     * - `maxValue`: 100
+     * - `suffix`: %
+     * - `percentageColors`: false
+     *
+     * Colors change with intervals of 10 %.
+     */
+    @Prop()
+    public progress?: InfoTileProgress;
+
+    public render() {
+        const extendedAriaLabel =
+            this.checkProps(this?.prefix) +
+            this.value +
+            ' ' +
+            this.checkProps(this?.suffix) +
+            this.checkProps(this?.label) +
+            '. ' +
+            this.checkProps(this?.progress?.prefix) +
+            this.checkProps(this?.progress?.value) +
+            this.checkProps(this?.progress?.suffix) +
+            this.checkProps(this?.link?.title);
+
+        const link = !this.disabled ? this.link?.href : '#';
+
+        return [
+            <a
+                title={this.link?.title}
+                href={link}
+                target={this.link?.target}
+                tabindex="0"
+                aria-label={extendedAriaLabel}
+                aria-disabled={this.disabled}
+                class={{
+                    'is-clickable': !!this.link?.href && !this.disabled,
+                    'has-circular-progress':
+                        !!this.progress?.value || this.progress?.value === 0,
+                }}
+            >
+                {this.renderIcon()}
+                {this.renderProgress()}
+                <div class="value-group">
+                    {this.renderPrefix()}
+                    <div class="value-and-suffix">
+                        {this.renderValue()}
+                        {this.renderSuffix()}
+                    </div>
+                    {this.renderSpinner()}
+                </div>
+                {this.renderLabel()}
+            </a>,
+            this.renderNotification(),
+        ];
+    }
+
+    private checkProps(propValue) {
+        return !propValue ? '' : propValue + ' ';
+    }
+
+    private renderPrefix = () => {
+        if (this.prefix) {
+            return <span class="prefix">{this.prefix}</span>;
+        }
+    };
+
+    private renderValue = () => {
+        const characterCount = this.value.toString().length;
+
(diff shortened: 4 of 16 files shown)

### CODE REVIEW

File: src/components/info-tile/info-tile.tsx [hunk 1]
```diff
@@ -113,6 +119,7 @@ export class InfoTile {
                 tabindex="0"
                 aria-label={extendedAriaLabel}
                 title={this.linkTitle}
+                class={{ 'clickable-infotile': !!this.href || this.clickable }}
```
Kiarokh, line 122: Good idea! However, I think this should be more automated, like if there is an onClick defined. Maybe that's not possible. Given the fact that that can be set on the component from outside as well 🤔 We can wait to see what we will end up doing for `limel-dashabord-shortcut`

Also, if we have this, we can also check for `!this.disabled` in the same place, and remove that `box-shadow: unset` rule from the CSS

Kiarokh: for these kind of classes, we use a convention like: `is-xxx` or `has-xxx`, so `is-clickable` would be a better class name

LucyChyzhova: > Good idea! However, I think this should be more automated, like if there is an onClick defined. Maybe that's not possible. Given the fact that that can be set on the component from outside as well 🤔 We can wait to see what we will end up doing for `limel-dashabord-shortcut`

I see your point, I also started from the `onClick`, but as you said it's not quite possible to manage at the Elements side so I talked to @adrianschmidt, and his suggestion is to set up an extra prop, so we can apply the style based on it.

LucyChyzhova: > Also, if we have this, we can also check for `!this.disabled` in the same place, and remove that `box-shadow: unset` rule from the CSS

Agree! will add a fixup in a sec

src/components/info-tile/examples/info-tile-progress.tsx
```diff
@@ -0,0 +1,49 @@
...
+ * Displaying a progress bar instead of an icon
+ *
+ * It is possible to display a circular progress bar instead of an icon
+ * on the component. This can for instance help visualize how much of a set goal
+ * has been achieved, which together with the `value` will help users
+ * get a better overview of the provided data.
+ *
+ * :::tip
+ * It is possible to customize the progress bar's suffix, but it is
+ * set to display the percentage sign (**%**) by default.
```
Kiarokh, line 12: if I type this instead:
```md
set to display `%` by default.
```
Kompendium will crash and not display any DOCs for this example. @jgroth FYI

src/components/info-tile/info-tile.tsx
```diff
@@ -0,0 +1,265 @@
...
+     * The url that the info-tile leads to.
+     */
+    @Prop({ reflect: true })
+    public href?: string = null;
+
+    /**
+     * Set to `true` if info-tile is clickable.
+     */
+    @Prop({ reflect: true })
+    public clickable: boolean = false;
```
jgroth, line 85: Do we need this? If an `href` is provided, I would assume it will be clickable and not clickable if omitted

src/components/info-tile/info-tile.tsx
```diff
@@ -0,0 +1,265 @@
...
+     */
+    @Prop({ reflect: true })
+    public progressPrefix?: string;
+
+    /**
+     * When set to `true`, makes the progress bar's percentage colorful.
+     * Colors change with intervals of 10%.
+     */
+    @Prop({ reflect: true })
+    public displayPercentageColors: boolean = false;
```
jgroth, line 138: I think I would prefer if these were combined to a single property instead, e.g.
```ts
interface InfoTileProgress {
    value: number;
    maxValue: number;
    prefix: string;
    suffix: string;
    displayColors: boolean;
}
```

src/components/info-tile/info-tile.tsx
```diff
@@ -0,0 +1,265 @@
...
+     * Where to load the linked URL, as the name for a browsing context:
+     * - `_self`: in the current browsing context. (Default)
+     * - `_blank`: in a new tab.
+     * - `_parent`: in the parent browsing context of the current one.
+     * If no parent, behaves as `_self`.
+     * - `_top`: the topmost browsing context (the "highest" context
+     * that's an ancestor of the current one). If no ancestors, behaves as `_self`.
+     */
+    @Prop({ reflect: true })
+    public target: '_self' | '_blank' | '_parent' | '_top' = '_self';
```
jgroth, line 97: I think we should combine `linkTitle`, `href` and `target` in to a single property, e.g. `link`
```ts
interface InfoTileLink {
    title: string;
    href: string;
    target: string;
}
```

jgroth (commented):
The component itself looks amazing :star_struck: 
I did not look at the code in detail, just had a look at what the component API looks like and have a few comments about that

src/components/info-tile/info-tile.tsx
Kiarokh: It is possible to use the component for visualization purposes only. So there would be no need for the link `<a` functionality in that case, and or clickability. 

So, when there is a link (`href`) this adds an elevated effect using a shadow, as well as `cursor: pointer` by applying a class. So the component looks clickable. 

But the component can also be used without a link (`href`), in another webcomponent for instance, which only sets an `onClick` actions from outside. We can't know this from inside the component. That's why we made this a prop.

Kiarokh: Added a fixup with improved docs

src/components/info-tile/info-tile.tsx
Kiarokh: sounds super nice. can you show me how to do so and document it and use it. Do we have an example somewhere that I can copy?

src/components/info-tile/info-tile.tsx
jgroth: Ok! If it is for visual purposes only, can we make it with a CSS prop instead? Or maybe a class name?

jgroth: But should we even have the link functionality built into this component then? Maybe that is just making it too smart? The components in Lime Elements are supposed to be as dumb as possible to make them reusable, so maybe that functionality should be handled by the consumer instead.

Kiarokh: Yes, using a class name would also work the same in this case I think. But my experience helping the appers shows that developers use props more properly than class names. So for instance, they'd more easily set a prop to true programmatically, rather than applying a class name programatically.

Additionally, these styling CSS classes would need an unnecessary example too. They won't be reflected in our composite examples neither.

I feel we have done more example like this before in Elements. Right now too tired to remember them. Or to say for sure that a class name is better than a prop. Maybe we also need it for more things than mere styling.

src/components/info-tile/info-tile.tsx
```diff
@@ -0,0 +1,273 @@
...
+    };
+
+    private renderNotification = () => {
+        if (this.badge) {
+            return <limel-badge label={this.badge} />;
+        }
+    };
+
+    private renderSpinner = () => {
+        if (this.loading) {
```
Kiarokh, line 269: I also think we should add `aria-live` or something similar in the `value` div or the whole component, when the value has been updated (?) or while it's being updated (?). This guides is so poorly written (for me at least), which I can't really figure out what to do or when to. do it. Need to do more research.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live

src/components/info-tile/info-tile.tsx
jgroth: As I mentioned above, I think we should remove this functionality altogether and let the consumer handle anything that happens on click

src/components/info-tile/info-tile.tsx
adrianschmidt: I understand Johan's argument, and using a CSS class or CSS variable ought to be a good bit more performant, but I also see the point in props being easier to get people to use. If I completely disregard any performance aspect, I think I actually prefer to have it as a prop.

However, going from Johan's initial question—"Do we need this?"—and the answer that this can be used only for visualisation: can you elaborate? I understand that it _could_ be used for just about anything, but the question is whether it _should_. Is there a legitimate use case where a component called _shortcut_ should be used to only visualise something?

adrianschmidt: Crap, scratch my above comment… as you will have already figured out, I was confused 🤦‍♂️ 🤣

src/components/info-tile/info-tile.tsx
adrianschmidt: If the whole component should act as a single link, then yes, we could have the consumer wrap the whole thing in an `<a href="…">`.

But I'm not sure that's a very helpful approach towards the consumer. I generally agree about keeping components simple (or "stupid"), but they also need to be helpful enough to actually be valuable… 🤔 I'm not entirely sure where I stand on this yet…

src/components/info-tile/examples/info-tile-loading.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * through a process that requires some time. In such cases, it is
+ * a great idea to let the users know that the data is being updated.
+ *
+ * To do so, set the `loading` property to `true`. The component will then
+ * show an indeterminate progressbar indicating the data is being updated,
+ * while the older value is still being displayed.
+ *
+ * :::note
+ * Note that this does _not_ disable the link, and most probably you
+ * do not need to it to be disabled neither.
```
adrianschmidt, line 15: I had to look this up to be sure, but after a "not", "either" is the correct word to use, rather than "neither" 😊

```suggestion
 * do not need it to be disabled either.
```

src/components/info-tile/examples/info-tile-notification.tsx
```diff
@@ -0,0 +1,42 @@
+import { Component, h } from '@stencil/core';
+/**
+ * Displaying a notification badge
```
adrianschmidt, line 3: I'd avoid the word "notification" here. At least to me, the word "notification" means things like a push notification on my phone, or a banner on my computer screen.

Take this example:

image

(Ignore the fact that I opened the drawer of past notifications to get a banner for the screenshot. Just pretend it popped on screen just now.)

I wouldn't say it's _wrong_ to call the badge a "notification badge", but I just think it's increasing the risk of confusion without really bringing all that much value, compared to just calling it "badge".

adrianschmidt:
```suggestion
 * Displaying a badge
```

src/components/info-tile/examples/info-tile-notification.tsx
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * The component can display a notification badge, which could either be
+ * a `number` or a `string`. Read more about how the badge truncates
+ * or abbreviates the provided label [here](#/component/limel-badge/).
+ *
```
adrianschmidt, line 8:
```suggestion
 * The component can display a badge, which could either be a `number` or
 * a `string`. Read more about how the badge truncates or abbreviates the
 * provided label [here](#/component/limel-badge/).
```

src/components/info-tile/examples/info-tile-notification.tsx
```diff
@@ -0,0 +1,42 @@
...
+ */
+
+@Component({
+    tag: 'limel-example-info-tile-notification',
```
adrianschmidt, line 12: Regardless of whether we call the badge "badge" or "notification badge", I think the word "notification" _alone_, without "badge", definitely means something else.

```suggestion
    tag: 'limel-example-info-tile-badge',
```

src/components/info-tile/examples/info-tile-notification.tsx
```diff
@@ -0,0 +1,42 @@
...
+    shadow: true,
+    styleUrl: 'info-tile-notification.scss',
```
adrianschmidt, line 14:
```suggestion
    styleUrl: 'info-tile-badge.scss',
```

src/components/info-tile/examples/info-tile-notification.tsx
```diff
@@ -0,0 +1,42 @@
...
+})
+export class InfoTileNotificationExample {
```
adrianschmidt, line 16:
```suggestion
export class InfoTileBadgeExample {
```

src/components/info-tile/examples/info-tile-notification.scss
```diff
@@ -0,0 +1,7 @@
+:host(limel-example-info-tile-notification) {
```
adrianschmidt, line 1:
```suggestion
:host(limel-example-info-tile-badge) {
```

src/components/info-tile/examples/info-tile-notification.tsx
adrianschmidt: Don't forget to change the filename of this file and the scss file!

adrianschmidt:
I just had a look at the docs for this branch now, and I have to say: _DAMN_ the components you people build look good! 😍

src/components/info-tile/info-tile.tsx
jgroth: They would not have to wrap it in a link, the can just add an `onClick` handler depending on what the use case is. To me, the `clickable` prop is a strong indication that it's not clear how the component is supposed to be used and that it might be trying to do too much.

If we are not sure about the use case, I would say that we should start with the most simple solution (i.e. not having this functionality). We can always add it later if we find that we need it, it's a lot harder to remove it once we have introduced it.

adrianschmidt: > They would not have to wrap it in a link, the can just add an `onClick` handler depending on what the use case is.

Yeah, but we really want to start fostering the use of proper links, in my opinion. We need to move away from this trend of not using proper links for things that should cause navigation.

> To me, the `clickable` prop is a strong indication that it's not clear how the component is supposed to be used and that it might be trying to do too much.

I agree. I think we could simply use the presence of a value for `href`, as you proposed originally. People should supply a url, and if they really can't, but still want to use a clickHandler, and have the "clickable" styling, they can just supply `'#'` as the url, and then handle the click in their own click handler.

jgroth: Also, if we speak from the use case of Lime CRM webclient this will be hard to use today since clicking an anchor tag will probably reload the whole page. I think we need something like the [Navigation API](https://developer.chrome.com/docs/web-platform/navigation-api/) to intercept navigation events like that, but I have not read the spec in detail and it's not even implemented everywhere yet.

Kiarokh: My quick take on it

- `limel-shortcut` is quite similar to this one, structurally and when it comes to the usages of links.
- The main idea here (for the web client) is that this is supposed to be used as a link.
- Lime Elements is living between the two worlds. But I think we should mostly care about how things are sed in the web client, and its eco system.
- During many previous conversations, have have come to the conclusion that we need to use `<a>` for navigations, and not for example `<button>`s or `<div class="trigger" tabindex="0"`. One of the reasons is giving ability to META-key + click, and open things in new browser tabs.
- Not sure how much we have to care about how the web client behaves with anchor links. In that case, that sounds like a buggy behavior of the client and shouldn't have anything to do with us here. If it's not possible to fix, that's another story.
- Links come with accessibility concerns and responsibilities. If we leave all of that to the consultants, they will make inaccessible links. This component tries to address those internally, as good as "I" personally can. So we don't bother documenting too much or risk consultants making too inaccessible solutions. These things are becoming laws, and we will have too many problems.
- People can add onClick, I don't think it's a super good idea. But they can and sometimes have to and will 🤷‍♂️ so why not take it into account.
- Start small, then make things bigger. I'm not a big fan honestly. I need these features right now, since I have to implement this component already.

src/components/info-tile/examples/info-tile-loading.tsx
Kiarokh: 😳 I blame my English teacher 😄

src/components/info-tile/examples/info-tile-notification.tsx
Kiarokh: In my head, anything that "notifies" the user that something has happened or requires attention, is a notification.

So those on the top-right would be "notification cards"
The red badges wold be "notification badge".
Then we'd have "notification dot" (a badge without number or letters.
Notification emails
Notification sound
etc…

Anyhow, your suggestions seems to be what Material Design would like to call them https://m3.material.io/components/badges/guidelines

I'm fine either way

src/components/info-tile/info-tile.tsx
adrianschmidt: > Also, if we speak from the use case of Lime CRM webclient this will be hard to use today since clicking an anchor tag will probably reload the whole page. I think we need something like the [Navigation API](https://developer.chrome.com/docs/web-platform/navigation-api/) to intercept navigation events like that in order to push to history state, but I have not read the spec in detail and it's not even implemented everywhere yet.

I don't think we _need_ the Navigation API for that, although having it would probably make things easier.

jgroth: > I don't think we need the Navigation API for that, although having it would probably make things easier.

Yeah maybe we have some global click listener registered somewhere that can take care of it :smile:

adrianschmidt: > * Start small, then make things bigger. I'm not a big fan honestly. I need these features right now, since I have to implement this component already. I'd rather make the component `@private` first, if we are worried for consequences.

My thumbs up and heart on your post refers to everything you said, _except_ this 😅

I have much to say on this, and I'm starting to feel like I'm maybe, possibly, _perhaps_, starting to get some thoughts formulated in a way that I might be able to convey them in a way that could, if I'm lucky, make sense to people outside my own head… maybe… 🙈

Let's have a huddle sometime…

src/components/info-tile/examples/info-tile-notification.tsx
adrianschmidt: Yeah… as always, naming is hard 😅

The "notification dot" example does make me think "notification badge" makes sense. How about this: I'm fine with "notification badge", if you prefer that, but don't call the example "notification". Use either "notification badge", or just "badge". Good with you?

LucyChyzhova: I think when we combine them as two words "notification badge" it is fine. I also found this naming as a common practice in development 
> 
> Starting with 8.0 (API level 26), notification badges (also known as notification dots) appear on a launcher icon when the associated app has an active notification. Users can long-press on the app icon to reveal the notifications (alongside any [app shortcuts](https://developer.android.com/guide/topics/ui/shortcuts)), as shown in figure 1.
> 
> These dots appear by default in launcher apps that support them and there's nothing your app needs to do. However, there might be situations in which you don't want the to notification dot to appear or you want to control exactly which notifications to appear there.

https://developer.android.com/develop/ui/views/notifications/badges

src/components/info-tile/info-tile.tsx
Kiarokh: I added a fixup. But don't know how to use the new props in the example file. Can you add a code suggestion please?

src/components/info-tile/info-tile.tsx
Kiarokh: > I have much to say on this, 

I already know your standpoint on this. 😄 but a huddle would be nice. I also have a lot to say on this I guess. 

Regarding this particular component, features are based on current info-tile [^1]. I need to replace that for the multiple start pages to work smoothly. I don't wanna get stuck on technicalities, best practices, and bureaucracy for too long. We have already had a fair share of these types of discussions when I was trying to do this in the CRM-components repo.

I want move on.

But misunderstand me correctly now. 😸 I'm trying to say, please don't ask me to remove features to make the first PR small, and then gradually add more features, until I can bring it up to a level that I can use the component.

[^1]: and the needs that I am aware of, taking to others regarding other projects

src/components/info-tile/examples/info-tile-notification.tsx
Kiarokh: Already changed everything. It's OK 👌 😄

src/components/info-tile/info-tile.tsx
LucyChyzhova: the fixup doesn't work right now, I'm trying to figure it out.

src/components/info-tile/info-tile.tsx
jgroth: > I don't wanna get stuck on technicalities, best practices, and bureaucracy for too long. We have already had a fair share of these types of discussions when I was trying to do this in the CRM-components repo.
> 
> I want move on.

Sorry, but I do not agree at all. This is hardly technicalities nor bureaucracy. The public interface of the component is one of the most important aspects since it is what the consumer will use, so it's important that it's easy to use and makes sense. Once we have released something it can be extremely hard to remove or change without making a lot of people annoyed. That's why it's usually better to start small and add more stuff later as they are needed

Kiarokh: I already agree to that @jgroth 

> But misunderstand me correctly now. 😸 I'm trying to say, please don't ask me to remove features to make the first PR small, and then gradually add more features, until I can bring it up to a level that I can use the component.

This was a response or clarification to
> Start small, then make things bigger.

adrianschmidt: > > I don't think we need the Navigation API for that, although having it would probably make things easier.
> 
> Yeah maybe we have some global click listener registered somewhere that can take care of it 😄

I might have a look at creating one today…

src/components/info-tile/info-tile.tsx
LucyChyzhova: works now :)

src/components/info-tile/info-tile.tsx
```diff
@@ -0,0 +1,256 @@
...
+    /**
+     * -`value`: The value of the progress bar.
+     * - `maxValue`: The maximum value within the scale that the progress bar should visualize.
+     * Defaults to `100`.
+     * - `prefix`: The prefix which is displayed before the `progressValue`,
+     * must be a few characters characters long.
+     * `suffix`: The suffix which is displayed after the `value`, must be one or two characters long.
+     * Defaults to `%`
+     * - `percentageColors`: When set to `true`, makes the progress bar's percentage colorful.
+     * Colors change with intervals of 10%.
```
jgroth, line 123: These should be moved to the different properties on the `InfoTileProgress` interface

src/components/info-tile/examples/info-tile-progress.tsx
```diff
@@ -0,0 +1,64 @@
...
+})
+export class InfoTileProgressExample {
+    @State()
+    private progress: InfoTileProgress = {
+        value: 76,
+        maxValue: 100,
+        prefix: '↑',
+        suffix: 'sek',
+        displayPercentageColors: false,
+    };
```
Kiarokh, line 36: @LucyChyzhova seems you **HAVE to** add them, or you get an error. These should be optional.
image

src/components/info-tile/info-tile.tsx
Kiarokh: I think they should be repeated there as well, you mean? 
Because if I only have them there, then they won't show up in the DOCs.
image

Kiarokh: is this a missing feature in Kompendium? Or is the correct way repeating the same info twice?

Kiarokh: I added a fixup, but the docs for `progress` do not compile properly anyway. I don't know what kind of dark magic is in the piece of markdown text which I wrote for them. 😭 🤮

src/components/info-tile/info-tile.tsx
LucyChyzhova: it was not intentional 😂.  I will check what was before. and will put it back

src/components/info-tile/examples/info-tile-progress.tsx
```diff
@@ -26,11 +27,14 @@ import { Component, h, State } from '@stencil/core';
 })
 export class InfoTileProgressExample {
     @State()
-    private progressValue = 76;
+    private progress: InfoTileProgress = {
+        value: 76,
+        maxValue: 100,
+        prefix: '↑',
+        suffix: 'sek',
```
LucyChyzhova: > @LucyChyzhova why you change my stuff? 😅

it's only one place where I "changed your stuff" @Kiarokh - I put a suffix as `sek`.   I  wouldn't say "changed" since there was no suffix 😁, only the value. And because the suffix isn't optional I had to add it.  Yes, I didn't pay attention that it was `EUR` at the `info-tile` component, I was trying to make it work. 

I added a fixup.

Also, I have checked the history but couldn't find more changes to the props.

src/components/info-tile/info-tile.tsx
jgroth: This is how more or less how all coding documentation tools work :smile: 
In here, you should add general documentation about the `progress` property and what it is used for. Each individual property on the `InfoTileProgress` interface needs to be documented on the interface itself. Otherwise, when a coder is using the interface they will not see any docs for each individual property and it will be weird to force them to go somewhere else to read about it.

If you want to duplicate the information here you can do that, but I would not recommend since it will lead to either more work if it needs to be updated, or it will be out of sync if someone forgets. There should be a link to the full documentation for the interface in the generated docs, but in your image it looks like there is not link generated :thinking:

src/components/info-tile/examples/info-tile-progress.tsx
LucyChyzhova: yes, added a fixup

src/components/info-tile/info-tile.tsx
jgroth: Did we come to a conclusion about this? :smile: 
If you want to include the link functionality, I suggest grouping the props under a single prop named `link` like my initial comment. Also, I think we should remove the `clickable` prop and do what @adrianschmidt suggested, i.e. if the consumer want to have their own handler they have to add an empty `href` or similar and then implement their own `onClick`

src/components/info-tile/info-tile.types.ts
```diff
@@ -7,5 +7,5 @@ export interface InfoTileProgress {
     maxValue: number;
     prefix?: string;
     suffix: string;
-    displayPercentageColors?: boolean;
+    displayPercentageColors: boolean;
```
adrianschmidt: Without the `?`, the consumer _has_ to provide either `true` or `false`, instead of just leaving the property out if they don't want percentage colours. Is that really what we want?

src/components/info-tile/info-tile.tsx
```diff
@@ -226,8 +226,8 @@ export class InfoTile {
                     class="progress"
                     prefix={this.progress?.prefix}
                     value={this.progress.value}
-                    suffix={this.progress.suffix}
-                    maxValue={this.progress.maxValue}
+                    suffix={this.progress?.suffix}
+                    maxValue={this.progress?.maxValue}
```
adrianschmidt, line 229: These question marks shouldn't be necessary, because we've wrapped this in an `if` that checks that `this.progress.value` exists. If `this.progress.value` exists, clearly `this.progress` must exist 😄

src/components/info-tile/info-tile.types.ts
jgroth: I think it should be optional :slightly_smiling_face:

adrianschmidt: It seems it has already been made optional in a later fixup 👍

src/components/info-tile/examples/info-tile-progress.tsx
adrianschmidt: This has been changed, so I'll resolve this thread.

src/components/info-tile/info-tile.tsx
LucyChyzhova: ah, okay.  👍 I will add a fixup :)

src/components/info-tile/info-tile.tsx
adrianschmidt: Yeah, a link should suffice. It would be very nice if such a link was autogenerated, but if it's not, we can add it manually.

+1 on not duplicating.

src/components/info-tile/info-tile.tsx
adrianschmidt: I was hoping you'd come to that conclusion 😁

This will give you the functionality you need @Kiarokh 👍

src/components/info-tile/info-tile.tsx
adrianschmidt: Everything discussed here has been done, as far as I can see, so I'm resolving this.

src/components/info-tile/info-tile.tsx
adrianschmidt: ⚡

src/components/info-tile/info-tile.tsx
adrianschmidt: Yeah, it had to be added there, and then there seems to be some problem with the descriptions, that prevent the output from being generated. I'm guessing it's got something to do with the `%` characters, since that was problematic elsewhere… I'm working on it 🙂

src/components/info-tile/examples/info-tile-progress.tsx
adrianschmidt: Fixed 👍

src/components/info-tile/info-tile.types.ts
```diff
@@ -0,0 +1,30 @@
...
+     */
+    value: number;
+    /**
+     * The maximum value within the scale that the progress bar should visualize.
+     * Defaults to `100`.
+     */
+    maxValue?: number;
+    /**
+     * The prefix which is displayed before the `progressValue`,
+     * must be a few characters characters long.
```
adrianschmidt, line 17: I'll fix the `characters characters` typo, but apart from that, I have the following question:

What is "must be a few characters long" supposed to mean? It's an optional property, so clearly there doesn't have to be a prefix at all. Does this mean that _if_ there is a prefix, it has to be more than one character? 🤨

adrianschmidt: Ah! Just as I posted my question it hit me! I guess this is supposed to say something like "can be a few characters at most"?

LucyChyzhova: > "can be a few characters at most"?

yes, correct!

src/components/info-tile/info-tile.tsx
Kiarokh: As long as there is a way to get the interactive effects. The clickable was adding styles for mouse hover, mouse press, keyboard tab-focus, elevated shadow. 

https://user-images.githubusercontent.com/35954987/203765427-ab941a44-5c2b-4a76-b41a-9712d36f80d3.mov

I don't know how it works.

Can we automatically get these styles if there is an `onClick` defined from outside? 

In the video above, only one of them has a link. So all widgets look as if they are static.

src/components/info-tile/info-tile.tsx
Kiarokh: Do you understand what they are saying in this guide @adrianschmidt ? 

I don't understand which one we should do: 
- while `loading` is `true`, we should also have `aria-live="polite"` on the `<a>`, and then remove it as soon as `loading` is `false` again
- or `aria-live="polite"` should always be there 🤔 (this sounds quite stupid)

adrianschmidt: I'm not sure, but I think it needs to be there all the time, because if you remove it when removing the spinner, then you're telling the browser to no longer announce changes to that region, so the change you did while also removing the attribute probably won't get announced…

adrianschmidt: But we'd not want to announce every change perhaps, at least not the very first when creating the component. Perhaps it should be added when loading is true, and then removed with a timer after loading becomes false. I think we really need to test it with a screenreader to  be sure…

src/components/info-tile/info-tile.scss
```diff
@@ -0,0 +1,344 @@
...
+
+    * {
+        box-sizing: border-box;
+    }
+}
+
+:host(limel-info-tile[disabled]) {
+    a {
+        opacity: 0.5;
+        cursor: not-allowed;
```
adrianschmidt, line 72: This doesn't seem great? Didn't you, @Kiarokh, say that an info tile can be used for only illustrating something, and thus not have a url? In that case, we wouldn't want to dim the tile and show a `not-allowed`-cursor when hovering it? 🤔

src/components/info-tile/info-tile.tsx
adrianschmidt: > Can we automatically get these styles if there is an `onClick` defined from outside?

Yes, the consumer just has to set `link={{ href: '#' }}`.

adrianschmidt: The link properties has been combined to a single interface now. It's called `Link`, and there are probably other places we can use it too, like limel-shortcut for example 🙂

src/components/info-tile/info-tile.tsx
adrianschmidt: Fixed. I'll resolve this thread now.

src/components/info-tile/info-tile.types.ts
adrianschmidt: I made a small change to this description.

src/components/info-tile/info-tile.tsx
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    private renderPrefix = () => {
+        if (this.prefix) {
+            return <span class="prefix">{this.prefix}</span>;
+        }
+    };
+
+    private renderValue = () => {
+        const chCount = this.value.toString().length;
```
adrianschmidt, line 152: I assume this is "characterCount" abbreviated? That took me an inordinate amount of time to figure out, so please avoid abbreviations 😊 🙏

src/components/info-tile/info-tile.tsx
```diff
@@ -0,0 +1,214 @@
...
+                        !!this.progress?.value || this.progress?.value === 0,
+                }}
+            >
+                {this.renderIcon()}
+                {this.renderProgress()}
+                <div class="value-group">
+                    {this.renderPrefix()}
+                    <div class="value-suffix">
+                        {this.renderValue()}
+                        {this.renderSuffix()}
```
adrianschmidt, line 131: It's a little confusing that the div has a class `value-suffix`, but seems to hold both the value and the suffix?

src/components/info-tile/info-tile.tsx
Kiarokh: let's skip it for now. We add this as an issue for later.

Kiarokh: https://github.com/Lundalogik/lime-elements/issues/1973

src/components/info-tile/info-tile.scss
Kiarokh: Not sure if I understand. This is specifically for when it is `disabled`. So styles shouldn't be triggered, for "visualization mode only".

A disabled `<a` on the other hand doesn't really exist, theoretically; unless you are dealing with CRM and permissions user rights and so forth. So I couldn't decide what to do.

src/components/info-tile/info-tile.tsx
Kiarokh: oh yes

we must change this

src/components/info-tile/info-tile.tsx
Kiarokh: perhaps the class name should be called `value-and-suffix`

src/components/info-tile/info-tile.scss
adrianschmidt: Ah! Sorry, I was confused. I noticed at some point that the link was automatically set to `'#'` when `disabled` is true, and when I came across this a good deal later, I was thinking that `disabled` was set if the link is left empty. Just a complete mixup in my head. Move along, nothing to see here! 😝

src/components/info-tile/info-tile.tsx
LucyChyzhova: fixed

src/components/info-tile/info-tile.tsx
LucyChyzhova: fixed

src/components/info-tile/info-tile.tsx
adrianschmidt: > > Yes, the consumer just has to set link={{ href: '#' }}.
> 
> then we should perhaps document it? Can we reuse what I wrote about `isClickable`?

Fixed!



https://github.com/Lundalogik/lime-elements/pull/1911/files#diff-2b2e96118baabdc2489f06cff2a80821d813b0d72c449ae930a593a1f0e243c9R77

adrianschmidt: I've squashed the fixups, and will mark this thread as resolved. I'll then proceed to approve and merge. I don't think anyone will mind, right? 😝

adrianschmidt (approved):

src/global/shared-types/link.types.ts
```diff
@@ -0,0 +1,28 @@
...
+     * and users of assistive technologies.
+     */
+    title?: string;
+
+    /**
+     * Target for the link. See
+     * https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-target
+     * for more info.
+     */
+    target?: string;
```
Kiarokh, line 27: just curious. why didn't we do this: 
```tsx
target?: '_self' | '_blank' | '_parent' | '_top' = '_self';
```

src/components/info-tile/info-tile.tsx
```diff
@@ -0,0 +1,222 @@
...
+     *
+     * Supplying a value also adds an elevated effect using a shadow,
+     * as well as `cursor: pointer`, which appears on hover.
+     * While we strongly recommend supplying a link whenever the
+     * component should act as a link, if this is not possible, and
+     * you need to provide interaction through a click handler,
+     * you can still get the correct styling by supplying a `Link`
+     * object with the `href` property set to `'#'`.
+     */
+    @Prop()
```
Kiarokh, line 88: just curious. don't we need `@Prop({ reflect: true })` here? or we can't use it?

jgroth: You can only use `reflect` for primitives, i.e. `string`, `number` and `boolean`

adrianschmidt: Found an email notification about this thread when I was trying to clean up my inbox a bit. Just wanted to clarify that you _can_ use `reflect` for complex values (complex values = all values that are not primitives), but you should avoid it, because it leads to lower performance because these complex properties need to be serialised to strings all the time.

src/global/shared-types/link.types.ts
adrianschmidt: Because those are not the only values allowed. Any string is allowed, those are only the values that have special meaning. It was ages since I used anything but the default or `_blank`, so I'm not sure, but if I recall correctly, setting any other value will cause the link to open a new window or tab, and set the name of that "context" to the value you entered. If you then click another link on the first page, which uses the same target value, that link will open in the window or tab that was opened when you clicked the first link.

src/components/info-tile/info-tile.tsx
jgroth: I don't think you can since Stencil does not know how to serialize them, they will just en up as the string `[Object object]`

adrianschmidt: Ah, ok, that may well be true. I don't think I've actually looked at what happens, I only know the compiler doesn't balk if you try to do it 😄

PR #1934, opened 2022-11-03
feat(header): add named slot `actions` and deprecate unnamed slot

adrianschmidt:
I found this in a six month old branch… it doesn't seem to have been merged. I'm not sure if I never opened a pull request, or if I did but it didn't get merged for some reason… 🤷‍♂️

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/header/header.tsx
+++ b/src/components/header/header.tsx
@@ -44,7 +44,12 @@ import { Component, h, Prop } from '@stencil/core';
  * @exampleComponent limel-example-header-colors
  * @exampleComponent limel-example-header-responsive
  * @exampleComponent limel-example-header-narrow
- * @slot - Content (actions) to be put inside the far right surface of the header
+ * @slot actions - Content (actions) to be put inside the far right surface of
+ * the header
+ * @slot [no name] - DEPRECATED. The `actions` slot used to be unnamed. This
+ * behavior has been deprecated, and support will be dropped in a future
+ * version. Please add `slot="actions"` to your elements to ensure your code
+ * will continue to work with future versions of Lime Elements.
  */
 @Component({
     tag: 'limel-header',
@@ -90,9 +95,9 @@ export class Header {
                     </h2>
                 </div>
             </div>,
-            <div class="actions">
+            <slot name="actions">
                 <slot />
-            </div>,
+            </slot>,
         ];
     }
 
--- a/src/components/header/examples/header-responsive.tsx
+++ b/src/components/header/examples/header-responsive.tsx
@@ -64,7 +64,7 @@ export class HeaderExample {
         }
 
         return (
-            <div class="actions">
+            <div class="actions" slot="actions">
                 {this.actions.map(this.renderActionButton)}
             </div>
         );
--- a/src/components/header/examples/header.tsx
+++ b/src/components/header/examples/header.tsx
@@ -51,7 +51,7 @@ export class HeaderExample {
         }
 
         return (
-            <div class="actions">
+            <div class="actions" slot="actions">
                 {this.actions.map(this.renderActionButton)}
             </div>
         );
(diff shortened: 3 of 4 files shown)


adrianschmidt:
@Kiarokh I think you're the one who asked for this, but I'm not completely sure. Do you recognise it?

Kiarokh:
I have a vague memory that we wanted to have this feature of limel-header (actions) in limel-dialog as well. Limel-header is  used in that component too. But I don't remember depreciation of the actions slot :/

src/components/header/header.tsx
```diff
@@ -91,7 +96,9 @@ export class Header {
                 </div>
             </div>,
             <div class="actions">
-                <slot />
+                <slot name="actions">
+                    <slot />
+                </slot>
             </div>,
```
Kiarokh, line 102:
````suggestion
            <slot name="actions" class="actions">
                    <slot />
            </slot>,
````

Comment (Kiarokh) [reply src/components/header/header.tsx:102]: isn't this cleaner?

With the above, we would need this in `header.scss`
```scss
slot[name='actions'] {
```

instead of this 👇 
```scss
.actions {
```

in two places

adrianschmidt: If we don't need the div for styling, I'm only happy to get rid of it 👍

Kiarokh: will you fixup! so that I can approve

adrianschmidt: Sure, I'll fix! 👍

adrianschmidt: Hmm… 🤔 This doesn't seem to work 😞

A screenshot from VSCode, showing an error: Property 'class' does not exist on type 'SlotAttributes'

I didn't get an error when running `npm run build`, but when running `npm start`, I _do_ get an error, and it prevents the docs app from working when served locally, so that's a show-stopper.

adrianschmidt: A screenshot showing that running `npm start` fails with the same error as in the screenshot above

adrianschmidt: I think we need to keep it the way it is in this PR… 😞

adrianschmidt: @Kiarokh Approved?

Kiarokh: No worries. I think you missed that my suggestion didn't have the `class=`. I pushed a fixup to show what I mean

Kiarokh: I also couldn't control myself, and updated the basic example to be more simple and easier to grasp. So I pushed an additional `docs` commit as well. What do you think?

adrianschmidt: > I think you missed that my suggestion didn't have the `class=`.

Ah! Yes, or no, I did see that, but then thought "if we keep the class, we don't have to make those changes in the css you mentioned", but then when putting a `class` attribute on the slot didn't work I had completely forgotten about the fact that I had added that myself. I feel so stupid sometimes… 🤦‍♂️🤣

adrianschmidt: > I also couldn't control myself, and updated the basic example to be more simple and easier to grasp. So I pushed an additional `docs` commit as well. What do you think?

Oh, yes, so much better! Thank you! 👍

One thing I thought of now though: That first example is pretty weird. The information presented in that header looks like it would belong in a snackbar instead. We don't have to fix that here and now, but I think we could make that example a lot less misleading 😅

Kiarokh: totally agree!

Kiarokh (approved):

Issue #1942, opened 2022-11-09
circular-progress: add prefix functionality


LucyChyzhova:
## New feature motivation
To be more informative we would like to have an extra property before the value. By using a prefix we can emphasize how the value was changed.

## New feature description
![Screenshot 2022-11-09 at 12 47 05](https://user-images.githubusercontent.com/50618208/200826073-0349ef09-8423-476b-9f22-3352fabc3d53.png)

PR #1943, opened 2022-11-09
feat(circular-progress): add prefix functionality

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/1942

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/circular-progress/circular-progress.scss
+++ b/src/components/circular-progress/circular-progress.scss
@@ -94,7 +94,7 @@
 
 .value {
     display: flex;
-    font-size: calc(var(--size) * 0.25);
+    font-size: clamp(0.5rem, calc(var(--size) * 0.25), 4rem);
     color: var(--circular-progress-text-color, rgb(var(--contrast-1200)));
     z-index: 1;
     cursor: default;
--- a/src/components/circular-progress/circular-progress.scss
+++ b/src/components/circular-progress/circular-progress.scss
@@ -1,7 +1,6 @@
 @use '../../style/functions';
 @use '../../style/mixins';
 @use '../../style/internal/lime-theme';
-@use '../../style/internal/z-index';
 
 /**
  * @prop --circular-progress-size: Determines the visual size of the visualization. This does not override the `size` property if it is specified.
@@ -15,6 +14,7 @@
 :host {
     display: block;
     box-sizing: border-box;
+    isolation: isolate;
 }
 
 :host([size='x-small']) {
@@ -96,7 +96,7 @@
     display: flex;
     font-size: calc(var(--size) * 0.25);
     color: var(--circular-progress-text-color, rgb(var(--contrast-1200)));
-    z-index: z-index.$limel-circular-progress-value;
+    z-index: 1;
     cursor: default;
 }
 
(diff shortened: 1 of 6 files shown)


src/components/circular-progress/circular-progress.scss
```diff
@@ -76,6 +76,18 @@
...
     }
 }
+.prefix-value {
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+}
+
+.prefix {
+    color: var(--circular-progress-fill-color, var(--mdc-theme-primary));
```
Kiarokh, line 86: I don't think we should use either of these colors. 
It's better to give prefix and suffix their own CSS props if they should be styled.

src/components/circular-progress/circular-progress.tsx
```diff
@@ -83,11 +89,19 @@ export class CircularProgress {
                 aria-valuenow={this.value}
                 style={{ '--percentage': currentPercentage }}
             >
-                <span class="value">
-                    {abbreviate(value)}
-                    <span class="suffix">{this.suffix}</span>
-                </span>
+                <div class="prefix-value">
```
Kiarokh, line 92: adding this deep nesting doesn't seem necessary. I think we can create a nice layout with the need of complications in the HTML.

src/components/circular-progress/examples/circular-progress-props.tsx
```diff
@@ -43,6 +48,13 @@ export class CircularProgressPropsExample {
                 maxValue={this.maxStars}
                 suffix={this.stars}
             />,
+            <limel-circular-progress
+                size="x-large"
```
Kiarokh, line 52: not a big deal, but I'd say it's better that all have the same size

LucyChyzhova: it was only for testing purposes 😁 (easier to see) , I will remove it when it's done :)

src/components/circular-progress/circular-progress.tsx
```diff
@@ -46,6 +46,12 @@ export class CircularProgress {
     @Prop()
     public maxValue: number = PERCENT;
 
+    /**
+     * The prefix which is displayed before the `value`, must be one or two characters long. When the prefix is null, no prefix is used.
```
Kiarokh, line 50: the last sentence is explaining the "obvious", in a developer's way. We don't have to say how the code works, if you know what I mean :)

Kiarokh: Additionally, it's better to truncate it.

src/components/circular-progress/examples/circular-progress-props.tsx
Kiarokh: I added it as a fixup ;) made all of them big. That's what I meant. :)

Kiarokh:
Also change the commit message to 
```
feat(circular-progress): add `prefix` as a property
```

Issue #1960, opened 2022-11-17
Select: It's possible to interact even when`disabled` or `readonly`, using keyboard


Kiarokh:
https://user-images.githubusercontent.com/35954987/202482032-dadf6a57-d989-48fe-8943-7dd4aea160c6.mov

PR #1961, opened 2022-11-17
fix(select): forbid to use the keyboard navigation when `readonly` or…

LucyChyzhova:
… `disabled`

fix: https://github.com/Lundalogik/lime-elements/issues/1960

https://user-images.githubusercontent.com/50618208/202512538-fff8011b-58c9-49ac-81f2-edcb6c1221f3.mov

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/select/select.template.tsx
+++ b/src/components/select/select.template.tsx
@@ -82,17 +82,16 @@ const SelectValue: FunctionalComponent<
     };
 
     return (
-        <div
+        <button
             class={anchorClassList}
             tabindex="0"
             onClick={props.open}
             onKeyPress={props.onTriggerPress}
-            role="button"
             aria-haspopup="listbox"
             aria-expanded="false"
             aria-labelledby="s-label s-selected-text"
             aria-required={props.required}
-            aria-disabled={props.disabled}
+            disabled={props.disabled || props.readonly}
         >
             <span id="s-label" class={labelClassList}>
                 {props.label}
@@ -120,7 +119,7 @@ const SelectValue: FunctionalComponent<
                     ></polygon>
                 </svg>
             </span>
-        </div>
+        </button>
     );
 };
 
--- a/src/components/select/select.scss
+++ b/src/components/select/select.scss
@@ -63,6 +63,7 @@
     }
 
     .limel-select-trigger {
+        border: none;
         height: shared_input-select-picker.$height-of-mdc-text-field;
         display: inline-flex;
         align-items: center;


src/components/select/select.tsx
```diff
@@ -194,7 +194,7 @@ export class Select {
                 required={this.required}
                 invalid={this.invalid}
                 label={this.label}
-                helperText={this.helperText}
+                helperText={this.menuOpen ? this.helperText : ''}
```
Kiarokh, line 197: this doesn't seem to be necessary, or correct. Can you please explain what you want to achieve here?

src/components/select/select.tsx
```diff
@@ -295,6 +295,10 @@ export class Select {
     }
 
     private handleMenuTriggerKeyPress(event: KeyboardEvent) {
+        if (this.disabled || this.readonly) {
+            return;
+        }
+
```
Kiarokh, line 301: I believe this is treating the side-effect of the problem, rather than _fixing_ it the right way.

I guess the issue is that the trigger element is not a native HTML element such as  a`<button />` for instance, which could be used to perform an action (click, key-press, tap). Right now, it's a `div`. Therefore, it cannot get `disabled` for real. Currently it only has `aria-disabled` which doesn't really do much of anything.

If it was a `<button />` element instead, and would get the attribute of `disabled`, then we wouldn't need to add this special `if()` condition. The browser would rather take care of the correct disabled behavior for us.

image

## considerations
1. Adding this will introduce some visual bugs that we inherit from the user agent stylesheets that must be fixed separately. Perhaps only below is enough 👇 but we may need even more fixes. So visually inspect it so that it looks exactly as before. image
2. The[ MDC docs](https://m2.material.io/develop/web/components/input-controls/select-menus) tells us we should use a div for the trigger `<div class="mdc-select__anchor" aria-disabled="true">`. 
So it doesn't seem to be us, doing it the wrong way initially. 🤔 They also mention:
    > Add the mdc-select--disabled class to the mdc-select element, set the aria-disabled attribute on the mdc-select__selected-text element to be "true", and set the disabled attribute any hidden input element.

    and  also say:
    > NOTE: To programmatically set a select as disabled, use the disabled property in the MDCSelect API.

Therefore I am not really sure if using `<button` can create accessibility issues, or other problems for us? 
- Why do they say use `aria-disabled="true"`? 
- Why do their default example has `role="button"` on the anchor element,
   image
    But their `disabled` example doesn't have it. 
    image
    Is that what we should remove instead of the above fix, when the component is disabled?
    
@adrianschmidt what do you think? 

*****
Additionally the native select element should get this:
image

Kiarokh (changes requested):
have a few concerns described as comments

src/components/select/select.tsx
Kiarokh: i see… 🤔 
But isn't that just the symptom? It should not be showing up automatically, if we solve the focusability with keyboard. In other words, if we disable the trigger for real (as I explained in the other comment).

src/components/select/select.tsx
Kiarokh: in other words, the fix should be this: 
> Don't allow the trigger be focus-able using tabbing, when `readonly` or `disabled`.

Not this:
> Disable `handleMenuTriggerKeyPress`, when `readonly` or `disabled`. (this fix)


We could perhaps simply put `tabindex="-1"` on the trigger element. 🤔
This way, we don't need to that other commit for the `helperText`.

LucyChyzhova: Yes, it seems like a better solution. I have pushed these changes, and if Adrian doesn't object to them, I guess, we can proceed. 
Also, I couldn't notice any weird behavior or different styling after testing it.

Kiarokh (commented):
tnx Lucy. Wanna drop the first commit too?

src/components/select/select.tsx
Kiarokh: Cool. regarding accessibility, I think this change should be completely fine. 
I was reading on `aria-expanded` that is on the trigger element, which for us is always on `false`. This is another accessibility bug that we should fix: https://github.com/Lundalogik/lime-elements/issues/1966. However, reading on that, I saw some examples: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-expanded and it seems this change is better than what MDC suggests. Super strange to me.

vicgeralds: I think you're right about using button (at least if using `role="button"` was right to begin with). From what I understand, from accessibility point of view, `<div role="button">` is just the same thing as `<button>`

Disabling focus on a div seems to me like unneeded work to reinvent the button behaviour.

vicgeralds (commented):
Looks good to me! (if you remove first commit that it seems like you reverted)

Kiarokh (approved):

src/components/select/select.template.tsx
```diff
@@ -82,17 +82,16 @@ const SelectValue: FunctionalComponent<
     };
 
     return (
-        <div
+        <button
             class={anchorClassList}
             tabindex="0"
```
adrianschmidt, line 87: I guess we could probably remove this `tabindex` too, now that the button is a proper `button`.

It shouldn't do any harm if it's left in, other than possibly being a source of confusion some day in the future 😅

Kiarokh: yeah. totally. no tabindex seems to be needed really. I didn't pay attention to it. And I don't know why MDC has chosen a `div` for this. It sounds uncannily weird.

LucyChyzhova: Added a fix here: https://github.com/Lundalogik/lime-elements/pull/1969

PR #1976, opened 2022-11-28
1839 empty select

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/1839 

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/select/select.template.tsx
+++ b/src/components/select/select.template.tsx
@@ -30,11 +30,12 @@ interface SelectTemplateProps {
 export const SelectTemplate: FunctionalComponent<SelectTemplateProps> = (
     props
 ) => {
+    const value = props.value;
     let hasValue = !!props.value;
-    if (isMultiple(props.value)) {
-        hasValue = props.value.length > 0;
+    if (isMultiple(value)) {
+        hasValue = value.length > 0;
     } else if (hasValue) {
-        hasValue = !!props.value.value;
+        hasValue = !!value.value;
     }
 
     let isValid = !props.invalid;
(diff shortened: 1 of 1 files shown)


src/components/select/select.template.tsx
```diff
@@ -32,10 +32,12 @@ export const SelectTemplate: FunctionalComponent<SelectTemplateProps> = (
 ) => {
     const value = props.value;
     let hasValue = !!props.value;
+    let hasEmptyText = true;
```
LucyChyzhova, line 35: `hasEmptyText` sounds a bit confusing, what if we say `hasEmptyString` or `hasNoText` 🤔

Kiarokh: Good point. The reason is that items in the Select have both `value` and `text`.
Screenshot 2022-11-30 at 07 29 30
Even empty items which we wanna hide have `value` (some ID in the database in case of CRM). But their `text` is empty.

Why we say "empty" and for example not "hasNoText"?
1. it's inspired by the [`:empty` pseudo selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:empty).
2. It has a "text" string, but the string is empty.

LucyChyzhova: I see, I think this confusion was caused by the similarity of `value` and `text`, would be better to have `id` instead of `value`. 

but it's another topic. 
I would vote to have `hasNoText` but it's up to you.

LucyChyzhova (approved):

PR #1981, opened 2022-11-30
fix(form): enable `lint` and `fold` for code editor in form

Kiarokh:
The linting allows the user to see syntax errors in the JSON code shown in the editor.
Folding makes it possible to collapse pieces of code.

fix https://github.com/Lundalogik/crm-feature/issues/3105

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/widgets/code-editor.ts
+++ b/src/components/form/widgets/code-editor.ts
@@ -24,6 +24,8 @@ export class CodeEditor extends React.Component {
                 value: value,
                 language: 'json',
                 lineNumbers: true,
+                fold: true,
+                lint: true,
             },
             events: {
                 change: this.handleChange,


src/components/form/widgets/code-editor.ts
```diff
@@ -24,6 +24,8 @@ export class CodeEditor extends React.Component {
                 value: value,
                 language: 'json',
                 lineNumbers: true,
+                fold: true,
+                lint: true,
```
adrianschmidt, line 28: My initial reaction when realising this is in the form component was "These seem like good defaults, but should we really hardcode this here?". Then I realised we hardcode the language to be `json` 🙈

Unless it's somehow already possible to override these, we might want to add that possibility at some point. But until then, I don't see a problem hard-coding `fold` and `lint` to enabled 🤷‍♂️ 🙂

adrianschmidt, line 28: This is a `feat` not a `fix`… it's adding functionality, not fixing a bug.

Kiarokh: is it? 
in my opinion, we are fixing the non-existing linting and folding in code editors which are in forms. It's not a functionality for the consumer. They can't turn it off or have a say (as you also mentioned above). The day we make these like overridable pros, then that'd be a `feat`.

LucyChyzhova: I also think we are fixing the bug here, it should be a part of the expected behavior of the code editor - not an extra feature

LucyChyzhova (approved):

PR #1983, opened 2022-11-30
😫

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/picker/examples/picker-empty-suggestions.tsx
+++ b/src/components/picker/examples/picker-empty-suggestions.tsx
@@ -1,12 +1,14 @@
 import { LimelPickerCustomEvent, ListItem } from '@limetech/lime-elements';
 import { Component, h, State } from '@stencil/core';
 
+// 👇 Do NOT copy this to production code! It's a MOCK for this example ONLY!
 const NETWORK_DELAY = 500;
+// ☝️ Do NOT copy this to production code! It's a MOCK for this example ONLY!
 
 /**
  * With no suggestions and a message for empty search results
  *
- * :::important
+ * :::warning
  * This example simulates that searching is done on the server. Because these
  * examples do not _actually_ send requests to the server, we simulate a small
  * delay, using `setTimeout`. **Please do NOT copy that to production code!**
@@ -52,6 +54,7 @@ export class PickerExample {
     }
 
     private search = (query: string): Promise<ListItem[]> => {
+        // 👇 Do NOT copy this to production code! It's a MOCK for this example ONLY!
         return new Promise((resolve) => {
             if (query === '') {
                 // Simulate some network delay
@@ -70,6 +73,7 @@ export class PickerExample {
                 resolve(filteredItems);
             }, NETWORK_DELAY);
         });
+        // ☝️ Do NOT copy this to production code! It's a MOCK for this example ONLY!
     };
 
     private onChange = (event: LimelPickerCustomEvent<ListItem<number>>) => {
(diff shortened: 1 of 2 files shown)


jgroth (approved):

src/components/picker/examples/picker-empty-suggestions.tsx
```diff
@@ -1,12 +1,14 @@
 import { LimelPickerCustomEvent, ListItem } from '@limetech/lime-elements';
 import { Component, h, State } from '@stencil/core';
 
+// 👇 Do NOT copy this to production code! It's a MOCK for this example ONLY!
```
Kiarokh, line 4:
```suggestion
// 👇 Do NOT copy and paste this to production code! It's a MOCK for this example ONLY!
```

Kiarokh: I think you should write it so, to be more "correct".

Kiarokh:
😄

anderssonjohan (approved):

vicgeralds (commented):
Maybe, since this turns out to be a problem (I assume), we should instead/also do some changes to the code to make it clearer that we simulate a fake request.

What about

```ts
function fakeNetworkRequest(fakeResponseData = []) {
    return new Promise((resolve) => 
        setTimeout(() => resolve(fakeResponseData), NETWORK_DELAY)
    )
}
```

It would then be easier to spot `await fakeNetworkRequest()` and understand you shouldn't copy that line,

compared to seeing

```ts
// Simulate some network delay
setTimeout(() => {
   ...
}, NETWORK_DELAY);
```

everywhere and not really understanding what it does.

We could perhaps even move `fakeNetworkRequest` to another module, so they can't easily copy-paste it and they will be forced to remove the call to it.

jgroth:
> Maybe, since this turns out to be a problem (I assume), we should instead/also do some changes to the code to make it clearer that we simulate a fake request.
> 
> What about
> 
> ```ts
> function fakeNetworkRequest(fakeResponseData = []) {
>     return new Promise((resolve) => 
>         setTimeout(() => resolve(fakeResponseData), NETWORK_DELAY)
>     )
> }
> ```
> 
> It would then be easier to spot `await fakeNetworkRequest()` and understand you shouldn't copy that line,
> 
> compared to seeing
> 
> ```ts
> // Simulate some network delay
> setTimeout(() => {
>    ...
> }, NETWORK_DELAY);
> ```
> 
> everywhere and not really understanding what it does.
> 
> We could perhaps even move `fakeNetworkRequest` to another module, so they can't easily copy-paste it and they will be forced to remove the call to it.

Haha, this is more less exactly what I said to Adrian over Slack :smile:

src/components/picker/examples/picker-empty-suggestions.tsx
vicgeralds, line 4: I don't know, I'm a bit concerned that all these warning don't have a nice tone and assume the reader is stupid. We should be very clear and make it difficult to make this mistake (copying what should definitely not be copied), but in a friendly way if possible and without distracting too much from the example code.

adrianschmidt: True…

I just didn't feel I had the time to do the better solution right away, but I agree. It's perhaps better to not merge this in the interim, and just do the better solution when someone has a little bit of time 👍

PR #1990, opened 2022-12-05
feat(form): adds schema path to custom components form info

hannahu:
fix Lundalogik/crm-feature#3036

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/fields/schema-field.ts
+++ b/src/components/form/fields/schema-field.ts
@@ -197,8 +197,15 @@ export class SchemaField extends React.Component<FieldProps> {
     }
 
     private buildCustomComponentProps() {
-        const { disabled, readonly, name, registry, schema, errorSchema } =
-            this.props;
+        const {
+            disabled,
+            readonly,
+            name,
+            registry,
+            schema,
+            errorSchema,
+            idSchema,
+        } = this.props;
         const factoryProps = getFactoryProps(registry.formContext, schema);
 
         return {
@@ -216,6 +223,7 @@ export class SchemaField extends React.Component<FieldProps> {
                 errorSchema: errorSchema,
                 rootValue: registry.formContext.rootValue,
                 name: name,
+                schemaPath: this.getSchemaPath(idSchema.$id),
             },
         };
     }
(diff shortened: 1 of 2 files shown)


src/components/form/fields/schema-field.ts
```diff
@@ -260,4 +268,25 @@ export class SchemaField extends React.Component<FieldProps> {
...
+     * Gets the path to the current property within the schema
+     *
+     * @param {string} schemaId the id of the schema
+     * @returns {string[]} an array with the schema path for the current property
+     * @example
+     * const schemaId = 'root_sections_0_controls_0_name';
+     * const schemaPath = getSchemaPath(schemaId);
+     * // => ['sections', '0', 'controls', '0', 'name']
+     *
+     * The result is compatible with the Lodash `get` function
```
jgroth, line 282: Not sure if we need to mention this since it might create a weird dependency on Lodash. If they change their API this will be outdated unless we also change it.

hannahu: Removed!

src/components/form/fields/schema-field.ts
```diff
@@ -260,4 +268,30 @@ export class SchemaField extends React.Component<FieldProps> {
...
+     */
+    private getSchemaPath(schemaId: string): string[] {
+        const paramString = window.location.href.split('?')[1];
+        const selectedConfigPath = new URLSearchParams(paramString).get(
+            'selectedConfig'
+        );
+
+        if (!selectedConfigPath) {
+            return [];
+        }
```
civing, line 290: We can't couple the web client with lime elements like this. Can we get the full path in some other way?

civing: We can maybe look in `this.formInfo.rootValue` inside the component instead?

hannahu: Ah yeah I really don't know why I did it like this here since I was thinking about that when implementing it from the start, sorry. Too much context switching after the first time which was a long time ago :) I'll look at this again!

hannahu: > We can maybe look in `this.formInfo.rootValue` inside the component instead of `this.config`?

Excellent idea!

civing (dismissed):
Awesome! 😎

civing (approved):

PR #1994, opened 2022-12-13
docs(picker): add documentation for the `Searcher` type

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/picker/searcher.types.ts
+++ b/src/components/picker/searcher.types.ts
@@ -1,3 +1,11 @@
 import { ListItem } from '../list/list-item.types';
 
+/**
+ * A search function that takes a search-string as an argument, and returns
+ * a promise that will eventually be resolved with an array of `ListItem`:s.
+ *
+ * @param {string} query A search query. Typically what the user has written
+ * in the input field of a limel-picker.
+ * @returns {Promise<ListItem[]>} The search result.
+ */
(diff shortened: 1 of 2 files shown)


src/components/picker/searcher.types.ts
```diff
@@ -1,3 +1,11 @@
 import { ListItem } from '../list/list-item.types';
 
+/**
+ * A search function that takes a search-string as an argument, and returns
+ * a promise that will eventually be resolved with an array of `ListItem`:s.
+ *
+ * @param {string} query A search query. Typically what the user has written
+ * in the input field of a limel-picker.
+ * @returns {Promise<ListItem[]>} The search result.
```
adrianschmidt, line 9: Unfortunately the _types_ disappear. I'm guessing that's a bug, either in lime-elements or in kompendium.

I'm not sure if we should add the types to the description as a workaround, or just try to find and fix the bug instead? @jgroth @mjlime

mjlime: I'd love if it was all fixed, but do what you can.

jgroth (approved):

Issue #1995, opened 2022-12-14
select: label covers the selected value if multiple options are possible


LucyChyzhova:
## Current behavior

![set field bug](https://user-images.githubusercontent.com/50618208/207584280-8b722184-6d16-4d92-af03-2e8eb78f6c4d.gif)

1. …

## Environment

- **lime-elements** version: 
- Framework used: 
- Logs:

PR #1997, opened 2022-12-14
fix(select): do not cover a value by label if multiple selections are…

LucyChyzhova:
… possible

fix: https://github.com/Lundalogik/lime-elements/issues/1995

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/select/select.template.tsx
+++ b/src/components/select/select.template.tsx
@@ -86,9 +86,10 @@ const SelectValue: FunctionalComponent<
     const labelClassList = {
         'mdc-floating-label': true,
         'mdc-floating-label--float-above':
-            (!props.hasEmptyText && props.hasValue) ||
+            !props.hasEmptyText ||
             props.isOpen ||
-            props.readonly,
+            props.readonly ||
+            props.hasValue,
         'mdc-floating-label--active': props.isOpen,
     };
 


src/components/select/select.template.tsx
```diff
@@ -88,7 +88,8 @@ const SelectValue: FunctionalComponent<
         'mdc-floating-label--float-above':
             (!props.hasEmptyText && props.hasValue) ||
             props.isOpen ||
-            props.readonly,
+            props.readonly ||
+            props.multiple,
```
Kiarokh, line 92: this this will always push the label up 
image

Kiarokh (changes requested):

src/components/select/select.template.tsx
LucyChyzhova: right! added a fixup!

LucyChyzhova: https://user-images.githubusercontent.com/50618208/207591670-564455e9-c489-44b6-b404-5b0bdc1ad846.mov

src/components/select/select.template.tsx
```diff
@@ -88,7 +88,8 @@ const SelectValue: FunctionalComponent<
         'mdc-floating-label--float-above':
             (!props.hasEmptyText && props.hasValue) ||
```
Kiarokh, line 89: The problem is here. `hasValue` should already work. But it obviously doesn't. We should check why `hasValue` is ignored for "multiple" type.

src/components/select/select.template.tsx
```diff
@@ -88,7 +88,8 @@ const SelectValue: FunctionalComponent<
...
             props.isOpen ||
-            props.readonly,
+            props.readonly ||
+            (props.multiple && props.hasValue),
```
Kiarokh, line 92: this works, but the problem is not here. Read my other comment.

Kiarokh (changes requested):

src/components/select/select.template.tsx
LucyChyzhova: fixed 🫡

Issue #2004, opened 2022-12-30
Docs / Checkbox: remove all `limel-flex-container`s


Kiarokh:
<img width="1787" alt="Screenshot 2022-12-30 at 16 13 27" src="https://user-images.githubusercontent.com/35954987/210085266-5735e1c2-5834-4f16-99c4-1d977b864642.png">

PR #2005, opened 2023-01-02
Remove flex containers

Kiarokh:
fix https://github.com/Lundalogik/lime-elements/issues/2004

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/flex-container/examples/flex-container.tsx
```diff
@@ -90,31 +90,34 @@ export class FlexContainerExample {
 
     public render() {
         return [
-            <limel-flex-container justify="space-between">
```
Kiarokh, line 93: 🤣 I don't know if we should remove this example all together, or keep it still…

adrianschmidt: Well, we certainly shouldn't make _this_ change at least 🤣

But I guess `limel-flex-container` is deprecated, right? Then we can remove this example. We can make the docs for it private too.

adrianschmidt: Yes, `limel-flex-container` is deprecated, and it seems we've already made the docs for it private. So absolutely, feel free to just remove this example 👍

Kiarokh: But not sure if someone who still uses it comes and wants to see or test or fiddle around with it... maybe it should be there, until we actually just get rid of it totally? Maybe I'm too stupid. If we don't want people to use it, perhaps examples should just be removed.

Kiarokh: Ok, I did it as the last commit

Kiarokh:
image

I don't know why, but the online preview version of the build is not showing the new component correctly ☝️ this is not how this is supposed to look like. I also had trouble to see changes on my local dev. I think you have to checkout the branch locally, remove the `dist` folder, and then `npm ci` and `npm start` to see the changes.

src/components/flex-container/examples/flex-container.tsx
adrianschmidt: People can always go an look at the docs for older versions 🙂

src/examples/example-controls.scss
```diff
@@ -1,11 +1,19 @@
...
-    flex-wrap: wrap;
-}
+:host(limel-example-controls) {
+    display: grid;
+    grid-template-columns: repeat(
+        var(--example-controls-column-layout, auto-fill),
+        minmax(var(--example-controls-max-columns-width, 6rem), 1fr)
+    );
+    gap: 0 1rem;
+    align-items: center;
```
adrianschmidt, line 13: > image
> 
> I don't know why, but the online preview version of the build is not showing the new component correctly ☝️ this is not how this is supposed to look like. I also had trouble to see changes on my local dev. I think you have to checkout the branch locally, remove the `dist` folder, and then `npm ci` and `npm start` to see the changes.

"The online preview version"? 🤔 You mean the docs for the PR? Those are identical to what will be published for the release, so if the component doesn't look right there, then it won't in the release either… 🙁

However, I think the problem is that you are targeting the `host` here, but the component doesn't use shadow-DOM. I don't know why it doesn't use shadow-DOM to begin with. You can try enabling that, or target `limel-example-controls` rather than `:host(limel-example-controls)`. (I haven't tried it myself yet, but I think that ought to work.)

Kiarokh: >  the problem is that you are targeting the host here,

🙈 yes, in a last fixup, I removed `shadow: true,`. When I was struggling to understand why it doesn't work and clean-reinstalled everything it seemed to me that it doesn't have any effect. So that's why. 

Now I added it back.

src/examples/example-controls.tsx
```diff
@@ -1,15 +1,24 @@
...
     styleUrl: 'example-controls.scss',
 })
 export class ExampleControls {
     public render() {
-        return (
-            <limel-flex-container justify="end">
-                <slot />
-            </limel-flex-container>
-        );
+        return <slot />;
```
adrianschmidt, line 22: I think this commit (`feat(example-controls): make it more flexible and useful`) should be a `chore` or `refactor`. It doesn't affect the production code, so shouldn't trigger a new version. It could be a `docs`, if it makes a big enough change for the reader of the docs, that it's worth telling them about it. Basically, if it's something like "this thing was a significant problem when using the docs, and now we've fixed it, so you might want to have another look if there was something you couldn't do or understand before".

Kiarokh: Ok, sounds right. I was thinking maybe adding the new CSS variables makes it a `feat` for "us". But since it's a docs thing, perhaps better not to use that, and as you say use `chore` or `docs`. In this case I myself vote for `docs`.

Kiarokh: ⚡ changed to `docs`

adrianschmidt: `feat` and `fix` only applies to to changes that affects either the consumer or the user 🙂

src/components/button-group/examples/button-group-composite.tsx
```diff
@@ -51,22 +51,24 @@ export class ButtonCompositeExample {
     public render() {
         return [
             <limel-button-group {...this.props} onChange={this.handleChange} />,
-            this.renderForm(),
             <limel-example-event-printer
                 ref={(el) => (this.eventPrinter = el)}
             />,
+            this.renderForm(),
```
adrianschmidt, line 57: In the composite example for `limel-button`, the event printer is still below the form. I don't know where is better to put it, now that the form is no longer collapsible? I would probably still put it below the controls… 🤔

adrianschmidt, line 57: Also, this commit says `docs(banner)`, but it should say `docs(button-group)` 🙂

Kiarokh: good catch ⚡

Kiarokh: Since this new component doesn't have a margin bottom (it's usually the last, and a margin bottom makes the UI of the example quite weird). So when we put it on top of event printer, it becomes even move weird, because there won't be any space between them at all. I'll improve the `example-event-printer` and `example-value` in another PR to avoid such issues, and then will move them back to their previous place. However, I also think in composite examples, events aren't the main topic? Playing around with settings is the main purpose. Or?

src/components/date-picker/date-picker.tsx
```diff
@@ -48,6 +47,7 @@ const nativeFormatForType = {
  * @exampleComponent limel-example-date-picker-year
  * @exampleComponent limel-example-date-picker-formatted
  * @exampleComponent limel-example-date-picker-programmatic-change
+ * @exampleComponent limel-example-date-picker-composite
```
adrianschmidt, line 50: I think we need to add `style="--example-controls-column-layout:auto-fit;"` to the example-control element in this composite example 😅

image

src/components/table/examples/table-selectable-rows.tsx
```diff
@@ -78,6 +72,16 @@ export class TableExampleSelectableRows {
...
+            <limel-example-controls
+                style={{ '--example-controls-max-columns-width': '10rem' }}
+            >
+                <limel-button label="Select all" onClick={this.selectAll} />
+                <limel-button
+                    label="Clear selection"
+                    onClick={this.clearSelection}
+                />
+                <limel-button label="Reset" onClick={this.resetSelection} />
+            </limel-example-controls>,
```
adrianschmidt, line 84: The buttons look worse after this change 😕 Not the end of the world, but shouldn't there at least be some default gutter between rows? 🤔

Before:

image

After:

image

src/components/flex-container/flex-container.tsx
```diff
@@ -10,9 +10,13 @@ import { Component, h, Prop } from '@stencil/core';
  * Lime Elements. Please use CSS for your flexible container needs 🙂
  * https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Flexible_Box_Layout/Basic_Concepts_of_Flexbox
  *
+ * :::warning
+ * We will delete this component soon!
+ * Do not use it.
+ * :::
```
adrianschmidt, line 16: It's already marked as deprecated, and the description has already been changed to say that it will be removed. I don't think we need this extra warning, and I think it sets a precedent we don't want. Marking something as deprecated should be enough, we shouldn't teach people to expect that there will also be a big red warning label before we remove a component. (Also, no one will actually see this warning, since the component has been marked as private, so it's a bit of a moot point. I still wanted to explain my reasoning though 🤗)

image

src/components/table/examples/table-selectable-rows.tsx
Kiarokh: Oh I see. I intentionally made the example-controls' gap to be zero `gap: 0 1rem` just because for many checkboxes or toggle buttons it felt we don't need `row-gap`. They already had enough distance between them. But when there are buttons (or input fields), then we would need row gap. So we should either:
- set the row gap to be very little, like `gap: 0.25rem 1rem;` or 
- make a CSS variable for the row gap

Kiarokh:
great comments @adrianschmidt 
Tomorrow and the day after, I'll be attending the leadership course at Lime. So I think there'll be zero time to focus on this. I already had a non-stop day with no lunch pause and can't work more. Feel free to add fixups to this PR wherever you feel they are needed (if you want it to be merged and done). 😊 Otherwise I'll have to do it next week, and then the cycle has already started. :(

src/components/button-group/examples/button-group-composite.tsx
Kiarokh: 🙊 😅

src/components/table/examples/table-selectable-rows.tsx
adrianschmidt: Ah, yeah, I guess we might want to polish this a little later on then. But for now, I guess it's better that the checkboxes doesn't get loads of extra space between them, and this example can look a little off. Or maybe I'll do something about it, we'll see… 😄

adrianschmidt:
> great comments

🙏❤️

> Tomorrow and the day after, I'll be attending the leadership course at Lime. So I think there'll be zero time to focus on this. I already had a non-stop day with no lunch pause and can't work more. Feel free to add fixups to this PR wherever you feel they are needed (if you want it to be merged and done). 😊 Otherwise I'll have to do it next week, and then the cycle has already started. :(

I can probably fix any issues that might remain 😊👍

src/components/button-group/examples/button-group-composite.tsx
adrianschmidt: ⚡

src/components/date-picker/examples/date-picker-composite.tsx
```diff
@@ -52,13 +52,13 @@ export class DatePickerCompositeExample {
 
     private renderForm() {
         return (
-            <limel-collapsible-section header="Settings">
+            <limel-example-controls>
```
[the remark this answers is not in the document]
adrianschmidt: ⚡

src/components/table/examples/table-selectable-rows.tsx
adrianschmidt: ⚡

I set `gap: 0.25rem 1rem;`. It looks nice for both buttons and checkboxes 👍

src/components/flex-container/flex-container.tsx
adrianschmidt: ⚡

adrianschmidt (approved):
Excellent!

Issue #1982, opened 2022-11-30
Select: hide disabled option with empty `text`


Kiarokh:
## Current behavior

https://user-images.githubusercontent.com/35954987/204729396-91c27a9a-b76f-4ff0-83ae-2bb1afcef8ff.mov

## Expected behavior

If there is an item which has no `text` and it is `disabled`, it should not be displayed in the list of options. (If it had text, but was disabled, it could have been displayed).

## Environment

Planet earth

PR #2007, opened 2023-01-03
fix(select.template): hide disabled option with empty text

Befkadu1:
fix: #1982
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/select/select.template.tsx
```diff
@@ -248,19 +248,21 @@ function createMenuItems(
     options: Option[],
     value: Option | Option[]
 ): Array<ListItem<Option>> {
-    return options.map((option) => {
-        const selected = isSelected(option, value);
-        const { text, disabled } = option;
+    return options
+        .filter((option) => !option.disabled && !!option.text)
```
adrianschmidt, line 252: Disabled options should not be filtered out, and empty options should only be filtered out if `required` is `true`.

adrianschmidt (changes requested):

src/components/select/select.template.tsx
Befkadu1: The issue says

```
Expected behavior
If there is an item which has no text and it is disabled, it should not be displayed in the list of options. (If it had text, but was disabled, it could have been displayed).
```

I don't understand what you meant

adrianschmidt: Ah! I'm sorry, I noticed a bug, and saw that you are filtering on `disabled` and `text` here, but didn't notice the `&&`.

Let me backtrack and show the bug instead of making assumptions about the underlying problem 😊

Disabled options that are _not_ empty should be visible:

Latest `next`:
The disabled option 'Han Solo' is visible

This PR:
The disabled option 'Han Solo' is hidden

Unless the select's @Prop `required` has been set to `true`, it should still be possible to "unset" the value, by selecting the empty option:

Latest `next`:
Can select the empty option again after selecting an option

This PR:
Once an option has been selected, there is no way to select 'nothing'

Kiarokh: Sorry. Maybe my explanation has been bad. But Adrian summarizes the requirement  perfectly.

src/components/select/select.template.tsx
```diff
@@ -265,6 +267,23 @@ function createMenuItems(
...
+        if (option.text) {
+            // If the select component IS required, we keep only options
+            // that are not "empty". We only check the text property, because
+            // some systems use an "empty option" that will have a value for
+            // the `value` property, to signify "no option selected". Such
+            // an option should be treated as "empty".
+            return true;
+        }
+    };
+}
```
adrianschmidt, line 285: I made this fixup, so that instead of checking if the option is disabled, we check only if the select component itself is set to `required`, and if it is, we remove any options that have no text value.

As you can see, I also refactored a little bit, to have space for some comments.

src/components/select/examples/select-initially-empty.tsx
```diff
@@ -1,8 +1,14 @@
...
 
 /**
- * Initially Empty
+ * With Empty Option
+ *
+ * Adding an empty option makes it possible for the user to "unset"
+ * the value. Try selecting a value below, and then selecting the empty
+ * option again.
+ *
+ * If the component is set as required, the empty option is removed.
```
adrianschmidt: I updated this example, because it's not necessary to have an empty option to make the component initially empty. I think that was necessary when we first built the `limel-select` component, and that's why this example was here. But it doesn't really make sense anymore, since the first example, which doesn't have an empty value, is also initially empty.

So I changed this to make more sense and show off the functionality we just added in this PR 🙂

src/components/select/select.template.tsx
```diff
@@ -246,21 +246,42 @@ function isSelected(option: Option, value: Option | Option[]): boolean {
...
+                selected: selected,
+                disabled: disabled,
+                value: option,
+                icon: option.icon,
+                iconColor: option.iconColor,
+            };
+        });
+}
+
+function getMenuOptionFilter(selectIsRequired) {
```
Befkadu1, line 270: Just to make it more readable.

```suggestion
function getMenuOptionFilter(selectIsRequired: boolean) {
```

adrianschmidt: ⚡

adrianschmidt (approved):

PR #2032, opened 2023-01-13
Dark mode shadows

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/global/core-styles.scss
+++ b/src/global/core-styles.scss
@@ -27,6 +27,15 @@
     --lime-elevated-surface-background-color: rgb(var(--contrast-100));
 }
 
-@include mixins.in(dark-mode) {
-    --lime-elevated-surface-background-color: rgb(var(--contrast-200));
-}
+// ⚠️ This section below is commented out.
+// It's contents are moved to src/style/color-palette-extended.css
+//
+// Because we don't want to give dark-mode-specific shadows to those
+// clients that only import src/style/color-palette-extended-light-mode-only.css
+// Because if their OS is in dark mode, not their app,
+// The shadows will get rendered too dark, due to `@media (prefers-color-scheme: dark)`
+// rule from the mixin. The day we generate `.css` files from these `.scss` files,
+// we can do it differently.
+// @include mixins.in(dark-mode) {
+//     --lime-elevated-surface-background-color: rgb(var(--contrast-200));
+// }
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -12,6 +12,16 @@
 }
 
 @mixin in($media) {
+    // ⛔️ As long as we don't have a script that generates a
+    // `.css` files automatically, we cannot use this mixin.
+    // we need to manually write the dark-mode CSS variables
+    // inside this file: /style/color-palette-extended.css
+    //
+    // ⚠️ Also note that this mixin only puts styles
+    // on the `:root` which means the `<html` level.
+    // Therefore, it cannot be used inside components
+    // to generate custom dark-mode styles.
+
     @if $media == dark-mode {
         @media (prefers-color-scheme: dark) {
             :root:not([data-theme='force-light']) {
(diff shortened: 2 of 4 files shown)


src/style/mixins.scss
```diff
@@ -12,6 +12,16 @@
 }
```
adrianschmidt, line 12: I assume only these two last commits are actually new?

image

I can help rebase this branch correctly, if you want me to.

Kiarokh: oh yes. please do. No idea what happened  :/ did I rebase on an old branch?

adrianschmidt: > did I rebase on an old branch?

Something like that probably, yes 😅

No worries though! 🙂

adrianschmidt: Git even solved it automatically when I did an interactive rebase… 🤷‍♂️ 😄

src/style/shadows.scss
```diff
@@ -110,94 +110,105 @@
...
+//
+// Because we don't want to give dark-mode-specific shadows to those
+// clients that only import src/style/color-palette-extended-light-mode-only.css
+// Because if their OS is in dark mode, not their app,
+// The shadows will get rendered too dark, due to `@media (prefers-color-scheme: dark)`
+// rule from the mixin. The day we generate `.css` files from these `.scss` files,
+// we can do it differently.
+
+// @include mixins.in(dark-mode) {
+//     --button-shadow-normal: 0 #{functions.pxToRem(1.5)} #{functions.pxToRem(
```
adrianschmidt, line 125: Do we need to leave this commented out code here?

src/global/core-styles.scss
```diff
@@ -27,6 +27,15 @@
...
-}
+// ⚠️ This section below is commented out.
+// It's contents are moved to src/style/color-palette-extended.css
+//
+// Because we don't want to give dark-mode-specific shadows to those
+// clients that only import src/style/color-palette-extended-light-mode-only.css
+// Because if their OS is in dark mode, not their app,
+// The shadows will get rendered too dark, due to `@media (prefers-color-scheme: dark)`
+// rule from the mixin. The day we generate `.css` files from these `.scss` files,
+// we can do it differently.
```
adrianschmidt, line 38: What and how can we do things differently? These files are of course already being converted to .css, since browsers don't read .scss-files. But I suspect you refer to generating separate .css-files with _just_ the variable definitions?

src/style/shadows.scss
adrianschmidt: Yeah, the PR works, so I'll approve and merge it 👍

Then we can enable releases again! 🎉

adrianschmidt (approved):

src/global/core-styles.scss
Kiarokh: There are a few ways to treat this issue.

<details><summary>Dream scenario</summary>
I would have liked to have a file called `theme-light.css` that automatically is generated from a `theme-light.scss` file when the build goes through. Along with `theme-light-and-dark.scss` file, that also generates its own CSS version.

These files would then contain variables that we need, both colors, and shadows, and other probable things which we may need, like the elevated-surface variable.

But this could mean a breaking change, and consumers would now need to import other CSS files with other names. Maybe hard. I don't know.
</details>

<details><summary>More pragmatic way</summary>
We just make sure that there is an SCSS file for `color-palette-extended-light-mode-only.scss` and `color-palette-extended.scss` which imports colors and shadows and such variables in itself. These files would automatically then get converted to CSS when the build goes through.
</details>

<details><summary>Ultimate dream scenario</summary>
That we wouldn't have to deal with `force-dark` and `force-light` at all. We could just somehow tell the browser that user for this page prefers light or dark color them, no matter what their OS setting is saying.

We would anyway need to have some auto CSS generation, because that would help us keep different variables in different files and organize stuff better.
</details>

adrianschmidt (approved):

src/global/core-styles.scss
Kiarokh: In any case, we can let this go for now and open new issues for any other solution.

PR #2041, opened 2023-01-16
fix(info-tiles): adjust the layout if no container queries support

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/3144

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/info-tile/info-tile.scss
+++ b/src/components/info-tile/info-tile.scss
@@ -112,6 +112,7 @@ a {
 }
 
 .icon {
+    z-index: 1;
     position: absolute;
     top: 0.5rem;
     right: 0.75rem;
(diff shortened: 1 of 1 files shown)


src/components/info-tile/info-tile.scss
```diff
@@ -168,6 +174,7 @@ a {
         var(--label-max-size)
     );
     @supports not (container-type: size) {
+        width: 100%;
```
Kiarokh, line 172: I don't think this is needed. If we wanted to truncate, this would be needed.

Kiarokh (approved):

Issue #2043, opened 2023-01-16
Header: simplify the basic example


LucyChyzhova:
## Current behavior
<img width="841" alt="Screenshot 2023-01-16 at 14 44 51" src="https://user-images.githubusercontent.com/50618208/212692810-d88a46c5-be82-4dce-b003-52c71b8d6f26.png">

1. …

## Environment

- **lime-elements** version: 
- Framework used: 
- Logs:

adrianschmidt:
That header doesn't use the default colours. It's an example with a bunch of special colours set. So this is a documentation issue, not a bug.

adrianschmidt:
By the way, I also don't like that the first example has a bunch of weird colours. The first example should be as simple and representative as possible. That's why we usually put a "Basic Example" first.

PR #2047, opened 2023-01-17
fix(header): simplify the basic example

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/2043

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/header/examples/header.scss
```diff
@@ -1,7 +1,6 @@
 :host(limel-example-header) {
     --header-background-color: rgb(var(--contrast-1100));
     --header-heading-color: rgb(var(--contrast-100));
-    --header-subheading-color: rgb(var(--color-teal-lighter));
     --header-supporting-text-color: rgb(var(--color-red-lighter));
     --header-icon-color: rgb(var(--color-red-light));
 }
```
Kiarokh, line 6: This doesn't seem to have any effect anymore. 👇 Not 100% sure if it ever did.
````scss
limel-icon-button {
    color: rgb(var(--contrast-100));
}
````
We should probably add a new public variable to `limel-icon` (in another PR or commit). For example `--icon-fill-color`, and consequently add the same to variable to `limel-icon-button`, and use it here. If we really wan ti. 

Otherwise, it is better to remove it.

Kiarokh (changes requested):
It is great that you simplified the first example. But please do the following now:

1. Remove the style file from the first example and let it be all default colors.
2. Change the commit title to `docs`
3. Make a new example for colors, that has a SCSS file, like this one. And choose better colors. We can skip using `--contrast-x` and instead choose a `--color-x-x` and`--color-white` for the text for example. 
4. Also decide what to do regarding my other comment about `limel-icon-button` `color`.

LucyChyzhova:
> 3\. We can skip using `--contrast-x` and instead choose a `--color-x-x` and`--color-white` for the text for example.

if we skip using using `--contrast-x` the dark mode will not work... take a look at the screen record 

https://user-images.githubusercontent.com/50618208/212926639-85003db2-5f65-4370-a1d3-441252b1c79a.mov

Kiarokh (approved):
perfect 💯

Issue #2003, opened 2022-12-30
Button: with `mdc-button--unelevated` class does not get "our own" transitions styles.


Kiarokh:
https://user-images.githubusercontent.com/35954987/210080625-5f169aaf-c69f-4bcc-9996-5ccdb30daf29.mov

Kiarokh:
All of our buttons are elevated. We don't have unelevated buttons (except in icon-button, but that's another story).
So the question is whether we should have this class at all, or not?
![image](https://user-images.githubusercontent.com/35954987/210081028-677a7379-249e-4115-b621-2b3d547a590c.png)

adrianschmidt:
> So the question is whether we should have this class at all, or not?

No, that seems like there's been a mistake made somewhere… I think the buttons were once not elevated, and when changing them to be elevated, instead of changing which class was used, the styles for the `mdc-button--unelevated` class was changed? 🤷‍♂️

I've traced the addition of `mdc-button--unelevated` only for primary buttons back to commit 95a8e58c09e4164360e03e941fed1e2cf804f7e0, where I seem to have misinterpreted the meaning of `.classList.add('mdc-button--unelevated', 'primary');`… unless I changed that behaviour intentionally, but I don't really see why I would have.

Either way, feel free to remove that class! 🙂

adrianschmidt:
Btw, MDC offers the class `mdc-button--raised`, which you might want to use. But if it's somehow easier or better not to, then feel free not to use it 👍

PR #2050, opened 2023-01-18
2003 unelevated buttons

Kiarokh:
fix https://github.com/Lundalogik/lime-elements/issues/2003

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -32,32 +32,34 @@ button {
         width: 100%;
 
         &:not(:disabled) {
-            @include mixins.is-elevated-clickable(
-                $color: var(--mdc-theme-primary, rgb(var(--color-teal-default)))
-            );
             @include mixins.visualize-keyboard-focus;
+            @include mixins.is-elevated-clickable();
+        }
 
-            &.mdc-button--raised,
-            &.mdc-button--unelevated {
-                // Primary
+        :host(limel-button[primary]) & {
+            &:not(:disabled) {
                 color: var(--mdc-theme-on-primary, rgb(var(--color-white)));
                 background-color: var(
                     --mdc-theme-primary,
                     rgb(var(--color-teal-default))
                 );
             }
+            &:disabled {
+                background-color: rgba(var(--contrast-1700), 0.15);
+            }
+        }
+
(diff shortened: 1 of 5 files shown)


src/components/button/button.scss
```diff
@@ -32,32 +32,34 @@ button {
         width: 100%;
 
         &:not(:disabled) {
-            @include mixins.is-elevated-clickable(
-                $color: var(--mdc-theme-primary, rgb(var(--color-teal-default)))
-            );
             @include mixins.visualize-keyboard-focus;
+            @include mixins.is-elevated-clickable();
```
adrianschmidt, line 36: I've watched the video on the issue, but I don't understand what problem it's trying to show. When comparing this branch to `next`, I did notice that the problem with the text changing colour when hovering is gone in this PR, so that's nice, but apart from that, I really can't see what has changed!? 🤷‍♂️

src/components/button/button.tsx
```diff
@@ -73,7 +73,6 @@ export class Button {
             <button
                 class={{
                     'mdc-button': true,
-                    'mdc-button--unelevated': this.primary,
```
adrianschmidt, line 76: I renamed this from `fix` to `refactor`. It could have also been a `chore` I guess. But it's not a bug fix 😄

adrianschmidt (approved):

src/components/button/button.scss
Kiarokh: the difference is that MDC only animates `box-shadow`.
our mixins animate `box-shadow`, `background-color`, and `transform`. So the pressing-down animation looks better and smoother with ours.

src/components/button/button.tsx
Kiarokh: sure thing 👍

src/components/button/button.scss
Kiarokh: who will press rebase and merge? :D

PR #2098, opened 2023-02-01
Add languages to date picker

adrianschmidt:
See individual commits for details. (If you review this commit by commit, you will see that each change is _extremely_ small, and very easy to understand.)

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/date-picker/date.types.ts
+++ b/src/components/date-picker/date.types.ts
@@ -7,4 +7,4 @@ export type DateType =
     | 'quarter'
     | 'year';
 
-export type Languages = 'da' | 'en' | 'fi' | 'nb' | 'no' | 'nl' | 'sv';
+export type Languages = 'da' | 'en' | 'fi' | 'fr' | 'nb' | 'no' | 'nl' | 'sv';
--- a/src/components/date-picker/date.types.ts
+++ b/src/components/date-picker/date.types.ts
@@ -7,4 +7,13 @@ export type DateType =
     | 'quarter'
     | 'year';
 
-export type Languages = 'da' | 'en' | 'fi' | 'fr' | 'nb' | 'no' | 'nl' | 'sv';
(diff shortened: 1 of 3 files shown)


src/components/date-picker/date.types.ts
```diff
@@ -7,4 +7,13 @@ export type DateType =
...
+export type Languages =
+    | 'da'
+    | 'de'
+    | 'en'
+    | 'fi'
+    | 'fr'
+    | 'nb'
+    | 'no'
+    | 'nl'
+    | 'sv';
```
jgroth, line 19: This is out of scope for this PR, but I think we should deprecate this type and change the prop on the component to a string instead. The user could then use any locale they want without us having to update and maintain this list every time we need a new locale. It will be hard to do as long as we depend on Moment and all the locale imports it requires, but if we switch from Moment to `Intl.DateTimeFormatter` it would probably not be too hard.

adrianschmidt: Yeah, that's good input. I don't like that we do this either, but with Moment it felt like a necessity to not just import all available localisations.

In case `Intl.DateTimeFormat` doesn't solve the problem completely (it doesn't support week and quarter formats, for example), we could perhaps solve it by lazy-loading the requested localisation, instead of bundling them all…? 🤔

jgroth: > No approval though? grimacingsmile

Oh sorry, I got distracted by other stuff :smile:

jgroth (approved):

PR #2102, opened 2023-02-02
feat(form): add new layout type for rendering form fields in rows

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/3178

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/examples/row-layout-schema.ts
+++ b/src/components/form/examples/row-layout-schema.ts
@@ -1,0 +1,119 @@
+export const schema = {
+    description: 'This form has the row layout',
+    type: 'object',
+    properties: {
+        info: {
+            title: 'Preferences',
+            description:
+                'These settings will not affect how others see the data. These are only for you.',
+            type: 'object',
+            lime: {
+                layout: {
+                    type: 'row',
+                },
+            },
+            properties: {
+                language: {
+                    type: 'string',
+                    title: 'Language',
+                    description: 'Select the app language',
+                    default: 'ua',
+                    lime: {
+                        layout: {
+                            icon: 'globe',
+                        },
+                    },
+                    oneOf: [
+                        {
+                            type: 'string',
+                            const: 'sv',
+                            title: 'Swedish',
+                        },
+                        {
+                            type: 'string',
+                            const: 'ua',
+                            title: 'Ukrainian',
+                        },
+                        {
+                            type: 'string',
+                            const: 'en',
+                            title: 'English',
+                        },
+                        {
+                            type: 'string',
+                            const: 'am',
+                            title: 'Amharic',
+                        },
+                        {
+                            type: 'string',
+                            const: 'fa',
+                            title: 'Farsi',
+                        },
+                    ],
+                },
+                date: {
+                    type: 'string',
+                    title: 'Date format',
+                    default: 'yyyy-mm-dd',
+                    lime: {
+                        layout: {
+                            icon: 'calendar',
+                        },
+                    },
+                    oneOf: [
+                        {
+                            type: 'string',
+                            const: 'yyyy-mm-dd',
+                            title: '1999-01-31',
+                        },
+                        {
+                            type: 'string',
+                            const: 'yyyy.dd.mm',
+                            title: '1999.01.31',
+                        },
+                        {
+                            type: 'string',
+                            const: 'yyyy/dd/mm',
+                            title: '1999/01/31',
+                        },
+                    ],
+                },
+                notification: {
+                    type: 'boolean',
+                    title: 'Receive email notifications',
+                    lime: {
+                        component: {
+                            name: 'limel-switch',
+                        },
+                        layout: {
+                            icon: 'filled_message',
+                        },
+                    },
+                },
+                frequency: {
+                    type: 'number',
+                    title: 'Frequency of notifications',
+                    description:
+                        'How many times per day can we sent you email notifications at most?',
+                    minimum: 0,
+                    maximum: 10,
+                    multipleOf: 1,
+                    lime: {
+                        layout: {
+                            icon: 'notification_center_92458',
+                        },
+                    },
+                },
+                personalNumber: {
+                    type: 'integer',
+                    title: 'Personal identity number',
+                    lime: {
+                        layout: {
+                            icon: 'numbers_input_form',
+                        },
+                    },
+                },
+            },
+        },
+    },
+};
--- a/src/components/form/form.scss
+++ b/src/components/form/form.scss
@@ -5,6 +5,8 @@
 
 /**
  * @prop --form-body-padding: space around content of the form
+ * @prop --form-background-color-of-even-rows: Background of even rows in the form, when layout type is `row`. Defaults to `transparent`.
+ * @prop --form-background-color-of-odd-rows:Background of odd rows in the form, when layout type is `row`. Defaults to `--contrast-200`.
  */
 
 @include lime-typography.core-styles;
@@ -243,3 +245,5 @@
         }
     }
 }
+
+@import './row/row.scss';
--- a/src/components/form/form.types.ts
+++ b/src/components/form/form.types.ts
@@ -219,6 +219,15 @@ export interface GridLayoutOptions
     dense?: boolean;
 }
 
+export interface RowLayoutOptions
+    extends FormLayoutOptions<FormLayoutType.Row> {
+    /**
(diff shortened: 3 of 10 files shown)


src/components/form/form.types.ts
```diff
@@ -229,4 +229,6 @@ export enum FormLayoutType {
      * Render the form fields using a responsive grid layout
      */
     Grid = 'grid',
+
```
Befkadu1, line 232: Add a documentation here

src/components/form/form.types.ts
```diff
@@ -229,4 +229,13 @@ export enum FormLayoutType {
...
+    /**
+     * Render the form fields in full-width rows.
+     * Each row can have a leading `icon`, and a field.
+     * `title` and `description` provided by the schema will be placed
+     * on the row itself, and not on the field.
+     * This layout is good for creating UIs for user settings pages.
+     */
+    SettingsRow = 'settingsRow',
```
jgroth, line 240: I'm not sure `SettingsRow` is a good name for this, it's too coupled to our feature implementation and could have more use cases than that. I think the name should be more generic

src/components/form/form.types.ts
```diff
@@ -229,4 +229,13 @@ export enum FormLayoutType {
      * Render the form fields using a responsive grid layout
      */
     Grid = 'grid',
+
+    /**
+     * Render the form fields in full-width rows.
+     * Each row can have a leading `icon`, and a field.
```
jgroth, line 235: There's no indication here on how to provide an icon. We should probably add a type for it, just as we have done for the `GridLayoutOptions`

src/components/form/form.types.ts
Kiarokh: agree. what's a good name? 🤔 

Perhaps `rows`? _(reminds a little about the default layout, but perhaps it's still generic enough)_

Kiarokh: We used `Row`. I think it's good enough

src/components/form/form.types.ts
Kiarokh: Do you mean we add something like this:
```tsx
export interface RowLayoutOptions
    extends FormLayoutOptions<FormLayoutType.Row> {
    /**
     * gldkjgdlf
     */
    icon?: string;
}
```

But the icon is only for each field/row. Not for the whole form with `row` layout. Do we need to write something like this?

jgroth: Oh, yeah maybe this was something else :smile: 
But I think we had something that could be applied to each item when using the grid layout :thinking:

jgroth: The `rowSpan` and `colSpan` are used on the components when using the `GridLayoutOptions`, but we have documented in there how it should be used :slightly_smiling_face:

src/components/form/form.types.ts
Befkadu1: @jgroth can I resolve this if you are satisfied ?

src/components/form/form.types.ts
jgroth: Yes and no :smile: 
If we can add a type for it I think we should, and we did for the grid options :slightly_smiling_face:

Kiarokh: so are you requiring changes now or what? 😄 I added a new fixup, but don't know if we should do more or not?

jgroth: I think @Befkadu1 is working on it

src/components/form/examples/row-layout-schema.ts
```diff
@@ -0,0 +1,107 @@
...
+            type: 'object',
+            lime: {
+                layout: {
+                    type: 'row',
+                },
+            },
+            properties: {
+                language: {
+                    type: 'string',
+                    icon: 'globe',
```
jgroth, line 18: You cannot add `icon` here since it is not part of the JSON Schema specification. Any modifications needs to be done under the `lime` schema property

src/components/form/examples/row-layout-schema.ts
```diff
@@ -0,0 +1,107 @@
...
+                    title: 'Language',
+                    description: 'Select the app language',
+                    default: 'ua',
+                    lime: {
+                        component: 'limel-select',
```
jgroth, line 23: This is incorrect. It should be
```suggestion
                        component: { name: 'limel-select' },
```

src/components/form/examples/row-layout-schema.ts
```diff
@@ -0,0 +1,107 @@
...
+                        },
+                    ],
+                },
+                date: {
+                    type: 'string',
+                    icon: 'calendar',
+                    title: 'Date format',
+                    default: 'yyyy-mm-dd',
+                    lime: {
+                        component: 'limel-select',
```
jgroth, line 59:
```suggestion
                        component: {name: 'limel-select' },
```

Kiarokh: we should actually just remove this, it will automatically become `limel-select`

src/components/form/form.scss
```diff
@@ -5,6 +5,8 @@
 
 /**
  * @prop --form-body-padding: space around content of the form
+ * @prop --form-background-color-of-even-settings-rows: Background of even rows in the form, when layout type is `row`. Defaults to `transparent`.
+ * @prop --form-background-color-of-odd-settings-rows:Background of odd rows in the form, when layout type is `row`. Defaults to `--contrast-200`.
```
jgroth, line 9: We should remove the `settings` substring from these

src/components/form/examples/row-layout-schema.ts
```diff
@@ -0,0 +1,107 @@
...
+                language: {
+                    type: 'string',
+                    icon: 'globe',
+                    title: 'Language',
+                    description: 'Select the app language',
+                    default: 'ua',
+                    lime: {
+                        component: 'limel-select',
+                    },
+                    oneOf: [
```
Kiarokh, line 25: Screenshot 2023-02-07 at 16 15 29
It would be nice to inject this class on `limel-select` if it is used in a form with `row` layout. Is it possible?
Because otherwise they render mega big!

Or is it the responsibility of the consumer to add is a class via schema or via a custom component? @jgroth

adrianschmidt: It seems weird that the select component renders that big anyway… is there any case where we actually want that behaviour? Or is it a side-effect of some other behaviour we _do_ want? If not, maybe we can just make it not do that in the first place, and not need the extra class? 🤔

jgroth: I'm not sure there is an easy way to add the class the way things work now :thinking: One way would be to inspect the schema and inject the class that way, but it's not the prettiest solution. Another way could be to set the class in the react wrapper class for the select component, not sure if that is pretty or easy to do either.

I guess this would have been a lot easier to solve if the narrow style was applied using css properties instead, but not sure if that is possible to do or not :slightly_smiling_face:

Kiarokh: > It seems weird that the select component renders that big anyway…

All form fields (input, slider, select, etc…) have the same height, otherwise they'd look horrible on the Overview tab or other forms with a grid layout. This row layout is kind of an exception now.

> I'm not sure there is an easy way to add the class the way things work now 

Then I think we leave it for the consumers. They can always do so. 

> if the narrow style was applied using css properties instead

That's unfortunately not so easy right now. Many things change in the `is-narrow` UI. That might require a lot of refactoring. However, I have been thinking about enabling different densities in Lime Elements, using CSS props. Not sure how yet. But maybe some beautiful day we can find a solution.

src/components/form/form.scss
jgroth: Not sure why this was resolved without a fixup :smile:

Kiarokh: 😮 WTF. I'm 500% sure I fixed and pushed it 😓

Kiarokh: good catch!

Befkadu1: ⚡

jgroth (approved):

PR #2106, opened 2023-02-05
feat(date-picker): add custom formatter option

civing:
Adds the possibility to supply a custom formatter to your date-picker. If supplied all date formatting will be done with that function.

The formatter function will not be used if the date-picker is used on mobile screens, i.e. Android.

fix: Lundalogik/crm-feature#3164

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/date-picker/date.types.ts
```diff
@@ -8,3 +8,5 @@ export type DateType =
     | 'year';
 
 export type Languages = 'da' | 'en' | 'fi' | 'nb' | 'no' | 'nl' | 'sv';
+
+export type CustomDateFormatter = (date: Date) => string;
```
jgroth, line 12: Do we need to prefix it with `Custom`? I do not think it adds much :slightly_smiling_face:

vicgeralds: We already have a `DateFormatter` class, but I agree, that shouldn't really determine how we choose to name interfaces that we expose

vicgeralds: I just inlined this type, and it's only used in two places now.

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.tsx
```diff
@@ -132,6 +135,10 @@ export class DatePickerCalendar {
                 );
                 break;
         }
+
+        if (this.formatter) {
+            this.picker.formatDate = this.formatter;
```
civing, line 140: I really like this approach!

Right now `formatDate` is `protected` on the `Picker` class. We should either make it public or expose a setter method for the formatter?

src/components/date-picker/date.types.ts
civing: Yeah, I wanted to name it `DateFormatter` but I couldn't 🤣 
Inlining it is fine I guess.

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.tsx
vicgeralds: It seems like it's public, actually (I was a bit surprised about that)

vicgeralds: But I'm not sure if `picker.parseDate` still works 🤔

vicgeralds: It seems like it's not really needed when passing value as Date, so we should be fine leaving it unchanged.

vicgeralds (approved):
Looks good to me now 🙂

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.tsx
civing: Ah, yeah it was actually public, my bad 🙈 
But anyhow, I still think it might be more explicit to set the formatter with a setter from the `Picker`. For someone reading the code it might be easier if we leave some breadcrumbs don't you think?

civing, line 140: I don't know how to unresolve a conversation so I will post another comment 😵‍💫

I still think it might be more explicit to set the formatter with a setter from the Picker. For someone reading the code it might be easier if we leave some breadcrumbs don't you think?

vicgeralds: Sorry 😊 
What do you think about this? 0cad3575917589f6a4c5861c5b3bccafd1667de5

vicgeralds (approved):

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.tsx
civing: like it! 🍭

PR #2132, opened 2023-02-15
feat(button-group): add support for badges

### LINKED ISSUE #1880: limel-button-group support badges with text
## New feature motivation

Button-groups are a great way to display different views within the same context. A common use case for button groups is to use it as a type of filtering. What you usually want with filters is to indicate the number of results hiding behind that filter. 

Since you can't insert your own JSX to the values, one solution would be to add a `badge` property to the values list of the button-group. 

Here is an example from Go where it would be super useful:
image

## New feature description

Add the possibility of inserting a badge to limel-button-group. Probably limited to the type number. 
Something like this (quick HTML hack but you get the idea):
image

alexwero:
Fixes #1880 

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/button/button.types.ts
```diff
@@ -15,7 +15,12 @@ export interface Button {
     icon?: string;
 
     /**
-     * True if the button should be selected. only first button will be selected.
+     * True if the button should be selected. Only first button will be selected.
```
Kiarokh, line 18: too picky, but this should be a separate `chore` commit.

src/components/button/button.types.ts
```diff
@@ -15,7 +15,12 @@ export interface Button {
...
      */
     selected?: boolean;
+
+    /**
+     * The label displayed in the badge
+     */
+    badge?: number | string;
```
Kiarokh, line 25: If this is the correct way to do it, this change should be a separate `feat` commit. But I'm quite unsure about this part of the change. Could we perhaps do something similar to this one here?
https://github.com/Lundalogik/lime-elements/blob/bf484b2277cd29ba7e62b41ed16c373fb6af43e5/src/components/menu-list/menu-list-renderer.tsx#L139

Kiarokh: the above also renders a badge inside the list items that are displayed in a menu

LawrenceBorst (commented):
Good stuff!

src/components/button-group/button-group.scss
```diff
@@ -131,3 +139,11 @@
         @include shared_input-select-picker.looks-disabled;
     }
 }
+
+limel-badge {
+    margin-right: 0.25rem;
```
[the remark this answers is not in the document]
Kiarokh: Really good catch! An alternative would be using `pointer-events: none;` perhaps more correct, because the badge is just ornamental/visual and shouldn't react to clicks. By using `cursor: pointer;` we will get the same result, but if I was reading the styles code, I would assume that the badge is an interactive element; but it's not a button or link, so the browser doesn't add the pointer style automatically, that's why the write wrote it manually. So it's a bit misleading that way.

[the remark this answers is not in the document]
alexwero: Thank you both, I'm gonna add `pointer-events: none;`

src/components/button-group/examples/button-group-badges.tsx
```diff
@@ -0,0 +1,50 @@
+import { Component, h } from '@stencil/core';
+/**
+ * Button group with badges
+ *
+ * When used for filtering, badges can visulize the number of results
```
Kiarokh, line 5:
```suggestion
 * Badges can visualize the number of results
```

src/components/button-group/examples/button-group-badges.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * for each filter option in the button group. The badge can either
+ * have a `number` or `string` label.
+ * Read more about how the badge truncates or abbreviates the
+ * provided label [here](#/component/limel-badge/).
```
Kiarokh, line 9:
```suggestion
 * Badges can be used to add further contextual information. 
 * For example if the component is used to filter a set of data
 * the badges could visualize the number of entries
 * for each filter option. 
 * 
 * The badge can either
 * have a `number` or `string` label.
 * Read more about how the badge truncates or abbreviates the
 * provided label [here](#/component/limel-badge/).
```

src/components/button/button.types.ts
adrianschmidt: > the above also renders a badge inside the list items that are displayed in a menu

What do you mean? @Kiarokh

alexwero: I'm also a bit unsure how to change this

Kiarokh: > > the above also renders a badge inside the list items that are displayed in a menu
> 
> What do you mean? @Kiarokh

I mean my suggested way of doing it also is for rendering a badge in the list items.
image
https://lundalogik.github.io/lime-elements/versions/next/#/component/limel-menu/

src/components/button/button.types.ts
alexwero: I removed it for now :)

src/components/button/button.types.ts
alexwero: I will take a look at this today

src/components/button-group/examples/button-group-badges.tsx
alexwero: Done!

src/components/button/button.types.ts
adrianschmidt: > > > the above also renders a badge inside the list items that are displayed in a menu
> > 
> > 
> > What do you mean? @Kiarokh
> 
> I mean my suggested way of doing it also is for rendering a badge in the list items. image [lundalogik.github.io/lime-elements/versions/next/#/component/limel-menu](https://lundalogik.github.io/lime-elements/versions/next/#/component/limel-menu/)

I still don't understand what your suggested way of doing it is? Unless I'm completely missing something here, the `Button` type in this case corresponds to the `MenuItem` type in limel-menu, and the `MenuItem` type has a `.badge` property, just like what @alexwero has added here. So I still don't understand what it is you want to change or how? 😊

Maybe we can huddle about this on Monday?

Kiarokh: I see. Ok. Sorry, ignore what I wrote. 

But perhaps this should be a separate `feat` commit, since button type is used more universally by other components too?

adrianschmidt: Hmm… yeah, since this is no longer a group of normal limel-buttons (which is what this component started out as), it's perhaps a little bit weird that we still use the same type for these buttons as for limel-buttons… 🤔

How about `feat(button-group): add badge property to button type'? It's not perfect, but perhaps good enough?

adrianschmidt:
@Kiarokh, if you're happy with how Alexandra fixed something you commented on, I would suggest resolving the corresponding thread (if there isn't other things being discussed in it, but hopefully there shouldn't be). It makes it easier for everyone to see what's left, if anything.

src/components/button/button.types.ts
alexwero: I split the commit in two

adrianschmidt (changes requested):
One more small thing, otherwise I think it looks great!

src/components/button-group/button-group.scss
```diff
@@ -43,6 +43,14 @@
     label {
         cursor: pointer;
         transition: color 0.2s ease;
+        display: flex;
+        align-items: center;
+        &:has(> :last-child:nth-child(2)) {
```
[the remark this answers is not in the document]
alexwero: Maybe I could delete the comment on line 49 then?
`// When `label` has a limel-badge`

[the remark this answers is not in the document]
adrianschmidt: Yes! Please do 😊 👍

LawrenceBorst (approved):

olssoncharlie:
Yesssss! You guys rock <3

Issue #2070, opened 2023-01-23
Dock: "follow the red thread" enable notification badges for the buttons


Kiarokh:
## New feature motivation
Sometimes, the user might need to pay attention to an item inside one of the dock menus. Therefore, it'd be nice if they could show a red dot (too big in this mockup, but you understand).
<img width="371" alt="image" src="https://user-images.githubusercontent.com/35954987/213993832-76c5337c-f77d-43e9-9eee-a27e7221e47f.png">

## New feature description
The menus are not a part of the Dock, they are individual components as you know. So they should have their own way of handling the notifications, depending on their content. However, a dock button should be able to receive a notification, and display a badge.

## New feature implementation

We will be using `Limel-badge` which supports displaying number or a string. Sometimes we just need a red dot, and no info in it. Some undefined state. We just wanna say, there is something here. Just click and follow the red thread.

This may need a fix or feat in limel-badge.

PR #2139, opened 2023-02-20
feat(dock): enable displaying a notification badge on the buttons

Kiarokh:
fix: https://github.com/Lundalogik/lime-elements/issues/2070

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dock/dock-button/dock-button.tsx
+++ b/src/components/dock/dock-button/dock-button.tsx
@@ -124,13 +124,21 @@ export class DockButton {
                     selected: this.item?.selected,
                 }}
                 onClick={handleClick}
+                aria-live={this.item.badge ? 'polite' : 'off'}
             >
                 {this.renderIcon()}
                 {this.renderLabel()}
                 {this.renderTooltip()}
+                {this.renderNotification()}
             </button>
         );
     }
+
+    private renderNotification = () => {
+        if (this.item.badge || this.item.badge === '') {
+            return <limel-badge label={this.item.badge} />;
+        }
+    };
 
     private openPopover = (event: MouseEvent) => {
         event.stopPropagation();
--- a/src/components/dock/dock.types.ts
+++ b/src/components/dock/dock.types.ts
@@ -40,6 +40,11 @@ export interface DockItem {
      * Used to specify a custom component to render as a menu for the dock item.
      */
     dockMenu?: DockMenu;
+
+    /**
+     * If specified, will display a notification badge on the buttons in the dock.
+     */
+    badge?: number | string;
 }
 
 export interface DockMenu {
--- a/src/components/dock/examples/dock-notification.tsx
+++ b/src/components/dock/examples/dock-notification.tsx
@@ -1,0 +1,101 @@
+import { Component, h, State } from '@stencil/core';
+import { DockItem } from '../dock.types';
+
+/**
+ * Displaying a notification badge
+ *
+ * It is possible to display a notification badge on each individual
+ * button in the Dock. Badges are supposed to inform the user that
+ * there is something in the menu that requires their attention.
+ *
+ * This is typically done by displaying a number, which summarizes
+ * the quantity of the items that require user's attention.
+ *
+ * :::important
+ * The menus are not a part of the Dock. They are individual components
+ * that you develop separately. Make sure that the information
+ * and interactions regarding the notifications are correctly handled.
+ *
+ * For example, when the items that require user's attention are
+ * seen or handled by the user after opening the menu, the badge on the
+ * Dock button should disappear.
+ * :::
+ *
+ * When this quantity is unclear or undefined, you can simply pass an
+ * empty string (`badge: ''`), which will only render a circle on the button.
+ * This is enough to attract user's attention.
+ * However, it is also possible to use a short string such as "·" or "!"
+ * for such cases, if considered necessary.
+ *
+ * :::warning
+ * Do not negatively exploit this possibility and spam users' awareness.
+ * The Dock is the most important and most dominant structural part of
+ * the UI of your application. Therefore crowding it with too much noise
+ * _will_ negatively affect the user experience.
+ * :::
+ *
+ *
+ */
+@Component({
+    tag: 'limel-example-dock-notification',
+    shadow: true,
+    styleUrl: 'dock-notification.scss',
+})
+export class DockNotificationExample {
+    @State()
+    private dockItems: DockItem[] = [
+        {
+            id: 'home',
+            label: 'Home',
+            selected: true,
+            icon: '-lime-logo-go-filled',
+        },
+        {
+            id: 'tables',
+            label: 'Tables',
+            icon: 'insert_table',
+        },
+        {
+            id: 'search',
+            label: 'Search',
+            icon: 'search',
+            badge: '',
+        },
+    ];
+
+    @State()
+    private dockFooterItems: DockItem[] = [
+        {
+            id: 'user',
+            label: 'Account',
+            icon: 'user',
+            badge: 5,
+            dockMenu: { componentName: 'my-custom-menu-with-notifications' },
+        },
+    ];
+
+    public render() {
+        return (
+            <div class="application">
+                <limel-dock
+                    accessibleLabel="Dock Example: dock with notification badges"
+                    dockItems={this.dockItems}
+                    dockFooterItems={this.dockFooterItems}
+                    onItemSelected={this.handleSelected}
+                />
+            </div>
+        );
+    }
+
+    private handleSelected = (event: CustomEvent<DockItem>) => {
+        const setSelection = (item: DockItem) => {
+            return {
+                ...item,
+                selected: item.id === event.detail.id,
+            };
+        };
+
+        this.dockItems = this.dockItems.map(setSelection);
+        this.dockFooterItems = this.dockFooterItems.map(setSelection);
+    };
+}
--- a/src/components/dock/dock-button/dock-button.scss
+++ b/src/components/dock/dock-button/dock-button.scss
@@ -4,6 +4,8 @@
 
 .button {
     all: unset;
+    isolation: isolate;
+    position: relative;
 
     cursor: pointer;
     @include mixins.is-flat-inset-clickable(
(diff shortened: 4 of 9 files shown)


src/components/dock/dock-button/dock-button.tsx
```diff
@@ -124,14 +124,22 @@ export class DockButton {
                     selected: this.item?.selected,
                 }}
                 onClick={handleClick}
+                aria-live={this.item.badge ? 'polite' : 'off'}
```
Kiarokh, line 127: This is good because we want to let the assistive technologies know when there is an update in here (a notification). However, I don't know how the screen readers work really. :( When I test with Apple's screen readers, I feel everything is so freaking stupid. The reader doesn't even read out loud `aria-described-by`. We should really educate ourselves much more about this and go through the elements.

src/components/dock/examples/dock-notification.tsx
```diff
@@ -0,0 +1,96 @@
...
+        },
+    ];
+
+    @State()
+    private dockFooterItems: DockItem[] = [
+        {
+            id: 'user',
+            label: 'Account',
+            icon: 'user',
+            badge: '5',
```
Kiarokh, line 67: When we have this feature: https://github.com/Lundalogik/lime-elements/issues/2140 we should write more docs here and perhaps update this example too.

src/components/dock/examples/dock-notification.tsx
```diff
@@ -0,0 +1,96 @@
...
+ * For example, when the items that require user's attention are
+ * seen or handled by the user after opening the menu, the badge on the
+ * Dock button should disappear.
+ * :::
+ *
+ * :::warning
+ * Do not negatively exploit this possibility and spam users' awareness.
+ * The Dock is the most important and most dominant structural part of
+ * the UI of your application. Therefore crowding it with too much noise
+ * Will negatively affect the user experience.
```
vicgeralds, line 30: If you want to emphasize "will" here, I suggest instead 🙂 

```suggestion
 * _will_ negatively affect the user experience.
```

vicgeralds (commented):
Looks good to me overall, provided a few comments/questions about some details.

src/components/dock/dock-button/dock-button.tsx
vicgeralds:
```suggestion
                aria-live={this.item.badge !== undefined ? 'polite' : 'off'}
```

Kiarokh: I actually have no clue. These things are so poorly explained in their docs. I never know how they work. My assumption is that as soon as there is a notification (which didn't exist before), the user with assistive technologies must realize it too. In case of urgent changes, we would need `role="alert"` or `aria-live="assertive"` [more...](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Alert_Role). 

But since the notification badge is not too serious to scream out loud, they have the `aria-live="polite"`. How screen-readers actually tell the blind user about this little change in the content, I have no idea. 

I even inspected the web-based version of Apple's icloud to see if they have something similar, where there are badges 👇 :
image
There was nothing in there.

Then I tried inspecting the web-based version of Microsoft Teams to see how they do it
image
But their shit didn't even load in the browser anymore and kept showing unknown errors. So I gave up.
image

Kiarokh: regarding:
`aria-live={this.item.badge !== undefined ? 'polite' : 'off'}`

I still don't know how we will do with `limel-badge`. What if we can just do so in the config of the dock:
```tsx
private dockItems: DockItem[] = [
        {
            id: 'home',
            label: 'Home',
            selected: true,
            icon: '-lime-logo-go-filled',
            badge: '',
        },
```

Is this counted as `undefined`?

src/components/dock/dock-button/dock-button.tsx
```diff
@@ -124,14 +124,22 @@ export class DockButton {
...
                 {this.renderIcon()}
                 {this.renderLabel()}
                 {this.renderTooltip()}
+                {this.renderNotification()}
             </button>
         );
     }
 
+    private renderNotification = () => {
+        if (this.item.badge) {
```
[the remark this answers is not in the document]
Kiarokh: yes if you ask me. But that depends on how we do in limel-badge, or?

src/components/dock/dock-button/dock-button.tsx
adrianschmidt: In case this isn't where you've been reading already, this page is helpful: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live

vicgeralds: My point is that if we set `aria-live="off"` while we're not displaying a notification badge, and then `aria-live="polite"` when we show a notification badge, I believe we're only marking it as a "live region" _after_ we have dynamically updated it, so there's nothing. If we use `aria-live` we should have it set regardless of whether we're currently showing a badge or not. We should then set `aria-live="polite"` if we expect that we _might_ show a notification badge.

That is why I suggested that we change the check to `this.item.badge !== undefined` so we would still set `aria-live="polite"` when `this.item.badge === 0` for example.

src/components/dock/examples/dock-notification.tsx
```diff
@@ -0,0 +1,96 @@
...
+/**
+ * Displaying a notification badge
+ *
+ * It is possible to display a notification badge on each individual
+ * button in the Dock. Badges are supposed to inform the user that
+ * there is something in the menu that requires their attention.
+ *
+ * This is typically done by displaying a number, which summarizes
+ * the quantity of the items that require user's attention.
+ * However, it is also possible to use a short string such as "·" or "!"
```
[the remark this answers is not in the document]
Kiarokh: Sure. Maybe better to remove this line. Or remove it when we enable that other thing. Either way, not a big deal. This is still ok to say: 
> possible to use a short string such as "·" or "!"

src/components/dock/dock.types.ts
```diff
@@ -40,6 +40,11 @@ export interface DockItem {
      * Used to specify a custom component to render as a menu for the dock item.
      */
     dockMenu?: DockMenu;
+
+    /**
+     * If specified, will display a notification badge on the buttons in the dock.
```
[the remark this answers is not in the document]
Kiarokh: how do you mean "other than 0"?

src/components/dock/dock.types.ts
```diff
@@ -40,6 +40,11 @@ export interface DockItem {
...
+     */
+    badge?: number | string;
```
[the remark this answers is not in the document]
Kiarokh: do it later in this case.

src/components/dock/dock-button/dock-button.tsx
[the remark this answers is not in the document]
vicgeralds: limel-badge actually allows label to be empty or missing already and it will be rendered the same way. It was only the docs that suggested that label was required I think.

```suggestion
        if (this.item.badge || this.item.badge === '') {
```

src/components/dock/dock-button/dock-button.tsx
adrianschmidt: Maybe we don't need to nitpick this too deeply, since we all agree that we're basically just guessing… I already worry about this going down unnecessary rabbit holes… 😬

src/components/dock/dock.types.ts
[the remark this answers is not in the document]
vicgeralds:
```suggestion
    badge?: boolean | number | string;
```

src/components/dock/dock-button/dock-button.tsx
Kiarokh, line 127:
```suggestion
                aria-live='polite'
```

Kiarokh: I read the guidelines again and how I understand it this time is that we should always have the `polite` mode ON. Since we are telling the assistive technology **_in advance_** that this area can get a live update. Announce it properly, when it is updated.

Previously I thought once it gets the `polite` activated, the AT will see this immediately and properly and non-intrusively announces it to the user. Which was why I wrote the check the way I wrote it.

However, it seems to be unnecessary (if I understand the guidelines correctly).

src/components/dock/dock.types.ts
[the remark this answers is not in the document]
Kiarokh: What about this one?

src/components/dock/dock.types.ts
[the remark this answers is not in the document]
Kiarokh: or should I first make a PR for the other one?

src/components/dock/dock-button/dock-button.tsx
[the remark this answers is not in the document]
Kiarokh: Ah OK. Cool. Yes 👇 
image
I just think it unnecessarily renders too big. I'll change this in a new PR.

src/components/dock/dock-button/dock-button.tsx
vicgeralds: But I'm not sure we want to communicate `aria-live='polite'` about _all_ dock buttons?

My idea was that we would use `this.item.badge: false` or `this.item.badge: 0` to indicate that a badge might be shown later.

vicgeralds: Ok, let's just do `aria-live='polite'`. I'm ok with that.

vicgeralds (approved):

src/components/dock/dock.types.ts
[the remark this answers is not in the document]
vicgeralds: Sorry, I missed this. Currently, the badge will not be shown when `badge === 0`. Because 0 is a falsy value when you check

```
if (this.item.badge) {}

PR #2144, opened 2023-02-21
feat(badge): make label optional

### LINKED ISSUE #2140: Badge: render even without label.
Screenshot 2023-02-20 at 16 51 30

There are many scenarios where we want to display a badge, but empty. The reason could be indetermination. Like, when the system knows that there is something that requires user's attention, but there is no quantity for it. We don't know how many there are... we just know that there is something.

vicgeralds:
This updates the documentation to make the `label` property on limel-badge optional.

fix #2140

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/badge/badge.tsx
+++ b/src/components/badge/badge.tsx
@@ -2,8 +2,10 @@ import { Component, Prop, h, Host } from '@stencil/core';
 import { abbreviate } from './format';
 
 /**
- * The Badge component can display both `number` and `string` as `label`.
+ * The Badge component can be used to display a notification badge,
+ * optionally with a number or a text label.
  *
+ * @exampleComponent limel-example-badge
  * @exampleComponent limel-example-badge-number
  * @exampleComponent limel-example-badge-string
  */
@@ -20,7 +22,7 @@ export class Badge {
      * six characters.
      */
     @Prop({ reflect: true })
-    public label: number | string;
+    public label?: number | string;
 
     public render() {
         return (
--- a/src/components/badge/examples/badge.scss
+++ b/src/components/badge/examples/badge.scss
@@ -1,0 +1,3 @@
+:host(limel-example-badge) {
+    --badge-background-color: rgb(var(--color-red-default));
+}
--- a/src/components/badge/examples/badge.tsx
+++ b/src/components/badge/examples/badge.tsx
@@ -1,0 +1,28 @@
+import { Component, h } from '@stencil/core';
+
+/**
+ * Badge without a `label`
+ * When no `label` is provided, the badge will only render as a circle.
+ * This is a convention which is used in many applications to attract the
+ * user's attention to a certain element on the user interface; typically to
+ * menus or buttons that navigate the user to another pane or screen.
+ *
+ * In such cases, the idea is to provide the users with a "red thread"
+ * and help them find something that requires their attention, but is located
+ * on another place in the app, and not directly visible.
+ *
+ * :::tip
+ * Make sure that the dot is noticeable, by providing an
+ * eye-catching background color, as shown in this example.
+ *:::
+ */
+@Component({
+    tag: 'limel-example-badge',
+    styleUrl: 'badge.scss',
+    shadow: true,
+})
+export class BadgeExample {
+    public render() {
+        return [<limel-badge />];
+    }
+}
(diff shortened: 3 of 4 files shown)


src/components/badge/examples/badge.tsx
```diff
@@ -0,0 +1,18 @@
+import { Component, h } from '@stencil/core';
+
+/**
+ * Badge without a label
```
Kiarokh, line 4:
```suggestion
 * Badge without a `label`
 * When no `label` is provided, the badge will only render as a circle.
 * This is a convention which is used to in many applications to attract 
 * user's attention to a certain element on the user interface; typically to
 * menus or buttons that navigate the user to another pane or screen. 
 *
 * In such cases, the idea is to provide the users with a "red thread"
 * and help them find something that requires their attention, but is located 
 * on another place in the app, and not directly visible.
 *
 * :::tip
 * In such cases, make sure that the dot is noticeable, by providing an
 * eye-catching background color, as show in this example.
 *:::
```

src/components/badge/examples/badge.scss
```diff
@@ -0,0 +1,9 @@
+:host(limel-example-badge) {
+    display: flex;
+    justify-content: space-between;
+}
+
+limel-badge {
+    --badge-background-color: rgb(var(--color-red-default));
+    --badge-text-color: rgb(var(--color-yellow-lighter));
+}
```
Kiarokh, line 9:
```suggestion
:host(limel-example-badge) {
    --badge-background-color: rgb(var(--color-red-default));
}
```

Kiarokh (commented):
1. What do you think about my suggestions?
2. Should this be a `feat`? I feel it should be `docs`, since the feature already existed but wrongly documented.

Kiarokh (commented):
Also, there seems to be a lint error

src/components/badge/badge.tsx
```diff
@@ -20,7 +22,7 @@ export class Badge {
      * six characters.
      */
     @Prop({ reflect: true })
-    public label: number | string;
+    public label?: number | string;
```
Kiarokh, line 25: didn't [you guys say that](https://github.com/Lundalogik/lime-crm-building-blocks/pull/24#discussion_r1104502622) `?` are unnecessary because Stencil already considers these optional. If we want them to be required, we should just add `!` instead? 😅

vicgeralds: Yes! But... TypeScript thinks otherwise from the component's point of view with `strictNullChecks: true`. 

In other words, adding `?` doesn't make any difference for the consumer. But it gives us better typing when using `this.label`.

src/components/badge/examples/badge.scss
```diff
@@ -0,0 +1,3 @@
+:host(limel-example-badge) {
```
Kiarokh, line 1: woops
Screenshot 2023-02-21 at 18 35 51

I think this is a buggy behaviour which we have never noticed before. Because all previous examples had `display: flex` on the host.

The fix for it is to go to `badge.scss` and change `display: flex` to `display: inline-flex`. This is more correct and prevents such eff-ups.

Can you add please a new commit to this PR Victor?

Kiarokh: `fix(badge): prevent stretching through the width of block containers`

Kiarokh: Screenshot 2023-02-22 at 09 19 55

vicgeralds: Good catch! I will do that

vicgeralds: 64fa53dc068afbb1d96d320f0b2fe8a650f19320

Kiarokh (dismissed):

Kiarokh (approved):

src/components/badge/badge.tsx
adrianschmidt: It also makes a light difference for the docs. If we have neither `!` nor `?`, the docs show this:

image

Ideally, we should fix that in Kompendium, and only show `Optional: true/false`, but I guess it's not been the highest of priorities.

Kiarokh: https://github.com/jgroth/kompendium/issues/120

PR #2146, opened 2023-02-22
fix(badge): render it smaller, when there is no label

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/3227

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/badge/badge.scss
+++ b/src/components/badge/badge.scss
@@ -12,7 +12,7 @@
     display: none;
 }
 
-:host {
+:host(limel-badge) {
     @include lime-typography.base;
     cursor: default;
     box-sizing: border-box;
@@ -23,12 +23,11 @@
 
     font-size: functions.pxToRem(11);
 
-    height: functions.pxToRem(20);
-    line-height: 1;
-    min-width: functions.pxToRem(20);
-    padding: 0 functions.pxToRem(5);
+    max-height: 1rem;
+    min-width: 0.5rem;
+    min-height: 0.5rem;
 
-    border-radius: functions.pxToRem(20);
+    border-radius: 2rem;
 
     background-color: var(--badge-background-color, rgb(var(--contrast-500)));
     color: var(--badge-text-color, rgb(var(--contrast-1200)));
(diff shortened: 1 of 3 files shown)


src/components/badge/badge.scss
```diff
@@ -23,21 +23,27 @@
 
     font-size: functions.pxToRem(11);
 
-    height: functions.pxToRem(20);
-    line-height: 1;
-    min-width: functions.pxToRem(20);
-    padding: 0 functions.pxToRem(5);
+    max-height: 1.25rem;
```
vicgeralds, line 26: changing height to max-height makes badges with a label a few pixels smaller (20px to 16.5px). Is this intentional?

vicgeralds (approved):
Looks good to me, just a question about the height when labels are used

src/components/badge/badge.scss
Kiarokh: I'm so happy that you noticed this! Yes it is intentional. I feel they were too clunky previously. Unnecessarily too big.

But you got me thinking, should I announce a BREAKING CHANGE in the commit message? Something like:

```
BREAKING CHANGE:
Badge height is decreased from `1.25rem` to `1rem`. 
This may have a slight effect on your UI design.
```

adrianschmidt:
@Kiarokh Please consider disabling the `use-first-commit-message-for-new-prs` option in your Refined GitHub 🙏 😅

I often find it very confusing when a PR title makes it look like there's only one commit in it, often with something trivial like "fix typo", and then there are more significant changes hiding in later commits 😄

src/components/badge/badge.scss
adrianschmidt: No, unless I'm totally underestimating the impact here, I don't think that should be considered a breaking change.

vicgeralds (approved):

Kiarokh:
It's already disabled @adrianschmidt 
image
it's pretty bad. I also always become very confused. But not sure what to do

vicgeralds (approved):

adrianschmidt:
> It's already disabled @adrianschmidt image it's pretty bad. I also always become very confused. But not sure what to do

Could it be that you are creating the PR when there's just one commit, and then adding more commits to it? Because GitHub will automatically use the commit message for a PR if there's just a single commit.

But you can always edit the PR title 🙂

Issue #2161, opened 2023-03-03
input type `textarea`: the `placeholder` text does not show up


Kiarokh:
## Current behavior

All input types show the placeholder text, except the `<limel-input-field type="textarea"` />`

## Expected behavior
☝️ they should work: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#attributes

## Environment

- **lime-elements** version: 
- Framework used: 
- Logs:

Kiarokh:
@LucyChyzhova do you think you can work on this one? @john-traas would need this here: https://github.com/Lundalogik/lime-crm-building-blocks/pull/42/files#r1124478915

I'll be on leadership courses, so I won't be able to contribute.

adrianschmidt:
(Man, don't link to w3schools, you'll help their google ranking 😣 😄)

adrianschmidt:
<img width="800" alt="A screenshot of MDN's description of the placeholder attribute: 'A hint to the user of what can be entered in the control. Carriage returns or line-feeds within the placeholder text must be treated as line breaks when rendering the hint. Note: Placeholders should only be used to show an example of the type of data that should be entered into a form; they are not a substitute for a proper label element tied to the input. See input labels for a full explanation.'" src="https://user-images.githubusercontent.com/154725/223061220-45747bb4-ff01-46af-b00e-319f07004f9f.png">
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea#attributes

Kiarokh:
> (Man, don't link to w3schools, you'll help their google ranking 😣 😄)

it was already the first thing coming up 😅

vicgeralds:
> it was already the first thing coming up 😅

(yes, it's been like that for ages, and people that know better complained about it for ages, because others use is as a reference)


![Image](https://user-images.githubusercontent.com/5563232/223361453-aab6c65b-c0c2-4549-9edd-33b754b4d2bb.png)

PR #2162, opened 2023-03-06
fix(input-field): make possible to use placeholder in textarea type

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/2161#issuecomment-1455723243

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/examples/input-field-textarea.tsx
+++ b/src/components/input-field/examples/input-field-textarea.tsx
@@ -25,9 +25,10 @@ export class InputFieldTextareaExample {
 
         return [
             <limel-input-field
-                label="Text Field"
+                label="Job description"
                 type="textarea"
-                helperText="Please enter a useful message!"
+                placeholder="What is your dream job? Describe it here..."
+                helperText="This text will be displayed in your profile"
                 maxlength={MAX_LENGTH}
                 value={this.value}
                 required={this.required}
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -406,7 +406,9 @@ export class InputField {
 
         return (
             <span class="mdc-text-field__resizer">
-                <textarea {...properties}>{this.value}</textarea>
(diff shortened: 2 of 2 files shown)


src/components/input-field/examples/input-field-textarea.tsx
```diff
@@ -28,6 +28,7 @@ export class InputFieldTextareaExample {
                 label="Text Field"
                 type="textarea"
                 helperText="Please enter a useful message!"
+                placeholder="Comment text"
```
adrianschmidt, line 31: "Comment text" is a little bit confusing as a placeholder. Ideally, we would have something that makes sense as a placeholder, but I can't really think of anything. Perhaps something like `Write a message here…`?

Otherwise, something like "You can put a placeholder here…" would at least make the intention a little clearer. Although I promise someone is going to copy and paste that into their production add-on… 🙈

src/components/input-field/examples/input-field-textarea.tsx
```diff
@@ -25,9 +25,10 @@ export class InputFieldTextareaExample {
 
         return [
             <limel-input-field
-                label="Text Field"
+                label="Job description"
                 type="textarea"
-                helperText="Please enter a useful message!"
+                placeholder="What is your dream job? Describe it here..."
+                helperText="This text will be used in you profile"
```
Kiarokh, line 31:
```suggestion
                helperText="This text will be displayed in your profile"
```

Kiarokh (approved):

Kiarokh:
For me this looks good. i don't think this is a `feat` since it was expected to exist. What do you think  @adrianschmidt

adrianschmidt (approved):

PR #2167, opened 2023-03-08
feat(admonition): new component

TommyLindh2:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


blombergedvin:
FYI @FlippAre

Kiarokh:
😍 Wow… cool stuff! I'll need to do this during the cooldown. I'm going to a ski trip now

src/components/admonition/admonition.scss
```diff
@@ -0,0 +1,41 @@
+.admonition {
+    display: flex;
+    width: 100%;
```
adrianschmidt, line 3: `display:flex` already makes this element block-level, so setting `width:100%` shouldn't be necessary, and can lead to problems. Was there something that didn't work as it should without this?

https://www.youtube.com/watch?v=-st14lUQD3U

src/components/admonition/admonition.scss
```diff
@@ -0,0 +1,41 @@
...
+    }
+    &.note {
+        background-color: rgb(var(--color-gray-default));
+    }
+    &.tip {
+        background-color: rgb(var(--color-green-dark));
+    }
+    &.warning {
+        background-color: rgb(var(--color-orange-dark));
+    }
```
adrianschmidt, line 19: I'm thinking we might want to allow overriding the default colours by doing something like:

```suggestion
    &.danger {
        background-color: var(--admonition-danger-color, rgb(var(--color-red-dark)));
    }
    &.info {
        background-color: var(--admonition-info-color, rgb(var(--color-cyan-default)));
    }
    &.note {
        background-color: var(--admonition-note-color, rgb(var(--color-gray-default)));
    }
    &.tip {
        background-color: var(--admonition-tip-color, rgb(var(--color-green-dark)));
    }
    &.warning {
        background-color: var(--admonition-warning-color, rgb(var(--color-orange-dark)));
    }
```

What do you think @Kiarokh?

src/components/admonition/admonition.scss
```diff
@@ -0,0 +1,41 @@
...
+        padding-left: 0.5rem;
+        width: 100%;
+
+        .admonition-main-title {
+            margin: 0.1rem;
+            margin-top: 0.25rem;
+        }
+        .admonition-main-description {
+            margin: 0.1rem;
+            white-space: pre-line;
```
adrianschmidt, line 38: Could we support markdown here, instead of just supporting linebreaks? 🤔

src/components/admonition/admonition.scss
```diff
@@ -0,0 +1,41 @@
...
+        background-color: rgb(var(--contrast-300));
+        padding-left: 0.5rem;
+        width: 100%;
+
+        .admonition-main-title {
+            margin: 0.1rem;
+            margin-top: 0.25rem;
+        }
+        .admonition-main-description {
+            margin: 0.1rem;
```
adrianschmidt, line 37: This is a matter for @Kiarokh and @LucyChyzhova, but is it really meaningful to have a 1.6px margin?

src/components/admonition/admonition.scss
```diff
@@ -0,0 +1,41 @@
...
+    }
+    &.tip {
+        background-color: rgb(var(--color-green-dark));
+    }
+    &.warning {
+        background-color: rgb(var(--color-orange-dark));
+    }
+
+    .admonition-side {
+        mix-blend-mode: screen;
```
adrianschmidt, line 22: Very nice! I haven't seen this before. I noticed that browser support for `mix-blend-mode` on _svg_ is so-so, so we might want to just have a quick look in BrowserStack to make sure the icons don't look horrible in Safari for example.

https://developer.mozilla.org/en-US/docs/Web/CSS/mix-blend-mode#browser_compatibility

src/components/admonition/admonition.scss
```diff
@@ -0,0 +1,41 @@
...
+        background-color: rgb(var(--contrast-800));
+        padding: 0.25rem;
+    }
+    .admonition-main {
+        mix-blend-mode: screen;
```
adrianschmidt, line 27: I think I like it better without the blend here… it makes the text easier to read…

UX peeps, what do you say?

src/components/admonition/admonition.scss
adrianschmidt, line 37: This is a matter for @Kiarokh and @LucyChyzhova, but is it really meaningful to have a 1.6px margin?

src/components/admonition/admonition.scss
Kiarokh: It could be good. I agree

FlippAre:
Awesome! A concrete implementation lives in Product Lime 

Can be seen live here:

https://product.lime-crm.com/client/object/release/1005

And code: 

https://github.com/Lundalogik/solution-product/tree/main/frontend/src/components/lwc-solution-product-admonition

src/components/admonition/admonition.scss
Kiarokh: I'll have to do this during the cooldown, but have you checked how we do it in Kompendium? I don't want to re-invent the wheel 😅 https://github.com/jgroth/kompendium/blob/master/src/components/markdown/markdown.scss

src/components/admonition/admonition.tsx
```diff
@@ -0,0 +1,86 @@
...
+            case 'danger':
+                return 'fire_element';
+            case 'info':
+                return 'info';
+            case 'note':
+                return 'info';
+            case 'tip':
+                return 'light_on';
+            case 'warning':
+                return 'error';
```
adrianschmidt, line 60: The `danger` icon is nice, my only small objection is that it is perhaps a bit _too_ nice. The `warning` icon looks more dangerous than the `danger` icon 😅

FlippAre:
TBH I liked my approach of having color and icon as variables, instead of just supporting a few different modes. But that’s just my opinion
________________________________
Från: Adrian Schmidt ***@***.***>
Skickat: Wednesday, March 8, 2023 4:59:52 PM
Till: Lundalogik/lime-elements ***@***.***>
Kopia: Filip Arenbo ***@***.***>; Mention ***@***.***>
Ämne: Re: [Lundalogik/lime-elements] feat(admonition): new component (PR #2167)


@adrianschmidt commented on this pull request.

________________________________

In src/components/admonition/admonition.tsx<https://eur02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2FLundalogik%2Flime-elements%2Fpull%2F2167%23discussion_r1129672339&data=05%7C01%7Cfilip.arenbo%40lime.tech%7Cc40186e93b2c44a36ab508db1fee283f%7C15f5bb2107a442c6ada208965bd62081%7C0%7C0%7C638138879968214112%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C7000%7C%7C%7C&sdata=HlQQSBrdJU2tt3TtZDRHAhJ0zxw6z6NLRTBWkSj0TmM%3D&reserved=0>:

> +            case 'danger':
+                return 'fire_element';
+            case 'info':
+                return 'info';
+            case 'note':
+                return 'info';
+            case 'tip':
+                return 'light_on';
+            case 'warning':
+                return 'error';


The danger icon is nice, my only small objection is that it is perhaps a bit too nice. The warning icon looks more dangerous than the danger icon 😅

src/components/admonition/admonition.tsx
Kiarokh: We also have a legal problem of redistributing icons form icons8 lib.
So, we must do something about this. If we reuse icons from Kompendium, they're done by myself. So it's oK. If we wanna allow consumers to add their own icons, then we have to figure out a way to allow that. Maybe using _CSS part_ or like the lime admin hack we did for including customer logotypes.

src/components/admonition/admonition.types.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * The type of the list, omit to get a regular list. Available types are:
+ * `note`: a grayish note.
+ * `tip`: a greenish tip.
+ * `info`: a blueish info.
+ * `warning`: a yellowish warning.
+ * `danger`: a redish danger.
```
adrianschmidt, line 7:
```suggestion
 * `danger`: a reddish danger.
```

But I don't know if we really need the `-ish` suffixes… 🤔 🤷‍♂️

src/components/admonition/examples/admonition-composite.tsx
```diff
@@ -0,0 +1,68 @@
...
+                    type: 'grid',
+                },
+            },
+        };
+    }
+
+    public render() {
+        const description = this.props?.description?.replace(
+            '[type]',
+            this.props?.type ?? ''
```
adrianschmidt, line 40: Aww, I thought this was part of the component… 😞

But, I guess this wouldn't actually be very useful in normal use, and it would spend runtime resources for no good reason. So let's keep it to the example only 👍

src/components/admonition/examples/admonition-composite.tsx
```diff
@@ -0,0 +1,68 @@
...
+        );
+
+        return [
+            <limel-admonition
+                {...{ ...this.props, description: description }}
```
adrianschmidt, line 45: Haha! This double spread made my brain put a knot in itself, before I realised what's going on! 😂

Doesn't the following work? I think it's a little clearer to read 🙂

```suggestion
                { ...this.props }
                description={ description }
```

src/components/admonition/examples/admonition-composite.tsx
```diff
@@ -0,0 +1,68 @@
...
+            />,
+            this.renderForm(),
+        ];
+    }
+
+    private renderForm() {
+        return (
+            <limel-example-controls
+                style={{ '--example-controls-column-layout': 'auto-fit' }}
```
adrianschmidt, line 54: @Kiarokh I don't remember seeing this before. From the name alone, `auto-fit` sounds like a reasonable default though? What's the current default? 😊

src/components/admonition/examples/admonition-simple-warning.tsx
```diff
@@ -0,0 +1,21 @@
...
+@Component({
+    tag: 'limel-example-admonition-simple-warning',
+    shadow: true,
+})
+export class AdmonitionSimpleWarning {
+    public render() {
+        return (
+            <limel-admonition
+                type="warning"
+                description="I warn you to not go any further"
```
adrianschmidt, line 17:
```suggestion
                description="I warn you not to go any further"
```

Kiarokh:
> Awesome! A concrete implementation lives in Product Lime
> 
> Can be seen live here:
> 
> [product.lime-crm.com/client/object/release/1005](https://product.lime-crm.com/client/object/release/1005)
> 
> And code:
> 
> [Lundalogik/solution-product@`main`/frontend/src/components/lwc-solution-product-admonition](https://github.com/Lundalogik/solution-product/tree/main/frontend/src/components/lwc-solution-product-admonition?rgh-link-date=2023-03-08T15%3A58%3A30Z)

Really nice initiative that you have brought this to Lime Elements. Such UI components really belong here! 
Awesome ⭐ ⭐ ⭐ ⭐ ⭐

src/components/admonition/admonition.scss
adrianschmidt: Those are just admonitions inside markdown though, while this is a separate "admonition component". Although, one way to implement it would certainly be to stick the input in some markdown, which is then rendered with the markdown admonition plugin. One significant benefit with that would also be that if we want to support admonitions in markdown elsewhere, like in the history feed, we can use the same plugin, and thus get identical admonitions.

src/components/admonition/admonition.tsx
adrianschmidt: > We also have a legal problem of redistributing icons form icons8 lib.

Yeah, but we're not including the icons here, we only set the names of the icons to use. A consumer that doesn't have access to `lime-icons8` will just have to provide their own icons with these names.

adrianschmidt:
> > Awesome! A concrete implementation lives in Product Lime
> > Can be seen live here:
> > [product.lime-crm.com/client/object/release/1005](https://product.lime-crm.com/client/object/release/1005)
> > And code:
> > [Lundalogik/solution-product@`main`/frontend/src/components/lwc-solution-product-admonition](https://github.com/Lundalogik/solution-product/tree/main/frontend/src/components/lwc-solution-product-admonition?rgh-link-date=2023-03-08T16%3A12%3A22Z)
> 
> Really nice initiative that you have brought this to Lime Elements. Such UI components really belong here! Awesome ⭐ ⭐ ⭐ ⭐ ⭐

I wholeheartedly agree! Love this! 😍

src/components/admonition/admonition.scss
adrianschmidt: That's an implementation issue though, so we don't _have_ to change that now. However, I think we might want to check what types are supported there, and limit this component to those types in step 1. That way we can postpone the decision on whether to use markdown admonitions, without having to make a breaking change if we decide to go that route.

LucyChyzhova:
> Maybe it should have round corners instead? I don’t know.
> 
> 38C699E0-11AE-438A-BF75-280C9210EECF

Yes, rounded corners are better! 😍

src/components/admonition/admonition.tsx
Kiarokh: gotcha

src/components/admonition/admonition.types.ts
Kiarokh: that probably comes from the mix blend mode :P :D

src/components/admonition/admonition.scss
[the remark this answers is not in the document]
LucyChyzhova: I also agree we should aim for higher readability here

src/components/admonition/examples/admonition-composite.tsx
Kiarokh: To follow the confusion tradition of CSS, these are just confusing names doing confusing things that is hard to guess or remember 😅 . Here are the docs: 

https://github.com/Lundalogik/lime-elements/blob/19465a9ca45f152790496f9449e35bd2a832a22d/src/examples/example-controls.scss#L3

But we have used it before in other examples and this usage seems to be correct.

src/components/admonition/admonition.scss
Kiarokh: You're right. We should follow our [size rhythms](https://lundalogik.github.io/lime-elements/versions/next/#/DesignGuidelines/size-rhythms.md/). And I'd say let's just copy paste from Kompendium, for margins, paddings, sizes and colors where possible.

Screenshot 2023-03-09 at 09 49 23

src/components/admonition/admonition.scss
adrianschmidt: This is also something we can add later though. Better to make it simpler now, and add features _later_.

Kiarokh: I'll add this as a fixup now. For me it's easier to think about these features already when the component is being prepared and the design is being massaged. :)

src/components/admonition/admonition.types.ts
Kiarokh: I don't think colors matter here in the docs. We don't wanna keep a track of styles inside the text documentations, unless really needed for super good reasons. If we decide to change the colors, the docs will mismatch, because we will certainly forget to update them.

src/components/admonition/admonition.types.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * The type of the list, omit to get a regular list. Available types are:
```
Kiarokh, line 2: Type of "list"? Copy/Paste 🤓?

Kiarokh: I'll fixup and make the docs here more clear.

src/components/admonition/admonition.tsx
```diff
@@ -0,0 +1,105 @@
...
+    tag: 'limel-admonition',
+    shadow: true,
+    styleUrl: 'admonition.scss',
+})
+export class Admonition {
+    /**
+     * Title of the admonition
+     */
+    @Prop({ reflect: true })
+    public title?: string;
```
Kiarokh, line 38: image

I'll change it to `heading`. It makes more sense too

src/components/admonition/admonition.scss
Kiarokh: I also would prefer markdown support. Also due to this, I'd say that this `<p class="description">` shouldn't be a `<p` element. It should be just a span or something.

Is it hard to add markdown support?

src/components/admonition/admonition.scss
Kiarokh: I removed this effect due to accessibility issues. So I'm hiding these discussions.

src/components/admonition/examples/admonition-custom-heading.tsx
```diff
@@ -0,0 +1,29 @@
...
+ */
+@Component({
+    tag: 'limel-example-admonition-custom-heading',
+    shadow: true,
+})
+export class AdmonitionCustomTitleExample {
+    public render() {
+        const description = `Lorem ipsum dolor sit amet, consectetur adipiscing elit. Vestibulum viverra magna non pretium condimentum. Integer a nunc magna. In commodo elit turpis, porttitor vulputate odio pretium a. Ut at sapien a massa convallis commodo eu sed ligula. Curabitur non sodales neque. Sed id nisl vel ex tempor euismod. Nulla nec lorem dui. Cras tincidunt urna nec velit pretium maximus.
+
+        Praesent sed cursus lorem. Phasellus lobortis dolor vitae pretium bibendum. Vivamus non augue in urna consequat dapibus at quis diam. Duis tristique lacinia felis, quis condimentum urna interdum sit amet. Suspendisse facilisis pulvinar suscipit. Pellentesque quis velit feugiat, bibendum erat at, ornare tortor. Nullam sed risus a enim tempor eleifend nec quis nulla. Cras quis pellentesque justo. Maecenas id justo a eros consequat auctor.
```
Kiarokh, line 18: Even though this is a good example and shows a smart hack to get line breaks, I'd still argue that this component without Markdown support is waaaaay too weak really. For my taste, we are below an MVP level.

We can lunch it this way, but as long as adding markdown support does not add a BREAKING CHANGE. I can't assess that of course, and don't know how much time or effort is needed add this support.

A breaking change could be for instance that we think it's better to call this prop `content` instead of `description`, since we now have markdown support. (I already think `description` is a bad prop name actually.)

As soon as we lunch this, I'm sure users will soon want to add bullet point list, links, etc… in the text.

Kiarokh:
I have "_reviewed_" and since the changes that I wanted were too many, I just added them as fixups myself. I think another person should now review and approve this. 

Here is my summary:
1. From usability, and UI design perspective, I think we are in good shape now.
2. From documentations point of view, also better now. I added more docs both in the code, and in the published examples.
3. I think this component without markdown support is waaaaay too weak really. Read more [here](https://github.com/Lundalogik/lime-elements/pull/2167/files#r1146409749).

src/components/admonition/admonition.tsx
```diff
@@ -0,0 +1,124 @@
...
+     * Title of the admonition
+     */
+    @Prop({ reflect: true })
+    public heading?: string;
+
+    /**
+     * Description of the admonition
+     */
+    @Prop({ reflect: true })
+    public description?: string;
```
Kiarokh, line 50: I think we should call this `content`. This is because:
1. if we add markdown support, that'd be a more logical choice of name
2. it works better with `heading` (I had to change the `title` to `heading`, due to build warnings which I commented on earlier)

TommyLindh2:
@Kiarokh the component we made this from supports markdown by a few lines of code:

```
import { marked } from 'marked';


 <p innerHTML={marked.parse(this.config.body)}></p>
```

by using an external library called: [marked](https://www.npmjs.com/package/marked)

not sure how we handle external libraries in elements though

src/components/admonition/admonition.tsx
```diff
@@ -0,0 +1,124 @@
...
+                return 'high_priority';
+            case 'example':
+                return 'test_tube';
+            case 'quote':
+                return 'quote_left';
+            case 'question':
+                return 'help';
+            default:
+                return 'info';
+        }
```
Befkadu1, line 91:
```suggestion
      const icons: Record<string, string> = {
         danger: 'poison',
         info: 'info',
         ....
      }

    return icons[this.type] || 'info'
```

src/components/admonition/admonition.tsx
jgroth: I think this should probably be removed entirely, and all the content should be moved to the child elements slot instead. That means the consumer can format the content however they please, e.g.

```html
<limel-admonition>
    This text is <strong>bold</strong>
</limel-admonition>
```

It also means we can add stuff like markdown support later, e.g.
```html
<limel-admonition>
    <limel-markdown>
        # Lorem ipsum
    </limel-markdown>
</limel-admonition>
```

Kiarokh: Great idea! I'll fix up

src/components/admonition/admonition.tsx
Kiarokh: Awesome! I'll do

src/components/admonition/admonition.tsx
```diff
@@ -0,0 +1,124 @@
+import { Component, h, Prop } from '@stencil/core';
+import { AdmonitionType } from '@limetech/lime-elements';
+
+/**
+ * Admonitions, also known as call-outs, very useful for including supportive content
```
Kiarokh, line 5: We had a quick discussion during our standup and @vicgeralds suggested that we should perhaps change the name of this component. This idea was supported by many others, including myself. Can't we just call this component `limel-callout`? 

It's more humanly and universally understandable; and honestly, more correct.

Meaning of _admonition_: https://www.ldoceonline.com/dictionary/admonition
Meaning of _callout_: https://www.ldoceonline.com/dictionary/call-out#call-out__4


Any objections @FlippAre ? 😊

src/components/admonition/admonition.tsx
Kiarokh: @jgroth but what if they wanna use this component in let's say a start page or overview tab of the CRM, and wanna quickly add some content as a string?

Can we just keep this prop, as is (maybe rename it to `content`), and then add a `<slot name="content" />` as well, for those who want to add complex content?

jgroth: In theory it would of course be possible to do that, but the API would perhaps feel a bit weird and conflicting.
How should this scenario be handled?
```html
<limel-admonition content="Is this my content?">
    Maybe this is my content?
</limel-admonition>
```

Instead, perhaps it should be possible to configure children when configuring components in lime-admin and then it would be easy to do something like that. But that's another topic on its own :smile:

Kiarokh: Well, the case you show is a bit strange. I don't think the user will do so. :) 
If they do, the component will then show them 👇 
image


I think this discussion is a little bit like another discussion we had regarding `limel-tooltip`. We wanted to enable a `slot` for that component as well, so that consumers can put custom and complex content in it. 

I think it's great that the component gives basic functionality out of the box, but still leaves space for consumers, to help them handle complexities like this. 

This way, they can more easily make something like this: 
image

Kiarokh: honestly, I don't know what you mean 😓 but maybe you will see what I mean, when I add my fixup 😅

jgroth: My point is that we should not add both a prop and allow slotted content if they are meant to be used for the same thing, especially not if the only purpose is to make it easier to configure in lime-admin. If there is a need to be able to configure child elements/content in lime-admin, we should add support for that rather than designing an API for this component that is hard to understand.

src/components/admonition/admonition.types.ts
```diff
@@ -27,3 +27,41 @@ export type AdmonitionType =
...
+    return admonitionIcons[type] || 'info';
+}
+
+export function getHeading(heading: string): string {
+    if (heading) {
+        return heading;
+    }
+
+    return admonitionHeadings[heading] || 'Info';
+}
```
jgroth, line 67: This file should only contain types. Helper functions needs to be moved to a separate file

src/components/admonition/admonition.tsx
Kiarokh: You're right. I understand. Usage is Lime Admin was just an example. That is not why I was suggesting this approach.

Again, I want to refer to the discussion we had to the tooltip. You remember? We wanted to display hotkey visualizations (like: <kbd>Alt</kbd> + <kbd>Shift</kbd> + <kbd>S</kbd>) inside the tooltip. 

Sometimes, what's offered out of the box is not enough for the customer, and we never know what is enough or can't really support it fully. This is specially the case for components that are heavily dependent on displaying content, description, or rich-text. This component is one of those (which is why the Markdown discussions started).

So, hopefully, my last fixups and example will explain what I mean. :)

src/components/admonition/admonition.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+    public render() {
+        return [
+            <div class="side">
+                <limel-icon name={getIcon(this.type)} />
+            </div>,
+            <div class="main">
+                <h1 class="heading">{getHeading(this.heading)}</h1>
+                {this.renderDescription()}
+                <slot name="content" />
```
Kiarokh, line 64: why this `slot` has a `name`? Read more here: https://github.com/Lundalogik/lime-elements/pull/1934#issuecomment-1386922939

src/components/admonition/admonition.tsx
TommyLindh2: @Kiarokh I think your example in lime-admin could be done using the property `innerText` or `innerHTML`

Kiarokh: how do you mean @TommyLindh2 ? Can you add an example code? Or some screenshot?

src/components/admonition/admonition.types.ts
Kiarokh: OK. I'll fix ;)

src/components/admonition/admonition.tsx
```diff
@@ -0,0 +1,74 @@
...
+    tag: 'limel-admonition',
+    shadow: true,
+    styleUrl: 'admonition.scss',
+})
+export class Admonition {
+    /**
+     * Title of the admonition
+     */
+    @Prop({ reflect: true })
+    public heading?: string;
```
Kiarokh, line 45: I also believe these should be translated. Right now, they are all hard-coded in English. The fact that consumer can override them is another story, and should not be considered as a localization feature.

src/components/admonition/admonition.scss
adrianschmidt: There are too many cooks in this soup at the moment, so I'm withdrawing.

src/components/admonition/admonition.tsx
jgroth: > @Kiarokh I think your example in lime-admin could be done using the property `innerText` or `innerHTML`

`innerText` would work, but it feels a bit hackish :smile: 
`innerHtml` should almost never be used, and here it would render any HTML as text instead of creating nested tags

src/components/admonition/admonition.tsx
TommyLindh2: I like this idea, much easier to understand

src/components/admonition/admonition.tsx
TommyLindh2: I realized that you cannot use `innerHTML` and `innerText` on custom components, so ignore my comment.
But you cannot add a lime element directly in the view-config anyways, so you'll need a wrapper component, and then you can add your own content property. 

I don't think this will be an issue.

src/components/admonition/admonition.tsx
Kiarokh: Nice. I'll change them all then.

src/components/callout/callout.e2e.ts
```diff
@@ -0,0 +1,70 @@
...
+        await page.setContent('<limel-callout type="warning"></limel-callout>');
+        const element = await page.find('limel-callout');
+        expect(element).toHaveClass('warning');
+    });
+
+    it('renders a callout with the correct icon based on type', async () => {
+        const page = await newE2EPage();
+        await page.setContent('<limel-callout type="example"></limel-callout>');
+        const element = await page.find('limel-callout');
+        const icon = await element.find('limel-icon');
```
Kiarokh, line 57: when I run the tests, I get this `console.error`s
image
So I added these 3 files to `/src/icons`. Hope it's OK.

Kiarokh: the question is if these svgs should have content, or should be just dummy empty svg fiels

adrianschmidt: You should _definitely not_ add any icons from icons8 to this repo, ever, not even in a branch. If that's what you did, they are now public, and need to be removed asap. If you need files to make the tests pass, you can duplicate one of the existing files in `src/icons/` and just rename it. Those files contain a royalty-free image of some yoga-dude.

adrianschmidt: I've overwritten the files you added with the yoga-dude. Since there are references to the pushed commit on this PR, to be really thorough, we'd have to purge those from GitHub's history too, bit that's a massive pita. Hopefully those links stop working after a while anyway…

adrianschmidt: Oh noes… they keep working forever… I just checked on a PR which was merged and had the branch removed in 2018, and a commit that was replaced by a force-push was still accessible 😖

It's a very minor issue with only three icons though, so I don't think we have to worry too much about it. But do use dummies next time 🤗

Kiarokh: Ah, thanks. Well, even a monkey with basic Adobe Illustrator skills can create the icons that I copied from them here. And they weren't really exactly the same icons as their. But good to know this. Thanks. 

I've now changed their contents to things that I personally created before.

src/components/callout/callout.e2e.ts
```diff
@@ -0,0 +1,77 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('limel-callout', () => {
+    it('renders a callout with the default "Info" heading and no description', async () => {
+        const page = await newE2EPage({ timeout: 10000 });
```
Kiarokh, line 5: I added unit tests with help of Chat GPT. Perhaps I'm officially the first one at Lime who does this? 😅 
Anyhow, its suggestion was adding `timeout: 10000` to all tests. Is it too long? Does it make `npm run test` run slowly? Is there a better way? You tell me. I'm not a good test writer.

Actually, don't tell me. Just fix it up please. I don't care. ;)

vicgeralds: > I added unit tests with help of Chat GPT. Perhaps I'm officially the first one at Lime who does this? 😅

way to go!

> Is it too long?

yeah, 10 secs seems excessive. I'm not familiar with the e2e tests yet, so I might leave that for someone else 😊

Kiarokh: This was what the AI said: 👇 😄  (First, there was not timeout.)
image

Kiarokh:
Questions left to answer:

- [ ] should we keep `public description?: string;` or is `slot="content"` enough?
- [ ] make the language picker work in the composite example 
- [ ] get translations reviewed if possible (for `fi`, `no`, `nl`). If not, we just move on. This shouldn't be show stopper. I've already tried to reach folks and ask on [Teams](https://teams.microsoft.com/l/message/19:802ebb7143e841d782d4920a5dc6258d@thread.skype/1679912474019?tenantId=15f5bb21-07a4-42c6-ada2-08965bd62081&groupId=5936a987-61f4-4688-8cba-c242d841c111&parentMessageId=1679912474019&teamName=Lime&channelName=Translation%20help&createdTime=1679912474019&allowXTenantAccess=false).

jgroth:
> Questions left to answer:
> 
> * [ ]  should we keep `public description?: string;` or is `slot="content"` enough? This would require rewiriting the ChatGPT tests though, and I'm not willing to play around with tests. :D
> * [ ]  make the language picker work in the composite example
> * [ ]  get translations reviewed if possible (for `fi`, `no`, `nl`). If not, we just move on. This shouldn't be show stopper. I've already tried to reach folks and ask on Teams

1. No, remove `description` I don't think it's needed
2. It's weird that this component has a language prop in the first place, the consumer should not have to think about stuff like this and provide the language to use all over the place. We really need to fix this over all components in Lime Elements. We should either read the browser language, or as a fallback let the consumer override it by setting some setting in the `limel-config` component

src/components/callout/callout.e2e.ts
adrianschmidt: I don't think Chat GPT is good enough to help beginners to write code like that. It still makes too many mistakes. Fun to try though, and it certainly won't be long before it performs _much_ better…

Copilot X for GitHub was revealed recently, and from the little I saw in a video, it seems very competent, and since it lives in VS Code (or other IDE), it automatically has basically the whole repo as "context" for the suggestions it makes.

adrianschmidt: For now though, I think you're better off looking att our existing code and asking for help 😊

I know you asked us to just fix it. If it's still here when I get around to it, I will, but I have other stuff too.

adrianschmidt:
> 2. It's weird that this component has a language prop in the first place, the consumer should not have to think about stuff like this and provide the language to use all over the place. We really need to fix this over all components in Lime Elements. We should either read the browser language, or as a fallback let the consumer override it by setting some setting in the `limel-config` component

We could default to reading from the browser, but it has to be possible to override, because if your app doesn't support Finnish, but Lime Elements does, you might not want random elements of your app display stuff in Finnish, when the rest of your app is in English, for example.

I hadn't thought about the possibility to override via the config component though! 🤯 That's brilliant! And kind of obvious now that you've said it 😂 👌

src/components/callout/callout.e2e.ts
TommyLindh2: Removed the timeout and do it like the limel-button does

TommyLindh2:
> > 2. It's weird that this component has a language prop in the first place, the consumer should not have to think about stuff like this and provide the language to use all over the place. We really need to fix this over all components in Lime Elements. We should either read the browser language, or as a fallback let the consumer override it by setting some setting in the `limel-config` component
> 
> We could default to reading from the browser, but it has to be possible to override, because if your app doesn't support Finnish, but Lime Elements does, you might not want random elements of your app display stuff in Finnish, when the rest of your app is in English, for example.
> 
> I hadn't thought about the possibility to override via the config component though! 🤯 That's brilliant! And kind of obvious now that you've said it 😂 👌

As of now we don't have another way than to pass the language.
When Lime elements support this feature that you're talking about, we should fix it in all affected elements.

So I would that this discussion is off topic in this PR.

TommyLindh2:
@jgroth description is now removed.

src/components/callout/callout.helpers.ts
```diff
@@ -0,0 +1,40 @@
...
+        return heading;
+    }
+
+    const key = `callout.${type}`;
+
+    try {
+        return translate.get(key, language);
+    } catch (error) {
+        return translate.get(key, 'en');
+    }
```
Kiarokh, line 39: @TommyLindh2 Don't forget to write about why you did this hack, and that we think this is not the place to do this hack.

From what I remember, if someone chose for example `fr`, `de`, or `nb` from the dropdown, the example crashed. Because `fr.po` and or `nb.po` files do no exist in Lime Elements, but the list of languages states that they are supported.

TommyLindh2, line 39: I did this because the component breaks if a language is given that doesn't have a translation.
So I will catch that error and default it to english.

I'm not sure that this component should handle this, maybe it would be better to have it in the translation function.
But this works and won't affect any other components, so that's why I did it

src/components/callout/callout.e2e.ts
```diff
@@ -0,0 +1,49 @@
+import { newE2EPage } from '@stencil/core/testing';
```
adrianschmidt, line 1: Nice! Seems like we have this in a pretty good place now 👍

The one thing I would like to change before we merge this, is that I would like to limit the number of supported types.

It's super-easy to just add more supported types in the future, but once added, removing one is a breaking change, and once it's widely used by Lime CRM add-ons, we basically can't ever remove it.

I'm worried that this many types will be more confusing than helpful, and I'm not sure all of the current types are well thought through. Why is `tip` green, but `success` blue? Do we _really_ need a `danger` when we have `warning`, and if we do, why is it the second most muted colour, after only `quote`? Should we really have a `quote` type when the semantic element to use for block quotes is `<blockquote>`?

The following are the ones supported by default by the plugin we use in Kompendium. I don't _love_ them, but at least it's a limited set. Could we go for something a little closer to those five, rather than our current _ten_?

image

adrianschmidt: If you're OK with this @Kiarokh and @TommyLindh2, I can make the change.

adrianschmidt (changes requested):

src/components/callout/callout.tsx
```diff
@@ -0,0 +1,86 @@
...
+    public render() {
+        return [
+            <div class="side" role="presentation">
+                <limel-icon name={getIcon(this.icon, this.type)} />
+            </div>,
+            <div class="main">
+                <h1 class="heading">
+                    {getHeading(this.heading, this.type, this.language)}
+                </h1>
+                <slot name="content" />
```
jgroth, line 82: Is there a reason to why we named the slot? Since there is only one slot it will be easier to use if the consumer don't have to add the `slot` attribute to their tags.

TommyLindh2: I agree that an unnamed slot is much easier to use, @Kiarokh said that there was some kind of discussion regarding this that led to that all slots should be named in lime-elements.

But for me it doesn't matter.

src/components/callout/callout.e2e.ts
TommyLindh2: It doesn't matter at all for me, please do what you guys think is best.

src/components/callout/callout.tsx
jgroth: Will it work to do something like this? Otherwise, you would be forced to add an inner tag as well just to specify the slot which is not ideal
```html
<limel-callout>Hello, world!</limel-callout>
```

Kiarokh: https://github.com/Lundalogik/lime-elements/pull/2167#discussion_r1147762984

jgroth: > Or we don't name the main slot (the content slot), but name every other slot.

If there's more than one slot, one of them needs to be named or it won't work

For all cases where it's quite obvious what the main content is (like in this component), I see no reason to why we need to name the slot

adrianschmidt: I'm fine with not naming this slot, as it seems pretty obvious that if more slots are ever added, this will remain the main one. The problem was in some other case, where we had only one slot, but that slot was for action buttons or something, and then, later, we wanted to add another slot for something like the content of the "body", and all of a sudden we had a situation where the default slot was very much an aside kind of thing, while the clearly main slot was named 🙈

> Will it work to do something like this? Otherwise, you would be forced to add an inner tag as well just to specify the slot which is not ideal
> 
> ```
> <limel-callout>Hello, world!</limel-callout>
> ```

I think this argument is powerful. Let's remove the name 👍

I'll fix it, since I'll remove some of the extra types anyway.

jgroth: :a:

adrianschmidt: ⚡

TommyLindh2:
I’m sorry, but why did we remove the possibility to set a custom icon and header?

1. Be able to set custom is really nice to be able to follow our core value make a mark.
2. It’s a fairly low complex feature, or am I wrong?

adrianschmidt:
By the way, the original commit, with all the extra features I stripped out, is still available on the https://github.com/Lundalogik/lime-elements/tree/callout-backup branch.

TommyLindh2:
> > I’m sorry, but why did we remove the possibility to set a custom icon and header?
> 
> Because _adding_ features is the easiest thing in the world. But every feature, however small, comes with a cost. That's a cost of added complexity and added maintenance.
> 
> _Removing_ already released features, on the other hand, is disproportionally difficult.
> 
> This very simple component has now been in review for three weeks. That's frankly absurd for how simple the component should be. The easiest way to cut down on review time is to trim everything that isn't strictly necessary. Always strive to think a new component through, to find the cases that must be included, then remove everything else. Then see if you really can't trim it a bit more. Then come back, and do the same thing one additional feature at a time. Once this has been merged, we can absolutely consider adding support for custom heading and icon.
> 
> > 1. Be able to set custom is really nice to be able to follow our core value make a mark.
> 
> I think it's highly debatable if the best way for our end-product to make a mark is for us to give our appers every conceivable choice. But that's not the point here. The point isn't that we should never support custom types. The point is to slim this PR down as much as possible, so we can get it out of this review purgatory…
> 
> > 2. It’s a fairly low complex feature, or am I wrong?
> 
> There are no low complexity features. If you think you've found a low complexity feature, I guarantee you've just not found the complexities yet 😞
> 
> But again, this is not about preventing the feature, it's about helping the component get into the product. Then we'll look at features.
> 
> EDIT: In fact, I was on the fence about the custom colours, but had landed on letting them stay. But to be consistent, I'm going to cut those out too, even though I know it will make Kia angry with me. Just make smaller PRs, and they'll go through quicker.

Thanks for a clear and good answer. I buy everything you say :-) Expect a new PR in the near future ;-)

adrianschmidt (approved):

src/translations/nl.ts
```diff
@@ -1,4 +1,9 @@
 export default {
+    'callout.note': 'Opmerking',
+    'callout.important': 'Belangrijk',
+    'callout.tip': 'Tip',
+    'callout.example': 'Waarschuwing',
```
Kiarokh, line 5: This key should have been 'callout.caution'

adrianschmidt: Ah crap… 😅

Good find! I'll fix 👍

adrianschmidt: Fix here: https://github.com/Lundalogik/lime-elements/pull/2204

src/components/callout/examples/callout-caution.tsx
```diff
@@ -0,0 +1,16 @@
...
+ * Type: `caution`
+ */
+@Component({
+    tag: 'limel-example-callout-caution',
+    shadow: true,
+})
+export class CalloutCautionExample {
+    public render() {
+        return (
+            <limel-callout type="caution">I hope you read this.</limel-callout>
```
Kiarokh, line 13: 😄 way better than Lorem Ipsum! Thanks

Kiarokh:
Thanks for the review Adrian. 
🥰
🙏 

I hope what I write below doesn't come as an unfriendly complain. 😊 I just want to reflect a few of my thoughts, and I might be wrong in my perception.

When I came to Sweden as s student, I had a fellow country mate whose passport expired. He was a student too. He sent his passport to the embassy to renew it, and they posted the new passports back to a post office in Lund. Then he went there, showed the notice from the postman, and asked them to give him the package. 

The person who worked there asked him to show his ID card. He said I don't have any other ID card. The only ID card which I have, that's accepted in Sweden, is my passport in the envelop that you are holding. She said, I'm sorry. I can't give you the package. I need to see your ID card! That's it.

He tried several days, asked her many times, brought his friends and neighbors with himself in the post office to testify that he was really himself, brought letters from the university, he did everything. But she wouldn't give him the envelope. The last time, she told him: "At this point, I know that you are really the person that you say you are. But I'm sorry, I can't give you the package. This is the rules, I must see your ID card! That's it." My friend was upset, furious, frustrated, and totally helpless. He went to the post office every day, until finally one day, she wasn't there. Then he received the package from her colleague instead.


I understand the argument that smaller PRs are better _because_ you can add features at any time, etc, etc.. and following conventions is useful. But rules and conventions shouldn’t turn into religiously valuing a _process_. For me, the goal is not following processes, the goal is getting good results.

Please help me understand this... Why put all that effort first in understanding how all the code in the PR works. Then removing our already-done-work (which you consider unnecessary for a first PR), rewriting the tests, and refactoring stuff? How will this help in any way, if we are just gonna redo the exact same thing again next week, copy/pasting the exact same code from a backup? 😔 

Maybe we didn't do this according to the holly book of best practices in software development. Maybe this was too much code for what's considered to be necessary or good-enough for a "first PR" for a new component. But dude, you basically reviewed the the whole thing to just remove some parts of it. 😞 To me this sounds like “I want to punish you because your PR was too big and that’s not by the book!”. 

Now I don’t mean this is your intention! But I can't help not experiencing it this way (assuming that you must have gone through all the code once 😅 to be able to remove it).

> This very simple component has now been in review for three weeks. That's frankly absurd for how simple the component should be. 

There is no doubt that a small PR is easier to review. But the fact that this PR stayed in review column for a while has _absolutely_ nothing to do with those props and features that were removed now. If you remember, I was on vacation for one week, and this PR happened mid-cycle just before that. Then the cooldown started with a ton of things we needed to do came back up. We all work on different things in parallel, and it's very hard to get everyone that has an opinion (or should give their opinion) about this PR in sync. Specially when we have to do that through GitHub notifications with optional time to answer and comment on stuff.

If we had already not done the work and were still arguing whether X or Y features should be there or not; then we could say "hey, let's leave this for later, we can't decide now and don't have to fix or add it." 

Now I totally understand that you may not know all the background story behind this PR. But at least four different people worked on it before (which also mean it was reviewed by them too already). But I feel I have to tell the tale, so you know why those features must have been there. We had reasons for those decisions; otherwise we wouldn't do them. 

It's fine if you didn't know, but I wish you asked us before removing them. 

Unfortunately now, this component in its current form is still not useful for the folks who needed it. 

When you go to the [link](https://product.lime-crm.com/client/object/release/1005) in Flip's [comment](https://github.com/Lundalogik/lime-elements/pull/2167#issuecomment-1460399337) this comes up 👇 
image
So we wanted to replace the UI parts of their component, with a Lime Elements component instead. This is why this PR was born originally. But with the merged PR, it's not possible to generate what we wanted in that solution. Which is for example at least the ability to add a custom icon, and a custom heading.

Furthermore, all Admonition components that I have seen offer the possibility of adding a custom header. Additionally, I as a user have used the possibility of custom headings countless times in my docs. So that's the bare minimum of such component. See this from your own example 👇 
image
Having a custom icon is more rare. Perhaps one could argue that. But I can't motivate why it is a bad idea neither, given how we/consultants/appers do things in the CRM (and given the original use case). Having icons was good, specially because you decided to reduce the `type`s which I had put in initially. 

This reminds me of the story of removal of the Alternative UI for Progress Flow: https://github.com/Lundalogik/lime-elements/pull/1327 With the same kind of reasoning, that too was removed just before merging. I still see cases where add-ons and appers would need it in their creations, and they instead have to use the existing progress flow and lock its steps (make it non-interactive) to use it for visualization-only. Here 👇 is an example of it, I just took the screen shot during the last hack Tuesday event.
Screenshot 2023-03-30 at 09 37 07


For the Progress Flow, firstly, I become extremely demotivated and I generally really **_HATE_** doing the same thing twice. Later I never had the energy to add the alternative UI again, because as we changed things through time in the main UI of the component, the structure and styles were changed and today they differ from that excluded backup version of the PR. The difference is to such an extent that I don't even want to go back and try to understand what's different and what's working, and adapt the old code to the newer code, resolve conflicts, etc… I actually tried once, and gave up. While if it was there from the beginning, any changes happening would already be verified with that "alternative UI" too, so that would also grow with the component itself. Secondly, I don't want to bring up these "design" issues for review, and write a book every time to motivate why they should be there.

Regarding the **CSS variables.** We definitely needed the variable for the background color. This was a measure to boost the quality of the original PR. In fact we need such a variable in most of our components. For instance I even added it to `limel-button-group` a while ago. It's not that I enjoy giving too much design freedom to consumers. Some consumers easily eff-up their UIs with bad graphical design decisions. That's unfortunate, but that's not the fault of us providing flexible tools to design a UI with. That's the fault of the consumer. If someone kills another person with a  knife or hurt themselves with it, that's not the fault of the knife-maker who made the tool too sharp. In Lime's context, all bad UI designs are the fault of the company which doesn't _want to_ hire designers for their projects, and rather asks consultants to be designers, managers, marketing and sales representative, etc….

 _So back to the CSS variable story for the background..._ it is needed because we have a minimalistic flat UI design language. It means elements are usually perceived and differentiated from each other by very minimal visual details, such as their background colors (and not-so-often by the shadows around them, or borders). So not being able to change the default hard-coded background color from `--contrast-300` means you can not use this component on a background color that is `--contrast-200`, `--contrast-300` or `--contrast-400`, due to lack of contrast between the surrounding UI and this component. So, the consumer should have the possibility to tune that color. And with that, there comes a natural need to tune the contrast of the text too (to make it readable on a darker or lighter background).

I can't write this ☝️  down for each and every PR or feature that is about CSS variables or new components. I expect my colleagues to either know this, or trust me that some of my choices are due to my professionalism, not naivety or being a trigger-happy with offering features. I also expect them to ask or challenge me, before just considering something unnecessary instinctively.

I have more reasons for the other CSS variables, and the removed types too. But I simply don't have the energy to argue for everything and prove myself every time. So I back off and let things go, even if I believe they are needed. I already feel we have an extremely bureaucratic process of working which is anti-creativity and counterproductive in many ways. I just don't want my wishes to be in the way too much and make things even harder. But that doesn't mean I like this way of working or doing things. I'm often alone in my side of the game, and already fight too many fights with too many players... so no energy for more struggle. 

To sum up my comments…
I believe rules and conventions are there to help people organize generally, and provide guidelines and make one think about why something is expected to be in a certain way. The goal shouldn't be literally following rules. It should rather be achieving great results, and actually sometimes smartly ignoring, bending and bypassing rules. Once you realize why a stop sign is invented and used, there will be no point in fully stopping the car and make sure it stands still for a few seconds before a stop sign in the middle of nowhere on a rural road, at 02:00AM in a summer night, when you see all the conditions are right and there is no sound or sight of a car nearby. 

We have components and solutions that are written following all rules and best practices of coding, but users still get confused when trying to use them. I wish we were more sensitive about these things, instead of being sensitive about following abstract coding standards. 

And I wish I was more trusted when it comes to features and design choices, and I was asked more often to clarify things, rather than to be assumed wrong.

Anyhow, let's not make this a big thing. It's not. I felt just frustrated and demotivated, specially since I know it'll have to be me who is making the new PRs and do all the boring bookkeeping again. This is while comparing code (text) is way out of my comfort zone. It's like a torture for my brain 😅

src/components/callout/examples/callout-caution.tsx
adrianschmidt: I totally stole that from [remark-admonitions](https://github.com/elviswolcott/remark-admonitions#infima-docusaurus-v2), and yes, I really like it too! 👍

adrianschmidt:
Regarding Kia's comment: we're talking about it in a huddle, (we're currently pausing because of other meetings, but we'll resume our conversation), and I might come back and sum up some thoughts here afterwards. I just didn't want to leave this thread looking like I didn't respond.

One clarification I can make already is that my intention was not that the extra features should not be added, or should not be added very soon. Quite the opposite. I wanted to get the part everyone seemed to agree on merged asap, and expected that it will be followed up with the additional features today or throughout this week 🙂

Kiarokh:
Thanks a ton Adrian! ❤️  🤗 You're the best.

I'll try to become better in making more minimal PRs whenever I can in future. ;)

adrianschmidt:
PRs adding functionality removed from this one:

#2206
#2207

PR #2194, opened 2023-03-28
fix(snackbar): make snackbar's label center alignment

LucyChyzhova:
When a `snakbar` has only a few words we have way too much empty space on the right side, I would suggest aligning a label to the center.

Before:
Screenshot 2023-03-28 at 13 30 17

Screenshot 2023-03-28 at 13 24 15

After:
Screenshot 2023-03-28 at 13 18 36

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


adrianschmidt (approved):
Nice and simple 👍

Kiarokh:
Sorry I cam in too late before you approved this. 

This change looks nice in this example. But when the snackbar has buttons on the side, then this wouldn't look so nice anymore and alignments would be graphically off.

image

Instead of this approach, we can do one of the following

- make sure that the width of the snackbar is more dynamic based on the content. 
- or make sure that these styles are applied only if the text content is "small" and there are no buttons

LucyChyzhova:
Good point! Thanks, Kia! Yes, you are right,  I will fix it and will make a new pr.

PR #2195, opened 2023-03-28
fix(button): hide the icon when `has-reduced-presence` class is true …

LucyChyzhova:
…& the button is `disabled`

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/partial-styles/_has-reduced-presence.scss
+++ b/src/components/button/partial-styles/_has-reduced-presence.scss
@@ -34,7 +34,8 @@ $speed-of-reducing-presence: 0.3s;
             }
             limel-icon,
             limel-spinner,
-            svg {
+            svg,
+            .mdc-button__icon {
                 transition: all $speed-of-reducing-presence ease;
                 transition-delay: $speed-of-reducing-presence;
 


src/components/button/partial-styles/_has-reduced-presence.scss
```diff
@@ -34,7 +34,8 @@ $speed-of-reducing-presence: 0.3s;
             }
             limel-icon,
             limel-spinner,
-            svg {
+            svg,
+            .mdc-button__icon {
```
adrianschmidt, line 38: Looks good and safe, but I can't reproduce it, presumably because all the icons in the docs examples are svgs, so are covered by the old selector anyway? Where or how did you discover this was a problem? 😄

adrianschmidt (approved):

src/components/button/partial-styles/_has-reduced-presence.scss
LucyChyzhova: I wanted to work on our action buttons slot, mainly to improve consulters' experience and UI of this area, and faced with some empty divs even if some actions (like reset or save) are not presented. It's here:

Screenshot 2023-03-29 at 09 59 06

adrianschmidt: Ah! I see! Nice! 👍

PR #2201, opened 2023-03-30
make mixins better so that they can be imported directly from `node_modules` folder in other repos.

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/3293

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -1,3 +1,12 @@
+/**
+ * Note! This file is exported to `dist/scss/` in the published
+ * node module, for consumer projects to import.
+ * That means this file cannot import from any file that isn't
+ * also exported, keeping the same relative path.
+ *
+ * Or, just don't import anything, that works too.
+ */
+
 @mixin visualize-keyboard-focus {
     &:focus {
         outline: none;
--- a/stencil.config.dist.ts
+++ b/stencil.config.dist.ts
@@ -6,7 +6,10 @@ export const config: Config = {
     outputTargets: [
         {
             type: 'dist',
-            copy: [{ src: 'style/' }],
+            copy: [
+                { src: 'style/' },
+                { src: 'style/mixins.scss', dest: '../scss/mixins.scss' },
+            ],
         },
     ],
     plugins: [sass()],


adrianschmidt:
Please update the PR title so it doesn't look like there's only one commit in the PR 🙏

src/style/mixins.scss
```diff
@@ -156,14 +154,14 @@
 
     cursor: pointer;
 
-    height: functions.pxToRem(20);
-    width: functions.pxToRem(20);
+    height: 1.25rem;
+    width: 1.25rem;
```
adrianschmidt, line 162: I don't want to be too nitpicky, and since both commit types behave the same, you don't have to change it, I'm just commenting here for the sake of knowledge sharing 🤗:

This is a textbook example of a `refactor`—the end result, the "behaviour", is _identical_, but you've change how the code gets us there 😊

Kiarokh: Thanks. I thought we don't wanna mess up the change log too much for the imaginary consumers. But maybe refactor won't go there neither? or?

src/components/select/select.scss
```diff
@@ -68,7 +68,6 @@
         display: inline-flex;
         align-items: center;
 
-        cursor: pointer;
```
adrianschmidt, line 71: There doesn't seem to be a matching mixin here? Not saying you did anything wrong, just wondering if there's something I'm missing, or if this was just not needed anymore? 🙂

Kiarokh: Good catch! I actually checked, it's further up or down in the code in the same file ;)

src/style/mixins.scss
adrianschmidt: Exactly, refactor also doesn't go there, because it's not something any consumer needs to know about 👍

stencil.config.dist.ts
```diff
@@ -6,7 +6,7 @@ export const config: Config = {
     outputTargets: [
         {
             type: 'dist',
-            copy: [{ src: 'style/' }],
+            copy: [{ src: 'style/' }, { src: 'style/mixins.scss', dest: '../../mixins.scss' }],
```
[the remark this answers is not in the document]
adrianschmidt: Since the `dist` folder already has folders like `cjs` ("common JS", aka node's kind of javascript), `esm`, and `types`, it feels reasonable to put a folder called `scss` there too, and put any exported scss files in that.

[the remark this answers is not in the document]
Kiarokh: This looks good to me of course. The point was not copy-pasting manually. If we can get the same file automatically from Elements, that'd be awesome. 

Do you want me to accept the suggestion and change all the import links?

The only thing is how to make sure that people don't add new mixins into this file? Is it an obvious thing, or does it need some comments where the file is imported?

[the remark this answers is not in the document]
adrianschmidt: (So I'm doing that 😄. Will push in a few minutes.)

src/components/select/select.scss
adrianschmidt: Ah, yeah, it's probably this one:

image

src/style/mixins.scss
adrianschmidt: Since I was fiddling around anyway, I changed the commits from `chore` to `refactor`.

stencil.config.dist.ts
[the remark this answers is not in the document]
adrianschmidt: > Do you want me to accept the suggestion and change all the import links?

Sorry, I didn't see that you had written this. You don't have to do anything, I've already updated all the imports in both repos.

> The only thing is how to make sure that people don't add new mixins into this file? Is it an obvious thing, or does it need some comments where the file is imported?

Ah, good point. Adding new mixins isn't a problem in itself, but they can't import anything from any file that isn't also exported, since the compilation of the file won't happen when _this_ repo is built, but when the _consumer_ repo is built.

So, yeah, I guess a comment about that might be good to have. I'll add something! 👍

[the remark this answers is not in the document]
adrianschmidt: ⚡

adrianschmidt (approved):

PR #2214, opened 2023-04-04
Fix date and time formats

### LINKED ISSUE #1489: date-picker: `format` property doesn't work as documented when date-picker is used inside limel-form
[Originally reported](https://forum.lime-crm.com/t/show-date-only-in-date-picker-in-web-component-schema/4141) by @amandaolsson.

## Current behavior

image

1. It's already reproduced for you, right in the first example of the docs for `limel-form`: https://lundalogik.github.io/lime-elements/versions/next/#/component/limel-form/

## Expected behavior

- ***Either*** the configuration is correct, and the date-picker should only display a date
- ***or***, the configuration is incorrect and should be fixed (I'm thinking maybe `format` has to be set on `lime.component.props`?)

## Environment

- **lime-elements** version: Verified in `v34.1.0-next.18` and `v32.4.4` (latest version before we updated MDC).
- Framework used: Reproducible in the docs.
- Logs:

### LINKED ISSUE #2232: Date-only strings are displayed as local time but parsed as UTC
The date-picker widget used in limel-form parses date format strings as UTC, by just passing them to the Date constructor, but displays and handles dates as local time after that.

This is usually not a problem in Europe where we have a positive time offset, for example: 

```
> new Date("2023-04-13")
Thu Apr 13 2023 02:00:00 GMT+0200 
```

but in the US, the local date will be 12 April.

vicgeralds:
This fixes some issues in `limel-form` when using the date picker widget or when using a custom component for 
the string formats `date-time`, `date` or `time`.

The `type` date picker prop is now set depending on the schema format (`date-time` or `date`) in the date widget. This has been broken since c290470676896cc965e2ac3b487fae6783100e6e

Fixes #1489
Fixes #2232

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/widgets/date-picker.ts
+++ b/src/components/form/widgets/date-picker.ts
@@ -1,23 +1,26 @@
 import moment from 'moment/moment';
 import React from 'react';
+import { DateType } from '../../date-picker/date.types';
 import { WidgetProps } from './types';
 import { LimeElementsWidgetAdapter } from '../adapters';
-import { FormSchema } from '../form.types';
 
-export class DatePicker extends React.Component {
+type DateWidgetProps = WidgetProps & {
+    type?: DateType;
+};
+
+export class DatePicker extends React.Component<DateWidgetProps> {
     public refs: any;
     public state = {
         modified: false,
     };
 
-    constructor(public props: WidgetProps) {
+    constructor(public props: DateWidgetProps) {
         super(props);
         this.handleChange = this.handleChange.bind(this);
     }
 
     public render() {
-        const props: WidgetProps = this.props;
-        const additionalProps = getAdditionalProps(props.schema);
+        const { type = 'datetime', ...props } = this.props;
 
         return React.createElement(LimeElementsWidgetAdapter, {
             name: 'limel-date-picker',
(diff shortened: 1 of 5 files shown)


src/components/form/widgets/date-picker.ts
```diff
@@ -61,7 +61,7 @@ export class DatePicker extends React.Component<DateWidgetProps> {
         const formatMapping = {
             'date-time': 'YYYY-MM-DDTHH:mm:ssZ',
             date: 'YYYY-MM-DD',
-            time: 'HH:mm:ss',
+            time: 'HH:mm:ssZ',
```
vicgeralds, line 64: The meaning of time zone offset when converting to local time is ambiguous without knowing the date, so I don't think it makes sense to include the time zone for the time picker. Unless we made it possible to also provide the date somehow 🤔

vicgeralds: da212df8b1a655fe65fe50395edfc8f4b32f776c

adrianschmidt (approved):
Looks good to me 👍

PR #2222, opened 2023-04-11
fix(slider): update the value when releasing the drag handle

Befkadu1:
It happens when you click and drag the slider, but don't click on an empty area of the screen to "release" it. See the attached video in the issue

fix Lundalogik/crm-feature#3282

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/slider/slider.tsx
+++ b/src/components/slider/slider.tsx
@@ -105,7 +105,6 @@ export class Slider {
     private percentageClass: string;
 
     public constructor() {
-        this.inputHandler = this.inputHandler.bind(this);
         this.getContainerClassList = this.getContainerClassList.bind(this);
     }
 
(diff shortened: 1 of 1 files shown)


Befkadu1:
The issue is in the MDC component. They have only 2 events `change` and `input`. If they have an event like `mouseup` it would have been easy to set the value while leaving the slider component. We could have emitted the changes in the `inputHandler()` function [here](https://github.com/Lundalogik/lime-elements/blob/a665fa199a53fb0957976b3bccc61a5e4462d853/src/components/slider/slider.tsx#L331) but this makes it difficult to set the value of the slider also.
 
It seems it is difficult to find a fix because of that I am closing this issue.

@Kiarokh you can add your point if I forget something

PR #2223, opened 2023-04-11
fix: trigger a resize event to get the slider to calculate correctly

omaralweli:
doesn't really fix [#3287](https://github.com/Lundalogik/crm-feature/issues/3287)

## Review:

### DIFF
--- a/src/components/slider/slider.tsx
+++ b/src/components/slider/slider.tsx
@@ -10,6 +10,8 @@ import {
     Watch,
 } from '@stencil/core';
 import { getPercentageClass } from './getPercentageClass';
+import { dispatchResizeEvent } from '../../util/dispatch-resize-event';
+
 /**
  * @exampleComponent limel-example-slider
  * @exampleComponent limel-example-slider-multiplier
@@ -165,6 +167,9 @@ export class Slider {
         }
 
         this.createMDCSlider();
+
+        const waitForUiToRender = 100;
+        setTimeout(dispatchResizeEvent, waitForUiToRender);
     }
 
     public componentWillLoad() {


src/components/slider/slider.tsx
```diff
@@ -165,6 +167,9 @@ export class Slider {
         }
 
         this.createMDCSlider();
+
+        const waitForUiToRender = 100;
+        setTimeout(dispatchResizeEvent, waitForUiToRender);
```
jgroth, line 172: Using `setTimeout` together with rendering is almost always a hack, especially when any other value than `0` is used. 

I'm not sure what the correct solution here is since I don't know the issue in detail, but here are some suggestions that could be worth considering

1. Is it possible to call `dispatchResizeEvent` in `componentDidRender` instead? We probably don't need `setTimeout` at all then.
2. `requestAnimationFrame` can be used instead of `setTimeout` if code needs to be run before the next browser paint cycle
3. If we need to track if something has changed size, `ResizeObserver` could be worth looking into

Kiarokh: I hope what I say is helpful. I'm not so good with these kinds of things.

I think you should do something like what we are doing here:

https://github.com/Lundalogik/lime-elements/blob/e29080eb7002839018793ffbf3a070afe38c262f/src/components/dock/dock.tsx#L116

However, not for the entire viewport, but for the container. I think we should observe the container, and see whether it has changed the size or not. And as soon as the size has changed, we re-render the component.

We do something similar in `limel-tab-bar.tsx` as well.

src/components/slider/slider.tsx
```diff
@@ -165,6 +167,9 @@ export class Slider {
         }
 
         this.createMDCSlider();
```
Kiarokh, line 169: I was checking [MDC's html structure](https://m2.material.io/components/sliders/web#sliders), and saw that they have a simple html template like this:
```html
<div class="mdc-slider">
  <input class="mdc-slider__input" type="range" min="0" max="100" value="50" name="volume" aria-label="Continuous slider demo">
  <div class="mdc-slider__track">
    <div class="mdc-slider__track--inactive"></div>
    <div class="mdc-slider__track--active">
      <div class="mdc-slider__track--active_fill"></div>
    </div>
  </div>
  <div class="mdc-slider__thumb">
    <div class="mdc-slider__thumb-knob"></div>
  </div>
</div>
```

We are wrapping this in two other divs, and then inside a shadow dom.

This might actually be the reasons that their own inbuilt observers (or whatever script they have to calculate the location of the knob) are not doing what they should do. 

Give me some time. I'll try to refactor the component with less nested structure, and see if that has any effect. Then we can continue from there, if things don't work.

src/components/slider/slider.tsx
adrianschmidt: I agree with @jgroth. Having a `setTimeout` with a timeout of 0 can be valid, but using any other value means we are building an intentional race condition*, and can not be considered a solution 😞

*) Unless we are also controlling the timing of the thing we need to wait for. Like, perhaps you trigger a css animation, that you know takes 750 ms, and then you use a `setTimeout` with a timeout of 750, to trigger something when the animation has finished. Then it _might_ be ok. (Although timing in javascript and css is not exact, so even this scenario is quite shaky, and should probably be avoided if possible.)

These issues typically happen when the component is loaded while hidden, like when a component is put inside a limel-dialog or limel-collapsible-section, for example. To fix such problems, the solution I've been using is to have the component that does the hiding and showing emit a resize-event when the content has been revealed.

This is done in both limel-dialog and limel-collapsible-section, so examples can be found there.

---

While I wouldn't want to preclude the possibility, I think that it's unlikely that a fix can be wholly built-in in the slider component itself, because the revealing of the component may happen at any time after the component has been loaded, or it may be tied to some kind of timed transition that means that you need to wait for an unknown amount of time after the component technically became visible on the page.

_If_ there are appropriate observers nowadays, it _might_ be possible. But it won't be possible with timeouts.

adrianschmidt (changes requested):

src/components/slider/slider.tsx
Kiarokh: that didn't help

Issue #2052, opened 2023-01-19
limel-collapsible-section: clicking a disabled action button lets the click through to the collapsible-section and opens or closes it


adrianschmidt:
## Current behavior

Clicking a disabled action button lets the click through to the collapsible-section and opens or closes it.

https://user-images.githubusercontent.com/154725/213476516-75f06877-3289-440f-a068-495664b39081.mov

1. Go to the _Example with actions_ example in [the docs](https://lundalogik.github.io/lime-elements/versions/next/#/component/limel-collapsible-section/).
2. Click the disabled action button (the one with the refresh-icon).

- The collapsible section expands, or collapses. Also, the mouse cursor has the wrong icon, indicating that interaction is possible.

## Expected behavior

The click should be captured and prevented from expanding or collapsing the section. The mouse cursor should be one that does not indicate that interaction is possible.

## Environment

- **lime-elements** version: v36.3.0-next.2
- Framework used: Reproducible in the docs
- Logs:

Kiarokh:
It is possible to treat the symptom of this issue in the collapsible section, in one or another way (like the creative way @LucyChyzhova has suggested in the linked draft-PR). 

But the source of this problem is here: https://github.com/Lundalogik/crm-feature/issues/3317 We need to solve it once for all interactive components like buttons, icon-buttons, shortcut, toggle, etc… Otherwise, we will have this problem happening in other contexts too. We need to get rid of the CSS hack that we have done on all these components.

LucyChyzhova:
@Befkadu1 and I were working on this: https://github.com/Lundalogik/crm-feature/issues/3317 and it seems like a non-related issue.  After implementing a solution described in that thread it was still possible to click on a disabled button and got a collapsible section opened.

LucyChyzhova:
@adrianschmidt Could you please check two linked PRs - they both fix this issue but in a slightly different ways. 
I will unassign myself and put it in the review column. However, would love to get your input ❤️ (whenever you have time) 
Thanks.

adrianschmidt:
> Could you please check two linked PRs - they both fix this issue but in a slightly different ways.

I think we should certainly do #2230 👍

We might want to do #2235 _as well_, but preferably without making it "un-shadowed".

LucyChyzhova:
cool! 🚀 thanks a lot for your input! ❤️

Also for the #2235 we need more fixes, I just added a fixup. However, even now it's quite tricky.

Kiarokh:
[...]

PR #2235, opened 2023-04-14
fix(icon-button): prevent clicks on `disabled` button to reach elemen…

LucyChyzhova:
…ts in the background
fix: https://github.com/Lundalogik/lime-elements/issues/2052#issuecomment-1508064559

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/icon-button/icon-button.tsx
```diff
@@ -9,7 +9,7 @@ import { Component, Element, h, Prop } from '@stencil/core';
  */
 @Component({
     tag: 'limel-icon-button',
-    shadow: true,
+    shadow: false,
```
adrianschmidt, line 12: Why do we have to make it "un-shadowed"? 🤔

Befkadu1: We forgot to get it back!. The PR is still in draft and not ready for review I think

LucyChyzhova: no no @Befkadu1 , it was intentionally, @Kiarokh suggested to play with shadow-root here.  But this hack only works with the new onClick event, which you and I were working on.  I and Kia wanted to have Adrian’s input earlier to not dig deeper if it's not the right way.

PR #2236, opened 2023-04-14
refactor(slider): get rid of nested structure & simplify things

Kiarokh:
This was an effort to fix some rendering bugs, which would happen when the container of the slider was
resized. While trying to find a solution,
we noticed that there are many nested
containers and hoped that these simplifications
could help solve the issue. We hoped to get something perhaps from the MDC's inbuilt tech. But so far
we don't see a noticeable improvement.
However, the code is simpler now.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/slider/partial-styles/_helper-text.scss
+++ b/src/components/slider/partial-styles/_helper-text.scss
@@ -8,6 +8,7 @@
 
 .mdc-slider-helper-line {
     @include shared_input-select-picker.looks-like-helper-line;
+    order: 3;
 }
 .mdc-slider-helper-text {
     @include shared_input-select-picker.looks-like-helper-text;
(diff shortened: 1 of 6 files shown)


LucyChyzhova:
minValue behaves differently now 🤔
The value should not be possible to set less than the minimum value.

https://user-images.githubusercontent.com/50618208/232776994-659056aa-702e-453e-8446-c2ff300d4857.mov

LucyChyzhova (approved):
Approved! 😀

PR #2251, opened 2023-04-21
Use the new `limel-helper-line` component (which is `@private`) in different input fileds

Kiarokh:
related to: https://github.com/Lundalogik/crm-feature/issues/3334
waiting for: https://github.com/Lundalogik/lime-elements/pull/2269

fix: https://github.com/Lundalogik/lime-elements/issues/2257
fix: https://github.com/Lundalogik/lime-elements/issues/2258
fix: https://github.com/Lundalogik/lime-elements/issues/2263

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -454,15 +454,21 @@ export class InputField {
     };
 
     private renderHelperLine = () => {
+        const text: string = this.value || '';
+        const length = text.length;
+
         if (!this.maxlength && !this.hasHelperText()) {
             return;
         }
 
         return (
-            <div tabIndex={-1} class="mdc-text-field-helper-line">
-                {this.renderHelperText()}
-                {this.renderCharacterCounter()}
-            </div>
+            <limel-helper-line
+                helperTextId={helperTextId}
+                helperText={this.helperText}
+                length={length}
+                maxLength={this.maxlength}
+                invalid={this.isInvalid()}
+            />
         );
     };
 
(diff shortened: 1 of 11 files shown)


src/components/helper-line/helper-line.spec.tsx
Kiarokh, line 1: just mucking some ai generated stupid tests for now.
some show errors, but we can fix them later

src/components/input-field/input-field.tsx
```diff
@@ -454,15 +454,21 @@ export class InputField {
...
             return;
         }
 
         return (
-            <div tabIndex={-1} class="mdc-text-field-helper-line">
-                {this.renderHelperText()}
-                {this.renderCharacterCounter()}
-            </div>
+            <limel-helper-line
+                helperTextId={helperTextId}
```
Kiarokh, line 466: I kept this as is, but I don't think the implementation on line 38 is correct actually 👇 
```tsx
const helperTextId = 'tf-helper-text';
```
We need to generate random IDs for each `<limel-input-field />` on the page. Otherwise all input filed components will have identical IDs this way, and the intended usage of the ID will fail 👇 
```tsx
 if (this.hasHelperText()) {
            properties['aria-controls'] = helperTextId;
            properties['aria-describedby'] = helperTextId;
        }
```

Kiarokh: More on this here: https://github.com/Lundalogik/lime-elements/issues/2255

src/components/helper-line/helper-line.tsx
```diff
@@ -0,0 +1,118 @@
...
+    };
+
+    private renderHelperText = () => {
+        if (!this.helperText) {
+            return;
+        }
+
+        return (
+            <span class="helper-text" id={this.helperTextId}>
+                {this.helperText}
```
Kiarokh, line 104: I wonder if we should use `.trim()` here.

```
(method) String.trim(): string
Removes the leading and trailing white space and line terminator characters from a string.
```

Kiarokh: Hmm… they are actually already trimmed away. I don't know why. But this doesn't seem necessary

LucyChyzhova (approved):
🚀 great!

PR #2256, opened 2023-04-24
feat(bulk-create): add connect-to feature

LawrenceBorst:
Fixes Lundalogik/crm-feature#3329

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/list/list-item.types.ts
+++ b/src/components/list/list-item.types.ts
@@ -45,6 +45,11 @@ export interface ListItem<T = any> {
      * Component used to render in the list
      */
     primaryComponent?: ListComponent;
+
+    /**
+     * True if the list item should be removable (only works for limel-chips)
+     */
+    removeable?: boolean;
 }
 
 export interface ListSeparator {
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -321,7 +321,7 @@ export class Picker {
         return {
             id: `${listItem.value}`,
             text: listItem.text,
-            removable: true,
+            removable: listItem.removeable ?? true,
             icon: listItem.icon,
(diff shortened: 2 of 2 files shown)


src/components/list/list-item.types.ts
```diff
@@ -45,6 +45,11 @@ export interface ListItem<T = any> {
      * Component used to render in the list
      */
     primaryComponent?: ListComponent;
+
+    /**
+     * True if the list item should be removable (only works for limel-chips)
+     */
+    removeable?: boolean;
```
jgroth, line 52: I'm not sure this makes much sense at the moment from the lists point of view, so I don't think we should add this here. Also, I'm not sure about the description, how is this connected to the chips? It seems to be used by the picker and the `Chip` interface :thinking:

jgroth (commented):
A comment about the commit message, Lime Elements does not know what "bulk create" is :smile:

src/components/list/list-item.types.ts
LawrenceBorst: Ah good point! I double checked, it seems we could instead pass in the `removable` (I misspelled it in the commit 💀 ) as a prop into the `limel-picker` in `relation-picker.tsx`. The only reason we need it is for `createChip()` in `picker.tsx`, and I just blindly added it to the interface

LawrenceBorst: Looked into it more

We can close this PR, all the changes can be handled by just passing in`readonly=true`

PR #2269, opened 2023-04-26
chore(helper-line): add new `@private` component

Kiarokh:
This is going to replace all helper-lines and their styles from various components.
related to: https://github.com/Lundalogik/crm-feature/issues/3334

⚠️ To see it implemented, you can look in this Draft PR: https://github.com/Lundalogik/lime-elements/pull/2251

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/helper-line/helper-line.tsx
```diff
@@ -0,0 +1,119 @@
...
+        }
+    };
+
+    private renderHelperText = () => {
+        if (!this.helperText) {
+            return;
+        }
+
+        return (
+            <span class="helper-text" id={this.helperTextId}>
```
Kiarokh, line 104: I actually think that the `id` should be on the host, not here. Technically, the parent component is controlling the whole helper line, and is described byt he whole helper line. So it should be on all of it. 

The original MDC way was on the helperText section, so I kept it here.

src/components/helper-line/helper-line.tsx
```diff
@@ -0,0 +1,119 @@
...
+                {this.renderHelperText()}
+                {this.renderCharacterCounter()}
+            </Host>
+        );
+    }
+
+    private hasContent = () => {
+        if (
+            this.maxLength ||
+            this.helperText.length > 0 ||
```
LucyChyzhova, line 90: I think these two checks would be enough
`this.maxLength ||
  this.helperText.length > 0`

src/components/helper-line/helper-line.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+        return (
+            <span class="helper-text" id={this.helperTextId}>
+                {this.helperText}
+            </span>
+        );
+    };
+
+    private renderCharacterCounter = () => {
+        const counter = `${this.length} / ${this.maxLength}`;
```
LucyChyzhova, line 111: Would it make sense (in some cases) also to have only `this.length` and no `maxLength`, so we count characters inside the input field? 🤔

src/components/helper-line/helper-line.tsx
Kiarokh: I also think so. But since we have this check in `limel-input-field`: 

https://github.com/Lundalogik/lime-elements/blob/944258e787c61bf3d7aa2be51d963a24028ec6ca/src/components/input-field/input-field.tsx#L497

I also made the same check here as well. I don't know why the original solution has those. Maybe it was needed for some reason? I don't know. Maybe helperText would output `undefined` if it was undefined?

src/components/helper-line/helper-line.tsx
Kiarokh: It would I guess in some cases. But I think that'd be a new feature. 
By default, we don't want to show a counter and mess up the UI. So the consumer who has a good reason should enable the feature deliberately.

Anyhow, it'll be way easier to add that feature, when we have this component used by all other input fields.

src/components/helper-line/helper-line.tsx
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * @exampleComponent limel-example-helper-line
+ * @exampleComponent limel-example-helper-line-invalid
+ * @exampleComponent limel-example-helper-line-long-text
+ * @exampleComponent limel-example-helper-line-long-text-no-counter
+ * @exampleComponent limel-example-helper-line-character-counter
+ * @private
+ */
+@Component({
+    tag: 'limel-helper-line',
```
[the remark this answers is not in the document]
Kiarokh: p.s.
I don't see this problem in my dev env

src/components/helper-line/helper-line.tsx
LucyChyzhova: > I actually think that the id should be on the host, not here

Should we ask someone or it's okay to merge like this now?

Kiarokh: No, this is a bug already and there is an issue for it. So we fix it later.

LucyChyzhova (approved):
🚀👍 cool!

PR #2274, opened 2023-05-02
Snackbar: Add some good documentations and usage guides

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/snackbar/snackbar.tsx
+++ b/src/components/snackbar/snackbar.tsx
@@ -29,7 +29,7 @@ export class Snackbar {
     public message: string;
 
     /**
-     * The amount of time in milliseconds to show the snackbar.
+     * The amount of time in milliseconds to show the snackbar. Defaults to `5000` (5 seconds).
      */
     @Prop()
     public timeout: number;
--- a/src/components/snackbar/snackbar.tsx
+++ b/src/components/snackbar/snackbar.tsx
@@ -13,6 +13,8 @@ import translate from '../../global/translations';
 
 /**
  * @exampleComponent limel-example-snackbar
+ * @exampleComponent limel-example-snackbar-with-action
+ * @exampleComponent limel-example-snackbar-with-changing-messages
  */
 @Component({
     tag: 'limel-snackbar',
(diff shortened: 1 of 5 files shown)


src/components/snackbar/snackbar.tsx
```diff
@@ -32,7 +32,7 @@ export class Snackbar {
      * The amount of time in milliseconds to show the snackbar. Defaults to `5000` (5 seconds).
      */
     @Prop()
-    public timeout: number;
+    public timeout?: number;
```
Kiarokh, line 35: MDC seems to be doing this in the background. Read commit the message for more info.

Kiarokh, line 56: or maybe it's better that I write it this way:
` public timeout?: number = 5000;`

Kiarokh: it will also help for a little future feature

LucyChyzhova: I think we should have a default delay (like you said 5000) but still give the opportunity to override it.

LucyChyzhova: Should we have it optional only in case the dismiss button is provided? So a snack-bar should be removed after a few seconds or by the dismiss button.

LucyChyzhova: what I think should not happen is that the timeout will override the dismiss button. for instance, if a consultant provides a dismiss button - so it means the message should be read and closed intentionally. H|owever our default 5000 will close this notification no matter the dismiss button.

LucyChyzhova (dismissed):
Great, you have improved the examples! 🚀🤩💫

LucyChyzhova (approved):

Issue #2040, opened 2023-01-16
Select: width of chosen option, defines the width of dropdown list that opens below, causing truncation "sometimes unwanted"


Kiarokh:
## Current behavior
In the select component, width of the chosen option displayed in the trigger element, defines the width of dropdown list that opens below it. This is naturally causing truncation of text in all available options in the list, which is **_sometimes unwanted_**. 

![image](https://user-images.githubusercontent.com/35954987/212644375-f7938f0f-3014-4984-8cb8-80b701b2b6b7.png)

## Expected behavior
This is actually expected behavior most of the times. You don't normally want a dropdown menu that is bigger than its own trigger. Dropdown menus are always as wide as what trigger them. _**But**_ there are exceptions, and they happen usually when the trigger is too small, and that normally happens when the chosen option is too small.

<img width="565" alt="image" src="https://user-images.githubusercontent.com/35954987/212644981-2a15703e-ccbc-44ca-9c01-2516a9de6d22.png">
For example here, we have had to hack this by adding a `in-width: 12.1875rem;` to the element; which also results in ugly UIs, such as this one, where the select element is too wide initially. But without that, we would be getting a dropdown list that doesn't show the content properly, like below:
<img width="581" alt="image" src="https://user-images.githubusercontent.com/35954987/212645666-fbc468c0-361b-46c4-ae92-6f35d14e903d.png">

So we would sometimes (optionally) need dropdown lists that do not have the same width as their own trigger. This should be up to the consumer to choose.

eketorp:
A humerous example where we show "Back off" instead of "Back office"

![image](https://user-images.githubusercontent.com/58422324/212646367-3cb3772f-16cb-40bc-be72-eb7320c16f01.png)

PR #2294, opened 2023-05-10
fix(select): prevent truncation in dropdown options, when the compone…

Kiarokh:
…nt has small width

In the select component, width of the chosen option displayed in the trigger element, used to define the width of dropdown list that opens below it. This naturally caused truncation of text in all of available options in the list, when the dropdown list was opened. If the chosen option was too small in width, other options would be unreadable. This is why the consumers kept setting a `min-width` where they used `limel-select`, which resulted in more layout issues.

fix https://github.com/Lundalogik/lime-elements/issues/2040

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/select/examples/select-narrow.scss
+++ b/src/components/select/examples/select-narrow.scss
@@ -4,6 +4,5 @@ limel-header {
 }
 
 limel-select {
-    min-width: 12rem;
     margin-right: 0.25rem;
 }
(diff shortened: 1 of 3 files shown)


LucyChyzhova (approved):
Yay! We finally fixed it! 🤩🚀❤️

PR #2306, opened 2023-05-16
Create markdown viewer

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/markdown-viewer/markdown-viewer.tsx
```diff
@@ -0,0 +1,56 @@
...
+ * @exampleComponent limel-example-markdown-viewer-images
+ * @exampleComponent limel-example-markdown-viewer-code
+ * @exampleComponent limel-example-markdown-viewer-footnotes
+ * @exampleComponent limel-example-markdown-viewer-tables
+ * @exampleComponent limel-example-markdown-viewer-html
+ * @exampleComponent limel-example-markdown-viewer-blockquotes
+ * @exampleComponent limel-example-markdown-viewer-horizontal-rule
+ */
+@Component({
+    tag: 'limel-markdown-viewer',
```
Kiarokh, line 21:
```suggestion
    tag: 'limel-markdown',
```

Kiarokh: Let's rename all files and tag names.

src/components/markdown/examples/markdown-headings.tsx
```diff
@@ -0,0 +1,31 @@
...
+##### H5
+###### H6
+
+Alternatively, for H1 and H2, an underline-ish style:
+
+Alt-H1
+======
+
+Alt-H2
+------
```
Kiarokh, line 17: these aren't working, so I'll remove them

src/components/markdown/examples/markdown-blockquotes.tsx
Kiarokh, line 1: source code for the examples are taken from here: https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet 

to verify that everything renders fine.

src/components/markdown/examples/markdown-footnotes.tsx
```diff
@@ -0,0 +1,29 @@
...
+A footnote can also have multiple lines[^2].
+
+You can also use words, to fit your writing style more closely[^note].
+
+[^1]: My reference.
+[^2]: Every new line should be prefixed with 2 spaces.
+  This allows you to have a footnote with multiple lines.
+[^note]:
+    Named footnotes will still render with numbers instead of the text but allow easier identification and linking.
+    This footnote also has been made with a different syntax using 4 spaces for new lines.
```
Kiarokh, line 15: according to the source, they should be rendered like this 
image
But not sure how to test and see them. 🤔 maybe we should just remove this example.

src/components/markdown/examples/markdown-html.tsx
Kiarokh, line 1: ain't rendering currently

src/components/markdown/markdown.tsx
```diff
@@ -0,0 +1,55 @@
...
+    shadow: true,
+})
+export class Markdown {
+    /**
+     * The input text. Treated as GitHub Flavored Markdown, with the addition
+     * that any included HTML will be parsed and rendered as HTML, rather than
+     * as text.
+     */
+    @Prop()
+    public content: string;
```
jgroth, line 31: There are both pros and cons for having this as a prop vs having the content provided as child elements instead. How did the discussion go about this topic when it was implemented?

Pros for having it as child elements is that it makes more sense, especially if HTML is allowed as well. Text is flowing content, and forcing the consumer to put it all in a string first is cumbersome. E.g.

```html
<limel-markdown>
    Hello, 
    <span class="green">World!</span>
</limel-markdown>
```
vs
```html
<limel-markdown 
    content="Hello, \n<span class=\"green\">World!</span>">
</limel-markdown>
```

Cons might be that it is a bit harder to implement for us.

If we insist on keeping this as a prop, I think we should make it more compatible with our other components and name this `value` instead of `content`

Kiarokh: > If we insist on keeping this as a prop, I think we should make it more compatible with our other components and name this value instead of content

Agree. Check this out, kind of relevant https://github.com/Lundalogik/crm-feature/issues/3480#issuecomment-1724922053

john-traas: If I recall correctly, we went with passing it as a prop because it allowed us to escape the linter/formatter that would automatically reformat the markdown text, which would then break the markdown. 
The only way to force the linter not to automatically reformat everything was a succession of ugly 'ignore' comments.

Because of that, we felt made more sense to pass it in as a prop.

jgroth: I still think we should rename the `content` prop to `value`

Kiarokh: will u do this @john-traas please? there seems to be some conflicts in the branch as well, that needs resolving.

john-traas: Yes, I am on it @Kiarokh

src/components/markdown/examples/markdown-footnotes.tsx
john-traas: I have got footnotes working @Kiarokh 🥳

Kiarokh: Wonderful! <3

src/components/markdown/markdown.tsx
adrianschmidt: It looks to me like this was resolved in favour of keeping it as a property, rather than a slot, but just in case:

I would assume that anytime someone wants to use this element, they have the content they need to present as a string, most likely from the database?

Also, I don't see why you'd ever need to put html into this component if you already have it as a DOM-structure? If it's for the sake of sanitising the html, that will already be too late by the time this element picks it up, as you've already injected the html into the DOM before the slot can "pick it up" for processing. (Or at least I'd presume it would be too late? And even if it wasn't, we'd presumably only be a small change to the `slot` functionality away from a security problem.)

Anyway, this might feel like I'm trying to hammer home that I think using `slot` is a bad idea, but it's more that I'm wondering if there's something I'm missing? 🤗

package-lock.json
adrianschmidt, line 1: @john-traas I've co-assigned myself to this PR, as I'd like to be involved before we merge it. This is a component we've wanted for a long time. I think it will start to be used pretty much immediately when we add it, and it also has security implications, so getting it right feels pretty important 😊

adrianschmidt: (Just based on the packages added though, this looks promising 😊 We seem to have used the same things we already used for the history note stuff 👍)

john-traas: @adrianschmidt that's great, we definitely need to get it right. 😃

src/components/markdown/markdown.tsx
john-traas: I think we're on the same page, keeping it as a `property` is the approach we're going with.

package-lock.json
adrianschmidt: Sorry, I don't think I'll have the time to look at this one in any depth. It looks like we've basically lifted the implementation from the one we currently use for history in the web client, and if so, I guess this should be fine 👍

Kiarokh:
So does it mean you have approved this now @adrianschmidt ? :)

stencil.config.ts
```diff
@@ -1,11 +1,20 @@
...
 export const config: Config = {
     hashFileNames: false,
     namespace: 'lime-elements',
+    plugins: [sass()],
+    rollupPlugins: {
+        after: [nodePolyfills()],
+    },
+    nodeResolve: {
+        preferBuiltins: true,
+    },
```
jgroth, line 17: Did we still need this? If so, should it not be needed in the other config file as well? :slightly_smiling_face:

john-traas: We don't need it anymore so I removed it.

jgroth (approved):

john-traas (approved):

src/components/markdown/examples/markdown-code.tsx
```diff
@@ -0,0 +1,29 @@
+import { Component, h } from '@stencil/core';
+
+const markdown = `
```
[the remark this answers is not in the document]
adrianschmidt: Doesn't that work with single backticks?

Like this:
```js
const markdown = 'testing some `code` here';
```

Or like this:
```js
const markdown = `
    testing some \`code\` here
`;
```

adrianschmidt:
> So does it mean you have approved this now @adrianschmidt ? :)

I can't approve it since I haven't reviewed it, but I certainly don't _disapprove_! 😄

PR #2310, opened 2023-05-19
feat(picker): fix picker clearing input and search results

john-traas:
This draft PR addresses only the User interaction with the component. If there are items that have been returned in the search, it will not clear the input and reset the search. Behaviour for this issue is demonstrated [here](https://limetechab.sharepoint.com/:v:/r/sites/LimeCRM/Shared%20Documents/Developer%20Open%20Hours/msedge_7ITo2mDjeu.mp4?csf=1&web=1&e=ebyL87).

The search is still triggering a filter query, and subsequent reload of the table view in the background, and that I'm still looking in to.

Connected to https://github.com/Lundalogik/lime-crm-components/issues/1612

There still seems to be an underlying issue in that a filter query, and subsequently a reload of the table view is triggered. 

The steps are:
1) Input a search for the first time, eg "test"
2) Input the same search: "test"
3) Again input the same search: "test"

The first time the search is run, a filter query is triggered, searches run after that are either the same search, or that return results that are already in the suggestions do not trigger the filter query. 
Screenshot 2023-05-23 at 14 54 40

The filter query is initiated from the picker:
Screenshot 2023-05-23 at 14 58 38

The same behaviour occurs when a new search is triggered, for example "x":
Screenshot 2023-05-23 at 14 55 59

It does not get triggered if the results of the query are objects that were part of a previous search, which leads me to believe that this might be intentional behaviour, but I thought it best to check first. 

Perhaps you could comment @jgroth ?


john-traas:
I've put this up for review because, though it does not tackle the underlying issue this will at least allow the customers to input their search queries, without the frustration of having the component resetting. 

This has been an issue for customers for a couple of weeks now, it coincided with our Cabin days, and then other more pressing issues came up, which is why I'd like to implement this temporary fix now, as we will go full into the cycle work for the next weeks. 

I will directly create a new issue for the underlying problem, and add it to the cooldown board, assigning myself to get back to work on it so that we don't leave this temporary fix hanging.

jgroth (approved):

PR #2315, opened 2023-05-23
feat(action-bar): add new component

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/3379

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/action-bar/action-bar-item/action-bar-item.tsx
+++ b/src/components/action-bar/action-bar-item/action-bar-item.tsx
@@ -1,0 +1,130 @@
+import { Component, Event, EventEmitter, h, Prop } from '@stencil/core';
+import { ActionBarItem } from '../action-bar.types';
+import { createRandomString } from '../../../util/random-string';
+import { ListSeparator } from 'src/interface';
+
+/**
+ * @private
+ */
+@Component({
+    tag: 'limel-action-bar-item',
+    shadow: false,
+    styleUrl: 'action-bar-item.scss',
+})
+export class ActionBarButton {
+    /**
+     * Item that is placed in the action bar.
+     */
+    @Prop()
+    public item!: ActionBarItem | ListSeparator;
+
+    /**
+     * Fired when a action bar item has been clicked.
+     */
+    @Event()
+    public select: EventEmitter<ActionBarItem | ListSeparator>;
+
+    /**
+     * When the item is displayed in the available width,
+     * this will be `false`.
+     */
+    @Prop({ reflect: true })
+    public isVisible: boolean = true;
+
+    /**
+     * Used to attach the right tooltip to the right button
+     */
+    private tooltipId: string;
+
+    constructor() {
+        this.tooltipId = createRandomString();
+    }
+
+    public render() {
+        if (!this.isItem(this.item) && this.item.separator) {
+            return <div role="separator" />;
+        }
+
+        return (
+            <button
+                id={this.tooltipId}
+                type="button"
+                onClick={this.handleClick}
+                disabled={this.isDisabled()}
+            >
+                {this.renderIcon()}
+                {this.renderLabel()}
+                {this.renderTooltip()}
+            </button>
+        );
+    }
+
+    private handleClick = (event: MouseEvent) => {
+        event.stopPropagation();
+        this.select.emit(this.item);
+    };
+
+    private isItem(item: ActionBarItem | ListSeparator): item is ActionBarItem {
+        return !('separator' in item);
+    }
+
+    private isDisabled() {
+        if (this.isItem(this.item) && this.item.disabled) {
+            return true;
+        }
+
+        if (!this.isVisible) {
+            return true;
+        }
+    }
+
+    private renderIcon() {
+        if (this.isItem(this.item) && !this.item.icon) {
+            return;
+        }
+
+        if ('icon' in this.item) {
+            return (
+                <limel-icon
+                    name={this.item.icon}
+                    style={{
+                        '--action-bar-item-icon-color': `${this.item.iconColor}`,
+                    }}
+                />
+            );
+        }
+    }
+
+    private renderLabel() {
+        if (!this.isItem(this.item) || this.item.iconOnly) {
+            return;
+        }
+
+        return <span class="text">{this.item.text}</span>;
+    }
+
+    private renderTooltip() {
+        if (!this.isItem(this.item)) {
+            return;
+        }
+
+        if (this.item.text) {
+            return (
+                <limel-tooltip
+                    elementId={this.tooltipId}
+                    label={this.item.text}
+                    helperLabel={this.item.commandText}
+                />
+            );
+        }
+
+        if (this.item.commandText) {
+            return (
+                <limel-tooltip
+                    elementId={this.tooltipId}
+                    label={this.item.commandText}
+                />
+            );
+        }
+    }
+}
--- a/src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
+++ b/src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
@@ -1,0 +1,65 @@
+import { Component, Prop, h, Event, EventEmitter } from '@stencil/core';
+import {
+    MenuItem,
+    LimelMenuCustomEvent,
+    ListSeparator,
+    ActionBarItem,
+    OpenDirection,
+} from '@limetech/lime-elements';
+
+/**
+ * @private
+ */
+@Component({
+    tag: 'limel-action-bar-overflow-menu',
+    shadow: false,
+})
+export class ActionBarOverflowMenu {
+    /**
+     * List of the items that should be rendered in the overflow menu.
+     */
+    @Prop()
+    public items: Array<MenuItem | ListSeparator>;
+
+    /**
+     * Defines the location that the content of the overflow menu
+     * appears, in relation to its trigger.
+     * It defaults to `bottom-end`, since in normal scenarios
+     * (for example when the action bar is not floating at the bottom of the screen)
+     * this menu is the right-most item in the user interface of the component.
+     */
+    @Prop({ reflect: true })
+    public openDirection: OpenDirection = 'bottom-end';
+
+    @Event()
+    public select: EventEmitter<ActionBarItem>;
+
+    public render() {
+        return [
+            <limel-menu
+                openDirection={this.openDirection}
+                items={this.items}
+                onSelect={this.handleSelect}
+            >
+                <button slot="trigger">{this.countOverflowedItems()}</button>
+            </limel-menu>,
+        ];
+    }
+
+    private countOverflowedItems = () => {
+        return `+${this.numberOfMenuItems}`;
+    };
+
+    private handleSelect = (event: LimelMenuCustomEvent<MenuItem>) => {
+        event.stopPropagation();
+        this.select.emit(event.detail);
+    };
+
+    private get numberOfMenuItems() {
+        return this.items.filter((item) => this.isMenuItem(item)).length;
+    }
+
+    private isMenuItem(item: MenuItem | ListSeparator): item is MenuItem {
+        return !('separator' in item);
+    }
+}
--- a/src/components/action-bar/action-bar.scss
+++ b/src/components/action-bar/action-bar.scss
@@ -1,0 +1,57 @@
+/**
+* @prop --action-bar-item-text-color: Text color of action bar items, defaults to `--contrast-1100`.
+* @prop --action-bar-item-icon-color: Color of the icons displayed on each action bar item. Defaults to the text color. To specify a color for an individual item, use the `iconColor` prop instead.
+* @prop --action-bar-item-max-width: Maximum width of a button in the action bar. Defaults to `10rem`. Keep in mind that the buttons should not appear too big.
+* @prop --action-bar-border-radius: Defines the roundness of the corners of the action bar. Defaults to `0`.
+* @prop --action-bar-background-color: Background color of the whole component. Defaults to `--contrast-100`.
+*/
+
+:host(limel-action-bar) {
+    --action-bar-item-height: 2rem;
+
+    --limel-action-bar-item-text-color: var(
+        --action-bar-item-text-color,
+        rgb(var(--contrast-1100))
+    );
+
+    box-sizing: border-box;
+
+    display: inline-flex;
+    align-items: center;
+    padding: 0.125rem 0.25rem;
+    max-width: 100%;
+    border-radius: var(--action-bar-border-radius);
+
+    background-color: var(
+        --action-bar-background-color,
+        rgb(var(--contrast-100))
+    );
+}
+
+:host(limel-action-bar),
+.items {
+    gap: 0.25rem;
+    @media (pointer: coarse) {
+        gap: 0.5rem;
+    }
+}
+
+.items {
+    display: inline-flex;
+    max-width: 100%;
+    min-width: 0;
+}
+
+:host(limel-action-bar.is-full-width) {
+    width: 100%;
+}
+
+:host(limel-action-bar.is-floating) {
+    --action-bar-border-radius: 100vw;
+
+    padding-right: 0.125rem;
+    padding-left: 0.125rem;
+
+    max-width: calc(100% - 2rem);
+    box-shadow: var(--shadow-depth-16);
+}
--- a/src/components/action-bar/action-bar.tsx
+++ b/src/components/action-bar/action-bar.tsx
@@ -1,0 +1,232 @@
+import {
+    Component,
+    Event,
+    EventEmitter,
+    h,
+    Host,
+    Prop,
+    State,
+    Element,
+} from '@stencil/core';
+import { ActionBarItem } from './action-bar.types';
+import { MenuItem, OpenDirection } from '../menu/menu.types';
+import { ListSeparator } from '../list/list-item.types';
+import { isItem } from './isItem';
+
+/**
+ * An action bar is a user interface element commonly found in software applications and websites.
+ * It typically appears at the top of the screen or within a specific section
+ * and serves as a centralized hub for accessing various actions and commands
+ * relevant to the current context or page.
+ *
+ * The action bar often contains a set of clickable icons or buttons (icons + labels)
+ * that represent specific actions, such as saving, deleting, editing, sharing,
+ * or bulk operations for selected items.
+ *
+ * The purpose of an action bar is to provide quick and convenient access to
+ * frequently used functionalities, enabling users to perform common tasks efficiently.
+ * It enhances usability by organizing important actions in a visually prominent and easily accessible location.
+ *
+ * The action bar's design and layout can vary based on the platform or application,
+ * but its primary goal remains consistent—to
+ * empower users to interact with the software and perform desired actions effortlessly.
+ * @exampleComponent limel-example-action-bar
+ * @exampleComponent limel-example-action-bar-overflow-menu
+ * @exampleComponent limel-example-action-bar-colors
+ * @exampleComponent limel-example-action-bar-floating
+ * @exampleComponent limel-example-action-bar-styling
+ * @exampleComponent limel-example-action-bar-as-primary-component
+ * @private
+ */
+@Component({
+    tag: 'limel-action-bar',
+    shadow: true,
+    styleUrl: 'action-bar.scss',
+})
+export class ActionBar {
+    /**
+     * Items that are placed in the action bar.
+     * These represent primary actions.
+     */
+    @Prop()
+    public actions: Array<ActionBarItem | ListSeparator> = [];
+
+    /**
+     * A label used to describe the purpose of the element to users
+     * of assistive technologies, like screen readers.
+     * Example value: "toolbar"
+     */
+    @Prop({ reflect: true })
+    public accessibleLabel?: string;
+
+    /**
+     * - When set to `fullWidth`, the component will take the
+     * entire width of its container.
+     * - When set to `floating`, the component will get basic stylings
+     * to visualize the floating state.
+     * :::note
+     * You should still properly position the component
+     * according to the structure of your user interface.
+     * For example, use an `absolute` or `fixed` position.
+     * :::
+     */
+    @Prop({ reflect: true })
+    public layout?: 'fullWidth' | 'floating';
+
+    /**
+     * Defines the location that the content of the overflow menu
+     * appears, in relation to its trigger.
+     */
+    @Prop({ reflect: true })
+    public openDirection: OpenDirection;
+
+    /**
+     * Fired when a action bar item has been clicked.
+     */
+    @Event()
+    public itemSelected: EventEmitter<ActionBarItem>;
+
+    @Element()
+    private host: HTMLElement;
+
+    @State()
+    private overflowCutoff: number = this.actions.length;
+
+    private intersectionObserver: IntersectionObserver;
+    private firstRender = true;
+    private actionBarItems: HTMLLimelActionBarItemElement[] = [];
+
+    public render() {
+        let overflowActions: Array<MenuItem | ListSeparator> = [];
+        if (this.actions.length) {
+            overflowActions = this.actions.slice(this.overflowCutoff);
+        }
+
+        return (
+            <Host
+                aria-label={this.accessibleLabel}
+                class={{
+                    'is-full-width': this.layout === 'fullWidth',
+                    'is-floating': this.layout === 'floating',
+                }}
+            >
+                <div class="items">
+                    {this.actions.map(this.renderActionBarItem)}
+                </div>
+                {this.renderOverflowMenu(overflowActions)}
+            </Host>
+        );
+    }
+
+    public componentDidRender() {
+        if (this.haveItemsChanged()) {
+            this.intersectionObserver?.disconnect();
+            this.createIntersectionObserver();
+        }
+    }
+
+    public disconnectedCallback() {
+        this.intersectionObserver?.disconnect();
+        this.actionBarItems = [];
+    }
+
+    private renderActionBarItem = (item: ActionBarItem, index: number) => {
+        return (
+            <limel-action-bar-item
+                item={item}
+                onSelect={this.handleSelect}
+                isVisible={this.isVisible(index)}
+            />
+        );
+    };
+
+    private renderOverflowMenu = (items: Array<MenuItem | ListSeparator>) => {
+        if (!(this.actions.length - this.overflowCutoff)) {
+            return;
+        }
+
+        return (
+            <limel-action-bar-overflow-menu
+                openDirection={this.openDirection}
+                items={items}
+                onSelect={this.handleSelect}
+            />
+        );
+    };
+
+    private isVisible(index: number) {
+        return index < this.overflowCutoff;
+    }
+
+    private handleSelect = (
+        event: CustomEvent<ActionBarItem | ListSeparator>
+    ) => {
+        event.stopPropagation();
+        if (isItem(event.detail)) {
+            this.itemSelected.emit(event.detail);
+        }
+    };
+
+    private handleIntersection = (entries: IntersectionObserverEntry[]) => {
+        const intersectingItems = entries.filter(
+            (entry) => entry.isIntersecting
+        );
+
+        const notIntersectingItems = entries.filter(
+            (entry) => !entry.isIntersecting
+        );
+
+        if (this.firstRender) {
+            this.overflowCutoff = intersectingItems.length;
+        } else {
+            this.overflowCutoff =
+                this.overflowCutoff +
+                intersectingItems.length -
+                notIntersectingItems.length;
+        }
+
+        this.firstRender = false;
+    };
+
+    private createIntersectionObserver() {
+        const options = {
+            root: this.host.shadowRoot.querySelector('.items'),
+            rootMargin: '0px',
+            threshold: 1.0,
+        };
+
+        this.actionBarItems = [];
+
+        this.intersectionObserver = new IntersectionObserver(
+            this.handleIntersection,
+            options
+        );
+
+        this.host.shadowRoot
+            .querySelectorAll('limel-action-bar-item')
+            .forEach((actionBarItem) => {
+                this.observe(actionBarItem);
+            });
+    }
+
+    private observe(actionBarItem: HTMLLimelActionBarItemElement) {
+        this.intersectionObserver.observe(actionBarItem);
+        this.actionBarItems.push(actionBarItem);
+    }
+
+    private haveItemsChanged() {
+        const someItemRemoved = this.actionBarItems.some(
+            (actionBarItem: HTMLLimelActionBarItemElement) =>
+                !this.host.shadowRoot.contains(actionBarItem)
+        );
+
+        const someItemAdded = Array.from(
+            this.host.shadowRoot.querySelectorAll('limel-action-bar-item')
+        ).some(
+            (actionBarItem: HTMLLimelActionBarItemElement) =>
+                !this.actionBarItems.includes(actionBarItem)
+        );
+
+        return someItemRemoved || someItemAdded;
+    }
+}
--- a/src/components/action-bar/action-bar.types.ts
+++ b/src/components/action-bar/action-bar.types.ts
@@ -1,0 +1,9 @@
+import { MenuItem } from '../menu/menu.types';
+
+export interface ActionBarItem extends MenuItem {
+    /**
+     * Renders the button in the action bar without their labels.
+     * Does not affect the items that are overflown into the overflow menu.
+     */
+    iconOnly?: boolean;
+}
--- a/src/components/action-bar/examples/action-bar-floating.tsx
+++ b/src/components/action-bar/examples/action-bar-floating.tsx
@@ -1,0 +1,74 @@
+import { Component, h, State } from '@stencil/core';
+import { ActionBarItem } from '../action-bar.types';
+import { ListSeparator } from 'src/interface';
+
+/**
+ * Floating Example
+ *
+ * For some designs, it may make sense to display the action bar as
+ * a floating element on top of the page's content.
+ * Set the `layout` prop to `floating` to get the basics styles of
+ * a floating bar.
+ *
+ * :::note
+ * 1. In this case, the action bar gets some elevation effect
+ * using a `box-shadow`. This is to properly separate the action bar
+ * form its surrounding context. You can override this by setting another
+ * `box-shadow`.
+ * 2. Make sure to use a proper `openDirection` for the
+ * overflow menu.
+ * 3. Make sure there is space on the sides of the action bar,
+ * so that it doesn't stretch out completely from left edge to the right
+ * edge. The component is already doing so using a `max-width`,
+ * but you can override it by providing another `max-width`.
+ * :::
+ */
+@Component({
+    tag: 'limel-example-action-bar-floating',
+    shadow: true,
+    styleUrl: 'action-bar-floating.scss',
+})
+export class ActionBarFloatingExample {
+    @State()
+    private actionBarItems: Array<ActionBarItem | ListSeparator> = [
+        {
+            text: 'Add',
+            icon: 'plus_math',
+            iconOnly: true,
+        },
+        {
+            text: 'Refresh',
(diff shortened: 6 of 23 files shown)


src/components/action-bar/action-bar.types.ts
```diff
@@ -0,0 +1,14 @@
+import { MenuItem } from '../menu/menu.types';
+
+export interface ActionBarItem extends MenuItem {
+    /**
+     * A non-changing value to uniquely identify each item.
+     */
+    id?: string;
```
vicgeralds, line 7: Why is this needed? We already have `value` from `MenuItem` that's more flexible and easier to use for storing `action.command` in our case.

Kiarokh: I'm not sure if the ID is needed. But we have an ID for a similar case in `limel-dock` 🤷‍♂️ 
I thought this was used to identify which action was clicked.
However, the menuItem types don't have any `id`s. We're sending the same ids to the menu too. So they have no effect there. I don't know how all of this is connected.

vicgeralds: > Will it always be unique? For bulk actions for instance we will use the same command ID several times right?

No, command id will not always be unique. Another good reason for removing id here I think. It would be annoying if you think you have to generate a unique id when you don't have one.

What you should do is just provide any useful unique value as `actionBarItem.value`.

vicgeralds: > I thought this was used to identify which action was clicked.

Yes, we did do that with dock, but `DockItem` doesn't have `value` (as it doesn't extend `MenuItem`).

Kiarokh: OK, then I'll remove that from types and all examples

src/components/action-bar/action-bar.scss
```diff
@@ -37,6 +37,10 @@ limel-action-bar-button {
     min-width: 0;
 }
 
+.invisible {
```
Kiarokh, line 40:
```suggestion
.is-invisible {
```

Kiarokh: we typically call such classes this way

src/components/action-bar/action-bar.types.ts
Kiarokh: ⚡ done

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
```diff
@@ -0,0 +1,62 @@
...
+     */
+    @Prop({ reflect: true })
+    public openDirection: OpenDirection;
+
+    @Event()
+    public select: EventEmitter<ActionBarItem>;
+
+    public render() {
+        return [
+            <limel-menu
```
Kiarokh, line 36: We need to add an accessible label for this (for screen readers). For example: "See 4 additional actions". Or at least this needs a prop so that the consumer can add it.

src/components/action-bar/action-bar-item/action-bar-item.tsx
```diff
@@ -0,0 +1,123 @@
...
+        return (
+            <button
+                id={this.tooltipId}
+                type="button"
+                onClick={handleClick}
+                disabled={this.isDisabled()}
+            >
+                {this.renderIcon()}
+                {this.renderLabel()}
+                {this.renderTooltip()}
```
vicgeralds, line 54: Do we really need the tooltip? It's a bit annoying
image

Kiarokh: yes we need. for reasons that I don't want to explain 😄 the tooltip has a [shitty disgusting bug](https://github.com/Lundalogik/lime-elements/issues/1408) that we never have fixed. This is why it becomes annoying. We should fix that bug instead. 😞 I can't, otherwise I'd have done it 2 years ago

src/components/action-bar/action-bar.scss
Kiarokh: I think this class is unnecessary. We should instead add a `visibility: hidden;` style on the item itself. That'll also ensure that the item cannot receive focus, such as when navigating through tab indexes. I'll add a fixup

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
```diff
@@ -0,0 +1,62 @@
...
+    ActionBarItem,
+    OpenDirection,
+} from '@limetech/lime-elements';
+
+/**
+ * @private
+ */
+@Component({
+    tag: 'limel-action-bar-overflow-menu',
+    shadow: false,
```
Befkadu1, line 15: Setting `shadow` to false, is this intentional ?

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
```diff
@@ -0,0 +1,62 @@
...
+})
+export class ActionBarOverflowMenu {
+    /**
+     * List of the items that should be rendered in the overflow menu.
+     */
+    @Prop()
+    items: Array<MenuItem | ListSeparator>;
```
Befkadu1, line 22:
```suggestion
   public items: Array<MenuItem | ListSeparator>;
```

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,262 @@
...
+            this.overflowRecalculationScheduled = true;
+            setTimeout(() => {
+                this.overflowCutoff = this.actionBarItems.length;
+                this.calculatingOverflow = true;
+                this.overflowRecalculationScheduled = false;
+            }, this.RESIZE_RATE_MS);
+        }
+    };
+
+    handleIntersection = (entries: IntersectionObserverEntry[]) => {
```
Befkadu1, line 196:
```suggestion
    private handleIntersection = (entries: IntersectionObserverEntry[]) => {
```

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,262 @@
...
+    shadow: true,
+    styleUrl: 'action-bar.scss',
+})
+export class ActionBar {
+    /**
+     * Items that are placed in the action bar.
+     * These represent primary actions.
+     */
+    @Prop()
+    public actionBarItems: Array<ActionBarItem | ListSeparator> = [];
```
jgroth, line 50: I propose changing this to just `items`. Short and concise names are usually better, and the extra `actionBar` prefix does not add anything in this scenario. There should be no need to repeat the name of the component for this property

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,262 @@
...
+     * of assistive technologies, like screen readers.
+     * Example value: "toolbar"
+     */
+    @Prop({ reflect: true })
+    public accessibleLabel?: string;
+
+    /**
+     * - When set to `fullWidth`, the component will take the
+     * entire width of its container.
+     * - When set to `true`, the component will get basic stylings
```
jgroth, line 63:
```suggestion
     * - When set to `floating`, the component will get basic stylings
```

src/components/action-bar/action-bar.types.ts
```diff
@@ -0,0 +1,9 @@
+import { MenuItem } from '../menu/menu.types';
+
+export interface ActionBarItem extends MenuItem {
+    /**
+     * Renders the button in the action bar without their labels.
+     * Does not affect the items that are overflown into the overflow menu.
+     */
+    iconOnly?: boolean;
```
jgroth, line 8: Should this be set for each individual item or should it be a prop on the component instead?
Also, not sure about the name here. The name of the property refers to the icon, but it controls the visibility of the label. So maybe `displayLabel` would make more sense (with a default value of `true`)? :thinking:

LawrenceBorst:
Fixup fixes that resize issue with the rerenders if you drag/resize for too long without letting go

LawrenceBorst:
Okay final fixup, didn't like the way rendering the items took a little longer to start, couldn't fix without undoing item animation from last fixup, so removed it

src/components/action-bar/action-bar.tsx
jgroth: Another option is to name it `actions`

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
Kiarokh: That's how we did in the Dock. I don't know. But do we have any good reasons why this is a bad idea?

src/components/action-bar/action-bar.types.ts
Kiarokh: Yes, for each individual item. The reason is that we are extending the types from MenuItem `export interface ActionBarItem extends MenuItem {` to be able to send the same array to the overflow menu, and get things rendered nice and tidy there. in MuneItem, the `text` is required. So consumers need to make sure that they provide a text. Which is also good here due to accessibility reasons (we use the provided `text` in the tooltip, which then also is used by ariadescribedby... you know the story).

`icon` is however option. Which is also good for the action bar. 

But the consumer may not want to display the text for all items, only for some of them. Like 👇 
image

Which is why we added `iconOnly`.

When they go to the overflow menu, the `iconOnly` has no effect, and it shouldn't.

I have no strong opinions about the name. It can be another thing, but I feel this one is just fine.

vicgeralds:
I (still) get this error when trying out the latest changes. 😞 

Error: Failed to create resizeObserver, no parent found for limel-action-bar. Consider wrapping the action-bar in an element

src/components/action-bar/action-bar-item/action-bar-item.tsx
vicgeralds: ok! 🙂

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
vicgeralds: I don't see any reason to use shadow dom if we don't have styling. This component is private, so we have nothing hide in shadows.

(I strongly believe shadow dom by default is absurd. Stencil supports scoped css without shadow dom, so if that's the only reason, we don't strictly need it for scoped styles.)

src/components/action-bar/action-bar.tsx
Kiarokh: since there are dividers (`ListSeparator`) in there too, maybe `items` is still a better name?

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
Kiarokh: In here, when the shadow is false, we can easily share the styles between the overflow menu and the buttons without doing anything. 

But if we make shadow true, then we will have to import the action button styles into the .scss file of the overflow menu component. It makes it a bit messier.

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,275 @@
...
+    private lastOverflowCutoff: number = this.actionBarItems.length;
+    private lastResize = Date.now();
+
+    public componentWillLoad() {
+        this.resizeObserver = new ResizeObserver(this.handleResize);
+    }
+
+    disconnectedCallback() {
+        this.resizeObserver.disconnect();
+        this.intersectionObserver.disconnect();
```
vicgeralds, line 112: I got this:

image

You definitely should check that `this.intersectionObserver` is set, since it's conditionally set after render.

vicgeralds (changes requested):

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,275 @@
...
+                }}
+            >
+                {visibleItems.map(this.renderActionBarItem)}
+                {this.renderOverflowMenu(overflowItems)}
+            </Host>
+        );
+    }
+
+    public componentDidRender() {
+        const element = this.element.parentElement;
```
jgroth, line 134: This will not work if the component is an immediate child of a web-component with a shadow DOM. In those cases the parent element can be fetched by
```ts
element.parentNode.host
```

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,275 @@
...
+    @State()
+    private calculatingOverflow: boolean = false;
+
+    private resizeObserver: ResizeObserver;
+    private intersectionObserver: IntersectionObserver;
+    private overflowRecalculationScheduled: boolean;
+    // eslint-disable-next-line no-magic-numbers
+    private RESIZE_RATE_MS = 50;
+    // eslint-disable-next-line no-magic-numbers
+    private RESIZE_TIMEOUT = 250;
```
jgroth, line 102: These can be moved out as `const`s outside the class and then we don't have to silence the linter :slightly_smiling_face:

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,275 @@
...
+    private lastOverflowCutoff: number = this.actionBarItems.length;
+    private lastResize = Date.now();
+
+    public componentWillLoad() {
+        this.resizeObserver = new ResizeObserver(this.handleResize);
```
jgroth, line 107: If the component is disconnected from the DOM and then connected again, the resize observer will not listen for any more events. This should be moved to `connectedCallback` instead

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,275 @@
...
+    public componentDidRender() {
+        const element = this.element.parentElement;
+
+        if (element === null) {
+            throw new Error(
+                'Failed to create resizeObserver, no parent found for limel-action-bar. Consider wrapping the action-bar in an element'
+            );
+        }
+
+        this.resizeObserver.observe(element);
```
jgroth, line 142: Should we really call this after each render? :thinking: Feels like should only have to setup this once

src/components/action-bar/action-bar.tsx
jgroth: Or, since it also relies on the parent element, we can conditionally create it in `componentDidRender` if it has not already been created

jgroth: I'm also curious about why we need both `ResizeObserver` and `IntersectionObserver` :thinking: 
Would it not be enough to create menu items for icons that are not visible? If the container is resized so that a new item would be visible or hidden, would we not get a callback from the `IntersectionObserver` so that we can add/remove items from the menu? :thinking: I'm probably missing someting :smile:

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
Befkadu1: > I don't see any reason to use shadow dom if we don't have styling.

Yes I agree, but I noticed it in many of the new components here in this PR it set to false I was just wondering why we set it to false. The question was not specifically to this line.

@Kiarokh The shadow DOM hides and separates the DOM of a component in order to prevent clashing styles or unwanted side effects. We have building so many components (...Grand parent -> Parent -> child ), we want to separate the DOM of a component. For example we might use same class name in the DOM tree which might have a side effect somewhere

> Stencil supports scoped css without shadow dom

Yes it is "an alternative to using the shadow DOM is using scoped components"

src/components/action-bar/action-bar.tsx
civing: Oh this is really nifty! You are right in your analysis when the problem occur. Let's try this out @LawrenceBorst

Kiarokh: if we do so, we need to update the docs too. just keep in mind

src/components/action-bar/action-bar.tsx
LawrenceBorst: I think that would work, but you might get visible re-renders as you resize. Say you make it smaller, it rerenders to fit the items but also suddenly adds the menu item at the end, which suddenly falls slightly outside of the container. So it requires another render. But now there'll be a visible "glitching" for a split second.

Worst case scenario is if we have this happen near a separator, it might take 3 rerenders.

We also couldn't wait to render until someone has stopped resizing before we "reveal" the result, but this is more of an optional feature (I quite like it though)

But I might be wrong as well there 😅 , it's hard to be sure of anything with this overflow business without explicitly trying

src/components/action-bar/action-bar.tsx
LawrenceBorst: I think you're right! Blindly copied our strategy for the intersection observer here 😅

src/components/action-bar/action-bar.tsx
LawrenceBorst: Or no ignore that first and third paragraph, it seems to me you could possibly mitigate that. But the second paragraph still stands 😅

LawrenceBorst: Wait I'll backpeddle my last comment again, we might need it for when we drag the container to become bigger, because the onResize handler sets the cutoff point to the number of items, which needs to happen for the algorithm to work (the algorithm cuts down on the number of items till things fit)

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,282 @@
...
+
+    @Element()
+    public element: HTMLElement;
+
+    /**
+     * Defines the location that the content of the overflow menu
+     * appears, in relation to its trigger.
+     */
+    @Prop({ reflect: true })
+    public openDirection: OpenDirection;
```
Kiarokh, line 90: we should default this to `bottom-end`

src/components/action-bar/action-bar.tsx
jgroth: I think my point is that we don't need to calculate if anything fits at all :smile: Right now it seems that we render the icons that fits and then those that do not fit inside the menu. With this approach I'm not sure why we need the `IntersectionObserver`.
On the other hand, if we would render _all_ icons _all_ the time, the `IntersectionObserver` will tell us which items that are not visible, and we can then render "copies" of those items inside the menu. If the container is resized, we will again get notified by the `IntersectionObserver` if something would be come visible or hidden, so we can adjust the menu accordingly. In this scenario, there should be no need to use `ResizeObserver`. 

Again, I might be missing something, but it seems that we should use either `ResizeObserver` or `IntersectionObserver`, not both. `IntersectionObserver` is probably the better option, so we do not have to calculate any sizes ourselves.

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
adrianschmidt: > (I strongly believe shadow dom by default is absurd. Stencil supports scoped css without shadow dom, so if that's the only reason, we don't strictly need it for scoped styles.)

(To my knowledge) they didn't support scoped CSS when we started, which is why we used shadow-DOM. Since scoped CSS is now supported, switching to that is perfectly fine with me. I assume there would be a performance boost in using less shadow-DOM. We don't need to rewrite anything either, I think? We could just start using scoped CSS with new components instead, right?

Anyway, one prerequisite for using scoped CSS is that we actually learn know how to use it. You seem to already know, so perhaps you would like to teach the rest of us? I'm thinking that one good way could be to select one of our existing components, based on it having a suitable complexity (not too complex, but not so simple as to make the example meaningless), and change it from using shadow-DOM to using scoped CSS, and then walk us through the changes in a show-and-tell?

src/components/action-bar/action-bar-item/action-bar-item.tsx
```diff
@@ -0,0 +1,123 @@
...
+     */
+    private tooltipId: string;
+
+    constructor() {
+        this.tooltipId = createRandomString();
+    }
+
+    public render() {
+        return this.renderButton(this.handleClick);
+    }
```
adrianschmidt, line 38: It seems pretty meaningless to break out _the whole_ render function into a single separate function…?

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
adrianschmidt: > In here, when the shadow is false, we can easily share the styles between the overflow menu and the buttons without doing anything.
> 
> But if we make shadow true, then we will have to import the action button styles into the .scss file of the overflow menu component. It makes it a bit messier.

Yeah, this isn't really a separate component, it's more of a sub-component that's broken out for the sake of keeping each component simple to grasp, kind of like breaking out a part of a function into a named helper function.

So I think it makes sense to not have this shadowed 👍

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,283 @@
...
```
adrianschmidt, line 90: I dislike that this isn't grouped with the other `@Prop`s…

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,283 @@
...
+    public layout?: 'fullWidth' | 'floating';
+
+    /**
+     * Fired when a action bar item has been clicked.
+     */
+    @Event()
+    public itemSelected: EventEmitter<ActionBarItem>;
+
+    @Element()
+    public element: HTMLElement;
```
adrianschmidt, line 83: We usually call this `host`.

Also, I'm not sure this needs to be public… 🤔

src/components/action-bar/action-bar.tsx
adrianschmidt: Came here to say this ☝️ 😄

src/components/action-bar/action-bar.tsx
adrianschmidt:
```suggestion
        this.intersectionObserver?.disconnect();
```

That ought to do it, right? 🤔

src/components/action-bar/action-bar.tsx
vicgeralds: I also had the idea that you might not need intersection observer in addition to resize observer. If you have the parent width, I guess it's possible to check the offset position relative to the parent somehow. Maybe you can also cache the offset of items, so you don't have to render overflowing items repeatedly?

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,283 @@
...
+        );
+    }
+
+    public componentDidRender() {
+        const element =
+            this.element.parentElement ?? (this.element.parentNode as any).host;
+
+        if (element === null) {
+            throw new Error(
+                'Failed to create resizeObserver, no parent found for limel-action-bar. Consider wrapping the action-bar in an element'
```
adrianschmidt, line 145: This seems to me like a curious error message… When will this happen? When someone uses document.createElement?

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,283 @@
...
+                }}
+            >
+                {visibleItems.map(this.renderActionBarItem)}
+                {this.renderOverflowMenu(overflowItems)}
+            </Host>
+        );
+    }
+
+    public componentDidRender() {
+        const element =
```
adrianschmidt, line 140: `element` is a really generic name for this. It's especially unfortunate as we also have a `this.element`, which points to a different element. Please be kind to thy reader! 😊

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,283 @@
...
+
+        this.resizeObserver.observe(element);
+        if (this.calculatingOverflow) {
+            this.createIntersectionObserver();
+        }
+    }
+
+    @Watch('actionBarItems')
+    protected handleActionBarItemsChange() {
+        setTimeout(this.handleResize, 0);
```
adrianschmidt, line 157: I added this. Without it, the bar wouldn't handle being created without items and then having them added very well. I found it while trying to write tests, but it is a legit use case when it's optional to provide items when the component is created (it's also good to keep it optional, as it not straightforward to supply non-primitive values at the time of creation in tests (if it's even possible… 🤔)

src/components/action-bar/isItem.ts
```diff
@@ -0,0 +1,8 @@
+import { ListSeparator } from '../list/list-item.types';
+import { ActionBarItem } from './action-bar.types';
+
+export function isItem(
+    item: ActionBarItem | ListSeparator
+): item is ActionBarItem {
+    return !('separator' in item);
```
adrianschmidt, line 7: I broke this out, to make it testable. There seem to be at least a few more functions or parts of functions that could be broken out and thus made more easily testable.

src/components/action-bar/action-bar.tsx
jgroth: I think it's due to this issue :smile: https://github.com/Lundalogik/lime-elements/pull/2315#discussion_r1214217958

The consumer is forced to put the element inside a `div` or similar (something without a shadow DOM). But we should fix that since it's a weird restriction :slightly_smiling_face:

src/components/action-bar/action-bar.tsx
civing: I already gave that comment but I think it was forgotten 🥲

src/components/action-bar/action-bar.tsx
civing: After talking to Johan we might be able solve this only using the intersection observer. For that to work we need to rearrange how we render the menu a bit. Now we always place it in the end of all the `action-bar-items`. But if we are going to render the ones that are invisible as well we need to do it slightly different I think.

Now both the menu and the action bar items are rendered at the same level in the DOM and the host is the element we are intersecting on.

```html
<Host>
    <action-bar-item />
    <action-bar-item />
    <action-bar-item />
    <action-bar-menu />
</Host>
```

Could we do it something like this instead?

```html
<Host>
    <div class='div-to-observe'>
        <action-bar-item />
        <action-bar-item />
        <action-bar-item />
    </div>
    <action-bar-menu />
</Host>

If we don't need to render the action bar menu that space will be claimed by the div that we observe intersections on. I don't know, just shower thoughts. But if we don't do it somehting in line with this it might get tricky to calculate were to put the menu when it's needed.

File: src/components/action-bar/action-bar.tsx (hunk 20 above)
Comment (civing) [reply src/components/action-bar/action-bar.tsx]: What @jgroth said 🙂 

At first we didn't know how to get some element to pass on to the resize observer if the parent element was a shadow dom. But it's solved now.

**Comment by LawrenceBorst** [comment]
Oh I apparently pushed the big fixup 2 days ago. 😅 Shoutout to Johan for seeing the approach.

In any case, this last fixup completely changes the code, as we wrote it from scratch. There are now probably chunks of CSS that are leftover from the previous iteration, so we need to look into what can be removed there.

File: src/components/action-bar/action-bar.scss [hunk 24]
```diff
@@ -0,0 +1,66 @@
...
+}
+
+limel-action-bar-button {
+    min-width: 0;
+    display: absolute;
+}
+
+.visible-items {
+    display: inline-flex;
+    overflow-x: hidden;
```
Kiarokh, line 43: image
This will cause a rendering problem for the shadows (hovered buttons).
We should conditionally set overflow while resizing the component. When the resize and calculations are done, remove the overflow.

src/components/action-bar/action-bar-item/action-bar-item.tsx
```diff
@@ -0,0 +1,134 @@
...
+     */
+    @Event()
+    public select: EventEmitter<ActionBarItem | ListSeparator>;
+
+    /**
+     * When the item is displayed in the available width,
+     * this will be `false`.
+     */
+    @Prop({ reflect: true })
+    public isVisible: boolean = true;
```
Kiarokh, line 32: this strategy gives us many things, but it also creates empty wholes in the UI like this:
image
@LawrenceBorst and I are trying to see how we can mitigate this.

Kiarokh: I think the current behavior 👇 

https://github.com/Lundalogik/lime-elements/assets/35954987/ff684a7e-f5e0-4738-8757-e96cb53ccc25

is still better than the alternative _jumping_ solution that we can do now 👇 

https://github.com/Lundalogik/lime-elements/assets/35954987/79257559-db76-43a7-b9e7-4523334c8a2e

I'm still not sure how we can achieve a nice and smooth transition, when an item disappears. We don't want tit to occupy visible space anymore. I can think of two ways to do it with CSS, one would be setting `display: none` on it, and the other is using `position: absolute` on the `<button />`. But both of these will mess up the intersection observer. Which means the observer never puts back the "removed" item, because it can't see it anymore.

I think we can skip fixing this issue for now. Maybe some beautiful day we can make it work using more advanced CSS and JS.

LawrenceBorst:
Force pushed to change the commit message of the original commit. No changes.

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,204 @@
...
+                    {this.actions.map(this.renderActionBarItem)}
+                </div>
+                {this.renderOverflowMenu(overflowActions)}
+            </Host>
+        );
+    }
+
+    public componentDidLoad() {
+        this.createIntersectionObserver();
+    }
```
vicgeralds, line 122: This doesn't create a new intersection observer after disconnect (the same problem that mentioned in https://github.com/Lundalogik/lime-elements/pull/2315#discussion_r1214224949), but what's more important, it doesn't account for any changed actions bar items after first render.

vicgeralds (changes requested):

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,204 @@
...
+                intersectingItems.length -
+                notIntersectingItems.length;
+        }
+
+        this.firstRender = false;
+    };
+
+    private createIntersectionObserver() {
+        const options = {
+            root: this.host.parentElement,
```
vicgeralds, line 188: Didn't you want to handle being rendered inside a shadow dom?

```suggestion
            root: this.host.parentNode instanceof ShadowRoot
                ? this.host.parentNode.host
                : this.host.parentElement,
```

src/components/action-bar/action-bar.tsx
vicgeralds: A suggestion is to do something in `componentDidRender()` instead, and only create a new intersection observer (or maybe just update which elements it observes, if that's possible) if the inputs changed.

So you could compare the result of `this.host.shadowRoot.querySelectorAll('limel-action-bar-item')` to see if they are the same as the observed elements or not.

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
Kiarokh: > prerequisite for using scoped CSS is that we actually learn know how to use it. You seem to already know, so perhaps you would like to teach the rest of us?

I can't say I know that much. But having a session about it would be nice. Then I'll have to go and learn too 😊

src/components/action-bar/action-bar-item/action-bar-item.tsx
Kiarokh: I agree

Kiarokh: ⚡

src/components/action-bar/action-bar.scss
Kiarokh: fixed now

Kiarokh:
Known & non-blocking issue, which is probably hard to fix or not worth to fix: https://github.com/Lundalogik/crm-feature/issues/3443

Putting it there to keep things more nice and tidy.

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,237 @@
...
+        this.host.shadowRoot
+            .querySelectorAll('limel-action-bar-item')
+            .forEach((actionBarItem: HTMLLimelActionBarItemElement) => {
+                if (!this.actionBarItems.includes(actionBarItem)) {
+                    return true;
+                }
+            });
+
+        return false;
+    }
```
jgroth, line 236: This method will always return `false`. The inner returns from the `forEach` calls will never return from the `haveItemsChanged` method. We can replace the `forEach` calls with `find` or `some` which will return a value which we can check before returning from `haveItemsChanged`

src/components/action-bar/action-bar.tsx
```diff
@@ -0,0 +1,237 @@
...
+        this.host.shadowRoot
+            .querySelectorAll('limel-action-bar-item')
+            .forEach((actionBarItem) => {
+                this.observe(actionBarItem);
+            });
+    }
+
+    private observe(actionBarItem: HTMLLimelActionBarItemElement) {
+        this.intersectionObserver.observe(actionBarItem);
+        this.actionBarItems.push(actionBarItem);
```
jgroth, line 215: We should probably reset this list every time we create a new intersection observer, otherwise we will just keep adding items to it so it might both contain duplicates and items that should no longer be included in it.

src/components/action-bar/examples/action-bar-floating.tsx
```diff
@@ -0,0 +1,74 @@
+import { Component, h, State } from '@stencil/core';
+import { ActionBarItem } from '../action-bar.types';
+import { ListSeparator } from 'src/interface';
+
+/**
+ * Floating Example
+ *
+ * For some designs, it may make sense to display the action bar as
+ * a floating element on top of the page's content.
+ * Set the `isFloating` prop to `true` to get the basics styles of
```
jgroth, line 10:
```suggestion
 * Set the `layout` prop to `floating` to get the basics styles of
```

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
Kiarokh: I actually change my mind about this. The button is already like this: `<button>+11</button>`. It should be accessible enough I guess.
image

src/components/action-bar/action-bar.tsx
vicgeralds: done in 5f115dc7e947662a23e081198e90f730534593fa

src/components/action-bar/action-bar.tsx
vicgeralds: fixed in 5f115dc7e947662a23e081198e90f730534593fa

vicgeralds:
> Known & non-blocking issue, which is probably hard to fix or not worth to fix: [Lundalogik/crm-feature#3443](https://github.com/Lundalogik/crm-feature/issues/3443)
> 
> Putting it there to keep things more nice and tidy.

I think this might have been fixed in the latest fixup.

vicgeralds (approved):
The parts that I've reviewed looks good to me

Kiarokh:
> > Known & non-blocking issue, which is probably hard to fix or not worth to fix: [Lundalogik/crm-feature#3443](https://github.com/Lundalogik/crm-feature/issues/3443)
> > Putting it there to keep things more nice and tidy.
> 
> I think this might have been fixed in the latest fixup.

Nope

vicgeralds:
> > > Known & non-blocking issue, which is probably hard to fix or not worth to fix: [Lundalogik/crm-feature#3443](https://github.com/Lundalogik/crm-feature/issues/3443)
> > > Putting it there to keep things more nice and tidy.
> > 
> > 
> > I think this might have been fixed in the latest fixup.
> 
> Nope

My mistake 😊

vicgeralds (dismissed):
Looks good to me, and works well as far as I can see. 🙂

src/components/action-bar/action-bar-item/action-bar-item.tsx
vicgeralds: While I agree with Lucy to some extent, I don't think this is about handling resizing as much as it's about handling re-rendering and having something that's not too complex and not shaky.

I think we are happy with the empty hole for now, so that this can be merged? And hopefully we'll find a way to improve it.

civing: I would say let's be fine with the empty hole. 🕳️ 
If the apatite grows to fill it we could do it in a cooldown.

src/components/action-bar/action-bar.types.ts
vicgeralds: https://github.com/Lundalogik/crm-feature/issues/3452

vicgeralds (approved):

PR #2345, opened 2023-06-08
Tests for action bar

civing:
This PR adds unit tests for the `limel-action-bar`. It's mostly concerned about how we handle overflow of the `limel-action-bar-items` into the `limel-action-bar-overflow-menu`.

The reason for creating a follow up PR for this was due to the large size of PR containing the action bar.

Fixes Lundalogik/crm-feature#3379
Connected to Lundalogik/lime-elements#2315
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/action-bar/action-bar.spec.tsx
+++ b/src/components/action-bar/action-bar.spec.tsx
@@ -1,0 +1,188 @@
+import { h } from '@stencil/core';
+import { newSpecPage, SpecPage } from '@stencil/core/testing';
+import { ActionBarItem, ListSeparator } from '@limetech/lime-elements';
+import { ActionBar } from './action-bar';
+
+let page: SpecPage;
+let intersectionObserverMockedFunctions;
+let intersectionObserverMock;
+let intersectionCallback;
+const actionsIn: Array<ActionBarItem | ListSeparator> = [
+    {
+        text: 'Add',
+        icon: 'info',
+    },
+    {
+        text: 'Edit',
+        icon: 'idea',
+    },
+    {
+        text: 'Delete',
+        icon: 'multiply',
+    },
+    {
+        separator: true,
+    },
+    {
+        text: 'Settings',
+        icon: 'external_link',
+    },
+];
+describe('action-bar', () => {
+    beforeEach(async () => {
+        intersectionObserverMockedFunctions = {
+            observe: jest.fn(),
+            unobserve: jest.fn(),
+            disconnect: jest.fn(),
+        };
+        intersectionObserverMock = jest.fn((callback) => {
+            intersectionCallback = callback;
+
+            return intersectionObserverMockedFunctions;
+        }) as any;
+        global.IntersectionObserver = intersectionObserverMock;
+
+        page = await newSpecPage({
+            components: [ActionBar],
+            template: () => <limel-action-bar actions={actionsIn} />,
+        });
+
+        await page.waitForChanges();
+    });
+
+    it('renders', () => {
+        expect(page.root).toEqualHtml(`
+            <limel-action-bar>
+                <mock:shadow-root>
+                    <div class=items>
+                        <limel-action-bar-item isvisible=""></limel-action-bar-item>
+                        <limel-action-bar-item isvisible=""></limel-action-bar-item>
+                        <limel-action-bar-item isvisible=""></limel-action-bar-item>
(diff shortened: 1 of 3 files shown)


src/components/action-bar/action-bar.spec.tsx
```diff
@@ -0,0 +1,184 @@
...
+
+    it('handles all actions intersecting', async () => {
+        triggerIntersection([true, true, true, true, true]);
+
+        await page.waitForChanges();
+        expect(getNumVisibleActions()).toBe(5);
+    });
+
+    /**
+     * How can I get the disconnectedCallback lifecycle method to be called? //Pelle
```
jgroth, line 149: `disconnectedCallback` will only be invoked when the component is removed from the DOM. You can try something like this

```ts
page.body.removeChild(page.root)
```

src/components/action-bar/action-bar.spec.tsx
```diff
@@ -0,0 +1,184 @@
...
+        ).toHaveBeenCalledTimes(1);
+    });
+});
+
+const triggerIntersection = (intersectingElements: boolean[]) => {
+    const entries = intersectingElements.map((entry) => ({
+        isIntersecting: entry,
+    }));
+
+    // Set only two of the items visible as a start
```
jgroth, line 163: Does this comment belong here? :smile:

src/components/action-bar/action-bar.spec.tsx
```diff
@@ -0,0 +1,184 @@
...
+            intersectionObserverMockedFunctions.observe
+        ).toHaveBeenCalledTimes(5);
+    });
+
+    it('hides non-visable items in the menu', async () => {
+        triggerIntersection([true, true, false, false, false]);
+
+        await page.waitForChanges();
+        expect(getNumVisibleActions()).toBe(2);
+        expect(hasOverflowMenu()).toBeTruthy();
```
jgroth, line 80: This can be removed since it's part of the next test which has identical setup (or the other test can be removed :smile: )

src/components/action-bar/action-bar.spec.tsx
civing: Yeah, that was some lingering comment that I don't have a good explanation as to why it ended up there 😅 
I will remove it

src/components/action-bar/action-bar.spec.tsx
civing: Yes, you are right.

vicgeralds (commented):
I approve the reconnected fix

src/components/action-bar/action-bar.tsx
```diff
@@ -127,7 +129,9 @@ export class ActionBar {
 
     public disconnectedCallback() {
         this.intersectionObserver?.disconnect();
+        this.intersectionObserver = undefined;
         this.actionBarItems = [];
+        this.connectedCallback = () => this.createIntersectionObserver();
```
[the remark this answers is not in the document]
jgroth: Perhaps we should move this line into `connectedCallback` instead? :thinking: 
That way we will at least reduce the risk of errors in the future if someone would add code into it without realizing that it is overwritten

jgroth (approved):
A+ on the tests, easy to follow :slightly_smiling_face: 
I will approve this and you can decide if you want to move that line I commented on or not

src/components/action-bar/action-bar.tsx
```diff
@@ -195,6 +199,9 @@ export class ActionBar {
             threshold: 1.0,
         };
 
+        this.overflowCutoff = this.actions.length;
```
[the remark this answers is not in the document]
civing: Agree, but now we didn't follow Johan's suggestion 😊

src/components/action-bar/action-bar.tsx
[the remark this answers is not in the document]
jgroth: They fail? That's a bit surprising :thinking: 
To me it looks like the behavior would be the same. Maybe I was unclear in what I meant :smile:

PR #2348, opened 2023-06-12
feat(icon): use browser request cache to cache icons

Kiarokh:
The request cache is more persistent than the in-memory implementation we have today. Icons should now only be cached once until the cache is cleared, instead of once per page load

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/global/icon-cache.ts
```diff
@@ -1,53 +1,51 @@
...
-        return new Promise((resolve) => {
-            if (!this.resolveFunctions[name]) {
-                this.resolveFunctions[name] = [];
-                this.fetchData(name, path);
-            }
-
-            this.resolveFunctions[name].push(resolve);
-        });
+    private async fetchData(url: string, cache: Cache): Promise<Response> {
+        let requestPromise = this.promises[url];
```
LawrenceBorst, line 34: I'm probably missing something, but I'm not sure why we need this.promises.

Couldn't we do away with the condition here as well as `this.promises` entirely?

LawrenceBorst (approved):
Tested on empty cache, full cache, mixed cache. 🎉 

Looks good, works fine but just one small comment

src/global/icon-cache.ts
LawrenceBorst: I see, that makes sense.

Let's ship it then 🚢

PR #2353, opened 2023-06-13
feat(action bar): fix list item type definition

john-traas:
Currently the interface definition for the ListSeparator is explicitly setting 'true' when it should instead be 'boolean'

Fixes [Fix list-item.types in lime-elements](https://github.com/Lundalogik/crm-feature/issues/3454)

```
    separator: true;
}
```

->

```
    separator: boolean;
}
```

### DIFF
--- a/src/components/list/list-item.types.ts
+++ b/src/components/list/list-item.types.ts
@@ -48,7 +48,7 @@ export interface ListItem<T = any> {
 }
 
 export interface ListSeparator {
-    separator: true;
+    separator: boolean;
 }
 
 export interface ListComponent {


src/components/list/list-item.types.ts
```diff
@@ -48,7 +48,7 @@ export interface ListItem<T = any> {
 }
 
 export interface ListSeparator {
-    separator: true;
+    separator: boolean;
```
adrianschmidt, line 51: Typing `separator` to `true` is intentional, as the property should either exist with a value of `true`, or not exist.

If you got some kind of error on this, I can help out trying to find some kind of fix 👍

adrianschmidt (changes requested):

PR #2362, opened 2023-06-22
feat(snackbar): add CSS variables for positioning the snackbar

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/3360

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/snackbar/examples/snackbar-positioning.scss
+++ b/src/components/snackbar/examples/snackbar-positioning.scss
@@ -1,0 +1,5 @@
+limel-snackbar {
+    --snackbar-top: 0;
+    --snackbar-left: auto;
+    --snackbar-bottom: auto;
+}
--- a/src/components/snackbar/examples/snackbar-positioning.tsx
+++ b/src/components/snackbar/examples/snackbar-positioning.tsx
@@ -1,0 +1,56 @@
+import { Component, Element, h } from '@stencil/core';
+
+/**
+ * Positioning on large and small screens
+ *
+ * Snackbars are by default center-aligned and placed at the bottom of the screen.
+ * However, on larger screens, they can optionally be displayed on the leading edge
+ * which would be the left side in LTR, or the right side in RTL.
+ *
+ * To do so, you can take advantage of the provided CSS variables,
+ * and keep in mind that the Snackbar uses `position: fixed;`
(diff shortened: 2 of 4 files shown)


LawrenceBorst (dismissed):
Tested. Looks good

src/components/snackbar/examples/snackbar-positioning.tsx
```diff
@@ -0,0 +1,56 @@
...
+    shadow: true,
+    styleUrl: 'snackbar-positioning.scss',
+})
+export class SnackbarPositioningExample {
+    @Element()
+    private host: HTMLLimelExampleSnackbarElement;
+
+    private triggerSnackbarWithoutAction: (event: MouseEvent) => void;
+
+    constructor() {
```
LawrenceBorst, line 34: I wonder if we should use `connectedCallback` instead, knowing that the appers tend to copy code from the documentation and it might be more "correct". I was surprised this works 😅, as `constructor` could technically be called before we attach the snackbar to the DOM.

src/components/snackbar/examples/snackbar-positioning.scss
```diff
@@ -0,0 +1,5 @@
+limel-snackbar {
```
LawrenceBorst, line 1: Should we add the default `0` for `--snackbar-right` here, to be more clear?

src/components/snackbar/examples/snackbar-positioning.tsx
LawrenceBorst: I take it back, I looked through the source and it seems this is the way :)

LawrenceBorst (approved):

src/components/snackbar/examples/snackbar-positioning.scss
Kiarokh: it's already inherited from the main component. My expectation is that the consumer knows CSS and is aware of what they are doing. So they only override what they have to override.

PR #2372, opened 2023-06-27
fix(chip-set): rename event from `input` to `chipSetInput`

adrianschmidt:
Stencil has long complained about custom events that shadow native events, that is, have the same name as a native event. When upgrading to Stencil v3, this causes build errors if the event handler uses typed arguments, as the types of the custom events conflict with the types of the native events.

BREAKING CHANGE: The custom event `input` has been renamed to `chipSetInput`.

This change is necessary in order for both us and our consumers to upgrade to StencilJS v3 and beyond, as StencilJS v3 uses strongly typed arguments for the `input` event, and thus causes a build error if the arguments to the event handler used with our custom event uses conflicting types.

Example of code that needs to be changed:

```ts
public render() {
    return (
        <limel-chip-set
            ...
            onInput={this.handleInput}
        />
    );
}
```

Updated code:

```ts
public render() {
    return (
        <limel-chip-set
            ...
            onChipSetInput={this.handleInput}
        />
    );
}
```

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -177,7 +177,7 @@ export class ChipSet {
      * Dispatched when the input is changed for type `input`
      */
     @Event()
-    private input: EventEmitter<string>;
+    private chipSetInput: EventEmitter<string>;
 
     @Element()
     private host: HTMLLimelChipSetElement;
@@ -537,7 +537,7 @@ export class ChipSet {
         event.stopPropagation();
         this.inputChipIndexSelected = null;
         this.textValue = event.target.value;
-        this.input.emit(event.target.value && event.target.value.trim());
+        this.chipSetInput.emit(event.target.value && event.target.value.trim());
     }
 
     private handleInteractionEvent(event: MDCChipInteractionEvent) {
--- a/src/components/chip-set/examples/chip-set-input-type-search.tsx
+++ b/src/components/chip-set/examples/chip-set-input-type-search.tsx
@@ -43,7 +43,7 @@ export class ChipSetInputExample {
                 maxItems={this.maxItems}
                 value={this.value}
                 onChange={this.handleChange}
-                onInput={this.handleInput}
+                onChipSetInput={this.handleInput}
                 onKeyUp={this.onKeyUp}
                 emptyInputOnBlur={this.emptyInputOnBlur}
             />,
--- a/src/components/chip-set/examples/chip-set-input-type-text.tsx
+++ b/src/components/chip-set/examples/chip-set-input-type-text.tsx
@@ -63,7 +63,7 @@ export class ChipSetInputExample {
                 value={this.value}
                 maxItems={this.maxItems}
                 onChange={this.handleChange}
-                onInput={this.handleInput}
+                onChipSetInput={this.handleInput}
                 onKeyUp={this.onKeyUp}
                 emptyInputOnBlur={this.emptyInputOnBlur}
             />,
--- a/src/components/chip-set/examples/chip-set-input.tsx
+++ b/src/components/chip-set/examples/chip-set-input.tsx
@@ -63,7 +63,7 @@ export class ChipSetInputExample {
                 leadingIcon={this.hasLeadingIcon ? 'search' : null}
                 maxItems={this.maxItems}
                 onChange={this.handleChange}
-                onInput={this.handleInput}
+                onChipSetInput={this.handleInput}
                 onInteract={this.handleInteraction}
                 onKeyUp={this.onKeyUp}
                 emptyInputOnBlur={this.emptyInputOnBlur}
--- a/src/components/icon/examples/icon-finder.tsx
+++ b/src/components/icon/examples/icon-finder.tsx
@@ -47,7 +47,7 @@ export class IconFinder {
                 type="input"
                 value={this.value}
                 onChange={this.chipSetOnChange}
-                onInput={this.onInput}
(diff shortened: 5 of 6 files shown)


src/components/chip-set/chip-set.tsx
```diff
@@ -177,7 +177,7 @@ export class ChipSet {
      * Dispatched when the input is changed for type `input`
      */
     @Event()
-    private input: EventEmitter<string>;
+    private chipSetInput: EventEmitter<string>;
```
adrianschmidt, line 180: Two questions:

1. Should we name it `chipSetInput` or `limelChipSetInput`?
2. This is the only one that we have to update because it causes build errors with Stencil v3, but should we rename other events too while we're at it?
  - Argument for: We would like to change the names, at least for all events that are shadowing native events, and perhaps for all events, to avoid future conflicts. It's always a horrible struggle to get people to update their customisations, let's do all the changes we know we want at once.
  - Argument against: It's always a horrible struggle to get people to update their customisations, so let's not make this any more difficult than it has to be. If we force too many changes at once, people are probably just going to protest and refuse instead.

adrianschmidt, line 180: Two questions:

1. Should we name it `chipSetInput` or `limelChipSetInput`?
2. This is the only one that we have to update because it causes build errors with Stencil v3, but should we rename other events too while we're at it?
  - Argument for: We would like to change the names, at least for all events that are shadowing native events, and perhaps for all events, to avoid future conflicts. It's always a horrible struggle to get people to update their customisations, let's do all the changes we know we want at once.
  - Argument against: It's always a horrible struggle to get people to update their customisations, so let's not make this any more difficult than it has to be. If we force too many changes at once, people are probably just going to protest and refuse instead.

Ping @jgroth @civing @eketorp

civing: 1. What would be the reason to prefix with `limel`? I like `chipSetInput`
2. What more events do we have? `change` and `select`? It would be really good to get rid of all the events that shadows native event in my opinion. It's a lot of components that would be affected by a breaking change like that though.

- `limel-slider`
- `limel-select`
- `limel-progress-flow`
- `limel-picker`
- `limel-menu`
- `limel-list`
- `limel-input-field`
- `limel-form`
- `limel-file`
- `limel-color-picker`
- `limel-code-editor`
- `limel-date-picker`
- `limel-chip-set`
- `limel-checkbox`
- `limel-button-group`

What would be the upgrade procedure for our users? Build brakes and they update accordingly? Could we do something with doctor health?

jgroth: If we are doing a change like this, I think we should first introduce the new event and deprecate the old one and have the components emit both for a while letting consumers have a period where they can change to the new one. Just forcing everyone to change from one day to another is not going to work. Then at some point in the future we remove all events that we have deprecated

adrianschmidt: > 1. What would be the reason to prefix with `limel`? I like `chipSetInput`
> 2. What more events do we have? `change` and `select`? It would be really good to get rid of all the events that shadows native event in my opinion. 

1. I'm just thinking that if we should rename the events, we might want to just make sure we don't conflict with events from other libraries either, while we're at it.
2. Apart from `input`, `change` and `select` seem to be the only ones. We also have a few components using a property called `prefix`, which is a reserved public name, so we should change that too.

> It's a lot of components that would be affected by a breaking change like that though.
> 
> * `limel-slider`
> * `limel-select`
> * `limel-progress-flow`
> * `limel-picker`
> * `limel-menu`
> * `limel-list`
> * `limel-input-field`
> * `limel-form`
> * `limel-file`
> * `limel-color-picker`
> * `limel-code-editor`
> * `limel-date-picker`
> * `limel-chip-set`
> * `limel-checkbox`
> * `limel-button-group`
> 
> What would be the upgrade procedure for our users? Build brakes and they update accordingly? Could we do something with doctor health?

With the hard switch I proposed here, the upgrade procedure would likely be terrible, because builds may very well _not_ break, since there are native events with the same name. If it's not caught at build-time, problems would manifest at run-time instead. Worst case scenario is that it doesn't even cause a run-time error, but just fail silently.

But:

> If we are doing a change like this, I think we should first introduce the new event and deprecate the old one and have the components emit both for a while letting consumers have a period where they can change to the new one.

Oh, yeah, that's very true!

I didn't think of that because I got stuck on the thought that there's no way we can print a warning for if the old event is used, since we can only emit it, we can't know if someone uses it (although marking it as deprecated in a docblock ought to make IntelliSense tell people, right?). From there, I guess I jumped to the conclusion that a hard switch was the only option, but of course it's not.

I'll get right on to doing that for all events shadowing native events right away. There's no reason to postpone that part of the work. Thanks for that input @jgroth! 🙏

I'm not quite sure right know what we could do with `doctor health`, but we could at the very least use it to output a warning if we see any uses of `onInput`, `onChange`, `onSelect` or `prefix` on any of the affected components. It's probably messier to try to detect if people are registering event listeners using `addEventListener`, but I wouldn't expect that way to be very common… 🤔

adrianschmidt:
This is no longer needed.

PR #2374, opened 2023-06-28
Make buttons clickable on enter

LawrenceBorst:
Make buttons clickable when focused and pressing enter.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/style/mixins.scss
Kiarokh, line 1: Reminder: we should not commit this file in this PR. The styles would be a separate PR.

src/components/button/button.tsx
```diff
@@ -77,6 +93,7 @@ export class Button {
                     'just-loaded': this.justLoaded && !this.loadingFailed,
                     'just-failed': this.justLoaded && this.loadingFailed,
                     'mdc-button--outlined': this.outlined,
+                    active: this.isActive,
```
Kiarokh, line 96: reminder: we don't need this class in this PR. We would do it in a separate PR, when working on the styles

src/components/button/button.tsx
```diff
@@ -102,6 +119,39 @@ export class Button {
         }
     }
 
+    @Listen('keydown')
+    handleKeyDown(ev: KeyboardEvent) {
+        if (ev.key === 'Enter' && !this.isActive) {
+            this.isActive = true;
+
+            this.host.focus();
```
LawrenceBorst, line 127: When I last worked on this issue it seemed like it took away focus from the host, but now when I test it in lime-elements on main it doesn't 🤔
Maybe we don't need this line after all

src/components/button/button.tsx
Kiarokh: also when we add it, the class should be called `is-active`, not `active` ;)

src/components/button/button.tsx
LawrenceBorst: Update: yeah, you don't need it :)

src/components/button/button.tsx
LawrenceBorst: Is it alright if I split the PR into two commits, one for the application logic and one for the styles?

We can create a larger PR where we make this approach more scalable after this gets merged :)

Kiarokh: No, I need to do some changes for the styles and mixins, or at least be able to review it. Since I'm on vacation, it's better to get this one merged, and we can do next one as soon as I come back. ;)

LawrenceBorst:
Tested in Chrome and Edge, but in Safari I couldn't even tab into the button because, well, Safari

LucyChyzhova (approved):
Works well! really nice! 🤩

LawrenceBorst:
> 

Hm I was thinking of merging this then fixing the merge conflict later

But actually closing this would be much simpler

PR #2376, opened 2023-06-29
Deprecrate events

adrianschmidt:
fix: #2377

Deprecate events that shadow native events (event that have the same name as native events):

- `change`
- `input`
- `select`

New events have been added, where the names have been prefixed.

Also, the property `prefix` is a "reserved public name", so that's also been deprecated and a new property, `limelPrefix` has been added.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/examples/custom-component-picker.tsx
+++ b/src/components/form/examples/custom-component-picker.tsx
@@ -99,7 +99,7 @@ export class CustomPickerExample implements FormComponent<number> {
                 disabled={this.disabled}
                 readonly={this.readonly}
                 required={this.required}
-                onChange={this.handleChange}
+                onLimelChange={this.handleChange}
                 searcher={this.search}
                 helperText={this.helperText}
             />
--- a/src/components/form/examples/custom-component-picker.tsx
+++ b/src/components/form/examples/custom-component-picker.tsx
@@ -44,10 +44,17 @@ export class CustomPickerExample implements FormComponent<number> {
     public helperText?: string;
 
     /**
+     * @deprecated Use `limelChange` instead.
      * Emitted when the value is changed
      */
     @Event()
     public change: EventEmitter<number>;
+
+    /**
+     * Emitted when the value is changed
+     */
+    @Event()
+    public limelChange: EventEmitter<number>;
 
     private heroes: Array<ListItem<number>> = [
         {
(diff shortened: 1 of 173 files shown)


src/components/form/examples/custom-component-picker.tsx
```diff
@@ -81,6 +88,8 @@ export class CustomPickerExample implements FormComponent<number> {
     ) => {
         event.stopPropagation();
         this.change.emit(event.detail?.value);
+        // When the event below is emitted, the whole form is emptied 😲
+        this.limelChange.emit(event.detail?.value);
```
adrianschmidt, line 92: I've tracked this down so far as to be reasonably sure this is because the event is not caught by the form, and it's therefore propagating past the form, and being caught by the handler listening for a `limelChange` coming from the form. Since that handler expects the whole formdata, the response when receiving a `null` because the value of this specific field is empty, is to empty all of the form-data.

However, I haven't quite figured out how to prevent that. I tried adding a listener for `limelChange`, but that didn't seem to do anything at all:

image

I'm thinking I probably need some help with this… @jgroth? 🙏

jgroth:
I'm not sure what to think of all these changes :smile: It's so much breaking stuff, are we really ever going to be able to remove the old events? We have both our own code and code of other consumers such as addons that need updating. Is it realistic and worth the effort just to get rid of the warnings?

I found some discussions on the topic [here](https://github.com/ionic-team/stencil/issues/2715) and [here](https://github.com/ionic-team/stencil/issues/2822), but there does not seem to be a conclusion. I think it's a bit weird that Stencil emits these warnings since it should be possible for both the implementer and the consumer to work around any eventual problems

Some other arguments for just keeping the old names.

- Is it a problem for anyone today that we use these event names? I don't think I have heard of any issues, and if a problem with this arises it should be possible to fix.
  - We can make sure our components with these event names never emit the native events by `stopPropagation` so there will not be any conflicts
  - Native events are always of the `Event` class, while our events are always of the `CustomEvent` class so consumers can already check if it is a custom event from a web component or not.
- The behavior of these events align with the behavior of the native events, having another name for them makes it less intuitive to use them.
- `CustomEvent` has the `detail` property which is not present on the regular `Event` class, which is meant for custom data and should alleviate the need for prefixed event names

adrianschmidt:
All good points. The problem arose when trying to upgrade to Stencil 3, because they have added strong typing for the `input` event, which made the compiler give an error when that event was listened to with an event handler that uses the typing from our custom event.

But that's perhaps possible to just circumvent with `// @ts-ignore`… 🤔

I'll try that!

adrianschmidt:
It turned out this wasn't necessary… 😂

Issue #2255, opened 2023-04-24
input field: helper texts need unique (random) IDs


Kiarokh:
I don't think the implementation of IDs is correct actually 👇 
```tsx
const helperTextId = 'tf-helper-text';
```
We need to generate random IDs for the helper text blocks of each `<limel-input-field />` on the page. Otherwise all input filed components will have helper texts with identical IDs this way, and the intended usage of the ID will fail 👇 

```tsx
 if (this.hasHelperText()) {
            properties['aria-controls'] = helperTextId;
            properties['aria-describedby'] = helperTextId;
        }
```

`aria-controls` and `aria-describedby` are supposed to give clues to assistive technologies, and with the current implementation, we are basically giving all existing helper text blocks on the page the exact same IDs (while they have different contents).

______
## Additionally
We should also add these `aria` properties to all other components that the `helperText` is used and is displayed.

PR #2379, opened 2023-07-03
Helper text ids

Kiarokh:
fix https://github.com/Lundalogik/lime-elements/issues/2255

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.e2e.ts
+++ b/src/components/input-field/input-field.e2e.ts
@@ -22,7 +22,6 @@ describe('limel-input-field', () => {
 
     types.forEach((type) =>
         describe(`with type="${type.name}"`, () => {
-            const expectedLabelId = 'tf-input-label';
             beforeEach(async () => {
                 page = await createPage(`
                     <limel-input-field
@@ -30,6 +29,7 @@ describe('limel-input-field', () => {
                         label="Test"
                     ></limel-input-field>
                 `);
+
                 limelInput = await page.find('limel-input-field');
                 inputContainer = await page.find(
                     'limel-input-field>>>label.mdc-text-field'
@@ -93,12 +93,6 @@ describe('limel-input-field', () => {
                     });
                 }
 
-                it('has the expected value for `aria-labelledby`', () => {
-                    expect(nativeInput).toEqualAttribute(
-                        'aria-labelledby',
-                        expectedLabelId
-                    );
-                });
                 it('has the class `mdc-text-field__input`', () => {
                     expect(nativeInput).toHaveClass('mdc-text-field__input');
                 });
@@ -113,9 +107,6 @@ describe('limel-input-field', () => {
                     expect(label).not.toHaveClass(
                         'mdc-floating-label--float-above'
                     );
-                });
-                it('has the expected id', () => {
-                    expect(label).toEqualAttribute('id', expectedLabelId);
                 });
                 describe('after focusing', () => {
                     beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


Kiarokh:
@adrianschmidt you might be interested about these series of PRs

adrianschmidt (dismissed):
Looks good to me. There are some tests failing because we compare html though, and obviously, the string is different every time… 😕

I don't know if there's some way to get around that? Seems like there ought to be… 🤔

Kiarokh:
oh, that might become an issue for [other PRs here](https://github.com/Lundalogik/crm-feature/issues/3502) as well.

src/util/random-string.ts
```diff
@@ -1,4 +1,15 @@
 export const createRandomString = () => {
+    if (
+        !('crypto' in window) ||
+        typeof window.crypto?.randomUUID !== 'function'
+    ) {
+        return legacyCreateRandomString();
+    } else {
```
Befkadu1, line 7: You don't need to have the else statement

src/util/random-string.ts
```diff
@@ -1,4 +1,15 @@
 export const createRandomString = () => {
+    if (
+        !('crypto' in window) ||
+        typeof window.crypto?.randomUUID !== 'function'
+    ) {
+        return legacyCreateRandomString();
+    }
+
+    return crypto.randomUUID().replace(/^\d/, 'a'); // ids must start with letters
```
adrianschmidt, line 9: Any particular reason why you've chosen to regex match for a digit and then replace it if found, rather than just prepending all UUIDs output with a letter? Seems like that would be both faster and simpler? 😊

LawrenceBorst: I like to live on the edge 😅, no good reason. It felt off to me to change the length/format of the id but there's no real reason to worry about that

Befkadu1: I personally prefer to avoid using a regular expression, if I can easily do the same thing with common string operations.

adrianschmidt: I'd even argue that it's better to break the UUID spec, since it's no longer a UUID when you change one of the characters to a fixed value. We might even want to prefix it with something like `a_` to make it even more obvious that it's not a UUID.

Befkadu1: That is a good point. However the consumer of `createRandomString` has no idea about `UUID` and it doesn't care if the random string is UUID or something else 😆.  Of course it becomes clear/readable and clean if we do it according to your suggestion. We strive for the best

src/util/random-string.ts
```diff
@@ -1,4 +1,15 @@
 export const createRandomString = () => {
+    if (
+        !('crypto' in window) ||
+        typeof window.crypto?.randomUUID !== 'function'
+    ) {
+        return legacyCreateRandomString();
```
john-traas, line 6: Is it expected that crypto.randomUUID won't be available in many cases? 

If so, why use it at all? Why not just use the legacyCreateRandomString function?

john-traas (commented):
I think it looks good. Just had a question about the two different random string/id generators used.

src/util/random-string.ts
LawrenceBorst: I suppose there is full support on all major browsers since about the end of 2021, although better safe than sorry. Still, the test suite still falls back on the legacy one (maybe it's not running a "secure context", I'm not sure)

src/components/input-field/input-field.e2e.ts
```diff
@@ -360,3 +354,10 @@ describe('limel-input-field', () => {
 async function createPage(content: string) {
     return newE2EPage({ html: content });
 }
+
+function replaceLabelId(HTML: string) {
+    return HTML.replace(
+        /"a_(\d|[a-f]){8}-(\d|[a-f]){4}-(\d|[a-f]){4}-(\d|[a-f]){4}-(\d|[a-f]){12}"/g,
+        '"tf-input-label"'
+    );
+}
```
Befkadu1, line 363: Why can't you just do it simply like this instead of using regex?
```suggestion
function replaceLabelId(labelId: string) {
    return `${a_labelId}` // add a doc here why we add the prefix
    );
}
```

john-traas (approved):

PR #2381, opened 2023-07-03
fix(checkbox): create a unique id for the helper text & use it by `aria` attributes

Kiarokh:
The `aria-controls` and `aria-describedby` are using these IDs to give clues to assistive technologies.
In the former implementation, we were lacking ids
for the helper text elements.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/checkbox/checkbox.template.tsx
+++ b/src/components/checkbox/checkbox.template.tsx
@@ -11,6 +11,7 @@ interface CheckboxTemplateProps {
     onChange?: (event: Event) => void;
     label?: string;
     helperText?: string;
+    helperTextId?: string;
 }
 
 export const CheckboxTemplate: FunctionalComponent<CheckboxTemplateProps> = (
@@ -41,6 +42,8 @@ export const CheckboxTemplate: FunctionalComponent<CheckboxTemplateProps> = (
                     disabled={props.disabled || props.readonly}
                     required={props.required}
                     onChange={props.onChange}
+                    aria-controls={props.helperTextId}
+                    aria-describedby={props.helperTextId}
                     {...inputProps}
                 />
                 <div class="mdc-checkbox__background">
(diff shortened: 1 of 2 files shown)


LawrenceBorst (commented):
Good stuff, but I only had one comment

I might be wrong, but I thought the `helpertextId` should be passed in as a prop to `limel-checkbox`, since we want to use it as an `id` on some other component to allow the `labelledby`/`describedby` attribute to point to it, which would mean some parent component should hold onto the `id`

Also, we might want to set `aria-required` if the field is required, and maybe `aria-disabled` if disabled in a future issue so we don't forget 😅

LawrenceBorst (approved):

Kiarokh:
thanks 😊 

### regarding `aria-required`, this is what I know:

If the `input` element of the checkbox component already has the standard HTML `required` attribute, there is no need to add the `aria-required` attribute. The required attribute is sufficient for indicating that the checkbox input must be filled out by the user before submitting the form.

The required attribute is a native HTML attribute that is understood by web browsers and is used for form validation. When the required attribute is present on an input element, it indicates to the browser that the input must have a value selected or entered before the form can be submitted. If the user tries to submit the form without filling out the required checkbox, the browser will display an error message prompting them to complete the input.

In most cases, using the `required` attribute alone is enough for form validation and accessibility purposes. The aria-required attribute is generally used for more complex custom form components where the native HTML attributes may not be sufficient to convey the required state to screen readers. But for a checkbox input, the required attribute should be enough, and adding aria-required in this case would be redundant.

PR #2389, opened 2023-07-10
feat(button-like-components): add helper function for enter clicking

LawrenceBorst:
Added helper function that you would put in componentWillLoad that adds the event listeners needed to make buttons clickable with enter.

We add it to limel-buttons to make testing easy (this will probably conflict with another similar issue, and we need to fix this conflict before merge).

Fixes Lundalogik/crm-feature#3518

## How to test this PR:
I tested it by throwing in a timeout in `componentWillLoad` that calls the add/remove event. I suggest trying something similar, calling them in different orders, maybe repeatedly, and see if it breaks.

The docs have a really good playground for the limel-button to test this functionality.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -1,4 +1,8 @@
-import { Component, h, Prop, State, Watch } from '@stencil/core';
+import { Component, h, Prop, State, Watch, Element } from '@stencil/core';
+import {
+    makeEnterClickable,
+    removeEnterClickable,
+} from '../../util/makeEnterClickable';
 
 /**
  * @exampleComponent limel-example-button-basic
@@ -66,7 +70,18 @@ export class Button {
     @State()
     private justLoaded = false;
 
+    @Element()
+    private host: HTMLElement;
+
     private justLoadedTimeout?: number;
+
+    public componentWillLoad() {
+        makeEnterClickable(this.host);
+    }
+
+    public disconnectedCallback() {
+        removeEnterClickable(this.host);
+    }
 
     public render() {
         return (
--- a/src/util/makeEnterClickable.ts
+++ b/src/util/makeEnterClickable.ts
@@ -1,0 +1,73 @@
+import { forceUpdate } from '@stencil/core';
+
+const eventHandlers: WeakMap<HTMLElement, CallBacks> = new WeakMap();
+
+export function makeEnterClickable(element: HTMLElement) {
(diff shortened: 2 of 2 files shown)


src/components/button/button.tsx
```diff
@@ -66,8 +67,15 @@ export class Button {
     @State()
     private justLoaded = false;
 
+    @Element()
+    private host: HTMLElement;
+
     private justLoadedTimeout?: number;
 
+    public componentWillLoad() {
+        makeClickable(this.host);
```
LawrenceBorst, line 76: Hm come to think of it, this is a bad name 😅 
I'll probably change it to something like `makeEnterClickable`

src/util/makeClickable.ts
```diff
@@ -0,0 +1,46 @@
+import { forceUpdate } from '@stencil/core';
+
+export function makeClickable(host) {
+    let isActive;
+    let hasJustReleasedEnter;
+
+    if (isActive === undefined || hasJustReleasedEnter === undefined) {
```
LawrenceBorst, line 7: This we can take out. Just initialise it above

src/util/makeClickable.ts
```diff
@@ -0,0 +1,46 @@
...
+                hasJustReleasedEnter = false;
+
+                return;
+            }
+
+            event.stopImmediatePropagation();
+        },
+        true
+    );
+}
```
jgroth, line 46: We should probably allow for some cleanup as well, if a consumer no longer wants the element to have this behavior. Now, once you have called this function you are stuck with the behavior forever.

src/components/button/button.tsx
```diff
@@ -1,5 +1,5 @@
 import { Component, h, Prop, State, Watch, Element } from '@stencil/core';
-import { makeClickable } from '../../util/makeClickable';
+import { makeEnterClickable } from '../../util/makeClickable';
```
LucyChyzhova, line 2: don't forget to change the file name to `makeEnterClickable.ts`

LucyChyzhova (approved):
Tested, works well!  🚀

PR #2392, opened 2023-07-12
feat(menu-list): add identifier to MenuItem interface

Befkadu1:
fix: https://github.com/Lundalogik/crm-feature/issues/3265
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

## Summary by CodeRabbit

- **New Features**
	- Enhanced menu functionality by adding a unique identifier for each item, paving the way for improved element targeting and future enhancements.

### DIFF
--- a/src/components/menu-list/menu-list-renderer.tsx
+++ b/src/components/menu-list/menu-list-renderer.tsx
@@ -138,6 +138,7 @@ export class MenuListRenderer {
                 role="menuitem"
                 aria-disabled={item.disabled ? 'true' : 'false'}
                 aria-selected={item.selected ? 'true' : 'false'}
+                data-item={item.identifier ? item.identifier : null}
                 data-index={index}
                 data-text={item.text}
                 {...attributes}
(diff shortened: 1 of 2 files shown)


jgroth (changes requested):
It's already possible to identify an item through the `value` property which the consumer has full control over, so adding this seem a bit superfluous

jgroth:
> I am not sure if I get what I need from `value`. 

No, you cannot, which is the whole point, we are trying to add something here that does not make much sense for the consumer. We are trying to shoehorn in a feature (tracking what a user has clicked on) in a place where it does not belong (rendering a menu item). 

> This needs thorough thinking how to fix the issue. I am glad to get suggestions.

It's already possible to track stuff today using commands. A command is used to capture user intent. Heap is used to track user actions. Those two already go hand in hand, so instead of using a generic command like `OpenBrowserTab` that we are using today, we should use a more specialized command so tracking it will be easer.

adrianschmidt:
Here's a comment I wrote on the issue:

https://github.com/Lundalogik/crm-feature/issues/3265#issuecomment-1477563014

Do you feel that makes sense @jgroth?

PR #2396, opened 2023-07-17
Make collapsible section enter clickable

LawrenceBorst:
Fixes Lundalogik/crm-feature#3535

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/collapsible-section/collapsible-section.tsx
+++ b/src/components/collapsible-section/collapsible-section.tsx
@@ -1,6 +1,17 @@
-import { Component, Event, EventEmitter, h, Prop } from '@stencil/core';
+import {
+    Component,
+    Event,
+    Element,
+    EventEmitter,
+    h,
+    Prop,
+} from '@stencil/core';
 import { dispatchResizeEvent } from '../../util/dispatch-resize-event';
 import { Action } from './action';
+import {
+    makeEnterClickable,
+    removeEnterClickable,
+} from 'src/util/makeEnterClickable';
 
 /**
  * @slot - Content to put inside the collapsible section
@@ -51,6 +62,25 @@ export class CollapsibleSection {
      */
     @Event()
     private action: EventEmitter<Action>;
+
+    @Element()
+    private host: HTMLElement;
+
+    public componentDidRender() {
+        const button = this.host.shadowRoot.querySelector(
+            '.open-close-toggle'
+        ) as HTMLElement;
+
+        makeEnterClickable(button);
+    }
+
+    public disconnectedCallback() {
+        const button = this.host.shadowRoot.querySelector(
+            '.open-close-toggle'
+        ) as HTMLElement;
+
+        removeEnterClickable(button);
+    }
 
     public render() {
         return (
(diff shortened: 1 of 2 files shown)


src/util/makeEnterClickable.ts
```diff
@@ -2,6 +2,12 @@ import { forceUpdate } from '@stencil/core';
 
 const eventHandlers: WeakMap<HTMLElement, CallBacks> = new WeakMap();
 
+/**
+ * Overrides the default browser behavior for clickable elements
+ * When focused and pressing down enter, avoids calling onClick repeatedly
+ * @param {HTMLElement} element the clickable element
+ */
+// eslint-disable-next-line sonarjs/cognitive-complexity
```
jgroth, line 10: We should not disable this rule :smile:

LawrenceBorst:
To test the changes made in the refactor, I suggest just having a `setTimeout` in `componentWillLoad`

e.g., make enter clickable, wait 5 seconds, remove enter clickable behaviour, wait 3 more seconds, make enter clickable (maybe twice, see what happens), remove it again and so on and so forth

It's worth making an issue to test that functionality, although it sounds potentially challenging

src/util/makeEnterClickable.ts
Kiarokh, line 1: maybe this commit should be `docs`?

src/util/makeEnterClickable.ts
```diff
@@ -2,6 +2,12 @@ import { forceUpdate } from '@stencil/core';
 
 const eventHandlers: WeakMap<HTMLElement, CallBacks> = new WeakMap();
 
+/**
+ * Overrides the default browser behavior for clickable elements
+ * When focused and pressing down enter, avoids calling onClick repeatedly
+ * @param {HTMLElement} element the clickable element
+ */
```
Kiarokh: maybe this first commit should be `docs`?

Kiarokh (approved):

src/components/collapsible-section/collapsible-section.tsx
```diff
@@ -52,6 +63,25 @@ export class CollapsibleSection {
...
 
+    @Element()
+    private host: HTMLElement;
+
+    public componentDidRender() {
+        const button = this.host.shadowRoot.querySelector(
+            '.open-close-toggle'
+        ) as HTMLElement;
+
+        makeEnterClickable(button);
```
adrianschmidt, line 74: `componentDidRender` is called every single time the component is re-rendered. While `makeEnterClickable` has an early exit for when the supplied element has already been "enhanced", we still query the DOM after every single re-render.

Also, if the implementation of `makeEnterClickable` were ever to change (intentionally or not), so that we no longer exit early but actually attach new listeners each time, even if the same element is passed, we might end up with a pretty awful leak here 😕

So:
1. Does this _really_ have to be called in `componentDidRender`?
2. If it does, please make sure we only do the DOM querying and calling of `makeEnterClickable` _once_ 🙂 (for example by adding a `private initialized = false`, which is checked before in initialising, and flipped to `true` when initialisation is completed)

@LawrenceBorst

src/util/makeEnterClickable.ts
adrianschmidt: Hang on, what happened here?

image

Was this thread resolved by mistake, or was there a good reason to resolve it with no change? @LawrenceBorst

PR #2403, opened 2023-07-19
Snackbar persistent

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/3355

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/snackbar/examples/snackbar-persistent-non-dismissible.tsx
+++ b/src/components/snackbar/examples/snackbar-persistent-non-dismissible.tsx
@@ -1,0 +1,66 @@
+import { Component, State, h } from '@stencil/core';
+
+/**
+ * Persistent and non-dismissible
+ * In some scenarios, a Snackbar that is persistent
+ * can simultaneously be non-dismissible.
+ *
+ * This is good for handling cases in which the system tries to
+ * communicate an ongoing background process or a status that the user
+ * cannot take an action on, but should be kept informed about.
+ *
+ * In such cases, you should programmatically close the Snackbar
+ * when the process is completed or the status changes!
+ */
+@Component({
+    tag: 'limel-example-snackbar-persistent-non-dismissible',
+    shadow: true,
+})
+export class SnackbarPersistentNonDismissibleExample {
+    @State()
+    private isOpen = false;
+
+    public render() {
+        return [
+            <limel-button
+                label="Show a persistent & non-dismissible alert"
+                onClick={this.triggerSnackbar}
+            />,
+            <limel-snackbar
+                message="Your internet connection was lost! Hang on while we're trying to reconnect…"
+                timeout={0}
+                open={this.isOpen}
+                dismissible={false}
+                onHide={this.handleHide}
+            />,
+            <limel-example-controls
+                style={{ '--example-controls-column-layout': 'auto-fit' }}
+            >
+                <limel-switch
+                    label="Toggle online status"
+                    onChange={this.handleChange}
+                />
+            </limel-example-controls>,
+        ];
+    }
+
+    private triggerSnackbar = () => {
+        this.isOpen = true;
+    };
+
+    private handleChange = (event: CustomEvent<boolean>) => {
+        if (!this.isOpen) {
+            return;
+        }
+
+        if (!event.detail) {
+            return;
+        }
+
+        this.isOpen = false;
+    };
+
+    private handleHide = () => {
+        this.isOpen = false;
+    };
+}
--- a/src/components/snackbar/examples/snackbar-persistent.tsx
+++ b/src/components/snackbar/examples/snackbar-persistent.tsx
@@ -1,0 +1,55 @@
+import { Component, State, h } from '@stencil/core';
+
+/**
+ * Persistent
+ * Dismissive Snackbars appear temporarily, and disappear automatically
+ * without requiring users to manually dismiss them.
+ * However, sometimes you may need to display a Snackbar
+ * which keeps being displayed persistently until the user deliberately takes
+ * an action or clicks the dismiss button. This could be to ensure that
+ *
+ * - the user actually reads message that is being displayed.
+ * - has time to take an informed decision to press the action button.
+ *
+ * To make a Snackbar persistent, set the `timeout` property to `null`.
+ */
+@Component({
+    tag: 'limel-example-snackbar-persistent',
+    shadow: true,
+})
+export class SnackbarPersistentExample {
+    @State()
+    private isOpen = false;
+
+    public render() {
+        return [
+            <limel-button
+                label="Show a persistent alert"
+                onClick={this.triggerSnackbar}
+            />,
+            <limel-snackbar
+                message="We use cookies to deliver services and analyze traffic."
+                actionText="Okay, I got it!"
+                open={this.isOpen}
+                timeout={null}
+                onAction={this.snackbarOnAction}
+                onHide={this.handleHide}
+            />,
+        ];
+    }
+
+    private triggerSnackbar = () => {
+        this.isOpen = true;
+    };
+
+    private snackbarOnAction = () => {
+        console.log(
+            'Great! You have just agreed with our terms & conditions without reading it.',
+        );
+        this.isOpen = false;
+    };
+
+    private handleHide = () => {
+        this.isOpen = false;
+    };
+}
--- a/src/components/snackbar/snackbar.tsx
+++ b/src/components/snackbar/snackbar.tsx
@@ -9,6 +9,7 @@ import {
     Method,
     Prop,
     State,
+    Watch,
 } from '@stencil/core';
 import translate from '../../global/translations';
 import { SnackbarContainer } from './container';
@@ -51,6 +52,12 @@ const hideAnimationDuration = 300;
 })
 export class Snackbar {
     /**
+     * `true` if the snackbar is open, `false` otherwise.
+     */
+    @Prop({ reflect: true })
+    public open: boolean = false;
+
+    /**
      * The text message to display.
      */
     @Prop()
@@ -58,8 +65,13 @@ export class Snackbar {
 
     /**
      * The amount of time in milliseconds to show the snackbar.
+     * If set to `null`, the snackbar will be persistent.
+     * This means:
+     * - either the end user will need to close is manually,
+     * which requires the `dismissible` property to be set to `true`.
+     * - or the snackbar needs to be closed programmatically.
      */
-    @Prop()
+    @Prop({ reflect: true })
     // eslint-disable-next-line no-magic-numbers
     public timeout?: number = 5000;
 
@@ -108,7 +120,7 @@ export class Snackbar {
     private offset: number = 0;
 
     @State()
-    private open: boolean = false;
+    private isOpen: boolean = false;
 
     @State()
     private closing: boolean = true;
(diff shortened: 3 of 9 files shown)


src/components/snackbar/snackbar.tsx
```diff
@@ -183,27 +183,27 @@ export class Snackbar {
 
         return (
             <div class="mdc-snackbar__actions" aria-atomic="true">
-                {this.renderActionButton(actionText)}
-                {this.renderDismissButton(dismissible)}
+                {this.renderActionButton()}
+                {this.renderDismissButton()}
```
LawrenceBorst, line 201: This is more a matter of style, some programmers prefer to expose all the used parameters in a function whereas others encapsulate the behaviour regardless. But I think if we change it here, it would be more consistent to do the same on line 180 and not pass anything into renderActions()

Kiarokh: I think in all of our newest components, we do as I did now in this refactor. That should mean our team prefers this stye 😄  I feel the other one requires a lot of bookkeeping and remembering stuff

src/components/snackbar/examples/snackbar-persistent-non-dismissible.tsx
```diff
@@ -0,0 +1,70 @@
+import { Component, Element, State, h } from '@stencil/core';
+
+/**
+ * Persistent and non-dismissible
+ * In some scenarios, a Snackbar that is persistent
+ * can simultaneously be non-dismissible.
+ *
+ * This is good for handling cases in which the system tries to
+ * communicate an ongoing background process or a status that the user
+ * cannot take an action on.
```
adrianschmidt, line 10: Is snackbar a good component to use for this case? I guess it's less intrusive than a banner, but the banner is intended to be persistent and non-blocking to begin with. The snackbar is intended to be transient and non-blocking. I think it feels less like a jump to me, to make it possible to have a banner that cannot be dismissed by the user, than to have a snackbar that's persistent… 🤔

I'm a little worried over two things when it comes to using a snackbar for this:
1. the snackbar might cover things
2. you're only supposed to show one snackbar at a time (we don't have good functionality for solving that, but adding that shouldn't be too hard, and is something we ought to do), and since the snackbar is fairly frequently used, I think we run a rather significant risk of conflicts

Now, obviously, with a banner, we also run the risk of covering things (or did we solve that already?). But since the banner is _supposed_ to be persistent, having it move things out of the way (by adding some padding to the top of the page, for example), makes sense for its design regardless of whether it's dismissible by the user or not (imo). That's not the case for snackbar.

adrianschmidt: > When you say "intended", I need to ask "intended by whom?" 😊

😄

> If you ask _me_, the banner (specially the current one, which is a ginormous block, covering and blocking everything below, specially the navigation passibility) is a horrible component, both when it comes to design and usability. Persistent banners would work perhaps a bit better if they would pop-up as a `sticky` component on the UI, so that they would still allow the user to scroll and interact with the rest of the page, and hopefully by doing so even be able to "fix" the issue/problem that the banner is warning about. But getting this behavior is so hard, specially from a component library's perspective. Because we never know how consumers will place the banner in the DOM and what scrolling structure or logic it will have. So the component itself cannot be so useful in this case. At least not in the CRM web client.

Yeah, makes sense. I agree, the banner shouldn't cover things, and it should _definitely_ be possible to scroll while it's showing. (But it's not preventing scrolling now, is it!? 😲)

> (we can't show several banners neither)

Yeah, absolutely true 👍

> When it comes to showing multiple snackbars, I wouldn't follow Material Design's old and silly conventions or rules here ("Don't show several snackbars!" or "snackbars shouldn't have icons" or "they should just have one action"). It took them 10 years to finally realize that _input fields cannot be just one line and a label on top of it_, even though they still don't acknowledge that and offer it as a valid style in their components.

I think there is at least one really good reason not to show several snackbars at once, and that's tied to the fact that they are (typically) only shown for a certain amount of time. If we show a snackbar for some given amount of time, let's say 10 seconds, you have 10 seconds to notice and read it. However, if we show two snackbars at the same time, each for 10 seconds, you effectively have only 5 seconds to read each.

So I don't think it's a silly convention that only one snackbar should be displayed at a time.

Kiarokh: > However, if we show two snackbars at the same time, each for 10 seconds, you effectively have only 5 seconds to read each.

Hm... Great argument. But there are many scenarios where information in the second snackbar could explain why the first snackbar was displayed. e.g.

_Snackabr 1:_

<table>
<tr><td>💀 Upload failed.    <kbd>Cancel</kbd> <kbd>Retry</kbd>
</table>



_Snackbar 2: (pending)_

<table>
<tr><td>🌍 You're currently offline…   <kbd>OK</kbd>
</table>

There are a ton of examples like this, where the user should take a decision, gathering information communicated from different sources. While your argument is true, I think the default time-out should be increased or programmatically controlled. 

Additionally, a healthy app needs a Notifications Center, in which all notifications will end up. Then users can always check again what the dismissed/missed info was about.

adrianschmidt: Yeah, the cool thing about us having a service for all the different parts of the app to use to send notifications, is that we can absolutely implement these things nowadays. We could cue up notification after each other, or add time to one that's already being displayed, and store the same message somewhere so it can be viewed in a notification centre at a later time.

This is starting to sound like a pitch!

(Also, this stuff might be really useful to have sorted the day we want to enable server-side push, since we might want to notify the user that certain data updated in the background.)

adrianschmidt:
Err… wasn't this already merged? I could swear I saw a notification that it had been merged… 🤨

src/components/snackbar/examples/snackbar-persistent-non-dismissible.tsx
Kiarokh: We actually got stuck on some limitation on our notification service in the web-client (or something like that). I don't really remember why we stopped working on this 😄

src/components/snackbar/examples/snackbar-persistent-non-dismissible.tsx
```diff
@@ -0,0 +1,68 @@
...
+ * Persistent and non-dismissible
+ * In some scenarios, a Snackbar that is persistent
+ * can simultaneously be non-dismissible.
+ *
+ * This is good for handling cases in which the system tries to
+ * communicate an ongoing background process or a status that the user
+ * cannot take an action on, but should be kept informed about.
+ *
+ * In such cases, you should programmatically close the Snackbar
+ * when the process is completed or the status changes!
```
Kiarokh, line 13: @jgroth I think this is were we got stuck last time if I remember correctly. How to close/dismiss a snackbar programmatically?

src/components/snackbar/examples/snackbar-persistent.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    public render() {
+        return [
+            <limel-button
+                label="Show a persistent alert"
+                onClick={this.triggerSnackbarWithAction}
+            />,
+            <limel-snackbar
+                message="We use cookies to deliver services and analyze traffic."
+                actionText="Okay, I got it!"
```
Kiarokh, line 39: or how to close it when an action button inside the snackbar itself is clicked

src/components/snackbar/examples/snackbar-persistent-non-dismissible.tsx
jgroth: It's not possible with how the component API is right now, we would have to add either an `open` property to replace the `show` method, or add a `hide` method in addition to the `show` method we already have.

Kiarokh: I can add. But I guess you prefer adding a `hide` method avoid breaking change, right?

jgroth: I would prefer the `open` property approach :smile: We can deprecate the `show` method to not make it breaking

src/components/snackbar/snackbar.tsx
```diff
@@ -76,6 +78,12 @@ export class Snackbar {
     @Prop()
     public dismissible: boolean = true;
 
+    /**
+     * Whether to show the snackbar with space for multiple lines of text
+     */
+    @Prop({ reflect: true })
+    public persistent: boolean = false;
```
jgroth, line 85: I don't think this is needed, it makes the API a bit ambiguous. The snackbar should be persistent if no timeout has been set

Kiarokh: isn't it more abstract and confusing to ask the consumer to set some value to `0`, if they don't the snackbar to disappear. Also that needs reading the docs and look at the examples to even understand that this is possible. You can't know it by just looking at the props.

😅 My brain's logic also thinks that if `timeout` is `The amount of time in milliseconds to show the snackbar.` the closer you get to zero, the shorter the snackbar will be displayed. And then if you set to to `0`, it should mean that the snackbar is not displayed at all. 🙃 Am I alone in thinking like this?

adrianschmidt: It's _very_ standard in programming that in cases where _actually zero_ would make no sense, 0 actually means unlimited. That's not something I'd expect the average programmer to have to look up, but if they do, we can just write it in the docblock, and it will show up in their editor's intellisense.

Having a separate property for persistent is ambiguous because what if I set `timeout` to `5000ms` and `persistent` to `true`? Those are contradictory, which is something we don't like in programming 🙂

adrianschmidt: To clarify, if this was user-facing, in the Admin interface for example, I would _absolutely_ argue for having a separate control for `persistent`, like a checkbox, and then either emptying in the timeout value or ideally even hiding the timeout input completely if `persistent` was `true` 🙂

Kiarokh: Ok, I'll remove it then. 

I always consider these props like UI controls, where toggling a switch (e.g. `persistent`) would "naturally" disable the `timeout` input field. But if `0` makes sense for the right audience (you guys), then we keep it.

p.s.
Checking back the old code we had here, I remember that MDC's original solution was setting the timeout to `-1` to make it persistent.

jgroth: I would say that manually setting timeout to `0` is a bit confusing, I would expect it to just disappear right away even if that makes no sense whatsoever for this component :smile: Setting timeout to `null`, `undefined` or `false` makes more sense to have it persistent, but I don't think we need to have any special handling for the `0` case and just let it be persistent then as well :smile:

jgroth: I guess it's also a bit confusing that the timeout has a default value of `5000`, so if they want it to be persistent they are forced to set a value like `<limel-snackbar timeout={null} />`

jgroth (approved):

jgroth (approved):

jgroth (approved):

PR #2405, opened 2023-07-20
Switch vs checkbox

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/3539

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/checkbox/checkbox.tsx
+++ b/src/components/checkbox/checkbox.tsx
@@ -14,6 +14,22 @@ import { createRandomString } from '../../util/random-string';
 import { CheckboxTemplate } from './checkbox.template';
 
 /**
+ * The Checkbox component is a classic and essential element in UI design that allows
+ * users to make multiple selections from a predefined list of options. The Checkbox component is commonly used in forms and settings interfaces to enable users to
+ * select one or more items from a list of choices.
+ *
+ * ## States of a Checkbox
+ * When a user clicks or taps on the box, it toggles between two states:
+ * Checked and Unchecked.
+ *
+ * However, a Checkbox can visualize a third state called the "Indeterminate" state.
+ * In this state, the checkbox appears as a filled box with a horizontal line or dash inside it.
+ *
+ * The Indeterminate state is typically used when dealing with checkbox groups
+ * that have hierarchical relationships or when the group contains sub-items.
+ * This state is used to indicate that some, but not all, of the items in a group are selected.
+ *
+ * <limel-example-switch-vs-checkbox />
  * @exampleComponent limel-example-checkbox
  * @exampleComponent limel-example-checkbox-helper-text
  */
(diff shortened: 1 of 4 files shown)


src/components/checkbox/checkbox.tsx
```diff
@@ -14,6 +14,20 @@ import { createRandomString } from '../../util/random-string';
...
+ * ## States of a Checkbox
+ * When a user clicks or taps on the box, it toggles between two states:
+ * Checked and Unchecked.
+ *
+ * However, a Checkbox can visualize a third state called the "Indeterminate" state.
+ * In this state, the checkbox appears as a filled box with a horizontal line or dash inside it.
+ *
+ * The Indeterminate state is typically used when dealing with checkbox groups
+ * that have hierarchical relationships or when the group contains sub-items.
+ * This state is used to indicate that that some, but not all, of the items in a group are selected.
```
LucyChyzhova, line 30:
```suggestion
This state is used to indicate that some, but not all, of the items in a group are selected.
```

src/components/switch/switch.tsx
```diff
@@ -12,6 +12,13 @@ import {
...
 
 /**
+ * The Switch component is a fundamental element in UI design that serves as a toggle switch
+ * to control the state of a specific setting or option in an application or website.
+ * The two distinct positions of the Switch are visually indicative of the two states:
+ * ON and OFF; making it easy for users to understand the current state of the controlled feature.
+ *
+ * The Switch component is widely used in user interfaces to enable users to
+ * quickly and intuitively change binary settings.
  * @exampleComponent limel-example-switch
```
LucyChyzhova, line 24: Maybe a bit more crm specific examples? 

Here are some situations where you might consider using a Switch component:
1. Settings and Preferences: Use a Switch component to allow users to enable or disable specific settings or 
preferences in your application. For example, you could use a Switch to let users toggle the "Dark Mode," 
"Notifications," or "Remember Me" options. 
2. Filters and Search Options: In search interfaces, you might use a Switch to enable or disable specific filters or 
search options. For example, you could use a Switch to let users toggle between "All" and "Mine" modes.

Kiarokh: But Lime Elements is not only for the CRM, so we cannot refer to things that require the reader to have a background knowledge of the CRM. We need to be more generic. Also "Dark Mode," "Notifications," or "Remember Me" options could be visualized either by using a switch or checkbox, depending on the implementation. There is nothing that says they must be _only_ done by using a switch.

LucyChyzhova (dismissed):

PR #2412, opened 2023-07-21
fix(switch): improve understanding of `readonly` & `disabled` states

LucyChyzhova:
make the component look more like a switch in `readonly` & `disabled` states

Fix: https://github.com/Lundalogik/crm-feature/issues/3312

Before:
Screenshot 2023-07-21 at 15 00 34

After:

Screenshot 2023-07-21 at 16 05 24

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/switch/partial-styles/_readonly.scss
+++ b/src/components/switch/partial-styles/_readonly.scss
@@ -2,13 +2,17 @@
     &.lime-switch--readonly {
         --mdc-switch-disabled-selected-track-color: var(--mdc-theme-primary);
         --mdc-switch-disabled-unselected-track-color: rgb(
-            var(--color-red-default)
+            var(--color-red-default),
+            0.6
         );
         --mdc-switch-handle-surface-color: transparent;
-        --mdc-switch-disabled-selected-handle-color: transparent;
-        --mdc-switch-disabled-unselected-handle-color: transparent;
+        --mdc-switch-disabled-selected-handle-color: var(--mdc-theme-primary);
+        --mdc-switch-disabled-unselected-handle-color: rgb(
+            var(--color-red-default),
+            0.8
+        );
 
-        --mdc-switch-disabled-track-opacity: 1;
+        --mdc-switch-disabled-track-opacity: 0.6;
 
         + label.disabled {
             cursor: default;
--- a/src/components/switch/switch.scss
+++ b/src/components/switch/switch.scss
@@ -66,7 +66,7 @@ $scale-factor: 0.8;
     --mdc-switch-selected-hover-track-color: var(--mdc-theme-primary);
 
     --mdc-switch-handle-elevation: var(--button-shadow-normal);
-    --mdc-switch-disabled-track-opacity: 0.4;
+    --mdc-switch-disabled-track-opacity: 0.3;
 }
 
 .mdc-switch {


Kiarokh (changes requested):
Thanks Lucy. Sorry for short response. I'm leaving for vacation soon, and can't elaborate a lot. But my intention was particularly _NOT TO_ render the `readonly` Switch, like a toggle switch. So this was so by design. 

`Readonly` is not a common state on such components. It's something we invented pretty much because of the CRM use-cases. When something is `readonly` it is not that _thing_ anymore. It is only visualization of data. While `disabled` is still that _thing_ that the component _is_ when it is enabled. Disabled is a real state of the component. So visualization  should resemble the same thing, but "disabled". It should mean, if something happens, the component can be "enabled" again.

Can you please add some background information about what was not good with the previous design and why you changed it?


Now since I am leaving for vacation and don't want this to be merged in the current shape, I'll request changes. We can go through it when both of us are back ;)

LawrenceBorst (approved):
Looks good, in dark and bright mode

LucyChyzhova:
> `Readonly` is not a common state on such components. It's something we invented pretty much because of the CRM use-cases. 
> 
For sure, users have never seen such a UI, and exactly because of it, I want to provide at least an understanding of what it is. 

> When something is `readonly` it is not that _thing_ anymore. It is only visualization of data. 
> 
I agree with this, however, I think the visualization is hard in the current UI. If we imagine using this element in a form, we could see something like this: 
Screenshot 2023-07-24 at 09 56 29

Will users understand that the newsletter feature is not available for that company? (based on Workplace comments this state is not clear for customers).  My intention was to make it more visually similar to an ordinary Switch component. So the user understands the concept - it was set to "false" (or better to say turned off) and it's not available, so there is nothing you can do to enable it.  I think it's more clear if we make it like this:

Screenshot 2023-07-24 at 10 35 39


> While `disabled` is still that _thing_ that the component _is_ when it is enabled. Disabled is a real state of the component. So visualization should resemble the same thing, but "disabled". It should mean, if something happens, the component can be "enabled" again.
> 
Yes, definitely. I didn't change that much for this `disabled` visualization. 

Screenshot 2023-07-24 at 10 42 23

> Can you please add some background information about what was not good with the previous design and why you changed it?
> 
Yes, as I mentioned, I'm most concerned about customer feedback as the "false" read-only state confuses users.
Screenshot 2023-07-24 at 11 04 13
I believe it happens because it was mostly our invention and the second reason is that the `read-only` state doesn't look like a `Switch` so users don't get the nature of it.


> Now since I am leaving for vacation and don't want this to be merged in the current shape, I'll request changes. We can go through it when both of us are back ;)

Sure, let's talk when we both back ;)

Kiarokh:
Thanks for the additional info.

Do you have a link to this Workplace post too? I've forgotten the background story of what they were complaining about. But I'm not sure if the confused users will _not_ be confused anymore, if they see the Switch with the new proposed design. 

I believe that the confusion comes mainly from the red color and the choice of label. Here is [another example](https://limetech.workplace.com/groups/limecrm/posts/5304196923017981/?comment_id=5304582582979415&reply_comment_id=5305048419599498).

image

In here, the UI says: 
> Skuld (Röd = Nej) [_translation_: Debt (Red = No)]. 

So they are using a boolean field to visualize whether a customer has any debts. When a normal brain sees a red color, they perceive it as a warning. So for an end-user it will be translated to:

🔴 Debt (`false`)
> Some users: 🤔 oh, it's red. I'm being warned that this customer has debts.

✅ Debt (`true`)
> Some users: 🤔 oh, it's green. I'm being told that everything is good, and this customer has no debts.

This is while the true and false statuses are opposite to what some users perceive. So these users will get confused about the meaning, and the "designer" has chosen to explain things with _text_ and say "when it's red, it means NO". This is to help the user to understand, but it's like explaining a joke. If you have to explain a joke, it's a bad joke.

Explaining things in text is always a sign of bad design. But the bad design here is very complicated actually. Almost nobody have managed to fix this issue completely, and, I'd like to argue that there is no bulletproof way to visualize the boolean field or boolean settings and be 100% sure that nobody will misunderstand what it means.

Users will always wonder, is this setting ON or OFF? Do they mean this is True or False? No matter how you do, if you are only relying on static colors and icons and labels.

A good example of this phenomenon is the [famous confusing microphone icon ](https://www.nngroup.com/articles/state-switch-buttons/)which is supposed to "mute". Not only nobody gets this right, but also different apps and systems randomly use one way or another for offering mute and unmute buttons in their UIs, which only add to the global confusion. 

Find more interesting examples [here](https://uxdesign.cc/mute-or-unmute-dilemma-during-the-work-from-home-era-94055f82cc74).

If the above design was like this 🔴 Debt (`true`) / ✅ Debt (`false`) it would be still confusing to another group of people. Even your new proposed design would be still confusing I'd say.



I think a better solution would be having a label that is dynamically changing in these cases (`readonly` boolean fields in cards). Something like:

Has debts (No) / Has debts (Yes) _or_ Has debts / Doesn't have debts

Receives Newsletters (No) / Receives Newsletters (Yes) _or_ Subscribed for Newsletters / Unsubscribed for Newsletters

Doing so would be easier for these cases, because the fields are `readonly` and not interactive. If they were interactive (they were toggle switches), then I'd recommend them to read those ☝️ articles that I linked above.

LucyChyzhova:
> Do you have a link to this Workplace post too? I've forgotten the background story of what they were complaining about.

sure, this is the post: https://limetech.workplace.com/groups/productupdate/posts/2432176610262850/?comment_id=2433426793471165&reply_comment_id=2433604703453374

LucyChyzhova:
> I believe that the confusion comes mainly from the red color and the choice of label. Here is [another example](https://limetech.workplace.com/groups/limecrm/posts/5304196923017981/?comment_id=5304582582979415&reply_comment_id=5305048419599498).
> 
> image
> 
> In here, the UI says:
> 
> > Skuld (Röd = Nej) [_translation_: Debt (Red = No)].

Very interesting case, thank you. Yes, colors are super confusing here, since "red" is not just a "no" sign but also an attention or even a danger mark. So having a debt (debt =true) is naturally a bad thing we should inform a user and show it as a red dot( or something) however if true is gree then we messed up. I 100% agree with you. My first thought was "Is it the right component for rendering debt then?" should not it be just a read-only input field? 

Screenshot 2023-08-21 at 10 53 56

Screenshot 2023-08-21 at 10 53 46

Otherwise having a more detailed or extended label is definitely a good idea!

Kiarokh:
Thanks @LucyChyzhova 
Your suggestion can also work. It really depends on the case. I don't think there will ever be a right design for the `readonly` of any boolean switch like this. The consumer should choose to go for a custom design. Or choose to set it on `disabled` if that helps better.

Can we agree to close this PR without merging it then?

Kiarokh:
We maybe found a better visualization here: https://github.com/Lundalogik/lime-elements/pull/2479

However, the consumers should still use correct labels to make this less confusing for their users

PR #2414, opened 2023-07-24
Fix icon cache not opening properly

LawrenceBorst:
Fixes Lundalogik/crm-feature#3546

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/global/icon-cache/in-memory-icon-cache.ts
```diff
@@ -0,0 +1,83 @@
...
+            iconPath = `${path}/`;
+        }
+
+        const response = await fetch(`${iconPath}assets/icons/${name}.svg`);
+
+        let svgData = await response.text();
+
+        // Some of the icons in the Icons8 library have hard coded black color on some of the paths.
+        // In order to apply coloring with CSS, these have to be set to 'currentColor'
+        svgData = svgData.replace(/#000000/g, 'currentColor');
```
LucyChyzhova, line 55: Could we apply our dark grey color here instead of black? Like:  rgb(var(--color-gray-darker)) ?
So it will work with the dark mode as well

LucyChyzhova (approved):

PR #2426, opened 2023-08-01
fix(slider): problems with resizing window

LawrenceBorst:
Fixes Lundalogik/crm-feature#3550
Maybe fixes Lundalogik/crm-feature#3287

## Review:

### Browsers tested:
Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/slider/slider.tsx
+++ b/src/components/slider/slider.tsx
@@ -4,6 +4,7 @@ import {
     Element,
     Event,
     EventEmitter,
+    forceUpdate,
     h,
     Host,
     Prop,
@@ -12,6 +13,7 @@ import {
 } from '@stencil/core';
 import { getPercentageClass } from './getPercentageClass';
 import { createRandomString } from '../../util/random-string';
+import { debounce } from 'lodash-es';
 
 /**
  * @exampleComponent limel-example-slider
@@ -110,14 +112,22 @@ export class Slider {
     private percentageClass: string;
 
     public constructor() {
+        const debounceTimeout = 200;
         this.inputHandler = this.inputHandler.bind(this);
         this.getContainerClassList = this.getContainerClassList.bind(this);
+        this.handleResize = debounce(
+            this.handleResize.bind(this),
+            debounceTimeout
+        );
+
         this.labelId = createRandomString();
         this.helperTextId = createRandomString();
     }
 
     public connectedCallback() {
         this.initialize();
+        window.addEventListener('resize', this.handleResize);
+        window.addEventListener('wheel', this.handleResize);
     }
 
     public componentDidLoad() {
@@ -180,6 +190,8 @@ export class Slider {
 
     public disconnectedCallback() {
         this.destroyMDCSlider();
+        window.removeEventListener('resize', this.handleResize);
+        window.removeEventListener('wheel', this.handleResize);
     }
 
     private getContainerClassList() {
@@ -261,6 +273,10 @@ export class Slider {
         );
     }
 
+    public componentDidRender() {
+        this.mdcSlider.initialSyncWithDOM();
+    }
+
     private renderHelperLine() {
         if (!this.helperText) {
             return;
@@ -304,6 +320,11 @@ export class Slider {
 
         this.reCreateSliderWithStep();
     }
+
+    private handleResize = () => {
+        this.mdcSlider.initialSyncWithDOM();
+        forceUpdate(this.rootElement);
+    };
 
     private updateDisabledState() {
         if (!this.mdcSlider) {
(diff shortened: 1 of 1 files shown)


src/components/slider/slider.tsx
```diff
@@ -261,6 +273,10 @@ export class Slider {
         );
     }
 
+    public componentDidRender() {
```
LawrenceBorst, line 276: For some reason it doesn't work quite as well without this extra synchronisation step. Is the debounce missed, or the synchronisation step called too early as Adrian suggested?

In any case, this fixes it

src/components/slider/slider.tsx
adrianschmidt, line 1: > Fixes https://github.com/Lundalogik/crm-feature/issues/3550
> Maybe fixes https://github.com/Lundalogik/crm-feature/issues/3287

Hmm… unfortunately this doesn't seem to fix either problem for me. I have only tested with the old webclient frontend though.

I've verified that it seems to be loading my local lime-elements, but I'll add some console output to the `handleZoom` function just to make absolutely sure I'm running the correct code.

src/components/slider/slider.tsx
```diff
@@ -105,19 +107,31 @@ export class Slider {
...
 
     @State()
     private percentageClass: string;
 
     public constructor() {
+        const debounceTimeout = 200;
         this.inputHandler = this.inputHandler.bind(this);
         this.getContainerClassList = this.getContainerClassList.bind(this);
+        this.handleZoom = debounce(
+            this.handleZoom.bind(this),
```
adrianschmidt, line 120: You shouldn't have to use `bind` when the original function is an arrow-function, as those bind `this` automatically. You don't have to change anything right now, but I wanted to comment on it for knowledge sharing, and also in case there is actually some reason this bind does actually need to be here 😊

src/components/slider/slider.tsx
```diff
@@ -105,19 +107,31 @@ export class Slider {
...
         this.labelId = createRandomString();
         this.helperTextId = createRandomString();
     }
 
     public connectedCallback() {
         this.initialize();
+        this.observer = new ResizeObserver(this.handleResize) as any;
+        this.observer.observe(this.rootElement);
+        window.addEventListener('resize', this.handleZoom);
+        window.addEventListener('wheel', this.handleZoom);
```
adrianschmidt, line 134: Hmm… here we're setting up listeners for `resize` and `wheel` events on `window`, but in `this.handleZoom`, we run `event.stopPropagation`. I guess it might not matter since we're catching the events in the bubbling phase, and presumably there's nothing left to bubble to after `window`, but it feels a bit dangerous to me to be running `stopPropagation` on those kinds of events… 🤔

LawrenceBorst: We can skip stopPropagation then. I think it makes the code a little simpler as well, since we can now use the same callback for the resize, wheel and resizeObserver callback.

In fact, I wrote.it on autopilot and it didn't occur to me that we were doing it on the window where it wouldn't propagate further up 😅

src/components/slider/slider.tsx
adrianschmidt: Ah, when I'm running a dev-build and log stuff to the console, it does seem to fix the problems! 👍
I wonder if I was running the wrong code somehow, or if it's a matter of timing differences or something… 🤔

LawrenceBorst: Did it work out in the end? I tried to see if it stopped working for me, but it seems to work, although I'm running on the new webclient

It's worth adding that there is this one glitch where you have to click twice to trigger the correct positioning of the slider. The slider has a tendency to jump to the left once when you zoom in, and right when you zoom out. In the end Lucy seemed okay with that minor bug. I couldn't quite bash it

I thought the componentDidRender() call would fix that particular problem, but it seems like it doesn't get called in Safari after a render, but it does in Chrome(!!!)

adrianschmidt: I've now played around a bit, and I've found a smaller solution, which I believe solves the same things this PR does, but with less code, and (I _think_) no new errors or warnings in the console (I might be wrong about that though. There's a frightening amount of stuff being printed in the console… 😬)

Please have a look, and see if you can find anything I've missed: #2429

One thing it doesn't fix is when you pinch-zoom in Safari, but that's because Safari only sends a single `wheel`  event when you start pinching, and then doesn't send any new events for the gesture, until you stop and begin a new pinching gesture. So that's just a bug in Safari, and nothing we should try to work around.

LawrenceBorst: I will have a look :)

LawrenceBorst:
Close in favor of #2429

PR #2434, opened 2023-08-07
Some small cleanups in limel-portal

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/portal/portal.tsx
```diff
@@ -44,38 +44,38 @@ export class Portal {
      * Decides which direction the portal content should open.
      */
     @Prop({ reflect: true })
-    public openDirection: OpenDirection = 'bottom';
+    public openDirection?: OpenDirection = 'bottom';
```
Befkadu1, line 47: Does it really matter adding `?` here? It seems unnecessary.
`openDirection= 'bottom'` when no value is passed to this component anyway. The default values here can be overridden by specifying a value for a prop with a default value.
The same comment for the changes on 53, 65, 71, 78 and 87

Befkadu1:
This is not a comment in this PR rather a question. You can omit it if you don't agree. I saw the other PR you made for `lime-tooltip` [here](https://github.com/Lundalogik/lime-elements/pull/2435/commits) and added `-basic` to filename of basic example. Maybe good to do it here as well?

src/components/portal/portal.tsx
adrianschmidt: Yeah, reading the typescript docs [here](https://www.typescriptlang.org/docs/handbook/2/objects.html#optional-properties), it seems like `?` should be used when a property is optional but there is _not_ a default value. So I guess this is unnecessary, yes. I'll remove it again 👍

adrianschmidt: ⚡

I dropped this commit entirely 👍

src/components/portal/portal.tsx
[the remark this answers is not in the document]
adrianschmidt: ⚡

Befkadu1 (approved):
👏

Issue #2446, opened 2023-08-16
Tooltip: add `openDirection` as a prop


Kiarokh:
## New feature motivation
Depending on where the trigger is located on the UI, you need different `openDirection`s for the tooltip content.

Right now, the default openDirection is `bottom-start`. Which will not look so nice very often:
<img width="237" alt="image" src="https://github.com/Lundalogik/lime-elements/assets/35954987/07715616-ad1e-477d-9f2d-de260536efa2"> <img width="444" alt="image" src="https://github.com/Lundalogik/lime-elements/assets/35954987/29309fce-b9c4-4b59-97c3-0e25e1ec0e34"> <img width="176" alt="image" src="https://github.com/Lundalogik/lime-elements/assets/35954987/9a963ca9-72ff-47b7-8fcb-2af9f361cc9e">

## New feature description
1. We want the default tooltip location to be `top`, since it is most of the the time the most un-intrusive place to open a tooltip, (since it doesn't cover useful info below in the UI, or doesn't end up below the cursor, etc…)
2. We want to allow customizing the tooltips's location, for example, instead of this: 
<img width="312" alt="image" src="https://github.com/Lundalogik/lime-elements/assets/35954987/9910613b-e1aa-4a8c-b810-9ee52e857e0d">
we want to render the tooltips like this:
<img width="252" alt="image" src="https://github.com/Lundalogik/lime-elements/assets/35954987/cb39d00d-510b-4ecd-b5c7-c207f51d035b">

PR #2447, opened 2023-08-16
feat(tooltip): add `openDirection` prop

Kiarokh:
Enables consumers to decide where to visually
render the tooltip, in relation to its trigger element.

fix https://github.com/Lundalogik/lime-elements/issues/2446

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tooltip/tooltip.tsx
+++ b/src/components/tooltip/tooltip.tsx
@@ -1,6 +1,7 @@
 import { Component, h, Prop, Element, State } from '@stencil/core';
 import { JSX } from 'react';
 import { createRandomString } from '../../util/random-string';
+import { OpenDirection } from '../menu/menu.types';
 
 const DEFAULT_MAX_LENGTH = 50;
 
@@ -82,6 +83,12 @@ export class Tooltip {
     @Prop({ reflect: true })
     public maxlength?: number = DEFAULT_MAX_LENGTH;
 
+    /**
+     * Decides the tooltip's location in relation to its trigger.
+     */
+    @Prop({ reflect: true })
+    public openDirection: OpenDirection = 'top';
+
     @Element()
     private host: HTMLLimelTooltipElement;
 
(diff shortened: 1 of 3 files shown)


src/components/tooltip/tooltip.tsx
```diff
@@ -82,6 +83,12 @@ export class Tooltip {
     @Prop({ reflect: true })
     public maxlength?: number = DEFAULT_MAX_LENGTH;
 
+    /**
+     * Decides the tooltip's location in relation to its trigger.
+     */
+    @Prop({ reflect: true })
+    public openDirection: OpenDirection = 'top';
```
adrianschmidt, line 90: Hmm… it's a bit of a pity that we didn't include this when making the component public, because now we're changing the default behaviour to be something different than what the behaviour was before… 😣

I guess it can't be helped, and it's not likely to affect anyone, but still.

adrianschmidt (dismissed):

src/components/tooltip/tooltip.tsx
Kiarokh: I thought about it a bit actually, but decided to keep things simple and trackable. But since this component has been a private component and consumers have used it at their own risk, I think this change is fine. Also, it's not really a breaking change I'd say. The previous behavior was so bad anyway... So this is a feature that makes things better.

But it would have been nicer to release the component publicly, already with this feature.

src/components/tooltip/tooltip.tsx
```diff
@@ -1,6 +1,7 @@
 import { Component, h, Prop, Element, State } from '@stencil/core';
 import { JSX } from 'react';
 import { createRandomString } from '../../util/random-string';
+import { OpenDirection } from '../menu/menu.types';
```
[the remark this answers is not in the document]
Kiarokh: I ~always~ make issues for things that I wanna do the next day after the day after tomorrow's, next year.

adrianschmidt (approved):

PR #2450, opened 2023-08-17
fix(table): differentiate interactivity of rows vs. components in cells

Kiarokh:
related to https://github.com/Lundalogik/crm-feature/issues/2757

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.scss
+++ b/src/components/table/table.scss
@@ -130,6 +130,10 @@ $table--limel-table--row-selector: 1;
             justify-content: center;
         }
     }
+
+    .tabulator-col-resize-handle {
+        display: none;
+    }
 }
 
 .tabulator-col,


src/components/table/table.tsx
```diff
@@ -683,6 +683,11 @@ export class Table {
         } else {
             row.getElement().classList.remove('active');
         }
+
+        const element = row.getElement().ownerDocument.createElement('div');
+        element.classList.add('interactive-feedback');
+
+        row.getElement().prepend(element);
```
Kiarokh, line 690: the problem with this implementation is that when the table loads, you get one such a div inside each row. 

But every time you click on a row, a new div with this class will be added to that row again
image

src/components/table/table.scss
```diff
@@ -130,6 +130,10 @@ $table--limel-table--row-selector: 1;
             justify-content: center;
         }
     }
+
+    .tabulator-col-resize-handle {
+        display: none;
+    }
```
Kiarokh, line 136: removes this annoying effect:
https://github.com/Lundalogik/lime-elements/assets/35954987/16317cbb-c59d-4653-a954-0c10c7526f96

src/components/table/partial-styles/tabulator-custom-styles.scss
```diff
@@ -1,32 +1,87 @@
...
+    // We use it to visualize interactive feedback.
+    position: absolute;
+    inset: 0;
+    min-height: 0.125rem;
+
+    // We could use `@include mixins.is-flat-clickable()`,
+    // But the "arguments" of the mixin would make the CSS for
+    // even and odd row colors and background color very complicated.
+    // So instead, I've copy/pasted the useful parts of the mixin here
+    transition: color 0.2s ease, background-color 0.2s ease,
```
LawrenceBorst, line 15: But do we need transform here? It never really transforms (as in scales/rotates etc.)?

src/components/table/partial-styles/tabulator-custom-styles.scss
```diff
@@ -1,32 +1,87 @@
...
+.tabulator-row:not(.tabulator-calcs) {
+    .tabulator-cell {
+        // We don't want the cells to block
+        // the hovering possibility of `interactive-feedback` div.
+        pointer-events: none;
+        // but we want the content in them
+        // (e.g. links, phone numbers, chips, components)
+        // to remain interactive.
+        > * {
+            pointer-events: all;
```
LawrenceBorst, line 28: Should we use auto here?

src/components/table/partial-styles/tabulator-custom-styles.scss
```diff
@@ -1,32 +1,87 @@
...
+            }
+
+            &:before {
+                $width-of-selected-row-indicator: 0.2rem;
+                content: '';
+                display: inline-block;
+                box-sizing: border-box;
+                position: sticky;
+                z-index: $table--has-interactive-rows--selectable-row--hover + 1;
+                inset: 0 0 auto 0.25rem;
```
LawrenceBorst, line 52: But the inset used to be 0 0 auto 0, just curious why it changed 😅

src/components/table/partial-styles/tabulator-custom-styles.scss
```diff
@@ -1,32 +1,87 @@
...
-
             &:not(.active) {
-                &:hover {
-                    background-color: rgb(
-                        var(--table-row-background-color--hover)
-                    );
+                .interactive-feedback {
+                    &:hover {
+                        background-color: var(
+                            --table-row-background-color--hover
```
LawrenceBorst, line 82: Nice to remove those unnecessary "rgb"'s :)
There's a few left, but this doesn't really matter

LawrenceBorst (commented):
Overall looks good
I'll make a fixup for the tests and maybe we can huddle

src/components/table/partial-styles/tabulator-custom-styles.scss
Kiarokh: oh yes. good point!

src/components/table/partial-styles/tabulator-custom-styles.scss
Kiarokh: Oh I was hoping the reviewer will miss this 😄 

I was too lazy and didn't really want to make a new commit for this change.

I just felt that the green circle was too close to the edge.
image
But I'll take it out from this commit now.

src/components/table/partial-styles/tabulator-custom-styles.scss
Kiarokh: https://github.com/Lundalogik/lime-elements/assets/35954987/28ab243e-7e57-4c4f-be88-6a4b0208eaf5

LawrenceBorst (approved):

src/components/table/partial-styles/tabulator-custom-styles.scss
Kiarokh: when the row is pressed (`active`), it slightly moves down. that's the transform

LawrenceBorst (approved):

PR #2463, opened 2023-08-29
Avoid rendering the component unless the value is a number

Befkadu1:
fix https://github.com/Lundalogik/crm-feature/issues/3585

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/linear-progress/linear-progress.tsx
+++ b/src/components/linear-progress/linear-progress.tsx
@@ -56,6 +56,10 @@ export class LinearProgress {
     }
 
     public render() {
+        if (!this.isFinite(this.value)) {
+            return;
+        }
+
         const classList = {
             'mdc-linear-progress': true,
             'mdc-linear-progress--indeterminate': this.indeterminate,
@@ -85,10 +89,14 @@ export class LinearProgress {
 
     @Watch('value')
     protected watchValue(newValue) {
-        if (!this.mdcLinearProgress) {
+        if (!this.mdcLinearProgress || !this.isFinite(newValue)) {
             return;
         }
 
         this.mdcLinearProgress.progress = newValue;
+    }
+
+    private isFinite(value: unknown) {
+        return Number.isFinite(value);
     }
 }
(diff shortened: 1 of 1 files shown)


src/components/linear-progress/linear-progress.tsx
```diff
@@ -56,6 +56,13 @@ export class LinearProgress {
     }
 
     public render() {
+        /**
+         * Avoid rendering the component when the prop 'value' is explicitly set to null.
+         */
+        if (this.isValueNull(this.value)) {
```
jgroth, line 62: What about `undefined`? Or what if it is an `object`? :smile:  
According to the type definition, the prop needs to be a `number`, so it might be better to check for that instead, e.g.
```suggestion
        if (!Number.isFinite(this.value)) {
```

Befkadu1: I will change the commit message later

src/components/linear-progress/linear-progress.tsx
```diff
@@ -57,9 +57,10 @@ export class LinearProgress {
 
     public render() {
         /**
-         * Avoid rendering the component when the prop 'value' is explicitly set to null.
+         * Avoid rendering the component when the prop 'value' explicitly set to null,
+         * undefined or object.
```
jgroth, line 61: This is a good example of why comments in code is almost never a good idea :smile: 
We made a change in the code, which forced us to also update the comment. However, the comment is not actually explaining what the code is doing and is confusing rather than helping. If we have a comment here it should probably say something like "Avoid rendering unless the value is a number". This is obvious from the code though, so there is no need for a comment at all. 
Comments in code is only warranted to answer _why_ it is doing something. Not _what_ it is doing, since that should already be obvious by reading the code.

jgroth (approved):

PR #2467, opened 2023-08-31
feat(breadcrumbs): add new component

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/3042

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/breadcrumbs/breadcrumbs.spec.tsx
+++ b/src/components/breadcrumbs/breadcrumbs.spec.tsx
@@ -1,0 +1,198 @@
+import { h } from '@stencil/core';
+import { newSpecPage, SpecPage } from '@stencil/core/testing';
+import { BreadcrumbsItem } from '@limetech/lime-elements';
+import { Breadcrumbs } from './breadcrumbs';
+
+let page: SpecPage;
+let handleSelect: jest.Mock;
+const buttonLikeItems: BreadcrumbsItem[] = [
+    {
+        text: 'step 1',
+        icon: {
+            name: 'fish',
+            color: 'rgb(var(--color-red-default))',
+        },
+    },
+    {
+        text: 'Step 2',
+        icon: {
+            name: 'cat',
+            color: 'rgb(var(--color-orange-default))',
+        },
+    },
+    {
+        text: 'Step 3',
+        icon: {
+            name: 'dog',
+            color: 'rgb(var(--color-blue-default))',
+        },
+    },
+];
+
+const hyperlinkItems: BreadcrumbsItem[] = [
+    {
+        text: 'Home',
+        link: {
+            href: '../../../..',
+            title: 'Start',
+        },
+    },
+    {
+        text: 'Products',
+        link: {
+            href: '../../../',
+            title: 'See all of our products',
+        },
+    },
+    {
+        text: 'Phones',
+        link: {
+            href: '../../',
+        },
+    },
+    {
+        text: 'Accessories',
+        link: {
+            href: '../',
+        },
+    },
+    {
+        text: 'Earphones',
+    },
+];
+
+describe('limel-breadcrumbs', () => {
+    describe('when button like items are provided', () => {
+        const ids = [];
+        beforeEach(async () => {
+            await initializeComponent(buttonLikeItems);
+            getIDs(page.root.shadowRoot.children[0], ids);
+        });
+
+        afterEach(() => {
+            jest.clearAllMocks();
+        });
+        it('renders the breadcrumbs with texts, icons and icon colors', () => {
+            expect(page.root).toEqualHtml(
+                `<limel-breadcrumbs>
+                    <mock:shadow-root>
+                        <ol aria-label="Breadcrumb" role="navigation" style="--limel-breadcrumbs-divider: '›';">
+                            <button class="step" id="${ids[0]}" role="listitem">
+                                <limel-icon name="fish" style="color: rgb(var(--color-red-default));"></limel-icon>
+                                <span class="text">step 1</span>
+                            </button>
+                            <button class="step" id="${ids[1]}" role="listitem">
+                                <limel-icon name="cat" style="color: rgb(var(--color-orange-default));"></limel-icon>
+                                <span class="text">Step 2</span>
+                            </button>
+                            <li aria-current="step" class="last step" tabindex="-1">
+                                <limel-icon name="dog" style="color: rgb(var(--color-blue-default));"></limel-icon>
+                                <span class="text">Step 3</span>
+                            </li>
+                        </ol>
+                    </mock:shadow-root>
+                </limel-breadcrumbs>`
+            );
+        });
+
+        it('emits select event when "step 1" button is clicked', async () => {
+            const button: HTMLButtonElement =
+                page.root.shadowRoot.querySelectorAll('button')[0];
+            button.click();
+
+            page.waitForChanges();
+
+            expect(handleSelect).toBeCalledTimes(1);
+            expect(handleSelect.mock.calls[0][0].detail).toEqual({
+                icon: {
+                    name: 'fish',
+                    color: 'rgb(var(--color-red-default))',
+                },
+                text: 'step 1',
+            });
+        });
+    });
+
+    describe('with custom divider', () => {
+        it('renders the breadcrumbs with the chosen divider', async () => {
+            await initializeComponent(buttonLikeItems, '+');
+
+            const oList: HTMLOListElement =
+                page.root.shadowRoot.querySelector('ol');
+            expect(
+                // eslint-disable-next-line @typescript-eslint/dot-notation
+                oList.style['_styles'].get('--limel-breadcrumbs-divider')
+            ).toEqual("'+'");
+        });
+    });
+
+    describe('when hyperlink items are provided', () => {
+        it('renders the breadcrumbs with links', async () => {
+            const ids = [];
+            await initializeComponent(hyperlinkItems);
+            getIDs(page.root.shadowRoot.children[0], ids);
+
+            expect(page.root).toEqualHtml(
+                `<limel-breadcrumbs>
+                    <mock:shadow-root>
+                        <ol aria-label="Breadcrumb" role="navigation" style="--limel-breadcrumbs-divider: '›';">
+                            <a class="step" href="../../../.." id=${ids[0]} role="listitem" title="Start">
+                                <span class="text">Home</span>
+                            </a>
+                            <a class="step" href="../../../" id=${ids[1]} role="listitem" title="See all of our products">
+                                <span class="text">Products</span>
+                            </a>
+                            <a class="step" href="../../" id=${ids[2]} role="listitem">
+                                <span class="text">Phones</span>
+                            </a>
+                            <a class="step" href="../" id=${ids[3]} role="listitem">
+                                <span class="text">Accessories</span>
+                            </a>
+                            <li aria-current="page" class="last step" tabindex="-1">
+                                <span class="text">Earphones</span>
+                            </li>
+                        </ol>
+                    </mock:shadow-root>
+                </limel-breadcrumbs>`
+            );
+        });
+    });
+});
+
+/**
+ * Ids are randomly generated and we do not have access to them as they
+ * are inside the breadcrumbs component. But we can get them this way.
+ * Retrieves ids via inorder traversal
+ * @param {Element} node base node
+ * @param {string[]} ids id array that you wish to populate
+ * @return void
+ */
+function getIDs(node: Element, ids: string[]) {
+    if (node.children && node.id) {
+        ids.push(node.id);
+    }
+
+    // eslint-disable-next-line @typescript-eslint/prefer-for-of
+    for (let i = 0; i < node.children.length; i++) {
+        getIDs(node.children[i], ids);
+    }
+}
+
+async function initializeComponent(
+    items: BreadcrumbsItem[],
+    divider: string = '›'
+) {
+    handleSelect = jest.fn();
+    page = await newSpecPage({
+        components: [Breadcrumbs],
+        template: () => (
+            <limel-breadcrumbs
+                items={items}
+                onSelect={handleSelect}
+                divider={divider}
+            />
+        ),
+    });
+
+    await page.waitForChanges();
+}
--- a/src/components/breadcrumbs/breadcrumbs.tsx
+++ b/src/components/breadcrumbs/breadcrumbs.tsx
@@ -1,0 +1,198 @@
+import {
+    Component,
+    Element,
+    h,
+    Event,
+    EventEmitter,
+    Prop,
+} from '@stencil/core';
+import { BreadcrumbsItem } from './breadcrumbs.types';
+import {
+    makeEnterClickable,
+    removeEnterClickable,
+} from '../../util/make-enter-clickable';
+import { createRandomString } from '../../util/random-string';
+
+/**
+ * A Breadcrumb consists of a list of distinct "places" that a user has gone through,
+ * before ending up where they are right now, in a website or an application.
+ *
+ * These "places" can be for example _pages_ of a website, which are hierarchically
+ * laid out before the current page that the user is looking at.
+ * They could also be _steps_ which the user has gone through, which perhaps have no
+ * hierarchical relation with each other, but has eventually led the user "here".
+ *
+ * :::note
+ * - Where the user currently is, is always the last step of the breadcrumb.
+ * - A breadcrumbs never shows where users can go after this place.
+ * It only illustrates where user has been before ending up here.
+ * If the path that a user can take is not changing and if next steps are clear,
+ * you can use the [Progress flow component](#/component/limel-progress-flow) instead.
+ * :::
+ *
+ * Breadcrumbs are often placed horizontally before the main content of the current screen.
+ * @exampleComponent limel-example-breadcrumbs-links
+ * @exampleComponent limel-example-breadcrumbs-buttons
+ * @exampleComponent limel-example-breadcrumbs-icons
+ * @exampleComponent limel-example-breadcrumbs-divider
+ * @exampleComponent limel-example-breadcrumbs-icon-color
+ * @exampleComponent limel-example-breadcrumbs-styling
+ */
+
+@Component({
+    tag: 'limel-breadcrumbs',
+    shadow: true,
+    styleUrl: 'breadcrumbs.scss',
+})
+export class Breadcrumbs {
+    /**
+     * List of items in the breadcrumbs,
+     * each representing a step or a page.
+     */
+    @Prop()
+    public items: BreadcrumbsItem[];
+
+    /**
+     * The visual divider that separates items.
+     * It must be a single character such as `-` or `,`.
+     */
+    @Prop()
+    public divider: string = '›';
+
+    /**
+     * Fired when clicking on buttons (not links!)
+     * inside the breadcrumbs.
+     */
+    @Event()
+    public select: EventEmitter<BreadcrumbsItem>;
+
+    @Element()
+    private host: HTMLLimelBreadcrumbsElement;
+
+    private button: HTMLButtonElement;
+    private anchor: HTMLAnchorElement;
+
+    public render() {
+        return (
+            <ol
+                role="navigation"
+                aria-label="Breadcrumb"
+                style={{ '--limel-breadcrumbs-divider': `'${this.divider}'` }}
+            >
+                {this.renderSteps()}
+                {this.renderLastStep()}
+            </ol>
+        );
+    }
+
+    public componentWillLoad() {
+        makeEnterClickable(this.host);
+    }
+
+    public disconnectedCallback() {
+        this.removeEnterClickable();
+    }
+
+    private renderSteps = () => {
+        const allStepsWithoutLast = this.items.slice(0, -1);
+
+        if (this.areItemsLinks(this.items)) {
+            return allStepsWithoutLast.map(this.renderAsLink);
+        }
+
+        return allStepsWithoutLast.map(this.renderAsButton);
+    };
+
+    private renderAsButton = (item: BreadcrumbsItem) => {
+        const tooltipId = createRandomString();
+
+        return [
+            <button
+                role="listitem"
+                id={tooltipId}
+                class="step"
+                onClick={this.handleClick(item)}
+            >
+                {this.renderIcon(item)}
+                {this.renderLabel(item)}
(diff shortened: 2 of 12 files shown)


Kiarokh:
We only need a few "**_useful_**" tests.

Kiarokh:
@jgroth can you please have a look and see if there is any obviously stupid thing in this component and its APIs? If not, We'd like to merge this.

src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,216 @@
...
+            <li
+                class="last step"
+                tabindex="-1"
+                aria-current={this.areItemsLinks(items) ? 'page' : 'step'}
+            >
+                {this.renderIcon(lastItem[0])}
+                <span class="text">{lastItem[0].text}</span>
+            </li>
+        );
+    };
```
Kiarokh, line 163: The reason we did it was [this guideline from nngroup](https://www.nngroup.com/articles/breadcrumbs/), which states:
> **5. In the breadcrumb trail, the breadcrumb corresponding to the current page should not be a link.**
You should never have a link that does nothing. The last breadcrumb (denoting the current page) should not be a link. To avoid confusing users, visually differentiate between the current page and the preceding linked breadcrumb items, preferably using underlined or blue text.


If we _only_ had buttons in the structure, we could probably just make the last `button` `disabled` and give it give it a `tabindex="-1"`. But the component can also have links. The HTML <a> (anchor) element itself doesn't have a built-in `disabled` attribute like some other form elements such as `<input>` or `<button>`. It's also not enough to rely on using CSS hacks (such as adding `pointe-events: none;`) to make the link non-clickable. Because users would still be able to tab into it and press <kbd>Enter</kbd> to activate the link. 

Therefore, we decided to render the last item as a `li` that holds a `span`.

src/components/breadcrumbs/breadcrumbs.types.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * Icon of the step.
+     */
+    icon?: string;
+
+    /**
+     * If `true`, text will be rendered as a tooltip
+     */
+    iconOnly?: boolean;
```
jgroth, line 17: I think I mentioned this in another PR, but will mention it here as well. It feels wrong to name a property like this, because it might become mutually exclusive with other props that gets introduced later. What would happen if all these are set? 

```ts
{
    iconOnly: true,
    textOnly: true,
    linkOnly: true,
}
```

It becomes hard to know what behavior this will give us since these props will not work together at the same time. This might not be a realistic example for this component, but I hope you get my point anyway :smile: 

A better alternative could be to either have a `type` property on the item that would give an indication on how it should be used/rendered, or have a property named `disableText` or similar

Kiarokh: 🤔 it makes sense. But I don't remember why we didn't change the `type` after [the discussion](https://github.com/Lundalogik/lime-elements/pull/2315#discussion_r1214000403) for Action Bar. Why did we let the other one in the Actionbar be merged with this name? 

Is it better to keep it consistent? Or are you suggesting to rename both?

In that case, I'd prefer a type like `hideText`, but that doesn't look nice neither? Because one could argue, what about `hideIcon` and `hideLink` and so forth. Or? In this case, a better name could be `renderTextAsTooltip` or `textAsTooltip` maybe?

Anyhow, not sure what to do here.... Continue, or change this in both components?

Kiarokh: Oh I see now what you mean with `type`. Yes that's perhaps a better way. But we have two types right now. It's either `iconAndText` (the default) or `iconOnly` (which will render the text in the tooltip). So it's possible to change that.

> We can't change the name in the action bar now without making a breaking change, but we can add an additional property with the new name and deprecate the old one and then remove it later.

Nobody is using the component yet except us. 
https://github.com/search?q=org%3ALundalogik%20limel-action-bar&type=code

So I don't think we should be bureaucratic about this at all, and change it in Elements and CRM components in a quick PR now. It's fine if it's a BREAKING CHANGE. Since nobody is using it anyways. Please let's not deprecate it. We just make it a breaking change.

Green light?

src/components/breadcrumbs/examples/breadcrumbs-buttons.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+    public render() {
+        return [
+            <limel-breadcrumbs
+                items={this.items}
+                onSelect={this.handleSelect}
+            />,
+            <limel-example-value
+                label={'last select item'}
+                value={this.selectedItem?.text}
```
Befkadu1, line 52:
```suggestion
                value={this.selectedItem.text}
```

src/components/breadcrumbs/breadcrumbs.types.ts
jgroth: > Oh I see now what you mean with type. Yes that's perhaps a better way. But we have two types right now. It's either iconAndText (the default) or iconOnly (which will render the text in the tooltip). So it's possible to change that.

I don't think you should be forced to set a type. We can include `iconAndText` if you want, but I think it should default to that anyway if nothing has been set

> So I don't think we should be bureaucratic about this at all, and change it in Elements and CRM components in a quick PR now. It's fine if it's a BREAKING CHANGE. Since nobody is using it anyways. Please let's not deprecate it. We just make it a breaking change.

This is why we should have stuff in beta for a while before releasing it :smile:

Kiarokh: > This is why we should have stuff in beta for a while before releasing it 😄

You're technically right 😄 But practically not 🤣 because nobody is using Lime Elements, except Lime CRM folks 🤪

src/components/breadcrumbs/examples/breadcrumbs-links.tsx
```diff
@@ -0,0 +1,65 @@
...
+            text: 'Products',
+            link: {
+                href: '../../../',
+                title: 'See all of our products',
+            },
+        },
+        {
+            text: 'Phones',
+            link: {
+                href: '../../',
```
Befkadu1, line 48: Some of the items have title and here no title. Is it intentional ? It looks inconsistent

src/components/breadcrumbs/examples/breadcrumbs-icon-only.tsx
```diff
@@ -0,0 +1,49 @@
...
+        {
+            text: 'Products',
+            icon: 'shop',
+        },
+        {
+            text: 'Phones',
+            icon: 'google_mobile_copyrighted',
+        },
+        {
+            text: 'Accessories',
```
Befkadu1, line 42: In the note it say "The last item (current step) will always display both an icon and the text." but there is no icon in the last item.
Screenshot 2023-09-20 at 4 03 04 PM

src/components/breadcrumbs/examples/breadcrumbs-buttons.tsx
Kiarokh: this change generates an error @Befkadu1 
image

src/components/breadcrumbs/breadcrumbs.types.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * If supplied, the breadcrumbs steps will be a clickable links.
+     */
+    link?: Omit<Link, 'target' & 'text'>;
```
jgroth, line 22:
```suggestion
    link?: Omit<Link, 'target' | 'text'>;
```

Kiarokh, line 22: not sure if this is the right way to do it, and not sure if we need to add something to the docs above. For example, write something like 
> only `title` and `href` are allowed in the link interface

?

Kiarokh: I'm not sure how it helps, since no errors or warnings are thrown anywhere, if I add `target` in an example, like below

image

src/components/breadcrumbs/examples/breadcrumbs-buttons.tsx
Befkadu1: Shouldn't we have a selectedItem from the beginning ? Here it is undefined at the beginning

src/components/breadcrumbs/examples/breadcrumbs-buttons.tsx
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Keep in mind that the last item will not be rendered as an
+ * HTML button and and therefore won't be clickable.
+ */
+@Component({
+    tag: 'limel-example-breadcrumbs-buttons',
+    shadow: true,
+})
+export class BreadcrumbsExampleButtons {
+    private items: BreadcrumbItem[] = [
```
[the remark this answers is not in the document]
Kiarokh: it was your fault 😅 🤣 made a separate comment about this and revereted a suggested change

src/components/breadcrumbs/examples/breadcrumbs-buttons.tsx
Kiarokh: 🤔 there is no selected item, or? 
The last step is always the "current" step, but it is never "selected". It's just an static thing. Or perhaps I misunderstand what you mean

src/components/breadcrumbs/examples/breadcrumbs-links.tsx
```diff
@@ -0,0 +1,65 @@
...
+    tag: 'limel-example-breadcrumbs-links',
+    shadow: true,
+})
+export class BreadcrumbsExampleLinks {
+    @State()
+    private items = [
+        {
+            text: 'Home',
+            link: {
+                href: '../../../..',
```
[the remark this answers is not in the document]
Kiarokh: The idea was to show the developers how links are logically organized. Like the first one is 4 steps back, next goes 3 steps back and so forth. Is it bad?

src/components/breadcrumbs/examples/breadcrumbs-links.tsx
Kiarokh: intentional

src/components/breadcrumbs/examples/breadcrumbs-icon-only.tsx
Kiarokh: Good catch! Will fixup

src/components/breadcrumbs/examples/breadcrumbs-buttons.tsx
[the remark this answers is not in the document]
Befkadu1: Yea I suggested to remove the decorator `@State` because it is not a state.

src/components/breadcrumbs/examples/breadcrumbs-links.tsx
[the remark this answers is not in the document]
Befkadu1: It is not bad but good to add a doc maybe? People tend to click it to try it out and then ended up to the broken link

src/components/breadcrumbs/breadcrumbs.types.ts
jgroth: There is no type for the `items` array in your example, so how can it complain? You are allowing anything :smile: 
Try this
```ts
class BreadcrumbsExample {
    private items: BreadcrumbItem[] = [...]
}
```

src/components/breadcrumbs/breadcrumbs.types.ts
jgroth: This is still an issue

Kiarokh: Ok, I've now added a `type`. So questions again:
1. What convention should we use for types? `icon-only` or `iconOnly`?
2. Can we decide a name? I don't find a good name.

src/components/breadcrumbs/breadcrumbs.types.ts
```diff
@@ -0,0 +1,31 @@
...
+export interface BreadcrumbsItem {
+    /**
+     * label displayed on the step.
+     */
+    text: string;
+
+    /**
+     * Icon of the step.
+     */
+    icon?: string;
```
Kiarokh, line 12: we may wanna wait for this new icon interface: https://github.com/Lundalogik/lime-elements/pull/2489 to be merged, so that we can use it here instead?

src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,221 @@
...
+                {this.renderSteps(this.items)}
+                {this.renderLastStep(this.items)}
+            </ol>
+        );
+    }
+
+    public componentWillRender() {
+        // Avoid polluting the "clickable registry" with unreferenced nodes
+        this.removeEnterClickable();
+    }
```
Befkadu1, line 91: Is this by mistake? I don't think we need it

src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,221 @@
...
+    public componentDidRender() {
+        this.button = this.host.shadowRoot.querySelector('button');
+        this.anchor = this.host.shadowRoot.querySelector('a');
+        if (this.button) {
+            makeEnterClickable(this.button);
+        }
+
+        if (this.anchor) {
+            makeEnterClickable(this.anchor);
+        }
```
Befkadu1, line 102: The function can handle the conditions
```suggestion
     this.removeEnterClickable();
```

src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,221 @@
...
+    };
+
+    private removeEnterClickable() {
+        if (this.button) {
+            removeEnterClickable(this.button);
+        }
+
+        if (this.anchor) {
+            removeEnterClickable(this.anchor);
+        }
```
Befkadu1, line 219:
```suggestion
   const element = this.button ?? this.anchor;
   removeEnterClickable(element);
```

src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,221 @@
...
+            return;
+        }
+
+        return <span class="text">{item.text}</span>;
+    };
+
+    private renderTooltip = (item: BreadcrumbsItem) => {
+        if (!item.type) {
+            return;
+        }
```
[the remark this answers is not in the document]
Kiarokh: seems to be cause due to IDs 
image
The `element-id` is different than the id of the trigger :/

src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,218 @@
...
+        if (this.anchor) {
+            makeEnterClickable(this.anchor);
+        }
+    }
+
+    public disconnectedCallback() {
+        this.removeEnterClickable();
+    }
+
+    private renderSteps = (items: BreadcrumbsItem[]) => {
```
Kiarokh, line 109: I think we should refactor and move out the code for items to another file called `breadcrumbs-item.tsx` (like what we do for `action-bar-item.tsx`). We should handle things in there, including IDs and such. What do you think?

src/components/breadcrumbs/examples/breadcrumbs-links.tsx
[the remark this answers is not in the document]
Kiarokh: I think it's fine

src/components/breadcrumbs/breadcrumbs.tsx
Kiarokh: this is needed when we want to prevent the <kbd>Enter</kbd> to trigger the action too many times. (when the key is pressed down and held)

src/components/breadcrumbs/breadcrumbs.tsx
Kiarokh: r you sure? the original one is `makeEnterClickable` and your suggestion is `removeEnterClickable`. Also it looks weird to have the 2 lines before this suggestion if we change it

src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        if (this.anchor) {
+            makeEnterClickable(this.anchor);
+        }
+    }
+
+    public disconnectedCallback() {
+        this.removeEnterClickable();
+    }
```
Kiarokh, line 113: @LucyChyzhova @LawrenceBorst we changed this section to:
```tsx
    public componentWillLoad() {
        makeEnterClickable(this.host);
    }

        this.removeEnterClickable();
    }
```
which seems to produce the exact same effect. If you have any objections let us know. we don't exactly know why the code was written that way.

src/components/breadcrumbs/breadcrumbs.types.ts
```diff
@@ -0,0 +1,27 @@
+import { Link } from '@limetech/lime-elements';
+import { Icon } from 'src/global/shared-types/icon.types';
```
Kiarokh, line 2: @jgroth suggested that this import should be like this:
`import { Link, Icon } from '@limetech/lime-elements';`

But that generates this error:
image

I don't really know how it's possible to write `import { Link } from '@limetech/lime-elements';`. Because that is removed from exports here: https://github.com/Lundalogik/lime-elements/commit/60578ba0501c9a9cd8d0de10bd1bbfbc207c49af#diff-4f3136d98245087ba43f112ee10bfb0fcc0147dff7afa6a36c8f3f61555821cbL2

This is anyhow a discussion for the engineers. I leave it like this for now

jgroth: Importing from `@limetech/lime-elements` is more important in the examples so that they look correct from a consumer perspective. 

`Icon` is not exported from there yet I think because it generated some build error, probably the same kind of error we got when working on #2370

Kiarokh: I see. But we don't import in examples. So this should be alright then. If something should happen regarding the export of `Link` and `Icon`, I suspect you guys will do it elsewhere.

src/components/breadcrumbs/breadcrumbs.tsx
```diff
@@ -0,0 +1,203 @@
...
+     * The visual divider that separates items.
+     * It must be a single character such as `-` or `,`.
+     */
+    @Prop()
+    public divider: string = '›';
+
+    /**
+     * Used to attach the right tooltip to the right item
+     */
+    private tooltipId: string;
```
Befkadu1, line 65: remove this, we don't need to have it. We can declare it inside the function on line 112 and 129.  No need to have book-keeping

src/components/breadcrumbs/breadcrumbs.spec.tsx
```diff
@@ -0,0 +1,247 @@
...
+            handleSelect = jest.fn();
+            page = await newSpecPage({
+                components: [Breadcrumbs],
+                template: () => (
+                    <limel-breadcrumbs
+                        items={buttonLikeItems}
+                        onSelect={handleSelect}
+                    />
+                ),
+            });
```
Befkadu1, line 76: We are re-writing these lines of codes multiple time. I think we can have a function like `createComponent(items BreadcrumbsItem[], divider?: string)` and put these lines of codes there

src/components/breadcrumbs/examples/breadcrumbs-buttons.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * will automatically generate a list of `buttons`s.
+ *
+ * Keep in mind that the last item will not be rendered as an
+ * HTML button and and therefore won't be clickable.
+ */
+@Component({
+    tag: 'limel-example-breadcrumbs-buttons',
+    shadow: true,
+})
+export class BreadcrumbsExampleButtons {
```
Befkadu1, line 24: We usually add `Example` as postfix. Check it in other component examples
```suggestion
export class BreadcrumbsButtonsExample {
```

src/components/breadcrumbs/examples/breadcrumbs-buttons.tsx
```diff
@@ -0,0 +1,66 @@
...
+                value={this.selectedItem.text}
+            />,
+        ];
+    }
+
+    private handleSelect = (
+        event: LimelBreadcrumbsCustomEvent<BreadcrumbsItem>
+    ) => {
+        this.selectedItem = event.detail;
+        console.log('jellit');
```
Befkadu1, line 64: What is this 😛 ?

src/components/breadcrumbs/breadcrumbs.tsx
Kiarokh: We do the exact same thing here https://github.com/Lundalogik/lime-elements/blob/cf6fc241059b2a6cac46f7d67212a02372812060/src/components/action-bar/action-bar-item/action-bar-item.tsx#L51

Befkadu1: That is different. In `action-bar-item` its value is saved throughout the component's life time. You can see that we initialized in the constructor. In your case, we assign it for every button since every button has its own tooltipId

src/components/breadcrumbs/breadcrumbs.spec.tsx
Kiarokh: thanks for the helps 🙏

Befkadu1 (approved):
It looks ok now 👏

Kiarokh:
Thanks a lot for all the help @Befkadu1

src/components/breadcrumbs/breadcrumbs.tsx
adrianschmidt: @Kiarokh (and whoever the rest of "we" is, @Befkadu1 perhaps? 😊) It seems there was some cleanup missed when this was done. The `removeEnterClickable` still tried to remove the listeners from the anchor or button, which meant we had a leak here. But yes, not running two DOM queries and then setting up new event listeners on _every_ re-render is much better, so thanks for fixing that! 🙏 ❤️

I've opened #2748, which fixes the leak. Feel free to review! 🙏

Kiarokh: Thanks Adrian. I don't know anything about these lifecycle stuff and what leaks where or how to fix it. I hope someone else can look into these issues.

PR #2472, opened 2023-09-08
fix(form): separate code editor from other elements

LucyChyzhova:
Fix: https://github.com/Lundalogik/crm-feature/issues/3602
Screenshot 2023-09-08 at 17 37 51

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/form.scss
+++ b/src/components/form/form.scss
@@ -25,6 +25,10 @@
     }
 }
 
+limel-code-editor {
+    margin-bottom: 0.75rem;
+}
+
 .limel-form-array-item--object {
     margin-bottom: 0.25rem;
 }


src/components/form/form.scss
```diff
@@ -25,6 +25,10 @@
     }
 }
 
+limel-code-editor {
```
LawrenceBorst, line 28: Tested. Works :)

Not sure why we only want it on code editors nested in forms, but don't know the context 😅

LawrenceBorst (approved):

src/components/form/form.scss
LucyChyzhova: we have it for the `check-box` as well. 😉
In our case all input fields have a `helper-text`, so extra space is reserved for it, however, `limel-code-editor` doesn't have this feature, so the gap looks significantly smaller. Until now we were using this component only at the bottom of the form, so there was no need for adding a margin-bottom.

PR #2473, opened 2023-09-11
Feat/limel menu cascading items

TommyLindh2:
fix: https://github.com/Lundalogik/solution-helpdesk-poc/issues/23

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu/examples/menu-subitems-lazy-loading.tsx
+++ b/src/components/menu/examples/menu-subitems-lazy-loading.tsx
@@ -1,0 +1,81 @@
+import {
+    MenuItem,
+    ListSeparator,
+    LimelMenuCustomEvent,
+} from '@limetech/lime-elements';
+import { Component, State, h } from '@stencil/core';
+
+const NETWORK_DELAY = 1000;
+/**
+ * Lazy loading infinite amount of sub menu items.
+ */
+@Component({
+    tag: 'limel-example-menu-subitems-lazy-loading',
+    shadow: true,
+})
+export class MenuSubItemsLazyLoadingExample {
+    private items: Array<MenuItem | ListSeparator> = [];
+
+    @State()
+    private lastSelectedItem: string;
+
+    public componentWillLoad() {
+        this.items = [
+            {
+                text: 'Item 1',
+                items: this.handleLoadSubItems,
+            },
+            {
+                text: 'Item 2',
+                items: this.handleLoadSubItems,
+            },
+            {
+                text: 'Item 3',
+                items: this.handleLoadSubItems,
+            },
+            {
+                text: 'Item 4',
+                items: this.handleLoadSubItems,
+            },
+            {
+                text: 'Item 5',
+                items: this.handleLoadSubItems,
+            },
+        ];
+    }
+
+    public render() {
+        return [
+            <limel-menu items={this.items} onSelect={this.handleSelect}>
+                <limel-button label="Menu" slot="trigger" />
+            </limel-menu>,
+            <limel-example-value
+                label="Last selected item"
+                value={this.lastSelectedItem}
+            />,
+        ];
+    }
+
+    private handleLoadSubItems = async (
+        item: MenuItem
+    ): Promise<MenuItem[]> => {
+        return new Promise<MenuItem[]>((resolve) => {
+            // Simulate some network delay
+            setTimeout(() => {
+                const subItems = [];
+                for (let i = 1; i < 6; i++) {
+                    subItems.push({
+                        text: `${item.text}.${i}`,
+                        items: this.handleLoadSubItems,
+                    });
+                }
+
+                resolve(subItems);
+            }, NETWORK_DELAY);
+        });
+    };
+
+    private handleSelect = (event: LimelMenuCustomEvent<MenuItem>) => {
+        this.lastSelectedItem = event.detail.text;
+    };
+}
--- a/src/components/menu/examples/menu-subitems.tsx
+++ b/src/components/menu/examples/menu-subitems.tsx
@@ -1,0 +1,86 @@
+import {
+    MenuItem,
+    ListSeparator,
+    LimelMenuCustomEvent,
+} from '@limetech/lime-elements';
+import { Component, State, h } from '@stencil/core';
+
+/**
+ * With Sub items
+ */
+@Component({
+    tag: 'limel-example-menu-subitems',
+    shadow: true,
+})
+export class MenuSubItemsExample {
+    private items: Array<MenuItem | ListSeparator> = [
+        {
+            text: 'Format',
+            items: [
+                {
+                    text: 'Bold',
+                    icon: 'bold',
+                },
+                {
+                    text: 'Italic',
+                    icon: 'italic',
+                },
+                {
+                    text: 'Lists',
+                    icon: 'bulleted_list',
+                    items: [
+                        {
+                            text: 'Numbered list',
+                            icon: 'numbered_list',
+                        },
+                        {
+                            text: 'Bullet list',
+                            icon: 'bulleted_list',
+                        },
+                        {
+                            text: 'Checklist',
+                            icon: 'todo_list',
+                        },
+                    ],
+                },
+            ],
+        },
+        {
+            text: 'Edit',
+            items: [
+                {
+                    text: 'Copy',
+                    icon: 'copy',
+                },
+                {
+                    text: 'Cut',
+                    icon: 'cut',
+                },
+                { separator: true },
+                {
+                    text: 'Paste',
+                    icon: 'paste',
+                },
+            ],
+        },
+    ];
+
+    @State()
+    private lastSelectedItem: string;
+
+    public render() {
+        return [
+            <limel-menu items={this.items} onSelect={this.handleSelect}>
+                <limel-button label="Menu" slot="trigger" />
+            </limel-menu>,
+            <limel-example-value
+                label="Last selected item"
+                value={this.lastSelectedItem}
+            />,
+        ];
+    }
+
+    private handleSelect = (event: LimelMenuCustomEvent<MenuItem>) => {
+        this.lastSelectedItem = event.detail.text;
+    };
+}
--- a/src/components/menu/menu.tsx
+++ b/src/components/menu/menu.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable sonarjs/no-duplicate-string */
 import {
     Component,
     Event,
@@ -6,15 +7,31 @@ import {
     Prop,
     Element,
     Watch,
+    State,
 } from '@stencil/core';
 import { createRandomString } from '../../util/random-string';
-import { zipObject } from 'lodash-es';
+import { zipObject, isFunction } from 'lodash-es';
+import { LimelBreadcrumbsCustomEvent } from '@limetech/lime-elements';
+
+import { BreadcrumbsItem } from '../breadcrumbs/breadcrumbs.types';
+import { ListSeparator } from '../list/list-item.types';
 import {
-    ListSeparator,
-    MenuItem,
     OpenDirection,
+    MenuItem,
+    MenuLoader,
     SurfaceWidth,
-} from '../../interface';
+} from './menu.types';
+
+import {
+    ARROW_LEFT,
+    ARROW_LEFT_KEY_CODE,
+    ARROW_RIGHT,
+    ARROW_RIGHT_KEY_CODE,
+} from '../../util/keycodes';
+
+interface MenuCrumbItem extends BreadcrumbsItem {
+    menuItem: MenuItem;
+}
 
 /**
  * @slot trigger - Element to use as a trigger for the menu.
@@ -29,6 +46,9 @@ import {
  * @exampleComponent limel-example-menu-hotkeys
  * @exampleComponent limel-example-menu-secondary-text
  * @exampleComponent limel-example-menu-notification
+ * @exampleComponent limel-example-menu-subitems
+ * @exampleComponent limel-example-menu-custom-navigation
+ * @exampleComponent limel-example-menu-subitems-lazy-loading
  * @exampleComponent limel-example-menu-composite
  */
 @Component({
@@ -80,6 +100,20 @@ export class Menu {
     public gridLayout = false;
 
     /**
+     * Making it possible for a consumer to tell the component that it's waiting for something.
+     * For example loading items/root items.
+     */
+    @Prop()
+    public loading: boolean;
+
+    /**
+     * Making it possible for the consumer so explicitly tell the component
+     * what has been selected in the menu.
+     */
+    @Prop({ mutable: true })
+    public selectedMenuItem: MenuItem;
+
+    /**
      * Is emitted when the menu is cancelled.
      */
     @Event()
@@ -89,35 +123,82 @@ export class Menu {
      * Is emitted when a menu item is selected.
      */
     @Event()
-    private select: EventEmitter<MenuItem | MenuItem[]>;
+    public select: EventEmitter<MenuItem>;
+
+    /**
+     * Is emitted when a menu item with sub items is selected.
+     */
+    @Event()
+    public navigateMenu: EventEmitter<MenuItem>;
 
     @Element()
     private host: HTMLLimelMenuElement;
+
+    /**
+     * True while sub items are loading.
+     */
+    @State()
+    private loadingSubItems: boolean;
+
+    /**
+     * The current items in the menu list
+     */
+    @State()
+    private currentItems: Array<MenuItem | ListSeparator> = null;
 
     private list: HTMLLimelMenuListElement;
 
     private portalId: string;
     private triggerElement: HTMLSlotElement;
 
+    private get visibleItems(): Array<MenuItem | ListSeparator> {
+        if (Array.isArray(this.currentItems)) {
+            return this.currentItems;
+        }
+
+        return this.items;
+    }
+
     constructor() {
         this.portalId = createRandomString();
+    }
+
+    @Watch('items')
+    protected itemsWatcher() {
+        this.currentItems = null;
+
+        this.setFocus();
+    }
+
+    @Watch('currentItems')
+    protected currentItemsWatcher() {
+        this.setFocus();
     }
 
     @Watch('open')
     protected openWatcher() {
         if (!this.open) {
+            this.currentItems = null;
+
             return;
         }
 
-        const observer = new IntersectionObserver(() => {
-            observer.unobserve(this.list);
-            this.focusMenuItem();
-        });
-        observer.observe(this.list);
+        this.setFocus();
     }
+
+    private setFocus = () => {
+        setTimeout(() => {
+            const observer = new IntersectionObserver(() => {
+                observer.unobserve(this.list);
+                this.focusMenuItem();
+            });
+            observer.observe(this.list);
+        }, 0);
+    };
 
     public render() {
         const cssProperties = this.getCssProperties();
+
         const dropdownZIndex = getComputedStyle(this.host).getPropertyValue(
             '--dropdown-z-index'
         );
@@ -150,17 +231,9 @@ export class Menu {
                             'has-grid-layout': this.gridLayout,
                         }}
                     >
-                        <limel-menu-list
-                            class={{
-                                'has-grid-layout has-interactive-items':
-                                    this.gridLayout,
-                            }}
-                            items={this.items}
-                            type="menu"
-                            badgeIcons={this.badgeIcons}
-                            onSelect={this.handleSelect}
-                            ref={this.setListElement}
-                        />
+                        {this.renderBreadcrumb()}
+                        {this.renderLoader()}
+                        {this.renderMenuList()}
                     </limel-menu-surface>
                 </limel-portal>
             </div>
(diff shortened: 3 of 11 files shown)


src/components/menu-list/menu-list.tsx
adrianschmidt, line 1: The `limel-menu-list` component is a private sub-component of `limel-menu`, so referencing `menu-list` in the changelog (which is what you do when you add a commit called `feat(menu-list): …`) won't really make sense to the reader.

However, it seems to me like the content of this commit should be part of the preceding commit (`feat(menu): support sub items`). I assume that the menu doesn't really support sub items if the menu-list doesn't support _rendering_ those sub items?

src/components/menu-surface/menu-surface.scss
```diff
@@ -2,6 +2,11 @@
 @use '@material/elevation';
 @use '@material/menu';
 
+/**
+* @prop --limel-menu-surface-display: defines whether the surface is treated as a block, flex or grid. It affects layout used to display its children. Defaults to `block`.
+* @prop --limel-menu-surface-flex-direction: defines the direction of menu-surface layout.
+*/
```
adrianschmidt, line 8: Just in case someone else thinks, as I did, "isn't this a new feature?": no, because it's only documented on this private component, consumers won't know about them, and also, if we make sure we always define a value for these somewhere inside the limel-menu component, any value the consumer sets won't actually get here.

src/components/menu/examples/menu-size.tsx
```diff
@@ -0,0 +1,57 @@
+import { MenuItem, ListSeparator } from '@limetech/lime-elements';
+import { Component, Element, Host, State, h } from '@stencil/core';
+
+/**
+ * Resize the menu drop-down
```
adrianschmidt, line 5: I suggest calling this example "Set custom width of drop-down" or something along those lines. "Resize" makes me think the user can resize it or something.

src/components/menu/menu.tsx
```diff
@@ -198,13 +206,15 @@ export class Menu {
             '--list-grid-gap',
             '--notification-badge-background-color',
             '--notification-badge-text-color',
-        ];
+        ] as const;
```
adrianschmidt, line 663: # 😍

I didn't know about this, so had to look it up. This is awesome! I'll be using this a _ton_ from now on!

Docs for anyone else wondering: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions

src/components/menu/menu.tsx
```diff
@@ -198,13 +206,15 @@ export class Menu {
...
         const style = getComputedStyle(this.host);
         const values = propertyNames.map((property) => {
             return style.getPropertyValue(property);
         });
 
-        return zipObject(propertyNames, values);
+        type PropName = (typeof propertyNames)[number];
```
adrianschmidt, line 669: Oooh, another thing I didn't know about! _Nice!_

src/components/menu/examples/menu-size.tsx
adrianschmidt: The commit message could also be improved I think. From `feat(menu): support custom width of content` to something like `feat(menu): support setting custom width of drop-down`, perhaps? 🤷‍♂️

src/components/menu-surface/menu-surface.scss
```diff
@@ -11,6 +16,8 @@
 @include menu.core-styles;
 
 .mdc-menu-surface {
+    display: var(--limel-menu-surface-display, block);
+    flex-direction: var(--limel-menu-surface-flex-direction, row);
```
adrianschmidt, line 20: Hmm… I'm not sure I'm thinking correctly about this, but doesn't `flex-direction: row` mean that items get placed in a row, after each other, rather than in a list from top to bottom? 🤔

`row` doesn't really make sense as a default in my head…

The commit message says this just adds new css vars, but the `display` and `flex-direction` attributes weren't set on this element before. Was the element somehow already defaulting to `display: block` and `flex-direction: row`?

src/components/menu/examples/menu-size.tsx
```diff
@@ -13,9 +13,70 @@ export class MenuSubItemsExample {
...
+            subItems: [
+                { text: 'Superman' },
+                { text: 'Iron Man' },
+                { text: 'Spider-Man' },
+                { text: 'The Flash' },
+                { text: 'Wonder Woman' },
+                { text: 'Hulk' },
+                { text: 'Captain America' },
+                { text: 'Batman' },
+                { text: 'Thor' },
```
adrianschmidt, line 25: Thor is a _god_ not some measly super hero… bloody Marvel… tsk tsk…

src/components/menu/examples/menu-subitems-lazy-loading.tsx
```diff
@@ -0,0 +1,74 @@
...
+            setTimeout(() => {
+                const subItems = [];
+                for (let i = 1; i < 6; i++) {
+                    subItems.push({
+                        text: `${item.text}.${i}`,
+                    });
+                }
+
+                resolve(subItems);
+            }, 1000);
```
adrianschmidt, line 67: We've had some trouble with people blindly copy-pasting example code into production, even though you would think it's been made pretty obvious:

image

So please try to make it a little harder to miss that this timeout is there for _demo purposes only_. 😅

(See the _With no suggestions and a message for empty search results_ example for `limel-picker`, but also feel free to come up with something better than what we have there.)

src/components/menu-list/menu-list.tsx
```diff
@@ -59,6 +59,9 @@ export class MenuList {
     @Prop()
     public type: MenuListType;
 
+    @Prop()
```
adrianschmidt, line 62:
```suggestion
    @Prop({ reflect: true })
```

src/components/menu/menu.tsx
```diff
@@ -70,6 +101,29 @@ export class Menu {
...
+     * See the docs for the type `Searcher` for type information on
+     * the searcher function itself.
+     */
+    @Prop()
+    public searcher: MenuSearcher;
+
+    @Prop()
+    public loadSubItems: SubItemsLoader;
+
+    @Prop()
```
adrianschmidt, line 118:
```suggestion
    @Prop({ reflect: true })
```

src/components/menu/menu.tsx
```diff
@@ -72,6 +96,15 @@ export class Menu {
     @Prop({ reflect: true })
     public gridLayout = false;
 
+    @Prop()
+    public loadSubItems: SubItemsLoader;
```
adrianschmidt, line 116: I get that _internally_ in the component, `loadSubItems` is a nice name, because that's what happens when you call `this.loadSubItems()` (well, _hopefully_ it is, you can never trust them _consumers_ 😝). But for the consumer it's a bit of a weird name for the prop, as it's where you're supposed to provide a callback function.

In `limel-picker`, we've called the property where you're supposed to provide a searcher function, `searcher`, along with a (hopefully) clear description in the docs.

So this property could be called `subItemsLoader` for example, and we'll just have to accept that it looks less neat when we call it inside this component.

src/components/menu/menu.tsx
```diff
@@ -72,6 +96,15 @@ export class Menu {
...
+
+    @Prop()
+    public lazyLoadItems: boolean;
+
+    @Prop()
+    public loading: boolean;
```
adrianschmidt, line 122: There's quite lot of new functionality being introduced all at once here. Are these all necessary for the sub-menu functionality, or are they separate features that just happen to become want-to-have's when we add the sub-menu feature? Could we split this PR down a bit?

I noticed that on the `Files` tab, the diff for `menu.tsx` doesn't even render unless you click it, because it's so big. That's a little frightening… 😅

src/components/menu/menu.tsx
```diff
@@ -80,34 +134,110 @@ export class Menu {
      * Is emitted when a menu item is selected.
      */
     @Event()
-    private select: EventEmitter<MenuItem | MenuItem[]>;
+    public select: EventEmitter<MenuItem | MenuItem[]>;
```
adrianschmidt, line 137: I know @jgroth tends to mark these as public, but I don't know why (Johan, want to chime in here?), as the docs include the events even if the event _emitter_ is private, and I don't understand why anyone from outside the component should be allowed to call the event emitter.

Plus, if this is marked `public`, we get no warning if it's unused inside the component. I think it's quite handy to know if it's unused, as we could then remove it. No outside code should call it anyway, right? 🤷‍♂️

src/components/menu/menu.tsx
```diff
@@ -80,34 +134,110 @@ export class Menu {
...
+
+    @Event()
+    public parentSelect: EventEmitter<MenuItem>;
```
adrianschmidt, line 140: I know I'm reviewing stuff that's not been put up for review yet, sorry about that, but I wanted to understand this PR before trying to review #2482. Anyway, all these new props and events really need to get some descriptions, so one can understand what they are for 😊

src/components/menu/menu.tsx
jgroth: No one will ever be able to use any of the public properties on the component class, not even the `@Prop`s, since Stencil will just use the class we are implementing and do some magic with it and put it inside the actual web component. My guess is that it will work no matter if we use `public`, `private` or `protected`, since it is just synthetic sugar for TypeScript anyway. If we start making them private for real, using the `#property` syntax, it will probably start to fail.

I like to make them `public` just because they are part of the components public API :smile: In the Stencil docs they don't specify anything, which makes them implicitly `public`

src/components/menu/menu.types.ts
```diff
@@ -57,4 +57,28 @@ export interface MenuItem<T = any> {
      * Value of the menu item.
      */
     value?: T;
+
+    /**
+     * What sub items the item contains
+     * Don't set if using lazy loading.
+     */
+    subItems?: Array<MenuItem<T>>;
```
jgroth, line 65: How about just naming this `items`?

src/components/menu/menu.tsx
jgroth: Should this be a public prop? :thinking: I'm not quite sure how it should be used since there are no docs for it, but maybe it should be a private state of the component instead? :thinking:

src/components/menu/menu.tsx
```diff
@@ -70,6 +101,29 @@ export class Menu {
...
+     * the searcher function itself.
+     */
+    @Prop()
+    public searcher: MenuSearcher;
+
+    @Prop()
+    public loadSubItems: SubItemsLoader;
+
+    @Prop()
+    public lazyLoadItems: boolean;
```
jgroth, line 119: Not sure this is needed :smile:

src/components/menu/menu.types.ts
jgroth, line 65: How about we make this something like this
```suggestion
    items?: Array<MenuItem<T>> | () => Promise<Array<MenuItem<T>>>;
```

This way, the consumer can either just send in a list of items for the nested menu if they are available, or give us a function to call that will load the items asynchronously when needed. This will allow us to get rid of both the `loadSubItems`, `lazyLoadItems` and `loading` props. The function will probably need some kind of input argument, but I don't know what that is yet :smile: Perhaps the parent item? :thinking:

src/components/menu/menu.types.ts
```diff
@@ -57,4 +57,28 @@ export interface MenuItem<T = any> {
...
+
+    /**
+     * What parent the item has.
+     * It's used to render the breadcrumbs history
+     * Mostly handled by the menu itself
+     */
+    parentItem?: MenuItem<T>;
```
jgroth, line 72: Not sure I understand why this is needed. If a consumer does not need it we should put it in a private type instead that is not exported

src/components/menu/menu.types.ts
```diff
@@ -57,4 +57,28 @@ export interface MenuItem<T = any> {
...
+
+    /**
+     * Only used if using lazyLoading
+     * It prevents the menu from rendering the carret icon
+     * that visualize sub items
+     */
+    isLeafNode?: boolean;
```
jgroth, line 79: Is this needed? :thinking: Can we just not set the `items`/`subItems` prop to `null`/`undefined`/`[]` to indicate it is a leaf? :thinking:

jgroth (commented):
This is not a full review :smile: Just had a look at the public API and have some questions and suggestions :slightly_smiling_face:

src/components/menu-surface/menu-surface.scss
TommyLindh2: The `.mdc-menu-surface` is a div and it's default display is block, so this wouldn't change in this case.

In this component `column` maybe makes more sense to use as default, but we used `row` as default because that is the default value when you use `display: flex`

src/components/menu/menu.tsx
TommyLindh2: We want the end user to be able to show a loader in the menu surface if they for example loads the root items dynamically.
Our hierarchy picker (why we did this PR in the beginning) for example loads the root items from limeobjects that is async, and instead of just an empty menu we would like to support the loader in the menu.

Does it make sense?

src/components/menu/menu.tsx
TommyLindh2: Ì agree with @jgroth to have them public.

I like your angle with the unused stuff @adrianschmidt but in some cases you must define the event because a child you render emits an event that you just pass up through your component, and in that case you define it and won't "use" it. And you will get a warning that is incorrect.

Do we agree on having them public?

src/components/menu/menu.types.ts
TommyLindh2: Yeah we weren't 100% happy with this solution, but the reason it's public is because that is the property that is used to build the "history" using the breadcrumb component.

So again in our hierarchy picker it's possible to open the menu that should "start" in a sub menu (not in the root but maybe in level 2 or 3) and this is then how you can visualize that from the outside.

We're very open for suggestion/discussions about this solution.

src/components/menu/menu.types.ts
jgroth: Maybe to root `items` prop also should have this type to allow lazy loading of the first level of items, not sure if it will make the code easier or harder to implement :smile:

src/components/menu-list/menu-list.tsx
BjornOstberg: fixed now

src/components/menu/examples/menu-size.tsx
BjornOstberg: fixed

src/components/menu/examples/menu-subitems-lazy-loading.tsx
BjornOstberg: fixed

src/components/menu/menu.tsx
BjornOstberg: prop removed

src/components/menu/menu.types.ts
BjornOstberg: fixed

src/components/menu/menu.tsx
BjornOstberg: removed props

src/components/menu/menu.types.ts
BjornOstberg: Implemented this solution and removed unnecessary props

src/components/menu/menu.types.ts
BjornOstberg: removed when implementing being able to send in a function to the item prop in the menuItem Interface

src/components/menu/menu.tsx
BjornOstberg: Hopefully fixed :)

src/components/menu/menu.tsx
```diff
@@ -77,34 +139,122 @@ export class Menu {
...
      */
     @Event()
-    private select: EventEmitter<MenuItem | MenuItem[]>;
+    public select: EventEmitter<MenuItem>;
+
+    /**
+     * Is emitted when a menu item with sub items is selected.
+     */
+    @Event()
+    public parentSelect: EventEmitter<MenuItem>;
```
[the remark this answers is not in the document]
BjornOstberg: Wont that make it a breaking change? And also, we feel that for readability a separate event is nicer.

Let's discuss if you don't agree.

src/components/menu/menu.tsx
BjornOstberg: It would be possible to remove the loadingSubItems, but then the consumer have to control when the component is in a loading state form outside.

src/components/menu/menu.types.ts
BjornOstberg: Maybe we should have a discussion regarding this as well.

src/components/menu/menu.types.ts
```diff
@@ -57,4 +59,23 @@ export interface MenuItem<T = any> {
...
+     * Mostl y handled by the menu itself
+     */
+    parentItem?: MenuItem<T>;
 }
+
+export type MenuSearcher = (query: string) => Promise<MenuItem[]>;
+
+export type SubItemsLoader = (
+    item: MenuItem
+) => Promise<Array<MenuItem | ListSeparator>>;
```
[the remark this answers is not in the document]
BjornOstberg: We renamed them according to your suggestion

src/components/menu/menu.tsx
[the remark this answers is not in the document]
BjornOstberg: Renamed this to navigateMenu to better fit what it actually does

src/components/menu/menu.tsx
```diff
@@ -38,7 +76,7 @@ export class Menu {
     public items: Array<MenuItem | ListSeparator> = [];
 
     /**
-     * Sets the disabled state of the menu.
+     * Sets the disabled st ate of the menu.
```
jgroth, line 79:
```suggestion
     * Sets the disabled state of the menu.
```

src/components/menu/menu.tsx
```diff
@@ -50,7 +88,7 @@ export class Menu {
     public openDirection: OpenDirection = 'bottom-start';
 
     /**
-     * Sets the open state of the menu.
+     * Sets the open state of the  menu.
```
jgroth, line 91:
```suggestion
     * Sets the open state of the menu.
```

src/components/menu/menu.types.ts
```diff
@@ -57,4 +59,23 @@ export interface MenuItem<T = any> {
...
+    /**
+     * What sub items the item contains
+     * Don't set if using lazy loading.
+     */
+    items?: Array<MenuItem<T> | ListSeparator> | MenuLoader;
+
+    /**
+     * What parent the item has.
+     * It's used to render the breadcrumbs history
+     * Mostl y handled by the menu itself
```
jgroth, line 72:
```suggestion
     * Mostly handled by the menu itself
```

src/components/menu/menu.types.ts
jgroth: Since there is still no example of how this is supposed to be used, can we perhaps mark it as `@private` so a consumer does not use it by mistake?

src/components/menu/menu.tsx
```diff
@@ -67,6 +105,37 @@ export class Menu {
     @Prop({ reflect: true })
     public gridLayout = false;
 
+    /**
+     * A search function that takes a search-string as an argument,
+     * and returns a promise that will eventually be resolved with
+     * an array of `MenuItem`:s.
+     *
+     * See the docs for the type `Searcher` for type information on
```
jgroth, line 113: Not sure if this works or not, otherwise it should just be `MenuSearcher` 
```suggestion
     * See the docs for the type {@link MenuSearcher} for type information on
```

src/components/menu/menu.tsx
```diff
@@ -67,6 +105,37 @@ export class Menu {
...
+     */
+    @Prop()
+    public emptyResultMessage?: string;
+
+    /**
+     * Making it possible for the consumer so explicitly tell the component
+     * what has been clicked in the menu.
+     */
+    @Prop({ mutable: true })
+    public clickedMenuItem: MenuItem;
```
jgroth, line 137: I don't think we should use words such as `click` in the props, since you can navigate the menu with the keyboard as well. Perhaps `selectedMenuItem`?

src/components/menu/menu.tsx
```diff
@@ -67,6 +105,37 @@ export class Menu {
...
+     */
+    @Prop()
+    public searcher: MenuSearcher;
+
+    /**
+     * Making it possible for a consumer to tell the component that it's waiting for something.
+     * For example loading items/root items.
+     */
+    @Prop()
+    public loading: boolean;
```
jgroth, line 124: What happened with this? Did you try and change the `item` prop to `Array<MenuItem<T> | ListSeparator> | MenuLoader` to get rid of this prop?

jgroth (commented):
Not sure why there are spaces in some random places :smile:

src/components/menu/menu.tsx
BjornOstberg: yes, we tried it out and it got really complex really fast for the consumer. And also pretty complex in the component itself :/

src/components/menu/menu.tsx
BjornOstberg: fixed!

src/components/menu/menu.tsx
BjornOstberg: fixed!

src/components/menu/menu.types.ts
BjornOstberg: fixed!

src/components/menu/menu.tsx
BjornOstberg: It works, and is fixed now.

src/components/menu/menu.tsx
BjornOstberg: Good point, fixed this as well.

src/components/menu/menu.types.ts
BjornOstberg: We added an example now. I must've totally forgotten about it after our meeting.

src/components/menu/menu.tsx
BjornOstberg: We tried it and it turned really complex here, but mostly for the consumer.

src/components/menu/examples/menu-subitems.tsx
```diff
@@ -0,0 +1,86 @@
...
+ * With Sub items
+ */
+@Component({
+    tag: 'limel-example-menu-subitems',
+    shadow: true,
+})
+export class MenuSubItemsExample {
+    private items: Array<MenuItem | ListSeparator> = [
+        {
+            text: 'Format',
```
Kiarokh, line 18: image
There is an empty icon in this 🤔

Kiarokh: same in `menu-custom-navigation.tsx`

Kiarokh: @BjornOstberg 's fix here: https://github.com/Lundalogik/lime-elements/pull/2550

src/components/menu-surface/menu-surface.scss
Kiarokh, line 20: something is wrong here 
image

src/components/menu/menu.tsx
```diff
@@ -150,6 +308,301 @@ export class Menu {
...
+        }
+
+        if (!breadCrumbItems.length) {
+            return;
+        }
+
+        return (
+            <limel-breadcrumbs
+                style={{
+                    'border-bottom': 'solid 1px rgb(var(--contrast-500))',
```
Kiarokh, line 359:
```suggestion
                    'border-bottom': 'solid 1px rgb(var(--contrast-500))',
                    'flex-shrink': '0',
```

Kiarokh: fixes this problems
image

BjornOstberg: Nice find, fixed now.

Kiarokh:
There seems to be a few comments from Johan that are not answered yet.

After fixing them, I’d suggest that you squash all the fixup commits, and then split the PR into to few meaningful PRs that can be reviewed and merged separately. This makes the work less scary for the reviewers.

For example, you can have one PR for enabling search, with a couple of examples.
Next PR could be for enabling lazy loading + examples.
Another PR can be about adding nested menus.

Which one you put up first for review is up to you. You know better how things are connected. Just keep in mind that the repo should be able to run or be deployed free of bugs or problems, after each _commit_ that you add.

TommyLindh2:
This PR have been through a lot and it looks like it's huge when looking at the comments.
That was a review from the public API/interfaces of the branch, So I will close this and create a separate one for the actual code.

Issue #1468, opened 2021-11-24
limel-list: keyboard navigation is broken for lists with checkboxes or radio buttons (affects typeahead too)


adrianschmidt:
## Current behavior

https://user-images.githubusercontent.com/154725/143221299-48ba2760-adae-4270-87d9-d54f1bf43fba.mov

1. Go to the [docs for limel-list](https://lundalogik.github.io/lime-elements/versions/next/#/component/limel-list/) and scroll to the example called "List with checkboxes".
2. Focus the list in some way (clicking the disabled option "Charmander" or clicking on an item in the preceding list and then tabbing to the checkbox list both work).
3. You can now move the focus between all list items using the up and down arrow keys.
4. Move to "Pikachu".
5. Press `space` or `enter` to deselect "Pikachu".
  - The list loses focus, which is the first bug here.
6. Select "Pikachu" again, either by focusing the list and then using the keyboard, or simply by clicking on the checkbox, it doesn't matter.
7. Try to navigate to the bottom of the list using the arrow keys.

- You can't move past "Yoshi". The two last items in the list can no longer be focused. Other combinations of selecting and deselecting items result in other items becoming unreachable.

## Expected behavior

The list should not lose focus when items are deselected, and all list items should always be reachable by keyboard.

## Environment

- **lime-elements** version: At least v34.0.2-next.10 and later, but I think the issue is older.
- Framework used: Can be reproduced in the docs.
- Logs:

adrianschmidt:
Check if this is related to using separators and/or disabled items.

Kiarokh:
This seems to be because the of how `tabindex` dynamically gets applied list items. 

I can't find a solid pattern to describe exactly what happens. But as you tab throw the list items, the tab index number gets changed for all items. It seems the item that is having the keyboard focus gets `tabindex="0"`, and others get `"-1"`. Until you select an item and press enter (to close the menu or dropdown). Then when you press enter again to open it, some items lose the `tabindex`. Thus they become totally unavailable for keyboard navigation.

I actually don't think it has anything to do with having `disabled` items or dividers. I can reproduce it in `limel-select`, both with and without disabled items too.

https://user-images.githubusercontent.com/35954987/149151176-172ebdb8-0209-4bcc-b558-a1f2591e37a2.mov


[...]

PR #2482, opened 2023-09-15
Fix/cascading menu list keyboard navigation

TommyLindh2:
The keyboard navigation doesn't work when we change the items in the menu, in the cascading menu PR.
This because the MDCMenu.items isn't in sync with what is actually rendered from the `this.items`.

We couldn't find a way to "force" a refresh, so we tried re-creating it which seems to fix our issues.

Here you can see how it doesn't work in [PR#2473](https://github.com/Lundalogik/lime-elements/pull/2473), and how it does work in [This PR#2482](https://github.com/Lundalogik/lime-elements/pull/2482)

https://github.com/Lundalogik/lime-elements/assets/1814702/b91b08ed-3de9-47ce-836a-f4edccec2931

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu-list/menu-list.tsx
+++ b/src/components/menu-list/menu-list.tsx
@@ -4,10 +4,8 @@ import {
     MenuItem,
     MenuListType,
 } from '../../interface';
-import { MDCList, MDCListActionEvent } from '@material/list';
 import { MDCMenu, MDCMenuItemEvent } from '@material/menu';
 import { MDCRipple } from '@material/ripple';
-import { strings as listStrings } from '@material/list/constants';
 import { strings as menuStrings } from '@material/menu/constants';
 import {
     Component,
@@ -21,7 +19,6 @@ import {
 import { MenuListRenderer } from './menu-list-renderer';
 import { MenuListRendererConfig } from './menu-list-renderer-config';
 
-const { ACTION_EVENT } = listStrings;
 const { SELECTED_EVENT } = menuStrings;
 
 /**
@@ -73,7 +70,6 @@ export class MenuList {
 
     private config: MenuListRendererConfig;
     private MenuListRenderer = new MenuListRenderer();
-    private mdcList: MDCList;
     private mdcMenu: MDCMenu;
 
     /**
@@ -112,17 +108,7 @@ export class MenuList {
     }
 
     @Watch('items')
-    protected itemsChanged() {
-        if (!this.mdcList) {
-            return;
-        }
-
-        const MenuItems = this.items.filter(this.isMenuItem);
-
-        this.mdcList.selectedIndex = MenuItems.findIndex(
-            (item: MenuItem) => item.selected
-        );
-    }
+    protected itemsChanged() {}
 
     private setup = () => {
         this.setupMenu();
@@ -151,15 +137,8 @@ export class MenuList {
     };
 
     private teardown = () => {
-        this.mdcList?.unlisten(ACTION_EVENT, this.handleAction);
-        this.mdcList?.destroy();
-
         this.mdcMenu?.unlisten(SELECTED_EVENT, this.handleMenuSelect);
         this.mdcMenu?.destroy();
-    };
-
-    private handleAction = (event: MDCListActionEvent) => {
-        this.handleSingleSelect(event.detail.index);
     };
 
     private handleMenuSelect = (event: MDCMenuItemEvent) => {
--- a/src/components/menu-list/menu-list.tsx
+++ b/src/components/menu-list/menu-list.tsx
@@ -108,7 +108,11 @@ export class MenuList {
     }
 
     @Watch('items')
-    protected itemsChanged() {}
+    protected itemsChanged() {
+        setTimeout(() => {
+            this.setup();
+        }, 0);
+    }
 
     private setup = () => {
         this.setupMenu();
@@ -116,6 +120,11 @@ export class MenuList {
     };
 
     private setupMenu = () => {
+        if (this.mdcMenu) {
+            this.teardown();
+            this.mdcMenu = null;
+        }
+
         const element = this.element.shadowRoot.querySelector('.mdc-menu');
         if (!element) {
             return;


src/components/menu-list/menu-list.tsx
adrianschmidt, line 1: There doesn't seem to be an example of cascading menus in the docs yet? That would be quite handy for testing this… 😉

If the cascading menus feature is hidden because it's not released yet, could you add a temporary example component in this PR, so I can test the fix? Keep it as a separate commit please, since I would need to cherry-pick it to a different branch where I can reproduce the original bug 😊

TommyLindh2: I've changed the base of the PR so it now goes to the correct branch and also rebased it.
So this branch should include an example called `With sub items` in the menu component where you can see that it works.

And in this [PR#2473](https://github.com/Lundalogik/lime-elements/pull/2473) you can see that it doesn't work.

TommyLindh2: I've also update the description with a video

Kiarokh: Wonderful! this must also fix this old issue. https://github.com/Lundalogik/lime-elements/issues/1468#issuecomment-1719311883

adrianschmidt: > Wonderful! this must also fix this old issue. [#1468 (comment)](https://github.com/Lundalogik/lime-elements/issues/1468#issuecomment-1719311883)

It's possible that making the same changes in `limel-list` would solve that problem too, yes. But the code involved are in different files, so this doesn't automatically solve that issue. (For anyone following that link, it's very easy to get the impression that that issue is indeed about the menu, because it's been very heavily side-tracked, but the issue itself is actually about limel-list.)

src/components/menu-list/menu-list.tsx
```diff
@@ -116,17 +112,7 @@ export class MenuList {
...
-            return;
-        }
-
-        const MenuItems = this.items.filter(this.isMenuItem);
-
-        this.mdcList.selectedIndex = MenuItems.findIndex(
-            (item: MenuItem) => item.selected
-        );
-    }
+    protected itemsChanged() {}
```
adrianschmidt: Why do we need an empty watcher that does nothing?

adrianschmidt: Ah, I saw the change in the next commit now.

src/components/menu-list/menu-list.tsx
Kiarokh: Ah I just took it for granted that @TommyLindh2 did the same change in both, since we talked about it before the PR. Maybe good to do a commit for that as well, in this PR Tommy?

TommyLindh2: When I tried it out it didn't solve the issue in the limel-list if I'm not mistaken.
The issue with the limel-list is the tab-index stuff. I think that this is something else.

Kiarokh: Tnx. Ok, then that is most probably in the `menu-list` too. But we take that issue elsewhere then.

adrianschmidt: Can you have a menu-list with checkboxes or radio-buttons? 🤔

Kiarokh: I don't think that was only for lists with Checkboxes or Radio buttons. It was in normal lists too. 

Here is an example that looks similar to what Tommy and Björn have fixed in the menu, recorded in this PR.

https://github.com/Lundalogik/lime-elements/assets/35954987/fa1a6d0d-cc6b-4fdf-8769-c327c2141b01

There is a `limel-list` inside a `limel-menu-surface` ☝️ 

- The first dropdown list has 3 items.
- The second dropdown list has 4 items, but the user cannot go past the 3rd item in the list.

Is it the same issue?

adrianschmidt: Ah! Yes, that certainly looks like it's the same issue 👍

TommyLindh2: Yes, we can probably solve that issue the same way, you are correct @Kiarokh but there was other issues with the limel-list as well I think.

TommyLindh2: @adrianschmidt do you think that fix in limel-list should be in this PR or a separate PR?

Kiarokh: If the exact same fix can be applied there, I'd suggest to do it in this PR. Or just make a new one immediately after or request the same reviewer, and reference to this PR.
The point is that there is only a few of us (already in this PR) who know what this is about. So it's easier to get a review  when everyone is onboard already. 😊 

If you don't do it now, it'll never get done basically. We tend to not make descriptive issues, and everything relies on me or Adrian to remember that we had some problem somewhere about something, and then make an effort to do it.

Believe me if I knew how to do these things, I'd do them myself and would spam people with PRs for review. I just don't know how to fix them 😓 and now that you're on it, I don't want to simply ignore it and procrastinate it. Because I know nobody will care to fix or prioritize these later. There is no later!

TommyLindh2: @adrianschmidt @Kiarokh I will create a separate PR for the keyboard navigation in the limel-list.
So can any of you approve this so we can merge it?

Kiarokh: sure. just resolve the conflicts 😅

TommyLindh2: Yeah, Björn did some rebasing right after I wrote xD I will have to do that tomorrow though. Get back to you then

Kiarokh: Do it ✊ Do it ✊ Do it ✊ 😄

adrianschmidt: Sorry, I've been out of action for a week 🤒

Sounds absolutely fantastic that you can fix this in both places. I agree with Kia that it's best to get both fixed at once. If they are in the same PR or not is less important. I can see pros and cons with either, so just pick one 😄

TommyLindh2: @adrianschmidt now this is ready for review again

Kiarokh:
Since this was already approved, I rebased and merged it

Issue #1998, opened 2022-12-21
The scroll indicator shadow in limel-dialog can be obscured by elements inside the scrolling area


adrianschmidt:
## Current behavior

The scroll indicator shadow in limel-dialog can be obscured by elements inside the scrolling area.

https://user-images.githubusercontent.com/154725/208860351-f7d68996-1730-4e3f-b0f3-97bb5c10331c.mov

<img width="820" alt="A screenshot of the 'Custom size' example in the docs for limel-dialog, where an input field has been added to show that it is rendered above the scroll indicator shadow." src="https://user-images.githubusercontent.com/154725/208864723-566ced4c-b954-4b0f-bc0f-529a0f9e71a4.png">

1. Open the dialog in the _Custom size_ example in [the limel-dialog docs](https://lundalogik.github.io/lime-elements/versions/NEXT-36.2.0-next.4/#/component/limel-dialog/).
2. Inspect the content, and edit one of the paragraphs to add a `<input type"text">` to it.
3. Scroll the content so that the input intersects with either the top or the bottom scroll indicator shadow.

## Expected behavior

The shadow should be rendered on top of any element inside the scrolling area. Note that the shadow should not interact with clicks in any way, so we should probably set `pointer-events: none;` on it.

Note: The scrolling indicator shadow is a bit of a special hack using the css `background` property ([implemented here](https://github.com/Lundalogik/lime-elements/blob/v36.2.0-next.4/src/components/dialog/dialog.scss#L81)). As is, it's probably not possible to "lift" the shadows on top of the content. But the shadows can probably be put in `::before` and `::after` pseudo-elements instead, and those should be possible to lift above the content.

## Environment

- **lime-elements** version: 36.2.0-next.4
- Framework used: Can be reproduced in the docs

adrianschmidt:
Here's another example from @LucyChyzhova:

https://user-images.githubusercontent.com/50618208/215755026-08d1f2eb-502b-4650-94c3-3bb289265273.mov

LucyChyzhova:
Another example from Addon installer https://github.com/Lundalogik/limepkg-addon-installer/issues/191

adrianschmidt:
The current implementation is one that doesn't use pseudo-elements. I'm thinking that if we change it to use pseudo-elements, we can perhaps z-index those to go above the content of the scroll area… 🤔

adrianschmidt:
[...]

Issue #1953, opened 2022-11-15
Dialog: Add CSS props for content padding


Kiarokh:
Many times the default padding of `.mdc-dialog__content` (which is by the way is in `px`s) has been disturbing in the UI of what you are creating. Also it's unresponsive, creating some mobile layout issues.

<img width="1139" alt="Screenshot 2022-11-15 at 17 27 04" src="https://user-images.githubusercontent.com/35954987/201973405-93a88d59-8e67-4564-8009-c2dbabb9f841.png">

For instance in this app 👇 we had to add a negative margin around the `tab-panel` inside the dialog's body to compensate for the hard-coded paddings.

![MicrosoftTeams-image](https://user-images.githubusercontent.com/35954987/201973601-9d95d5eb-dc58-4f2a-9dc7-b8f3554d95c5.png)

Kiarokh:
from the PR:


https://user-images.githubusercontent.com/35954987/202479389-17a40947-56d4-499c-b10d-6ad740fcd5a0.mov

PR #2494, opened 2023-09-20
Fade out content that overflow out of the top and bottom edges of a dialog's content

Kiarokh:
fix https://github.com/Lundalogik/lime-elements/issues/1998

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dialog/dialog.scss
+++ b/src/components/dialog/dialog.scss
@@ -1,4 +1,5 @@
 @use '../../style/functions';
+@use '../../style/mixins';
 
 @use '../../style/internal/lime-theme';
 
@@ -16,7 +17,7 @@
  * @prop --dialog-max-width: Max width of the dialog.
  * @prop --dialog-max-height: Max height of the dialog.
  * @prop --dialog-border-radius: Border radius of the dialog corners
- * @prop --dialog-padding-top-bottom: Padding on top and bottom of dialog content. Defaults to `1.5rem`.
+ * @prop --dialog-padding-top-bottom: Padding on top and bottom of dialog content. Affects the height of fade-out effects on top and bottom edges when the content is scrollable and has overflowed out of the content area. Defaults to `1.5rem`. Note that if you use this variable and set it to numbers smaller than 1rem, you will loose the fade-out effects on the edges. If you have set these paddings to `0`, losing the fade out effects should be however fine for your use case! Because in such a case your intention is to handle the `overflow` internally in the component that is displayed in the dialog's content.
  * @prop --dialog-padding-left-right: Padding on the sides of dialog content. Defaults to `1.25rem`.
  */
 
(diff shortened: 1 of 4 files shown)


src/components/dialog/dialog.scss
```diff
@@ -16,7 +17,7 @@
  * @prop --dialog-max-width: Max width of the dialog.
  * @prop --dialog-max-height: Max height of the dialog.
  * @prop --dialog-border-radius: Border radius of the dialog corners
- * @prop --dialog-padding-top-bottom: Padding on top and bottom of dialog content. Defaults to `1.5rem`.
+ * @prop --dialog-padding-top-bottom: Padding on top and bottom of dialog content. Defaults to `1.5rem`. Note that if you use this variable and set it to numbers smaller than 1rem, you will loose the edge fade effects when the content is scrollable and has overflowed out of the dialog's content area. This should be however fine! Because you must then be handling the overflow internally in the component that is sent to the dialog's content.
```
Kiarokh, line 20: Maybe I should also explain how these affect the fade-out effect of the edges, generally.

LucyChyzhova: I think this is a good explanation. Maybe we could add the preferred padding ranges.

Kiarokh: Maybe. The more padding, the larger the fade-out effect will be.

LucyChyzhova (approved):

src/components/dialog/dialog.scss
```diff
@@ -16,7 +17,7 @@
  * @prop --dialog-max-width: Max width of the dialog.
  * @prop --dialog-max-height: Max height of the dialog.
  * @prop --dialog-border-radius: Border radius of the dialog corners
- * @prop --dialog-padding-top-bottom: Padding on top and bottom of dialog content. Defaults to `1.5rem`.
+ * @prop --dialog-padding-top-bottom: Padding on top and bottom of dialog content. Affects the height of fade-out effects on top and bottom edges when the content is scrollable and has overflowed out of the content area. Defaults to `1.5rem`. Note that if you use this variable and set it to numbers smaller than 1rem, you will loose the fade-out effects on the edges. If you have set these paddings to `0`, losing the fade out effects should be however fine for your use case! Because in such a case your intention is to handle the `overflow` internally in the component that is displayed in the dialog's content.
```
Kiarokh, line 20: it might be hard to know what these docs are trying to say. but this is the background story of adding those variables: https://github.com/Lundalogik/lime-elements/issues/1953 some appers user the whole dialog surface to render their "fullscreen apps", and they don't want our paddings. They just set it to zero: https://github.com/search?q=org%3ALundalogik%20--dialog-padding-top-bottom&type=code

adrianschmidt:
This version has its own problems:

https://github.com/Lundalogik/lime-elements/assets/154725/b3f8198a-5b4d-44e4-a866-ed1e243fccc8

But if we think this version is less problematic than the old one, I'm happy to merge it 👍

Kiarokh:
You're right. The perfect solution would be having a reusable script that adds some sort of gradient shadow  on the edge on each side/edge of the scroll area when the content has scrolled out (and also manipulates the gradient masks using the CSS variables that I added in the mixin). However, I can't write such a script myself. If you wanna join me for a session, I can explain what I mean.

However, as you said this is less problematic than what we had, and more globally usable. I added a new commit where we use the same mixin in the recently added breadcrumbs component, to show case the reusability of this mixin better.

LucyChyzhova (approved):

adrianschmidt:
> The perfect solution would be having a reusable script that adds some sort of gradient shadow on the edge on each side/edge of the scroll area when the content has scrolled out (and also manipulates the gradient masks using the CSS variables that I added in the mixin). However, I can't write such a script myself. If you wanna join me for a session, I can explain what I mean.

I'm think I know exactly what you mean. I've been thinking more or less the same myself. Let's do it some day! Next hackday maybe? This is an improvement in the meantime 👍

PR #2499, opened 2023-09-25
fix(picker): be able to clear all items in the picker when clicking 'clear all' button

Befkadu1:
**EDIT:** _Apparently the [original bug](https://github.com/Lundalogik/lime-crm-components/issues/1612) (which this revert would reintroduce) was worse than I had understood. So we shouldn't merge this without also having a fix or workaround for that bug. /Ads_

When we have multiple items in the picker and clicking the Clear all button in the picker not clearing all the items at all. This bug causes freezing of the whole component.

fix Lundalogik/crm-feature#3559

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -693,10 +693,6 @@ export class Picker {
     }
 
     private handleCloseMenu() {
-        if (this.items.length > 0) {
-            return;
-        }
-
         this.clearInputField();
     }
 


adrianschmidt:
Closing this, since Kia will be making a PR for simply hiding the clear-button while the drop-down is visible. That feels very much like the kind of patch-on-patch-on-patch solution I wanted to avoid when I said to revert the bugfix that caused this issue, but when I said that I hadn't realised that the previous bug happened whenever you started typing. I thought it was just when you clicked in the input field while the picker was already open, which seemed like an edge case.

Sorry for making you jump through hoops @Befkadu1! 🙇‍♂️

PR #2502, opened 2023-09-26
Icon button: remove dependency of MDC

Kiarokh:
This component is very minimal and simple, and we are completely overwriting MDC styles with our own styles. 
We don't use MDC ripple effect, and nothing else from the MDC. 
There is no need for us to be dependent on MDC here anymore.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


adrianschmidt (approved):
Looks good to me 👍

adrianschmidt (approved):

PR #2506, opened 2023-09-26
feat(table): add variable for `padding-bottom`

Kiarokh:
In a certain use cases the scrollable area
of a table may need an extra `padding-bottom`.
For example, when another element with a `fixed` or `absolute` `position` is overlayed on the table.

fix https://github.com/Lundalogik/crm-feature/issues/3636

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.scss
+++ b/src/components/table/table.scss
@@ -5,6 +5,7 @@
 
 /*
 * @prop --table-max-column-width: defines a maximum width for columns using standard size units, to prevent the table from growing too wide. Set to `auto` if you do not need this limitation. Defaults to `40rem`.
+* @prop --table-scrollable-area-padding-bottom: Allows adding extra space at the bottom of the scrollable area of the table. This space will be added between the last row and the aggregations row (if present), or the bottom edge of the table. Defaults to `0`.
 */
 
 $width-of-sorter-arrow: 0.5rem;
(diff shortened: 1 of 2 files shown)


src/components/table/table.scss
```diff
@@ -50,6 +51,10 @@ $table--limel-table--row-selector: 1;
         color: var(--table-text-color);
     }
     .tabulator-tableHolder {
+        padding-bottom: calc(
```
Kiarokh, line 54: https://github.com/Lundalogik/lime-crm-components/assets/35954987/81ba853b-a136-4d30-8977-1f243e648c48

We get this annoying bug where tabulator scrolls to top, every time one of the rows is selected :( 
I've tried to use this as a `margin-bottom`, on the nested container, or as a `margin-bottom` on the last row, as well as other tricks. But nothing works well.

So we should just ignore this, and close the issue. I consider this too hard and not worth fixing.

Kiarokh:
closing due to unwanted buggy behaviors that it generates.

PR #2508, opened 2023-09-27
Code diff component

john-traas:
Create new component to show code differences in a way similar to the way github shows differences. 

Fixes https://github.com/Lundalogik/hack-tuesday/issues/321


john-traas:
New component implemented here: https://github.com/Lundalogik/lime-elements/pull/3949

Issue #2512, opened 2023-09-28
List: add support for having separators with titles.


Kiarokh:
## New feature motivation
Separators are a simple yet powerful design elements that can be employed in lists of items. They offer significant usability advantages by providing valuable visual cues that aid users in perceiving and navigating through lists.
 
- **Grouping and Hierarchy:**
Separators can be used to group related items, signaling to users that those items share a common attribute or purpose. This grouping effect aids in creating a hierarchical structure within the list, making it simpler for users to grasp relationships and make informed decisions.
- **Visual Scannability:**
When users quickly scan a list, their eyes naturally use the separator lines as visual anchors, making it easier to find items and remember their whereabouts next time they revisit the same list.
- **Reduced Cognitive Effort:**
Separators contribute to a user's overall comprehension and experience
by reducing the cognitive effort required to process the information.
 
## New feature description
You should be able to optionally add a title to the separators, to clarify further what each group of items is about,
and by doing so improve the users perception and experience.

In our sketches, we have many designs that require this feature. For example, the arrow number 5 here in this PR is pointing to that:
![image](https://github.com/Lundalogik/lime-elements/assets/35954987/d3c44241-b297-465d-bf19-5b889bf34997)

## New feature implementation
Screenshot from the PR
<img width="475" alt="image" src="https://github.com/Lundalogik/lime-elements/assets/35954987/4d229342-f4f5-478d-98c4-f601c09f5929">

PR #2513, opened 2023-09-28
List separators

Kiarokh:
fix https://github.com/Lundalogik/lime-elements/issues/2512

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/list/examples/list-separator.tsx
+++ b/src/components/list/examples/list-separator.tsx
@@ -1,0 +1,83 @@
+import { ListItem, ListSeparator } from '@limetech/lime-elements';
+import { Component, h } from '@stencil/core';
+
+/**
+ * List with separators
+ *
+ * Separators are simple yet powerful design elements that can be
+ * employed in lists of items. They offer significant usability advantages
+ * by providing valuable visual cues that aid users in perceiving
+ * and navigating through lists.
+ *
+ * - **Grouping and Hierarchy:**
+ * Separators can be used to group related items, signaling to users that
+ * those items share a common attribute or purpose.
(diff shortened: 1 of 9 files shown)


src/components/list/list-item.types.ts
```diff
@@ -49,6 +49,8 @@ export interface ListItem<T = any> {
 
 export interface ListSeparator {
     separator: true;
+
+    separatorTitle?: string;
```
jgroth, line 53: It might make sense to just name this `text` as for regular `ListItem`s

jgroth: It should at least not have `separator` in the name, we already know that it's a separator so we do not need to state that again :slightly_smiling_face:

Kiarokh: ⚡

src/components/list/examples/list-separator.tsx
```diff
@@ -0,0 +1,83 @@
+import { ListItem, ListSeparator } from '@limetech/lime-elements';
+import { Component, h } from '@stencil/core';
+
+/**
+ * List with separators
+ *
+ * Separators are a simple yet powerful design elements that can be
```
john-traas, line 7:
```suggestion
 * Separators are simple yet powerful design elements that can be
```

PR #2514, opened 2023-09-28
fix(picker): don't display clear all button when the suggestions are …

Kiarokh:
…being shown

This has two reasons:
1. The button is not functional, because the dropdown should close first for the
user to be actually able to press the clear all button.
2. It causes another bug, where you cannot close the suggestion list if you have clicked on the close all button, while the list was open.

fix https://github.com/Lundalogik/crm-feature/issues/3559

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -269,7 +269,7 @@ export class Picker {
                 onStartEdit={this.handleInputFieldFocus}
                 onStopEdit={this.handleStopEditAndBlur}
                 emptyInputOnBlur={false}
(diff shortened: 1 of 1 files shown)


adrianschmidt (approved):
_Excellent!_ 👌

Issue #2510, opened 2023-09-27
Action bar: hidden overflowed items show tooltip when hovered


Kiarokh:
https://github.com/Lundalogik/lime-elements/assets/35954987/a60d966d-050e-4a45-a06e-64261c1e886e

PR #2515, opened 2023-09-29
fix(action-bar): do not display tooltip of invisible action buttons

Kiarokh:
fix https://github.com/Lundalogik/lime-elements/issues/2510

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/action-bar/action-bar-item/action-bar-item.scss
+++ b/src/components/action-bar/action-bar-item/action-bar-item.scss
@@ -9,6 +9,7 @@ limel-action-bar-item {
 
     &:not([is-visible]) {
         opacity: 0;
+        pointer-events: none;
     }
 }
 


src/components/action-bar/action-bar-item/action-bar-item.scss
```diff
@@ -9,6 +9,7 @@ limel-action-bar-item {
 
     &:not([is-visible]) {
         opacity: 0;
+        visibility: hidden;
```
Kiarokh, line 12: https://developer.mozilla.org/en-US/docs/Web/CSS/visibility

LawrenceBorst (dismissed):
This works, I'll approve but should note that we lose the fade-out effect when an action bar item becomes invisible if we change this

Kiarokh:
> This works, I'll approve but should note that we lose the fade-out effect when an action bar item becomes invisible if we change this

😮 WOt? I checked specifically for that. Are you sure? 

Ok now checked again. It seems to have fade-in, but not fade-out. So I added an alternative fixup, which shouldn't have that problem.

src/components/action-bar/action-bar-item/action-bar-item.scss
Kiarokh, line 12:
```suggestion
        pointer-events: none;
```

LawrenceBorst (approved):

PR #2518, opened 2023-10-02
fix(markdown): update remark versions to latest

jgroth:
There were both build errors and runtime errors with the versions Dependabot had updated to. The versions have been updated to the latest in package.json, the code has been updated to accommodate for the breaking changes, and the build was updated to get rid of some runtime errors

The `remark-footnotes` package could not be updated, so it was removed. It also have been deprecated https://github.com/remarkjs/remark-footnotes. The footnotes seem to work anyway, perhaps since we already use `remark-gfm`

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,15 +8,11 @@
       "name": "@limetech/lime-elements",
       "version": "37.1.0-next.35",
       "license": "Apache-2.0",
-      "dependencies": {
-        "@rollup/plugin-commonjs": "^25.0.4",
-        "rollup": "^3.29.3"
-      },
       "devDependencies": {
         "@commitlint/config-conventional": "^17.7.0",
         "@popperjs/core": "^2.11.8",
         "@rjsf/core": "^2.4.2",
-        "@rollup/plugin-node-resolve": "^15.2.1",
+        "@rollup/plugin-node-resolve": "^13.3.0",
         "@stencil/core": "^3.4.1",
         "@stencil/sass": "^3.0.6",
         "@types/codemirror": "^5.60.2",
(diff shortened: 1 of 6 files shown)


Kiarokh:
should we use `fix`? since the component is still `@private`? I think it's better to use `chore`

john-traas:
I could be wrong, but isn't `chore` for things like only updating versions, but this update has some changes to the code as well that `fix` the issue?

jgroth:
> should we use `fix`? since the component is still `@private`? I think it's better to use `chore`

Well, the original commit was a `feat` so... :smile: 

> I could be wrong, but is `chore` for things like updating versions, but this update has some changes to the code as well that `fix` the issue?

`fix` is for fixing bugs. That could be just updating a version of some package, but more likely it contains changes to the source files as well. `chore` is more for maintenance stuff. To me this issue was clearly fixing a bug since neither the build worked, or the component in the latest docs :smile: 

This is what `git cz` says:

```
  feat:     A new feature 
  fix:      A bug fix 
  docs:     Documentation only changes 
  style:    Changes that do not affect the meaning of the code (white-space, formatting, missing semi-colons, etc) 
  refactor: A code change that neither fixes a bug nor adds a feature 
  perf:     A code change that improves performance 
  test:     Adding missing tests or correcting existing tests 
  build:    Changes that affect the build system or external dependencies (example scopes: gulp, broccoli, npm) 
  ci:       Changes to our CI configuration files and scripts (example scopes: Travis, Circle, BrowserStack, SauceLabs) 
  chore:    Other changes that don't modify src or test files 
  revert:   Reverts a previous commit 
```

john-traas (approved):
Looks really good.

PR #2522, opened 2023-10-03
Fix exported types

jgroth:
A lot of the exported types did not work, probably since #2370 was merged. The types could be imported by consumers, but many interfaces had lost all type information, basically converting the types to `any`. It appears if types are not imported using relative paths in our source files, the types in `components.d.ts` are generated incorrectly.

Also, the examples should import all types from the public package, otherwise it will be confusing for consumers that are reading the documentation.

I will investigate if we can create some kind of lint rule that will prevent us from making this mistake again in the future.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/action-bar/action-bar-item/action-bar-item.tsx
+++ b/src/components/action-bar/action-bar-item/action-bar-item.tsx
@@ -6,9 +6,8 @@ import {
     h,
     Prop,
 } from '@stencil/core';
-import { ActionBarItem } from '../action-bar.types';
+import { ActionBarItem, ListSeparator } from '../../../interface';
 import { createRandomString } from '../../../util/random-string';
-import { ListSeparator } from 'src/interface';
 import {
     makeEnterClickable,
     removeEnterClickable,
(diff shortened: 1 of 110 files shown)


jgroth:
This should also fix some broken types in the documentation, e.g. here is what `ListItem` looks like before and after (there are probably more broken types)
image

jgroth:
It also seems the types might be broken when using the dev build (`npm start`). Not sure how to get around that, but it's more important that the production build works. Most types seem to work for the dev build as well, but `ListItem` and `Tab` seem to be incorrectly exported from `components.d.ts`

```ts
export { ListItem, Tab } from "@limetech/lime-elements";
```

Kiarokh (approved):

adrianschmidt:
Nice! I didn't get to the email notification I had received about this until now, but really good find! And thanks for the fix! 🙏 ❤️

PR #2540, opened 2023-10-16
fix(input-field): apply the user's locale when formatting numbers

Befkadu1:
The property "locale" specifies the user's locale and/or preferences in which numbers should be formatted. This affect the formatting of integers and decimals

fix Lundalogik/crm-feature#3673

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -29,6 +29,7 @@ import { getHref, getTarget } from '../../util/link-helper';
 import { JSXBase } from '@stencil/core/internal';
 import { createRandomString } from '../../util/random-string';
 import { LimelListCustomEvent } from 'src/components';
+import config from '../../global/config';
 
 interface LinkProperties {
     href: string;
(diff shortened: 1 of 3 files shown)


src/components/input-field/input-field.tsx
```diff
@@ -217,6 +217,12 @@ export class InputField {
     @Prop({ reflect: true })
     public showLink = false;
 
+    /**
+     * The user's locale that determines how numbers are formatted.
+     */
+    @Prop({ reflect: true })
+    public locale: string = navigator.language;
```
vicgeralds, line 224: We could use global config here (`import config from '../../global/config';`)

```suggestion
    public locale: string = config.defaultLocale;
```

The icon component is an example that uses global config, which is provided using `limel-config`

Befkadu1: Would it be better to initialize `defaultLocale` to `en`? 

```
export class Config {
    public iconPath = '';
    public defaultLocale: string
    public featureSwitches: any = getFeatureSwitches(localStorage);
}

vicgeralds: no, `navigator.language` is better I think, as it's the user's preferred language

vicgeralds (approved):
Looks good

Issue #2505, opened 2023-09-26
Use the new `Icon` interface in more components.


Kiarokh:
## New feature motivation
Consistency! So instead of writing
```tsx
…
text: 'test',
iconFillColor: 'black',
iconBackgroundColor: 'white',
iconTitle: 'Useful title',
link: {
    href: 'www.google.com',
    target: '_blank',
},
…
```
Just like using the `Link` interface, one would do the same for icons:

```tsx
…
text: 'test',
icon: {
    color: 'black',
    backgroundColor: 'white',
    title: 'Useful title'
},
link: {
    href: 'www.google.com',
    target: '_blank',
},
…
```

## New feature description
We have [a new interface for Icon](https://github.com/Lundalogik/lime-elements/issues/1303), here src/global/shared-types/icon.types.ts 
Currently, the new `Icon` interface is _ONLY_ used in the `limel-breadcrumbs` only. But we need to use it in more components.

## New feature implementation
We don't wanna make a breaking change. Using this new API will only be a breaking change if we force someone to use the new type without allowing the previous ways of doing it.

PR #2542, opened 2023-10-16
Deprecate `iconColor`, `iconBackgroundColor` etc, in favour of using the `Icon` interface

Kiarokh:
fix https://github.com/Lundalogik/lime-elements/issues/2505

💡 Don't get scared of the number of files changed in this PR. Most of the files are the updated examples that now use the new `Icon` interface. The logic of deprecating the old props works exactly the same for all components. Once you see and understand one of them, you will have an easy time understanding the rest of them.

### Which components are updated:
This is the list of component that are affected:

- [x] menu
- [x] list
- [x] action bar
- [x] chipset
- [x] select
- [x] progress flow
- [x] picker
- [x] tab bar
- [x] file

# How to test this PR?

`iconColor`, `iconFillColor`, `iconBackgroundColor` and `iconTitle` props are deprecated now in different components. But they should still work if they are used.

The old way may look like this:
```tsx
icon: 'name',
iconFillColor: 'black',
iconBackgroundColor: 'white',
iconTitle: 'title',
```
And the new way looks like this:
```tsx
icon: {
	name: 'name',
    color: 'black',
    backgroundColor: 'white',
    title: 'title',
},
```

⚠️ Keep in mind that the old way still work, but the combination of the old and new will not work! So for example a config like below wouldn't work and the `iconColor` would not get applied:
```tsx
icon: {
	name: 'name',
},
iconColor: 'color'
```

An easy test is to link this to the web client (where all the configs in crm-components are still old) and see if each component still show the colors as expected. 

Otherwise, a manual test is to change back the "examples" in Lime Elements docs to what they were before and make sure that each component renders the colors as expected.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/action-bar/action-bar-item/action-bar-item.tsx
+++ b/src/components/action-bar/action-bar-item/action-bar-item.tsx
@@ -12,6 +12,7 @@ import {
     makeEnterClickable,
     removeEnterClickable,
 } from 'src/util/make-enter-clickable';
+import { getIconColor, getIconName } from '../../icon/get-icon-props';
 
 /**
  * @private
@@ -55,6 +56,10 @@ export class ActionBarButton {
 
     public componentWillLoad() {
         makeEnterClickable(this.host);
+    }
+
+    public componentDidLoad() {
+        this.triggerIconColorWarning();
     }
 
     public disconnectedCallback() {
@@ -105,11 +110,14 @@ export class ActionBarButton {
         }
 
         if ('icon' in this.item) {
+            const name = getIconName(this.item.icon);
+            const color = getIconColor(this.item.icon, this.item.iconColor);
+
             return (
                 <limel-icon
-                    name={this.item.icon}
+                    name={name}
                     style={{
-                        '--action-bar-item-icon-color': `${this.item.iconColor}`,
+                        '--action-bar-item-icon-color': `${color}`,
                     }}
                 />
             );
@@ -145,6 +153,15 @@ export class ActionBarButton {
                     elementId={this.tooltipId}
                     label={this.item.commandText}
                 />
+            );
+        }
+    }
+
+    private triggerIconColorWarning() {
+        if (this.isItem(this.item) && this.item.iconColor) {
+            /* eslint-disable-next-line no-console */
+            console.warn(
+                "The `iconColor` prop is deprecated now! Use the new `Icon` interface and instead of `iconColor: 'color-name'` write `icon {name: 'icon-name', color: 'color-name'}`."
             );
         }
     }
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -20,6 +20,12 @@ import { handleKeyboardEvent } from './chip-set-input-helpers';
 import translate from '../../global/translations';
 import { getHref, getTarget } from '../../util/link-helper';
 import { isEqual } from 'lodash-es';
+import {
+    getIconBackgroundColor,
+    getIconColor,
+    getIconName,
+    getIconTitle,
+} from '../icon/get-icon-props';
 
 const SELECTED_CHIP_CLASS = 'mdc-chip--selected';
 const INPUT_FIELD_TABINDEX = 1;
@@ -262,6 +268,8 @@ export class ChipSet {
     }
 
     public componentDidLoad() {
+        this.triggerIconColorWarning(this.value);
+
         if (this.type === 'input') {
             this.mdcTextField = new MDCTextField(
                 this.host.shadowRoot.querySelector('.mdc-text-field')
@@ -700,23 +708,30 @@ export class ChipSet {
     }
 
     private renderChipIcon(chip: Chip) {
+        const name = getIconName(chip.icon);
+        const color = getIconColor(chip.icon, chip.iconFillColor);
+        const backgroundColor = getIconBackgroundColor(
+            chip.icon,
+            chip.iconBackgroundColor
+        );
+        const title = getIconTitle(chip.icon, chip.iconTitle);
         const style = {};
-        if (chip.iconFillColor) {
-            style['--icon-color'] = chip.iconFillColor;
+        if (color) {
+            style['--icon-color'] = color;
         }
 
-        if (chip.iconBackgroundColor) {
-            style['--icon-background-color'] = chip.iconBackgroundColor;
+        if (backgroundColor) {
+            style['--icon-background-color'] = backgroundColor;
         }
 
         return (
             <limel-icon
                 class="mdc-chip__icon mdc-chip__icon--leading"
-                name={chip.icon}
+                name={name}
                 style={style}
                 size="small"
                 badge={true}
-                title={chip.iconTitle}
+                title={title}
             />
         );
     }
@@ -804,5 +819,21 @@ export class ChipSet {
         }
 
         return <limel-badge label={chip.badge} />;
+    }
+
+    private triggerIconColorWarning(value: Chip[]) {
+        for (const chip of value) {
+            if (
+                chip.icon &&
+                (chip.iconFillColor ||
+                    chip.iconBackgroundColor ||
+                    chip.iconTitle)
+            ) {
+                /* eslint-disable-next-line no-console */
+                console.warn(
+                    "The `iconFillColor`, `iconBackgroundColor`, and `iconTitle` props are deprecated now! Use the new `Icon` interface and instead of `iconColor: 'color-name', `iconBackgroundColor: 'color-name', and `iconTitle: 'title'`, write `icon { name: 'icon-name', color: 'color-name', backgroundColor: 'color-name', title: 'title' }`."
+                );
+            }
+        }
     }
 }
--- a/src/components/file/file.tsx
+++ b/src/components/file/file.tsx
@@ -240,11 +240,13 @@ export class File {
                 ...DEFAULT_FILE_CHIP,
                 text: this.value.filename,
                 id: this.value.id,
-                icon: getFileIcon(this.value),
-                iconFillColor: getFileColor(this.value),
-                iconBackgroundColor: getFileBackgroundColor(this.value),
+                icon: {
+                    name: getFileIcon(this.value),
+                    title: getFileExtensionTitle(this.value),
+                    color: getFileColor(this.value),
+                    backgroundColor: getFileBackgroundColor(this.value),
+                },
                 href: this.value.href,
-                iconTitle: getFileExtensionTitle(this.value),
             },
         ];
     }
--- a/src/components/icon/get-icon-props.ts
+++ b/src/components/icon/get-icon-props.ts
@@ -1,0 +1,117 @@
+import { Icon } from '../../interface';
+
+/**
+ * Get the icon name.
+ * This helps in setting the right icon, both if the consumer only types `icon: string`,
+ * and if they use the `Icon` interface for writing the icon name.
+ * @param {string | Icon | undefined} icon - The icon to retrieve the name from.
+ * @returns {string | undefined} The icon name or the provided string, or `undefined` if `icon` is falsy.
+ */
+export function getIconName(
+    icon: string | Icon | undefined
+): string | undefined {
+    if (typeof icon === 'object' && 'name' in icon) {
+        return icon.name;
+    }
+
+    if (typeof icon === 'string') {
+        return icon;
+    }
+
+    return undefined;
+}
+
+/**
+ * Get the icon color.
+ * This is added because the old way of specifying `iconColor`
+ * deprecated in our components. So consumers should now use the new
+ * `Icon` interface instead. But our components must still support the
+ * old way of writing `iconColor: string`.
+ * @param {string | Icon | undefined} icon - The icon to retrieve the color from.
+ * @param {string | undefined} [iconColor] - The color to use when the deprecated `iconColor` is used.
+ * @returns {string | undefined} The icon color or the provided color string, or `undefined` if `iconColor` is falsy.
+ */
+export function getIconColor(
+    icon: string | Icon | undefined,
+    iconColor?: string | undefined
+): string | undefined {
+    if (typeof icon === 'object' && 'color' in icon) {
+        return icon.color;
+    }
+
+    if (typeof icon === 'string') {
+        return iconColor;
+    }
+
+    return undefined;
+}
+
+/**
+ * Get the icon color.
+ * This is added because the old way of specifying `iconFillColor` is
+ * deprecated in our components. So consumers should now use the new
+ * `Icon` interface instead. But our components must still support the
+ * old way of writing `iconFillColor: string`.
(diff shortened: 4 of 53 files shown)


src/components/picker/examples/picker-icons.tsx
Kiarokh, line 1: Selecting items in this example doesn't work anymore, and I have no idea why

Kiarokh:
I've update to the `Icon` everywhere except `limel-file`!

src/components/icon/get-icon-props.ts
```diff
@@ -0,0 +1,117 @@
+import { Icon } from '../../interface';
+
+/**
+ * Get the icon name.
+ * This helps in setting the right icon, both if the consumer only types `icon: string`,
```
Kiarokh, line 5: now after this PR, consumers can either write `icon: 'icon-name'` or `icon: {name: 'icon-name'}`.

However the only component that only works by writing `icon: {name: 'icon-name'}` is `limel-breadcrumbs`. I wonder if we should do so that both ways work for breadcrumbs as well? 🤔 otherwise, it'll be weird.

src/components/menu/menu.tsx
```diff
@@ -90,6 +90,13 @@ export class Menu {
         this.portalId = createRandomString();
     }
 
+    public componentWillLoad() {
+        /* eslint-disable-next-line no-console */
+        console.warn(
+            "The `iconColor` prop is deprecated now! Use the new `Icon` interface and instead of `iconColor: 'color-name'` write `icon {name: 'icon-name', color: 'color-name'}`."
+        );
+    }
```
adrianschmidt, line 98: We could warn only if the `iconColor` attribute is used.

Kiarokh: oh yes please! how to do that. give me an example and I'll fix it everywhere

src/components/icon/get-icon-props.ts
Kiarokh: I buy that. But from a consumer's perspective it also felt convenient (in a smart way) not to _have to_ specify `icon` and then `name`, if all you need is the `name` (you're not interested in `title`, `color`, etc…). Also in the docs, we don't say that this -> `icon: {name: 'icon-name'}` is also deprecated.

src/components/menu/menu.tsx
adrianschmidt: I'm not sure if the input values are available yet in `componentWillLoad`, so you might have to do it in `componentDidLoad` instead, but you could just look at `this.iconColor`. If it has a truthy value, log the warning.

src/components/icon/get-icon-props.ts
adrianschmidt: Ah! I think I misunderstood. Yeah, I think the property `icon` should be of type `string | Icon` everywhere, and if it's just a string, we use that for the name of the icon to show. Absolutely 👍

I thought you were asking whether we should implement `iconColor` etc on breadcrumbs. Sorry, I clearly wasn't paying attention to what I was reading 🙈

src/components/menu/menu.tsx
Kiarokh: I'll test

src/components/menu-list/menu-list-renderer.tsx
```diff
@@ -17,6 +18,15 @@ export class MenuListRenderer {
...
     private applyTabIndexToItemAtIndex: number;
 
+    public componentDidLoad(item) {
+        if (item.iconColor) {
+            /* eslint-disable-next-line no-console */
+            console.warn(
+                "The `iconColor` prop is deprecated now! Use the new `Icon` interface and instead of `iconColor: 'color-name'` write `icon {name: 'icon-name', color: 'color-name'}`."
+            );
+        }
+    }
```
Kiarokh, line 28: @adrianschmidt I cannot figure out how to do it for menu-list and list. Can you add a suggestion or fixup plz?

adrianschmidt: I don't have time right now. I can look at it during cooldown if you remind me. I think it's ok to leave those two for now though…

Or maybe someone else can find the time to help you? Someone working on the pothole cycle maybe? 🤞 🤗

src/components/picker/picker.tsx
```diff
@@ -318,12 +323,18 @@ export class Picker {
...
+        );
+
         return {
             id: `${listItem.value}`,
             text: listItem.text,
             removable: true,
-            icon: listItem.icon,
-            iconBackgroundColor: listItem.iconColor,
+            icon: name,
+            iconBackgroundColor: backgroundColor,
```
Kiarokh, line 337: something is wrong here. the chosen chip's color is not applied
image

also tested this: 
```tsx
return {
            id: `${listItem.value}`,
            text: listItem.text,
            removable: true,
            icon: {
                name: name,
                backgroundColor: backgroundColor,
            },
            value: listItem,
        };
``` 
but got the same result

src/components/picker/picker.tsx
```diff
@@ -221,10 +221,6 @@ export class Picker {
 
     public componentWillLoad() {
         this.chips = this.createChips(this.value);
-        /* eslint-disable-next-line no-console */
-        console.warn(
-            "The `iconColor` prop is deprecated now! Use the new `Icon` interface and instead of `iconColor: 'color-name'` write `icon {name: 'icon-name', color: 'color-name'}`."
-        );
```
Kiarokh: isn't it enough that we have the `console.warn` in limel-chipset? Since picker is using Chipset anyways?

src/components/menu-list/menu-list-renderer.tsx
Kiarokh: found a way

src/components/select/select.template.tsx
```diff
@@ -324,3 +324,15 @@ function getSelectedIcon(value: any) {
...
+    const selectedValues = Array.isArray(values) ? values : [values];
+    for (const value of selectedValues) {
+        if (value.iconColor) {
+            /* eslint-disable-next-line no-console */
+            console.warn(
+                "The `iconColor` prop is deprecated now! Use the new `Icon` interface and instead of `iconColor: 'color-name'` write `icon {name: 'icon-name', color: 'color-name'}`."
+            );
+        }
+    }
+}
```
Kiarokh, line 338: This doesn't generate the console.warn for limel-select if `iconColor` is used, and I have no idea how to fix this. I tried a lot.

src/components/icon/get-icon-props.ts
Kiarokh: np!

I added a commit here for ti [`6841531` (#2542)](https://github.com/Lundalogik/lime-elements/pull/2542/commits/68415317595d5c0b8ca057668e8dcf2387403b02)

@BjornOstberg I hope this does not override [your recent fix for breadcrumbs](https://github.com/Lundalogik/lime-elements/pull/2550)?

BjornOstberg: I think it will work the same, so it shouldn't be a problem.

src/components/picker/examples/picker-icons.tsx
adrianschmidt: Has this been resolved?

adrianschmidt:
This doesn't really look ready to review to me. I would suggest moving it back into _In Progress_, and if you need help, grab a developer that can help you out.

Also, if that first commit (enabling all the other commits) works on its own, I would suggest putting just that up for review first, so that can get merged. Then you can put up all of the other commits that don't cause any problems, and get those merged while trying to work out how to fix those that _did_ cause problems.

src/components/action-bar/action-bar-item/action-bar-item.tsx
```diff
@@ -105,11 +110,14 @@ export class ActionBarButton {
...
+            const name = getIconName(this.item.icon);
+            const color = getIconColor(this.item.icon, this.item.iconColor);
+
             return (
                 <limel-icon
-                    name={this.item.icon}
+                    name={name}
                     style={{
-                        '--action-bar-item-icon-color': `${this.item.iconColor}`,
+                        '--action-bar-item-icon-color': `${color}`,
```
Kiarokh, line 120: 🤔 now if the `iconColor` is defined (the old way of doing it), the `--action-bar-item-icon-color` becomes `undefined`. 

image

src/components/picker/examples/picker-icons.tsx
Kiarokh: this works now again 🤷

src/components/select/select.template.tsx
```diff
@@ -305,17 +308,30 @@ function getSelectedIcon(value: any) {
...
+export function triggerIconColorWarning(options: Option[]) {
+    options.forEach((option) => {
+        if (option.iconColor) {
+            /* eslint-disable-next-line no-console */
+            console.warn(
+                "The `iconColor` prop is deprecated now! Use the new `Icon` interface and instead of `iconColor: 'color-name'` write `icon {name: 'icon-name', color: 'color-name'}`."
+            );
+        }
+    });
+}
```
Kiarokh, line 337: This should generate console.warn for limel-select **ONLY** when `iconColor` is used. But the warning in the console constantly shows up and I have no idea how to fix this. I tried a lot.

src/components/chip-set/chip-set.tsx
```diff
@@ -789,4 +804,20 @@ export class ChipSet {
...
+    private triggerIconColorWarning(value: Chip[]) {
+        for (const chip of value) {
+            if (
+                chip.icon &&
+                (chip.iconFillColor ||
+                    chip.iconBackgroundColor ||
+                    chip.iconTitle)
+            ) {
+                /* eslint-disable-next-line no-console */
+                console.warn(
```
Kiarokh, line 817: it is enough that we have the `console.warn` in `limel-chip-set` only. Bacuse `limel-picker` and `limel-file` are using `limel-chip-set`, and will show the console warning

src/components/select/select.template.tsx
Kiarokh: @jgroth do you have any magic here? I can't figure out what the problem is here, and why the console warning is always being displayed.

src/components/file/file.tsx
```diff
@@ -235,8 +235,8 @@ export class File {
                 icon: getFileIcon(this.value),
                 iconFillColor: getFileColor(this.value),
                 iconBackgroundColor: getFileBackgroundColor(this.value),
-                href: this.value.href,
                 iconTitle: getFileExtensionTitle(this.value),
```
Kiarokh, line 238: I don't know how to make these work with the new Icon interface in a backwards compatible way. 

But also this one could be a breaking change if we want. The impact is super low. I don't think anyone have customized their icon's color in limel-file. They should be really "sick" if they have done so! 😅 And if they upgrade and get no colors, it won't be a big deal.

Kiarokh: @jgroth I'd also need help with this one, if you have time.

Kiarokh:
> […] if that first commit (enabling all the other commits) works on its own, I would suggest putting just that up for review first, so that can get merged. 

It works on its own, but has no effect. It should at least be applied somewhere in a component to make sense.
Since our components use each other to render their own UIs, we get console warnings in many places as a reuslt. These warnings are inherited from the internal-reused components which now have deprecated props. 

I could theoretically deprecate the props in 1 component, together with the first commit and make a PR for review (for example in `limel-menu-list`). But if I do so, then I should take out `getIconFillColor`, `getIconBackgroundColor`, and `getIconTitleColor` out from the first commit, and add them later.

Anyhow, this could be a good start as you suggest! It just makes more paper work, and the next reviewer may have to go through everything and relearn them again, to be able to review deprecations in the other componentns. So I'm tentative.

> Then you can put up all of the other commits that don't cause any problems, and get those merged while trying to work out how to fix those that _did_ cause problems.

Currently everything works fine. The only place I couldn't fix is `limel-file`. We can **not** leave it out, since `limel-chipset` is throwing console warnings in limel-file, and says the props are deprecated. But limel-file doesn't support the new Icon interface yet. So this has to be fixed together with chipset and picker.

src/components/select/select.template.tsx
```diff
@@ -254,14 +255,16 @@ function createMenuItems(
...
 
         return {
             text: text,
             selected: selected,
             disabled: disabled,
             value: option,
-            icon: option.icon,
-            iconColor: option.iconColor,
+            icon: name,
+            iconColor: color,
```
jgroth, line 267: Why are we using `iconColor` here if we have deprecated it :smile: 
This is why you get the warning in the console you asked about here https://github.com/Lundalogik/lime-elements/pull/2542#discussion_r1376445815

src/components/select/select.template.tsx
jgroth: https://github.com/Lundalogik/lime-elements/pull/2542/files#r1377545428

src/components/file/file.tsx
jgroth: Should it not just be something like this? :thinking: 
```
icon: {
    name: getFileIcon(this.value),
    title: getFileExtensionTitle(this.value),
    color: getFileColor(this.value),
    backgroundColor: getFileBackgroundColor(this.value),
},
```

src/components/select/select.template.tsx
Kiarokh: Ah I see 😓 tnx

src/components/chip-set/chip-set.tsx
john-traas: I think we need to reconsider where to place the console.warn because right now it feels a bit aggressive and the console just fills with the warnings. 😅

john-traas: It's also not very easy to read/understand because of the lack of formatting in the console messages. 
Maybe we could create a short documentation that explains the difference and link there from the console.warns instead?

john-traas: It's not terrible, but this is only after clicking a couple of times in the browser. 
Screenshot 2023-11-09 at 11 35 02

Kiarokh: Well, I'd say this is actually the least aggressive way of doing it 😄 and the experience shows that this is not enough for consultants and appers to update their components. After this PR is merged, we should update our code in crm-components and building-blocks, and use the new `Icon` interface ourselves. Then we won't see these errors anymore. :)

john-traas: Do we already have issues created for those?

Kiarokh: no. But I won't forget this :D

john-traas (approved):
Looks good and works. Maybe we can update the relevant components with the new syntax over the coming week.

PR #2560, opened 2023-10-26
feat(column sorting): set default sort direction to descending

john-traas:
This sets the default sorting direction, when clicking on the tabulator in the table view, to descending. 

Fixes https://github.com/Lundalogik/crm-feature/issues/3655

### DIFF
--- a/src/components/table/columns.ts
+++ b/src/components/table/columns.ts
@@ -24,6 +24,7 @@ export class ColumnDefinitionFactory {
             field: column.field,
             hozAlign: column.horizontalAlign,
             headerSort: column.headerSort,
+            headerSortStartingDir: 'desc',
         };
 
         if (column.headerComponent) {
--- a/src/components/table/table.e2e.ts
+++ b/src/components/table/table.e2e.ts
@@ -81,20 +81,20 @@ describe('limel-table', () => {
             headerA.click();
             await page.waitForChanges();
             rowData = await getFirstRowContent();
-            expect(rowData[0]).toEqual('1');
+            expect(rowData[0]).toEqual('2');
             headerA.click();
             await page.waitForChanges();
             rowData = await getFirstRowContent();
-            expect(rowData[0]).toEqual('2');
+            expect(rowData[0]).toEqual('1');
 
             headerB.click();
             await page.waitForChanges();
             rowData = await getFirstRowContent();
-            expect(rowData[1]).toEqual('ascending');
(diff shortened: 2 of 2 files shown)


src/components/table/columns.ts
```diff
@@ -24,6 +24,7 @@ export class ColumnDefinitionFactory {
             field: column.field,
             hozAlign: column.horizontalAlign,
             headerSort: column.headerSort,
+            headerSortStartingDir: 'desc',
```
vicgeralds, line 27: Do we want to change the default in lime-elements without providing a way to change it? Also seems like a breaking change 🤔 

I think we rather want to add `headerSortStartingDir` to our `Column` type and set that instead (possibly depending on the field type) when passing the `columns` prop the the table component.

Or maybe call it `headerSortDescending: boolean` instead 🙂

vicgeralds (commented):
Good job finding how to change this!

I still think limel-table needs to flexible and allow different sort starting directions, not just change it to descending always. So we need a way to pass the desired sort starting direction in `columns` I think.

john-traas:
> Good job finding how to change this!
> 
I can't take the credit, I was close but in the end @LawrenceBorst found where the trail eventually led 😅

src/components/table/columns.ts
john-traas: I agree, in order to implement it for specific field types we'd have to do that. I'll give that approach a try. 😃 

This first approach won't set the default sorting on columns to descending when the table is loaded, it simply ensures that descending is the first direction the tabulator sets on a column when the user clicks on it to sort it.

vicgeralds: > I'm not sure if this approach is a breaking changes as it won't set the default sorting on columns to descending when the table is loaded, it simply ensures that descending is the first direction the tabulator sets on a column when the user clicks on it to sort it.

You're right. It's probably not a big deal to change that default behaviour. 🙂

vicgeralds (approved):
I encourage you try the approach that enables passing a sorting starting direction per column, but I'm still ready to approve this if it turns out too difficult or you don't have the time. 🙂

john-traas:
On speaking with Tomas about this the other day, I think we can go ahead and merge this one. I'll create a separate issue for handling it for specific fields, I'll explore that at the start of cooldown and if it's a simple fix, we can implement it, but if not then Tomas has said that we can wait for an upcoming pitch regarding custom layouts that Product wants us to build.

PR #2562, opened 2023-10-27
fix(form): animate sorting of collapsible sections

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/form.scss
+++ b/src/components/form/form.scss
@@ -286,3 +286,14 @@ limel-code-editor {
 }
 
 @import './row/row.scss';
+
+limel-collapsible-section {
+    transition: transform 0.3s ease-in-out;
+
+    &.move-up-transition {
+        transform: translateY(-100%);
+    }
+    &.move-down-transition {
+        transform: translateY(100%);
+    }
+}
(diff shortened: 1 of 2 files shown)


src/components/form/templates/array-field-collapsible-item.ts
```diff
@@ -103,7 +103,53 @@ export class CollapsibleItemTemplate extends React.Component {
...
+            if (previousSection) {
+                previousSection.classList.remove(down);
+            }
+
+            if (nextSection) {
+                nextSection.classList.remove('move-up-transition');
+            }
+
+            // Perform the action
+            event.detail.run(event);
```
Kiarokh, line 151: We have to wait for the transitions to happen, and then update the index number and simultaneously and instantly remove the transition classes. This timing however doesn't seem to work always. Maybe the implementation should be totally different. I don't know

src/components/form/form.scss
```diff
@@ -286,3 +286,14 @@ limel-code-editor {
...
+
+limel-collapsible-section {
+    transition: transform 0.3s ease-in-out;
+
+    &.move-up-transition {
+        transform: translateY(-100%);
+    }
+    &.move-down-transition {
+        transform: translateY(100%);
+    }
```
Kiarokh, line 298: this only works when all collapsible sections have the same size. So for example if one is closed and one is open, this doesn't work and make a silly visual effect.

PR #2565, opened 2023-10-30
docs(CONTRIBUTING.md): add section about how we structure our code

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,6 +115,33 @@ To ensure consistency and quality throughout the source code, before being accep
 **Please note**:
 We will be happy to help you out in meeting these requirements. Don't be afraid to share your code in a pull request even if it doesn't meet the requirements yet.
 
+#### Code Structure
+
+We try to follow a practice known as the "Newspaper code structure", which in short means putting all non-function properties on top; and ordering them with public properties first, then more important private properties, then less important private properties.
+
+So that gives us roughly:
+1. `@Prop` properties. (Typically all public properties are decorated with `@Prop`.)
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -115,6 +115,35 @@ To ensure consistency and quality throughout the source code, before being accep
 **Please note**:
 We will be happy to help you out in meeting these requirements. Don't be afraid to share your code in a pull request even if it doesn't meet the requirements yet.
 
+#### Code Structure
+
+We try to follow the "Newspaper code structure", see https://kentcdodds.com/blog/newspaper-code-structure. (Yes, we are aware that the author has archived that blog post because he doesn't follow that principle anymore, but we still think it's a good fit for us.)
+
+We put all non-function properties on top, so we want to order them with public properties first, then more important private properties, then less important private properties.
```
Kiarokh, line 122:
```suggestion
We try to follow the "Newspaper code structure", which in short means putting all non-function properties on top; and ordering them with public properties first, then more important private properties, then less important private properties.
```

Kiarokh: People can always Google stuff. If we think using this is good, there is no need to bring up excuses. Let's keep it simple.

adrianschmidt: Ok 👍

Kiarokh (approved):

PR #2567, opened 2023-10-30
Set up api extractor

adrianschmidt:
### Why?

The point of this PR is to be able to check if lime-elements' public API is changed by a PR. If so, we want to require a code review by the code owners (the lime-elements-maintainers team). This way, we can get rid of the need for making pre-releases from the `next` branch. Instead, any changes that do not affect the public API can be reviewed by anyone with write access. And for any changes that do affect the public API, anyone from the lime-elements-maintainers team can simply review the changes to the public API, while the review of the internal implementation can be made by someone else.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


Kiarokh:
I remember that we have linked to `next` in a couple of places from our Docs.

One of them is here: 

- https://github.com/Lundalogik/lime-crm-components/blob/1b5a7deb74572e0e95ec5783ecb037fa161be850/src/history-view/history-item/markdown/markdown.tsx#L35

and in a few places in lime-crm docs:

- https://github.com/search?q=repo%3ALundalogik%2Flime-crm+%2Fversions%2Fnext%2F&type=code

and maybe also elswhere.

I wonder if we should fix them, or if we can redirect `next` urls to `main`?

adrianschmidt:
> I wonder if we should fix them, or if we can redirect next urls to main?

Yeah, we can keep the `next`symlink, but point it to the latest regular release instead 👍

adrianschmidt:
> > I wonder if we should fix them, or if we can redirect next urls to main?
> 
> Yeah, we can keep the `next`symlink, but point it to the latest regular release instead 👍

(Note that _this_ PR does not change or require us to change anything about the release branches. It just means that we can get rid of the `next`-branch when we want.)

src/components/list/list.tsx
adrianschmidt, line 1: Hmm… the links to types no longer work… 😕 🤔

image

adrianschmidt: I suspect this will require some update in Kompendium? 🤔

adrianschmidt: Ah! The whole _Types_ section of the navigation menu is actually empty. So I guess types just aren't extracted correctly. Which makes sense, since I had to remove the types from the docblocks when switching from jsdoc to tsdoc.

So I guess we'll have to look over how to extract those again.

adrianschmidt: I found what the problem was. I had to rename `src/interface.d.ts` to `src/interface.ts`, because api-extractor would incorrectly include it otherwise. But I hadn't updated the value in `stencil.config.ts` and `stencil.config.docs.ts`.

Now the types are both extracted and linked to again 👍

PR #2570, opened 2023-10-31
Feat/menu custom width dropdown

TommyLindh2:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu-surface/menu-surface.scss
+++ b/src/components/menu-surface/menu-surface.scss
@@ -2,6 +2,11 @@
 @use '@material/elevation';
 @use '@material/menu';
 
+/**
+* @prop --limel-menu-surface-display: defines whether the surface is treated as a block, flex or grid. It affects layout used to display its children. Defaults to `block`.
+* @prop --limel-menu-surface-flex-direction: defines the direction of menu-surface layout.
+*/
+
 :host(limel-menu-surface) {
     display: block;
     max-height: inherit;
@@ -11,6 +16,8 @@
 @include menu.core-styles;
 
 .mdc-menu-surface {
+    display: var(--limel-menu-surface-display, block);
+    flex-direction: var(--limel-menu-surface-flex-direction, row);
     max-height: inherit;
     position: relative;
     --mdc-menu-max-width: var(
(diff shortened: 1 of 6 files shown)


src/components/menu/examples/menu-size.tsx
```diff
@@ -0,0 +1,57 @@
+import { MenuItem, ListSeparator } from '@limetech/lime-elements';
+import { Component, Element, Host, State, h } from '@stencil/core';
+
+/**
+ * Resize the menu drop-down
```
Kiarokh, line 5: We decided not to go this way, and instead create a prop or something. (work in progress)

Because this required too much documentation and explaining what you should do and why, to achieve this effect.

example of a draft that I wrote:
******
 
 In this code, `this.host?.offsetWidth` is used to access the
 `offsetWidth` property of the `host` element with optional chaining.
 Here's how it works:
 1. `this.host` refers to the host property of the component,
 which is assigned to the `HTMLElement` that represents the component itself.
 In Stencil, the `@Element()` decorator is used to
 obtain a reference to the host element for the component.
 1. The `?.` is the optional chaining operator, introduced in JavaScript to
 safely access properties or methods of an object.
 It allows you to avoid runtime errors when accessing properties
 on potentially null or undefined objects.
 1. `offsetWidth` is a property of an HTML element,
 and it represents the width of the element in pixels.
 It tells you how wide the element is in the layout.
 
 So, in the context of the code, `this.host?.offsetWidth` is checking
 whether the host element exists and is not `null` or `undefined`.
 If it does exist, it will access its `offsetWidth` property
 (the width of the host element).
 If `this.host` is null or undefined, it will not throw an error;
 instead, it will return `undefined`.
 
 This value is then used to set the `--menu-surface-width` CSS variable in
 the `style` of the `limel-menu` element.
 This is the CSS variable which can be used to control the width of
 the menu dropdown.
 If `width` has a value (i.e., this.host is not null or undefined),
 it will set the width of the menu dropdown to be the same as
 the width of the host element.
 If width is undefined, no specific width is set, and the menu
 dropdown will inherit its width from its contents.
 This is a technique used to make the menu dropdown as wide as
 the trigger element in this specific example.
 ******

Kiarokh (approved):

PR #2572, opened 2023-10-31
Revert bump ajv 8

vicgeralds:
This reverts the upgrade from ajv 6 to ajv 8 and related changes.

Note that we depend on an old version of rjsf that needs an old version of ajv. In newer versions of rjsf, the validation has been decoupled and you can instead provide the validator that you want to use.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


jgroth:
We might need to tell the bot to not update Ajv again, I'm not sure how to do that

vicgeralds:
> We might need to tell the bot to not update Ajv again, I'm not sure how to do that

Yes. I did that by removing ajv as direct dependency.  🙂

adrianschmidt:
> > We might need to tell the bot to not update Ajv again, I'm not sure how to do that
> 
> Yes. I did that by removing ajv as direct dependency. 🙂

You can also specifically tell Dependabot to not upgrade a dependency by editing the repo's `.github/dependabot.yml` file. Just in case removing the direct dependency isn't possible or isn't the right solution.

vicgeralds:
> You can also specifically tell Dependabot to not upgrade a dependency by editing the repo's .github/dependabot.yml file. Just in case removing the direct dependency isn't possible or isn't the right solution.

I removed it because I thought it was better, but that might be up for discussion. To me it seems like we're only using ajv in the form together with rjsf, and it needs a specific version of ajv, so we should use the same version. It's not clear why using our own ajv 8 validator doesn't work, but it also doesn't make sense to include both ajv 6 and ajv 8 I think.

So removing the direct dependency seems reasonable to me to reduce the risk for problems.

adrianschmidt:
Yeah, I didn't mean to imply that I had an opinion one way or the other in this specific case 😅

vicgeralds:
> Yeah, I didn't mean to imply that I had an opinion one way or the other in this specific case 😅

I can see that there's an argument to make that technically we're using ajv as a direct dependency, so we should list it as a direct dependency. I've made that argument to myself 🙂 

A counter argument is that it gives the false impression that we can use any version of ajv.

adrianschmidt:
I think this fix is unnecessary, because the fix for Lundalogik/crm-feature/issues/3370 seems to have fixed this bug too. I had a suspicion it might, because when trying to reproduce it before testing this branch, I noticed that I could reproduce it using a text field (phone number in this case), but not using a relation field (responsible coworker). So I made sure to update my lime-webclient to the version containing the fix that was just merged in lime-crm-components, and I could no longer reproduce the bug with the text field either 🥳

vicgeralds:
> I think this fix is unnecessary, because the fix for https://github.com/Lundalogik/crm-feature/issues/3370 seems to have fixed this bug too. I had a suspicion it might, because when trying to reproduce it before testing this branch, I noticed that I could reproduce it using a text field (phone number in this case), but not using a relation field (responsible coworker). So I made sure to update my lime-webclient to the version containing the fix that was just merged in lime-crm-components, and I could no longer reproduce the bug with the text field either 🥳

Interesting. I couldn't reproduce the bug with ignored validation errors with that fix either.

But I looked a bit closer, and it turns out the errors returned by `getValidationErrors()` are slightly different than the errors returned by `getCustomErrorMessages()` in the form component. That is, the errors returned when we're using an ajv 8 validator are different than the ones returned from rjsf.

The property is using `/` as separator in one version, and `.` as separator in the other.

image
image

So the fix in https://github.com/Lundalogik/crm-feature/issues/3370 is really just a workaround for a breaking change that was introduced when upgrading to ajv 8.

When setting a value to `null` we get a validation error from our own validator (ajv 8). When setting a value to `undefined` we get a validation error from rjsf (ajv 6).

We can make changes to handle that inconsistency (seems simple), but I see no good reason for using ajv 8 together with an old version of rjsf. So I still think we should revert.

adrianschmidt:
> but I see no good reason for using ajv 8 together with an old version of rjsf. So I still think we should revert.

Yeah, that makes sense 👍

PR #2574, opened 2023-10-31
Feat/limel menu cascading items

TommyLindh2:
Reviews of the public API/Interfaces have been done in a separate [PR#2473 ](https://github.com/Lundalogik/lime-elements/pull/2473)

In this PR we have two different features separated by two commits:
- Searchable items
- Enable Sub items

## Review:
- [ ] Commits are [atomic](https://seesparkbox.com/foundry/atomic_commits_with_git)
- [ ] Commits have the correct *type* for the changes made
- [ ] Commits with *breaking changes* are marked as such

### Browsers tested:
(Check any that applies, it's ok to leave boxes unchecked if testing something didn't seem relevant.)

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

PR #2576, opened 2023-11-01
feat(menu): enable having nested menus (sub-menus)

TommyLindh2:
This is one of three PRs that are connected and extracted from #2574 
They are chained in 3 branches and is best reviewed in this order:

1. https://github.com/Lundalogik/lime-elements/pull/2576 **This PR**
2. https://github.com/Lundalogik/lime-elements/pull/2577
3. https://github.com/Lundalogik/lime-elements/pull/2578

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu-list/menu-list-renderer.tsx
+++ b/src/components/menu-list/menu-list-renderer.tsx
@@ -140,6 +140,7 @@ export class MenuListRenderer {
             >
                 {item.icon ? this.renderIcon(this.config, item) : null}
                 {this.renderText(item)}
+                {this.renderSubMenuIcon(item)}
                 {this.renderNotification(item)}
                 {this.twoLines && this.avatarList ? this.renderDivider() : null}
             </li>
@@ -171,6 +172,14 @@ export class MenuListRenderer {
                 </div>
             </div>
         );
+    };
+
+    private renderSubMenuIcon = (item: MenuItem) => {
+        if (!this.hasSubItems(item)) {
+            return;
+        }
+
+        return <limel-icon class="sub-menu-icon" name="-lime-caret-right" />;
     };
 
     private rendertext = (item: ListSeparator) => {
@@ -245,5 +254,9 @@ export class MenuListRenderer {
         }
 
         return <hr class={classes} />;
+    };
+
+    private hasSubItems = (item: MenuItem): boolean => {
+        return Array.isArray(item.items) && item.items.length > 0;
     };
 }
--- a/src/components/menu/examples/menu-open-sub-menu-programmatically.tsx
+++ b/src/components/menu/examples/menu-open-sub-menu-programmatically.tsx
@@ -86,15 +86,15 @@ export class MenuOpenSubMenuProgrammaticallyExample {
             (i) => i.text === 'Format'
         ) as MenuItem;
 
-        const menuToOpen = formatMenuItem.items.find(
-            (i) => i.text === 'Bullets and numbering'
-        ) as MenuItem;
+        const menuToOpen = (
+            formatMenuItem.items as Array<MenuItem | ListSeparator>
+        ).find((i) => i.text === 'Bullets and numbering') as MenuItem;
 
         this.currentSubMenu = {
             ...menuToOpen,
             parentItem: formatMenuItem,
         };
-        this.items = menuToOpen.items;
+        this.items = menuToOpen.items as Array<MenuItem | ListSeparator>;
         this.openMenu = true;
     };
 
--- a/src/components/menu/examples/menu-open-sub-menu-programmatically.tsx
+++ b/src/components/menu/examples/menu-open-sub-menu-programmatically.tsx
@@ -1,0 +1,107 @@
+import {
+    MenuItem,
+    ListSeparator,
+    LimelMenuCustomEvent,
+} from '@limetech/lime-elements';
+import { Component, State, h, Host } from '@stencil/core';
+import { CascadingMenuItems } from './item-constants';
+
+/**
+ * Opening sub-menus programmatically
+ *
+ * **This example is currently not in use because it's an experimental feature**
+ *
+ * It is possible to open any sub-menu in the menu-hierarchy.
+ * This is done by using the parentItem property of the MenuItem class.
+ * @link item-constants.ts
+ */
+@Component({
+    tag: 'limel-example-menu-open-sub-menu-programmatically',
+    shadow: true,
+    styleUrl: 'menu-open-sub-menu-programmatically.scss',
+})
+export class MenuOpenSubMenuProgrammaticallyExample {
+    private readonly rootItems: Array<MenuItem | ListSeparator> =
+        CascadingMenuItems;
+
+    @State()
+    private items: Array<MenuItem | ListSeparator> = this.rootItems;
+
+    @State()
+    private lastSelectedItem: string;
+
+    @State()
+    private currentSubMenu: MenuItem;
+
+    @State()
+    private openMenu: boolean = false;
+
+    public render() {
+        return (
+            <Host>
+                {this.renderMenu()}
+                <limel-example-value
+                    label="Last selected item"
+                    value={this.lastSelectedItem}
+                />
+            </Host>
+        );
+    }
+
+    private renderMenu() {
+        return (
+            <div class="menu-container">
+                <limel-menu
+                    items={this.items}
+                    open={this.openMenu}
+                    currentSubMenu={this.currentSubMenu}
+                    onSelect={this.handleSelect}
+                    onNavigateMenu={this.handleNavigateMenu}
+                    onCancel={this.handleMenuCancel}
+                >
+                    <limel-button label="Menu" slot="trigger" />
+                </limel-menu>
+                <limel-button
+                    label='Shortcut to "Bullets and numbering"'
+                    primary={true}
+                    onClick={this.buttonClick}
+                />
+            </div>
+        );
+    }
+
+    private handleNavigateMenu = (event: LimelMenuCustomEvent<MenuItem>) => {
+        if (!event.detail) {
+            this.items = this.rootItems;
+        }
+    };
+
+    private handleMenuCancel = () => {
+        this.items = this.rootItems;
+        this.openMenu = false;
+    };
+
+    private buttonClick = () => {
+        const formatMenuItem = this.rootItems.find(
+            (i) => i.text === 'Format'
+        ) as MenuItem;
+
+        const menuToOpen = formatMenuItem.items.find(
+            (i) => i.text === 'Bullets and numbering'
+        ) as MenuItem;
+
+        this.currentSubMenu = {
+            ...menuToOpen,
+            parentItem: formatMenuItem,
+        };
+        this.items = menuToOpen.items;
+        this.openMenu = true;
+    };
+
+    private handleSelect = (event: LimelMenuCustomEvent<MenuItem>) => {
+        this.currentSubMenu = null;
+        this.items = this.rootItems;
+        this.lastSelectedItem = event.detail.text;
+        this.openMenu = false;
+    };
+}
--- a/src/components/menu/menu.tsx
+++ b/src/components/menu/menu.tsx
@@ -89,7 +89,7 @@ export class Menu {
      * Is emitted when a menu item is selected.
      */
     @Event()
-    private select: EventEmitter<MenuItem | MenuItem[]>;
+    private select: EventEmitter<MenuItem>;
 
     @Element()
     private host: HTMLLimelMenuElement;
--- a/src/components/menu/menu.tsx
+++ b/src/components/menu/menu.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable sonarjs/no-duplicate-string */
 import {
     Component,
     Event,
@@ -6,15 +7,26 @@ import {
     Prop,
     Element,
     Watch,
+    State,
 } from '@stencil/core';
 import { createRandomString } from '../../util/random-string';
 import { zipObject } from 'lodash-es';
+import { LimelBreadcrumbsCustomEvent } from '@limetech/lime-elements';
+
+import { BreadcrumbsItem } from '../breadcrumbs/breadcrumbs.types';
+import { ListSeparator } from '../list/list-item.types';
+import { OpenDirection, MenuItem, SurfaceWidth } from './menu.types';
+
 import {
-    ListSeparator,
-    MenuItem,
-    OpenDirection,
-    SurfaceWidth,
-} from '../../interface';
+    ARROW_LEFT,
+    ARROW_LEFT_KEY_CODE,
+    ARROW_RIGHT,
+    ARROW_RIGHT_KEY_CODE,
+} from '../../util/keycodes';
+
+interface MenuCrumbItem extends BreadcrumbsItem {
+    menuItem: MenuItem;
+}
 
 /**
  * @slot trigger - Element to use as a trigger for the menu.
@@ -29,6 +41,7 @@ import {
  * @exampleComponent limel-example-menu-hotkeys
  * @exampleComponent limel-example-menu-secondary-text
  * @exampleComponent limel-example-menu-notification
+ * @exampleComponent limel-example-menu-sub-menus
  * @exampleComponent limel-example-menu-composite
  */
 @Component({
@@ -80,6 +93,18 @@ export class Menu {
     public gridLayout = false;
 
     /**
+     * :::warning Internal Use Only
+     * This property is for internal use only. We need it for now, but want to
+     * find a better implementation of the functionality it currently enables.
+     * If and when we do so, this property will be removed without prior
+     * notice. If you use it, your code _will_ break in the future.
+     * :::
+     * @internal
+     */
+    @Prop({ mutable: true })
+    public currentSubMenu: MenuItem;
+
+    /**
      * Is emitted when the menu is cancelled.
      */
     @Event()
@@ -91,8 +116,17 @@ export class Menu {
     @Event()
     public select: EventEmitter<MenuItem>;
 
+    /**
+     * Is emitted when a menu item with a sub-menu is selected.
+     */
+    @Event()
+    public navigateMenu: EventEmitter<MenuItem>;
+
     @Element()
     private host: HTMLLimelMenuElement;
+
+    @State()
+    private menuBreadCrumb: MenuCrumbItem[] = [];
 
     private list: HTMLLimelMenuListElement;
     private portalId: string;
@@ -109,6 +143,7 @@ export class Menu {
 
     public render() {
         const cssProperties = this.getCssProperties();
+
         const dropdownZIndex = getComputedStyle(this.host).getPropertyValue(
             '--dropdown-z-index'
         );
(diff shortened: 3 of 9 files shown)


src/components/menu-list/menu-list-renderer.tsx
```diff
@@ -242,4 +251,8 @@ export class MenuListRenderer {
 
         return <hr class={classes} />;
     };
+
+    private hasSubItems = (item: MenuItem): boolean => {
```
Kiarokh, line 255: I'd change the commit title to `feat(menu): enable having nested menus (sub-menus)`

src/components/menu/menu.tsx
```diff
@@ -6,15 +7,26 @@ import {
...
-    ListSeparator,
-    MenuItem,
-    OpenDirection,
-    SurfaceWidth,
-} from '../../interface';
+    ARROW_LEFT,
+    ARROW_LEFT_KEY_CODE,
+    ARROW_RIGHT,
+    ARROW_RIGHT_KEY_CODE,
+} from '../../util/keycodes';
```
Kiarokh, line 25: https://github.com/Lundalogik/lime-elements/assets/35954987/5532be11-822b-492d-9b3d-20cb94e67fea

There is a bug in the keyboard navigation. When navigating the menu with keys, if I click outside to close the menu, then I can't open it again by clicking on it. (I think this happens if I have used the left and right arrow keys in my keyboard navigation)

adrianschmidt: Has this been fixed, or is this a call for help from the reviewer?

Kiarokh: No I just saw that this bug exists.

adrianschmidt: Ah, I understand. You've helped Tommy and Björn so much with this I stopped thinking of you as a reviewer and started thinking about you as one of the creators of the PR 🤦‍♂️ 😅

For me, it seems like it's pretty consistent that once I've opened the menu through keyboard navigation, regardless of how I closed it (clicking outside, clicking on a (leaf node) menu item, pressing escape, or pressing enter to select a (leaf node) menu item), I can't open the menu again by clicking the menu-button. If I tab to the menu button, I have to press enter twice to open the menu. This also seems consistent for me.

I've also tried opening the menu by clicking the button, and _then_ navigating with the arrow keys, and selecting a (leaf node) menu item, and then I can still open the menu by clicking it again. So there seems to be something that is mucked up by opening the menu using the keyboard. I would guess that there is some internal state that is set only when the menu is opened using the keyboard, and that is only reset by again trying to open the menu using the keyboard (which would explain why the first enter-press does nothing).

adrianschmidt: Further evidence: if I get into the bugged state, then tab to the button and press enter _once_, I can then open the menu by clicking on it, as usual.

src/components/menu/examples/menu-open-sub-menu-programatically.tsx
```diff
@@ -0,0 +1,113 @@
...
+            },
+            {
+                text: 'Checklist',
+                icon: 'todo_list',
+                parentItem: selectedItem,
+            },
+        ];
+        this.selectedMenuItem = selectedItem;
+        this.items = selectedItems;
+        this.openMenu = true;
```
adrianschmidt, line 104: I find it confusing that we have both `selectedItem` and `selectedItems` here.

The `selectedItems` array is, unless I'm mistaken, simply the items the menu should be showing at the moment. No one has selected them yet.

The name `selectedMenuItem` for the property that determines what sub-menu to show is also really confusing. Sure, I can see that you probably have selected that item to get to the state where the sub-menu is shown (unless the menu was opened programmatically), but I think a name like `submenuToDisplay` or `currentSubmenu` would be better. I know it's a little weird since that item doesn't really _contain_ the menu items shown, but that's what we convey to the user, so I think it's easier to grasp the code if the code is somewhat aligned with that concept.

src/components/menu/menu.tsx
TommyLindh2: After looking into it a bit it seems like what's causing this is the `makeEnterClickable` class that is supposed to prevent holding the Enter to produce a lot of click events, that is used for `limel-button`

Sometime the Up/Down events are triggered in the opposite order i.e. First "Up", then "Down", which will tell the click handler that the button is already "active" and should not let click events passed. So when you use Enter again, the up/down is triggered in the right order that will make it work again. Not sure how to fix it though.

TommyLindh2: I can re-create it in this docs as well: https://lundalogik.github.io/lime-elements/versions/NEXT-37.1.0-next.52/#/component/limel-menu/

src/components/menu/examples/menu-open-sub-menu-programatically.tsx
TommyLindh2: Makes sense, will fix that

adrianschmidt: Hmm… hang on… in the `CascadingMenuItems` imported from `item-constants.ts`, the sub-menu items actually do _contain_ the items included in that sub-menu. Why is it different here?

Another difference is that all the items in `selectedItems` have a reference to their `parentItem`, which apparently isn't necessary when not opening the menu programmatically?

The "vanilla" example for cascading menu is so clean, you just supply a nested structure of items, and everything is just taken care of, but here, there is so much extra stuff to do, you have to create new items yourself, and replace the normal structure of items with this specially crafted one, and then remember to restore the normal structure again at the right moments (plural!).

Have you explored the possibility of keeping the simple "rootItems" and just telling the menu where to navigate to, somehow? I imagine you could provide it with a "path". I guess since it's all arrays, and `value` is optional, rather than a nice `'format.lists'`, you'd give it something like `'0.2'` or `'[0][2]'`, but still, it would be much less work on the consumer.

By the way, I just noticed this little curiosity, which I think highlights why it would be nice to not have the consumer do all this "manual labour":

https://github.com/Lundalogik/lime-elements/assets/154725/e698622a-73bf-4d0e-abed-f3ca118aab6d

Since I feel this feedback might be a little hard to hear, considering it might require some rework, I wan't to express that the cascading menu feature in itself is absolutely fantastic, and you've really done some great work here!

Is the ability to programmatically open the menu to any sub-menu critical for your use-case? If not, I would suggest simply leaving that functionality out for now, and add the feature once it's ready and stable.

src/components/menu/menu.tsx
adrianschmidt: I've created a new issue for this: https://github.com/Lundalogik/lime-elements/issues/2579

Resolving this thread.

src/components/menu/examples/menu-open-sub-menu-programatically.tsx
```diff
@@ -0,0 +1,103 @@
...
+                <limel-button
+                    label='Shortcut to "Bullets and numbering"'
+                    primary={true}
+                    onClick={this.buttonClick}
+                />
+            </div>
+        );
+    }
+
+    private buttonClick = () => {
```
adrianschmidt, line 80: I'm moving this comment since the thread it was in was resolved. Since the code has been updated, it's not describing the current state of the code 100 % correctly anymore, but the gist of it is still relevant:

---

Hmm… hang on… in the `CascadingMenuItems` imported from `item-constants.ts`, the sub-menu items actually do _contain_ the items included in that sub-menu. Why is it different here?

Another difference is that all the items in `selectedItems` have a reference to their `parentItem`, which apparently isn't necessary when not opening the menu programmatically?

The "vanilla" example for cascading menu is so clean, you just supply a nested structure of items, and everything is just taken care of, but here, there is so much extra stuff to do, you have to create new items yourself, and replace the normal structure of items with this specially crafted one, and then remember to restore the normal structure again at the right moments (plural!).

Have you explored the possibility of keeping the simple "rootItems" and just telling the menu where to navigate to, somehow? I imagine you could provide it with a "path". I guess since it's all arrays, and `value` is optional, rather than a nice `'format.lists'`, you'd give it something like `'0.2'` or `'[0][2]'`, but still, it would be much less work on the consumer.

By the way, I just noticed this little curiosity, which I think highlights why it would be nice to not have the consumer do all this "manual labour":

https://github.com/Lundalogik/lime-elements/assets/154725/e698622a-73bf-4d0e-abed-f3ca118aab6d

Since I feel this feedback might be a little hard to hear, considering it might require some rework, I wan't to express that the cascading menu feature in itself is absolutely fantastic, and you've really done some great work here!

Is the ability to programmatically open the menu to any sub-menu critical for your use-case? If not, I would suggest simply leaving that functionality out for now, and add the feature once it's ready and stable.

TommyLindh2: Yeah, I know that it isn't a very good design on this specific feature, YES it is critical in our implementation of the hierarchy picker. But maybe we could solve it from the outside some how. The reason the "path thing" doesn't work is because when we later enable lazy-loading which we already have in the next PR, the items isn't arrays anymore but functions that load the data when clicked. Because the limel-menu doesn't have all the data of all hierarchies it will be hard to define. We've tried to simplify it but haven't find a good way.

I have updated the example so it fetches the items from the rootItems instead of defining them again, that might make the example a little bit easier to read. If you have any ideas on how to make it easier I'm all ears, or if you want to talk and discuss it.

adrianschmidt: Ah, I understand. Yeah that seems like it would make the path approach more difficult.

Then I would suggest that we leave the functionality in, but actually remove the example, and add something to the docs about `selectedMenuItem` being an experimental property which shouldn't be used, and which we may remove without warning in the future. Then _you_ can use it, but we don't promise anyone else we'll keep it, if we find a better solution. And if we don't, we can make it permanent and add the example back at a later time.

Sound ok?

TommyLindh2: Yeah, sounds like a plan.
Then we can make the `parentItem` property on `MenuItem` private as well.

TommyLindh2:
@adrianschmidt why did you change the base to `next`? I created a separate branch to merge all these three so that we can test it all together in the end and not release parts of the changes in the menu but everything all at once.

src/components/menu/examples/menu-open-sub-menu-programatically.tsx
TommyLindh2: Done, make sure to verify that you agree with the changes, and if so, Resolve the conversation :)

adrianschmidt:
> why did you change the base to `next`? I created a separate branch to merge all these three so that we can test it all together in the end and not release parts of the changes in the menu but everything all at once.

Why does it need to be released all at once? We generally want to get each mergeable part merged as soon as possible, to avoid having giant PRs and risk massive merge conflicts that can be very difficult to resolve. The longer a branch stays unmerged, the greater the risk of conflicting changes being developed and merged, causing even further delays, increasing the risk of even more conflicting changes. We want all development to happen on branches that remain as small as possible and as short-lived as possible.

Are there changes in the upcoming PRs that will be breaking if this is merged first?

TommyLindh2:
> > why did you change the base to `next`? I created a separate branch to merge all these three so that we can test it all together in the end and not release parts of the changes in the menu but everything all at once.
> 
> Why does it need to be released all at once? We generally want to get each mergeable part merged as soon as possible, to avoid having giant PRs and risk massive merge conflicts that can be very difficult to resolve. The longer a branch stays unmerged, the greater the risk of conflicting changes being developed and merged, causing even further delays, increasing the risk of even more conflicting changes. We want all development to happen on branches that remain as small as possible and as short-lived as possible.
> 
> Are there changes in the upcoming PRs that will be breaking if this is merged first?

No I guess not, maybe it's just because they all belong together in my head. But there shouldn't be any issue releasing them one-by-one

src/components/menu/menu.types.ts
```diff
@@ -95,5 +95,5 @@ export interface MenuItem<T = any> {
      * :::
      * @private
      */
-    parentItem?: MenuItem<T>;
+    parentItem?: MenuItem;
```
TommyLindh2: It would be weird if the didn't but yes, there's nothing saying that they must.

src/components/menu/examples/menu-open-sub-menu-programmatically.tsx
```diff
@@ -31,7 +31,7 @@ export class MenuOpenSubMenuProgrammaticallyExample {
     private lastSelectedItem: string;
 
     @State()
-    private selectedMenuItem: MenuItem;
+    private currentSubMenu: MenuItem;
```
TommyLindh2, line 34: I really like this rename. We didn't come up with a good name.
But you did ❤️

src/components/menu/examples/menu-open-sub-menu-programmatically.tsx
```diff
@@ -0,0 +1,105 @@
...
+                        event: LimelMenuCustomEvent<MenuItem>
+                    ) => {
+                        if (!event.detail) {
+                            this.items = this.rootItems;
+                        }
+                    }}
+                    onCancel={() => {
+                        this.items = this.rootItems;
+                        this.openMenu = false;
+                    }}
```
adrianschmidt, line 69: We don't need to change anything here, since we're currently not using this example anyway, but we don't want people to use arrow-functions or `.bind()` in templates, because both of them cause a new function to be created every time the template is re-rendered, which eats up resources and performance. And since we don't want people to do that, we also really don't want to teach them to do it by doing it in the code examples 😊

src/components/menu/menu.tsx
```diff
@@ -202,10 +386,34 @@ export class Menu {
         this.open = !this.open;
     };
 
-    private handleSelect = (event: CustomEvent<MenuItem>) => {
-        event.stopPropagation();
-        this.select.emit(event.detail);
+    private handleSelect = async (
```
adrianschmidt, line 389: This doesn't need to be async yet. I'll fix 👍

src/components/menu/menu.types.ts
```diff
@@ -75,4 +77,21 @@ export interface MenuItem<T = any> {
      * Value of the menu item.
      */
     value?: T;
+
+    /**
+     * A way of defining a sub-menu for an item.
+     * Set it to an array of `MenuItem`s
+     */
+    items?: Array<MenuItem<T> | ListSeparator>;
```
adrianschmidt, line 85: It seems like the change in type of this property from `Array<MenuItem<T> | ListSeparator>` to `Array<MenuItem<T> | ListSeparator> | MenuLoader` (in the next PR) might actually be a breaking change* if we don't release those PRs together, so I'm holding off on merging this for a little bit anyway.

*) Evidence: see the change I had to make to the "open sub-menu programmatically"-example in the next PR: https://github.com/Lundalogik/lime-elements/pull/2577/files#diff-2add54a9e21a898b280c548832b0213ea57812aa63e2a29aecabbc0afd6389e8R87

image

src/components/menu/menu.tsx
```diff
@@ -150,28 +179,181 @@ export class Menu {
...
+            // No need to load a sub-menu.
+            this.currentSubMenu = null;
+            this.currentItems = null;
+            this.navigateMenu.emit(null);
+
+            return;
+        }
+
+        this.handleSelect(grandParent);
+    };
```
adrianschmidt, line 355: I don't understand why we're going to "grandParent" here? If we want to navigate one step up from "current", surely that must be "parent" then? 🤔

adrianschmidt (commented):
Oh, apparently I "started a review". It's so annoying that that's the default option when commenting 😣

src/components/menu/menu.tsx
```diff
@@ -150,28 +179,181 @@ export class Menu {
...
+            breadCrumbItems.unshift({
+                text: currentItem.text,
+                icon: {
+                    name: currentItem.icon,
+                    color: currentItem.iconColor,
+                },
+                menuItem: currentItem,
+            });
+            currentItem = currentItem.parentItem;
+        }
```
adrianschmidt, line 230: Having to redo this loop on every render feels a little painful. I'm not saying we necessarily need to optimise it _now_, but its good to be aware of whenever you are in any piece of code that is part of the render-loop, as that means the code will be run pretty incessantly, and will have a disproportionate performance impact.

src/components/menu/menu.tsx
```diff
@@ -150,28 +179,181 @@ export class Menu {
...
+        if (!breadCrumbItems.length) {
+            return;
+        }
+
+        return (
+            <limel-breadcrumbs
+                style={{
+                    'border-bottom': 'solid 1px rgb(var(--contrast-500))',
+                    'flex-shrink': '0',
+                }}
```
adrianschmidt, line 241: Styles belong in the stylesheet 🙂

adrianschmidt: Oh, or… these can't be in the stylesheet, because this stuff is lifted out to the portal, right?

That brings me to the point that this seems complex enough that it might warrant breaking it out into its own (internal) component. And it just so happens that doing so would let us avoid the inline styles as well!

src/components/menu/menu.tsx
```diff
@@ -150,28 +179,181 @@ export class Menu {
...
+                items={[
+                    {
+                        text: '',
+                        icon: {
+                            name: 'home',
+                        },
+                        type: 'icon-only',
+                    },
+                    ...breadCrumbItems,
+                ]}
```
adrianschmidt, line 264: If we optimise the breadcrumbs-creating loop so it doesn't have to be run on every single render, we should move this out too, for the same reason.

We should be able to update the breadcrumbs when navigation happens, and just keep that state between renders.

src/components/menu/menu.tsx
adrianschmidt: Ah, is it because we we're not going to the parent, we're selecting the grandparent item, which den loads the parent? Is that also why there is loading of items when moving outwards from a sub-menu to its parent menu (in the lazy-loading example in the next PR)?

adrianschmidt:
I'm going to log off for today after writing this comment, and perhaps it will feel different after resting my brain until tomorrow, but I'm worried about how complex this whole thing is. Like, what's in the first PR is fairly manageable, but adding in the lazy-loading it's starting to feel like it's melting my brain, and I haven't even gotten to the search-functionality yet.

Don't get me wrong, the functionality provided to the user is _excellent_! The problem I see is how are we ever going to maintain this code? Who's going to dare to make any changes to it?

I have a bunch of more or less vague ideas for how things might potentially be possible to simplify, but I'm not sure if they'll work or not, and I feel awful about once again being such an absolute pita before something you built can be merged 😭

src/components/menu/menu.tsx
TommyLindh2: Yes, exactly. We "simulate" a click on the grandParent to go to the parent so to say.

src/components/menu/menu.tsx
TommyLindh2: Yeah, before the logic for this was very different, but now when we have the `currentSubMenu` in a prop it's easier to do that. I can try it out.

src/components/menu/menu.tsx
```diff
@@ -150,28 +176,181 @@ export class Menu {
...
+            />
+        );
+    };
+
+    private renderMenuList = () => {
+        const items = this.visibleItems;
+
+        return (
+            <limel-menu-list
+                tabindex="1"
```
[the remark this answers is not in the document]
adrianschmidt: > The failing test seems to be failing because of this tabindex. (Although, I tried running the tests locally, both with this left in, and with it removed, and all tests passed regardless 🤷‍♂️)

Ah, no, I tried re-running the tests now, and they passed, so it was just a timing-thing or something I guess. The tests can be a little bit flaky sometimes… 😒

I'd still like to know what this does, or remove it 🙂

src/components/menu/menu.tsx
adrianschmidt: ⚡

adrianschmidt: Or, well, no, not quite zapped. I've just improved it a little bit. It still runs on every render.

adrianschmidt: Sorry, I'm too tired and don't read properly. This is indeed zapped, because, while the loop is still run on every render, this specific thing has been moved, so it's done next to the loop. So I'll resolve this after all. Sorry for spamming… 🙇‍♂️

src/components/menu/menu.tsx
```diff
@@ -150,28 +179,181 @@ export class Menu {
...
+                    if (!event.detail.menuItem) {
+                        this.currentItems = null;
+                        this.currentSubMenu = null;
+                        this.navigateMenu.emit(null);
+
+                        return;
+                    }
+
+                    this.handleSelect(event.detail.menuItem);
+                }}
```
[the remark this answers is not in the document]
adrianschmidt: ⚡

src/components/menu/examples/menu-open-sub-menu-programmatically.tsx
TommyLindh2: We add the file, so I think it's a good idea to fix it anyways. It's quick.

src/components/menu/menu.tsx
TommyLindh2: Yeah, me and Kia discussed that, but it would only be a very small wrapper of the limel-breadcrumbs which seems kind of risky for only 2 style properties.

src/components/menu/menu.tsx
```diff
@@ -200,8 +200,8 @@ export class Menu {
     }
 
     @Watch('open')
-    protected openWatcher() {
-        if (!this.open) {
+    protected openWatcher(newValue: boolean) {
+        if (!newValue) {
```
adrianschmidt: This change is to avoid any potential risk of getting out of sync.

src/components/menu/menu.tsx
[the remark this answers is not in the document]
adrianschmidt: We can remove it for now. It might be needed once search is added in a later PR.

adrianschmidt (approved):

Kiarokh:
😮 Merged! Wow... yeay....! 🥳 

Great job guys! ❤️

TommyLindh2:
It's only nested menu, without Search and lazy loading. So a little more work needs to be done :)

PR #2580, opened 2023-11-01
refactor(input-field): simplify the logic for rendering the portal

Kiarokh:
The portal is currently used to show auto-completion suggestions. But the plan is to use it for showing even the helper-line. This change will make the next thing easier.
I believe this change provides a better code organization, and achieves the same functionality
in terms of rendering the autocomplete list with the same conditions and handling the same interactions.
The core structure and logic of rendering the list, controlling its visibility, and handling user interactions remain consistent in both versions.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -802,10 +802,6 @@ export class InputField {
             return;
         }
 
-        return this.renderDropdown();
-    };
-
-    private renderPortal = (content = null) => {
         const dropdownZIndex = getComputedStyle(
             this.limelInputField
         ).getPropertyValue('--dropdown-z-index');
@@ -827,16 +823,10 @@ export class InputField {
                     }}
                     onDismiss={this.handleCloseMenu}
                 >
-                    {content}
+                    {this.renderListResult()}
                 </limel-menu-surface>
             </limel-portal>
         );
-    };
-
-    private renderDropdown = () => {
-        const content = this.renderListResult();
-
-        return this.renderPortal(content);
     };
 
     private renderListResult = () => {


src/components/input-field/input-field.tsx
```diff
@@ -827,32 +823,23 @@ export class InputField {
...
-        return this.renderPortal(content);
-    };
-
     private renderListResult = () => {
-        const filteredCompletions: ListItem[] = this.filterCompletions(
-            this.value
-        );
-        if (!filteredCompletions || filteredCompletions.length === 0) {
-            return null;
+        if (!this.filterCompletions(this.value)) {
```
vicgeralds, line 833: Do we want to show an empty `limel-list` when the filtered completions are empty? This seems like an unintentional change, not refactoring.

Kiarokh: This doesn't return an empty limel-list. This prevents the limel-list from being rendered, if the search in auto-completions is not finding anything (both the old and the refactored code). 

However, they are both bad anyway. Because we should basically not render the portal at all, if:
1. there are no auto complete items
2. there are no filtered items (search results) 
3. And I think the check for `if (this.type === 'textarea' ` then return is also completely unnecessary.

 
I don't like this original code at all. 
I would have written it in a different way.

However, what is bad now, will be good for my next PR. 
I want to render the limel-helper-line in the portal. So these bad choices will be useful later. ;P

vicgeralds (approved):

Issue #2583, opened 2023-11-02
Input field: the empty auto suggestion list doesn't really dissapear


Kiarokh:
## Current behavior

https://github.com/Lundalogik/lime-elements/assets/35954987/01754918-a756-4209-a1c6-c333f15952de

## Expected behavior

That shadow shouldn't be there anymore. It looks like a bug.

We should not return the limel-portal at all, when the filtered results (search results) are empty.

## Environment

- **lime-elements** version: 
- Framework used: 
- Logs:

Kiarokh:
related to https://github.com/Lundalogik/lime-elements/issues/3057

PR #2589, opened 2023-11-06
only render portal when the element is actually needed to be displayed

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/3703
fix https://github.com/Lundalogik/lime-elements/issues/2583

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -802,6 +802,13 @@ export class InputField {
             return;
         }
 
+        const filteredCompletions: ListItem[] = this.filterCompletions(
+            this.value
+        );
+        if (!filteredCompletions || filteredCompletions.length === 0) {
+            return null;
+        }
+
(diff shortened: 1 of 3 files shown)


john-traas:
After investigating the side effects of these proposed changes, it seems to unlock potential issues and bugs that could have wide reaching effects. Much as it might be inefficient to have so many instances of the portal open, it is the underlying structure on which much of the web client is built. 

With these proposed changes errors occur in other components specifically with the intersection observer. I think we're opening a potential rabbit hole here and I don't believe the rewards are worth the risk.

PR #2590, opened 2023-11-06
fix(menu): allow clicks for permitted elements from open menu surface

john-traas:
When the click event originates from a component that is allowed, such as a chip-set, then the menu should allow the click. It should only kill click events that originate from elements that are not allowed.

Fixes [Table filters bugs out immediately](https://github.com/Lundalogik/crm-feature/issues/3678)

How to test:
You should be able to remove chips now even if the menu is open. 

https://github.com/Lundalogik/lime-elements/assets/122509723/3675a66b-e9c4-40e7-91e2-f25ac6819d5e

Also manually test that other components using the menu-surface still work as intended. I've manually tested but it can't hurt to test a bit more.


LawrenceBorst (approved):
Tested it manually. It works like a charm

src/components/menu-surface/menu-surface.tsx
```diff
@@ -126,7 +126,9 @@ export class MenuSurface {
         }
 
         this.dismiss.emit();
-        this.preventClickEventPropagation();
+        if (!this.allowClicksElement) {
+            this.preventClickEventPropagation();
```
adrianschmidt, line 130: This doesn't look like the right solution.

We already check (or try to check) if the click originated in the allowed element, and if it did, we return early:

Screenshot of the code from line 107 until and including this code change

Apparently, that code doesn't work correctly. Unless what you are really trying to do is something other than allow clicks on the supplied "allowed element"?

adrianschmidt (dismissed):

src/components/menu-surface/menu-surface.tsx
john-traas: Now that I look further into it, the original code works correctly except in the case of `limel-chipset`. Where I think this is falling down is, when a chip is clicked the `elementPath` does not include the `allowClicksElement`. The `allowClicksElement` is there and is correct, but the path shows simply as an array `[]` and this is causing the guard on `line 123` to be bypassed. 

At this point, `allowClicksElement` is `true`, but `clickedInAllowedElement` is `false`, the function does not return, and hence the issue is caused with the `preventClickEventPropogation` function being invoked. 

I guess perhaps I should look into the chipset component instead?

adrianschmidt: > I guess perhaps I should look into the chipset component instead?

Yeah, if we either could fix the problem there, or at least find out enough to somehow make some kind of targeted workaround, that would be great!

john-traas: I'll give it a go

john-traas: That was a great call @adrianschmidt. The issue was still in the menu-surface, but higher up the chain and simply due to event.path not being the right way to get an element path from the shadowdom.

adrianschmidt: Nice!

You forgot to remove the old attempt at a fix though 😊

john-traas: Ahahah, oops 😆

adrianschmidt:
@john-traas "table filter" isn't a concept that exists in lime-elements. I think this should be a `fix` and that the scope should be `menu` 🙂

adrianschmidt (approved):
I haven't tested that this actually solves the problem in lime-webclient, but it certainly solves a bug here in lime-elements, so I'm approving it for that reason.

@LawrenceBorst, can you check that it's still solving the problem in the web client? If not, please re-open the issue, if it has been closed (I'm setting this to automerge).

PR #2591, opened 2023-11-06
Transition helper line

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/3537
closes: https://github.com/Lundalogik/crm-feature/issues/3334

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


Kiarokh:
There will be a **know** issue in **Input Field > Auto complete** and in **Select**. The issue is that _when there are auto completion suggestions_ or _the select dropdown menu_ (basically, when the portal is open), the helper line is not displayed directly. 

But I consider it as "very unimportant", and we can ignore it.

The issue is that when there is a helperText && there are suggestions, the portal (with the list of suggestions for example) is rendered before the helper text animation is run. This means the height of the component is not yet "finalized" and the portal doesn't know it yet. The portal doesn't know that it should wait for the helperText to get rendered and get done with increasing the accumulated height of the component. Therefore, the portal opens right below the input-field, and covers the helper-line.

However, this is not a serious problem, because:

- **For the auto completions:** as soon as the user types something (or if they scroll), the portal recalculates its location and aligns itself correctly. So the users will see the helper text immediately after that.

https://github.com/Lundalogik/lime-elements/assets/35954987/935dbeab-9496-4ec9-9179-9e5d4f9e5feb

- **For the Select:** when you have selected an option, the dropdown closes and you see the helper text anyways.
- 
Therefore I believe that we should not try to fix this issue by setting a timeout or similar, and just move on.

LucyChyzhova (approved):
Looks really good! 💫

PR #2595, opened 2023-11-07
feat(chip-set): enable consumers to make it `invalid`

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -92,6 +92,13 @@ export class ChipSet {
     public readonly: boolean = false;
 
     /**
+     * Set to `true` to indicate that the current value of the input field is
+     * invalid.
+     */
+    @Prop({ reflect: true })
+    public invalid = false;
+
+    /**
      * For chip-sets of type `input`. Value to use for the `type` attribute on the
      * input field inside the chip-set.
      */
@@ -466,6 +473,15 @@ export class ChipSet {
     }
 
     private isInvalid() {
+        if (this.readonly) {
+            // A readonly field can never be invalid.
+            return false;
+        }
+
+        if (this.invalid) {
+            return true;
+        }
+
         if (!this.required) {
             return false;
         }
--- a/src/components/file/examples/file.tsx
+++ b/src/components/file/examples/file.tsx
@@ -21,6 +21,9 @@ export class FileExample {
     @State()
     private readonly = false;
 
+    @State()
+    private invalid = false;
+
     public render() {
         return [
             <limel-file
@@ -30,6 +33,7 @@ export class FileExample {
                 value={this.value}
                 disabled={this.disabled}
                 readonly={this.readonly}
+                invalid={this.invalid}
             />,
             <limel-example-controls>
                 <limel-checkbox
@@ -46,6 +50,11 @@ export class FileExample {
                     checked={this.required}
                     label="Required"
                     onChange={this.setRequired}
+                />
+                <limel-checkbox
+                    checked={this.invalid}
+                    label="Invalid"
+                    onChange={this.setInvalid}
                 />
             </limel-example-controls>,
             <limel-example-value value={this.value} />,
@@ -70,5 +79,10 @@ export class FileExample {
     private setReadonly = (event: CustomEvent<boolean>) => {
         event.stopPropagation();
         this.readonly = !!event.detail;
+    };
+
+    private setInvalid = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.invalid = event.detail;
     };
 }
--- a/src/components/chip-set/examples/chip-set-input.tsx
+++ b/src/components/chip-set/examples/chip-set-input.tsx
@@ -26,6 +26,9 @@ export class ChipSetInputExample {
     private readonly: boolean = false;
 
     @State()
+    private invalid: boolean = false;
+
+    @State()
     private disabled: boolean = false;
 
     @State()
@@ -59,6 +62,7 @@ export class ChipSetInputExample {
                 value={this.value}
                 required={this.required}
                 readonly={this.readonly}
+                invalid={this.invalid}
                 disabled={this.disabled}
                 leadingIcon={this.hasLeadingIcon ? 'search' : null}
                 maxItems={this.maxItems}
@@ -97,6 +101,11 @@ export class ChipSetInputExample {
                     label="Required"
                     onChange={this.setRequired}
                     checked={this.required}
+                />
+                <limel-checkbox
+                    label="Invalid"
+                    onChange={this.setInvalid}
+                    checked={this.invalid}
                 />
                 <limel-checkbox
                     label={'Leading icon'}
@@ -162,6 +171,10 @@ export class ChipSetInputExample {
 
     private setRequired = (event: CustomEvent<boolean>) => {
         this.required = event.detail;
+    };
+
+    private setInvalid = (event: CustomEvent<boolean>) => {
+        this.invalid = event.detail;
     };
 
     private setEmptyInputOnBlur = (event: CustomEvent<boolean>) => {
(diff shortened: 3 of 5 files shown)


LawrenceBorst (dismissed):
This looks good. We need to make a future issue for the "red border" :)

src/components/chip-set/chip-set.tsx
```diff
@@ -466,6 +473,15 @@ export class ChipSet {
...
     private isInvalid() {
+        if (this.readonly) {
+            // A readonly field can never be invalid.
+            return false;
+        }
+
+        if (this.invalid) {
+            return true;
+        }
+
```
Kiarokh, line 484: if you wonder about why, it could be good to look at:

https://github.com/Lundalogik/lime-elements/blob/5744bb981d253866cdc7553fb3f6e01ab2ae3296/src/components/input-field/input-field.tsx#L544C9-L544C9

src/components/file/examples/file.tsx
```diff
@@ -71,4 +80,9 @@ export class FileExample {
         event.stopPropagation();
         this.readonly = !!event.detail;
     };
+
+    private setInvalid = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.invalid = !!event.detail;
```
Befkadu1, line 86: `event.detail` is already boolean
```suggestion
        this.invalid = event.detail;
```

src/components/chip-set/chip-set.tsx
```diff
@@ -466,6 +473,15 @@ export class ChipSet {
     }
 
     private isInvalid() {
+        if (this.readonly) {
+            // A readonly field can never be invalid.
+            return false;
+        }
+
+        if (this.invalid) {
```
Befkadu1, line 481: Can disabled field be invalid?

Kiarokh: Good question. I don't know. I don't see any logic handling this scenario in the input-field. So I guess it can be invalid yes. Probably it won't show anything in the UI when it's disabled && invalid. What do you think?

vicgeralds: seems like the disabled styling overrides the invalid styling, but I don't think that's something we need to be concerned with here.

If it's set as both invalid and disabled, I see no reason why `isInvalid()` shouldn't return true.

src/components/chip-set/chip-set.tsx
vicgeralds: I still don't understand why we need a special case for readonly. It would be weird to show a readonly field as invalid to the user, because they can't change it to be valid, but technically it seems possible to me. Readonly and invalid would indicate something that's outside the control of the user is in an inconsistent state.

vicgeralds (approved):
I approve this because it's consistent with input-field, but I still don't understand the need for the readonly special case 😄

src/components/chip-set/chip-set.tsx
Kiarokh: a `readonly` field is a stupid/smart internal notion that we have invented, due to CRM. An input field which is inherently interactive cannot be `readonly`. It doesn't make sense. A readonly field is basically something that is for visualization of data. That's it. It's a dead component, non-interactive. 

So when something is non-interactive and dead, the system cannot tell the user to fix it, because it is now `invalid`! Users cannot do anything about the validity of the field. 😅 

Hope I could explain this in short @vicgeralds

vicgeralds: > So when something is non-interactive and dead, the system cannot tell the user to fix it, because it is now invalid! Users cannot do anything about the validity of the field. 😅

Yes, sounds reasonable. That still doesn't fully motivate the special case inside `isInvalid()` though. 🙂 

Let me put it this way: would it make any difference if we removed these lines?

```js
        if (this.readonly) {
            // A readonly field can never be invalid.
            return false;
        }
```

If that would mess up styling, or causes some other problem when rendering, fine. I think I would still prefer just handling that in css if possible. 🙂

simonedeblasio: Just some food for thought:
Lets say the customers have logic on their deals. Any deal in the segment X, needs to have a `Total Value` higher than `100 000`. The `Total Value` is never set by the user, it is calculated from the `Price per Unit` x `Amount`. So once the user tries to save the custom validation logic stops them. 

Which field should we then show as the invalid one? In my opinion it can only be the `Total Value` field.

src/components/chip-set/chip-set.tsx
Befkadu1: This argument can be used here as well maybe?
> So when something is non-interactive and dead, the system cannot tell the user to fix it, because it is now invalid! Users cannot do anything about the validity of the field.

src/components/chip-set/chip-set.tsx
vicgeralds: > Which field should we then show as the invalid one? In my opinion it can only be the `Total Value` field.

In other words, a readonly field. Good point!

Another approach is to make the section of group of fields that has an error appear as invalid. But currently, a validation error has to belong to a field, from what I understand, and a limeobject is just a flat collection of fields.

simonedeblasio: Agree, and you could in this case raise the errors on both the `Price per Unit` and `Amount`, but that is not always intuitive for the user.

Kiarokh: I understand what you mean. 

I don't want to be religious about it, but I still disagree (or agree with what I said 😅 ).

But for interactive components (e.g. input elements) `invalid` is a system/developer thing really, which also manifests itself to the user. It is **NEVER** good enough however, because merely by making the field 🔴 the UI will be unable to communicate to the user "what is exactly wrong with this field?". 

In a good design, the system / developer / designer will use a proper helper text to explain what is wrong. This helper text is not the helpertext that usually is provided with the input field. It is rather the description of the problem. Are we doing that now anywhere in the web client? I don't think so.

But that is when the field is interactive. The invalid state can become valid, by interacting with the field **_itself_**!!!! Not by doing something else somewhere else in the UI.

For what you explained @simonedeblasio , the filed is non-interactive and it is just a visualization of data. So we can't use the same convention and make it "invalid" and expect the user to fix it. The user cannot do anything about that particular field itself, and their attempts of fixing the problem should be in other input field. 

I would argue that the problem that you describe has a complexity that perhaps needs a set of UI changes, to communicate or guide the user about how they can fix the invalid state. Maybe by showing a snackbar or modal with a more clear and more elaborated message for the user, or by setting some sort of highlight on all fields that are involved in the validation. 

In your case, the `Price per Unit` & `Amount` are the problems, not the `total value`. 

Can the total value become 🔴 as well? Well, yes, maybe. 
Is it enough for the user to know what to do? I don't think so. But maybe?
Are there better ways? I think so.
Should we allow readonly fields to become red and show a validation error? I'm not convinced yet, but maybe.

This is how far my rationale reaches right now. Open for suggestions or counter arguments.

src/components/chip-set/chip-set.tsx
vicgeralds: Good find. I am convinced that setting something as both disabled an invalid is a mistake or the result of some inconsistent state.

Should we try to hide the mistake or inconsistency in lime-elements, without clear documentation?

Or is there some valid use case when a field has errors, and it should still be disabled? I guess it won't be possible to save a form then, and that would be even more confusing if we hide the error.

Kiarokh: I think if a field can be invalid, then it shouldn't be disabled then. Doesn't it look like a bad design, done by the consumer?

Should the invalid styles be applied to a disabled && invalid field? Yes. Because then we show them that they've done something stupid, or something is effed up somewhere along the way. I think it's good.

Note that `readonly` state is another story. A disabled field can become enabled, if user does something. But a readonly filed can never become enabled.

Issue #2597, opened 2023-11-08
limel-button: Disabling vs Hiding  guideline 


LucyChyzhova:
## New feature motivation

https://github.com/Lundalogik/lime-elements/assets/50618208/2233ede1-f774-49bc-96b6-3fbb7ab496dc

PR #2598, opened 2023-11-08
feat(button): add guidelines for `disable` and `hide` buttons

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/2597

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/guides.ts
+++ b/guides.ts
@@ -13,6 +13,7 @@ export default [
             'src/design-guidelines/size/size-rhythms.md',
             'src/design-guidelines/action-buttons/action-buttons.md',
             'src/design-guidelines/declutter/declutter.md',
+            'src/design-guidelines/disabled-hidden/disabled-hidden.md',
         ],
     },
 ] as Guide[];
--- a/src/components/button/examples/button-disabled.tsx
+++ b/src/components/button/examples/button-disabled.tsx
@@ -2,6 +2,9 @@ import { Component, h } from '@stencil/core';
 
 /**
  * Disabled
+ * :::note
+ * Discover when to utilize the disabled state and when it is preferable to hide a button by reading our guidelines [Disabled vs. Hidden](#/DesignGuidelines/disabled-hidden.md/).
+ * :::
  */
 @Component({
     tag: 'limel-example-button-disabled',
--- a/src/design-guidelines/disabled-hidden/disabled-hidden.md
+++ b/src/design-guidelines/disabled-hidden/disabled-hidden.md
@@ -1,0 +1,3 @@
+# Hiding or Disabling an interactive element in User Interfaces
+
+<limel-example-button-disabled-vs-hidden></limel-example-button-disabled-vs-hidden>
--- a/src/design-guidelines/disabled-hidden/examples/button-disabled-vs-hidden.scss
+++ b/src/design-guidelines/disabled-hidden/examples/button-disabled-vs-hidden.scss
@@ -1,0 +1,50 @@
+@import '../../../design-guidelines/_generic-example-styles/do-dont';
+
+.do,
+.do-not {
+    width: 30rem;
+    margin: 1rem 0 2rem 0;
+    background-color: rgb(var(--contrast-200));
+
+    limel-button {
+        padding: 0 0 1rem 0.5rem;
+    }
+
+    .relevant-button {
+        margin-top: 1rem;
+    }
+
+    .assign {
+        padding-bottom: 0;
+    }
+}
+
+#tooltip-example {
+    margin-top: 1rem;
+    width: 5rem;
+}
+
+#tooltip {
+    cursor: help;
+}
+
+.button-icon {
+    display: flex;
+    align-items: center;
+
+    .disabled-button {
+        padding-right: 0.5rem;
+        margin-top: 1rem;
+    }
+}
+
+.split-example,
+.relevan-buttons-example {
+    display: flex;
+    gap: 5rem;
+
+    limel-split-button {
+        margin-top: 1rem;
+        padding: 0 0 1rem 0.5rem;
+    }
+}
--- a/src/design-guidelines/disabled-hidden/examples/disabled-vs-hidden.tsx
+++ b/src/design-guidelines/disabled-hidden/examples/disabled-vs-hidden.tsx
@@ -1,0 +1,206 @@
+import { Component, h, State } from '@stencil/core';
+import { ListSeparator, MenuItem } from 'src/interface';
+
+@Component({
+    tag: 'limel-example-button-disabled-vs-hidden',
+    shadow: true,
+    styleUrl: 'button-disabled-vs-hidden.scss',
+})
+export class ButtonDisabledVsHideExample {
+    @State()
+    private value: boolean = false;
+
+    @State()
+    private clicked: boolean = false;
+
+    @State()
+    private clickedDontExample: boolean = false;
+
+    @State()
+    public required: boolean;
+
+    private items: Array<ListSeparator | MenuItem> = [
+        { text: 'Save as', icon: 'save_as' },
+    ];
+    private disabledItem: Array<ListSeparator | MenuItem> = [
+        { text: 'Save as', icon: 'save_as', disabled: true },
+    ];
+
+    public render() {
+        return [
+            <h3>Disabled vs. Hidden</h3>,
+            <p>
+                When should an interactive element –like a button– be "Disabled"
+                and when should we avoid displaying it? This is a crucial
+                question in user interface design, and the choice between the
+                two should always be context-dependent. These guidelines provide
+                insights into when to apply each approach to optimize user
+                experiences.
+                <br />
+                <br />
+                <b> Disabling an interactive element:</b>
+            </p>,
+            <p>
+                Use button disabling when the action the button represents is
+                not currently available but may become available in the future.
+                For example, you can disable a "Submit" button until all
+                required form fields are filled out. Keep in mind that in such
+                cases, users should be able to do something to enable the
+                <code>disabled</code> element!
+            </p>,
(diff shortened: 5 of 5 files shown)


src/components/button/examples/button-disabled-vs-hidden.tsx
```diff
@@ -0,0 +1,160 @@
...
+    ];
+
+    public render() {
+        return [
+            <h3>Disabled vs. Hidden</h3>,
+            <p>
+                Effective use of 'Disabled' and 'Hidden' buttons is crucial in
+                user interface design, and the choice between the two should
+                always be context-dependent. These guidelines provide insights
+                into when to apply each approach to optimize user experiences.
```
Kiarokh, line 34:
```suggestion
                When should an interactive element –like a button– be "Disabled" and
                when should we avoid displaying it? This is a crucial question in
                user interface design, and the choice between the two should
                always be context-dependent. These guidelines provide insights
                into when to apply each approach to optimize user experiences.
```

src/components/button/examples/button-disabled-vs-hidden.tsx
```diff
@@ -0,0 +1,160 @@
...
+                <br />
+                <br />
+                <b> Disabling a Button:</b>
```
Kiarokh, line 37:
```suggestion
                <b>Disabling an interactive element:</b>
```

src/components/button/examples/button-disabled-vs-hidden.tsx
```diff
@@ -0,0 +1,160 @@
...
+                <div id="tooltip-example">
+                    <limel-button
+                        class="disabled-button"
+                        disabled
+                        icon="phone"
+                        label="Call"
+                    />
+                </div>
+                <limel-tooltip
+                    label="Select recipient to make a call"
```
Kiarokh, line 78:
```suggestion
                    label="Select a recipient to make a call"
```

src/components/button/examples/button-disabled-vs-hidden.tsx
```diff
@@ -0,0 +1,160 @@
...
+                />
+                <limel-button
+                    label="Submit"
+                    disabled={this.value ? false : true}
+                />
+            </div>,
+            <p>
+                It's helpful when you want to communicate to the user why they
+                can't perform a certain action. Disabled buttons could have a
+                tooltip or message explaining why they are disabled.
```
Kiarokh, line 61:
```suggestion
                Simply showing a disabled element in the user interface might not 
                be enough for the user to realize what the should do to enable it.
                Therefore, it's very helpful to communicate to the user why they
                can't perform that certain action. Disabled buttons for instance, 
                could have a tooltip or message explaining why they are disabled.
```

src/components/button/examples/button-disabled-vs-hidden.tsx
```diff
@@ -0,0 +1,160 @@
...
+                </div>
+                <limel-tooltip
+                    label="Select recipient to make a call"
+                    elementId="tooltip-example"
+                />
+            </div>,
+            <p>
+                It can be suitable for indicating to the user that they need to
+                complete a certain step or meet specific conditions before
+                proceeding.
```
Kiarokh, line 85:
```suggestion
                Another idea could be to display a more noticeable visual element 
                next to the disabled element, which hints about an explanation 
                or further information.
                For example, indicating to the user that they need to
                complete a certain step or meet specific conditions before
                proceeding in this case.
```

src/components/button/examples/button-disabled-vs-hidden.tsx
```diff
@@ -0,0 +1,160 @@
...
+                    <limel-button
+                        class="disabled-button"
+                        disabled
+                        icon="plus_math"
+                        label="Add recipient"
+                    />
+                    <limel-icon name="info" id="tooltip" size="x-small" />
+                </div>
+                <limel-tooltip
+                    label="To enable the feature call our support"
```
Kiarokh, line 102:
```suggestion
                    label="To activate this feature, call our support!"
```

src/components/button/examples/button-disabled-vs-hidden.tsx
```diff
@@ -0,0 +1,160 @@
...
+                    elementId="tooltip"
+                />
+            </div>,
+            <h3>Hiding a Button:</h3>,
```
Kiarokh, line 106:
```suggestion
            <h3>Hiding an interactive element:</h3>,
```

src/components/button/examples/button-disabled-vs-hidden.tsx
```diff
@@ -0,0 +1,160 @@
...
+            <p>
+                Hide the button if it's not relevant to a context. If the user
+                is already assigned there is no sense in having the assign
+                button, it is better to show the <b> unassign</b> button
+                instead.
```
Kiarokh, line 111:
```suggestion
                Sometimes, displaying a disabled element does not make sense.
                This could be of course due to various reasons, 
                but a common scenario is coexistence of another "enabled" element 
                that does the opposite of what the disabled element does.
                
                In this example, there is no point to show two buttons, one for 
                "assigning" and one for "un-assigning". 
                The "Assign" button should be hidden if it's not relevant to a context. 
                If the user is already assigned there is no sense in having the assign
                button, it is better to show the <b>Unassign</b> button
                instead.
```

src/components/button/examples/button-disabled-vs-hidden.tsx
```diff
@@ -0,0 +1,160 @@
...
+            </p>,
+            <div class="do">
```
Kiarokh, line 113: I think you can have a <div class="dont"> example here as well, which shows both Assign and Unassign buttons.

src/components/button/examples/button-disabled-vs-hidden.tsx
```diff
@@ -0,0 +1,160 @@
...
+    private value: boolean = false;
+
+    @State()
+    private clicked: boolean = false;
+
+    @State()
+    public required: boolean;
+
+    private items: Array<ListSeparator | MenuItem> = [
+        { text: 'Edit', icon: 'save_as' },
```
Kiarokh, line 24: including an Edit option in the split button which is for Save is not a good design. I suggest creating a <div class="dont"> example in which the two buttons are available, but one of them is disabled.

Kiarokh: Also "Delete" would be a better example for what an admin may be allowed to do, but not a normal user

src/components/button/examples/button-disabled-vs-hidden.tsx
LucyChyzhova: Do you mean the "don't" example?

src/components/button/examples/button-disabled-vs-hidden.tsx
LucyChyzhova: > including an Edit option in the split button which is for Save is not a good design. I suggest creating a
> 
> example in which the two buttons are available, but one of them is disabled.

I think `Save` and `Save as` actually would be good show here. Especially since it's our real example from the webclient

LucyChyzhova: > I suggest creating a
> 
> example in which the two buttons are available, but one of them is disabled.

But what is the reason to have a disabled one, since for the user it will never be an active(or not disabled) button?

src/components/button/examples/button-disabled-vs-hidden.tsx
LucyChyzhova: I'm not sure this rewriting will make any difference :) It sounds the same to me. Or what was the intention of your version?😀

src/components/button/examples/button-disabled-vs-hidden.tsx
Kiarokh: to compare the two different **good** and **bad**, side by side.

src/components/button/examples/button-disabled-vs-hidden.tsx
Kiarokh: the first sentence 😊 
this is not a principle that applies on to "buttons". This is about anything that is interactive, and receives an input from the users. Right? Most of the times, it's about buttons. But we don't have "hidden" buttons really. Hidden is not a property of a button component. 

This guide is about: **_Why you sometimes should NOT simply disable something, and instead choose not to show it._** 

This is also why using "hidden" is a little bit like a miscommunication of the purpose. It's like setting `display: none;` on an element; while the elements shouldn't even have been there in the UI.

LucyChyzhova: > this is not a principle that applies on to "buttons".

The idea was to apply this guidance to the button case specifically, that’s why I used only different types of buttons and not other UI elements.

Of course, `disable/hide` could be applied to anything and be more generic.  The reason, why I stick to buttons is because I got the concrete question from our consultants about when to disable and when to hide buttons.  This is quite often confusions from their side. I think we should address it here and be specific.

Kiarokh: I understand that. 😊  Did you read this comment? https://github.com/Lundalogik/lime-elements/pull/2598#pullrequestreview-1722529492

LucyChyzhova: fixed

Kiarokh (approved):
👏 beautiful work!

PR #2601, opened 2023-11-08
Make non-empty and optional fields with errors invalid

vicgeralds:
This makes sure fields that have a value and an error are rendered as invalid.

Also, empty fields that are not required are always rendered as invalid when they have errors.

Fixes: Lundalogik/crm-feature#3478

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/fields/schema-field.ts
+++ b/src/components/form/fields/schema-field.ts
@@ -108,13 +108,6 @@ export class SchemaField extends React.Component<FieldProps> {
         this.handleChange = this.handleChange.bind(this);
         this.handleCustomComponentChange =
             this.handleCustomComponentChange.bind(this);
-        this.initState();
-    }
-
-    private initState() {
-        if (this.hasValue()) {
-            this.state.modified = true;
-        }
     }
 
     private hasValue() {
@@ -144,11 +137,7 @@ export class SchemaField extends React.Component<FieldProps> {
         const { modified } = this.state;
         const { errorSchema } = this.props;
 
-        if (!modified) {
-            return false;
-        }
-
-        return !isEmpty(errorSchema);
+        return (modified || this.hasValue()) && !isEmpty(errorSchema);
     }
 
     private isRequired() {
--- a/src/components/form/fields/schema-field.ts
+++ b/src/components/form/fields/schema-field.ts
@@ -137,7 +137,10 @@ export class SchemaField extends React.Component<FieldProps> {
         const { modified } = this.state;
         const { errorSchema } = this.props;
 
-        return (modified || this.hasValue()) && !isEmpty(errorSchema);
+        return (
+            (modified || this.hasValue() || !this.isRequired()) &&
+            !isEmpty(errorSchema)
+        );
     }
 
     private isRequired() {
--- a/src/components/form/adapters/widget-adapter.ts
+++ b/src/components/form/adapters/widget-adapter.ts
@@ -31,13 +31,6 @@ export class LimeElementsWidgetAdapter extends React.Component {
         super(props);
 
         this.handleBlur = this.handleBlur.bind(this);
-        this.initState();
-    }
-
-    private initState() {
-        if (this.hasValue()) {
-            this.state.modified = true;
-        }
     }
 
     private hasValue() {
@@ -71,11 +64,7 @@ export class LimeElementsWidgetAdapter extends React.Component {
         const { modified } = this.state;
         const { rawErrors } = this.props.widgetProps;
 
-        if (!modified) {
-            return false;
-        }
-
-        return !!rawErrors;
+        return !!rawErrors && (modified || this.hasValue());
     }
 
     private isRequired() {
--- a/src/components/form/adapters/widget-adapter.ts
+++ b/src/components/form/adapters/widget-adapter.ts
@@ -64,7 +64,9 @@ export class LimeElementsWidgetAdapter extends React.Component {
         const { modified } = this.state;
         const { rawErrors } = this.props.widgetProps;
 
-        return !!rawErrors && (modified || this.hasValue());
+        return (
+            !!rawErrors && (modified || this.hasValue() || !this.isRequired())
+        );
     }
 
     private isRequired() {


src/components/form/fields/schema-field.ts
```diff
@@ -144,11 +137,7 @@ export class SchemaField extends React.Component<FieldProps> {
         const { modified } = this.state;
         const { errorSchema } = this.props;
 
-        if (!modified) {
-            return false;
-        }
-
-        return !isEmpty(errorSchema);
+        return (modified || this.hasValue()) && !isEmpty(errorSchema);
```
Kiarokh: I don't remember the fulls story, and I don't know if what you did here is what we intentionally "did **not** want" to have. But I tell you the story, and the one who reviews this PR can keep it in mind.

- Let's say the admin changes a field that has always been optional to `required`; or adds a new required field to the card. 
- There are potentially already many cards on which those required fields are empty, and they are saved in the database.
- the user opens one of those cards that have required but empty fields. S/he changes something else in the card, and wants to save it. 
- S/he should be able to save and continue, even if there is an invalid field in the card. The UI should also not complain about the invalid field in the card neither, and should not make it 🔴 for example.
- Same thing if an integration that is trying to update a card.

The story of forms on Cards is however different than the use-case of Create New dialogs. The behavior should be different in create new dialogs. You cannot have invalid fields when you are creating.

Kiarokh: I guess @jgroth and @adrianschmidt remember the reasons better than, and they'r certainly better judges for this change.

vicgeralds: I haven't changed the designed behaviour as you describe it, from what I can tell.

I only (intentionally) change the behaviour when a field either has a value or is optional. @Kiarokh your story is about empty required fields.

> But they should of course be rendered as invalid if you've tried to save a create-form with them empty, even if you've never touched them.

This isn't something that I tried to fix in this PR, but I agree. We work around this problem by disabling the button until the form is valid, so the user needs to find all the required fields.

Kiarokh: > We work around this problem by disabling the button until the form is valid

From what [the issue](https://github.com/Lundalogik/crm-feature/issues/3478) was explaining, the "button" (the <kbd>Create</kbd> button in the Create New dialog) was not `disabled`, when the required fields were empty or had validation errors.

vicgeralds: > From what [the issue](https://github.com/Lundalogik/crm-feature/issues/3478) was explaining, the "button" (the Create button in the Create New dialog) was not `disabled`, when the required fields were empty or had validation errors.

I only see one required field in that issue (First name), and the save button is enabled because it is not empty.

We can't know what fields are "required" using backend validation. From my point of view, they are not required, they are invalid using custom validation. Which means they are what I call optional, so they should always be shown as invalid if they have errors with the changes in this PR.

adrianschmidt: Sorry if I confused things 🙈

vicgeralds: > Sorry if I confused things 🙈

No, I think you described the problem and desired behaviour really well.

> it's more important that they show up as invalid after saving. If that means they have to show as invalid from the moment you open the create-dialog, then so be it. imo.

So the remaining issue is making required fields be shown as invalid after saving. I didn't focus on fixing that specifically in this PR, because it doesn't seem to make any difference as long as we're also disabling the save button.

I also think I agree that it's better to always show them as invalid _if the save button is enabled_ while the form is invalid, even if it's annoying to have all required fields red before you had the chance to fill them. But then there's a conflict of interest, should we focus on make forms in Lime CRM work as designed, or should we make sure `limel-form` work better when used in a different way. Looking at the examples, they suggest that save button should be disabled until the form is valid, so I'm inclined to think this "different way" can be seen as a new feature that we probably want to support, so a "fix" is not urgently needed.

I would not say that this is "a combination of behaviours that is hard to implement". The only missing part is that `limel-form` needs to know whether it is in a submitted/saved state. Since it's decoupled from any submit button,  I think we could just add a boolean property, for example `saved` and leave it to the consumer set after saving. Maybe "saved" is confusing if we failed to save? Better to call it "validated"?

Kiarokh: I'm afraid fixing this might be really hard.

Honestly, I'm not a big fan of lonely disabled buttons (like <kbd>Save</kbd>, <kbd>Next</kbd>, or <kbd>Continue</kbd>). Because they don't tell the user what is wrong. They're just disabled and nasty. Especially annoying when there are no other _visuals (or tags for screen readers) in the UI that are hinting about the problem_ in the UI. 

One of such "visuals" is the `*` sign beside the required fields. I think this is _enough_ to have (even if it's too delicate) to tell the user in advance that they should fill in some information. So if the <kbd>Save</kbd> button can remain `disabled` if such visuals are present and visible in the form.

The problem is that the user might miss them due to mane reasons (like they being in a closed collapsible section or our of the viewport). However, those are fine.

For these custom validations, I think the current behavior is OK. I mean, the <kbd>Save</kbd> can be enabled, and once you press it, everything that is "invalid" because of some other logic can then turn red, and show the invalid styles. This way, we are not too screamy and intrusive from the beginning when the form is displayed for the first time, and we decently remind the user that we validated the input before saving, and something is wrong.

In a dream world, the UI would scroll to the first invalid field as well. But that's a longer story.

vicgeralds: > I'm afraid fixing this might be really hard.

What do you mean with "this"? I just wrote that "making required fields be shown as invalid after saving" is not hard, really. 😄 We only need a way to tell `limel-form` to show all errors after saving/loading.

> In a dream world, the UI would scroll to the first invalid field as well. But that's a longer story.

That's possible to implement. Why so negative? 🙂

vicgeralds: Btw, I would love if someone wanted to review the changes I _did make_ in this PR.

We could discuss forever the changes that we could make 😄

Kiarokh: I can't explain what I meant when I wrote that 🤣

Kiarokh (approved):

src/components/form/fields/schema-field.ts
Kiarokh: I'm absolutely not the right person to approve this. But since you verified that the things I was worried about aren't what this PR is touching, and in case you are confident and we're stuck... I approve.

adrianschmidt: > because it doesn't seem to make any difference as long as we're also disabling the save button.

Ah, yeah, of course 🤦‍♂️

PR #2606, opened 2023-11-10
Invalid switch and checkbox

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/switch/switch.tsx
+++ b/src/components/switch/switch.tsx
@@ -59,6 +59,12 @@ export class Switch {
     public readonly = false;
 
     /**
+     * Set to `true` to indicate that the current value is invalid.
+     */
+    @Prop({ reflect: true })
+    public invalid: boolean;
+
+    /**
      * The value of the switch
      */
     @Prop({ reflect: true })
--- a/src/components/checkbox/checkbox.tsx
+++ b/src/components/checkbox/checkbox.tsx
@@ -55,6 +55,12 @@ export class Checkbox {
     public readonly = false;
 
     /**
+     * Set to `true` to indicate that the current value is invalid.
+     */
+    @Prop({ reflect: true })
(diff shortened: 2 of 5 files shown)


src/components/switch/switch.tsx
```diff
@@ -115,7 +121,7 @@ export class Switch {
 
     public render() {
         return (
-            <Host>
+            <Host invalid={this.invalid}>
```
jgroth, line 124: Why is this needed? When the consumer set the `invalid` prop it will already be set on the host component :thinking:

Kiarokh: Oh… I didn't know. tnx

Kiarokh, line 124:
```suggestion
            <Host>
```

Kiarokh: ⚡

jgroth: Clicking the "invalid" checkbox in the switch example does not seem to do anything. 
Is it due to this? Maybe it's needed anyway? I don't see how though :smile:

Kiarokh: I only need to the `invalid` tag in the HTML:
```html
<limel-switch invalid></limel-switch>
```
If there is a helperText, then the helper text turns red. This will happen in [this PR](https://github.com/Lundalogik/lime-elements/pull/2591) as soon as it is merged.

However, we might also want to make the label of the Switch red. I can do that later in a follow up PR.

jgroth (approved):

Issue #2618, opened 2023-11-16
Select: empty icons are rendered in dropdown lists


Kiarokh:
<img width="497" alt="image" src="https://github.com/Lundalogik/lime-elements/assets/35954987/88866ef5-b57e-44ab-9b03-60e3e4972528">

PR #2620, opened 2023-11-16
fix(select): don't render empty icons

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/2618

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/select/select.template.tsx
+++ b/src/components/select/select.template.tsx
@@ -258,6 +258,15 @@ function createMenuItems(
         const name = getIconName(option.icon);
         const color = getIconColor(option.icon, option.iconColor);
 
+        if (!name) {
+            return {
+                text: text,
+                selected: selected,
+                disabled: disabled,
+                value: option,
+            };
+        }
+
         return {
             text: text,
             selected: selected,


jgroth (approved):

LawrenceBorst (commented):
I see Johan already reviewed this. It looks good but I just wanted to point out we can expect similar problems elsewhere where `getIconName` is being used, like `chip-set.tsx`, `list-renderer`, `menu-list-renderer` etc. Not sure if this should be handled in one go from within the icon component using some dynamic CSS or if we should do it this way in other places

Kiarokh:
🤔 maybe let's wait a little with merging this. 
When I added `getIconName` (when we deprecated `iconColor` `iconTitle` etc…) this problem didn't appear. I'm not sure why this has happened now. But if you see a potential that this happens in other places, maybe we should do a more global fix. Now I haven't really looked at why this is happening or what the fix is. Let's just do a visual inspection of other components and see if they also have any issues first.

Kiarokh:
I didn't find this issue on other places. So I think it's better to merge this for now

PR #2291, opened 2023-05-09
feat(help): add new component

Kiarokh:
Specially Intended to be used here: https://github.com/Lundalogik/lime-elements/pull/2622

## Review:
- [x] Commits are [atomic](https://seesparkbox.com/foundry/atomic_commits_with_git)
- [x] Commits have the correct *type* for the changes made
- [x] Commits with *breaking changes* are marked as such

### Browsers tested:
(Check any that applies, it's ok to leave boxes unchecked if testing something didn't seem relevant.)

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

PR #2622, opened 2023-11-17
feat(form): enable providing rich help next to form fields

Kiarokh:
requires: https://github.com/Lundalogik/lime-elements/pull/2291

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [x] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/adapters/widget-adapter.ts
+++ b/src/components/form/adapters/widget-adapter.ts
@@ -3,6 +3,7 @@ import { WidgetProps } from '../widgets/types';
 import { LimeElementsAdapter } from './base-adapter';
 import { capitalize } from 'lodash-es';
 import { LimeSchemaOptions } from '../../../interface';
+import { getHelpComponent } from '../help';
 
 interface WidgetAdapterProps {
     name: string;
@@ -109,20 +110,25 @@ export class LimeElementsWidgetAdapter extends React.Component {
             ...events,
         };
 
-        return React.createElement(LimeElementsAdapter, {
-            name: name,
-            elementProps: {
-                value: value,
-                label: this.getLabel(),
-                disabled: disabled,
-                readonly: readonly,
-                required: this.isRequired(),
-                invalid: this.isInvalid(),
-                'helper-text': this.getHelperText(),
-                ...extraProps,
-            },
-            events: newEvents,
-        });
+        return React.createElement(
+            React.Fragment,
+            {},
+            React.createElement(LimeElementsAdapter, {
+                name: name,
+                elementProps: {
+                    value: value,
+                    label: this.getLabel(),
+                    disabled: disabled,
+                    readonly: readonly,
+                    required: this.isRequired(),
+                    invalid: this.isInvalid(),
+                    'helper-text': this.getHelperText(),
+                    ...extraProps,
+                },
+                events: newEvents,
+            }),
+            getHelpComponent(this.props.widgetProps.schema),
+        );
     }
 
     private isDisabled() {
--- a/src/components/form/examples/form-with-help.tsx
+++ b/src/components/form/examples/form-with-help.tsx
@@ -1,0 +1,59 @@
+import { Component, h, State } from '@stencil/core';
+import { schema } from './help-form-schema';
+
+/**
+ * Form fields with help
+ * It's possible to add extensive help to any form element.
+ * The string you provide can be in Markdown format,
+ * empowering you to present a rich-text experience to the user,
+ * including bullet points, hyperlinks, etc…
+ * Read more on [Help](#/component/limel-help) component.
+ *
+ * :::note
+ * Do not confuse `help` and `helperText`!
+ * The helper text is a short description for the input fields
+ * that becomes visible when the user click on the fields to provide
+ * brief clues about the field or its expected value.
+ *
+ * It can also be used to display validation errors.
+ * These errors will be displayed in red below the fields, without
+ * requiring the users to click on the field.
+ *
+ * Check out the [Input field Component](#/component/limel-input-field)
+ * examples, where we describe how to properly use `helperText`, and `placeholder`.
+ * :::
+ *
+ * :::tip
+ * When rendering a form using a schema, the `helperText`s are automatically
+ * passed for all the fields based on the schema and validation errors.
+ * The `description` specified for a field in the schema is used as
+ * helper text while the field is shown as valid.
+ * When there is validation feedback to provide to the user,
+ * the field is instead marked as invalid with an error message that is displayed
+ * in the place of the helper text.
+ * :::
+ * @link help-form-schema.ts
+ */
+@Component({
+    tag: 'limel-example-form-with-help',
+    shadow: true,
+})
+export class NestedFormExample {
+    @State()
+    private formData: object = {};
+
+    public render() {
+        return [
+            <limel-form
+                onChange={this.handleFormChange}
+                value={this.formData}
+                schema={schema}
+            />,
+            <limel-example-value value={this.formData} />,
+        ];
+    }
+
+    private handleFormChange = (event) => {
+        this.formData = event.detail;
+    };
+}
--- a/src/components/form/examples/help-form-schema.ts
+++ b/src/components/form/examples/help-form-schema.ts
@@ -1,0 +1,84 @@
+export const schema = {
+    type: 'object',
+    properties: {
+        address: {
+            type: 'object',
+            title: 'Book a demo',
+            properties: {
+                planet: {
+                    type: 'string',
+                    title: 'Name',
+                    description: 'Please write your name & surname',
+                    lime: {
+                        help: {
+                            value: '**Why do we need this information?** <br> This enables our sales representatives to contact you personally. We will keep all your personal information safe!',
+                            readMoreLink: {
+                                href: 'https://www.lime-technologies.com/en/av-saas/',
+                                title: 'This link opens in a new tab',
+                                target: '_blank',
+                                text: 'Our Terms & Privacy Policies',
+                            },
+                        },
+                    },
+                },
+                galaxy: {
+                    type: 'string',
+                    title: 'Company',
+                    description: 'Please write your company name',
+                },
+                email: {
+                    type: 'string',
+                    title: 'Work Email',
+                    format: 'email',
+                    lime: {
+                        help: {
+                            value: 'Enter a valid email address so that we can contact you. We will not share this information with any third parties.',
+                            readMoreLink: {
+                                href: 'https://www.lime-technologies.com/en/av-saas/',
+                                title: 'This link opens in a new tab',
+                                target: '_blank',
+                                text: 'Our Terms & Privacy Policies',
+                            },
+                        },
+                    },
+                },
+                race: {
+                    type: 'string',
+                    title: 'How did you hear about Lime CRM?',
+                    oneOf: [
+                        {
+                            type: 'string',
+                            const: 'search',
+                            title: 'Internet search',
+                        },
+                        {
+                            type: 'string',
+                            const: 'recommendation',
+                            title: 'Recommendation from a friend or a colleague',
+                        },
+                        {
+                            type: 'string',
+                            const: 'ads',
+                            title: 'Advertising campaigns',
+                        },
+                        {
+                            type: 'string',
+                            const: 'work',
+                            title: 'Worked with Lime CRM before',
+                        },
+                        {
+                            type: 'string',
+                            const: 'workshop',
+                            title: 'Watched a workshop online',
+                        },
+                    ],
+                    lime: {
+                        help: {
+                            value: '**Why do we need this information?** <br> This enables us to improve our marketing efforts and get a better understanding of how people perceive us as a brand.',
+                        },
+                    },
+                },
+            },
+        },
+    },
+};
--- a/src/components/form/examples/row-layout-schema.ts
+++ b/src/components/form/examples/row-layout-schema.ts
@@ -16,9 +16,11 @@ export const schema = {
                 language: {
                     type: 'string',
                     title: 'Language',
-                    description: 'Select the app language',
                     default: 'ua',
                     lime: {
+                        help: {
+                            value: 'This will affect both the language of the user interface of our app, and also the language of the emails we send you.',
+                        },
                         layout: {
                             icon: 'globe',
                         },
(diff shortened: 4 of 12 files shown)


src/components/form/examples/row-layout-schema.ts
```diff
@@ -19,6 +19,15 @@ export const schema = {
...
                     lime: {
+                        help: {
+                            value: '**Why do we need this information?** This enables our sales representatives to contact you personally. We will keep all your personal information safe!',
+                            readMoreLink: {
+                                href: 'https://www.lime-technologies.com/en/av-saas/',
+                                title: 'This link opens in a new tab',
+                                target: '_blank',
+                                text: 'Our Terms & Privacy Policies',
+                            },
+                        },
```
Kiarokh, line 30: we need another default placement for this layout type 👇 

image

src/components/form/examples/row-layout-schema.ts
```diff
@@ -82,6 +100,15 @@ export const schema = {
...
                         component: {
                             name: 'limel-switch',
                         },
```
Kiarokh, line 114: It seems when a component is specified, the help does not appear at all. image

Do you know why @civing ?

src/components/form/adapters/widget-adapter.ts
```diff
@@ -109,20 +110,23 @@ export class LimeElementsWidgetAdapter extends React.Component {
...
+                    disabled: disabled,
+                    readonly: readonly,
+                    required: this.isRequired(),
+                    invalid: this.isInvalid(),
+                    'helper-text': this.getHelperText(),
+                    ...extraProps,
+                },
+                events: newEvents,
+            }),
+            getHelpComponent(this.props.widgetProps.schema),
```
civing, line 128: @vicgeralds could you take a look at the mechanics surrounding how we insert the help component in the DOM?

src/components/form/examples/form-with-help.tsx
```diff
@@ -0,0 +1,47 @@
+import { Component, h, State } from '@stencil/core';
+import { schema } from './help-form-schema';
+
+/**
+ * Form elements with contextual help
```
vicgeralds, line 5: What's contextual about it?

To me contextual suggests that the help would somehow (dynamically) depend on the context, whatever context means here. So a helper text that depends on validation errors is _more_ contextual than this piece of help. This is just static info providing rich text "help".

Or did you mean to say that the help provides context?

vicgeralds (commented):
Providing help doesn't work for object and array fields. Is this an intentional limitation?

src/components/form/examples/form-with-help.tsx
Kiarokh: I didn't have any specific intentions. Maybe was just thinking that a form field in the context of that form, needs some further explanation / help :) We can change this heading.

src/components/form/examples/form-with-help.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * It's possible to add contextual help to any form element.
```
Kiarokh, line 6:
```suggestion
 * It's possible to add extensive help to any form element.
```

src/components/form/examples/form-with-help.tsx
```diff
@@ -0,0 +1,47 @@
...
+ * The string you provide can be in Markdown format,
+ * empowering you present a rich-text experience to the user,
+ * including bullet points, hyperlinks, etc…
+ * Read more on [Help](#/component/limel-help) component.
+ *
+ * :::note
+ * This "help" is not the same as providing a so called "helper text".
+ * Helper text is a short description of the input fields
+ * that becomes visible when the user click on the fields.
```
Kiarokh, line 15:
```suggestion
 * that becomes visible when the user click on the fields.
 * It is most often used to render validation errors, or provide
 * brief clues about the filed or its expected value.
```

src/components/form/examples/form-with-help.tsx
vicgeralds: > a form field in the context of that form, needs some further explanation

So it's providing help per form field then. When you say "the context of that form", I'm just getting more confused (I don't even know what form means here. The form design described by the schema?) 😄

src/components/form/examples/form-with-help.tsx
```diff
@@ -0,0 +1,49 @@
...
+ * It is most often used to render validation errors, or provide
+ * brief clues about the filed or its expected value.
+ *
+ * Check out the [Input field Component](#/component/limel-input-field)
+ * examples, where we describe how to properly use `helperText`, and `placeholder`.
+ *
+ * In the context of forms, helper texts are also used to
+ * provide validation feedback to the user, about an invalid input.
```
vicgeralds, line 23: This can be confusing. 

"In the context of forms" is vague. I would say the helper text is only ever used in the context of forms! But you mean when rendering a `limel-form` using a schema. Which leads to the question, how do you set `helperText` then? I couldn't find this explained anywhere, but in some examples we set `description` for a field in the schema which is used as helper text.

I think it would be helpful to clearly explain that limel-form will either pass a validation error as helper text, or the schema description when there's no error.

(maybe this example is not the right place for documenting helper texts in detail, but we currently don't have any other place for how they work in limel-form)

src/components/form/adapters/widget-adapter.ts
```diff
@@ -109,20 +110,23 @@ export class LimeElementsWidgetAdapter extends React.Component {
...
-                disabled: disabled,
-                readonly: readonly,
-                required: this.isRequired(),
-                invalid: this.isInvalid(),
-                'helper-text': this.getHelperText(),
-                ...extraProps,
-            },
-            events: newEvents,
-        });
+        return [
```
vicgeralds, line 113: It's better to render a fragment than an array in this case (avoids warning about missing `key`). Also, we could conditionally return just a single element when there's no help component.

src/components/form/adapters/widget-adapter.ts
vicgeralds: 0aa08cf68e1905e1dba4b720ee48236b1a496369

src/components/form/helper/helper.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export function getHelpComponent(schema: LimeJSONSchema) {
+    const help = schema.lime?.help;
+
+    if (!help) {
+        return;
+    }
+
+    const helpProps =
+        typeof help === 'string' ? { content: help } : { ...help };
```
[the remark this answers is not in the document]
vicgeralds: 0aa08cf68e1905e1dba4b720ee48236b1a496369

src/components/form/adapters/widget-adapter.ts
vicgeralds: 0aa08cf68e1905e1dba4b720ee48236b1a496369

src/components/form/examples/form-with-help.tsx
Kiarokh: ⚡

src/components/form/examples/form-with-help.tsx
```diff
@@ -0,0 +1,51 @@
+import { Component, h, State } from '@stencil/core';
+import { schema } from './help-form-schema';
+
+/**
+ * Form elements with help
```
Kiarokh, line 5:
```suggestion
 * Form fields with help
```

src/components/form/examples/form-with-help.tsx
Kiarokh: my intention was mostly providing mental help for decision making, like when to use `helperText`, `placeholder`, or `help`.

> I think it would be helpful to clearly explain that limel-form will either pass a validation error as helper text, or the schema description when there's no error.

But how would you explain this one? Can you add a suggestion please.

src/components/form/examples/form-with-help.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * Read more on [Help](#/component/limel-help) component.
+ *
+ * :::note
+ * This "help" is not the same as providing a so called "helper text".
+ * Helper text is a short description of the input fields
+ * that becomes visible when the user click on the fields.
+ * It is most often used to render validation errors, or provide
+ * brief clues about the field or its expected value.
+ * It is most often used to render validation errors, or provide
+ * brief clues about the filed or its expected value.
```
vicgeralds, line 19: Duplicated sentence
```suggestion
```

src/components/form/examples/form-with-help.tsx
```diff
@@ -0,0 +1,51 @@
...
+ * empowering you present a rich-text experience to the user,
+ * including bullet points, hyperlinks, etc…
+ * Read more on [Help](#/component/limel-help) component.
+ *
+ * :::note
+ * This "help" is not the same as providing a so called "helper text".
+ * Helper text is a short description of the input fields
+ * that becomes visible when the user click on the fields.
+ * It is most often used to render validation errors, or provide
+ * brief clues about the field or its expected value.
```
vicgeralds, line 17: Validation errors is maybe not the best first example after explaining that helper text becomes visible when clicking the field. That's not the behaviour when displaying errors. I suggest shorten this sentence, to just mention that it's _also_ used for validation errors. We're already pointing the input field examples for more details.

```suggestion
 * It is also used to render validation errors when a form field has invalid input.
```

src/components/form/examples/form-with-help.tsx
vicgeralds:
```suggestion
 * When rendering a form with a schema, helper texts are automatically passed for all the fields based on the schema and validation errors. The schema `description` for a field is used when there are no errors, otherwise the first validation error is displayed in the place of the helper text.
```

src/components/form/examples/form-with-help.tsx
Kiarokh: > Validation errors is maybe not the best first example after explaining that helper text becomes visible when clicking the field.

That's not true. We force-show the helper text, when field is invalid.

vicgeralds: > That's not true. We force-show the helper text, when field is invalid.

That's what I meant. So it's inconsistent with the preceding definition of helper text as "a short description of the input fields that becomes visible when the user click on the fields."

My suggestion is that we present validation errors as an alternative use of helper text, that's an exception (always shown when field is invalid), rather than as the main use.

src/components/form/examples/form-with-help.tsx
vicgeralds: This description is not completely accurate though, because the `isInvalid` check is a bit more complicated than that:
https://github.com/Lundalogik/lime-elements/pull/2601

I don't think the consumer needs to be bothered with the exact details of when errors are shown, but maybe this is better:

"The schema `description` for a field is used as helper text while the field is shown as valid. When there is validation feedback to provide to the user, the field is instead marked as invalid with an error message that is displayed in the place of the helper text."

src/components/form/helper/helper.ts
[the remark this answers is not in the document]
Kiarokh: ⚡

src/components/form/examples/form-with-help.tsx
Kiarokh: I re-edited the info

src/components/form/examples/form-with-help.tsx
```diff
@@ -0,0 +1,59 @@
...
+ *
+ * :::note
+ * Do not confuse `help` and `helperText`!
+ * The helper text is a short description for the input fields
+ * that becomes visible when the user click on the fields to provide
+ * brief clues about the field or its expected value.
+ *
+ * It can also be used to display validation errors.
+ * These errors will be displayed in red below the fields, without
+ * the requiring the users to click on the field.
```
vicgeralds, line 20:
```suggestion
 * requiring the users to click on the field.
```

vicgeralds:
> I think it'd better if we add markdown support there at that level, instead of rendering a `<p>` element.

Do you mean we should support markdown in schema `description`?

This is confusing. Forms were never meant to be built from schemas. Schemas are for machines and apis. 😢

Kiarokh:
I think it's just fine if the "sections headers, **headings** + **subheadings** do not show any help. The idea of the help was to be together with the form's input fields that the user interacts with.

However, I think it's super nice if the description of the form section (the subheading) can show markdown. I'm not religious about schema and what the creator intended to do with them 😄 . But either way, that can be a future PR.

src/components/form/examples/form-with-help.tsx
```diff
@@ -0,0 +1,59 @@
+import { Component, h, State } from '@stencil/core';
+import { schema } from './help-form-schema';
+
+/**
+ * Form fields with help
+ * It's possible to add extensive help to any form element.
+ * The string you provide can be in Markdown format,
+ * empowering you present a rich-text experience to the user,
```
vicgeralds, line 8:
```suggestion
 * empowering you to present a rich-text experience to the user,
```

vicgeralds:
> I'm not religious about schema and what the creator intended to do with them 😄 

To me, working with schemas for building forms is the ceremony of a wicked cult that I want to save people from. 😄

vicgeralds (commented):
I don't think I have any more comments now 🙂

Kiarokh:
🤣 sounds like you wanna approve the PR then

src/components/form/examples/form-with-help.tsx
vicgeralds: looks like you accidentally undid this

vicgeralds (approved):

src/components/form/examples/form-with-help.tsx
Kiarokh: 🙈

src/components/form/examples/help-form-schema.ts
```diff
@@ -0,0 +1,84 @@
...
+                            readMoreLink: {
+                                href: 'https://www.lime-technologies.com/en/av-saas/',
+                                title: 'This link opens in a new tab',
+                                target: '_blank',
+                                text: 'Our Terms & Privacy Policies',
+                            },
+                        },
+                    },
+                },
+                race: {
```
adrianschmidt, line 45: `race` and `galaxy`?

This was perhaps copied and pasted in a bit of a hurry…? 😝

src/components/form/examples/help-form-schema.ts
adrianschmidt, line 1: For next time:

I think it's nice if all files that belong to the same example are sorted together, so I would have preferred this to be called `form-with-help-schema.ts`, for example. I know the name itself might actually be slightly worse, but I think the usefulness of keeping all files that are part of a specific example easily identifiable as such justifies it 😊

Issue #2093, opened 2023-01-31
Select: `Changing Available Options` example needs fix


LucyChyzhova:
## Current behavior

https://user-images.githubusercontent.com/50618208/215774893-77dd6d47-f7b2-4cec-938b-9fcf47c88f80.mov

1. …

## Environment

- **lime-elements** version: 
- Framework used: 
- Logs:

adrianschmidt:
That's per design actually, to make sure it doesn't crash if it's given an empty set of options 😊

But we should probably clarify that in the example description… 👍

LucyChyzhova:
aha! I see, yeah, I agree, then the description would help here 😀.
I will add it.

adrianschmidt:
@LucyChyzhova, you said you were going to add a description here, but I don't think you ever got around to it. Should we add it, or should I just close this issue? 😊

PR #2637, opened 2023-11-29
doc(select): improve example of `select-change-options`

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/2093

Screenshot 2023-11-29 at 10 17 39

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/select/examples/select-change-options.tsx
```diff
@@ -19,6 +23,12 @@ export class SelectExample {
     public currentOptionGroup = 0;
 
     private optionGroups: Option[][] = [
+        [],
+        [
+            { text: '', value: '', disabled: true },
+            { text: 'Bart Simpson', value: 'bart' },
+            { text: 'Ned Flanders', value: 'ned' },
+        ],
```
adrianschmidt, line 31: Why did you move this set of options? Did it not work as expected in the place it was? Or did things not work as expected when the set with "Luke Skywalker" came immediately after the empty set? Or some other reason? 😄

This set and the set with "Homer Simpson" should be next to each other, because it allows you to test what happens if you have "Ned Flanders" selected in one, and then switch to the other. "Ned Flanders" should be selected in that one as well. But I would assume that doesn't work if there's been a set with no "Ned Flanders" in it in-between.

LucyChyzhova: Yes, it should work, I moved it because I found another logic there 😀. 
Empty array ->  2 options plus an empty row -> 3 options -> 3 options plus an empty row -> 4 options.

However, if this makes sense only to me, I will put it back 🙂

adrianschmidt: Haha, yeah, they were actually sorted differently on purpose, because I didn't want the number of options to only increase.

But now I've update the example quite a lot, so it doesn't matter anymore. You can now switch from any group to any other group, and you can recreate the component at any state, to check it can load in that state 👍

adrianschmidt (approved):

PR #2638, opened 2023-11-29
fix(menu): be able to use ListSeparator in search result

TommyLindh2:
A very small typing issue, where we missed that the search results of the `MenuSearcher` type could contains `ListSeparators`

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu/menu.types.ts
+++ b/src/components/menu/menu.types.ts
@@ -114,9 +114,11 @@ export interface MenuItem<T = any> {
  * a promise that will eventually be resolved with an array of `MenuItem`:s.
  * @param {string} query A search query. What the user has written
  * in the input field of a limel-menu.
- * @returns {Promise<MenuItem[]>} The search result.
+ * @returns {Promise<Array<MenuItem | ListSeparator>>} The search result.
  */
-export type MenuSearcher = (query: string) => Promise<MenuItem[]>;
+export type MenuSearcher = (
+    query: string
+) => Promise<Array<MenuItem | ListSeparator>>;
 
 /**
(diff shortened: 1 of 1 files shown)


src/components/menu/menu.types.ts
adrianschmidt, line 1: When you fix the lint error, change this commit from a `chore` to a `fix` as well. Then I'll approve 😊

adrianschmidt (changes requested):

src/components/menu/menu.types.ts
TommyLindh2: I asked chatGPT to get the prefix right, but still don't get it right xD I will fix that

adrianschmidt: https://github.com/Lundalogik/lime-elements/blob/next/CONTRIBUTING.md#type

I ought to update that slightly though, because we don't follow it completely. For example, we use chore a bunch for things that does touch src or test files, but that doesn't need a version bump, and also doesn't feel like a refactor…

But at least it's something of a guide.

adrianschmidt (approved):

adrianschmidt:
No, since it's just the type of something the consumer will send in to our component, and the type has _expanded_, I think we're good 👍

adrianschmidt (approved):

PR #2659, opened 2023-12-19
fix(disabled-vs-hidden): make the guidelines responsible

LucyChyzhova:
fix: 

https://github.com/Lundalogik/lime-elements/assets/50618208/49ef53ff-d035-4e50-87db-970120b61cf5

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/design-guidelines/disabled-hidden/examples/button-disabled-vs-hidden.scss
+++ b/src/design-guidelines/disabled-hidden/examples/button-disabled-vs-hidden.scss
@@ -2,16 +2,9 @@
 
 .do,
 .do-not {
-    width: 30rem;
-    margin: 1rem 0 2rem 0;
-    background-color: rgb(var(--contrast-200));
-
+    .relevant-button,
     limel-button {
-        padding: 0 0 1rem 0.5rem;
-    }
-
-    .relevant-button {
-        margin-top: 1rem;
+        margin: 1rem;
     }
 
     .assign {
@@ -20,7 +13,6 @@
 }
 
 #tooltip-example {
-    margin-top: 1rem;
     width: 5rem;
 }
 
@@ -31,20 +23,11 @@
 .button-icon {
     display: flex;
     align-items: center;
-
-    .disabled-button {
-        padding-right: 0.5rem;
-        margin-top: 1rem;
-    }
 }
 
 .split-example,
-.relevan-buttons-example {
-    display: flex;
-    gap: 5rem;
-
+.relevant-buttons-example {
     limel-split-button {
-        margin-top: 1rem;
-        padding: 0 0 1rem 0.5rem;
+        margin: 1rem;
     }
 }
--- a/src/design-guidelines/disabled-hidden/examples/disabled-vs-hidden.tsx
+++ b/src/design-guidelines/disabled-hidden/examples/disabled-vs-hidden.tsx
@@ -48,18 +48,23 @@ export class ButtonDisabledVsHideExample {
                 cases, users should be able to do something to enable the
                 <code>disabled</code> element!
             </p>,
-            <div class="do">
-                <limel-header icon="ok" heading="Disabled button in the form" />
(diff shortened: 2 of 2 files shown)


src/design-guidelines/disabled-hidden/examples/disabled-vs-hidden.tsx
```diff
@@ -127,7 +127,7 @@ export class ButtonDisabledVsHideExample {
                 button, it is better to show the <b> unassign</b> button
                 instead.
             </p>,
-            <div class="relevan-buttons-example">
+            <div class="relevant-buttons-example">
```
Kiarokh, line 130: image

If you use the same classes and HTML structure that we use in the other examples ☝️ , then you don't need to make special CSS for this example. It's more maintainable to have the same source for all of these layouts, and not treat them individually. 😊

Kiarokh (changes requested):

src/design-guidelines/disabled-hidden/examples/disabled-vs-hidden.tsx
Kiarokh: as you see, they are wrapped in a div with class of `do-dont-container` which has all the necessary styles. 

Also make sure to remove any replicated styles which we don't need.

LucyChyzhova: The generic (from other examples) one didn't help me here, then would it be better just to have a customer here?

LucyChyzhova: I mean, I use some CSS from that import. But not all work for my examples like ` background-color: rgb(var(--contrast-100));` I have overwritten it here. 
I tried to use only `import` but it didn't bring desirable results.

LucyChyzhova: made a fixup

Kiarokh: this is great now. 

It's alright to overwrite things if needed. My point was not to force styles that aren't needed. ;) 

There is just this problem now, but it has nothing to do with your code. It's just a poor CSS that I wrote back in the day. 😅 👇 

image

we will fix it later

src/design-guidelines/disabled-hidden/examples/button-disabled-vs-hidden.scss
```diff
@@ -2,16 +2,11 @@
 
 .do,
 .do-not {
-    width: 30rem;
-    margin: 1rem 0 2rem 0;
     background-color: rgb(var(--contrast-200));
```
Kiarokh, line 5: this can also be removed 😸

src/design-guidelines/disabled-hidden/examples/button-disabled-vs-hidden.scss
```diff
@@ -31,20 +25,13 @@
...
-        margin-top: 1rem;
-    }
 }
 
 .split-example,
-.relevan-buttons-example {
-    display: flex;
-    gap: 5rem;
+.relevant-buttons-example {
+    margin: 1rem 0;
```
Kiarokh, line 32: this seems also unnecessary

src/design-guidelines/disabled-hidden/examples/button-disabled-vs-hidden.scss
LucyChyzhova: Background color?

Kiarokh (approved):
🚀 👩‍🎤

PR #2663, opened 2023-12-20
feat(input-field): add support for `file://` protocol

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/util/link-helper.ts
+++ b/src/util/link-helper.ts
@@ -39,6 +39,9 @@ function hasKnownProtocol(input: string) {
         'https',
         'http',
 
+        // `file` may or may not work, due to cross-origin restrictions and browser settings.
+        'file',
+
         // m-files is a protocol used by the M-Files desktop app or something.
         // It's not a web protocol, but it allows open M-Files links in their app.
         'm-files',


src/util/link-helper.ts
```diff
@@ -39,6 +39,9 @@ function hasKnownProtocol(input: string) {
         'https',
         'http',
 
+        // `file` may or may not work, due to cross-origin restrictions and browser settings.
+        'file',
```
adrianschmidt, line 43: Using `file://` links probably won't actually work for people, but by adding them here, lime-elements won't add `https://` in front of them, which at least make people see that their browser is blocking the request, rather than ping us about limel-input-field's weird behaviour 😅

See [thread on our forum](https://forum.lime-crm.com/t/urlastext-still-adding-https-before-my-url/5648/4?u=ads) (internal link)

hannahu (approved):

PR #2668, opened 2023-12-22
Dematerialize `limel-button`, remove its dependency on MDC

Kiarokh:
`limel-button` includes an internal `button` element. There is no reason to depend on MDC, merely to get a bunch of styles.  We don't get any specific functionality from MDC as far as I can say. Everything else is handled basically inside the component already.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -19,18 +19,17 @@ button {
     display: flex;
     align-items: center;
     justify-content: center;
-    gap: 0.375rem;
+    gap: 0.125rem;
 
     border: none;
     border-radius: 0.4rem;
-    min-width: functions.pxToRem(36);
     padding: {
         top: 0;
-        right: var(--button-padding-right, #{functions.pxToRem(12)});
+        right: var(--button-padding-right, 0.5rem);
         bottom: 0;
-        left: var(--button-padding-left, #{functions.pxToRem(12)});
+        left: var(--button-padding-left, 0.5rem);
     }
-    height: 100%;
+
     min-height: functions.pxToRem(36);
     width: 100%;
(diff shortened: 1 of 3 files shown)


src/components/button/button.scss
```diff
@@ -3,9 +3,6 @@
 
 @use '../../style/internal/lime-theme';
 
-@use '@material/button/styles';
-@use '@material/button';
```
Kiarokh, line 7: In other components like `limel-input-field` we are importing other functionality from MDC. 👇 

https://github.com/Lundalogik/lime-elements/blob/14eca6dce61dbb34519036d0453fee4996cd9b16/src/components/input-field/input-field.tsx#L1

But here, we are just importing some styles.

src/components/button/button.scss
```diff
@@ -19,63 +16,73 @@
...
+    align-items: center;
+    justify-content: center;
+    gap: 0.375rem;
+
+    border: none;
+    border-radius: 0.4rem;
+    min-width: functions.pxToRem(36);
+    padding: {
+        top: 0;
+        right: var(--button-padding-right, #{functions.pxToRem(12)});
```
LucyChyzhova: > {functions.pxToRem(12)});

could we put some number instead of the function?

LucyChyzhova (approved):
🤩

PR #2673, opened 2023-12-28
Remove dependency on Roboto font, and make all components font-agnostic. Enable consumers to have their desired font-family

Kiarokh:
requires: https://github.com/jgroth/kompendium/pull/130
fixes https://github.com/Lundalogik/crm-feature/issues/3820

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [ ] iOS

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,25 @@ For a full list of components, along with live examples, please visit the [docum
 
 #### 1. Font
 
-The "Roboto" font is included for development purposes, but is not included in the published package. This font should be supplied by the consuming application. If not supplied, texts will fall back to suitable alternatives.
+To achieve a blazing fast rendering, our components' user interface utilizes a default cross-browser sans-serif font stack. As web components typically inherit font-related styles such as `font-family`, `font-size`, and `color`, we recommend defining these styles at a higher level, such as the `<body>` element. This is because we do not specify these defaults on each individual component.
+
+To maintain consistency with the look & feel demonstrated in this documentation, we suggest incorporating the following styles into your project:
+
+```css
+font-family: ui-sans-serif, system-ui, sans-serif;
+font-size: 0.875rem;
+font-style: normal;
+font-weight: 400;
+color: rgb(var(--contrast-1500));
+```
+
+💡 About the `color` specified above, read more on [our color system](/#/DesignGuidelines/color-system.md/).
+
+Feel free to customize the font-family and related styles to suit your project's needs. For example, you might prefer a different typeface like below:
+
+```css
+font-family: 'Roboto', Arial, Verdana, sans-serif;
+```
 
 #### 2. Icons
 
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -17,6 +17,16 @@
 
 @include mixins.visualize-aria-expanded('button');
 
+button.mdc-button {
+    // As long as this component is depended on MDC,
+    // we need to force it to be font-agnostic.
+    // When MDC-dependency is removed, this block can also be removed.
+    // However, on removal of MDC-dependency, we should also make sure to check
+    // other font-related styles that might be set by MDC,
+    // such as `letter-spacing` or `font-size`.
+    font-family: inherit;
+}
+
 button {
     display: flex;
     align-items: center;
@@ -76,11 +86,9 @@ button {
 }
 
 .label {
-    font-family: Roboto, sans-serif;
     font-size: 0.875rem;
-    font-weight: 500;
-    letter-spacing: functions.pxToRem(1);
-    opacity: 1;
+    font-weight: 600;
+    letter-spacing: functions.pxToRem(0.5);
     padding: 0 0.25rem;
 }
 
--- a/src/components/date-picker/flatpickr-adapter/flatpickr-adapter.scss
+++ b/src/components/date-picker/flatpickr-adapter/flatpickr-adapter.scss
@@ -15,7 +15,6 @@
 
     * {
         box-sizing: border-box;
-        font-family: 'Roboto';
     }
 }
 
--- a/src/components/select/select.scss
+++ b/src/components/select/select.scss
@@ -24,6 +24,18 @@
     left: functions.pxToRem(16);
 }
 
+.mdc-select__anchor,
+.mdc-floating-label,
+.mdc-select__selected-text {
+    // As long as this component is depended on MDC,
+    // we need to force it to be font-agnostic.
+    // When MDC-dependency is removed, this block can also be removed.
+    // However, on removal of MDC-dependency, we should also make sure to check
+    // other font-related styles that might be set by MDC,
+    // such as `letter-spacing` or `font-size`.
+    font-family: inherit;
+}
+
 .mdc-select__dropdown-icon-graphic {
     transition: transform 0.2s ease;
 }
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -340,3 +340,19 @@
         }
     }
 }
+
+/**
+* This mixin creates a cross-browser font stack.
+* - `sans-serif` can be used for the UI of the components.
+* - `monospace` can be used for code.
+*/
+@mixin font-family($letterform) {
+    @if $letterform == sans-serif {
+        font-family: ui-sans-serif, system-ui, sans-serif;
+    } @else if $letterform == monospace {
+        font-family: ui-monospace, 'Cascadia Code', 'Source Code Pro', Menlo,
+            Consolas, 'DejaVu Sans Mono', monospace;
+    } @else {
+        @error "Please specify the font-family #{$letterform}!";
+    }
+}
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -345,6 +345,11 @@
 * This mixin creates a cross-browser font stack.
 * - `sans-serif` can be used for the UI of the components.
 * - `monospace` can be used for code.
+*
+* ⚠️ If we change the font stacks, we need to update
+* 1. the consumer documentation in `README.md`, and
+* 2. the CSS variables of `--kompendium-example-font-family`
+* in the `<style>` tag of `index.html`.
 */
 @mixin font-family($letterform) {
     @if $letterform == sans-serif {
--- a/src/components/badge/badge.scss
+++ b/src/components/badge/badge.scss
@@ -1,6 +1,5 @@
 @use '../../style/functions';
 @use '../../style/internal/lime-theme';
-@use '../../style/internal/lime-typography';
 @use '../../style/mixins';
 
 /**
@@ -23,7 +22,6 @@
 }
 
 span {
-    @include lime-typography.base;
     cursor: default;
     box-sizing: border-box;
 
--- a/src/components/banner/banner.scss
+++ b/src/components/banner/banner.scss
@@ -1,5 +1,4 @@
 @use '../../style/internal/lime-theme';
-@use '../../style/internal/lime-typography';
 @use '../../style/internal/variables';
 @use '../../style/functions';
 
@@ -9,7 +8,6 @@
  */
 
 .lime-banner {
-    @include lime-typography.typography(body2);
     min-height: variables.$lime-banner-height;
     display: none;
     background-color: rgba(var(--contrast-100), 0.7);
--- a/src/components/banner/banner.tsx
+++ b/src/components/banner/banner.tsx
@@ -49,7 +49,7 @@ export class Banner {
                 <div class="lime-banner__surface">
                     <div class="lime-banner__content">
                         {this.renderIcon()}
-                        <div class="mdc-typography--body2">{this.message}</div>
+                        <div>{this.message}</div>
                         <div class="lime-banner__actions">
                             <slot name="buttons" />
                         </div>
--- a/src/components/breadcrumbs/breadcrumbs.scss
+++ b/src/components/breadcrumbs/breadcrumbs.scss
@@ -60,6 +60,10 @@ ol {
 
         &:after {
             content: var(--limel-breadcrumbs-divider);
+            display: flex;
+            align-items: center;
+            justify-content: center;
+
             width: var(--limel-breadcrumbs-gap);
 
             position: absolute;
--- a/src/components/button-group/button-group.scss
+++ b/src/components/button-group/button-group.scss
@@ -9,6 +9,16 @@
 */
 
 .mdc-chip {
+    // As long as this component is depended on MDC,
+    // we need to force it to be font-agnostic.
+    // When MDC-dependency is removed, this block can also be removed.
+    // However, on removal of MDC-dependency, we should also make sure to check
+    // other font-related styles that might be set by MDC,
+    // such as `letter-spacing` or `font-size`.
+    font-family: inherit;
+}
+
+.mdc-chip {
     @include mixins.is-flat-inset-clickable($background-color: transparent);
     max-width: 100%;
     min-width: functions.pxToRem(32);
--- a/src/components/callout/callout.scss
+++ b/src/components/callout/callout.scss
@@ -77,4 +77,5 @@
 .heading {
     margin: 0;
     font-size: 1rem;
+    font-weight: 600;
 }
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -24,6 +24,16 @@
 @include form-field.core-styles;
 
 .mdc-form-field {
+    // As long as this component is depended on MDC,
+    // we need to force it to be font-agnostic.
+    // When MDC-dependency is removed, this block can also be removed.
+    // However, on removal of MDC-dependency, we should also make sure to check
+    // other font-related styles that might be set by MDC,
+    // such as `letter-spacing` or `font-size`.
+    font-family: inherit;
+}
+
+.mdc-form-field {
     display: flex;
 
     .mdc-checkbox {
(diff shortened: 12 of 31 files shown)


src/components/button/button.scss
```diff
@@ -18,6 +18,14 @@
...
 }
 
+button.mdc-button {
+    // As long as this component is depended on MDC,
+    // we need to force it to be font-agnostic.
+    // When MDC-dependency is removed, this block can also be removed.
+    font-family: inherit;
+    font-size: inherit;
+}
+
```
Kiarokh, line 28: I hope this pr: https://github.com/Lundalogik/lime-elements/pull/2668 gets merged before the current PR. Then we can revisit this block of code. Let's see which one goes in first.

Kiarokh:
The commits individually are not `feat`. But all together they enable consumers to have their desired font-family. So this is a `feat` at the end. Not sure what is best to do. Like squashing all comits into one and making it a `feat` sounds nice. But that makes the review process hard.

Maybe after the review is done, we should squash all and write this commit message:

```
feat: enable having a custom `font-family`

From now on, you do not have to rely on Roboto as a font, to render the UI correctly.
You can now specify any font-family that you like, for your project.
```

src/components/button/button.scss
adrianschmidt: I'm guessing you've already revisited this since #2668 was merged?

adrianschmidt:
> The commits individually are not `feat`. But all together they enable consumers to have their desired font-family. So this is a `feat` at the end. Not sure what is best to do. Like squashing all comits into one and making it a `feat` sounds nice. But that makes the review process hard.
> 
> Maybe after the review is done, we should squash all and write this commit message:
> 
> ```
> feat: enable having a custom `font-family`
> 
> From now on, you do not have to rely on Roboto as a font, to render the UI correctly.
> You can now specify any font-family that you like, for your project.
> ```

I'll have a look-think 😄

src/components/button/button.scss
adrianschmidt: Ah, actually, maybe not?

image

Kiarokh: Oh I forgot to update this comment. I've already done what was needed after merging https://github.com/Lundalogik/lime-elements/pull/2668

Not sure what the conflict is :/

Kiarokh: I fixed the conflict and pushed. Had nothing to to with that other PR. It was super weird why git couldn't already figure it out by itself. But fixed now.

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.scss
```diff
@@ -15,7 +15,11 @@
 
     * {
         box-sizing: border-box;
-        font-family: 'Roboto';
+        // To ensure that the date-picker's popover is font-agnostic,
+        // but still displays the same font as the rest of the components.
+        // This is needed, since the popover opens in the Portal,
+        // which inherits its font from the root element (Kompendium).
+        font-family: var(--kompendium-example-font-family, inherit);
```
adrianschmidt, line 22: I guess that, since we fall back to `inherit`, and we don't seem to expect that anyone should have to set `--kompendium-example-font-family` anywhere, apart from ourselves setting it in the docs, I guess we can live with this if it's _absolutely necessary_. But I really don't like that we are referring to kompendium stuff in our production code. Anything relating to kompendium belongs in the examples or other docs related places only 😞

adrianschmidt: I'm not sure I entirely understand why we can't just use `inherit` here though? Does that inherit the wrong font when we show the component in the docs? 🤔

adrianschmidt: (And if it does, can't that happen for a consumer as well?)

src/style/mixins.scss
```diff
@@ -340,3 +340,24 @@
...
+* in the `<style>` tag of `index.html`.
+*/
+@mixin font-family($letterform) {
+    @if $letterform == sans-serif {
+        font-family: ui-sans-serif, system-ui, sans-serif;
+    } @else if $letterform == monospace {
+        font-family: ui-monospace, 'Cascadia Code', 'Source Code Pro', Menlo,
+            Consolas, 'DejaVu Sans Mono', monospace;
+    } @else {
+        @error "Please specify the font-family #{$style}!";
```
adrianschmidt, line 361: What does `#{$style}` refer to? Should it say `#{$letterform}` instead?

README.md
```diff
@@ -22,7 +22,25 @@ For a full list of components, along with live examples, please visit the [docum
 
 #### 1. Font
 
-The "Roboto" font is included for development purposes, but is not included in the published package. This font should be supplied by the consuming application. If not supplied, texts will fall back to suitable alternatives.
+To achieve a blazing fast rendering, our components' user interface utilizes a default cross-browser sans-serif font stack. As web components typically inherit font-related styles such as `font-family`, `font-size`, and `color`, we recommend defining these styles at a higher level, such as the `<body />` element. This is because we do not specify these defaults on each individual component.
```
adrianschmidt, line 25:
```suggestion
To achieve a blazing fast rendering, our components' user interface utilizes a default cross-browser sans-serif font stack. As web components typically inherit font-related styles such as `font-family`, `font-size`, and `color`, we recommend defining these styles at a higher level, such as the `<body>` element. This is because we do not specify these defaults on each individual component.
```

`body` is not a self-closing tag, so it just looks weird to have it self-close when referring to it like this.

```
fixup! docs(README.md): explain how font-related styles should be supplied
```

src/components/popover-surface/popover-surface.scss
```diff
@@ -11,6 +11,12 @@
     max-height: inherit; // inherits it from the dynamically calculated `max-height` of `limel-portal--container`
     max-width: calc(100vw - 2rem);
     margin: 0 0.25rem;
+
+    // To ensure that the tooltip is font-agnostic,
```
adrianschmidt, line 15: Copy-paste mistake?

```suggestion
    // To ensure that the popover is font-agnostic,
```

README.md
adrianschmidt, line 1: limel-circular-progress:

https://github.com/Lundalogik/lime-elements/assets/154725/16b6603d-939a-46f4-ac31-10f994898969

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.scss
adrianschmidt: Ok, yeah, I've verified that this is needed in order to not get Roboto in the flatpickr in the docs. But it feels like there must be some other way to fix that. I'll have a look.

Kiarokh: the reason is that in Kompendium, the `--kompendium-example-font-family` is set in deeper container in some shadow dom that displays the "examples". But **_WE_** have this limel-portal that opens in the `body`, which has nothing to do with that deep container. So if we simply say `inherit`, then we will be inheriting from the `body` or `root`, which will give us Kompendium's font family.

src/style/mixins.scss
Kiarokh: 🙈 copy/paste mistake

Yes, it should say `#{$letterform}`

Kiarokh, line 361:
```suggestion
        @error "Please specify the font-family #{$letterform}!";
```

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.scss
adrianschmidt: I added a fixup that solves this with your font-family mixin instead. I've checked that it works for the datepicker, and I would expect that it works for the other cases too. Please have a look and see if you think that's going to work 🙏

src/components/tooltip/tooltip-content.scss
```diff
@@ -9,9 +11,9 @@
 
     // To ensure that the tooltip is font-agnostic,
     // but still displays the same font as the rest of the components.
-    // This is needed, since the tooltip opens in the Portal,
-    // which inherits its font from the root element (Kompendium).
-    font-family: var(--kompendium-example-font-family, inherit);
+    // This is needed, since the tooltip opens in a portal,
+    // which inherits its font from the root element.
+    @include mixins.font-family(sans-serif);
```
Kiarokh, line 16: this won't work, because the component is not font-agnostic anymore now; as you are telling it which font family it should use

Kiarokh: can't we dynamically set `font-family: var(--kompendium-example-font-family)` on the portal itself somehow?

Kiarokh: another alternative is also just not to care that portal inherits its font family from root. Litterally, NOBODY will notice this.

src/components/select/select.scss
adrianschmidt, line 1: Hmm…

Current NEXT:
image

This PR:
image

In the current NEXT-version, both the select and its dropdown have the same font (which doesn't look like it's Roboto, based on the nines 🤔), but in this PR, the select uses the expected font, while the dropdown uses Roboto. I guess the Roboto comes from Kompendium and we need to override it…

Curious that the font used in NEXT _isn't_ Roboto though… 🤨 🤷‍♂️

src/components/tooltip/tooltip-content.scss
adrianschmidt: > this won't work, because the component is not font-agnostic anymore now; as you are telling it which font family it should use

🙈

src/components/select/select.scss
Kiarokh: Ah this is also due to the surface being opened in the portal. Same story as the other ones

Kiarokh: Whatever we end up doing for popover and tooltip, etc, we should do for this as well. It feels like we should add `font-family: var(--kompendium-example-font-family, inherit);` to portal instead, which will handle all scenarios.

src/components/tooltip/tooltip-content.scss
adrianschmidt: I removed this fixup, but added another one instead, which ought to solve the problem properly: [`968fe8b` (#2673)](https://github.com/Lundalogik/lime-elements/pull/2673/commits/968fe8b15b84b6c3cdda6e9ce0c24f81b9e26020) ([perma-link](https://github.com/Lundalogik/lime-elements/commit/968fe8b15b84b6c3cdda6e9ce0c24f81b9e26020))

src/components/select/select.scss
adrianschmidt: In the fixup I just pushed, I've added a specific override for each component that has this problem. I considered using just one property, rather than one for each component, but was thinking that that might be annoying if you actually want different components to use different fonts, for whatever reason…

But yeah, we could still add one for the portal, and then have specific ones for the ones we know of too. Then you can just use the portal override to catch them all, but leave the portal override unset and set the component overrides if you really need that.

adrianschmidt: (I'll fix 👍)

README.md
Kiarokh: [`c9bfecd` (#2673)](https://github.com/Lundalogik/lime-elements/pull/2673/commits/c9bfecd18b8a829fe54f9ad76b30eda6559af00b)

src/components/popover-surface/popover-surface.scss
```diff
@@ -1,6 +1,10 @@
 @use '../../style/functions';
 @use '../../style/mixins';
 
+/**
+ * @prop --limel-popover-font-family: Font family override for the popover. Because the popover is rendered in a portal, it inherits its font from the body element. If the body's font is not what should be used for the popover, this property can be used to override the font.
```
Kiarokh, line 5: Since this var starts with `--limel-` prefix, it means this is an internal variable (which is perfect). But that also means we shouldn't add public docs for it, using `* @prop`. You could say this component itself is `private` so nobody sees this, but I'd still not do it. It's confusing.

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.scss
```diff
@@ -7,6 +7,7 @@
 
 /**
  * @prop --today-label: Tooltip label for "today".
+ * @prop --limel-date-picker-dropdown-font-family: Font family override for the date-picker dropdown. Because the dropdown is rendered in a portal, it inherits its font from the body element. If the body's font is not what should be used for the date-picker, this property can be used to override the font.
```
Kiarokh, line 10: this one is private prop, but it's publicly documented. See [`968fe8b` (#2673)](https://github.com/Lundalogik/lime-elements/pull/2673/commits/968fe8b15b84b6c3cdda6e9ce0c24f81b9e26020#r1442645698)

src/components/select/select.scss
Kiarokh: I like your idea. But I like it even more if we only set one on `div class="limel-portal--container". In this case, I think we would only need set the font-family of date-picker to `inherit`. Others will already inherit from limel-portal--container

adrianschmidt: I'll fix.

src/components/popover-surface/popover-surface.scss
adrianschmidt: No, this isn't an internal variable. But it has to start with a prefix because it has to be a global variable, so we need the prefix to avoid conflicts with global variables that other libs might provide.

Why are we using `--limel-` to denote that a variable is internal?

Kiarokh: Yes, throughout Lime Elements, we have always used `--limel-` as a convention for our internal stuff, and when we need public variables, we always use the component's name first, like `--popover-xxx-yyy`

src/components/select/select.scss
adrianschmidt: All right, I've relented and removed the override variables for the specific components, and just added one for the portal.

I'm convinced I saw a comment from you stating that we needed a specific variable for flatpickr, but I can't find that now? The date-picker dropdown also uses the portal though, so that's working now 👍

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.scss
adrianschmidt: Removed.

src/components/popover-surface/popover-surface.scss
adrianschmidt: I've removed this, including the variable itself.

src/components/select/select.scss
adrianschmidt: ⚡

(I also rebased on latest next)

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.scss
Kiarokh, line 22: this one still needs to be change.
```suggestion
        font-family: inherit;
```

src/components/select/select.scss
Kiarokh: > I'm convinced I saw a comment from you stating that we needed a specific variable for flatpickr, but I can't find that now?

I added a new comment :)

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.scss
adrianschmidt: Ah! I must have missed this one… 🤔

adrianschmidt: Ah, crap… I made a mistake when I removed my fixup. Never mind, I'll fix it 👍

adrianschmidt: ⚡

README.md
adrianschmidt, line 1: This is a pity:

Roboto:
image

This PR on macOS:
image

Definitely not worth loading Roboto for, but still a pity.

src/components/date-picker/flatpickr-adapter/flatpickr-adapter.scss
adrianschmidt: We don't need the `font-family: inherit;`, as far as I can see. It works correctly without it in the docs at least, and it's what I'd expect, since `inherit` is the default behaviour anyway, right?

README.md
Kiarokh: 😮 But that's a unicode. How would this happen? Does it mean that emojis won't show in the web-client neither?

adrianschmidt: I'm not sure, but I _think_ this is because that character technically isn't an emoji. I _think_ that emojis are recognised by the system, and always rendered using the same font, but this icon is older than emojis, and therefore the selected font is left in charge of how its rendered. Or _something_ like that anyway.

You have the same with ♥️. 👈 That isn't actually an emoji, it's a heart-character that existed before emojis. But once I post this comment that heart-character will look quite similar to the actual heart-_emoji_ that also exists, so I'll post a screenshot of this comment in edit mode too. Notice how the actual emoji doesn't change between edit mode and read mode, while the character _does_.

Char: ♥️
Emoji: ❤️

image

adrianschmidt: (I'm not sure _why_ the character changes between edit mode and read mode, since it looks like the same font is being used in both modes… 🤷‍♂️)

src/style/mixins.scss
```diff
@@ -340,3 +340,24 @@
...
+
+/**
+* This mixin creates a cross-browser font stack.
+* - `sans-serif` can be used for the UI of the components.
+* - `monospace` can be used for code.
+*
+* ⚠️ If we change the font stacks, we need to update
+* 1. the consumer documentation in `README.md`, and
+* 2. the CSS variables of `--kompendium-example-font-family`
+* in the `<style>` tag of `index.html`.
```
adrianschmidt, line 352: I added some fixups that places the addition of parts of this comment in the commits where the things referred to in the comment are introduced. For future traceability.

adrianschmidt (approved):

PR #2683, opened 2024-01-04
refactor(chip-set): correctly use MDC's HTML & classes

Kiarokh:
required for https://github.com/Lundalogik/crm-feature/issues/2723

We're inheriting styles from MDC, but since our HTML structure was not exactly similar to the one from
MDC's input-field, and some classes were missing,
then we had to make special styles to override what was coming from MDC, or to compensate for
what was missing, due to missing classes.

⚠️ to the reviewer: just make sure that the chipset (and picker) visually looks alright. It was impossible to break down this commit into several meaningful and easier to follow commits.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -390,13 +390,29 @@ export class ChipSet {
                     'force-invalid': this.isInvalid(),
                     'mdc-text-field--disabled': this.readonly || this.disabled,
                     'lime-text-field--readonly': this.readonly,
-                    'has-chips mdc-text-field--label-floating':
-                        this.value.length !== 0,
-                    'has-leading-icon': this.leadingIcon !== null,
+                    'has-chips': this.value.length !== 0,
+                    'mdc-text-field--label-floating':
+                        this.value.length !== 0 || this.editMode,
+                    'mdc-text-field--with-leading-icon':
+                        this.leadingIcon !== null,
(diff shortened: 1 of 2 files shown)


src/components/chip-set/chip-set.tsx
```diff
@@ -386,13 +386,29 @@ export class ChipSet {
...
+                        'mdc-text-field--required': this.required,
+                        'lime-notched-outline--notched': this.floatLabelAbove(),
+                    }}
+                    dropzone-tip={this.dropZoneTip()}
+                >
+                    <div class="mdc-notched-outline__leading" />
+                    {this.renderLabel()}
+                    <div class="mdc-notched-outline__trailing" />
+                </div>
+                {this.renderLeadingIcon()}
```
LucyChyzhova, line 411: This change causes the next issue 👇


https://github.com/Lundalogik/lime-elements/assets/50618208/97522a65-df1e-4c0a-830f-a79b566ddefa

Kiarokh:
we need to come back to this after we have:
1. used the new `limel-chip` in the chip-set
2. refactored `limel-input-field` so that it is not depending on MDC

Issue #2686, opened 2024-01-04
list: empty icons are rendered, when `icon.name` is `undefined`


Kiarokh:
<img width="1792" alt="image" src="https://github.com/Lundalogik/lime-elements/assets/35954987/eff9238a-e4fc-4c0c-a190-65fa8288a7ad">

PR #2688, opened 2024-01-04
More icon interface

Befkadu1:
fix https://github.com/Lundalogik/lime-elements/issues/2686
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu-list/menu-list-renderer.tsx
+++ b/src/components/menu-list/menu-list-renderer.tsx
@@ -139,7 +139,7 @@ export class MenuListRenderer {
                 data-index={index}
                 {...attributes}
             >
-                {item.icon ? this.renderIcon(this.config, item) : null}
+                {this.renderIcon(this.config, item)}
                 {this.renderText(item)}
                 {this.renderSubMenuIcon(item)}
                 {this.renderNotification(item)}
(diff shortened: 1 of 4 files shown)


Kiarokh (approved):

src/components/menu-list/menu-list-renderer.tsx
```diff
@@ -139,7 +139,7 @@ export class MenuListRenderer {
                 data-index={index}
                 {...attributes}
             >
-                {item.icon ? this.renderIcon(this.config, item) : null}
+                {this.renderIcon(this.config, item)}
```
adrianschmidt, line 142: 😬

This change makes the list crash if any item doesn't have `icon` specified 😞

Reports of several customers having this problem: https://forum.lime-crm.com/t/query-visualizer-not-working-with-newer-lime-elements/5723

The problem is that `renderIcon` uses `getIconName`, which in turn does:

```ts
if (typeof icon === 'object' && 'name' in icon) {
    return icon.name;
}
```

The problem is that `typeof null` returns `'object'`, so this tries to run `'name' in null`, which causes an error.

I think `get-icon-props.ts` is a pretty perfect example of how helper-functions can be broken out into a separate file, to make them testable. But unfortunately, there are no tests for that file 🙈 😄

I'll look into adding some tests and fixing the bug.

@Befkadu1 @Kiarokh

Kiarokh: 🙏

Issue #2539, opened 2023-10-16
feat(chip): add new component


Kiarokh:
## New feature motivation
Currently, Lime Elements offers chipsets
![image](https://github.com/Lundalogik/lime-elements/assets/35954987/3bd1d12a-a293-476a-8d9a-44fcfa5fee00)
![image](https://github.com/Lundalogik/lime-elements/assets/35954987/ca1dbce2-e4c5-412b-a83d-84be31140116)
But there is no stand-alone _Chip_ component to be used in other contexts.

## New feature description
Chips are fundamental parts of UI design. They can be used for many scenarios independently from a chipset.
But even within the chipset it'd be great to have a separate component for the chip part itself. This way, we would prevent repeating a lot of code in other places like Button-groups too.
![image](https://github.com/Lundalogik/lime-elements/assets/35954987/836a4152-9197-45d1-b455-ba10a42f5bfe)

## New feature implementation
In addition to onClick on chips, we also need them to be links in some scenarios. For example in the web-client, when a chip (inside a chipset) is meta-clicked (e.g. CMD + Click), the user wants to open that chip in a new tab. This is not possible today, ~as the component does not offer such interfaces~.

We would also like to have a contextual menu on the chip, which would help us not do hack like the one below, where we hacked a menu on the side of the chipset and `absolute`ly positioned it on top of it. The menu is about the chip, not the chipset, and we had no better way to do it:

![image](https://github.com/Lundalogik/lime-elements/assets/35954987/9871b485-5c36-4f38-93c9-c2ced5c1a5b5)

By implementing our own chips, we get a step closer to removing the MDC from Lime Elements.

PR #2707, opened 2024-01-12
Feat chip

Kiarokh:
fix https://github.com/Lundalogik/lime-elements/issues/2539

💡  **A Note to the reviewer:**
Each commit can practically become a new PR. I /we can make new PRs if you want to have an easier time for reviewing. I just did everything in one round, because based on previous experience of adding new components from scratch, it's way easier to add all the features you intend to add, all at once. This way, while work is in progress, you will just add fixups to previous commits where needed, instead of adding new `fix` or `refactor` commits several times after the first commit is merged. This way, the code becomes much cleaner.

⚠️ The ultimate goal of this PR is to create a component that can be replaced with the internal complexities of `limel-chip-set`. When reviewing, also cast an eye on `limel-chip-set` to be able to judge whether this component is good enough for that purpose, or whether changes are needed.

⏭️ **What's next**
It's good to know what's coming. I'll later make new PRs, implementing new features in this component.
1. Adding possibility of having a custom menu on the chip. This will enable us to create a more decent ui in `limec-relation-picker` instead of the crazy hack that we are doing:  
    image
    If there is an action menu && `removable={true}`, the `X` button will be one of the choices of the ellipsis menu `⋮` instead of being constantly displayed on the chip.
2. I might also add a `loading` state (which renders a spinner) on the chip, in another PR.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip/chip.e2e.ts
+++ b/src/components/chip/chip.e2e.ts
@@ -1,0 +1,57 @@
+import { newE2EPage } from '@stencil/core/testing';
+
+describe('limel-chip', () => {
+    let page;
+
+    beforeEach(async () => {
+        page = await newE2EPage();
+    });
+
+    it('renders with the correct text', async () => {
+        await page.setContent('<limel-chip text="Test chip"></limel-chip>');
+        await page.waitForChanges();
+
+        const element = await page.find('limel-chip >>> .chip');
+        expect(element.textContent).toEqual('Test chip');
+    });
+
+    it('renders as a link when the link prop is provided', async () => {
+        await page.setContent(
+            '<limel-chip text="Test chip" link="{ href: \'https://example.com\' }"></limel-chip>',
+        );
+
+        const element = await page.find('limel-chip');
+        element.setProperty('link', { href: 'https://example.com' });
+
+        await page.waitForChanges();
+
+        expect(element)
+            .toEqualHtml(`<limel-chip class="hydrated" language="en" link="{ href: 'https://example.com' }" text="Test chip">
+        <mock:shadow-root>
+          <a class="chip" href="https://example.com" tabindex="0">
+            <span class="text">
+              Test chip
+            </span>
+          </a>
+        </mock:shadow-root>
+      </limel-chip>`);
+    });
+
+    it('renders with a badge when the badge prop is provided', async () => {
+        await page.setContent(
+            '<limel-chip text="Test chip" badge="5"></limel-chip>',
+        );
+
+        const element = await page.find('limel-chip >>> limel-badge');
+        expect(element.getAttribute('label')).toEqual('5');
+    });
+
+    it('renders with a remove button when the removable prop is true', async () => {
+        await page.setContent(
+            '<limel-chip text="Test chip" removable></limel-chip>',
+        );
+
+        const element = await page.find('limel-chip >>> .remove-button');
+        expect(element).not.toBeNull();
+    });
+});
--- a/src/components/chip/chip.scss
+++ b/src/components/chip/chip.scss
@@ -1,0 +1,183 @@
+@use '../../style/mixins';
+
+/**
+* @prop --chip-max-width: Maximum width of the chip. Defaults to `10rem`. Keep in mind that the chips should not appear too big.
+*/
+
+:host(limel-chip) {
+    --limel-chip-height: 1.75rem;
+    --limel-chip-gap: 0.5rem;
+    isolation: isolate;
+    display: inline-flex;
+    align-items: center;
+}
+
+* {
+    box-sizing: border-box;
+}
+
+.chip {
+    all: unset;
+    position: relative;
+
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    gap: 0.25rem;
+
+    max-width: var(--chip-max-width, 10rem);
+    height: var(--limel-chip-height);
+
+    border-radius: var(--limel-chip-height);
+    font-size: 0.875rem;
+    padding: 0 0.125rem;
+
+    &:has(limel-icon) {
+        .text {
+            padding-left: 0.25rem;
+        }
+    }
+
+    &:not([disabled]) {
+        @include mixins.visualize-keyboard-focus;
+        @include mixins.is-elevated-inset-clickable;
+    }
+
+    &:has(limel-badge) {
+        padding-right: 0.375rem;
+
+        .text {
+            padding-right: 0;
+        }
+    }
+
+    &:has(+ .remove-button:hover) {
+        box-shadow: var(--shadow-depth-8-error);
+    }
+}
+
+:host(limel-chip[disabled]) {
+    .chip {
+        // Similar to `limel-button[disabled]`
+        color: rgba(var(--contrast-1600), 0.37);
+        background-color: rgba(var(--contrast-1600), 0.1);
+        box-shadow: none;
+    }
+}
+
+:host(limel-chip[readonly]) {
+    .chip {
+        box-shadow: 0 0 0 1px rgba(var(--contrast-800), 0.5);
+    }
+}
+
+:host(limel-chip[selected]) {
+    .chip {
+        box-shadow: var(--button-shadow-inset);
+
+        &:focus-visible {
+            box-shadow: var(--button-shadow-inset),
+                var(--shadow-depth-8-focused);
+        }
+
+        &:active {
+            box-shadow: var(--button-shadow-inset-pressed);
+        }
+    }
+    .text {
+        color: var(--mdc-theme-primary);
+    }
+}
+
+:host(limel-chip[removable]) {
+    .chip:not([disabled]) {
+        padding-right: calc(var(--limel-chip-height) + 0.125rem);
+
+        .text {
+            padding-right: 0;
+        }
+    }
+}
+
+:host(limel-chip[type='filter']) {
+    .chip {
+        border-top-left-radius: 0.5rem;
+        border-bottom-left-radius: 0.5rem;
+
+        &:after {
+            content: '';
+            transition:
+                background-color 0.4s ease 0.2s,
+                box-shadow 0.6s ease 0.2s;
+            box-sizing: border-box;
+            position: absolute;
+            bottom: 0.125rem;
+            left: 0.125rem;
+            width: 0.5rem;
+            height: 0.5rem;
+            border-radius: 50%;
+            background-color: rgb(var(--contrast-800), 0.8);
+            box-shadow: 0 0 0 1px rgb(var(--color-white)) inset;
+        }
+    }
+}
+
+:host(limel-chip[type='filter'][selected]) {
+    .chip {
+        &:after {
+            background-color: rgb(var(--color-green-default));
+            box-shadow:
+                0 0 0.375rem 0 rgb(var(--color-green-light)),
+                0 0 0 1px rgb(var(--color-white)) inset;
+        }
+    }
+}
+
+limel-icon {
+    flex-shrink: 0;
+    width: calc(var(--limel-chip-height) - 0.25rem);
+    height: calc(var(--limel-chip-height) - 0.25rem);
+    padding: 0.0625rem;
+}
+
+limel-badge {
+    pointer-events: none;
+}
+
+.text {
+    @include mixins.truncate-text;
+    line-height: 1.2;
+    padding: 0 0.5rem;
+}
+
+.trailing-button {
+    all: unset;
+    @include mixins.is-flat-clickable(
+        $color--hovered: rgb(var(--color-red-dark))
+    );
+
+    z-index: 1;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+
+    margin-left: calc(var(--limel-chip-height) * -1 + 0.125rem);
+    margin-right: 0.125rem;
+    width: calc(var(--limel-chip-height) - 0.25rem);
+    height: calc(var(--limel-chip-height) - 0.25rem);
+
+    border-radius: 50%;
+
+    svg {
+        transition:
+            color 0.2s ease,
+            transform 0.2s ease;
+        width: 1.25rem;
+    }
+
+    &:hover {
+        svg {
+            transform: scale(0.8);
+        }
+    }
+}
--- a/src/components/chip/chip.tsx
+++ b/src/components/chip/chip.tsx
@@ -1,0 +1,280 @@
+import {
+    Component,
+    Element,
+    Event,
+    EventEmitter,
+    h,
+    Host,
+    Prop,
+} from '@stencil/core';
+import { Icon, Languages, Link } from '../../interface';
+import { getIconName } from '../icon/get-icon-props';
+import {
+    makeEnterClickable,
+    removeEnterClickable,
+} from '../../util/make-enter-clickable';
+import translate from '../../global/translations';
+import {
+    BACKSPACE,
+    BACKSPACE_KEY_CODE,
+    DELETE,
+    DELETE_KEY_CODE,
+} from '../../util/keycodes';
+import { Chip as OldChipInterface } from '../chip-set/chip.types';
+
+interface ChipInterface extends Omit<OldChipInterface, 'id' | 'badge'> {
+    badge?: string | number;
+}
+
+/**
+ * Chips and buttons are both interactive elements in UI design,
+ * but they serve different purposes and are used in different contexts.
+ *
+ * :::warning
+ * Do not use the chip component carelessly, as an alternative for
+ * [`limel-button`](#/component/limel-button/) in the UI design!
+ *
+ * **Buttons:**
+ * Buttons are used to trigger actions. They are typically used to
+ * submit forms, open dialogs, initiate a process, or perform any action
+ * that changes the state of the application.
+ * Buttons' labels usually contain action words, in other words, the labels is
+ * a _verb in imperative mood_ such as "Submit" or "Delete".
+ * Buttons are placed in areas where it's clear they will initiate
+ * an action when clicked.
+ *
+ * **Chips:**
+ * Chips however are elements which may look like buttons, but they are
+ * representing choices, filters, or tags, in a small block
+ * or clearly bundled into a group. Chips are rarely used alone in the
+ * user interface.
+ * They are often used in a so called "chip-set", or placed together in
+ * a section of the UI, where the user can expect more than one chip to be present.
+ *
+ * For example, a chip may represent a filter in a filter bar, or a tag in a tag list,
+ * or an item in a shopping list.
+ * Clicking a chip can also trigger an action, for example toggling a filter ON or OFF,
+ * or opening a page with all posts tagged with the tag represented by the chip,
+ * or navigating to a page with more information about the item in the shopping list.
+ * :::
+ *
+ * @private
+ * @exampleComponent limel-example-chip-button
+ * @exampleComponent limel-example-chip-link
+ * @exampleComponent limel-example-chip-icon-colors
+ * @exampleComponent limel-example-chip-badge
+ * @exampleComponent limel-example-chip-filter
+ * @exampleComponent limel-example-chip-removable
+ * @exampleComponent limel-example-chip-aria-role
+ */
+@Component({
+    tag: 'limel-chip',
+    shadow: true,
+    styleUrl: 'chip.scss',
+})
+export class Chip implements ChipInterface {
+    /**
+     * Defines the language for translations.
+     * Will translate the translatable strings on the components.
+     */
+    @Prop({ reflect: true })
+    public language: Languages = 'en';
+
+    /**
+     * Label displayed on the chip
+     */
+    @Prop({ reflect: true })
+    public text: string;
+
+    /**
+     * Icon of the chip.
+     */
+    @Prop()
+    public icon: string | Icon;
+
+    /**
+     * If supplied, the chip will become a clickable link.
+     */
+    @Prop()
+    public link?: Omit<Link, 'text'>;
+
+    /**
+     * The value of the badge, displayed on the chip.
+     */
+    @Prop({ reflect: true })
+    public badge?: string | number;
+
+    /**
+     * Set to `true` to disable the chip.
+     */
+    @Prop({ reflect: true })
+    public disabled = false;
+
+    /**
+     * Set to `true` to render the chip as a static UI element.
+     * Useful when the parent component has a `readonly` state.
+     */
+    @Prop({ reflect: true })
+    public readonly = false;
+
+    /**
+     * Set to `true` to visualize the chip in a "selected" state.
+     * This is typically used when the chip is used in a chip-set
+     * along with other chips.
+     */
+    @Prop({ reflect: true })
+    public selected = false;
+
+    /**
+     * Set to `true` to render a remove button on the chip.
+     */
+    @Prop({ reflect: true })
+    public removable = false;
+
+    /**
+     * Set to `filter` to render the chip with a distinct style
+     * suitable for visualizing filters.
+     */
+    @Prop({ reflect: true })
+    public type: 'filter';
+
+    /**
+     * Fired when clicking on the remove button of a `removable` chip.
+     */
+    @Event()
+    public remove: EventEmitter<void>;
+
+    @Element()
+    private host: HTMLLimelChipElement;
+
+    public componentWillLoad() {
+        makeEnterClickable(this.host);
+    }
+
+    public disconnectedCallback() {
+        removeEnterClickable(this.host);
+    }
+
+    public render() {
+        return (
+            <Host onClick={this.filterClickWhenDisabled}>
+                {this.link ? this.renderAsLink() : this.renderAsButton()}
+            </Host>
+        );
+    }
+
+    private renderAsButton = () => {
+        return [
+            <button
+                class="chip"
+                role="button"
+                disabled={this.disabled || this.readonly}
+                onKeyDown={this.handleDeleteKeyDown}
+            >
+                {this.renderIcon()}
+                {this.renderLabel()}
+                {this.renderBadge()}
+            </button>,
+            this.renderRemoveButton(),
+        ];
+    };
+
+    private renderAsLink = () => {
+        return [
+            <a
+                class="chip"
+                href={this.link.href}
+                title={this.link.title}
+                target={this.link.target}
+                aria-disabled={this.disabled || this.readonly}
+                tabindex={this.disabled || this.readonly ? -1 : 0}
+                onKeyDown={this.handleDeleteKeyDown}
+            >
+                {this.renderIcon()}
+                {this.renderLabel()}
+                {this.renderBadge()}
+            </a>,
+            this.renderRemoveButton(),
+        ];
+    };
+
+    private renderLabel = () => {
+        return <span class="text">{this.text}</span>;
+    };
+
+    private renderIcon() {
+        const icon = getIconName(this.icon);
+
+        if (!icon) {
+            return;
+        }
+
+        return (
+            <limel-icon
+                badge={true}
+                name={icon}
+                style={{
+                    color: `${(this.icon as Icon)?.color}`,
+                    'background-color': `${
+                        (this.icon as Icon)?.backgroundColor
+                    }`,
+                }}
+            />
+        );
+    }
+
+    private renderBadge() {
+        if (!this.badge) {
+            return;
+        }
+
+        return <limel-badge label={this.badge} />;
+    }
+
+    private renderRemoveButton() {
+        if (!this.removable || this.readonly || this.disabled) {
+            return;
+        }
+
+        const svgData =
+            '<svg viewBox="0 0 32 32" xmlns="http://www.w3.org/2000/svg"><path fill="none" stroke="currentColor" stroke-width="2" d="m8 8 16 16M24 8 8 24"/></svg>';
+
+        return (
+            <button
+                class="trailing-button remove-button"
+                tabIndex={-1}
+                aria-label={this.removeChipLabel}
+                innerHTML={svgData}
+                onClick={this.handleRemoveClick}
+            />
+        );
+    }
+
+    private filterClickWhenDisabled = (e) => {
+        if (this.disabled || this.readonly) {
+            e.preventDefault();
+        }
+    };
+
+    private handleRemoveClick = (event: MouseEvent | KeyboardEvent) => {
+        event.stopPropagation();
+        this.remove.emit();
+    };
+
+    private handleDeleteKeyDown = (event: KeyboardEvent) => {
+        if (!this.removable) {
+            return;
+        }
+
+        const keys = [DELETE, BACKSPACE];
+        const keycodes = [DELETE_KEY_CODE, BACKSPACE_KEY_CODE];
+
+        if (keys.includes(event.key) || keycodes.includes(event.keyCode)) {
+            this.handleRemoveClick(event);
+        }
+    };
+
+    private removeChipLabel = (): string => {
+        return translate.get('chip-set.remove-chip', this.language);
+    };
+}
--- a/src/components/chip/examples/chip-aria-role.tsx
+++ b/src/components/chip/examples/chip-aria-role.tsx
@@ -1,0 +1,57 @@
+import { Component, h } from '@stencil/core';
+
+/**
+ * Correct usage of ARIA roles
+ *
+ * Chips represent choices, filters, or tags, organized in a block or bundled into a group.
+ * While sighted users see the visually bundled group of chips in a well-designed UI,
+ * screen reader users only hear the chip text, one at a time.
+ * This can make it difficult for users of assistive technologies to understand
+ * the context of the chip.
+ *
+ * To provide an accessible experience, it's important to place the chips in
+ * a semantically correct structure, such as a list or a table,
+ * or properly use ARIA roles on the chip and its container.
+ *
+ * In this example, we demonstrate how to use ARIA roles to improve accessibility for chips.
+ * However, it's recommended to read up on the subject to fully understand the
+ * implications of ARIA roles.
+ *
+ * For more information on ARIA roles, refer to the
+ * [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles).
+ */
+@Component({
+    tag: 'limel-example-chip-aria-role',
+    shadow: true,
+    styleUrl: 'limel-example-chip-aria-role.scss',
+})
+export class ChipAriaRoleExample {
+    public render() {
+        return (
+            <div role="grid" aria-label="Filter bar" aria-colcount="3">
+                <div role="rowgroup">
+                    <div role="row">
+                        <limel-chip
+                            role="gridcell"
+                            aria-colindex="1"
+                            text="Age > 20"
+                            removable={true}
+                        />
+                        <limel-chip
+                            role="gridcell"
+                            aria-colindex="2"
+                            text="Gender = Female"
+                            removable={true}
+                        />
+                        <limel-chip
+                            role="gridcell"
+                            aria-colindex="3"
+                            text="Income > $3000 / mo."
+                            removable={true}
+                        />
+                    </div>
+                </div>
+            </div>
+        );
+    }
+}
--- a/src/components/chip/examples/chip-button.tsx
+++ b/src/components/chip/examples/chip-button.tsx
@@ -1,0 +1,74 @@
+import { Component, State, h } from '@stencil/core';
+
+/**
+ * Chip as button
+ * Typically, a chip is used to trigger an action or act as an input element.
+ * This is why the component generates a `<button>` element in the DOM to give
+ * a more semantically correct clues to assistive technologies.
+ *
+ * To trigger these actions, you will only need to handle the `onClick`
+ * event on the component.
+ */
+@Component({
+    tag: 'limel-example-chip-button',
+    shadow: true,
+})
+export class ChipButtonExample {
+    @State()
+    private disabled: boolean = false;
+
+    @State()
+    private selected: boolean = false;
+
+    @State()
+    private readonly: boolean = false;
+
+    public render() {
+        return [
+            <limel-chip
+                text="FunnyCats"
+                icon="hashtag"
+                onClick={this.onClick}
+                disabled={this.disabled}
+                selected={this.selected}
+                readonly={this.readonly}
+            />,
+            <limel-example-controls>
+                <limel-checkbox
+                    checked={this.disabled}
+                    label="Disabled"
+                    onChange={this.setDisabled}
+                />
+                <limel-checkbox
+                    checked={this.readonly}
+                    label="Readonly"
+                    onChange={this.setReadonly}
+                />
+                <limel-checkbox
+                    checked={this.selected}
+                    label="Selected"
+                    onChange={this.setSelected}
+                />
+            </limel-example-controls>,
+        ];
+    }
+
+    private onClick() {
+        console.log('Chip is clicked.');
+    }
+
+    private setDisabled = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.disabled = event.detail;
+    };
+
+    private setSelected = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.selected = event.detail;
+    };
+
+    private setReadonly = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.readonly = event.detail;
+    };
+}
--- a/src/components/chip/examples/chip-filter.tsx
+++ b/src/components/chip/examples/chip-filter.tsx
@@ -1,0 +1,61 @@
+import { Component, State, h } from '@stencil/core';
+
+/**
+ * Chip as filter
+ * Chips are great candidates to visualize active filters.
+ * However, as chips are used for other purposes as well,
+ * we need to make sure that the user understands that the chip is a filter,
+ * just by the look of it.
+ *
+ * By setting the `type` to `filter`, the chip will be rendered with a distinct style
+ * suitable for visualizing filters.
+ *
+ * :::note
+ * In this mode, clicking on the chip should also toggle its `selected` state.
+ * :::
+ */
+@Component({
+    tag: 'limel-example-chip-filter',
+    shadow: true,
+    styleUrl: 'chip-badge.scss',
+})
+export class ChipFilterExample {
+    @State()
+    private selected: { [id: string]: boolean } = {};
+
+    public render() {
+        return [
+            <limel-chip
+                id="1"
+                text="photos"
+                icon="image_file"
+                type="filter"
+                onClick={() => this.handleClick('1')}
+                selected={this.selected['1']}
+            />,
+            <limel-chip
+                id="2"
+                text="videos"
+                icon="video_file"
+                type="filter"
+                onClick={() => this.handleClick('2')}
+                selected={this.selected['2']}
+            />,
+            <limel-chip
+                id="2"
+                text="audios"
+                icon="audio_file"
+                type="filter"
+                onClick={() => this.handleClick('3')}
+                selected={this.selected['3']}
+            />,
+        ];
+    }
+
+    private handleClick = (id: string) => {
+        this.selected = {
+            ...this.selected,
+            [id]: !this.selected[id],
+        };
+    };
+}
--- a/src/components/chip/examples/chip-link.tsx
+++ b/src/components/chip/examples/chip-link.tsx
@@ -1,0 +1,34 @@
+import { Component, h } from '@stencil/core';
+import { Link } from '@limetech/lime-elements';
+
+/**
+ * Chip as hyperlink
+ * For accessibility and usability alike, if clicking on a chip should
+ * result in any kind of navigation, it is preferable to use a link,
+ * rather than a button.
+ *
+ * That way, the user can choose to, for example, open the link in a new tab.
+ * For this reason, we suggest always providing a Link with
+ * the URL representing the target state of the navigation.
+ */
+@Component({
+    tag: 'limel-example-chip-link',
+    shadow: true,
+})
+export class ChipLinkExample {
+    link: Link = {
+        href: 'https://github.com',
+        title: 'Open Github',
+        target: '_blank',
+    };
+
+    public render() {
+        return (
+            <limel-chip
+                text="Github"
+                icon="github_copyrighted"
+                link={this.link}
+            />
+        );
+    }
+}
--- a/src/components/chip/examples/chip-badge.scss
+++ b/src/components/chip/examples/chip-badge.scss
@@ -1,0 +1,5 @@
+:host {
+    display: flex;
+    flex-wrap: wrap;
+    gap: 0.5rem;
+}
--- a/src/components/chip/examples/chip-badge.tsx
+++ b/src/components/chip/examples/chip-badge.tsx
@@ -1,0 +1,19 @@
+import { Component, h } from '@stencil/core';
+
+/**
+ * Chip with a badge
+ * Chips can display a badge with a number or a short text.
+ */
+@Component({
+    tag: 'limel-example-chip-badge',
+    shadow: true,
+    styleUrl: 'chip-badge.scss',
+})
+export class ChipBadgeExample {
+    public render() {
+        return [
+            <limel-chip text="Batman" icon="batman_old" badge={2005} />,
+            <limel-chip text="Batman" icon="batman_new" badge="NEW" />,
+        ];
+    }
+}
--- a/src/components/chip/examples/chip-icon-color.tsx
+++ b/src/components/chip/examples/chip-icon-color.tsx
@@ -1,0 +1,21 @@
+import { Component, h } from '@stencil/core';
+
+/**
+ * Icon color
+ * Using the `Icon` interface, you can specify colors for the icon.
+ */
+@Component({
+    tag: 'limel-example-chip-icon-colors',
+    shadow: true,
+})
+export class ChipIconColorsExample {
+    public render() {
+        const icon = {
+            name: 'filled_star',
+            color: 'rgb(var(--color-yellow-default))',
+            backgroundColor: 'rgb(var(--color-blue-dark))',
+        };
+
+        return <limel-chip text="Golden star" icon={icon} />;
+    }
+}
--- a/src/components/chip/examples/chip-remove.tsx
+++ b/src/components/chip/examples/chip-remove.tsx
@@ -1,0 +1,43 @@
+import { Component, State, h } from '@stencil/core';
+
+/**
+ * Removable chips
+ * Chips can display a remove button,
+ * when their `removable` prop is set to `true`.
+ *
+ * This is typically used when the chip is used in a chip-set,
+ * where each chip visualizes a chosen option.
+ *
+ * :::tip
+ * When the chip is focused using the keyboard, the user can press
+ * the <kbd>Delete</kbd> or <kbd>Backspace</kbd> keys to
+ * trigger the same remove `event`.
+ * :::
+ */
+@Component({
+    tag: 'limel-example-chip-removable',
+    shadow: true,
+})
+export class ChipRemoveExample {
+    @State()
+    private removeButtonClicked: boolean = false;
+
+    public render() {
+        return [
+            <limel-chip
+                text="My filter"
+                badge={123}
+                removable={true}
+                onRemove={this.handleRemove}
+            />,
+            <limel-example-value
+                label="Remove"
+                value={this.removeButtonClicked}
+            />,
+        ];
+    }
+
+    private handleRemove = () => {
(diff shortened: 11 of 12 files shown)


src/components/chip/chip.scss
```diff
@@ -103,6 +103,40 @@
     }
 }
 
+:host(limel-chip[type='filter']) {
```
Kiarokh, line 106: this is an intentional choice, and gives us a new UI 👇 
image


which will replace this old UI 👇  when we implement this component in the chip-set
image

src/components/chip/chip.tsx
```diff
@@ -123,6 +124,13 @@ export class Chip {
     @Prop({ reflect: true })
     public removable = false;
 
+    /**
+     * Set to `filter` to render the chip with a distinct style
+     * suitable for visualizing filters.
+     */
+    @Prop({ reflect: true })
+    public type: 'filter';
+
```
Kiarokh, line 133: Almost all the props above exist in src/components/chip-set/chip.types.ts , which are used in `limel-chip-set` today. However, I think it's better to have the props of this new component inside it.

Also, in src/components/chip-set/chip.types.ts, there is a `value` prop, which I don't really know why it is needed.

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,154 @@
...
+    tag: 'limel-chip',
+    shadow: true,
+    styleUrl: 'chip.scss',
+})
+export class Chip {
+    /**
+     * ID of the chip. Must be unique.
+     */
+    @Prop()
+    id: number | string;
```
Kiarokh, line 74: maybe this should be an optional prop instead

src/components/chip/chip.e2e.ts
```diff
@@ -0,0 +1,45 @@
...
+
+    it('should render as a link when the link prop is provided', async () => {
+        await page.setContent(
+            '<limel-chip text="Test chip" link="{ href: \'https://example.com\' }"></limel-chip>',
+        );
+        await page.waitForChanges();
+
+        const element = await page.find('limel-chip >>> a.chip');
+        expect(element.getAttribute('href')).toEqual('https://example.com');
+    });
```
Kiarokh, line 26: image
Don't know how to fix this

src/components/chip/chip.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+        expect(page.root).toEqualHtml(`
+            <limel-chip text="Test chip" link="{ href: 'https://example.com' }">
+                <mock:shadow-root>
+                    <a class="chip" href="https://example.com">
+                        <span class="text">Test chip</span>
+                    </a>
+                </mock:shadow-root>
+            </limel-chip>
+        `);
+    });
```
Kiarokh, line 41: don't know how to fix this:

```
 FAIL  src/components/chip/chip.e2e.ts
  ● limel-chip › should render as a link when the link prop is provided

    TypeError: Cannot read properties of null (reading 'getAttribute')

      23 |
      24 |         const element = await page.find('limel-chip >>> a.chip');
    > 25 |         expect(element.getAttribute('href')).toEqual('https://example.com');
         |                        ^
      26 |     });
      27 |
      28 |     it('should render with a badge when the badge prop is provided', async () => {

      at Object.<anonymous> (src/components/chip/chip.e2e.ts:25:24)
      ```

File: src/components/chip/chip.scss [hunk 6]
```diff
@@ -0,0 +1,69 @@
+@use '../../style/mixins';
+
+/**
+* @prop --chip-max-width: Maximum width of the chip. Defaults to `10rem`. Keep in mind that the chips should not appear too big.
+*/
+
+:host(limel-chip) {
+    --limel-chip-height: 1.75rem;
```
Kiarokh, line 8: 💡 _this is an internal variable, starting with `--limel-` prefix!_

The current `mdc-chip` is `32px` tall. Also, `limel-button` is `36px` tall. I intentionally want the new chips to be `28px` tall! 

**Why?**

1. I don't want consumers to use chips instead of buttons, or put them beside a button in the UI.
2. I want to make all input fields smaller in height. All should become `36px`, like `limel-button`. This means, in the chip-set `type="input"`, we will need more space around the chips. And this will give us that height.
     
    However, before resizing all input fields, we can use this internal CSS var to render the chips larger in the chip-set.

src/components/chip/chip.spec.ts
LawrenceBorst: Is the shadow root interfering maybe? One of the things it does, apart from namespacing CSS, is hide the whole shadow tree from queries to elements above it.

But you could access it like `el = page.root.querySelector('limel-chip-set').shadowRoot` and then try to find the anchor, maybe? I'm just guessing though 😛

Kiarokh: I don't know and don't want to know! Freaking tests. Just fix it! 🤣

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,279 @@
...
+    DELETE,
+    DELETE_KEY_CODE,
+} from '../../util/keycodes';
+
+/**
+ * Chips and buttons are both interactive elements in UI design,
+ * but they serve different purposes and are used in different contexts.
+ *
+ * :::warning
+ * Do not use the chip component, carelessly as an alternative for
```
adrianschmidt, line 28:
```suggestion
 * Do not use the chip component carelessly, as an alternative for
```

`fixup! chore(chip): add new private component`

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,279 @@
...
+ * Buttons are used to trigger actions. They are typically used to
+ * submit forms, open dialogs, initiate a process, or perform any action
+ * that changes the state of the application.
+ * Buttons' labels usually contain action words, in other words, the labels is
+ * a _verb in imperative mood_ such as "Submit" or "Delete".
+ * Buttons are placed in areas where it's clear they will initiate
+ * an action when clicked.
+ *
+ * **Chips:**
+ * Chips are compact elements that represent an input, attribute, or action;
```
adrianschmidt, line 43: I'd suggest that we probably don't mention "action" here. I understand the difference between "triggering an action" and "representing an action as a 'thing'", but I'm worried it's a little too confusing in this fairly compact context.

(I don't know, maybe it's not a problem, but it made me go "hang on, didn't he just say… ah, wait, no, I understand", and that makes me suspect that someone less familiar with the whole thing might be a lot more confused for a lot longer 😅)

src/components/chip/chip.tsx
adrianschmidt: Hmm… yeah… I'm slightly conflicted… I'm pretty confident we'll want a `Chip` type* in the future as well, because we need to be able to pass around arrays of them and stuff. And since it's annoying to have to define the same thing twice, and keep stuff in sync, that would seem like an argument for having the limel-chip component accept a Chip. And when used in something like the chip-set, it's surely going to be more convenient to just pass the `Chip` straight into the component. But when used stand-alone, it's probably nicer not to have to pass in a `Chip`, but pass the properties separately… 🤔

Or, maybe… perhaps that's just when we think about using limel-chip in the kind of way that you use a button. Which is exactly the way we _don't_ want people to use it. If you actually use it to represent entities, you're much more likely to have an array of things, and then you can just map those things to Chips, and, most of all, you can have a single listener for changes, that expects an updated `Chip`, rather than have one listener for each property.

I might flip-flop on this over the course of reviewing this, but for now, I'm leaning heavily towards accepting a `Chip`, rather then separate properties.

(* pedant's corner: the current one is an interface, I'll call it "type" when I'm talking about something that defines a data structure, as I tend to think of "interface" as "something that can be implemented")

src/components/chip/chip.tsx
adrianschmidt: Now I'm _very_ confused 😅

image

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,279 @@
...
+    public text: string;
+
+    /**
+     * Icon of the chip.
+     */
+    @Prop()
+    public icon: string | Icon;
+
+    /**
+     * If supplied, the chip will become a clickable links.
```
adrianschmidt, line 89:
```suggestion
     * If supplied, the chip will become a clickable link.
```

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,279 @@
...
+     */
+    @Prop()
+    link?: Omit<Link, 'text'>;
```
adrianschmidt, line 92:
```suggestion
    public link?: Omit<Link, 'text'>;
```

`fixup! chore(chip): add new private component`

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,279 @@
...
+        this.removeEnterClickable();
+    }
+
+    public render() {
+        if (this.link) {
+            return this.renderAsLink();
+        }
+
+        return this.renderAsButton();
+    }
```
adrianschmidt, line 161: Is there a positive reason you want to render a button when there is no link to render, or is it a matter of running into problems using an anchor tag when there is no link to use for the `href`?

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,279 @@
...
+                class="chip"
+                href={this.link.href}
+                title={this.link.title}
+                aria-disabled={this.disabled || this.readonly}
+                tabindex={this.disabled ? -1 : 0}
+                onClick={(e) => {
+                    if (this.disabled) {
+                        e.preventDefault();
+                    }
+                }}
```
adrianschmidt, line 191: Using arrow-functions in templates (or, rather, _defining_ arrow-functions in templates) is bad for performance, since it creates a new function every time the component is re-rendered. So this should be extracted to something like:

```tsx
                onClick={this.filterClick}
    …

    private filterClick = (e: Event) => {
        if (this.disabled) {
            e.preventDefault();
        }
    };
```

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,279 @@
...
+     * inside the breadcrumbs.
+     */
+    @Event()
+    public remove: EventEmitter<Chip>;
+
+    @Element()
+    private host: HTMLLimelChipElement;
+
+    private button: HTMLButtonElement;
+    private anchor: HTMLAnchorElement;
```
adrianschmidt, line 145: These don't seem to be defined anywhere?

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,279 @@
...
+     */
+    @Prop({ reflect: true })
+    public type: 'filter';
+
+    /**
+     * Fired when clicking on buttons (not links!)
+     * inside the breadcrumbs.
+     */
+    @Event()
+    public remove: EventEmitter<Chip>;
```
adrianschmidt, line 139: Surely the description for this event must be a copy-paste mistake or something? 😅

src/components/chip/examples/chip-link.tsx
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Chip as hyperlink
+ * A chip can be used to navigate to a different page, pane of the application
+ * or to an external site. To get the chip to act as a hyperlink, you need to
+ * provide a `Link`.
+ *
+ * This way, the component will automatically become a native HTML link.
+ * This gives the users the possibility of interacting with links
+ * in a natural way, for instance they can open any of the previous
+ * pages in a new browser tab. This can also have other accessibility benefits.
```
adrianschmidt, line 12: > to navigate to a different page, pane of the application or to an external site

Nitpick, but this sounds a bit clunky, I think, especially "pane of the application".  I understand that refers to the sidepanel in the case of Lime CRM, but without that reference, it would just seem a bit odd. It's a little unnecessary to describe what _kind_ of navigation can be performed with a link, as well. Like, "we know, it's called _the web_" 😅

> for instance they can open any of the previous pages in a new browser tab

🤨 😅

Overall, I would probably rephrase this whole section a bit. I'd like it more imperative (is that the right word for it?). Something like:

> For accessibility and usability alike, if clicking on a chip should result in any kind of navigation, it is preferable to use a link, rather than a button. That way, the user can choose to, for example, open the link in a new tab.
>
> For this reason, we suggest always providing a `Link` with the URL representing the target state of the navigation, if such a URL exists.

Although less pompously worded than I managed here perhaps… 🙈

src/components/chip/examples/chip-link.tsx
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * :::note
+ * Clicking a link will open it in current window by default,
+ * and this reloads the entire webpage.
+ * To avoid reloading the whole application (in the context of a single-page apps),
+ * you might want to handle the navigation with your application's router,
+ * :::
```
adrianschmidt, line 19: I don't know if we need this note here. I don't think it's going to help anyone. For anyone building stuff inside Lime CRM, the navigation interceptor should take care of this for them, and for anyone building their own app, if they don't know they need to consider this, talking about using the application's router probably isn't going to help them…

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+        return this.renderAsButton();
+    }
+
+    private renderAsButton = () => {
+        return [
+            <button
+                class="chip"
+                role="button"
+                disabled={this.disabled || this.readonly}
```
adrianschmidt, line 168: This doesn't make sense to me. When the component is readonly, there shouldn't be a remove-button rendered, but the component should still be clickable, right?

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,279 @@
...
+     * If supplied, the chip will become a clickable links.
+     */
+    @Prop()
+    link?: Omit<Link, 'text'>;
+
+    /**
+     * The value of the badge, displayed on the chip.
+     */
+    @Prop()
+    badge?: string | number;
```
adrianschmidt, line 98:
```suggestion
    public badge?: string | number;
```

```
fixup! chore(chip): add props and styles for `badge`
```

src/components/chip/chip.tsx
```diff
@@ -166,6 +175,14 @@ export class Chip {
         );
     }
 
+    private renderBadge() {
+        if (!this.badge) {
+            return;
+        }
```
adrianschmidt, line 229: It seems like a feature, rather than a bug, but just to make sure:

It's intentional (or a happy coincidence 😄) that `badge={0}` renders no badge, I assume?

src/components/chip/examples/chip-link.tsx
```diff
@@ -0,0 +1,42 @@
...
+        target: '_blank',
+    };
+
+    public render() {
+        return (
+            <limel-chip
+                id="1"
+                text="Github"
+                icon="github_copyrighted"
+                link={this.link}
```
adrianschmidt, line 38: We don't use the `target` property from the `Link`. I think we should 😊

src/components/chip/examples/chip-button.tsx
```diff
@@ -0,0 +1,72 @@
...
+                    checked={this.selected}
+                    label="Selected"
+                    onChange={this.setSelected}
+                />
+            </limel-example-controls>,
+        ];
+    }
+
+    private onClick() {
+        console.log('Chip is clicked.');
```
adrianschmidt, line 55: The current behaviour makes the chip in this example look _very_ much like a button, that you might want to use for triggering actions, just like we tell people they should _not_. I would suggest that we add a flip of the `selected` state here as well, to avoid that.

```suggestion
        console.log('Chip is clicked.');
        this.selected = !this.selected;
```

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,279 @@
...
+
+        if (keys.includes(event.key) || keycodes.includes(event.keyCode)) {
+            this.handleRemoveClick(event);
+        }
+    };
+
+    private removeEnterClickable() {
+        const element = this.button ?? this.anchor;
+        removeEnterClickable(element);
+    }
```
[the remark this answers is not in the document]
Kiarokh: this was just a copy/paste from "action bar item" if I'm not hallucinating. 🤔  If this is a problem, it should probably be fixed in several places.

src/components/chip/chip.tsx
Kiarokh: > using limel-chip in the kind of way that you use a button. Which is exactly the way we don't want people to use it.

I absolutely want people to be able to use it as an stand-alone component too. But if they need a chip-set, it's for sure better to use `limel-chip-set` to get more relevant features.

Kiarokh: My feeling is that the consumer component (e.g. chip-set) can import the stuff from `chip.types.ts`, but that should perhaps be located in this new component's folder instead.

src/components/chip/chip.tsx
Kiarokh: me too 😅

src/components/chip/chip.tsx
Kiarokh: I don't understand the question :/

src/components/chip/chip.scss
adrianschmidt: Gotta say, this new filter style looks great, and the transition when selecting and deselecting the chips is _beautiful_! Great work! 🙌

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,279 @@
...
+                class="chip"
+                href={this.link.href}
+                title={this.link.title}
+                aria-disabled={this.disabled || this.readonly}
+                tabindex={this.disabled ? -1 : 0}
+                onClick={(e) => {
+                    if (this.disabled) {
+                        e.preventDefault();
+                    }
+                }}
```
Kiarokh, line 191:
```suggestion
```

Good point, this was also an over-engineering. The combination of the tab-index (the line before) and the CSS styles for `[disabled]` already do the job. What do you think if we remove it totally?

src/components/chip/chip.tsx
Kiarokh: I don't remember really. But some faint memory tells me that it was intentional. :)

src/components/chip/chip.tsx
adrianschmidt: That seems reasonable to me.

`id` is not a great name when it's an `@Prop`, since it will conflict with the native `id` attribute. So, if we don't switch to accepting an object of type `Chip`, we should rename this property.

src/components/chip/chip.tsx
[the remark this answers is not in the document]
Kiarokh: by the way, just some clarifications, we only want to prevent repeatedly triggered actions when using the <kbd>Enter</kbd> on keyboard, when the host is clicked. Because the consumer will add the `onClick` on the host element, even though the component contains a `<button>` element in itself.

However, when the component renders a hyperlink (an `<a>` element), there is hopefully no onClick defined.

src/components/chip/chip.e2e.ts
adrianschmidt: The fact that the actual response has `tabindex="0"` should be easy to fix by just adding that to the expected output. (It looks like this comment is placed on the wrong test? If it's not, then I have no idea, to be honest.)

The fact that the actual response is _missing` the `href="https://example.com"` on the other hand, indicates either that the functionality we are testing for doesn't work (unlikely, since it seems to work in the PR docs), or that we need to wait for it to re-render or something.

I'm sure I or someone else can have a look 👍

src/components/chip/chip.tsx
Kiarokh: I don't think we want it to be clickable. Think about a readonly chip-set. 🤔 Maybe I'm wrong. But throughout our components, we have used `readonly` to render some sort of "visualization-only" mode. More info here: https://github.com/Lundalogik/lime-elements/pull/2412#pullrequestreview-1541249268

src/components/chip/chip.tsx
adrianschmidt: Do you specifically _want_ to use `<button>` when there is no url, or have you used `<button>` when there isn't a url simply because you had to for some practical reason, and you would be just as happy to use `<a>` in both cases otherwise?

I felt like it was a little complex and annoying to have to handle rendering either an `<a>` or a `<button>`, so I was wondering if we could perhaps make it use an `<a>` even when no url was supplied. But perhaps it's better to use a `<button>` then? 🤔

src/components/chip/chip.tsx
Kiarokh: see https://github.com/Lundalogik/lime-elements/pull/2707/files#r1453501618

src/components/chip/chip.tsx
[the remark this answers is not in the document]
adrianschmidt: I think we are leaking memory here, because I don't think the listeners we set with `makeEnterClickable` is ever removed, since the two functions are called on different elements. That's probably a problem in breadcrumbs too then, yes.

src/components/chip/chip.tsx
adrianschmidt: Hmm… oh… yeah… I guess it depends on what functionality is tied to clicking the chip… if it's a filter-chip, then no, it shouldn't be possible to click it, because it shouldn't be possible to change the chip's `selected` state. But if the chip represents a related object in a feed-item, it should still be possible to click on it to navigate to the related object, right?

Although, you perhaps don't consider that "readonly". That's just "the normal way chips are, when they don't have `removable=true`"? 🤔

src/components/chip/chip.tsx
adrianschmidt: Screen readers might still be able to focus the element… 🤔 (They _should_ be, since conveying that there's something there but that it's disabled is important, as far as I know.)

src/components/chip/chip.tsx
Kiarokh: Do we even need this? What's this `value?: T;` that the chip.types.ts has? Should it be that one instead? 

Note that we both `id` there too!
```ts
import { Icon } from '../../interface';

export interface Chip<T = any> {
    /**
     * ID of the chip. Must be unique.
     */
    id: number | string;

// ···

    /**
     * Value of the chip.
     */
    value?: T;

// ···
}

```

src/components/chip/chip.tsx
Kiarokh: Aha. Yes. for accessibility reasons, I need to render a `<button type="button"`, when the consumer doesn't provide a link.

src/components/chip/chip.tsx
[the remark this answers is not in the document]
Kiarokh: @LawrenceBorst you know this best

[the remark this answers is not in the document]
LawrenceBorst: I think it needs to be `this.host` as well :), but I'll play around with it. Maybe moving forward we should give the consumer a random `id` which they own with which they can destruct the object in the service, and use information hiding to avoid other consumers peeking around for other ids.

src/components/chip/chip.tsx
Kiarokh: [`27a6aaa` (#2707)](https://github.com/Lundalogik/lime-elements/pull/2707/commits/27a6aaac38dfee901c07ec5782ac75eee6356247)

src/components/chip/chip.tsx
Kiarokh: Exactly. In the feed, they are just fully functional and interactive chips. I'll resolve the similar comment further down, and you can resolve this when you have read it ;)

src/components/chip/chip.tsx
Kiarokh: Yes. I see. That's correct. But I just got a hallelujah idea. Can we define this on the `<Host>` instead?
```tsx
render() {
    return (
        <Host onClick={this.filterClick}>
            {/* rest of the component */}
        </Host>
    );
}
```

This way, we can remove the CSS rule 
```scss
:host(limel-chip[disabled]) { 
    pointer-events: none; 
}
```
, as the filterClick method will now handle preventing clicks when the component is disabled.

The reason I think this is better is that we have always used that CSS above, on components like `limel-button` or `limel-icon-button` or any similar component that the consumers will define an `onClick` on. This have always created a lot of problems, since when you set `pointer-events: none; ` you will lose every-wonderful-thing that you would get from pointers interacting with the component; e.g. `:focus` and or `:hover`, which could be so useful in communicating with users using micro interactions. Another thing you would lose is the possibility of attaching a `limle-tooltip` to the component. Because when it is `disabled`, the pointer-events would be none, which means you cannot hover the component to get a tooltip. The tooltip could be used smartly, to explain why the button is disabled for example, (or show the `title` or `label` of the icon-button, which is hard to know without hovering it).

If `<Host onClick={this.filterClick}>` will do the exact same thing (also handle the case both for link and button, then I can make follow up PRs in other components, and remove that super annoying `pointer-events: none; `.

src/components/chip/examples/chip-link.tsx
Kiarokh: Ok. This was also a copy/paste form other components too (which I think you and I did together actually 😅 ). But I think saying it is not bad, to remind the consumer what to think about, when designing interactions.

src/components/chip/examples/chip-aria-role.tsx
```diff
@@ -0,0 +1,60 @@
...
+export class ChipAriaRoleExample {
+    public render() {
+        return (
+            <div role="grid" aria-label="Filter bar" aria-colcount="4">
+                <div role="rowgroup">
+                    <div role="row">
+                        <limel-chip
+                            role="gridcell"
+                            aria-colindex="1"
+                            id="1"
```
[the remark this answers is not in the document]
Kiarokh: Good catch. I forgot to increment the `id`s and the `aria-colindex`! c4c5e6bc63863a60fef61b1ecf3ab74b847d49f6 

Why I chose this DOM structure, see the example in [role='gridcell'](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/gridcell_role).

Regarding `aria-posinset`
They [say](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-posinset):
> The `aria-posinset` attribute defines an element's number or position in the current set of `listitems` or `treeitems` when not all items are present in the DOM.
> If all the items in a list are present in the DOM, the browser can calculate the total and the position of each, making `aria-posinset` unnecessary.

src/components/chip/examples/chip-button.tsx
Kiarokh: But that's not what I want here. Toggling `selected` state is showcased in another example. 

Here, you should imagine a chip in a chip-set. You just wanna click it and something should happen. That's a hard fact and we can't change it. 

When I said I don't want people to use chips instead of buttons, I meant I don't want them for instance to use two chips like `❌ Cancel` and `💾 Save`, and put them in a modal. But I can't stop them from doing so neither. I think reading the docs should be enough. Maybe we can add a couple of DO and DON'T examples too.

In this example, maybe we can change the icon and label of the chip.

src/components/chip/examples/chip-filter.tsx
```diff
@@ -0,0 +1,61 @@
...
+ * Chips are great candidates to visualize active filters.
+ * However, as chips are used for other purposes as well,
+ * we need to make sure that the user understands that the chip is a filter,
+ * just by the look of it.
+ *
+ * By setting the `type` to `filter`, the chip will be rendered with a distinct style
+ * suitable for visualizing filters.
+ *
+ * :::note
+ * Normally, clicking on the chip should also toggle its `selected` state in this mode.
```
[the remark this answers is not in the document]
Kiarokh: agree

src/components/chip/examples/chip-link.tsx
Kiarokh: agree! ⚡

src/components/chip/examples/chip-link.tsx
Kiarokh: I removed it

src/components/chip/examples/chip-link.tsx
Kiarokh: ⚡

src/components/chip/chip.tsx
adrianschmidt: > Do we even need this? 

I'm not entirely sure what we use the id for. I guess we need to check the code.

> What's this `value?: T;` that the chip.types.ts has?

That's a "[generic](https://www.typescriptlang.org/docs/handbook/2/generics.html)". What that does is that it lets the _consumer_ specify what type the `value` property should have. So I could use it like this:

```ts
const myChip: Chip<string> = {
    id: 123,
    value: 'hello',
}
```

Or like this:

```ts
const myChip: Chip<LimeObject> = {
    id: 123,
    value: myCompany,
}
```

> Should it be that one instead?

Not sure what you're asking here. If `id` should have type `T`?

> Note that we both `id` there too!

Yes, but the `id` on `Chip` is never used an an attribute on an html tag. It's perfectly legal to have a property called `id` in javascript, the problem is that since Stencil connects all `@Prop` decorated properties to the correspondingly named html attribute, we get a conflict with the native `id` html attribute.

src/components/chip/chip.tsx
adrianschmidt: Well… it became a little less obviously self-contradictory, but I don't think it really became any clearer…

Perhaps we can have a huddle, and you can explain to me what you want to get across, and then I can perhaps help formulate that in some way? 😊

src/components/chip/chip.tsx
adrianschmidt: It seems like that ought to work, yes…

Sounds like a great idea! Please feel free to try it!

src/components/chip/chip.e2e.ts
```diff
@@ -19,10 +19,22 @@ describe('limel-chip', () => {
...
         );
+
+        const element = await page.find('limel-chip');
+        element.setProperty('link', { href: 'https://example.com' });
+
         await page.waitForChanges();
 
-        const element = await page.find('limel-chip >>> a.chip');
-        expect(element.getAttribute('href')).toEqual('https://example.com');
+        expect(element)
```
LawrenceBorst, line 28: The problem is that the shadowRoot is an empty dict. It seems that in other places we've opted to just use the HTML.

You might wonder why we need to invoke `setProperty`. I don't know. It just breaks if I don't do it here, even if we wait for changes. Manually everything works.

But then it also breaks if we don't put the link at the top even if we set th e property 🤷🏽‍♂️

src/components/chip/chip.tsx
Kiarokh: Ok I buy it. Added a new fixup c6b21dbf85a3a96d9635b3b2be198f37c44a86c1 but I can't really explain it more. These are like unwritten design rules. Maybe I'm one of the first people trying to write something meaningful about this topic 😅

Kiarokh: internally for our consultants and such, I have 0 hope that they even look at this docs. Sorry, I have -100 hope that they even care to read a second. Many seem to just go it and see if they like how it looks, then jut copy/paste the code into their project 😞

src/components/chip/chip.tsx
Kiarokh: yes, we need to control the readonly from outside, e.g. the consuming component like `limel-chip-set`. Also imagine a tag list of static hashtags, or static list of people joining an event. they could all be non-interactive set of chips

src/components/chip/examples/chip-button.tsx
adrianschmidt: OK. No worries, as long as you are happy with what's being conveyed, I'm perfectly OK with that 😊

src/components/chip/chip.e2e.ts
adrianschmidt: > The problem it seems is that the shadowRoot is an empty dict. It seems that in other places we've opted to just use the HTML where there was a shadow node.

Hmm… that's odd. In Jest (which .spec-files are run in) it makes sense, since Jest isn't running a real browser environment. But it's weird that there wouldn't be a real shadowRoot in e2e-tests, since those are run in a real browser by Puppeteer…

Although, the version of Puppeteer that's being used is some old version that still runs Chrome's old headless mode, which is apparently a completely separate browser implementation from the normal Chrome browser (which kind of defeats the whole purpose a bit, one would think 🤦‍♂️). That's really old, and is being replaced by a new headless mode that actually uses the regular Chrome browser. So, maybe that old headless mode just doesn't support shadow-DOM properly, and our tests are being so weird and hard to get to work because it's actually running some polyfill…? 🤔

> You might wonder why we need to invoke `setProperty`. I don't know. It just breaks if I don't do it here, even if we wait for changes. Manually everything works.

The reason is that the `Link` property is not a string. You can only pass strings as html attributes. Stencil magics it so that you seem to be able to pass anything when you do it in tsx, but here it's vanilla html, so it doesn't work.

The reason you have to use `.setProperty` rather than just do `element.link = { href: 'https://example.com/' }` is that the `element` isn't actually a DOM-node, it's a puppeteer object with references to the actual DOM-node.

> But then it also breaks if we don't put the link at the top even if we set the property 🤷🏽‍♂️

I don't understand what you mean by "put the link at the top"? 🤔

src/components/chip/chip.tsx
adrianschmidt: > Ok I buy it. Added a new fixup [c6b21db](https://github.com/Lundalogik/lime-elements/commit/c6b21dbf85a3a96d9635b3b2be198f37c44a86c1) but I can't really explain it better. These are like unwritten design rules. Maybe I'm one of the first people trying to write something meaningful about this topic 😅

I think the updated explanation is great! 🙌

> Many seem to just go it and see if they like how it looks, then jut copy/paste the code into their project 😞

Mmm… yeah, I know 😞

"simulating network delay"… 😣

src/components/chip/chip.spec.ts
adrianschmidt, line 1: @LawrenceBorst Yeah, it seems rather pointless to have the same things being tested both ways. Specs are faster to run, but they run in a simulated browser environment. Generally, stuff testing the html components is a better fit for e2e-tests, and the spec-tests is where more traditional unit-tests should go—things testing the logic of one specific function (that is not manipulating the page content directly). The kind of things that could be broken out into a separate file and imported from there. That's also how I would recommend making such functions testade, if they are private, because exposing the inner workings of a component just to be able to test it, isn't a wise choice, imo.

src/components/chip/chip.e2e.ts
LawrenceBorst: > ...maybe that old headless mode just doesn't support shadow-DOM properly...

Maybe the old one doesn't, that seems reasonable 🤔. Interesting stuff. I'll dig around for a bit, now I'm curious how exactly they implement their e2e test framework

The "link at the top" is the link attribute inside the chip element defined at the top when we initialise the page when we arrange this test. 😅 I should've probably used a more precise descriptor

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,289 @@
...
+    tag: 'limel-chip',
+    shadow: true,
+    styleUrl: 'chip.scss',
+})
+export class Chip {
+    /**
+     * Defines the language for translations.
+     * Will translate the translatable strings on the components.
+     */
+    @Prop()
```
adrianschmidt, line 74:
```suggestion
    @Prop({ reflect: true })
```

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,289 @@
...
+    public language: Languages = 'en';
+
+    /**
+     * ID of the chip. Must be unique.
+     */
+    @Prop()
```
adrianschmidt, line 80:
```suggestion
    @Prop({ reflect: true })
```

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,289 @@
...
+    id: number | string;
+
+    /**
+     * Label displayed on the chip
+     */
+    @Prop()
```
adrianschmidt, line 86:
```suggestion
    @Prop({ reflect: true })
```

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,289 @@
...
+    /**
+     * If supplied, the chip will become a clickable link.
+     */
+    @Prop()
+    public link?: Omit<Link, 'text'>;
+
+    /**
+     * The value of the badge, displayed on the chip.
+     */
+    @Prop()
```
adrianschmidt, line 104:
```suggestion
    @Prop({ reflect: true })
```

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,289 @@
...
+     * Icon of the chip.
+     */
+    @Prop()
+    public icon: string | Icon;
+
+    /**
+     * If supplied, the chip will become a clickable link.
+     */
+    @Prop()
+    public link?: Omit<Link, 'text'>;
```
adrianschmidt, line 99: No, I didn't forget these 😊 These should not be reflected, since they are not primitive values. (Well, `icon` _might_ be primitive, but also might not be…)

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,289 @@
...
+
+    private renderAsLink = () => {
+        return [
+            <a
+                class="chip"
+                href={this.link.href}
+                title={this.link.title}
+                target={this.link.target}
+                aria-disabled={this.disabled || this.readonly}
+                tabindex={this.disabled ? -1 : 0}
```
adrianschmidt, line 193: Should this also be `-1` for readonly?

```suggestion
                tabindex={(this.disabled || this.readonly) ? -1 : 0}
```

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,289 @@
...
+                aria-label={this.removeChipLabel}
+                innerHTML={svgData}
+                onClick={this.handleRemoveClick}
+            />
+        );
+    }
+
+    private filterClick = (e) => {
+        if (this.disabled) {
+            e.preventDefault();
```
adrianschmidt, line 259: Presumably we want to prevent default on readonly too?

src/components/chip/chip.e2e.ts
adrianschmidt: > > ...maybe that old headless mode just doesn't support shadow-DOM properly...
> 
> Maybe the old one doesn't, that seems reasonable 🤔. Interesting stuff. I'll dig around for a bit, now I'm curious how exactly they implement their e2e test framework

I think that when you run the e2e-tests, there is some info logged about the new mode, and that you can opt-in to it by using some config. Perhaps we can simply opt-in to the new mode? I'm not sure how to though, because I think Stencil "hides" a bunch of the Puppeteer config from us, so I'm not sure if we can set that property that needs setting… 🤷‍♂️

> The "link at the top" is the link attribute inside the chip element defined at the top when we initialise the page when we arrange this test. 😅 I should've probably used a more precise descriptor

Ah! Now I get what you meant. No worries, I'm just pretty rubbish at figuring out what people mean sometimes! 🙈😅

That gave me in inkling actually. The chip looks for the `this.link` property when it renders, and renders a button if the link property is falsy. It ought to swap that for an anchor element when the link property becomes truthy, but if we've already tried to find that element by then, or the page doesn't have time to update before we do, then we're not going to find it.

src/components/chip/chip.tsx
adrianschmidt: @jgroth @vicgeralds Can you chime in here? I'm thinking that, if you use limel-chip stand-alone (the way it's used in the examples), it's a minor inconvenience to have to create a `Chip` object to pass to `limel-chip`. But I feel like _most_ times someone would use a chip (and when we want to use it in a future version of `limel-chip-set`) you would have an array of `Chip`s. Again, having to apply those values as individual attributes is really only a minor inconvenience, but it's when it comes to event handlers that I'm getting a little worried that separate attributes might make things a major pain in the butt:

- If we pass a `Chip` as a single property, then we'd presumably pass that chip along with any event that happens. Any change to the chip's status could either be reflected in the `Chip` properties (we might have a property `Chip.selected`, for example), or by having separate events (`chipSelected` and `chipDeselected`, for example). You could then listen for any event on some container, rather than hooking up one listener for each event on each chip, because you would know _which_ chip the event applies to—the affected chip is attached to the event!
- If, on the other hand, we use separate attributes, our convention is to pass the value of the property that changed. So the event for a chip being selected would just be `selected`, with a value of `true` or `false`. We could of course choose to depart from that convention, but that would be unexpected and presumably confusing (including to our future selves, I believe). So, assuming we keep to our convention, we need to know from _which chip_ an event came. Which means either hooking up separate listeners for each chip, which isn't nice performance-wise, or inspecting the event to find out which DOM element was clicked, and from there figure out which chip that correlates to. "…and there shall be much wailing and gnashing of teeth…" 🙈

src/components/chip/chip.spec.ts
Kiarokh: I also like the e2e tests much more. I feel for UI components like Lime Elements, the spec tests are generally stupid. Or at least I don't see a point in them really. They always break (100% guarantied) as soon as you update something in the component. They are perhaps good if you write a test that says "make sure this is not a `span` or a `p` but an `h1` element (due to accessibility or whatever reasons). But otherwise, I'm not really sure how these tests help to be honest.

src/components/chip/chip.tsx
jgroth: I think having a single `chip` property is a bit annoying and not how we usually have done things. The chip component should probably implement the `Chip` interface to ensure the component and the interface are in sync

src/components/chip/chip.spec.ts
LawrenceBorst: > @LawrenceBorst Yeah, it seems rather pointless to have the same things being tested both ways. Specs are faster to run, but they run in a simulated browser environment. Generally, stuff testing the html components is a better fit for e2e-tests, and the spec-tests is where more traditional unit-tests should go—things testing the logic of one specific function (that is not manipulating the page content directly). The kind of things that could be broken out into a separate file and imported from there. That's also how I would recommend making such functions testade, if they are private, because exposing the inner workings of a component just to be able to test it, isn't a wise choice, imo.

This 100%. This is the standard

> I also like the e2e tests much more. I feel for UI components like Lime Elements, the spec tests are generally stupid. Or at least I don't see a point in them really. They always break (100% guarantied) as soon as you update something in the component. They are perhaps good if you write a test that says "make sure this is not a span or a p but an h1 element (due to accessibility or whatever reasons). But otherwise, I'm not really sure how these tests help to be honest.

Yeah these kinds of tests actually add net negative value to the company. Tests are code too and if the cost of writing and maintaining a test is higher than the gain in clarity, free documentation and assurance moving forward then it's not worth it. Of course, that's almost never the case--in fact the equation is probably even longer on the right hand side--except it is for dumb tests like this 😅

On that note I actually wish we did test more actually, but by that I still mean unit testing only the functional requirements of the system under test and not some other tangential system, and for e2e testing only the relevant user experience.

src/components/chip/chip.tsx
Kiarokh: why is reflecting useful here?

src/components/chip/chip.tsx
Kiarokh: OK I don't really know what the "primitive values" are. But in my world, if the reflected value can be used for styling, I reflect it. Otherwise, I keep it quite 😅 is that not enough? 

What I've learned as a general rule of thumb, you should only use `@Prop({ reflect: true })` for props that are used to control the visual appearance of the component. For all other props, you should use `@Prop()`. This will make your code more maintainable and easier to understand.

I'm thinking, knowing the `text` or the `badge` or `icon` or even the `id` has no value in styling the component. So why reflecting them?

src/components/chip/chip.tsx
Kiarokh: https://github.com/Lundalogik/lime-elements/pull/2707/files#r1457404844

src/components/chip/chip.tsx
Kiarokh: Yeay 152cc5630a546154b121e0f3f82b0c0a662a135f it works like a charm! 
Look @LucyChyzhova ! Finally, we found a solution [this super annoying issue](https://github.com/Lundalogik/crm-feature/issues/3317).

😊 You remember this: 

https://github.com/Lundalogik/lime-elements/blob/e1c555dabc0084927d39ec6755ccde86b3360512/src/design-guidelines/disabled-hidden/examples/disabled-vs-hidden.tsx#L83C47-L83C47

src/components/chip/chip.tsx
vicgeralds: > So, if we don't switch to accepting an object of type Chip, we should rename this property.

Can't we just remove it?

src/components/chip/chip.tsx
vicgeralds: Having an object as a single property makes it more difficult/annoying to detect changes as well. If the passed chip object isn't strictly equal to the old chip object, that will cause re-rendering even if there's no real changes.

> we need to know from _which chip_ an event came. Which means either hooking up separate listeners for each chip, which isn't nice performance-wise, or inspecting the event to find out which DOM element was clicked, and from there figure out which chip that correlates to. 

We don't make it any easier to identify which input field a change event originated from. Why should we be concerned then about chips?

I don't think the individual component should need to worry about identifying itself among a collection of others.

src/components/chip/chip.spec.ts
adrianschmidt: I completely agree @LawrenceBorst!

> I feel for UI components like Lime Elements, the spec tests are generally stupid. Or at least I don't see a point in them really. They always break (100% guarantied) as soon as you update something in the component. They are perhaps good if you write a test that says "make sure this is not a `span` or a `p` but an `h1` element (due to accessibility or whatever reasons). But otherwise, I'm not really sure how these tests help to be honest.

Yeah, the thing is, spec-tests shouldn't be used to test the html output at all, imo. Specs should be used when some logical function can be broken out and tested completely separately. To take a super simple example, let's say that we broke out the logic for "disabled or readonly" into a separate function:

So we might want to change the following:

```tsx
public render() {
    return (
        <a
            class="chip"
            href={this.link.href}
            aria-disabled={this.disabled || this.readonly}
            tabindex={(this.disabled || this.readonly) ? -1 : 0}
        >
    );
}
```

…to:
```tsx
public render() {
    return (
        <a
            class="chip"
            href={this.link.href}
            aria-disabled={this.isDisabled()}
            tabindex={this.isDisabled() ? -1 : 0}
        >
    );
}

private isDisabled = (isReadonly: boolean, isDisabled: boolean) => {
    return isReadonly || isDisabled;
};
```

But then `isDisabled` is private, and making it public just to test it is like cutting your foot off to get a closer look at the splinter in your toe. Instead, we move the function to a separate file, from which we can import it to both the main component, and to a test file:

```ts
export const isDisabled = (isReadonly: boolean, isDisabled: boolean) => {
    return isReadonly || isDisabled;
};
```

```tsx
import { isDisabled } from `./logicUnits.ts`;

public render() {
    return (
        <a
            class="chip"
            href={this.link.href}
            aria-disabled={this.isDisabled()}
            tabindex={this.isDisabled() ? -1 : 0}
        >
    );
}
```

```ts
import { isDisabled } from `./logicUnits.ts`;

describe('isDisabled', () => {
    it('blah blah', () => {
        …
```

src/components/chip/chip.tsx
adrianschmidt: > I don't really know what the "primitive values" are.

Basically it's strings, numbers, and booleans. Technically `null` and `undefined` are also primitive values, and I assume `NaN` is too, since it's considered a `number` (yes! "not a number" has the type `number` in javascript 😂)

(Technically, anything that isn't a function or an object is a primitive, and since functions are also objects, anything that isn't an object is a primitive 😝)

src/components/chip/chip.tsx
adrianschmidt: > I think having a single `chip` property is a bit annoying and not how we usually have done things.

> We don't make it any easier to identify which input field a change event originated from. Why should we be concerned then about chips?

Alright, then that's settled! Good thing I asked! 😄

Thank you! 🙏

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,289 @@
...
+ * @exampleComponent limel-example-chip-filter
+ * @exampleComponent limel-example-chip-removable
+ * @exampleComponent limel-example-chip-aria-role
+ */
+@Component({
+    tag: 'limel-chip',
+    shadow: true,
+    styleUrl: 'chip.scss',
+})
+export class Chip {
```
adrianschmidt, line 70: https://github.com/Lundalogik/lime-elements/pull/2707#discussion_r1457371745

This component should probably implement  the `Chip` interface (the existing one or a new one, but probably the existing one), to make sure the component and the interface are in sync.

src/components/chip/chip.tsx
adrianschmidt: There are a few places where we have a lone `this.disabled`, and they should all probably be paired up with a `this.readonly` then.

src/components/chip/chip.tsx
adrianschmidt: Yeah, since we're not using it inside the component, let's just remove it 👍

If we need it for the chip-set or something, we can deal with that then, probably without having to change this component.

Kiarokh: ⚡

src/components/chip/chip.tsx
Kiarokh: yes.

src/components/chip/chip.e2e.ts
```diff
@@ -0,0 +1,57 @@
...
+          <a class="chip" href="https://example.com" tabindex="0">
+            <span class="text">
+              Test chip
+            </span>
+          </a>
+        </mock:shadow-root>
+      </limel-chip>`);
+    });
+
+    it('should render with a badge when the badge prop is provided', async () => {
```
adrianschmidt, line 40: Regarding `it('should render …`, we typically just write `it('renders …`.

Just an FYI, I've already fixed it locally and will be pushing a fixup.

src/components/chip/chip.tsx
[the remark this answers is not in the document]
adrianschmidt: ⚡

src/components/chip/chip.tsx
adrianschmidt: ⚡

Since the `id` property is required in the `Chip` interface, but we removed it from here, I made a special solution that I think is good enough for now. The component is still private after all.

image

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,278 @@
...
+     * Icon of the chip.
+     */
+    @Prop()
+    public icon: string | Icon;
+
+    /**
+     * If supplied, the chip will become a clickable link.
+     */
+    @Prop()
+    public link?: Omit<Link, 'text'>;
```
adrianschmidt, line 97: Hmm… I just realised we probably want to adjust this typing just a tad. Right now, I'm pretty sure you can't supply a `Link`, and I think it would be quite nice if you could (just that if you do, we'll ignore the `text` property on it).

I really want to get this merged though, so might look at tweaking that in a separate PR.

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,278 @@
...
+    @Prop({ reflect: true })
+    public readonly = false;
+
+    /**
+     * Set to `true` to visualize the chip in a "selected" state.
+     * This is typically used when the chip is used in a chip-set
+     * along with other chips.
+     */
+    @Prop({ reflect: true })
+    public selected = false;
```
adrianschmidt, line 124: There are some very minor interaction-visuals happening when you click the chip if it's both `selected` and either `disabled` or `readonly`:

https://github.com/Lundalogik/lime-elements/assets/154725/df3405f7-9ddf-4b07-ab4c-866ecee35d3c

It's so minor I think we should make any fixes in a separate PR, so we don't hold up this big one even longer.

src/components/chip/chip.tsx
```diff
@@ -0,0 +1,278 @@
...
+     * suitable for visualizing filters.
+     */
+    @Prop({ reflect: true })
+    public type: 'filter';
+
+    /**
+     * Fired when clicking on the remove button of a `removable` chip.
+     */
+    @Event()
+    public remove: EventEmitter<Chip>;
```
adrianschmidt, line 143: In the docs, this makes it look like we emit an object of the type defined in `chip-set/chip.types.ts`, because that's the only `Chip` type we export.

image

We don't need to attach anything to the event, so I'll just remove it.

adrianschmidt: ⚡

adrianschmidt (approved):

src/components/chip/chip.tsx
adrianschmidt: Ah! OK, `text` is already optional on `Link`, so it's already possible to create a `Link` without a `text` property, and then pass it to `limel-chip`.

I wonder if it's worth omitting the `text` property then… 🤔

I guess it makes it _obvious_ that we won't use the `text` value, if nothing else 😄

adrianschmidt (approved):

PR #2720, opened 2024-01-16
docs(markdown): add composite example

TommyLindh2:
This is an example I have locally to test out how a specific markdown/HTML will be rendered in Lime CRM. We have custom styling which make it render in another way than a regular markdown.

So this example makes very much sense to have so that we can test content out directly in the docs.

https://github.com/Lundalogik/lime-elements/assets/1814702/9030f83f-f074-4efc-bcd9-395dd4a72762

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/markdown/examples/markdown-composite.scss
+++ b/src/components/markdown/examples/markdown-composite.scss
@@ -1,0 +1,21 @@
+:host(limel-example-markdown-composite) {
+    display: flex;
+    flex-direction: column;
+    gap: 1rem;
+
+    limel-input-field {
+        height: 10rem;
+    }
+
+    fieldset {
+        padding-block: 1rem;
+        padding-inline: 1rem;
+        margin-inline: 0;
+        border: 0.125rem dashed rgb(var(--contrast-700));
+
+        legend {
+            color: rgb(var(--contrast-1000));
+            text-align: center;
+        }
+    }
+}
--- a/src/components/markdown/examples/markdown-composite.tsx
+++ b/src/components/markdown/examples/markdown-composite.tsx
@@ -1,0 +1,37 @@
+import { LimelInputFieldCustomEvent } from '@limetech/lime-elements';
+import { Component, State, h } from '@stencil/core';
+
+/**
+ * Composite example
+ * Test your markdown code and see what you get in return in real-time.
+ */
+@Component({
+    tag: 'limel-example-markdown-composite',
+    styleUrl: 'markdown-composite.scss',
+    shadow: true,
+})
+export class MarkdownRenderContentExample {
+    @State()
+    private markdown = '# Hello, world!\n\nThis is **markdown**!';
+
+    public render() {
+        return [
+            <limel-input-field
+                label="Markdown to render"
+                type="textarea"
+                value={this.markdown}
+                onChange={this.handleMarkdownChange}
+            />,
+            <fieldset>
+                <legend>Rendered markdown</legend>
+                <limel-markdown value={this.markdown} />
+            </fieldset>,
+        ];
+    }
+
+    private handleMarkdownChange = (
+        event: LimelInputFieldCustomEvent<string>,
+    ) => {
+        this.markdown = event.detail;
+    };
+}
--- a/src/components/markdown/markdown.tsx
+++ b/src/components/markdown/markdown.tsx
@@ -16,6 +16,7 @@ import { markdownToHTML } from './markdown-parser';
  * @exampleComponent limel-example-markdown-html
  * @exampleComponent limel-example-markdown-blockquotes
  * @exampleComponent limel-example-markdown-horizontal-rule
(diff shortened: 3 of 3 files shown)


Kiarokh:
> We have custom styling which make it render in another way than a regular markdown.

Do you mean the bold text in the header which is not being rendered as bold?

If you list some of the issues, I can fix them. Now that we have changed the font-family, I have more tools to play with, when it comes to typography

src/components/markdown/examples/markdown-composite.tsx
```diff
@@ -0,0 +1,38 @@
...
+                    value={this.markdown}
+                    onChange={this.handleMarkdownChange}
+                />
+                <fieldset>
+                    <legend>Rendered markdown</legend>
+                    <limel-markdown value={this.markdown} />
+                </fieldset>
+            </Host>
+        );
+    }
```
Kiarokh, line 31: I normally don't add `<Host>` like this, unless I need a tag or attribute on the host itself. I'd recommend just using an array instead.

```suggestion
import { LimelInputFieldCustomEvent } from '@limetech/lime-elements';
import { Component, State, h } from '@stencil/core';

/**
 * Composite example 
 * Test your markdown code and see what you get in return in real-time.
 */
@Component({
    tag: 'limel-example-markdown-composite',
    styleUrl: 'markdown-composite.scss',
    shadow: true,
})
export class MarkdownRenderContentExample {
    @State()
    private markdown = '# Hello, world!\n\nThis is **markdown**!';

    public render() {
        return [
              <limel-input-field
                  label="Markdown to render"
                  type="textarea"
                  value={this.markdown}
                  onChange={this.handleMarkdownChange}
              />
              <fieldset>
                  <legend>Rendered markdown</legend>
                  <limel-markdown value={this.markdown} />
              </fieldset>
          ];
    }
```

src/components/markdown/examples/markdown-composite.tsx
```diff
@@ -0,0 +1,38 @@
+import { LimelInputFieldCustomEvent } from '@limetech/lime-elements';
+import { Component, Host, State, h } from '@stencil/core';
+
+/**
+ * Composite example with a markdown input field and a markdown renderer.
+ */
```
Kiarokh, line 6:
```suggestion
/**
 * Composite example 
 * Test your markdown code and see what you get in return in real-time.
 */
```

src/components/markdown/examples/markdown-composite.scss
```diff
@@ -0,0 +1,19 @@
...
+    flex-direction: column;
+    gap: 1rem;
+
+    limel-input-field {
+        height: 10rem;
+    }
+
+    fieldset {
+        padding: 0.5rem 0.5rem 3rem 0.5rem;
+        border-color: rgba(var(--contrast-1000), 0.3);
```
Kiarokh, line 12:
```suggestion
        border: 0.125rem dashed rgb(var(--contrast-700));
```

src/components/markdown/examples/markdown-composite.scss
```diff
@@ -0,0 +1,19 @@
...
+    display: flex;
+    flex-direction: column;
+    gap: 1rem;
+
+    limel-input-field {
+        height: 10rem;
+    }
+
+    fieldset {
+        padding: 0.5rem 0.5rem 3rem 0.5rem;
```
Kiarokh, line 11:
```suggestion
            padding-block: 1rem;
            padding-inline: 1rem;
            margin-inline: 0;
```

Since we are overriding the native styles and they are using the logical properties, it'd be nice to also override them with logical properties. :)

src/components/markdown/examples/markdown-composite.scss
```diff
@@ -0,0 +1,19 @@
...
+        border-color: rgba(var(--contrast-1000), 0.3);
+        border-radius: 0.25rem;
+
+        legend {
+            color: rgb(var(--contrast-1000));
```
Kiarokh, line 16:
```suggestion
            color: rgb(var(--contrast-1000));
            text-align: center;
```

TommyLindh2:
> > We have custom styling which make it render in another way than a regular markdown.
> 
> Do you mean the bold text in the header which is not being rendered as bold?
> 
> If you list some of the issues, I can fix them. Now that we have changed the font-family, I have more tools to play with, when it comes to typography

It mostly related to the communication flow we've noticed it, when we have e-mail content in HTML, where tables looks terrible when nested too much. I have an example I can show you from a customer case where I've masked sensitive data.

**How it looks in HTML in the browser**

image


**How it looks in our Markdown**

image

Kiarokh (approved):

PR #2746, opened 2024-01-30
Avoid future cases of broken type exports

adrianschmidt:
Fixes problem mentioned in [this Slack message](https://lime-technologies.slack.com/archives/C1K6TN9BK/p1705587480128879).

> Ok, I found out that some types exported from Lime Elements are broken again :disappointed: I fixed them a while back, but now some of them were just exported as any again, leading to a lot of confusion and loss of type hints. It happens if anything is imported from @limetech/lime-elements from within lime elements itself, so please keep this in mind and do not do it! :smile:  (it's ok to import from @limetech/lime-elements in the examples, and this is actually how we should do it in there). Here's a PR to fix it, we maybe could investigate if it's possible to check for this kind of error automatically since it's hard to discover manually and I just did it by accident now after a lot of debugging :smile:
https://github.com/Lundalogik/lime-elements/pull/2726

The first commit fixes the problem mentioned above. The second commit bans importing from `lodash` and `underscore`, with an instruction to import from `lodash-es` instead.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -121,6 +121,14 @@ module.exports = {
                 markers: ['/'],
             },
         ],
+        'no-restricted-imports': [
+            'error',
+            {
+                name: '@limetech/lime-elements',
+                message:
+                    'Production code should not import from `@limetech/lime-elements`. Please import from a relative path instead.',
+            },
+        ],
     },
     overrides: [
         {
@@ -185,6 +193,7 @@ module.exports = {
                     { name: ['test', 'only'], message: "don't focus tests" },
                     { name: 'ftest', message: "don't focus tests" },
                 ],
+                'no-restricted-imports': 'off',
             },
         },
     ],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -128,6 +128,16 @@ module.exports = {
                 message:
                     'Production code should not import from `@limetech/lime-elements`. Please import from a relative path instead.',
             },
+            {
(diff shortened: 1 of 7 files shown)


.eslintrc.js
adrianschmidt, line 1: The build is failing here: https://github.com/Lundalogik/lime-elements/actions/runs/7710765216/job/21014792252?pr=2746#step:9:509

I've tracked it down to be something that happens in Kompendium, most seemingly inside `typedoc`, because the error happens inside this call to `app.convert()`: https://github.com/jgroth/kompendium/blob/75489490bbceb3f39fee8025570e0f8fdcb28132/src/kompendium/typedoc.ts#L58C28-L58C35

I haven't been able to track it down any further. I'm sure the problem isn't actually in typedoc, and probably not in Kompendium either, but with no information about _what_ is actually `undefined` here, I have no idea what could be the actual cause of the problem 🤷‍♂️ 😞

adrianschmidt: Found it! Fixed. Now I just need to fix the commitlint problem too… 🙄

PR #2747, opened 2024-01-30
Readonly boolean

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/3312

instead of https://github.com/Lundalogik/lime-crm-building-blocks/pull/319

to test in Lime Admin, add this example config on a boolean field

```json
{
	"readonlyLabels": [
		{
			"value": true,
			"icon": "news",
			"text": "Is subscribed to receive newsletters"
		},
		{
			"value": false,
			"icon": {
				"name": "cancel_subscription",
				"color": "rgb(var(--color-orange-default))"
			},
			"text": "Is unsubscribed from newsletters"
		}
	]
}
``` 

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/checkbox/checkbox.scss
```diff
@@ -17,6 +17,7 @@
 }
 
 :host(limel-checkbox) {
+    min-height: 2.5rem; // ensures that the UI doesn't flicker too much when switching to `readonly`
```
Kiarokh, line 20: seems to cause this bug
image

src/components/checkbox/checkbox.tsx
```diff
@@ -94,6 +96,40 @@ export class Checkbox {
...
+
+    /**
+     * The icon to show, when the component is `readonly` and its `value` is `false`.
+     * This can be used to clarify what kind of data is being visualized.
+     * If not set, a default icon with a default color will be used.
+     * Colors can be customized using the `Icon` interface.
+     */
+    @Prop({ reflect: true })
+    public readonlyFalseIcon?: string | Icon;
+
```
Kiarokh, line 132: A config like this:
```json
{
    "readonlyTrueLabel": "Still works here",
    "readonlyFalseLabel": "Has quit their job",
    "readonlyTrueIcon": {
        "name": "inactive_state",
        "color": "rgb(var(--color-gray-default))"
    },
    "readonlyFalseIcon": {
        "name": "in_progress",
        "color": "rgb(var(--color-sky-default))"
    }
}
```
Doesn't do anything in Lime Admin at all.

I think these new props should somehow be added to the `extraProps` of the checkbox widget 👇 or somewhere in this file

https://github.com/Lundalogik/lime-elements/blob/0f56884b79f3149732d2710cd5ee327cb5a9b9b1/src/components/form/widgets/checkbox.ts#L20

### Other relevant and similar issues:
1. It's also worth nothing that adding `help` which we did here: https://github.com/Lundalogik/lime-elements/pull/2622, is not possible to do via Lime Admin on any component, but writing it in the schema in the code works:
     

2. Adding `helperText` on the boolean field is not possible via Lime Admin.
3. Adding styles (such as CSS variables of components) is not possible on any field via Lime Admin.

I think same things are wrong with out implementations about these.

Maybe you guys can help @adrianschmidt @vicgeralds @jgroth

jgroth: There's an issue about the CSS variables here Lundalogik/crm-feature#3772
I thought it was possible to specify props for the components in lime-admin, but maybe we need to fix some bug there as well :thinking:

src/components/checkbox/checkbox.tsx
```diff
@@ -94,6 +96,40 @@ export class Checkbox {
...
+    public readonlyTrueIcon?: string | Icon;
+
+    /**
+     * The icon to show, when the component is `readonly` and its `value` is `false`.
+     * This can be used to clarify what kind of data is being visualized.
+     * If not set, a default icon with a default color will be used.
+     * Colors can be customized using the `Icon` interface.
+     */
+    @Prop({ reflect: true })
+    public readonlyFalseIcon?: string | Icon;
```
jgroth, line 131: I'm not super happy with the naming of these and the fact that there are four of them. Having four properties controlling just this single aspect of the component feels a bit weird.

I think we should combine them to a single property and give it a good name/interface. I can't think of anything right now, but I'm thinking something along the lines of this

```ts
type ReadOnly {
    label: string;
    icon: string | Icon;
}

type ReadOnlyProps = Record<boolean, ReadOnly>
```

src/components/checkbox/checkbox.tsx
vicgeralds: It is possible to specify props for the component, and it also works (at least for text fields) without specifying a component name.

The problem with `help` is that it's not a property to pass. It's only a feature of limel-form when (ab)using schemas. :wink:

> Adding helperText on the boolean field is not possible via Lime Admin.

This should work, but I can see that it doesn't when trying it out. Not sure yet why 😕

Kiarokh: > The problem with help is that it's not a property to pass. It's only a feature of limel-form when (ab)using schemas. 😉

😮 Then we completely missed the target @civing @john-traas . One of the most important reasons for adding this was to be able to add contextual help to form elements, both on cards (by consultants) and in admin (link to platform docs by us).

vicgeralds: The reason specifying props for a boolean field didn't work, is that we don't pass a default custom component name, which we do apparently for text fields, for example. So it works if you specify limel-checkbox as component name.

jgroth: > The reason specifying props for a boolean field didn't work, is that we don't pass a default custom component name, which we do apparently for text fields, for example. So it works if you specify limel-checkbox as component name.

The problem with that is that `limel-checkbox` does not have a `value` prop (it has `checked` instead :disappointed: )

src/components/checkbox/checkbox.tsx
Kiarokh: I agree. Let me see what I can do

src/components/checkbox/checkbox.tsx
adrianschmidt: The reason limel-checkbox has `checked` instead of `value` is that that's what a native html checkbox has, and my idea originally was to keep Lime Elements' components as close to drop-in replacements as possible.

It feels like that was less possible than I originally intended, and I think we have already more or less decided to move away from that idea.

Adding a `value` property that takes precedence over `checked` ought to be simple, and I would suggest that we deprecate `checked` at the same time, if you all agree.

Ping @Lundalogik/lime-elements-maintainers.

jgroth:
This looks like a false positive :thinking: @adrianschmidt 
image

src/components/checkbox/checkbox.tsx
Kiarokh: I did what I think you meant @jgroth can you check?

src/components/checkbox/checkbox.tsx
civing: > 😮 Then we completely missed the target @civing @john-traas . One of the most important reasons for adding this was to be able to add contextual help to form elements, both on cards (by consultants) and in admin (link to platform docs by us).

Right now this is a feature of lime-forms where you specify the inputs for the help component in the schema, yes.

If we would like to make this generic for all components, we need to either support it as props for every component or do some other clever shenanigans. I really dislike the idea of violating DRY by extending every component with the same props.

Anyhow, we have the helper component so we can in theory add it somehow to components outside of lime-form.

adrianschmidt:
> This looks like a false positive 🤔 @adrianschmidt image

Fixed in #2765

src/components/switch/switch.tsx
```diff
@@ -124,58 +132,65 @@ export class Switch {
...
-                        'mdc-switch--selected': this.value,
-                    }}
-                    type="button"
-                    role="switch"
-                    aria-checked={this.value}
-                    disabled={this.disabled || this.readonly}
-                    onClick={this.handleClick}
+        if (this.readonly) {
+            return [
+                <limel-readonly-boolean
```
Kiarokh, line 137: by the way @jgroth, I don't fully buy the argument that the consumer who has control over the code can also create a separate design for the readonly mode. Of course your argument is 100% valid, but with the same argument we could say that the consumer can just use a native HTML checkbox, or input field, popover, or many of the things that this component library offers. 

One of the strong points of having a UI component library is that it helps keeping the UI consistent and clean, and making it easy for a developer to make a decent UI, without taking too many layout and ui decisions about minute graphical details. 

If everyone was forced to create their own stuff for everything, then the products using this library would become messy and inconsistent. 

Also, if **A:** the `readonly` is a feature of our boolean field, and **B:** the boolean fields create this classic usability issue and lead to confusions, then it make a lot of sense that the component itself offers a solution that prevent the usability issues. We can't expect the consumer to come up with solutions. If they could, we didn't have to make this PR neither.

jgroth: > by the way @jgroth, I don't fully buy the argument that the consumer who has control over the code can also create a separate design for the readonly mode. 

They don't have to create a separate design for it. There's already a new component added in this PR that could be used (if it's made public). That also opens the possibility to use it in more scenarios than for only boolean fields. Perhaps you want this design as well if you have a readonly option field?

> One of the strong points of having a UI component library is that it helps keeping the UI consistent and clean, and making it easy for a developer to make a decent UI, without taking too many layout and ui decisions about minute graphical details.

Yes, I agree, but that does not mean that every component needs to be handle every kind of scenario. Components should ideally be small and have a single purpose, and if you need additional functionality you should create compositions of different components. This is the same problem as with the `help` property that you described above. If we need to add these kinds of properties to all the components we are bloating them and not keeping them simple.

> If everyone was forced to create their own stuff for everything, then the products using this library would become messy and inconsistent.

I'm not sure what you mean by this, but if the new component is public no one has to create their own component. We can still add functionality to lime-admin to let this be configurable without consumers creating their own components.

Kiarokh: > They don't have to create a separate design for it. There's already a new component added in this PR that could be used (if it's made public). That also opens the possibility to use it in more scenarios than for only boolean fields. Perhaps you want this design as well if you have a readonly option field?

I have no objection if this component is public. But I don't think its usage applies to other components in Lime Elements, than our own boolean fields. The issue it is addressing is about the boolean field.

>  that does not mean that every component needs to be handle every kind of scenario. Components should ideally be small and have a single purpose, and if you need additional functionality you should create compositions of different components. 

I _fully_ agree with your sentiments, but I don't think this is the applicable place for it. In our case we offer a component that has a feature which in turn comes an _inherent_ problem. We cannot be saying to the consumer: 

"_Hey, that we are aware of this problem, and just wanted to let you know that this problem exists. You should make sure no to have this problem when you use our component. However, if you wanna fix it, you should go use another component._"

This would be both counter intuitive and wrong. It is like buying a smartphone which has a certain feature; let say a **Turbo Mode**. But when the user turns on the Turbo Mode, the phone overheats and turns off. The manufacturer says, we know that this problem exist. But we decided not to add a cooling mechanism inbuilt for it, to keep the device more simple. If you want to use the Turbo Mode, go buy a cooler shell for your smartphone. By the way, we also produce and sell cooler shells.

We know the the readonly mode of our boolean components comes with an inherent problem. So the component itself should have means of mitigating the problem. Therefore
>  that does not mean that every component needs to be handle every kind of scenario.

is not a valid argument here. Switch and Checkbox are already handling the scenario of being readonly by changing their visual design and interactivity. But they are handling it in an insufficient way, forcing the users to have problematic experiences, or forcing the consumer to do loads of extra work to fix it.

> We can still add functionality to lime-admin to let this be configurable without consumers creating their own components.
> create a component in lime-crm-building-blocks that combines the two

This is not only a Lime Admin problem or a CRM problem. This comes with boolean fields, specially when they're set to readonly. I don't know how you want to fix it or implement this in Lime Admin. I can't do it myself anyways. You guys should find a way for it. But as I tried to explain above, I believe we need to address the issue in the component, not outside of it.

vicgeralds: Ok. Lots of discussion here. Sorry if I missed something by not reading all the comments. But I have a suggestion.

Is it enough to handle readonly for checkboxes in limel-form? That is, if the schema type is boolean:
- render limel-checkbox if not readonly schema
- render limel-readonly-bool if readonly in schema

Currently in Lime CRM, we're not really specifying a custom component to use for boolean fields anyway.

I guess we don't really need a component that dynamically can toggle its readonly state?

Kiarokh: @vicgeralds that would certainly help Lime CRM, but it wouldn't fix the inherent problem of boolean fields, when they become readonly. That'd be we just add documentations to Switch and Checkbox, and ask consumers to not just mindlessly render the boolean field in readonly mode, and instead use `limel-readonly-boolean` (which means making it public, what Johan is saying) or another proper visualization.

I don't like it too much, but I also don't see this as a show stopper. We can do so too.

Anyhow, I don't know how to fix the issue in the limel-form anyways as you describe. You guys need to do it or help me do it.

Kiarokh:
1. can we squash?
2. how do we add this feature to the limel-form, and make it configurable via admin? Is that a next PR or should we give it a try here?

jgroth:
> 1. can we squash?
> 2. how do we add this feature to the limel-form, and make it configurable via admin? Is that a next PR or should we give it a try here?

1. Yes squash it :smile: 
2. Should it not work already to configure it? :thinking:

api-extractor.json
```diff
@@ -400,7 +400,7 @@
         "addToApiReportFile": false
       },
       "ae-incompatible-release-tags": {
-        "logLevel": "error",
+        "logLevel": "warning",
```
jgroth, line 403: @adrianschmidt I think we need to change this to `warning`. Stencil does not seem to output all the doc blocks, which makes API extractor complain about the release tags not being compatible

adrianschmidt: The way to "ignore" it, is to set `addToApiReportFile` to `true`. Then it adds the warning to the api report instead of outputting it on the command line, and that also makes the verify-check pass (at least if it's set to "warning").

That's what I've done with `ae-missing-release-tag` below (line 410-421).

Then we still get the warning in the api report, and can at least review it as part of the PR.

etc/lime-elements.api.md
adrianschmidt, line 1: The reason it looks like every singe line of this file has been changed is likely that someone is using the wrong node version.

jgroth: What Node version should be used? I've generated the file on Node 18

adrianschmidt: `.nvmrc` says `lts/*`, which is currently Node 20.

Aside: We should probably specify v20 instead in that file… 🤔

adrianschmidt: I fixed it and pushed the changes.

The thing is that once the file has been changed (in whatever way it is being changed to make it look like every single line has been changed), you have to remove the file and generate it anew for it to go back to what it was like. I don't know why. Perhaps we should update the `api:update` script to do that every time… 🤷‍♂️

src/design-guidelines/boolean/examples/radio-buttons.tsx
```diff
@@ -23,19 +23,6 @@ export class BooleanRadioButtonsExample {
         { text: 'Only myself', value: 2, selected: false },
     ];
 
-    @State()
-    private selectedItem1: ListItem | ListSeparator;
-    private selectedItem2: ListItem | ListSeparator;
```
jgroth: @Kiarokh I removed these since they were not used and caused the build to fail when building the docs
Please check if you forgot something or fix the example :smile:

Kiarokh: https://github.com/Lundalogik/lime-elements/assets/35954987/90637e6c-eedd-4d3d-af6f-6689231cc4e3

Without those, this will happen. 😓

jgroth: Well it was broken before as well :smile:

adrianschmidt: > Without those, this will happen. 😓 

Hmm… I assume that means that adding `selectedItem1` and `selectedItem2` solved that problem with the radio-buttons? 😲 🤨

Kiarokh:
I tried this 

```json
{
	"readonlyLabels": [
		{
			"value": true,
			"icon": "news",
			"text": "Is subscribed to receive newsletters"
		},
		{
			"value": false,
			"icon": {
				"name": "cancel_subscription",
				"color": "rgb(var(--color-orange-default))"
			},
			"text": "Is unsubscribed from newsletters"
		}
	]
}
```
in Lime Admin, and that shows only the default:
image

image


Am I doing something wrong?

jgroth:
> Am I doing something wrong?

It should work now :slightly_smiling_face:

Kiarokh:
> ```json
> {
> 	"readonlyLabels": [
> 		{
> 			"value": true,
> 			"icon": "news",
> 			"text": "Is subscribed to receive newsletters"
> 		},
> 		{
> 			"value": false,
> 			"icon": {
> 				"name": "cancel_subscription",
> 				"color": "rgb(var(--color-orange-default))"
> 			},
> 			"text": "Is unsubscribed from newsletters"
> 		}
> 	]
> }
> ```

Great! I've also added a final touch, and verified that this works. I'm happy. If you're happy too, please let's merge this so that I can write the guidelines for the platform.

jgroth (approved):

Issue #2749, opened 2024-01-31
limel-picker: shows empty icon in the chipset


LucyChyzhova:
## Current behavior

https://github.com/Lundalogik/lime-elements/assets/50618208/d044e525-045c-4cfd-a4a3-170be5857188

1. …

## Environment

- **lime-elements** version: 
- Framework used: 
- Logs:

PR #2750, opened 2024-01-31
fix(limel-picker): don't show an empty icon if no icon is provided

LucyChyzhova:
Fix: https://github.com/Lundalogik/lime-elements/issues/2749
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -321,14 +321,14 @@ export class Picker {
     }
 
     private createChip(listItem: ListItem): Chip {
+        const name = getIconName(listItem.icon);
+        const color = getIconFillColor(listItem.icon, listItem.iconColor);
+
         return {
             id: `${listItem.value}`,
             text: listItem.text,
             removable: true,
-            icon: {
-                name: getIconName(listItem.icon),
-                color: getIconFillColor(listItem.icon, listItem.iconColor),
-            },
+            icon: name ? { name: name, color: color } : undefined,
             value: listItem,
         };
     }


src/components/picker/picker.tsx
```diff
@@ -321,14 +321,14 @@ export class Picker {
     }
 
     private createChip(listItem: ListItem): Chip {
+        const name = getIconName(listItem.icon);
+        const color = getIconFillColor(listItem.icon, listItem.iconColor);
```
LawrenceBorst, line 325: Could we use `listItem.icon?.color` here instead? Since `iconColor` is marked as deprecated

src/components/picker/picker.tsx
```diff
@@ -321,14 +321,14 @@ export class Picker {
...
+
         return {
             id: `${listItem.value}`,
             text: listItem.text,
             removable: true,
-            icon: {
-                name: getIconName(listItem.icon),
-                color: getIconFillColor(listItem.icon, listItem.iconColor),
-            },
+            icon: name && color ? { name: name, color: color } : undefined,
```
LawrenceBorst, line 331: But colours are optional, so things would still render even if `color` were `falsy`?

src/components/picker/picker.tsx
LucyChyzhova: we should have both since `iconColor` still in used

src/components/picker/picker.tsx
LucyChyzhova: good catch, I added a fixup

LucyChyzhova: Uploading Screen Recording 2024-01-31 at 10.54.45.mov…

LawrenceBorst (approved):

PR #2771, opened 2024-02-07
Don't block all pointer events on disabled components, just block the click event instead

Kiarokh:
When the host element is disabled, but has an `onClick`
event defined, we filter that out now, instead of preventing
clicks using a CSS hack.

fix https://github.com/Lundalogik/crm-feature/issues/3317

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button-group/button-group.scss
+++ b/src/components/button-group/button-group.scss
@@ -146,6 +146,7 @@
 
     &.disabled {
         @include shared_input-select-picker.looks-disabled;
+        pointer-events: none;
     }
 }
 
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -184,6 +184,7 @@ limel-icon.mdc-chip__icon.mdc-chip__icon--leading {
 
     &.disabled:not(.mdc-chip-set--input) {
         @include shared_input-select-picker.looks-disabled;
+        pointer-events: none;
     }
 }
 
@@ -256,6 +257,7 @@ limel-icon.mdc-chip__icon.mdc-chip__icon--leading {
 
     &.disabled {
         @include shared_input-select-picker.looks-disabled;
+        pointer-events: none;
     }
 }
 
--- a/src/components/icon-button/icon-button.scss
+++ b/src/components/icon-button/icon-button.scss
@@ -8,18 +8,16 @@
     display: none;
 }
 
-:host([disabled]) {
-    pointer-events: none;
-}
-
 @include mixins.visualize-aria-expanded('button');
 
 button {
     all: unset;
-    @include mixins.is-flat-clickable(
-        $background-color: var(--icon-background-color, transparent)
-    );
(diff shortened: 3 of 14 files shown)


src/components/chip-set/chip-set.scss
```diff
@@ -184,6 +184,7 @@ limel-icon.mdc-chip__icon.mdc-chip__icon--leading {
 
     &.disabled:not(.mdc-chip-set--input) {
         @include shared_input-select-picker.looks-disabled;
+        pointer-events: none;
```
Kiarokh, line 187: First I wanted to refactor some stuff in the chip-set, as we are disgustingly handling the `disabled` stated on chips of different types, using CSS hacks and `tabindex` hacks. However, since we are gonna refactor the chip-set and use `limel-chip` hopefully soon, this won't be an issue in future. So I'm just recreating what we used to get from the mixin here instead.

src/components/button-group/button-group.scss
```diff
@@ -146,6 +146,7 @@
 
     &.disabled {
         @include shared_input-select-picker.looks-disabled;
+        pointer-events: none;
```
Kiarokh, line 149: same as [`03a0c96` (#2771)](https://github.com/Lundalogik/lime-elements/pull/2771/commits/03a0c964e4ccaac2d6489f4dfd1f22ed5595d6ef#r1481182430)

adrianschmidt: I don't understand this comment… 😅

adrianschmidt: Ah, you meant to link to the comment on `chip-set.scss` I assume? Then I understand 👍

src/components/icon-button/icon-button.scss
adrianschmidt, line 1: Regarding `fix(icon-button): correctly visualize disabled styles when elevated`: I can't see the difference between the PR-docs and the docs for latest. Is this just a fix for something that happened in the previous commit?

adrianschmidt (approved):
Fantastic! 👌

src/components/icon-button/icon-button.scss
Kiarokh: Oh you're right. This was introduced by the same PR 😓 Gonna squash it to its previous commit and push. OK?

src/components/split-button/split-button.scss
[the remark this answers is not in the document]
Kiarokh: Since I'm focused on something completely different now, I suggest we make a new issue for this. 🙏

src/components/icon-button/icon-button.scss
adrianschmidt: 👍

PR #2776, opened 2024-02-08
fix(components): decrease the height and headspace of input elements

Kiarokh:
The input field was necessarily tall and the height and the UI design of
all other input elements such as slider, select, picker, etc… was based
on the height & visual density of the `limel-input-field`. This resulted in
UI designs in consuming applications which were too "airy" and had a
low information density. This change will make all input elements smaller
which prevent such UI and layout design issues. Additionally, `limel-select`
used to offer a narrow UI design, which was enabled by adding an `is-narrow`
class on the component. That would render it smaller in height.
Since the component will render with the same narrow hight now by default,
that feature is removed and adding that class will not have any
visible effects.

BREAKING CHANGE:
If you have layout designs with hardcoded size values (for instance
heights, margins or paddings) which are added only for the purpose of
horizontally or vertically aligning the elements in the UI,
there is a big chance that your layouts now will render
in a messy way after this change and elements will not align as before.
Note that for UI designs nowadays,
you should not need to have hardcoded values. You must rather
think "responsive design", and incorporate `flex` or `grid` to achieve
a more flawless, smooth and responsive layout.

fix https://github.com/Lundalogik/crm-feature/issues/3916

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -39,6 +39,7 @@ $leading-icon-space: 1.5rem;
 
 .mdc-chip-set {
     display: flex;
+    flex-wrap: wrap;
     align-items: center;
     gap: 0.5rem;
     min-height: shared_input-select-picker.$height-of-mdc-text-field;
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -31,7 +31,7 @@
 @include shared_input-select-picker.lime-looks-like-input-value;
 
 $height-of-chip-set-input: functions.pxToRem(36);
-$leading-icon-space: functions.pxToRem(40);
+$leading-icon-space: 1.5rem;
 
 :host(limel-chip-set) {
     isolation: isolate;
@@ -51,8 +51,7 @@ $leading-icon-space: functions.pxToRem(40);
     }
 
     &.mdc-chip-set--input {
-        padding: functions.pxToRem(8);
-
+        padding: 0.4rem 0.5rem;
         width: 100%;
     }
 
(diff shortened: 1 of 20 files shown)


src/components/chip/chip.scss
Kiarokh, line 1: prevents these scenarios 
image

src/components/helper-line/helper-line.scss
```diff
@@ -9,6 +9,7 @@
     -moz-osx-font-smoothing: grayscale;
     -webkit-font-smoothing: antialiased;
     font-size: 0.6875rem;
+    line-height: normal;
```
Kiarokh, line 12: Screen Recording 2024-02-19 at 12.02.47

src/components/helper-line/helper-line.scss
Kiarokh, line 1: https://github.com/Lundalogik/lime-elements/assets/35954987/c0dcf151-c282-4a4f-bbb0-b2b73ed942ba

There is a known issue that probably needs some sort of refactoring in the input field. 
The height of `<input-field type="textarea"` is changing when user clicks on the field in a `limel-form`. However, this is not a big deal in the form, and the effect is kind of nice for now 😅 So I'll try to fix this later.

Kiarokh: this seems to happen **only** if number of rows is unspecified for the input field, in Lime Admin.

src/components/chip-set/chip-set.scss
adrianschmidt, line 1: I changed the commit message of the first commit so it's not a breaking change. Since it will only break layouts if you've styled things in ways you're not supposed to, I don't think we need to treat this as breaking. But I've made it so that the message from the breaking change is still included in the changelog.

I'm not entirely sure how that will render when the changelog is included in lime-webclient and lime-crm, but we'll see 😅

adrianschmidt (approved):

PR #2777, opened 2024-02-08
refactor(chip-set): replace mdc-chip with limel-chip 

Kiarokh:
Only possible to see this when the chip has a `href`,
is `removable`, and is `disabled`

fix https://github.com/Lundalogik/crm-feature/issues/3917

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -1,10 +1,5 @@
-import { Chip } from '../chip-set/chip.types';
+import { Chip, ChipType } from '../chip-set/chip.types';
 import { Languages } from '../date-picker/date.types';
-import {
-    MDCChipInteractionEvent,
-    MDCChipSelectionEvent,
-    MDCChipSet,
-} from '@material/chips/deprecated';
 import { MDCTextField } from '@material/textfield';
 import {
     Component,
@@ -21,14 +16,8 @@ import { handleKeyboardEvent } from './chip-set-input-helpers';
 import translate from '../../global/translations';
 import { getHref, getTarget } from '../../util/link-helper';
 import { isEqual } from 'lodash-es';
-import {
-    getIconBackgroundColor,
-    getIconColor,
-    getIconName,
-    getIconTitle,
-} from '../icon/get-icon-props';
+import { LimelChipCustomEvent } from 'src/components';
 
-const SELECTED_CHIP_CLASS = 'mdc-chip--selected';
 const INPUT_FIELD_TABINDEX = 1;
 
 /**
@@ -58,7 +47,7 @@ export class ChipSet {
      * Type of chip set
      *
      * - `choice` renders a set of selectable chips where only one is selectable. The `removable` property is ignored
-     * - `filter` renders a set of selectable chips where all are selectable. The `icon` property is ignored
+     * - `filter` renders a set of selectable chips where all are selectable.
      * - `input` renders a set of chips that can be used in conjunction with an input field
      *
      * If no type is set, a basic set of chips without additional functionality will be rendered
@@ -208,7 +197,9 @@ export class ChipSet {
     @State()
     private inputChipIndexSelected: number = null;
 
-    private mdcChipSet: MDCChipSet;
+    @State()
+    private selectedChipIds: Array<string | number>;
+
     private mdcTextField: MDCTextField;
     private handleKeyDown = handleKeyboardEvent;
 
@@ -216,8 +207,6 @@ export class ChipSet {
         this.renderChip = this.renderChip.bind(this);
         this.renderInputChip = this.renderInputChip.bind(this);
         this.isFull = this.isFull.bind(this);
-        this.handleInteractionEvent = this.handleInteractionEvent.bind(this);
-        this.handleSelection = this.handleSelection.bind(this);
         this.handleTextFieldFocus = this.handleTextFieldFocus.bind(this);
         this.handleInputBlur = this.handleInputBlur.bind(this);
         this.handleTextInput = this.handleTextInput.bind(this);
@@ -227,6 +216,18 @@ export class ChipSet {
         this.handleDeleteAllIconClick =
             this.handleDeleteAllIconClick.bind(this);
         this.renderDelimiter = this.renderDelimiter.bind(this);
+    }
+
+    public connectedCallback() {
+        this.initialize();
+    }
+
+    private initialize() {
+        if (this.value.length) {
+            this.selectedChipIds = this.value
+                .filter((chip) => chip.selected)
+                .map((chip) => chip.id);
+        }
     }
 
     /**
@@ -279,16 +280,9 @@ export class ChipSet {
                 this.host.shadowRoot.querySelector('.mdc-text-field'),
             );
         }
-
-        this.createMDCChipSet();
-    }
-
-    public componentWillUpdate() {
-        this.destroyMDCChipSet();
     }
 
     public componentDidUpdate() {
-        this.createMDCChipSet();
         const input = this.host.shadowRoot.querySelector('input');
         if (input && this.editMode) {
             input.focus();
@@ -296,8 +290,6 @@ export class ChipSet {
     }
 
     public disconnectedCallback() {
-        this.destroyMDCChipSet();
-
         if (this.mdcTextField) {
             this.mdcTextField.destroy();
         }
@@ -322,14 +314,25 @@ export class ChipSet {
             classes['chip-set--with-label'] = true;
         }
 
+        const value = this.getValue();
+
         return (
             <div class={classes} role="grid">
                 {chipSetLabel}
-                {this.value.map(this.renderChip)}
+                {value.map(this.renderChip)}
                 {this.renderHelperLine()}
             </div>
         );
     }
+
+    private getValue = () => {
+        return this.value.map((chip) => ({
+            ...chip,
+            ...(this.type && {
+                selected: this.selectedChipIds.includes(chip.id),
+            }),
+        }));
+    };
 
     @Watch('value')
     protected handleChangeChips(newValue: Chip[], oldValue: Chip[]) {
@@ -338,35 +341,7 @@ export class ChipSet {
         }
 
         this.syncEmptyInput();
-    }
-
-    private createMDCChipSet() {
-        this.mdcChipSet = new MDCChipSet(
-            this.host.shadowRoot.querySelector('.mdc-chip-set'),
-        );
-
-        if (!this.type || this.type === 'input') {
-            this.mdcChipSet.listen(
-                'MDCChip:interaction',
-                this.handleInteractionEvent,
-            );
-        }
-
-        if (this.type === 'choice' || this.type === 'filter') {
-            this.mdcChipSet.listen('MDCChip:selection', this.handleSelection);
-        }
-    }
-
-    private destroyMDCChipSet() {
-        if (this.mdcChipSet) {
-            this.mdcChipSet.unlisten(
-                'MDCChip:interaction',
-                this.handleInteractionEvent,
-            );
-            this.mdcChipSet.unlisten('MDCChip:selection', this.handleSelection);
-
-            this.mdcChipSet.destroy();
-        }
+        this.initialize();
     }
 
     private renderChipSetLabel() {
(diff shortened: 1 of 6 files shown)


Kiarokh:
@Befkadu1 I moved some of the commits to a separate PR https://github.com/Lundalogik/lime-elements/pull/2780

src/components/chip-set/chip-set.tsx
```diff
@@ -208,16 +196,19 @@ export class ChipSet {
     @State()
     private inputChipIndexSelected: number = null;
 
-    private mdcChipSet: MDCChipSet;
+    @State()
+    private selectedChip: Chip;
+
+    @State()
+    private selectedChipIds: Array<string | number>;
```
adrianschmidt, line 203: So, can there only be one chip selected, or can there be multiple chips selected?

adrianschmidt, line 203: Also, what does "selected" mean? Does it mean that the chip is highlighted, so that pressing backspace removes it, or does it mean that the chip is "in" the chip-set? Or something else?

src/components/chip-set/chip-set.tsx
```diff
@@ -653,89 +575,82 @@ export class ChipSet {
...
+                {...chipProps}
+                onRemove={this.removeChip(chip.id)}
+            />,
             this.renderDelimiter(),
         ];
     }
 
-    private catchInputChipClicks(event) {
-        event.stopPropagation();
+    private removeChip = (id: string | number) => () => {
```
adrianschmidt, line 594: I'm a little worried that this construct might cause _very_ many functions to be created, since it will create a new function not just for each chip, but each time a chip is re-rendered. I don't really know how often that might be. Perhaps it's only a few times in a chip's lifetime, but if it happens for all the chips anytime anything in the chip-set is updated, that could be a little bit nasty…

We'd avoid this completely if the chip could accept some kind of identifier, and then emit that same value on the `remove` event. It doesn't mean we have to make that property required. It could still be optional, and if it's set, we attach it to the `remove` event.

(Ping me if you would like me to help implement it.)

src/components/chip-set/chip-set.tsx
```diff
@@ -653,89 +575,82 @@ export class ChipSet {
...
 
-        if (backgroundColor) {
-            style['--icon-background-color'] = backgroundColor;
+        this.emitInteraction(chip);
+    };
+
+    private updateSelectedChipIds() {
+        const { id } = this.selectedChip;
+        if (this.type === 'choice') {
+            this.selectedChipIds = this.selectedChipIds[0] === id ? [] : [id];
```
adrianschmidt, line 642: Ah! Ok. Now I see what `this.selectedChipIds` is used for 🙂

src/components/chip-set/chip.types.ts
```diff
@@ -89,3 +89,5 @@ export interface Chip<T = any> {
      */
     href?: string;
 }
+
+export type ChipType = 'default' | 'filter';
```
adrianschmidt, line 93: This needs a docblock.

```suggestion
/**
 * This type is used to determine the visual style and behavior of a Chip component.
 *
 * @beta
 */
export type ChipType = 'default' | 'filter';
```

src/components/chip-set/chip-set.tsx
adrianschmidt: > So, can there only be one chip selected, or can there be multiple chips selected?

I found the answer to this further down in the code.

adrianschmidt: For `choice` and `filter`, I guess it's pretty straightforward what "selected" means. I had forgotten that we had those variants 😅

Does "selected" have any meanings other than that as well?

src/components/chip-set/chip.types.ts
Kiarokh: do we really need a `@beta` tag here? I know it's new and shiny, but I don't think we need it for this private component ;) :P

src/components/chip-set/chip-set.tsx
```diff
@@ -653,89 +575,82 @@ export class ChipSet {
...
-        const title = getIconTitle(chip.icon, chip.iconTitle);
-        const style = {};
-        if (color) {
-            style['--icon-color'] = color;
+    private catchInputChipClicks = (chip: Chip) => () => {
+        if (this.type !== 'input') {
+            const selected = chip.selected;
+            chip.selected = !selected;
+            this.selectedChip = chip;
+            this.updateSelectedChipIds();
```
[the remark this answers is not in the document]
Kiarokh: Here is the background story: 
I think previously mdc chipset's own scripts took care of a lot of things. We weren't sure how exactly they worked. Sometimes simply adding the class of `mdc-chip` on `limel-chip` made it work like before. But sometimes it didn't help. So we decided to remove everything that comes from MDC, and make sure that all examples and implementations work like before, but with our own logics instead of MDC's. Specially because we wanna get rid of MDC generally.

Now I cannot judge the implementation, as the King Bef has done it and my knowledge is too little for evaluating it. But as far as our tests showed, things worked like before when we were done.

But if things can be improved, just feel free to do so with fresh eyes.

src/components/chip-set/chip.types.ts
Kiarokh, line 1: **Rant:**
The tricky thing for us in this PR was the fact that the chipset does so many things and has all these different `type`s that are each doing their own thing. Also it's used in limel-file, limel-picker, and other components (?). It also leaves some of the implementation for the consumer, like selecting "all", in this example 👇 
image which I think has a bit strange interaction design.

As if these weren't not enough, we have this other component called `limel-button-group` which also somehow uses the chip interface and some sort of `mdc-chip`. It's basically another variation of a chip-set but with a slightly different behaviour 😓 😄 

I'm really not a big fan of this component, and honestly never understood why we have something called chipset, with type of input, and another component called picker that uses chipset. And why we make a fake copy of `limel-input-field`'s internal HTML structure in chip-set (but the structures don't exactly match). My brain still short-circuits when I try to think about all these parts and make sense of them. 😅 

Hopefully we can simplify things with time, and clean up some of the mess. 🤗

src/components/chip-set/chip.types.ts
adrianschmidt: We do and we don't. We don't technically _need_ it, as the PR-checks will be green anyway. And we get this little warning in the api-extractor output:

image

That is as it should be, because it should warn us about that. But, the problem is that when you later make the component and type public, and you perhaps forget to add a docblock to the type, there is no update in the api-extractor output, which means the maintainer reviewing it will not be notified that there should be a docblock added.

And the type is visible in the docs, even if you consider it internal:

image

If it's meant to be internal permanently, not just while the `limel-chip` component is private, it should be marked with `@internal`.

src/components/chip-set/chip-set.tsx
Befkadu1: `selected` mean when `chip.select = true`

src/components/chip-set/chip.types.ts
Kiarokh: I can't really answer the question. These things are really confusing for me. I trust you do what is right ;)

src/components/chip-set/chip-set.tsx
Befkadu1: Why didn't we pass the chip object to `<limel-chip>` instead of passing many properties? I could not understand the reason behind.

Kiarokh: yes, I also don't really understand why limel-chip is importing the `Chip` interface, but also repeats everything from the Chip interface, as a prop inside itself. But generally I don't have a good understanding of theses things. I just assumed that was needed. Maybe because limel-chip might become a public component in future? You guys discussed this here: https://github.com/Lundalogik/lime-elements/pull/2707#discussion_r1452633337

src/components/chip-set/chip-set.tsx
Kiarokh: selected also adds styles on `limel-chip`

src/components/chip-set/chip-set.tsx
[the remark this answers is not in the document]
Befkadu1: > The way the code currently looks, it seems a little unnecessary to have the this.selectedChip property at all, especially as a @State-prop, since all it does is hold the value of the selected chip from line 631, to line 632, where it is immediately deconstructed, and only the id is used 😅

The reason is when selecting(or highlighting or clicking whatever we call it 😅) the chips having the chip type = 'filter' for some reason it could not trigger re-rendering even if `this.selectedChipIds` is changed. You can remove the  state `this.selectedChip` and try it. I am interested to know why it could not re-render updating `this.selectedChipIds`

src/components/chip-set/chip-set.tsx
Befkadu1: He is asking why we add a prefix `selected` in `selectedChipIds` 😅

src/components/chip-set/chip-set.tsx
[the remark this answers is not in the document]
adrianschmidt: > I am interested to know why it could not re-render updating `this.selectedChipIds`

That's because the `splice` on line 650 and the `push` on line 652 mutates the existing array, instead of replacing it with a new array. Because it's the same array, only with changed content, Stencil has no way of detecting the change.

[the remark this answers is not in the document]
Befkadu1: Nice I did not know that Stencil doesn't detect the modification. It doesn't detect the changes since the array reference remains the same. 

I will change the `splice` code with

```
this.selectedChipIds = this.selectedChipIds
                .filter((chipId: number | string) => chipId !== id);
```

and the `push` one to `this.selectedChipIds = [...this.selectedChipIds, id];`.
Finally I will remove `this.selectedChip`

src/components/chip-set/chip-set.tsx
adrianschmidt: > He is asking why we added a prefix `selected` in `selectedChipIds` 😅

Not so much why you added the prefix, as: _What does it signify that a chip is "selected"?_

As I mentioned before, I had forgotten that a chip-set could be anything other than the `input` type 🙈

My question was more, does it mean all of these things?

A 'choice' chip-set, with the chips 'Lime', 'Apple', and 'Banana'. The 'Apple' chip looks like it's pressed down and is rendered with green text

A 'filter' chip-set, with the chips 'Lime', 'Apple', and 'Banana'. The 'Apple' chip looks like it's pressed down and is rendered with green text, and a green 'light' in the bottom left corner

An 'input' chip-set, with the chips 'Elephant', 'Caterpillar', 'Badger', and 'Fish'. The 'Caterpillar' chip has a red outline

The first two I could answer "yes" by looking at the examples in the docs, and the last one I found out is also "yes", by looking a little closer at the code.

I felt like at least the last example, with the red outline, is quite different from the others. If it was a completely new addition, I would have probably complained about it, but since it's just an adaptation of `chip.selected`, and that already worked like this, it's ok.

adrianschmidt: I consider this question resolved, so I'll mark the thread as such 🙂

src/components/chip-set/chip-set.tsx
Befkadu1: This runs only once when you click remove button otherwise I don't think it runs each time a chip is re-rendered. 

>We'd avoid this completely if the chip could accept some kind of identifier, and then emit that same value on the remove event. It doesn't mean we have to make that property required. It could still be optional, and if it's set, we attach it to the remove event.

That would be better to have an identifier prop. I did not understand why it was not passed from the beginning. That is why I used a curry here. You decide the best one I will add a fixup if you are not super happy

adrianschmidt: > This runs only once when you click remove button otherwise I don't think it runs each time a chip is re-rendered.

The inner function will run only when the button is clicked. The _outer_ function, that creates and returns the inner function, will run each time the template is rendered.

I'll experiment a little with it.

adrianschmidt: I added this `console.log` to the outer function:

image

…and it's called a _lot…_

https://github.com/Lundalogik/lime-elements/assets/154725/75fc7d14-df24-4b6a-9cc7-14bdba43cdfa

Befkadu1: Nice argument. Should we have an identifier prop in the child component then?

src/components/chip-set/chip.types.ts
adrianschmidt: Didn't you make the button-group component Kia? 🤨 😄

But yeah, I can only agree. This is the technical debt we settle ourselves with when we only rush, rush, rush… 🙈

At least we have the ability to spend a little bit of time cleaning up _sometimes…_

I'm sorry I haven't been able to put more focus on this the last few days. It's probably get calmer next week, at least towards the end of the week, when people start to settle in to the cycle. Cooldown is just insanely high stress for me. It's two weeks straight of trying to handle at least half a dozen simultaneous requests for attention, basically never getting anything actually finished, because new things pop up as soon as I think I'm about to have a chance to actually spend a little time on something 😭

src/components/chip-set/chip-set.tsx
adrianschmidt: Yes, I think that's the way to go 👍

But it should be optional, so people who might use the chip standalone (once we make it public), doesn't have to supply one.

Befkadu1: https://github.com/Lundalogik/lime-elements/pull/2777/commits/33e8bccc42a531327a5246f3f046b70682f71be0 I leave it for you if you don't mind. Both @Kiarokh and @adrianschmidt you guys can update the code if you find it unreadable or difficult to understand.

src/components/chip-set/chip.types.ts
Kiarokh: > Didn't you make the button-group component Kia? 🤨 😄

Technically, not really. I just worked on the UI. 😅 Completely innocent. 

For the rest, I understand what you mean. ⭐️ it's OK.

src/components/chip-set/chip-set.tsx
Befkadu1: @adrianschmidt are you aware that you removed this fixup when you squash it?

adrianschmidt: Oh, no, I'm sorry! I don't know how that happened 🙈

adrianschmidt: I didn't see that you had added that, so I was starting to fix it myself…

I'll add it back in 👍

adrianschmidt: I had made some additional changes, so I combined your changes and mine 👍

I've pushed them now.

adrianschmidt: Hmm… deleting chips with the keyboard doesn't work at the moment. I must have missed something.

adrianschmidt: Ah, now I see why you had added a `handleRemoveChip` 😄

adrianschmidt:
> Only possible to see this when the chip has a `href`,
> is `removable`, and is `disabled`

What does that mean? 🤔

src/components/chip-set/chip-set.tsx
adrianschmidt: ⚡

src/components/chip/chip.tsx
adrianschmidt, line 1: I think we can make limel-chip public now, right?

@Kiarokh

adrianschmidt (approved):

adrianschmidt (approved):

Issue #2420, opened 2023-07-27
Keyboard opens automatically when dialog is rendered in android


omaralweli:
## Current behavior

The bug was discovered in esign #https://github.com/Lundalogik/limepkg-esign/issues/1405 where a phone keyboard automatically pops up every time the `sign the document` button in the portal is pressed. But after further investigation it looks like the behaviour is linked to `limel-dialog`.  

1. Open any `limel-dialog` from an android phone.
2. 
![2](https://github.com/Lundalogik/lime-elements/assets/77667105/5b755652-cf5d-4513-b153-23e753794282)

## Expected behavior

The keyboard does not automatically pops up on an android phone when `limel-dialog` is rendered and there are no text input fields.

justFredster:
@Kiarokh @LucyChyzhova is this something you can have a look at? Or is it @adrianschmidt? 

It is really annoying in the eSign portal, when you click the "Sign" button, but shouldn't type any text. See issue in eSign repo for more info: https://github.com/Lundalogik/limepkg-esign/issues/1405

Kiarokh:
😮 is it because we are setting `focus` on the surface maybe Adrian?

adrianschmidt:
I know we had a problem that the focus-trap* that MDC was using didn't work because it didn't support shadow-DOM. I think that was later solved by someone, I think it might have been @maltja. If I recall correctly, the solution involved adding a hidden element, which we set focus to. I'm not sure about this, but it's what my vague memories suggest. If that's correct, and that element is something like a visually hidden input field, that would explain why the soft keyboard shows up on mobile devices… 😕

*) A focus-trap is a piece of code that sets focus within a certain element (like a modal), and prevents focus from leaving said element as long as it is open (focusing elements inside the modal should work fine, but tabbing out of the modal and focusing elements on the page behind should be prevented). If it's a really good focus-trap, it might also cover placing focus back on the element that was focused when the focus-trapped element (the modal) was opened.

PR #2779, opened 2024-02-09
fix(dialog): avoid keyboard automatically opening when dialog opens on android

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/2420

https://github.com/Lundalogik/lime-elements/assets/50618208/42daedb6-8f6b-40f3-bedf-72b699d878a9

but:

https://github.com/Lundalogik/lime-elements/assets/50618208/d322dfb9-024d-42bd-a284-1d3a259de210

However, if we change the input type to "button" everything magically works 🥹

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/dialog/dialog.tsx
+++ b/src/components/dialog/dialog.tsx
@@ -164,7 +164,7 @@ export class Dialog {
                             the focus trap to use. Read more here:
                             https://github.com/material-components/material-components-web/tree/v11.0.0/packages/mdc-dialog#handling-focus-trapping
                         */}
-                        <input type="text" id="initialFocusElement" />
+                        <input type="button" id="initialFocusElement" />
                         {this.renderHeading()}
                         <div
                             class="mdc-dialog__content"


LucyChyzhova:
There are two ways to solve it, the first one (fix) will break the focus trap - so it is not the best option

The second one (fixup) - should be judged by @adrianschmidt :) - works fine but feels too simple to be the truth 😄

adrianschmidt:
iOS doesn't seem to have this problem to begin with?

Either way, I could recreate it on Android using the latest release, and it was fixed in this PR. The focus trapping still works. Great job!

adrianschmidt (approved):
Such a perfect little fix! 😍

LucyChyzhova:
> iOS doesn't seem to have this problem to begin with?
> 
no, no problem on iOS

FredrikWallstrom:
Amazing, great thanks! ❤️

joelricci:
Sweet! Well done.

PR #2782, opened 2024-02-12
fix(config): revert unintentional breaking change in 37.2.0

adrianschmidt:
re 852282d

fix: #2781 

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/global/config.ts
+++ b/src/global/config.ts
@@ -17,7 +17,7 @@ export type Config = {
     /**
      * @internal
      */
-    featureSwitches?: any;
+    featureSwitches?: Record<string, boolean>;
 };
 
 class ConfigClass implements Config {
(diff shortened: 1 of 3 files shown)


src/global/config.ts
```diff
@@ -1,4 +1,26 @@
...
+
+    /**
+     * The default locale to use for components that support localization.
+     */
+    defaultLocale?: string;
+
+    /**
+     * @internal
+     */
+    featureSwitches?: any;
```
jgroth, line 20: This should probably be `Record<string, boolean>`

adrianschmidt: ⚡

jgroth (approved):

Issue #2758, opened 2024-02-02
limel-picker - multiple: remove one value clears all when value is object


BregenzerK:
## Current behavior

When you use `limel-picker` with `multiple` and an object as value and you remove one value, all values are removed.
![image](https://github.com/Lundalogik/lime-elements/assets/10260886/f54dab87-b43a-4a4e-84a3-077e2f337ba1)
 

1. `limel-picker` like this:
```
<limel-picker
                label={"Test"}
                multiple={true}
                searcher={this.search}
                onChange={this.handleChange}
                value={this.recipientsValue}
            />


private handleChange = (
        event: LimelPickerCustomEvent<Array<ListItem<RecipientValue>>>
    ) => {
        this.recipientsValue = event.detail;
}

export interface RecipientValue {
    id: number;
    limetype: string;
    address: string;
    name: string;
}
```
2. pick several values
3. remove one of them --> all values are cleared, as shown in the screenshot, the picker emits an empty array

## Expected behavior

Only remove the value that was clicked on. It works fine, when value on `ListItem` is a simple value e.g. a number instead. 

## Environment

- **lime-elements** version: "37.1.0-next.97"

- Framework used: StencilJs

Kiarokh:
@BregenzerK are you sure this is not dependent on your implementation? I'm testing it other places, but it seems to be working fine. 🤔

BregenzerK:
> No I was just testing with existing examples and implementations of the picker, such as in Lime Elements. I don't think I'm the right person for this task to be honest. It'd be way more effective if you or your team makes a PR i elements I'd say.

It doesn't have to be you and it isn't super urgent. Sure we can look at it, but so can anyone in the feature team I guess

LucyChyzhova:
UPD: Katja and I created an example that contains an object as a value and we could reproduce the problem  in `lime-elements`.  I will continue with an investigation.

https://github.com/Lundalogik/lime-elements/assets/50618208/7569adcb-2c14-475b-92ba-643b54685a9a

Kiarokh:
please send me the example code @LucyChyzhova and we will test it in this PR: https://github.com/Lundalogik/lime-elements/pull/2777

adrianschmidt:
[...]

PR #2784, opened 2024-02-13
fix(picker): be able to have value as object

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/2758

## Review:

https://github.com/Lundalogik/lime-elements/assets/50618208/9483c045-3c88-4c97-87f7-1f031d6f9859

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -38,6 +38,7 @@ const CHIP_SET_TAG_NAME = 'limel-chip-set';
  * @exampleComponent limel-example-picker-single
  * @exampleComponent limel-example-picker-multiple
  * @exampleComponent limel-example-picker-icons
+ * @exampleComponent limel-example-picker-value-as-object
  * @exampleComponent limel-example-picker-empty-suggestions
  * @exampleComponent limel-example-picker-leading-icon
  * @exampleComponent limel-example-picker-static-actions
@@ -106,10 +107,18 @@ export class Picker {
     public invalid = false;
 
     /**
-     * Currently selected value or values
+     * Currently selected value or values. Where the value can be an object.
      */
     @Prop()
-    public value: ListItem<number | string> | Array<ListItem<number | string>>;
+    public value:
+        | ListItem<
+              number | string | { id: string | number; [key: string]: any }
+          >
+        | Array<
+              ListItem<
+                  number | string | { id: string | number; [key: string]: any }
+              >
+          >;
 
     /**
      * A search function that takes a search-string as an argument,
@@ -301,7 +310,16 @@ export class Picker {
         return null;
     }
 
-    private createChips(value: ListItem | ListItem[]): Chip[] {
+    private getValueId = (item: ListItem) => {
+        const value = item.value;
+        if (!!value && typeof value === 'object') {
+            return value.id;
+        }
+
+        return value;
+    };
+
+    private createChips = (value: ListItem | ListItem[]): Chip[] => {
         if (!value) {
             return [];
         }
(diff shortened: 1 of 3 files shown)


src/components/picker/picker.tsx
```diff
@@ -320,9 +321,11 @@ export class Picker {
     private createChip(listItem: ListItem): Chip {
         const name = getIconName(listItem.icon);
         const color = getIconFillColor(listItem.icon, listItem.iconColor);
+        const value = listItem.value;
+        const id = typeof value !== 'object' ? value : value.id;
```
adrianschmidt, line 325:
```suggestion
        const id = (!!value && typeof value === 'object') ? value.id : value;
```

src/components/picker/picker.tsx
```diff
@@ -322,7 +322,7 @@ export class Picker {
         const name = getIconName(listItem.icon);
         const color = getIconFillColor(listItem.icon, listItem.iconColor);
         const value = listItem.value;
-        const id = typeof value !== 'object' ? value : value.id;
+        const id = (!!value && typeof value === 'object') ? value.id : value;
```
Kiarokh, line 325: I was wondering if the root of the problem is that `ListItem` interface doesn't have `id`, 

https://github.com/Lundalogik/lime-elements/blob/4b731c74df02b5e2519bb81f803a90266d7f9948/src/components/list/list-item.types.ts#L8

while the `Chip` interface which is used in `limle-chip-set` expects `id`s:
https://github.com/Lundalogik/lime-elements/blob/4b731c74df02b5e2519bb81f803a90266d7f9948/src/components/chip-set/chip.types.ts#L10

but in this example that you were testing on, the consumer has to include the `id` in the object, due to the lack of `id` in the `ListItem` interface that is used in `limel-picker` (picker is using chip-set).

Hope you followed what I was trying to say. But I just wonder if we should revisit `ListItem` instead.

LucyChyzhova: I'm changing a bit, I will add a fixup soon, so you can check if it is still relevant

Kiarokh: I haven't really looked into the problem or the solution, I just saw this was some missmatch between what LiteItem interface allows the users to do, and what Chip interface has, which forces the users to implement the code in a way that becomes broken. But what is a good fix, I have no idea. I can't say where this problem should be fixed.

src/components/picker/picker.tsx
```diff
@@ -301,7 +310,12 @@ export class Picker {
         return null;
     }
 
-    private createChips(value: ListItem | ListItem[]): Chip[] {
+    private getValueId = (item: any) => {
```
Befkadu1, line 313:
```suggestion
    private getValueId = (item: ListItem) => {
```

src/components/picker/picker.tsx
```diff
@@ -301,7 +310,16 @@ export class Picker {
...
     }
 
-    private createChips(value: ListItem | ListItem[]): Chip[] {
+    private getValueId = (item: ListItem) => {
+        const value = item.value;
+        if (!!value && typeof value === 'object') {
+            return value.id;
+        }
+
+        return value;
```
Befkadu1, line 319: @adrianschmidt is there any way the `item.value = {}` if so do we need to handle the case somewhere?

adrianschmidt (approved):
This looks good enough for me at the moment. There may be further improvements that can be made once #2777 is merged, but that's a question for then.

src/components/picker/picker.tsx
adrianschmidt: The types indicate that `value` must be a string, a number, or an object with a property `id`, which is a string or a number. So `value` should never be an empty object, or an object without an `id`. But if it is, then the same bug will still be there. I don't think we need to support that case though. We wouldn't really have had to support the case with an object that has an `id` either, since the types already specified what's valid and what's not. But since many will likely want to use these with limeobjects‚ I guess it's helpful to allow that.

Although, if they decide to use the picker with a mix of limeobjects from different limetypes, the bug could still appear, if two picked objects had the same `id`, but different `limetypes`.

Once we've merged the changes to chip-set, I think we might want to look that over. I don't think there's any particular reason we need to compare things as strings, but we couldn't change that here without changing the `Chip` interface.

PR #2795, opened 2024-02-21
Feat/form expose schema interface

TommyLindh2:
While building the ViewFactory for the CardView for the web client we're gonna return the JSON schema for a limel-form.
Then it would be nice to have a public interface to type it with. We found that we had almost a finished interface for internal use here so we fixed the last pieces and want to expose it publicly and type the `schema` prop in `limel-form` as well.

This will make it easier to use the limel-form and FormComponents because you will get help from your IDE while writing code.

The PR tests itself because we don't get any type errors in the examples after adding our interface to those schemas.

Fix [Lundalogik/limepkg-basic-helpdesk#19](https://github.com/Lundalogik/limepkg-basic-helpdesk/issues/19)

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/form.types.ts
+++ b/src/components/form/form.types.ts
@@ -1,5 +1,23 @@
+import { JSONSchema7 } from 'json-schema';
 import { Help } from '../help/help.types';
 import { EventEmitter } from '@stencil/core';
+
+declare module 'json-schema' {
+    interface JSONSchema7 {
+        /**
+         * @internal
+         * Unique identifier for the schema
+         */
+        id?: string;
+
+        /**
+         * Lime elements specific options that can be specified in a schema
+         */
+        lime?: Omit<LimeSchemaOptions, 'layout'> & {
+            layout?: Partial<LimeLayoutOptions>;
+        };
+    }
+}
 
 /**
  * @public
@@ -110,12 +128,12 @@ export interface FormInfo {
     /**
      * The schema of the current property
      */
-    schema?: object;
+    schema?: FormSchema;
 
     /**
      * The schema of the whole form
      */
-    rootSchema?: object;
+    rootSchema?: FormSchema;
 
     /**
      * A tree of errors for this property and its children
@@ -177,15 +195,21 @@ export interface LimeSchemaOptions {
      * When specified on an object it will render the sub components with the
      * specified layout
      */
-    layout?: FormLayoutOptions<any>;
+    layout?: LimeLayoutOptions;
 
     /**
      * Mark the field as disabled
      */
     disabled?: boolean;
 
-    help?: string | Help;
+    help?: string | Partial<Help>;
 }
+
+/**
+ * @public
+ * Options for a layout to be used in a form
+ */
+export type LimeLayoutOptions = GridLayoutOptions & RowLayoutOptions;
 
 /**
  * Options for a component to be rendered inside a form
@@ -208,7 +232,9 @@ export interface FormComponentOptions {
 /**
  * @public
  */
-export interface FormLayoutOptions<T = FormLayoutType.Default> {
+export interface FormLayoutOptions<
+    T extends FormLayoutType | `${FormLayoutType}` = FormLayoutType.Default,
+> {
     /**
      * The type of layout to use
      */
@@ -217,9 +243,10 @@ export interface FormLayoutOptions<T = FormLayoutType.Default> {
 
 /**
  * @public
+ * Layout options for a grid layout
  */
 export interface GridLayoutOptions
-    extends FormLayoutOptions<FormLayoutType.Grid> {
+    extends FormLayoutOptions<FormLayoutType | `${FormLayoutType}`> {
     /**
      * When specified on a component within the grid, the component will take
      * up the the specified number of columns in the form
@@ -249,9 +276,10 @@ export interface GridLayoutOptions
 
 /**
  * @public
+ * Layout options for a row layout
  */
 export interface RowLayoutOptions
-    extends FormLayoutOptions<FormLayoutType.Row> {
+    extends FormLayoutOptions<FormLayoutType | `${FormLayoutType}`> {
     /**
      * When specified on a field, the chosen icon will be displayed
      * on the left side of the row, beside the title.
@@ -261,6 +289,7 @@ export interface RowLayoutOptions
 
 /**
  * @public
+ * Represents the layout types for a form.
  */
 export enum FormLayoutType {
     /**
@@ -282,3 +311,9 @@ export enum FormLayoutType {
      */
     Row = 'row',
 }
+
+/**
+ * @public
+ * Represents the JSON schema with Lime specific options
+ */
+export interface FormSchema extends JSONSchema7 {}
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -4,6 +4,8 @@
 
 ```ts
 
+import { JSONSchema7 } from 'json-schema';
+
 // @public (undocumented)
 export interface Action {
     // (undocumented)
@@ -341,11 +343,8 @@ export namespace Components {
     export interface LimelForm {
         "disabled": boolean;
         "errors": ValidationError;
-        "propsFactory"?: (schema: Record<string, any>) => Record<string, any>;
-        "schema": {
-            id?: string;
-            [key: string]: any;
-        };
+        "propsFactory"?: (schema: FormSchema) => Record<string, any>;
+        "schema": FormSchema;
         "transformErrors"?: (errors: FormError[]) => FormError[];
         "value": object;
     }
@@ -737,26 +736,30 @@ export interface FormError {
 export interface FormInfo {
     errorSchema?: object;
     name?: string;
-    rootSchema?: object;
+    rootSchema?: FormSchema;
     rootValue?: any;
-    schema?: object;
+    schema?: FormSchema;
     schemaPath?: string[];
 }
 
 // @public (undocumented)
-export interface FormLayoutOptions<T = FormLayoutType.Default> {
+export interface FormLayoutOptions<T extends FormLayoutType | `${FormLayoutType}` = FormLayoutType.Default> {
     type: T;
 }
 
-// @public (undocumented)
+// @public
 export enum FormLayoutType {
     Default = "default",
     Grid = "grid",
     Row = "row"
 }
 
-// @public (undocumented)
-export interface GridLayoutOptions extends FormLayoutOptions<FormLayoutType.Grid> {
+// @public
+export interface FormSchema extends JSONSchema7 {
+}
+
+// @public
+export interface GridLayoutOptions extends FormLayoutOptions<FormLayoutType | `${FormLayoutType}`> {
     colSpan?: 1 | 2 | 3 | 4 | 5 | 'all';
     columns?: 1 | 2 | 3 | 4 | 5;
     dense?: boolean;
@@ -1170,11 +1173,8 @@ namespace JSX_2 {
         "errors"?: ValidationError;
         "onChange"?: (event: LimelFormCustomEvent<object>) => void;
         "onValidate"?: (event: LimelFormCustomEvent<ValidationStatus>) => void;
-        "propsFactory"?: (schema: Record<string, any>) => Record<string, any>;
-        "schema"?: {
-            id?: string;
-            [key: string]: any;
-        };
+        "propsFactory"?: (schema: FormSchema) => Record<string, any>;
+        "schema"?: FormSchema;
         "transformErrors"?: (errors: FormError[]) => FormError[];
         "value"?: object;
     }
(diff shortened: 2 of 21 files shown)


TommyLindh2:
It seems like the build check fails because of changes of `etc/lime-elements.api.md`:
I don't know how to solve that.

```
Run npm run api:verify
  npm run api:verify
  shell: /usr/bin/bash -e {0}

> @limetech/lime-elements@[3](https://github.com/Lundalogik/lime-elements/actions/runs/7991024617/job/21821360880?pr=2795#step:9:3)7.[4](https://github.com/Lundalogik/lime-elements/actions/runs/7991024617/job/21821360880?pr=2795#step:9:5).0 api:verify
> api-extractor run


api-extractor 7.40.3  - https://api-extractor.com/

Using configuration from ./api-extractor.json
Analysis will use the bundled TypeScript version [5](https://github.com/Lundalogik/lime-elements/actions/runs/7991024617/job/21821360880?pr=2795#step:9:6).3.3
Warning: You have changed the public API signature for this project. Please copy the file "temp/lime-elements.api.md" to "etc/lime-elements.api.md", or perform a local build (which does this automatically). See the Git repo documentation for more info.

API Extractor completed with warnings
Error: Process completed with exit code 1.
```

TommyLindh2:
I can't manage to run the api update script locally on windows. Can someone help me run it.
Maybe @adrianschmidt can?

TommyLindh2:
> @TommyLindh2 Sure! I'm hunting down a critical bug right now, but we can have a huddle afterward? If you are getting blocked by not being able to update the file, I'm sure one of my teammates can run the update script on your branch in the meanwhile.

There's no rush for me, so I can wait

src/components/form/form.types.ts
```diff
@@ -282,3 +294,17 @@ export enum FormLayoutType {
      */
     Row = 'row',
 }
+
+/**
+ * @public
+ * Represents the JSON schema with Lime specific options
+ */
+export interface LimeJSONSchema extends JSONSchema7 {
```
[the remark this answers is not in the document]
TommyLindh2: Yeah, I can understand that.
One idea I had that we could try is to create our own schema interface for what we use in our examples and maybe in the webclient and some customizations (the common ones).

And then also add like _unknowns_ like `any`
```typescript
export interface FormSchema {
    id?: string | undefined;
    $id?: string | undefined;
    
    /**
     * Lime elements specific options that can be specified in a schema
    */
    lime?: LimeSchemaOptions;
    type?: FormSchemaTypeName | FormSchemaTypeName[] | undefined;

    /* ... More properties/types ... */
    
    /* Allow whatever we haven't defined */
    [unknown: string]: any;
```

This way we get InteliSense for writing schemas, and it's easier to work with the FormComponents and so on.
BUT you can also set everything we haven't added yet.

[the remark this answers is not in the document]
TommyLindh2: This would mean that the entire typing could look something like this:

```typescript

/**
 * @public
 * What type the schema represents
 */
type FormSchemaTypeName =
    | 'string'
    | 'number'
    | 'integer'
    | 'boolean'
    | 'object'
    | 'array'
    | 'null';

/**
 * @public
 * Utility type for representing a form object with recruisive types
 */
export interface FormObject {
    [key: string]: FormSchemaType;
}

/**
 * @public
 * Utility type for representing a form array with recruisive types
 */
export interface FormArray extends Array<FormSchemaType> {}

/**
 * @public
 * Available types in the schema
 */
export type FormSchemaType =
    | string
    | number
    | boolean
    | FormObject
    | FormArray
    | null;

/**
 * Meta schema
 *
 * Recommended values:
 * - 'http://json-schema.org/schema#'
 * - 'http://json-schema.org/hyper-schema#'
 * - 'http://json-schema.org/draft-07/schema#'
 * - 'http://json-schema.org/draft-07/hyper-schema#'
 *
 * @see https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-5
 */
export type SchemaVersion = string;

/**
 * @public
 * Represents the schema to render a limel-form
 */
export interface FormSchema {
    id?: string | undefined;
    $id?: string | undefined;
    $ref?: string | undefined;
    $schema?: SchemaVersion | undefined;
    $comment?: string | undefined;

    /**
     * Lime elements specific options that can be specified in a schema
     */
    lime?: LimeSchemaOptions;

    /**
     * Handle all schema properties that are not explicitly defined
     */
    [unknown: string]: any;

    /**
     * @see https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-00#section-8.2.4
     * @see https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#appendix-A
     */
    $defs?:
        | {
              [key: string]: FormSchema;
          }
        | undefined;

    /**
     * @see https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6.1
     */
    type?: FormSchemaTypeName | FormSchemaTypeName[] | undefined;
    enum?: FormSchemaType[] | undefined;
    const?: FormSchemaType | undefined;

    /**
     * @see https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6.3
     */
    maxLength?: number | undefined;
    minLength?: number | undefined;
    pattern?: string | undefined;

    /**
     * @see https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6.4
     */
    items?: FormSchema | FormSchema[] | undefined;
    additionalItems?: FormSchema | boolean | undefined;
    uniqueItems?: boolean | undefined;
    contains?: FormSchema | undefined;

    /**
     * @see https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6.5
     */
    required?: string[] | undefined;
    properties?:
        | {
              [key: string]: FormSchema;
          }
        | undefined;
    additionalProperties?: FormSchema | boolean | undefined;
    dependencies?:
        | {
              [key: string]: FormSchema | string[];
          }
        | undefined;

    /**
     * @see https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6.7
     */
    allOf?: FormSchema[] | undefined;
    anyOf?: FormSchema[] | undefined;
    oneOf?: FormSchema[] | undefined;
    not?: FormSchema | undefined;

    /**
     * @see https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-7
     */
    format?: string | undefined;

    /**
     * @see https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-9
     */
    definitions?:
        | {
              [key: string]: FormSchema;
          }
        | undefined;

    /**
     * @see https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-10
     */
    title?: string | undefined;
    description?: string | undefined;
    default?: FormSchemaType | undefined;
    readOnly?: boolean | undefined;
}

```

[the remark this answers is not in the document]
jgroth: I think it would be quite safe to expose this one. They seem to create a new interface for each version of the JSON schema specification, so risk of breaking changes are probably quite slim

src/components/form/form.types.ts
```diff
@@ -208,18 +215,21 @@ export interface FormComponentOptions {
...
  */
-export interface FormLayoutOptions<T = FormLayoutType.Default> {
+export interface FormLayoutOptions<
+    T extends FormLayoutType | `${FormLayoutType}` = FormLayoutType.Default,
+> {
     /**
      * The type of layout to use
      */
-    type: T;
+    type?: T;
```
jgroth, line 224: This looks like a breaking change, not sure why it's needed :thinking:

src/components/form/form.types.ts
```diff
@@ -208,18 +215,21 @@ export interface FormComponentOptions {
 /**
  * @public
  */
-export interface FormLayoutOptions<T = FormLayoutType.Default> {
+export interface FormLayoutOptions<
+    T extends FormLayoutType | `${FormLayoutType}` = FormLayoutType.Default,
```
jgroth, line 219: Not sure if I understand this change and the ones for the grd/row layout options :thinking:

src/components/form/form.types.ts
TommyLindh2: The reason for this change is because it was "wrong" before.
The layoutOption doesn't actually need a `type` for example if you have a schema with `layout: grid` it's possible for all properties in that schema to set only `colSpan: all` for example. Then you don't have to set the type on that layout.

And are we sure it counts as a breaking change if we make something optional? The other way around would totally be breaking but I think that this direction is fine?

src/components/form/form.types.ts
TommyLindh2: This typing change is a side effect of us using `enum` [A video showing why we shouldn't use enums in TypeScript](https://www.youtube.com/watch?v=jjMbPt_H3RQ)

So basically `` `${FormLayoutType}` `` turns an enum type to a type with a union of the values of the enum.
So you can in this case write `FormLayoutType.Default` OR `'default'` and both are valid. Which I think makes a lot more sense, AND Also you get intellisense of the types instead of needing to import the enum type and use that.

src/components/form/form.types.ts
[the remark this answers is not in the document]
TommyLindh2: Ok, so would you prefer the extending of `JSONSchema7` instead of writing our own simpler implementation?

src/components/form/form.types.ts
TommyLindh2: Yeah, that's true. Then I will just change the type to Partial in our new Schema instead. Fix coming soon

src/components/form/form.types.ts
```diff
@@ -282,3 +294,113 @@ export enum FormLayoutType {
...
+    /**
+     * {@inheritdoc}
+     */
+    not?: LimeFormSchemaDefinition;
+
+    /**
+     * {@inheritdoc}
+     */
+    definitions?: Record<string, LimeFormSchemaDefinition>;
+}
```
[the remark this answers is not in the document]
TommyLindh2: Ooh, that looks pretty interesting.
I have another PR #2801 where I want to make the Schema Generic so you can say that you want a Schema of type `Person` for instance. That will then type the required, properties and so on for you.

To support that fully, we will need to define all properties either way. But that if only if we want a generic schema which I think would be nice.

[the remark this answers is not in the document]
jgroth: The interfaces should be merged so in my example above it will contain everything from the original interface plus the new declaration that includes `lime`, so you should not have to redefine anything :slightly_smiling_face: 
But maybe I misunderstood something about what you meant :smile:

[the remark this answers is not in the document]
TommyLindh2: Yeah, I understand that.
But in the other PR later where I try to make it recursively generic, I will need to specify each property with a reference to itself either way because of the Generic typing. BUT that is something we could  add in that PR instead and leave this PR in a more readable and simple version.

Let's do that. Update coming soon 👍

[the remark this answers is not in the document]
TommyLindh2: Fixed, I still want to keep the `LimeFormSchema` name so that we can extend it later, and that we have our own name for it.

src/components/form/form.types.ts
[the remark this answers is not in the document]
TommyLindh2: Now we went with the extension of `JSONSchema7`. Seems to work for all our examples.

src/components/form/form.types.ts
```diff
@@ -282,3 +311,9 @@ export enum FormLayoutType {
      */
     Row = 'row',
 }
+
+/**
+ * @public
+ * Represents the JSON schema with Lime specific options
+ */
+export interface LimeFormSchema extends JSONSchema7 {}
```
jgroth, line 319: Any reason why this is prefixed with `Lime`? Most other public types/interfaces are not :thinking:

TommyLindh2: No, we can remove it. `FormSchema` maybe is better. I will update it

TommyLindh2: Fixed!

src/components/form/form.types.ts
jgroth: Yes, enums are weird in typescript and we should probably not use them if we can avoid it :smile: 
I think we started using object literals in some other places as an alternative, e.g.

```ts
export const Environment = {
    development: 'dev',
    production: 'prod',
    testing: 'test',
} as const;

export type Environment = (typeof Environment)[keyof typeof Environment];
```

TommyLindh2: Sweet, but I don't want to change the existing enum. But good for future reference

jgroth: Yeah that's out of scope here :smile:

TommyLindh2:
I've finally managed to update my node/npm version so that I can build lime-elements locally and have now updated the `lime-elements.api.md`. BTW a really nice way of handling the breaking changes and easily see how the public api looks like

jgroth (approved):

PR #2796, opened 2024-02-21
visualize different types of progress on `limel-chip`

Kiarokh:
This renders the chip with an indeterminate progress indicator, showing to the end user that something is being loaded or calculated.

fix https://github.com/Lundalogik/crm-feature/issues/3933

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [x] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/chip/chip.scss
+++ b/src/components/chip/chip.scss
@@ -182,3 +182,5 @@ limel-badge {
         }
     }
 }
+
+@import './partial-styles/_loading.scss';
--- a/src/components/chip/chip.scss
+++ b/src/components/chip/chip.scss
@@ -2,6 +2,7 @@
 
 /**
 * @prop --chip-max-width: Maximum width of the chip. Defaults to `10rem`. Keep in mind that the chips should not appear too big.
+* @prop --chip-progress-color: Color of the progress bar. Defaults to `rgb(var(--contrast-700))`.
 */
 
 :host(limel-chip) {
@@ -184,3 +185,4 @@ limel-badge {
 }
 
 @import './partial-styles/_loading.scss';
+@import './partial-styles/_progress.scss';
--- a/src/components/chip/chip.tsx
+++ b/src/components/chip/chip.tsx
@@ -73,6 +73,7 @@ interface ChipInterface extends Omit<OldChipInterface, 'id' | 'badge'> {
  * @exampleComponent limel-example-chip-badge
  * @exampleComponent limel-example-chip-filter
  * @exampleComponent limel-example-chip-removable
+ * @exampleComponent limel-example-chip-loading
  * @exampleComponent limel-example-chip-aria-role
  */
 @Component({
@@ -149,6 +150,14 @@ export class Chip implements ChipInterface {
     public type?: ChipType = 'default';
 
     /**
+     * Set to `true` to put the component in the `loading` state,
+     * and render an indeterminate progress indicator inside the chip.
+     * This does _not_ disable the interactivity of the chip!
+     */
+    @Prop({ reflect: true })
+    public loading? = false;
+
+    /**
      * Identifier for the chip. Must be unique.
      */
     @Prop({ reflect: true })
@@ -189,6 +198,7 @@ export class Chip implements ChipInterface {
                 disabled={this.disabled || this.readonly}
                 onKeyDown={this.handleDeleteKeyDown}
             >
+                {this.renderSpinner()}
                 {this.renderIcon()}
                 {this.renderLabel()}
                 {this.renderBadge()}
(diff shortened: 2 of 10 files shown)


src/components/chip/chip.scss
```diff
@@ -2,6 +2,7 @@
 
 /**
 * @prop --chip-max-width: Maximum width of the chip. Defaults to `10rem`. Keep in mind that the chips should not appear too big.
+* @prop --chip-progress-color: Color of the progress bar. Defaults to the color of the icon if specified, otherwise defaults to `rgb(var(--contrast-700))`.
```
Kiarokh, line 5: > "Defaults to the color of the icon if specified,"

this is probably overkill, right?

src/components/chip/chip.tsx
```diff
@@ -315,5 +327,31 @@ export class Chip implements ChipInterface {
...
+                role="progressbar"
+                aria-label="%"
+                aria-valuemin="0"
+                aria-valuemax="100"
+                aria-valuenow={this.progress}
+                style={{
+                    '--limel-chip-progress-percentage': currentPercentage,
+                }}
+            >
+                <span>{abbreviate(progress)}%</span>
```
adrianschmidt, line 353: What's `abbreviate` used for here?

src/components/chip/examples/chip-progress.tsx
```diff
@@ -0,0 +1,66 @@
+import { Component, h, State } from '@stencil/core';
+
+/**
+ * Displaying a progress bar
+ * By defining a numeric `progress` (between `0` to `100`),
```
adrianschmidt, line 5: "between … to …" really is a pet peeve of mine, both in English and in Swedish… 🙈

```suggestion
 * By defining a numeric `progress` (between `0` and `100`),
```

or 

```suggestion
 * By defining a numeric `progress` (from `0` to `100`),
```

src/components/chip/examples/chip-progress.tsx
```diff
@@ -0,0 +1,66 @@
...
+ * you can display a progress bar on the chip
+ * to inform the user about an ongoing progress and also
+ * visualize the amount of progress that has been made so far.
+ *
+ * :::important
+ * 1. Do not use `loading={true}` and `progress` properties at the same time.
```
adrianschmidt, line 11:
```suggestion
 * 1. Do not use `loading={true}` and `progress` at the same time.
```

src/components/chip/partial-styles/_progress.scss
```diff
@@ -0,0 +1,64 @@
...
+            var(--limel-chip-progressbar-stripe-color) 50%,
+            var(--limel-chip-progressbar-stripe-color) 75%,
+            transparent 75%,
+            transparent
+        );
+        z-index: 1;
+        background-size: var(--limel-chip-progressbar-stripe-size)
+            var(--limel-chip-progressbar-stripe-size);
+        animation: move 2.5s linear infinite;
+        border-radius: inherit;
```
adrianschmidt, line 50:
```suggestion
        border-radius: inherit;

        @media (prefers-reduced-motion) {
            animation: none;
        }
```

Kiarokh: <3

src/components/chip/chip.tsx
Kiarokh: I copied this from `limel-circular-progress`. I think sometimes we could get numbers like `12.23679` and wanted to convert them to 1 decimal this way. So the UI would display `12.2`

adrianschmidt: Ah, yeah, that might be it. Let me give it a quick try 👍

adrianschmidt: Even with `abbreviate` it can show one decimal, and without it, we still ensure it can never show more than one decimal by rounding the value on line 340:

image

If we want to avoid ever showing decimals here, we can just adjust line 340 like this:

```diff
-        const progress = Math.round(this.progress * 10) / 10; // eslint-disable-line no-magic-numbers
+        const progress = Math.round(this.progress);
```

Wait… now that I look at that line… 🤨 😄

adrianschmidt: Should we just go with `const progress = Math.round(this.progress);` on line 340 then?

Kiarokh: this works absolutely in this case, because we won't need to show any decimals at all actually.
So getting this UI 👇 which the previous code generates is actually useless in this context.
image
I'll add a fixup!

Kiarokh: What do you think about this? 
image

It also looks unnecessary. We expect the consumer to give us a number between 0 to 100 anyways, right?

Kiarokh: Added a fixup and simplified all of these.

adrianschmidt (approved):

Issue #2793, opened 2024-02-21
Not possible to click on document chip to open file


FredrikWallstrom:
## Current behavior

Not possible to click on document chip to open file, nothing happens.

1. Click on a document chip on a document object

## Expected behavior

File should be opened/downloaded

## Environment

- **lime-elements** version: 37.4.0
- Framework used: StencilJS
- Logs: Nothing

adrianschmidt:
I've found the problem, and I'm working on finding a solution 😊

Befkadu1:
> I've found the problem, and I'm working on finding a solution 😊

I told you, you are awesome 😄

PR #2798, opened 2024-02-22
fix(file): fix bug where links on chips could not be opened

adrianschmidt:
fix: #2793

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -197,7 +197,6 @@ export namespace Components {
         // @beta
         "type"?: ChipType;
     }
-    // (undocumented)
     export interface LimelChipSet {
         "clearAllButton": boolean;
         "delimiter": string;
@@ -1000,7 +999,6 @@ namespace JSX_2 {
         // @beta
         "type"?: ChipType;
     }
-    // (undocumented)
     interface LimelChipSet {
         "clearAllButton"?: boolean;
         "delimiter"?: string;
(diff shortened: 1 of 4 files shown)


etc/lime-elements.api.md
```diff
@@ -1000,7 +999,6 @@ namespace JSX_2 {
         // @beta
         "type"?: ChipType;
     }
-    // (undocumented)
     interface LimelChipSet {
```
adrianschmidt, line 1002: @jgroth I have no idea why api-extractor suddenly thinks this is any less documented than it was before… 🤷‍♂️

The change means I'll need an approval from you though 😅

src/components/chip-set/chip-set.e2e.ts
```diff
@@ -291,6 +293,16 @@ describe('limel-chip-set', () => {
                     removable: true,
                 });
             });
+
+            // This test doesn't work at the moment, and I can't run the e2e
+            // tests locally to debug it. So I'm skipping it for now. /Ads
+            it.skip('attaches the chip to the click event', () => {
```
adrianschmidt, line 299: Oh ffs… it seems like this isn't possible to test, because the "event" object you get from the spy isn't actually the real event object which we added the `Lime` property to… 😣

adrianschmidt: Got an idea, and it worked! Pushing a fixup in a moment!

etc/lime-elements.api.md
adrianschmidt: The rest of the PR has been approved by Per, so I'm going to use my admin privileges to bypass the need for another maintainer than myself to review 😊

PR #2801, opened 2024-02-24
Feat/form generic schema

TommyLindh2:
Make the FormSchema generic and support more specific typing
Related to #2795 

Fix [Lundalogik/limepkg-basic-helpdesk#19](https://github.com/Lundalogik/limepkg-basic-helpdesk/issues/19)

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -772,8 +772,34 @@ export enum FormLayoutType {
 }
 
 // @public
-export interface FormSchema extends JSONSchema7 {
+export interface FormSchema<T extends Record<string, any> = any> extends JSONSchema7 {
+    $defs?: Record<string, FormSchema>;
+    additionalItems?: FormSchemaArrayItem<T>;
+    additionalProperties?: FormSchema | boolean;
+    allOf?: Array<FormSchemaArrayItem<T>>;
+    anyOf?: Array<FormSchemaArrayItem<T>>;
+    contains?: FormSchemaArrayItem<T>;
+    definitions?: Record<string, FormSchema>;
+    dependencies?: Record<string, FormSchema | string[]>;
+    else?: FormSchema;
+    if?: FormSchema;
+    items?: FormSchemaArrayItem<T> | Array<FormSchemaArrayItem<T>>;
+    not?: FormSchema;
+    oneOf?: Array<FormSchemaArrayItem<T>>;
+    patternProperties?: Record<string, FormSchema>;
+    properties?: ReplaceObjectType<T, FormSubKeySchema<T>, Record<string, FormSchema>>;
+    propertyNames?: FormSchema;
+    required?: Array<ReplaceObjectType<T, Extract<keyof T, string>, string>>;
+    then?: FormSchema;
 }
+
+// @public
+export type FormSchemaArrayItem<T> = T extends any[] ? FormSchema<T[Extract<keyof T, number>]> : FormSchema;
+
+// @public
+export type FormSubKeySchema<TObj> = Partial<{
+    [Key in Extract<keyof TObj, any>]: FormSchema<TObj[Key]>;
+}>;
 
 // @public
 export interface GridLayoutOptions extends FormLayoutOptions<FormLayoutType | `${FormLayoutType}`> {
@@ -1971,6 +1997,9 @@ interface Option_2<T extends string = string> {
     value: T;
 }
 export { Option_2 as Option }
+
+// @public
+export type ReplaceObjectType<T, AllowedType, ElseType> = T extends any[] ? ElseType : T extends Record<string, any> ? AllowedType : ElseType;
 
 // @public
 export interface RowLayoutOptions extends FormLayoutOptions<FormLayoutType | `${FormLayoutType}`> {
--- a/src/components/button-group/examples/button-group-composite.tsx
+++ b/src/components/button-group/examples/button-group-composite.tsx
@@ -1,5 +1,9 @@
 import { Component, h, Prop, State } from '@stencil/core';
-import { Button, LimelButtonGroupCustomEvent } from '@limetech/lime-elements';
+import {
+    Button,
+    FormSchema,
+    LimelButtonGroupCustomEvent,
+} from '@limetech/lime-elements';
 
 /**
  * Composite
@@ -12,7 +16,7 @@ import { Button, LimelButtonGroupCustomEvent } from '@limetech/lime-elements';
 })
 export class ButtonCompositeExample {
     @Prop()
-    public schema: any;
+    public schema: FormSchema;
 
     @State()
     private props = {
--- a/src/components/form/examples/help-form-schema.ts
+++ b/src/components/form/examples/help-form-schema.ts
@@ -1,6 +1,16 @@
 import { FormSchema } from '@limetech/lime-elements';
 
-export const schema: FormSchema = {
+type Race = 'search' | 'recommendation' | 'ads' | 'work' | 'workshop';
+export interface HelpFormData {
+    address?: {
+        planet?: string;
+        galaxy?: string;
+        email?: string;
+        race?: Race;
+    };
+}
+
+export const schema: FormSchema<HelpFormData> = {
     type: 'object',
     properties: {
         address: {
--- a/src/components/form/examples/help-form-schema.ts
+++ b/src/components/form/examples/help-form-schema.ts
@@ -1,12 +1,12 @@
 import { FormSchema } from '@limetech/lime-elements';
 
-type Race = 'search' | 'recommendation' | 'ads' | 'work' | 'workshop';
+type Source = 'search' | 'recommendation' | 'ads' | 'work' | 'workshop';
 export interface HelpFormData {
     address?: {
-        planet?: string;
-        galaxy?: string;
+        name?: string;
+        company?: string;
         email?: string;
-        race?: Race;
+        source?: Source;
     };
 }
 
@@ -17,7 +17,7 @@ export const schema: FormSchema<HelpFormData> = {
             type: 'object',
             title: 'Book a demo',
             properties: {
-                planet: {
+                name: {
                     type: 'string',
                     title: 'Name',
                     description: 'Please write your name & surname',
@@ -33,7 +33,7 @@ export const schema: FormSchema<HelpFormData> = {
                         },
                     },
                 },
-                galaxy: {
+                company: {
                     type: 'string',
                     title: 'Company',
                     description: 'Please write your company name',
@@ -54,7 +54,7 @@ export const schema: FormSchema<HelpFormData> = {
                         },
                     },
                 },
-                race: {
+                source: {
                     type: 'string',
                     title: 'How did you hear about Lime CRM?',
                     oneOf: [
--- a/src/components/form/examples/span-fields-schema.ts
+++ b/src/components/form/examples/span-fields-schema.ts
@@ -1,16 +1,6 @@
 import { FormSchema } from '@limetech/lime-elements';
 
-export interface SpanFieldsFormData {
-    field1?: string;
-    field2?: number;
-    field3?: string;
-    field4?: string;
-    field5?: number;
-    field6?: string;
-    field7?: number;
-}
-
-export const schema: FormSchema<SpanFieldsFormData> = {
+export const schema: FormSchema = {
     title: 'A form with fields that span columns and rows',
     description:
         'This main form has a grid layout with 5 columns. Notice how fields reorder to fill holes when the "Dense layout" is enabled. You may need to resize your browser window to see this responsive layout in effect.',
--- a/src/components/form/examples/span-fields-schema.ts
+++ b/src/components/form/examples/span-fields-schema.ts
@@ -1,6 +1,16 @@
 import { FormSchema } from '@limetech/lime-elements';
 
-export const schema: FormSchema = {
+export interface SpanFieldsFormData {
+    field1?: string;
+    field2?: number;
+    field3?: string;
+    field4?: string;
+    field5?: number;
+    field6?: string;
+    field7?: number;
+}
+
+export const schema: FormSchema<SpanFieldsFormData> = {
     title: 'A form with fields that span columns and rows',
     description:
         'This main form has a grid layout with 5 columns. Notice how fields reorder to fill holes when the "Dense layout" is enabled. You may need to resize your browser window to see this responsive layout in effect.',
--- a/src/components/form/form.types.ts
+++ b/src/components/form/form.types.ts
@@ -323,4 +323,367 @@ export enum FormLayoutType {
  * @public
  * Represents the JSON schema with Lime specific options
  */
-export interface FormSchema extends JSONSchema7 {}
+export interface FormSchema<T extends Record<string, any> = any>
+    extends JSONSchema7 {
+    /**
+     * The value of "items" MUST be either a valid JSON Schema or an array
+     * of valid JSON Schemas.
+     *
+     * This keyword determines how child instances validate for arrays, and
+     * does not directly validate the immediate instance itself.
+     *
+     * If "items" is a schema, validation succeeds if all elements in the
+     * array successfully validate against that schema.
+     *
+     * If "items" is an array of schemas, validation succeeds if each
+     * element of the instance validates against the schema at the same
+     * position, if any.
+     *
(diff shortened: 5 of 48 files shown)


src/components/button-group/examples/button-group-composite.tsx
adrianschmidt, line 1: I _love_ that you've made the effort to go through and update the examples! For the sake of easier overview of what's what, I'd like to separate the updates of the examples for other components than limel-form into a separate commit though. So I'll split this commit up in two, if that's OK with you 😊

TommyLindh2: Yeah, that makes sense, do you want me to do it?

adrianschmidt: No no, you shouldn't be saddled with more work because you stepped up and did extra to begin with 😊

I just pushed it 👍

etc/lime-elements.api.md
```diff
@@ -755,8 +755,101 @@ export enum FormLayoutType {
 }
 
 // @public
-export interface FormSchema extends JSONSchema7 {
-}
+export interface FormSchema<T extends Record<string, any> = any> extends JSONSchema7 {
+    // Warning: (ae-unresolved-inheritdoc-base) The @inheritDoc tag needs a TSDoc declaration reference; signature matching is not supported yet
```
jgroth, line 759: There seems to be a lot of warnings generated here :thinking:

src/interface.ts
```diff
@@ -11,7 +11,20 @@ export * from './components/dialog/dialog.types';
...
+    FormLayoutOptions,
+    FormLayoutType,
+    FormSchemaArrayItem,
+    FormSubKeySchema,
+    GridLayoutOptions,
+    RowLayoutOptions,
+    LimeLayoutOptions,
+    LimeSchemaOptions,
+    ReplaceObjectType,
+} from './components/form/form.types';
```
jgroth, line 27: Is there something from here we don't want to export? I could be easy to forget to add new types to this list in the future. Perhaps the type we don't want to export could be moved to another file?

TommyLindh2: Yeah, Maybe we can skip ALL the utility types, that is used just for the generic stuff.

etc/lime-elements.api.md
TommyLindh2: It looks like this is due to using inheritDoc on properties that doesn't have docs.
Maybe that can be fixed, by looking what properties that have docstrings. Either we add our own OR we just remove the inheritdoc tag.

As I cannot run the api-extractor locally maybe you can do this for me @jgroth?

src/interface.ts
TommyLindh2: I have fixed this in a fixup commit 👍 
But as before I cannot run the api-update so you have to do it for me 👎

etc/lime-elements.api.md
TommyLindh2: I have fixed this in a fixup commit 👍 
But as before I cannot run the api-update so you have to do it for me 👎

jgroth: Yeah I can run it :smile: 
Perhaps we should move the logics of this `.sh` file to a `.js` file instead :smile:

adrianschmidt: If you rebase on latest `main`, the `.sh` file is no more 😉

etc/lime-elements.api.md
```diff
@@ -755,7 +755,49 @@ export enum FormLayoutType {
...
+    // (undocumented)
+    not?: FormSchema;
+    // (undocumented)
+    oneOf?: Array<FormSchemaArrayItem<T>>;
+    // (undocumented)
+    patternProperties?: Record<string, FormSchema>;
+    // Warning: (ae-forgotten-export) The symbol "FormSubKeySchema" needs to be exported by the entry point index.d.ts
+    //
+    // (undocumented)
+    properties?: ReplaceObjectType<T, FormSubKeySchema<T>, Record<string, FormSchema>>;
```
adrianschmidt, line 792: Same thing here.

etc/lime-elements.api.md
```diff
@@ -755,7 +755,49 @@ export enum FormLayoutType {
...
+    // (undocumented)
+    propertyNames?: FormSchema;
+    // Warning: (ae-forgotten-export) The symbol "ReplaceObjectType" needs to be exported by the entry point index.d.ts
+    //
+    // (undocumented)
+    required?: Array<ReplaceObjectType<T, Extract<keyof T, string>, string>>;
```
adrianschmidt, line 798: And here.

src/interface.ts
adrianschmidt: Those "utility" types are part of public types, so they need to be public too. So I've moved them back into the `form.types.ts` and changed the export here to just export everything from it.

etc/lime-elements.api.md
adrianschmidt: ⚡

etc/lime-elements.api.md
adrianschmidt: ⚡

src/components/form/form.types.ts
```diff
@@ -320,7 +320,128 @@ export enum FormLayoutType {
 }
 
 /**
- * @public
  * Represents the JSON schema with Lime specific options
+ * @public
+ */
+export interface FormSchema<T extends Record<string, any> = any>
```
jgroth, line 326: Is this correct or should `T` extend `JSONSchema7Definition` instead? :thinking:

src/components/form/widgets/input-field.ts
```diff
@@ -71,19 +72,19 @@ function getInputType(schema: any): InputType {
...
-function getStepSize(schema: any): 'any' | number {
+function getStepSize(schema: FormSchema): 'any' | number {
     if (isNumberType(schema) && schema.multipleOf) {
-        return parseFloat(schema.multipleOf) || 'any';
+        return parseFloat(schema.multipleOf as unknown as string) || 'any';
     }
 
     if (isIntegerType(schema)) {
-        return parseInt(schema.multipleOf, 10) || 1;
+        return parseInt(schema.multipleOf as unknown as string, 10) || 1;
```
jgroth, line 81: We could get rid of these ugly typecasts by doing something like `String(schema.multipleOf)`

jgroth: Or checking if it is actually a string or not before trying to convert it :slightly_smiling_face:

src/components/form/form.types.ts
TommyLindh2: In this case T is not the Schema itself, but the object the schema represents.
So for example if you have the interface:

```typescript
interface Person {
    name: string;
    role: string;
}
```

then you can get a `FormSchema` based on the Person by `FormSchema<Person>`.
Then you get required and properties and such types with `keyof Person` and that kind of things

src/components/form/widgets/input-field.ts
TommyLindh2: Fixed in fixup

src/interface.ts
TommyLindh2: Thanks, nice!

etc/lime-elements.api.md
TommyLindh2: Awesome!

etc/lime-elements.api.md
TommyLindh2: Awesome!

etc/lime-elements.api.md
```diff
@@ -755,7 +755,49 @@ export enum FormLayoutType {
...
+    // (undocumented)
+    dependencies?: Record<string, FormSchema | string[]>;
+    // (undocumented)
+    else?: FormSchema;
+    // (undocumented)
+    if?: FormSchema;
+    // Warning: (ae-forgotten-export) The symbol "FormSchemaArrayItem" needs to be exported by the entry point index.d.ts
+    //
+    // (undocumented)
+    items?: FormSchemaArrayItem<T> | Array<FormSchemaArrayItem<T>>;
```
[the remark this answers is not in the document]
TommyLindh2: Awesome

etc/lime-elements.api.md
TommyLindh2: Trying it out as we speak 👍

TommyLindh2: It looks like it runs perfectly for me now ❤️

src/components/form/widgets/input-field.ts
```diff
@@ -71,19 +72,19 @@ function getInputType(schema: any): InputType {
...
-function getStepSize(schema: any): 'any' | number {
+function getStepSize(schema: FormSchema): 'any' | number {
     if (isNumberType(schema) && schema.multipleOf) {
-        return parseFloat(schema.multipleOf) || 'any';
+        return parseFloat(String(schema.multipleOf)) || 'any';
     }
 
     if (isIntegerType(schema)) {
-        return parseInt(schema.multipleOf, 10) || 1;
+        return parseInt(String(schema.multipleOf), 10) || 1;
```
adrianschmidt, line 81: An alternative to `parseInt` and `parseFloat`, which I _usually_ recommend, is `+` (apparently called "unary plus" in this case). Here's a good explanation of the differences, copied from https://stackoverflow.com/a/17106701/280972:

---

Well, here are a few differences I know of:

- An empty string `""` evaluates to a `0`, while `parseInt` evaluates it to `NaN`. IMO, a blank string should be a `NaN`.

        +'' === 0;              //true
        isNaN(parseInt('',10)); //true

- The unary `+` acts more like `parseFloat` since it also accepts decimals. 

    `parseInt` on the other hand stops parsing when it sees a non-numerical character, like the period that is intended to be a decimal point `.`.

        +'2.3' === 2.3;           //true
        parseInt('2.3',10) === 2; //true

- `parseInt` and `parseFloat` parses and builds the string *left to right*. If they see an invalid character, it returns what has been parsed (if any) as a number, and `NaN` if none was parsed as a number. 

    The unary `+` on the other hand will return `NaN` if the entire string is non-convertible to a number.

        parseInt('2a',10) === 2; //true
        parseFloat('2a') === 2;  //true
        isNaN(+'2a');            //true

- As seen in the comment of [@Alex K.](https://stackoverflow.com/users/246342), `parseInt` and `parseFloat` will parse by character. This means hex and exponent notations will fail since the `x` and `e` are treated as non-numerical components (at least on base10).

    The unary `+` will convert them properly though.

        parseInt('2e3',10) === 2;  //true. This is supposed to be 2000
        +'2e3' === 2000;           //true. This one's correct.

        parseInt("0xf", 10) === 0; //true. This is supposed to be 15
        +'0xf' === 15;             //true. This one's correct.

---

I don't know what behaviour is the more appropriate for this use case, but generally, I would say that `+` seems to do the more correct thing in all cases except when the input is an empty string, and that case could easily be tested for first, and be made to return `NaN` instead.

For the `isIntegerType` case, we might also want to use `Math.floor` or `Math.round`…

adrianschmidt: Oh, and, as an added bonus, you don't need to convert the value to a string before using unary plus 😊

And I realised I didn't show how it's used:

```suggestion
        return +schema.multipleOf || 1;
```

And I just now realised that we don't need to treat an empty string differently, as `0` is falsy, so it will return `1` instead 😄

src/components/form/widgets/input-field.ts
```diff
@@ -71,19 +72,19 @@ function getInputType(schema: any): InputType {
     return 'text';
 }
 
-function getStepSize(schema: any): 'any' | number {
+function getStepSize(schema: FormSchema): 'any' | number {
     if (isNumberType(schema) && schema.multipleOf) {
-        return parseFloat(schema.multipleOf) || 'any';
+        return parseFloat(String(schema.multipleOf)) || 'any';
```
adrianschmidt, line 77:
```suggestion
        return +schema.multipleOf || 'any';
```

Should this really return the _string_ `'any'`? 🤔

jgroth: Yes, the `step` property of `limel-input-field` is `"any" | number`

jgroth (approved):

src/components/form/widgets/input-field.ts
TommyLindh2: Fixed, did it on the `parseFloat` in a few lines above as well

TommyLindh2:
@adrianschmidt do you also approve this?
Johans approval disappeared after I used autosquash and fixed your suggestions

src/components/form/examples/help-form-schema.ts
```diff
@@ -1,6 +1,16 @@
 import { FormSchema } from '@limetech/lime-elements';
 
-export const schema: FormSchema = {
+type Race = 'search' | 'recommendation' | 'ads' | 'work' | 'workshop';
```
adrianschmidt, line 3: This raised my eyebrow a bit… 😆

I had to track it down, and this was clearly a copy-and-paste error made [here](https://github.com/Lundalogik/lime-elements/pull/2622/files#diff-ffd4c8bf682a861e2eec3afe1191504fa1625a4e6c5638b5414857a981617f3dR45). However, this PR makes the mistake a whole lot more visible, so it would be nice to correct it now.

TommyLindh2: Nice find 👍 I just continued with the copy/paste 😂
But I can fix it. I will just do a force push

TommyLindh2: Is now fixed

src/components/form/examples/span-fields-schema.ts
```diff
@@ -1,6 +1,16 @@
...
+    field1?: string;
+    field2?: number;
+    field3?: string;
+    field4?: string;
+    field5?: number;
+    field6?: string;
+    field7?: number;
+}
+
+export const schema: FormSchema<SpanFieldsFormData> = {
```
adrianschmidt, line 13: Is the typing required now? If it's not, I'm thinking we should perhaps just leave this un-typed… it brings so little value anyway, that it would kind of bring more value in showing that it's possible to not use such strict typing… 🤷‍♂️ I dunno…

TommyLindh2: I agree that the Generic typing don't bring that much value, even though I think it DOES bring value.
But the main issue now is that the current typing makes it really hard to work with and you need to cast a lot of stuff to get it to build. So if you don't want the generic stuff we should still fix the other stuff. i.e. sub schemas are FormSchema instead of the FormSchema | boolean.

But don't know why you want to wait? We already have merged the typing once.

TommyLindh2: Yes, I agree. We can remove it. I fix

src/components/form/form.types.ts
```diff
@@ -320,7 +320,128 @@ export enum FormLayoutType {
...
+
+    /**
+     * @see https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-6.7.3
+     */
+    oneOf?: Array<FormSchemaArrayItem<T>>;
+
+    /**
+     * @see https://datatracker.ietf.org/doc/html/draft-handrews-json-schema-validation-01#section-9
+     */
+    $defs?: Record<string, FormSchema>;
```
adrianschmidt, line 375: `$defs` and `definitions` (on line 420) link to the same section (where `definitions` is mentioned, but `$defs` is not). Is this correct? Do we need to keep both? If we need to keep both:

- Should we mention that they are the same thing, or what the differences are?
- We could perhaps place them next to each other, so it becomes a little easier to discover that there are two of them… 🙂

adrianschmidt: Oh, don't do anything right now. I'm making changes here. Let me push first! 😅

TommyLindh2: Nice, fixed. Not sure if it should be clarified even more or not.

adrianschmidt: Looks good 👍

I don't think it needs further clarification. I guess it's whether `definitions` should be marked as deprecated perhaps? 🤷‍♂️

adrianschmidt (approved):

PR #2802, opened 2024-02-25
Add a `file-dropzone` component

civing:
Adds a new `<limel-file-dropzone>` component. It can wrap any other components.

The `<limel-file-dropzone>` component will wrap all child components and create a file drop zone that accepts files. The `limel-file-input` component works in a similar way but will instead let the user select files using the native file browser dialog.

The file types to be dropable can be controlled with the `accept` prop. It should conform to the HTML attribute accept format https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept

The component emits a `filesSelected` event upon selection of files.
If any files does not conform with the optional `accept` property a `filesRejected` even will be emitted.

Following is example usage of the component.  A user can select files both by dragging them into the `<limel-file-dropzone>`

```tsx
<limel-file-dropzone onFilesSelected={files: FileInfo[] => console.log(files)}>
    <my-component />
 </limel-file-dropzone>
```

Fixes https://github.com/Lundalogik/crm-feature/issues/3941
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/util/file-metadata.ts
```diff
@@ -1,12 +1,12 @@
-import { FileInfo } from './file.types';
-import { getIconForFile } from './icons';
-import { getIconFillColorForFile } from './icon-fill-colors';
-import { getIconBackgroundColorForFile } from './icon-background-colors';
+import { FileInfo } from 'src/components/file/file.types';
```
adrianschmidt: Huh… is this path relative to the package root? I've never seen that before… 😮

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,123 @@
...
+    Component,
+    h,
+    Event,
+    EventEmitter,
+    Host,
+    Element,
+    Prop,
+} from '@stencil/core';
+import { FileInfo } from 'src/interface';
+import { createRandomString } from 'src/util/random-string';
```
Kiarokh, line 11: are we gonna use this `createRandomString`, or use the `crypto.randomUUID()` instead?

https://github.com/Lundalogik/lime-elements/blob/9ec29350164b3d7dfa1c03de6bae386a9d8d6c7e/src/components/chip/chip.tsx#L173

src/util/file-metadata.ts
civing: I hadn't pushed the latest, sorry 🙈 
The latest should be updated with correct imports.

src/components/file-input/file-input.tsx
```diff
@@ -0,0 +1,110 @@
...
+            >
+                <input
+                    hidden={true}
+                    id={this.fileInputId}
+                    onChange={this.handleFileChange}
+                    type="file"
+                    accept={this.accept}
+                    disabled={this.disabled}
+                />
+                <slot></slot>
```
Kiarokh, line 58:
```suggestion
                <slot />
```

src/components/file-dropzone/file-dropzone.scss
Kiarokh, line 1: there are some styles in `_file-picker.scss` for the old dropzone which should be removed as well.

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,92 @@
...
+ * After receiving the files, the component emits a `newFiles` event.
+ * The event detail would be an array of `FileInfo` objects,
+ * each representing a file dropped into the dropzone.
+ * @exampleComponent limel-example-file-dropzone
+ * @exampleComponent limel-example-file-dropzone-type-filtering
+ */
+@Component({
+    tag: 'limel-file-dropzone',
+    shadow: true,
+    styleUrl: 'file-dropzone.scss',
```
adrianschmidt, line 21: I'm guessing this property and the empty file it points to are remnants of earlier iterations? 🙂

adrianschmidt:
I squashed the fixups. Sorry for not asking first. I wanted to make it as simple as possible for myself to fix the issues that api-extractor was failing on.

src/components/file-dropzone/file-dropzone.tsx
civing: Nah it's going to be filled in eventually (hopefully tomorrow) :)

civing:
> I squashed the fixups. Sorry for not asking first. I wanted to make it as simple as possible for myself to fix the issues that api-extractor was failing on.

No worries!

This thing grew into a monster PR. I plan to split it into three PR's.
1. New file dropzone component
2. New file input component
3. Refactor of file componen

src/components/file-dropzone/file-dropzone.tsx
john-traas: The createRandomString function uses crypto.randomUUID() but also has a fallback in case the browser being used doesn't support crypto or have it available.

adrianschmidt: I was about to say that we don't need to support browsers that don't support `crypto.randomUUID()`, but thought I'd look at caniuse.com first, just to be sure, and it looks like `crypto.randomUUID()` only got support in Chrome and Firefox on Android during February _this year_ 😲

image

https://caniuse.com/?search=crypto.randomUUID

Although, I wonder if the lack of red blobs for earlier versions doesn't just mean they don't have data for those versions (though that would also surprise me… so I don't really know what to think… 🤷‍♂️)

We were using `createRandomString()` in an example, and I [replaced](https://github.com/Lundalogik/lime-elements/commit/b88977369b91696d1c1649293a03c35178fe85de) that with `crypto.randomUUID()` just this last Friday, because we don't export `createRandomString`, so it shouldn't be used in examples. (In this specific example, the value was used for the `key` attribute, so I didn't have to prepend the UUID with a letter, but when used with the `id` attribute, remember that IDs must begin with a letter, not a number, so the UUID would need to be prepended, like it is in `createRandomString`.)

I'll see if BrowserStack has older Chrome versions on Android that I can use to check…

john-traas: I've added `createRandomString() `as an exported function to a` utils.ts` in` lime-crm-components` in the PR I'm working on related to the new FileUpload class but maybe it should instead be in `lime-web-components` if we'll possibly need it elsewhere?

adrianschmidt: BrowserStack are running Chrome 120 on all the Android devices I tested just now, and `crypto.randomUUID()` works on all of them. I think we should just stop using `createRandomString`, rather than spread it to more places.

john-traas: I'll remove it from my PR in that case. I think actually now that I'm cleaning it up, I'm no longer using it anyways as we went a different direction for handling uploads.

civing: I agree with you all, please see https://github.com/Lundalogik/lime-elements/pull/2802/commits/aba85d16f2e3b4bd43bdc25e988c790914711e1a

src/util/files.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+        expect(result.id).toBeDefined();
+
+        const { id: h, ...resultWithoutId } = result;
+
+        expect(resultWithoutId).toMatchObject(expectedFileInfo);
+    });
+});
+
+describe('isTypeAccepted', () => {
+    const testCases = [
```
LawrenceBorst, line 33: NIT. We could use `test.each` as well here. It may be more standard

LawrenceBorst: NIT. Test have full coverage line coverage, but there are only four code paths to really consider, so I wonder if it's enough to write only ~4 test cases. I think if we factor these as separate tests we can get more free documentation, like "When accept is `undefined` or `'*'`".

src/util/files.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function createFileInfo(file: File): FileInfo {
+    let limeFile: FileInfo = {
+        id: createRandomString(),
+        filename: file.name,
+        contentType: file.type,
+        size: file.size,
+        fileContent: file,
+    };
+    limeFile = {
```
LawrenceBorst, line 17: I suppose this is one way, but I think we can just do `limeFile.icon = ...`

etc/lime-elements.api.md
```diff
@@ -322,6 +322,12 @@ export namespace Components {
         "required": boolean;
         "value": FileInfo;
     }
+    export interface LimelFileDropzone {
```
LawrenceBorst, line 325: I wonder if we will get declaration merging here and export the event handlers, because this interface is now defined twice in this file 🤔. Worth double checking

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,162 @@
...
+import { createFileInfo, isTypeAccepted } from '../../util/files';
+
+/**
+ * This component enables you to seamlessly convert any region of the user interface into
+ * a file dropzone area, just by wrapping it inside the `limel-file-dropzone`.
+ *
+ * The file dropzone can then be used to allow end-users to upload files
+ * by dragging and dropping them into the specified area.
+ *
+ * After receiving the files, the component emits a `filesSelecgted` event. For unsupported
```
LawrenceBorst, line 20: NIT. Small typo on this line.

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,162 @@
...
+                        name={this.onDropLayoutIcon}
+                    ></limel-icon>
+                    {this.renderOnDragLayoutText()}
+                </div>
+            );
+        }
+    };
+
+    private renderOnDragLayoutText = () => {
+        if (this.onDragLayoutText.length === 0) {
```
LawrenceBorst, line 118: Should we instead make the type of this prop `string | undefined`? It seems like `undefined` or `null` are more transparent "magic values" than `''`.

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,162 @@
...
+        this.hasFileToDrop = false;
+
+        if (this.disabled) {
+            return;
+        }
+
+        const files: File[] = Array.from(event.dataTransfer.files);
+        const fileInfos: FileInfo[] = files.map(createFileInfo);
+
+        const acceptedFileInfos: FileInfo[] = fileInfos.filter((file) =>
```
LawrenceBorst, line 137: Maybe try out Lodash's `partition` function for 137-142, it's perfect :)

LawrenceBorst (commented):
Some thoughts

src/components/file-dropzone/examples/file-dropzone-type-filtering.scss
```diff
@@ -0,0 +1,18 @@
...
+
+.video-picture-dropzone {
+    min-width: 3rem;
+    min-height: 10rem;
+    background-color: rgb(var(--contrast-600));
+    border-radius: 0.5rem;
+    border: 1px solid rgb(var(--contrast-800));
+    padding: 0.25rem;
+    margin-bottom: 0.5rem;
+}
```
Kiarokh, line 18: [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) 😊

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    private renderOnDragLayout = () => {
+        if (this.hasFileToDrop) {
+            return (
+                <div class="has-file-to-drop">
+                    <limel-icon
+                        class="dropzone-icon"
+                        name={this.onDropLayoutIcon}
+                    ></limel-icon>
```
Kiarokh, line 110:
```suggestion
                    />
```

aj aj aj

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,162 @@
...
+     * Set to `true` to disable the file dropzone.
+     */
+    @Prop()
+    public disabled: boolean = false;
+
+    /**
+     * Text to be displayed when the user is dragging a file over the dropzone.
+     */
+    @Prop()
+    public onDragLayoutText: string = '';
```
Kiarokh, line 60: this is a really strange name for a public prop 😅

```suggestion
    public label: string = '';
```

or 

```suggestion
    public text: string = '';
```

Kiarokh: In most of our components we use these two prop names. I prefer `text` myself. Because we wanna add another prop, preferably called `helperText`, which is something to display below this text.

image

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,162 @@
...
+} from '@stencil/core';
+import { FileInfo } from '../../global/shared-types/file.types';
+import { createFileInfo, isTypeAccepted } from '../../util/files';
+
+/**
+ * This component enables you to seamlessly convert any region of the user interface into
+ * a file dropzone area, just by wrapping it inside the `limel-file-dropzone`.
+ *
+ * The file dropzone can then be used to allow end-users to upload files
+ * by dragging and dropping them into the specified area.
```
Kiarokh, line 18:
```suggestion
 * by dragging and dropping them into the specified area, for example to trigger an upload process.
```

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,162 @@
...
+    public accept: string = '*';
+
+    /**
+     * Set to `true` to disable the file dropzone.
+     */
+    @Prop()
+    public disabled: boolean = false;
+
+    /**
+     * Text to be displayed when the user is dragging a file over the dropzone.
```
Kiarokh, line 57:
```suggestion
     * Is displayed when the user is dragging files over the dropzone.
     * A suitable text could for instance be "Drop your files here".
```

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,162 @@
...
+     */
+    @Prop()
+    public onDragLayoutText: string = '';
+
```
Kiarokh, line 61:
```suggestion
    /**
     * Is displayed to provide supplementary information to the end users,
     * for instance, which filetypes or file sizes are accepted.
     */
    @Prop()
    public helperText?: string = '';

```

src/components/file-dropzone/file-dropzone.tsx
Kiarokh, line 20:
```suggestion
 * After receiving the files, the component emits a `filesSelected` event. For unsupported
```

Kiarokh: I highly recommend everyone to install `streetsidesoftware.code-spell-checker`: https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,162 @@
...
+                        class="dropzone-icon"
+                        name={this.onDropLayoutIcon}
+                    ></limel-icon>
+                    {this.renderOnDragLayoutText()}
+                </div>
+            );
+        }
+    };
+
+    private renderOnDragLayoutText = () => {
```
Kiarokh, line 117:
```suggestion
    private renderText = () => {
```

src/components/file-dropzone/file-dropzone.tsx
Kiarokh, line 118:
```suggestion
        if (!this.onDragLayoutText) {
```

Kiarokh: of course this prop should be renamed as I suggested above. But you get the point

src/util/files.ts
civing: I like your way more. See https://github.com/Lundalogik/lime-elements/pull/2802/commits/892fe1625cd4620774e37ce7f70bdef11d3d9a24

etc/lime-elements.api.md
civing: That file is automatically generated by api-extractor. @adrianschmidt do you have any input to Lawrence concern?

src/components/file-dropzone/file-dropzone.tsx
civing: See https://github.com/Lundalogik/lime-elements/pull/2802/commits/40fd1dd4b68f8f2a2edc178ef85e4659cdc61c5d

src/components/file-dropzone/examples/file-dropzone.tsx
```diff
@@ -0,0 +1,69 @@
...
+    private readonly = false;
+
+    @State()
+    private files: FileInfo[] = [];
+
+    public render() {
+        return [
+            <limel-file-dropzone
+                onFilesSelected={this.handleDrop}
+                disabled={this.disabled || this.readonly}
```
Kiarokh, line 26: you have to write all of the required props in the first example. So the `text` prop or whatever we end up calling it should be here, saying "Drop your files here"

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,162 @@
...
+            </Host>
+        );
+    }
+
+    private renderOnDragLayout = () => {
+        if (this.hasFileToDrop) {
+            return (
+                <div class="has-file-to-drop">
+                    <limel-icon
+                        class="dropzone-icon"
```
Kiarokh, line 108:
```suggestion
                        class="icon"
```

The CSS is already scoped and the internal elements are isolated. There is no need for such specificity in web components.

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,162 @@
...
+            );
+        }
+    };
+
+    private renderOnDragLayoutText = () => {
+        if (this.onDragLayoutText.length === 0) {
+            return;
+        }
+
+        return <span class="dropzone-text">{this.onDragLayoutText}</span>;
```
Kiarokh, line 122:
```suggestion
        return <span class="text">{this.onDragLayoutText}</span>;
```

The CSS is already scoped and the internal elements are isolated. There is no need for such specificity in web components.

src/components/file-dropzone/file-dropzone.tsx
civing: Sure! Nice suggestion, see https://github.com/Lundalogik/lime-elements/pull/2802/commits/40fd1dd4b68f8f2a2edc178ef85e4659cdc61c5d

src/components/file-dropzone/file-dropzone.tsx
civing: I love it 😍
See https://github.com/Lundalogik/lime-elements/pull/2802/commits/05f80575ff2eec6620e0ea5ad1a9f7e5d0af1305

src/components/file-dropzone/examples/file-dropzone-type-filtering.scss
LucyChyzhova: haha, right, did it in a hurry

src/components/file-dropzone/file-dropzone.tsx
LucyChyzhova: fixed

src/components/file-dropzone/file-dropzone.tsx
LucyChyzhova: I left one icon 😢, and  will change it when we have a better one

src/components/file-dropzone/examples/file-dropzone.tsx
LucyChyzhova: The text is not a required prop.

src/components/file-dropzone/examples/file-dropzone.tsx
```diff
@@ -0,0 +1,69 @@
...
+    };
+
+    private setDisabled = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.disabled = !!event.detail;
+    };
+
+    private setReadonly = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.readonly = !!event.detail;
```
LawrenceBorst, line 67:
```suggestion
        this.readonly = event.detail;
```

src/components/file-dropzone/examples/file-dropzone.tsx
```diff
@@ -0,0 +1,69 @@
...
+        ];
+    }
+
+    private handleDrop = (event: CustomEvent<FileInfo[]>) => {
+        this.files = [...this.files.concat(event.detail)];
+    };
+
+    private setDisabled = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.disabled = !!event.detail;
```
LawrenceBorst, line 62:
```suggestion
        this.disabled = event.detail;
```

src/components/file-dropzone/file-dropzone.tsx
```diff
@@ -0,0 +1,169 @@
...
+                onDragLeave={this.handleDragLeave}
+            >
+                <slot />
+                {this.renderOnDragLayout()}
+            </Host>
+        );
+    }
+
+    private renderOnDragLayout = () => {
+        if (this.hasFileToDrop) {
```
LawrenceBorst, line 107: NIT. Maybe early return here, but only because it's more consistent with the other render functions.

LawrenceBorst (commented):
Screenshot 2024-03-05 at 13 19 51

^ Small aesthetic small problem in the documentation under "File Dropzone" -> "Overflow Menu"

Looked closely at everything. I'm eager to approve, but we should first resolve all the other comments

Would love to see a test of the dropzone component before the end of cycle 😅, I'll have a go at this

etc/lime-elements.api.md
LawrenceBorst: Ah nevermind, this is probably intended. I see the other components have this repeated declaration as well 😅

src/components/file-dropzone/examples/file-dropzone.tsx
Kiarokh: it should be 🤔 if it's not, add a `?` in front of it then 😅

Kiarokh: I feel it has to be a required pro, otherwise _nobody_ will understand why the border appeared there and what is going on (when they drag some file).

src/components/file-dropzone/file-dropzone.scss
```diff
@@ -0,0 +1,46 @@
...
+    z-index: 1;
+
+    limel-icon.icon {
+        height: clamp(2rem, 4vh, 6rem);
+        width: clamp(2rem, 4vh, 6rem);
+        color: rgb(var(--contrast-800));
+    }
+
+    .text {
+        font-size: clamp(1rem, 2vh, 1.75rem);
```
Kiarokh, line 37: 🥰 I love it that you did this! 😇

Kiarokh:
I'm gonna add something @LawrenceBorst . Don't approve. Wait

src/components/file-dropzone/file-dropzone.scss
Kiarokh: But I think these sizes are still too big when the container is small. Let's say we use this on `limel-file`, will the text be visible? 🤔

src/components/file-dropzone/file-dropzone.scss
```diff
@@ -0,0 +1,46 @@
+:host(limel-file-dropzone) {
+    position: relative;
+    display: grid;
+    position: relative;
+    height: 100%;
+    width: 100%;
```
Kiarokh, line 6: I think these two are unnecessary. It should be up-top the consumer to set a size on this element if they have to. But since this element is just a wrapper, it will inherit its size from their own children. They should just wrap them to provide functionality, and nothing more.

src/components/file-dropzone/examples/file-dropzone-type-filtering.tsx
```diff
@@ -0,0 +1,72 @@
...
+                onFilesSelected={this.handleAcceptedFiles}
+                onFilesRejected={this.handleRejectedFiles}
+                accept="image/*"
+                text="Drop pictures here"
+            >
+                <limel-input-field
+                    label="Dream pet description"
+                    type="textarea"
+                    placeholder="What is your dream pet? Describe it here and attach a picture of it."
+                    helperText="This text will be displayed in your profile"
```
Kiarokh, line 43: This is a good example where the user will not even dream about dragging and dropping a file for uploading it.

image

For now, I will add instructions to the example itself. But I think we should also add this note to the docs.

src/components/file-dropzone/examples/file-dropzone.tsx
LucyChyzhova: text already with `?`, but the icon is the required one, do you think we need to have it opposite? Like icon optional and text required?

src/components/file-dropzone/examples/file-dropzone.tsx
civing: Yea, transforming it to a boolean with double bang feels unnecessary since it's already typed as boolean 😅
See https://github.com/Lundalogik/lime-elements/pull/2802/commits/1a3536491ab5ea8b504a8066e766243722fc909c

src/components/file-dropzone/examples/file-dropzone.tsx
civing: See https://github.com/Lundalogik/lime-elements/pull/2802/commits/1a3536491ab5ea8b504a8066e766243722fc909c

src/components/file-dropzone/file-dropzone.tsx
civing: I complied with your NIT. Consistency is a good thing :)
See https://github.com/Lundalogik/lime-elements/pull/2802/commits/fd2943c0c65ae47108a0007bb29ac454fb0ea459

src/components/file-dropzone/examples/file-dropzone.tsx
LucyChyzhova: I don't think it's hard to understand without a text tbh,  😅 especially when we have an Icon 

https://github.com/Lundalogik/lime-elements/assets/50618208/dac9f625-e569-449b-9338-892686571caf

src/components/file-dropzone/file-dropzone.scss
LucyChyzhova: I will decrease the inset to 0.25rem, otherwise in our new input fields it will look terrible
before:

https://github.com/Lundalogik/lime-elements/assets/50618208/f35f9ec7-2e1b-4230-b3c4-c7769d47aaa0


after:


https://github.com/Lundalogik/lime-elements/assets/50618208/f8aac8a5-0269-4d51-b987-688238a2b6af

LucyChyzhova: It's a good size, I checked before. I think, we should have it this:  

https://github.com/Lundalogik/lime-elements/assets/50618208/30ddbfb1-b98e-4db1-9def-5bd9d4b0538f

src/components/file-dropzone/examples/file-dropzone.tsx
LucyChyzhova: Alright, I made the text required 😅
Let's be safe 😊

LawrenceBorst (approved):

adrianschmidt (approved):
Public interface approved 👍

PR #2820, opened 2024-03-04
Use new dropzone/input component in `limel-file`

civing:
Makes use of the new `limel-file-dropzone` and `limel-file-input` component in `limel-file`. All the other functionality of `limel-file` should be unchanged.

Fixes https://github.com/Lundalogik/crm-feature/issues/3941
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/file/examples/file.tsx
+++ b/src/components/file/examples/file.tsx
@@ -10,7 +10,11 @@ import { Component, h, State } from '@stencil/core';
 })
 export class FileExample {
     @State()
-    private value: FileInfo = { filename: 'letter.docx', id: 123 };
+    private value: FileInfo = {
+        filename: 'cute-cat.jpg',
+        id: 123,
+        href: 'https://www.boredpanda.com/blog/wp-content/uploads/2014/02/funny-wet-cats-36.jpg',
+    };
 
     @State()
     private required = false;
--- a/src/components/file/file.tsx
+++ b/src/components/file/file.tsx
@@ -9,9 +9,7 @@ import {
     EventEmitter,
     h,
     Prop,
-    State,
 } from '@stencil/core';
-import { createRandomString } from '../../util/random-string';
 import {
     getFileBackgroundColor,
     getFileColor,
@@ -132,32 +130,14 @@ export class File {
     @Element()
     private element: HTMLLimelFileElement;
 
-    @State()
-    private isDraggingOverDropZone = false;
-
-    private fileInput: HTMLInputElement;
-    private fileInputId = createRandomString();
     private chipSet;
     private mdcTextField;
-
-    constructor() {
-        this.handleFileSelection = this.handleFileSelection.bind(this);
-        this.handleFileChange = this.handleFileChange.bind(this);
-        this.handleChipSetChange = this.handleChipSetChange.bind(this);
-        this.handleFileDrop = this.handleFileDrop.bind(this);
-        this.handleKeyDown = this.handleKeyDown.bind(this);
-        this.handleKeyUp = this.handleKeyUp.bind(this);
-        this.handleChipInteract = this.handleChipInteract.bind(this);
-    }
 
     public connectedCallback() {
         this.initialize();
     }
 
     public componentDidLoad() {
-        this.fileInput = this.element.shadowRoot.getElementById(
-            this.fileInputId,
-        ) as HTMLInputElement;
         this.chipSet = this.element.shadowRoot.querySelector(CHIP_SET_TAG_NAME);
         this.initialize();
     }
@@ -179,61 +159,23 @@ export class File {
     }
 
     public render() {
-        return [
-            <input
-                hidden={true}
-                id={this.fileInputId}
-                onChange={this.handleFileChange}
-                type="file"
+        return (
+            <limel-file-dropzone
+                disabled={this.disabled || this.readonly || !!this.value}
                 accept={this.accept}
-                disabled={this.disabled || this.readonly}
-            />,
-            <limel-chip-set
-                class={{
-                    'is-file-picker': true,
-                    'shows-dropzone': true,
-                    'highlight-dropzone': this.isDraggingOverDropZone,
-                }}
-                disabled={this.disabled}
-                readonly={this.readonly}
-                invalid={this.invalid}
-                label={this.label}
-                leadingIcon="upload_to_cloud"
-                language={this.language}
-                onChange={this.handleChipSetChange}
-                onClick={this.handleClick}
-                onInteract={this.handleChipInteract}
-                onKeyDown={this.handleKeyDown}
-                onKeyUp={this.handleKeyUp}
-                required={this.required}
-                type="input"
-                value={this.chipArray}
-                title={this.getTranslation('drag-and-drop-tips')}
-                onDragEnter={this.handleDragEnter}
-                onDragOver={this.preventAndStop}
-                onDragLeave={this.handleDragLeave}
-                onDrop={this.handleFileDrop}
-            />,
-        ];
+                onFilesSelected={this.handleNewFiles}
+            >
+                {this.renderChipset()}
+            </limel-file-dropzone>
+        );
     }
 
-    private handleDragEnter = (event: DragEvent) => {
-        this.isDraggingOverDropZone = true;
+    private handleNewFiles = (event: CustomEvent<FileInfo[]>) => {
         this.preventAndStop(event);
+        this.change.emit(event.detail[0]);
     };
 
-    private handleDragLeave = () => {
-        this.isDraggingOverDropZone = false;
-    };
-
-    private handleFileDrop = (event: DragEvent) => {
-        this.preventAndStop(event);
-        this.isDraggingOverDropZone = false;
-        const dataTransfer = event.dataTransfer;
-        this.handleFile(dataTransfer.files[0]);
-    };
-
-    private get chipArray() {
+    private getChipArray(): Chip[] {
         if (!this.value) {
             return [];
         }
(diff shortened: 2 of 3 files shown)


src/components/file/file.tsx
```diff
@@ -254,80 +228,23 @@ export class File {
...
 
-    private handleChipInteract(event: CustomEvent<Chip>) {
+    private handleChipInteract = (event: CustomEvent<Chip>) => {
         event.stopPropagation();
         event.preventDefault();
         this.interact.emit(event.detail.id);
-    }
+    };
 
     private preventAndStop(event: Event) {
```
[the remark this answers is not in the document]
LawrenceBorst: But if you want to keep it, we should also intend to use it in `handleChipInteract` 😅

LawrenceBorst (commented):
Almost ready. I'll add fixups for some of the comments and you can decide if you want to keep them 😛

src/components/file/file.tsx
```diff
@@ -132,32 +130,14 @@ export class File {
...
     public connectedCallback() {
         this.initialize();
     }
 
     public componentDidLoad() {
-        this.fileInput = this.element.shadowRoot.getElementById(
-            this.fileInputId,
-        ) as HTMLInputElement;
         this.chipSet = this.element.shadowRoot.querySelector(CHIP_SET_TAG_NAME);
         this.initialize();
```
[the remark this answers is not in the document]
civing: Although it's good that you check the entirety of the component but this is old code that I haven't touched :)

src/components/file/file.tsx
```diff
@@ -179,58 +159,52 @@ export class File {
     }
 
     public render() {
-        return [
-            <input
-                hidden={true}
-                id={this.fileInputId}
-                onChange={this.handleFileChange}
-                type="file"
+        return (
```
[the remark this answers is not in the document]
civing: This is a good catch, I will disabled it if we have a value

src/components/file/file.tsx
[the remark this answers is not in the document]
civing: I reacted a bit on that function as well. But I decided to keep it. I will have a look at handleChipInteract as well

src/components/file/file.tsx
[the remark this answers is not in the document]
civing: Actually, I had thought about this and set it to disable in those cases. However, it seems that the drop zone didn't respect that property when rendering. I just now merged this PR that will fix it https://github.com/Lundalogik/lime-elements/pull/2831

src/components/file/file.tsx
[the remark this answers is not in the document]
adrianschmidt: You can't just remove the connectedCallback. This will break the component if it's ever detached from the DOM and then re-attached. Like if a consumer wants to move the component from one place in the DOM to another.

adrianschmidt (changes requested):

src/components/file/file.tsx
```diff
@@ -133,17 +133,13 @@ export class File {
     private chipSet;
     private mdcTextField;
 
-    public connectedCallback() {
-        this.initialize();
-    }
```
adrianschmidt, line 155: I realised my comment shows on the wrong code in the Conversation tab, so I'll add it here as well 🙂

> You can't just remove the connectedCallback. This will break the component if it's ever detached from the DOM and then re-attached. Like if a consumer wants to move the component from one place in the DOM to another.

src/components/file/file.tsx
```diff
@@ -133,17 +133,13 @@ export class File {
...
     public componentDidLoad() {
         this.chipSet = this.element.shadowRoot.querySelector(CHIP_SET_TAG_NAME);
         this.initialize();
     }
 
     private initialize() {
-        if (!this.chipSet) {
+        if (!this.mdcTextField) {
             return;
         }
```
adrianschmidt, line 144: I don't understand how this can ever initialise the text field? `this.mdcTextField` is undefined until the following has been run, isn't it?

```ts
        this.mdcTextField = new MDCTextField(
            this.chipSet.shadowRoot.querySelector('.mdc-text-field'),
        );
```

And that won't ever be run, because the function will exit here…?

Unless something else has changes so that `this.mdcTextField` is set somewhere else? I'm puzzled how this can work at all 😅 Maybe we no longer need the `MDCTextField` functionality at all?

But, if nothing relating to this has changed in this PR, I think it should be addressed in a separate PR.

LawrenceBorst: This is my mistake, my intention was to write `if (this.mdcTextField)` 😅

civing: I actually removed the usage of MDCTextField altogether. I didn't think it made much sense and when I tested it it actually behaved better with showing invalid states. Maybe I'm missing something crucial here 😅
See https://github.com/Lundalogik/lime-elements/pull/2820/commits/79e0db9dff4ce1994108055e08b6510f6bdd4f29

src/components/file/examples/file.tsx
```diff
@@ -10,7 +10,11 @@ import { Component, h, State } from '@stencil/core';
 })
 export class FileExample {
     @State()
-    private value: FileInfo = { filename: 'letter.docx', id: 123 };
+    private value: FileInfo = {
+        filename: 'letter.docx',
+        id: 123,
+        href: 'https://leekspin.com/',
```
adrianschmidt, line 16: It feels like this makes more sense if it points to a file…

```suggestion
    private value: FileInfo = {
        filename: 'cute-puppy.jpg',
        id: 123,
        href: 'https://fastly.picsum.photos/id/237/536/354.jpg',
```

src/components/file-input/file-input.tsx
```diff
@@ -100,9 +100,6 @@ export class FileInput {
 
     private handleClick = (event: Event) => {
         if (this.disabled) {
-            event.stopPropagation();
-            event.preventDefault();
-
```
civing, line 105: Is this expected behaviour? If we have the file-input component disabled that means that any click wont open the native file select dialog but will still propagate the click event.

As it was previously any click event were stopped, so if you had something like this the chip would not open a link

```tsx
<limel-file-input disabled={true}>
  <limel-chip text="cat.jpg", link={href: "/cat.jpg"} />
</limel-file-input>
```

src/components/file/file.tsx
```diff
@@ -129,111 +118,42 @@ export class File {
...
-                onDragLeave={this.handleDragLeave}
-                onDrop={this.handleFileDrop}
-            />,
-        ];
+                onFilesSelected={this.handleNewFiles}
+            >
+                <limel-file-input
+                    accept={this.accept}
+                    disabled={this.disabled || this.readonly || !!this.value}
+                >
```
civing, line 131: In one iteration of this PR I wrapped the `<limel-chip-set>` inside a `<limel-file-input>` only if we expected it to be clickable (no file selected). I changed it into this but just now realised we need this change https://github.com/Lundalogik/lime-elements/pull/2820/commits/a85e952b0f09a415ea4e25071db0f5ff2e371a02 to make it work.

It's maybe a bit strange to have it like this if we would like to make the component work with multiple files? Maybe that's a headache for another time ... ?

This comment is also relevant in this context https://github.com/Lundalogik/lime-elements/pull/2820/files#r1521124280

src/components/file/examples/file.tsx
adrianschmidt: Well… ok… but couldn't you have at least used a cute cat then? 🙈

adrianschmidt: (Noticed now that my link didn't work. I think it only works for embedding. I meant to link to [this picture](https://fastly.picsum.photos/id/237/3500/2095.jpg?hmac=y2n_cflHFKpQwLOL1SSCtVDqL8NmOnBzEW7LYKZ-z_o).)

src/components/file-input/file-input.tsx
adrianschmidt: I would say that expected behaviour for this PR is that the component keeps working as it did before… isn't this supposed to just be a refactor?

src/components/file/examples/file.tsx
civing: If you snooze you lose 😛

src/components/file-input/file-input.tsx
civing: Yeah, the implications discussed are in <limel-file-input>. Or what do you mean?

LawrenceBorst (dismissed):
Tested this last night. Everything works as expected. Code looks good, had some comments but they don't really concern the changes here :)

src/components/file/file.tsx
LawrenceBorst: > Maybe that's a headache for another time ... ?

Yes, for another time :)

src/components/file-input/file-input.tsx
adrianschmidt: Ah, ok, I didn't notice that this wasn't `limel-file` 😄

I'm not sure, since I haven't yet gathered enough context, but since this component is meant to wrap some other component, I think the consumer needs to be responsible for controlling the behaviour of the wrapped component. So if they don't want anything to be clickable inside limel-file-input when limel-file-input is disabled, they have to disable the component inside it themselves. If we block all clicks, we make it impossible for the consumer to implement at more nuanced behaviour, like being able to click on existing chips, but not add new files, for example.

adrianschmidt: Sum up of discussion we had in a huddle:

We'll pause this refactoring for now, because we don't use limel-file in the activities composer, so we don't need this refactoring right now. Then we will look into possibly changing the UI a bit. We would like there to be a simple button for browsing files to upload, rather than the currently quite weird interaction where you focus a text-field to get the file-dialog. With keyboard navigation, you currently have to tab to the input field, and then press enter to get the file-dialog. That doesn't seem very intuitive, and is probably really bad for accessibility…

civing:
I've been thinking a lot about the chipset that serves as both clickable chips but also as a file input handler. I would like to move the file input handling to a new button and let the chipset be a chipset and nothing else.

LawrenceBorst (approved):

PR #2832, opened 2024-03-07
Add text editor

LawrenceBorst:
Fixes Lundalogik/crm-feature#3977

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


package.json
```diff
@@ -118,5 +118,9 @@
   "bugs": {
     "url": "https://github.com/Lundalogik/lime-elements/issues"
   },
-  "homepage": "https://github.com/Lundalogik/lime-elements#readme"
+  "homepage": "https://github.com/Lundalogik/lime-elements#readme",
+  "dependencies": {
```
jgroth, line 122: We should not add dependencies as `dependencies`, the should be added as `devDependencies`

src/components/text-editor/text-editor.tsx
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * The `limel-text-editor` component wraps the toast-ui react editor. This editor
+ * allows a rich text editing experience with markdown support.
+ */
+@Component({
+    tag: 'limel-text-editor',
+    shadow: true,
+    styleUrl: 'text-editor.scss',
+})
+export class TextEditor {
```
jgroth, line 17: To make sure this is compatible to be used in a form, it should implement the `FormComponent<string>` interface

src/components/text-editor/text-editor.tsx
```diff
@@ -0,0 +1,98 @@
...
+    @Prop()
+    initialValue: string = '';
+
+    @Prop()
+    editType: EditorType = 'wysiwyg';
```
jgroth, line 22: Exposing external types like this can be dangerous, since if a breaking change is introduced we might not notice it. It's better to create our own type and map it to the original type

src/components/text-editor/text-editor.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+    @Prop()
+    theme: 'dark' | 'light' = 'light';
```
jgroth, line 25: I don't think our other components have properties for light/dark mode, we should make this component work the same way

src/components/text-editor/text-editor.tsx
```diff
@@ -0,0 +1,126 @@
...
+        };
+
+        ReactDOM.render(
+            createElement(Editor, props),
+            this.host.shadowRoot.querySelector('#editor'),
+        );
+    }
+
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    private handleChange = (_: EditorType): void => {
```
jgroth, line 112:
```suggestion
    private handleChange = (): void => {
```

src/components/text-editor/text-editor.tsx
```diff
@@ -0,0 +1,126 @@
...
+            initialEditType: this.editType,
+            language: LANGUAGE,
+            placeholder: this.placeholder,
+            previewStyle: PREVIEW_STYLE,
+            theme: 'light',
+            toolbarItems: TOOLBAR_ITEMS,
+            useCommandShortcut: true,
+            usageStatistics: false,
+
+            // @ts-ignore
```
jgroth, line 98: This looks like an error :smile:

src/components/text-editor/text-editor.tsx
```diff
@@ -0,0 +1,126 @@
...
+        this.editor.changeMode(newValue);
+    }
+
+    @Watch('placeholder')
+    public handlePlaceholderChange(newValue: string) {
+        this.editor.setPlaceholder(newValue);
+    }
+
+    constructor() {
+        this.editorRef = createRef();
```
jgroth, line 77: I don't think we need this, refs in React is just used to get references to DOM elements, and we have other ways of doing that in Stencil

src/components/text-editor/text-editor.tsx
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * Emitted when the editor type changes
+     */
+    @Event()
+    public changeMode: EventEmitter<EditorType>;
+
+    @Element()
+    private host: HTMLElement;
+
+    private editorRef: RefObject<Editor> = null;
```
jgroth, line 64: I think this can be removed

src/components/text-editor/text-editor.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+    @Watch('placeholder')
+    public handlePlaceholderChange(newValue: string) {
+        this.editor.setPlaceholder(newValue);
+    }
+
+    constructor() {
+        this.editorRef = createRef();
+
+        this.handleChange = this.handleChange.bind(this);
```
jgroth, line 79: Use an arrow function instead of binding the method manually like this

src/components/text-editor/text-editor.tsx
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    public render() {
+        return <div id="editor"></div>;
```
Kiarokh, line 83: I don't see all of their buttons neither 
theirs:
image
ours:
image

PR #2843, opened 2024-03-14
Add rich text editor (prosemirror version)

LawrenceBorst:
Fixes Lundalogik/crm-feature#3977

This PR in its current state just shows how the examples from the documentation might be added to the CRM, to show that things do work. But we still have to write a more robust editor from the ground up.

For reviewing:
To see what is available in Prosemirror, you may want to have a look at the Prosemirror repositories here:

https://github.com/orgs/ProseMirror/repositories?type=all 

The guide is excellent, and found at https://prosemirror.net/docs/guide/

The tutorials are also excellent https://prosemirror.net/examples/

From investigating this, I would say there are no real limitations. But implementation is generally much harder. In this issue we want to create a wrapper around a Prosemirror view, and construct the editor to have

1) Custom "lime-ish" icons
2) Custom buttons
3) Default schema (we could define our schema ourselves, but this is tricky and prone to error)
4) Emit custom events relevant to and at the right level of our use case
5) Headers, bold, italic and lined text
6) Images upload
7) Undo, redo (this comes quite readily with the history plugin)
8) Code blocks
9) Ordered and unordered lists
10) Lime-ish font and related serialisation choices for marks and nodes

In the future it will certainly be possible to add links to limeobjects, and even define their appearance in the editor.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,6 +83,8 @@
     "number-abbreviate": "^2.0.0",
     "parse-css-color": "^0.2.1",
     "prettier": "^3.2.5",
+    "prosemirror-example-setup": "^1.2.2",
+    "prosemirror-schema-basic": "^1.2.2",
     "puppeteer": "^19.11.1",
     "react": "^18.2.0",
     "react-dom": "^18.2.0",
--- a/src/components/text-editor/examples/text-editor-basic.tsx
+++ b/src/components/text-editor/examples/text-editor-basic.tsx
@@ -1,0 +1,31 @@
+import { LimelTextEditorCustomEvent } from '@limetech/lime-elements';
+import { Component, h, State } from '@stencil/core';
+/**
+ * Basic example
+ *
+ * Try typing and editing text, or copy & paste in some rendered HTML code
+ * from your browser into the editor to see how it is rendered and what you get
+ * as an output value.
+ */
+@Component({
+    tag: 'limel-example-text-editor-basic',
+    shadow: true,
+})
+export class BasicTextEditorBasicExample {
+    @State()
+    private text: { html: string } = { html: '' };
+
+    public render() {
+        return [
+            <limel-text-editor onChange={this.handleChange} />,
+            <limel-example-value value={this.text} />,
+        ];
+    }
+
+    private handleChange = (
+        event: LimelTextEditorCustomEvent<{ html: string }>,
+    ): void => {
+        event.stopPropagation();
+        this.text = event.detail;
+    };
+}
--- a/src/components/text-editor/partial-styles/_prosemirror-menu.scss
+++ b/src/components/text-editor/partial-styles/_prosemirror-menu.scss
@@ -1,0 +1,180 @@
+@use '../../../style/mixins';
+
+$size-of-caret: 0.25rem;
+
+div#editor {
+    .ProseMirror-menubar {
+        position: sticky !important;
+        z-index: 1;
+        top: 0;
+
+        &[style*='position: fixed'] {
+            // They add this dynamically as soon as you scroll.
+            // I want to catch this to detect the scroll,
+            // to add proper styles. This is a hack.
+            box-shadow: 0 0.25rem 0.5rem -0.5rem rgb(var(--color-black), 0.8);
+        }
+    }
+}
+
+.ProseMirror-menubar {
+    position: relative;
+    z-index: 1;
+    display: flex;
+    flex-wrap: wrap;
+    gap: 0.25rem;
+    align-items: center;
+
+    padding: 0.125rem;
+    background-color: rgb(var(--contrast-100));
+    border-top-left-radius: 0.5rem;
+    border-top-right-radius: 0.5rem;
+}
+
+.ProseMirror-menuitem {
+    &:not(:has(.ProseMirror-menu-disabled)) {
+        @include mixins.is-flat-clickable;
+    }
+    position: relative;
+    flex-shrink: 0;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+
+    white-space: nowrap;
+    line-height: 1;
+
+    min-height: 1.5rem;
+    min-width: 1.5rem;
+    border-radius: 0.25rem;
+    color: rgb(var(--contrast-1100));
+    font-size: 0.8125rem;
+
+    &:has(.ProseMirror-menu-active) {
+        box-shadow: var(--button-shadow-inset);
+        color: var(--mdc-theme-primary);
+
+        svg {
+            color: var(--mdc-theme-primary);
+        }
+    }
+
+    &:has(.ProseMirror-menu-dropdown-menu) {
+        box-shadow: var(--button-shadow-inset);
+    }
+
+    svg {
+        fill: currentColor;
+        height: 1rem;
+        color: rgb(var(--contrast-1100));
+    }
+}
+
+.ProseMirror-menuseparator {
+    border-radius: 1rem;
+    background-color: rgb(var(--contrast-600));
+    width: 0.125rem;
+    height: 1rem;
+}
+
+.ProseMirror-menu-dropdown {
+    position: relative;
+    display: flex;
+    align-items: center;
+    gap: 0.5rem;
+    padding: 0 0.5rem;
+
+    &:after {
+        content: '';
+        border-left: $size-of-caret solid transparent;
+        border-right: $size-of-caret solid transparent;
+        border-top: $size-of-caret solid currentColor;
+    }
+}
+
+.ProseMirror-menu-dropdown-menu,
+.ProseMirror-menu-submenu {
+    //the same shadow as limel-menu
+    box-shadow:
+        0px 5px 5px -3px rgba(0, 0, 0, 0.2),
+        0px 8px 10px 1px rgba(0, 0, 0, 0.14),
+        0px 3px 14px 2px rgba(0, 0, 0, 0.12);
+}
+
+.ProseMirror-menu-dropdown-menu {
+    position: absolute;
+    top: 100%;
+    background: rgb(var(--contrast-100));
+    border-radius: 0.25rem;
+    padding: 0.125rem;
+}
+
+.ProseMirror-menu-submenu {
+    position: absolute;
+    background: rgb(var(--contrast-100));
+    border-radius: 0.25rem;
+    padding: 0.125rem;
+    display: none;
+    min-width: 4em;
+    left: 100%;
+    top: -0.75rem;
+}
+
+.ProseMirror-menu-dropdown-item {
+    transition: background-color 0.2s ease;
+    cursor: pointer;
+    border-radius: 0.375rem;
+
+    min-height: 2.5rem;
+    display: flex;
+    align-items: center;
+    width: 100%;
+
+    > div {
+        padding: 0 1rem;
+        width: 100%;
+    }
+
+    &:not(:has(.ProseMirror-menu-disabled)) {
+        &:hover {
+            background-color: rgb(var(--contrast-300));
+        }
+    }
+}
+
+.ProseMirror-menu-submenu-wrap {
+    position: relative;
+    margin-right: -0.25rem;
+
+    &:hover {
+        .ProseMirror-menu-submenu {
+            display: block;
+        }
+    }
+}
+
+.ProseMirror-menu-submenu-label {
+    position: relative;
+    display: flex;
+    align-items: center;
+    justify-content: space-between;
+    gap: 0.5rem;
+
+    &:after {
+        content: '';
+        border-top: $size-of-caret solid transparent;
+        border-bottom: $size-of-caret solid transparent;
+        border-left: $size-of-caret solid currentColor;
+    }
+}
+
+.ProseMirror-menu-disabled {
+    opacity: 0.3;
+    cursor: default;
+}
+
+.ProseMirror-menu-submenu-wrap-active {
+    .ProseMirror-menu-submenu {
+        display: block;
+    }
+}
--- a/src/components/text-editor/partial-styles/_prosemirror.scss
+++ b/src/components/text-editor/partial-styles/_prosemirror.scss
@@ -1,0 +1,62 @@
+.ProseMirror {
+    position: relative;
+    word-wrap: break-word;
+    white-space: pre-wrap;
+    white-space: break-spaces;
+    -webkit-font-variant-ligatures: none;
+    font-variant-ligatures: none;
+    font-feature-settings: 'liga' 0;
+
+    border-bottom-left-radius: 0.5rem;
+    border-bottom-right-radius: 0.5rem;
+
+    padding: 0.5rem 1rem;
+    background-color: rgb(var(--contrast-100));
+
+    [draggable][contenteditable='false'] {
+        user-select: text;
+    }
+
+    &:focus-visible {
+        outline: none;
+    }
+}
+
+.ProseMirror-hideselection {
+    * {
+        &::selection {
+            background: transparent;
+        }
+
+        &::-moz-selection {
+            background: transparent;
+        }
+    }
+
+    caret-color: transparent;
+}
+
+.ProseMirror-selectednode {
+    outline: 0.125rem solid rgb(var(--color-sky-light));
+}
+
+li.ProseMirror-selectednode {
+    outline: none;
+
+    &:after {
+        content: '';
+        position: absolute;
+        left: -2rem;
+        right: -0.125rem;
+        top: -0.125rem;
+        bottom: -0.125rem;
+        border: 0.125rem solid rgb(var(--color-sky-light));
+        pointer-events: none;
+    }
+}
+
+img.ProseMirror-separator {
+    display: inline !important;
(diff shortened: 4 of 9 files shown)


src/components/file/file.types.ts
LawrenceBorst, line 1: Ignore this, I didn't mean to add this here but did since I got a "file doesn't exist" error locally.

package.json
```diff
@@ -120,5 +120,9 @@
   "bugs": {
     "url": "https://github.com/Lundalogik/lime-elements/issues"
   },
-  "homepage": "https://github.com/Lundalogik/lime-elements#readme"
+  "homepage": "https://github.com/Lundalogik/lime-elements#readme",
+  "dependencies": {
+    "prosemirror-example-setup": "^1.2.2",
+    "prosemirror-schema-basic": "^1.2.2"
```
adrianschmidt, line 126: Our tree-shaking ought to mean these only have to be dev-dependencies.

src/components/text-editor/text-editor/text-editor.tsx
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @exampleComponent limel-example-basic-text-editor
+ * @beta
+ */
+@Component({
+    tag: 'limel-text-editor',
+    shadow: true,
+    styleUrl: 'text-editor.scss',
+})
+export class TextEditor {
```
LawrenceBorst, line 27: Maybe we should implement `FormComponent` here, as they potentially wanted the editor in input fields 🤔

adrianschmidt: Yeah, that sounds like a good idea.

(Some components have a wrapper in /src/form/widgets/ instead, but I think that's probably only components that we already had before we added limel-form.)

src/components/text-editor/examples/basic-text-editor.tsx
```diff
@@ -0,0 +1,34 @@
...
+                <hr />
+                <h2>Preview using limel-markdown:</h2>
+                <limel-markdown value={this.text.html} />
+                <hr />
+                <limel-example-value value={this.text} />
+            </div>
+        );
+    }
+
+    handleChange = (
```
adrianschmidt, line 28:
```suggestion
    private handleChange = (
```

src/components/text-editor/partial-styles/_prosemirror.scss
adrianschmidt, line 1: There's currently quite a fat outline, which hides the cursor when the cursor is a the beginning of a line:

image

src/components/text-editor/examples/basic-text-editor.tsx
adrianschmidt, line 1: Inserting an image doesn't work, and will require interaction with the consumer's code, so I think that should be excluded from the basic example. We shouldn't focus on getting that part to work at this time, but once we do, we should add a separate example, that shows how to use that functionality.

adrianschmidt, line 1: There's a button to "Select parent node":

image

That seems like something we should remove from the basic example. I'm not entirely sure what it's intended for, but I think it's just going to confuse the average user.

src/components/text-editor/partial-styles/_prosemirror.scss
adrianschmidt, line 1: Btw, @LucyChyzhova perhaps, I think we need some styling to make this look more like a regular `textarea`. Currently, even though the menu bar is there, I'm missing the visual cue that tells me that I can click in the area and write stuff there.

image

src/components/text-editor/partial-styles/_prosemirror.scss
```diff
@@ -0,0 +1,66 @@
...
+li.ProseMirror-selectednode:after {
+    content: '';
+    position: absolute;
+    left: -32px;
+    right: -2px;
+    top: -2px;
+    bottom: -2px;
+    border: 2px solid #8cf;
+    pointer-events: none;
+}
```
adrianschmidt, line 58: The lone `-32px` kind of stands out from all the `-2px`, so I thought I'd ask if it's intentional or a typo 😊

src/components/text-editor/examples/text-editor.tsx
```diff
@@ -4,25 +4,22 @@ import { Component, h, State } from '@stencil/core';
  * A basic text editor example.
  */
 @Component({
-    tag: 'limel-example-basic-text-editor',
+    tag: 'limel-example-text-editor',
```
adrianschmidt, line 7: Is it OK if we call the example file `examples/text-editor-basic.tsx` and the tag `limel-example-text-editor-basic`?

I know we don't do that _everywhere_ yet, but I'm trying to work our way there.

(I'm happy to make the change myself, I just want to know if that's a change you would be OK with? 😊)

src/components/text-editor/examples/text-editor-prefilled-value.tsx
```diff
@@ -0,0 +1,54 @@
...
+    //     You can have properly indented paragraphs within list items. Notice the blank line above, and the leading spaces (at least one, but we'll use three here to also align the raw Markdown).
+
+    //     To have a line break without a paragraph, you will need to use two trailing spaces.
+    //     Note that this line is separate, but within the same paragraph.
+    //     (This is contrary to the typical GFM line break behaviour, where trailing spaces are not required.)
+
+    // * Unordered list can use asterisks
+    // - Or minuses
+    // + Or pluses
+    // `;
```
Kiarokh, line 32: This component should be able to get replaced with a `limel-input-field type="text"`.
This means also that if there is already some preset `value` (like in the input fields), we should be able to get that and display it inside the component. So I just added this example so that we can implement and test this.

src/components/text-editor/examples/text-editor-prefilled-value.tsx
```diff
@@ -0,0 +1,54 @@
...
+
```
Kiarokh, line 33: ☝️ that should also mean that we probably need some sort of support for props like `disabled`, `readonly` (maybe)?

As well as support for `placeholder` and `label` perhaps?

src/components/text-editor/examples/text-editor.tsx
Kiarokh: I'm fine either ways. But I think in most of our components we don't call the basic (first) example, "basic" 😄 So I tried to keep it consistent with them. But whatever works. We should just decide one day on all of these conventions and follow them. Also update all so that new things can follow the old conventions.

src/components/text-editor/partial-styles/_prosemirror.scss
Kiarokh: Agree. We're still not doe with the styling though. We will continue.

src/components/text-editor/partial-styles/_prosemirror.scss
Kiarokh: Ignore the styles all together for now. We're not done yet. 😊

src/components/text-editor/examples/text-editor.tsx
adrianschmidt: I made an "executive decision" years ago, but haven't had the time to go around and update every single instance. You can do it for me if you want to! 😝

src/components/text-editor/text-editor.tsx
```diff
@@ -0,0 +1,68 @@
...
+    @State()
+    private view: EditorView;
+
+    @Event()
+    private change: EventEmitter<{ html: string }>;
+
+    public componentWillLoad() {}
+
+    public render() {
+        return [<div id="editor" />, <div id="content" />];
```
Kiarokh, line 41: the generated HTML looks like this:
image

1. The `<div class="ProseMirror-menubar-wrapper">` contains the content too, and the `<div id="content"></div>` is rendered empty. So something is wrong here. 

When I check [their example](https://prosemirror.net/) they don't seem to have this `<div id="content">`. Why do WE have it?

src/components/text-editor/examples/text-editor-prefilled-value.tsx
adrianschmidt: Yeah, makes sense. We'll have to see how we solve that, I guess.

src/components/text-editor/examples/text-editor-prefilled-value.tsx
adrianschmidt: Yeah, true. I think we should focus on the basic functionality first though. We can add support for these later. Let's try to not make this first PR the mother of all PRs, OK? 🙏

src/components/text-editor/text-editor.tsx
```diff
@@ -0,0 +1,68 @@
...
+                    plugins: exampleSetup({ schema: mySchema }),
+                }),
+                dispatchTransaction: (transaction) => {
+                    const newState = this.view.state.apply(transaction);
+                    this.view.updateState(newState);
+
+                    this.change.emit({ html: this.view.dom.innerHTML });
+                },
+            },
+        );
```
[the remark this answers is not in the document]
adrianschmidt: I think it's a perfectly OK state for this PR to simply hide this functionality, and add it in a separate PR later on (still in this cycle obviously).

I _really_ don't want to have to keep re-re-re-reviewing this PR, the way it always ends up when a new large feature is being added and more stuff just keeps getting squeezed in.

src/components/text-editor/text-editor.tsx
```diff
@@ -0,0 +1,68 @@
...
+} from '@stencil/core';
+import { EditorState } from 'prosemirror-state';
+import { EditorView } from 'prosemirror-view';
+import { Schema, DOMParser } from 'prosemirror-model';
+import { schema } from 'prosemirror-schema-basic';
+import { exampleSetup } from 'prosemirror-example-setup';
+
+/**
+ * The `limel-text-editor` component wraps the toast-ui react editor. This editor
+ * allows a rich text editing experience with markdown support.
```
adrianschmidt, line 17: We should update this description before merging.

src/components/text-editor/examples/text-editor.tsx
```diff
@@ -0,0 +1,31 @@
...
+    shadow: true,
+})
+export class BasicTextEditorExample {
+    @State()
+    private text: { html: string } = { html: '' };
+
+    public render() {
+        return [
+            <limel-text-editor onChange={this.handleChange} />,
```
Kiarokh, line 16: the buttons in the menu-bar need clicking twice to do anything. 
So if you want to make something bold or un-bold it, you need to click on the <kbd>B</kbd> two times.

src/components/text-editor/text-editor.tsx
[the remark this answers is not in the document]
Kiarokh: ok

src/components/text-editor/text-editor/text-editor.tsx
LawrenceBorst: Did this in a separate PR, was quite a large change

src/components/text-editor/examples/text-editor-prefilled-value.tsx
LawrenceBorst: Solved in the other PR

src/components/text-editor/examples/text-editor-prefilled-value.tsx
LawrenceBorst: Solved in the other PR

src/components/text-editor/partial-styles/_prosemirror.scss
LucyChyzhova: sure, I agree as well

src/components/text-editor/text-editor.tsx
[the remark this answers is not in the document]
Kiarokh: https://github.com/Lundalogik/crm-feature/issues/4036

src/components/text-editor/text-editor.tsx
Kiarokh: ⚡

src/components/text-editor/examples/text-editor-prefilled-value.tsx
Kiarokh: I feel using `helperText` might be an overkill. Also perhaps `label` is an overkill. I don't know. But it's so common in forms, so we maybe need to do something about it. Another thing we need is `invalid` but let's talk about them in your other PR.

src/components/text-editor/text-editor.tsx
LawrenceBorst: Yeah it seems okay to not have it. The tutorial does it differently but everything continues working (and I think the button clicking problem you mentioned below even goes away?)

src/components/text-editor/examples/text-editor.tsx
LawrenceBorst: Is this still the case now?

src/components/text-editor/text-editor.tsx
LucyChyzhova: yes, the button-clicking problem no longer exists!

src/components/text-editor/examples/text-editor.tsx
LucyChyzhova: not anymore!

src/components/text-editor/examples/basic-text-editor.tsx
Kiarokh: https://github.com/Lundalogik/crm-feature/issues/4037

src/components/text-editor/examples/text-editor.tsx
Kiarokh: did it for this one.

src/components/text-editor/examples/text-editor-basic.tsx
```diff
@@ -0,0 +1,35 @@
...
+    @State()
+    private text: { html: string } = { html: '' };
+
+    public render() {
+        return [
+            <limel-text-editor onChange={this.handleChange} />,
+            <hr />,
+            <h2>Preview using limel-markdown:</h2>,
+            <limel-markdown value={this.text.html} />,
+            <hr />,
```
Kiarokh, line 24: We're importing the markdown styles into this component.
So this section shouldn't really be necessary.
We can keep it for now while we're testing things. 
However when we publish the component, we don't need to have this in the docs.

src/components/text-editor/partial-styles/_prosemirror-menu.scss
```diff
@@ -2,6 +2,21 @@
...
+    .ProseMirror-menubar {
+        position: sticky !important;
+        z-index: 1;
+        top: 0;
+
+        &[style*='position: fixed'] {
+            // They add this dynamically as soon as you scroll.
+            // I want to catch this to detect the scroll,
+            // to add proper styles. This is a hack.
+            box-shadow: 0 0.25rem 0.5rem -0.5rem rgb(var(--color-black), 0.8);
```
adrianschmidt, line 15: > This is a hack.

You don't say? 😝 🤣

adrianschmidt: "When you know the rules, you are allowed to break the rules" 😉

src/components/text-editor/examples/text-editor-basic.tsx
```diff
@@ -0,0 +1,35 @@
+import { LimelTextEditorCustomEvent } from '@limetech/lime-elements';
+import { Component, h, State } from '@stencil/core';
+/**
+ * A basic example
```
Kiarokh, line 4:
```suggestion
 * Basic example
```

src/components/text-editor/examples/text-editor-basic.tsx
```diff
@@ -0,0 +1,35 @@
+import { LimelTextEditorCustomEvent } from '@limetech/lime-elements';
```
Kiarokh, line 1: is this really needed? 🤔

adrianschmidt: There was some version of Stencil that started complaining if you used `CustomEvent` instead of the correct type output by Stencil (in this case `LimelTextEditorCustomEvent`). But, no it's not strictly necessary. I think you could generally use `CustomEvent` instead, and you could of course just type the event as `any` in the event handler. But this is the correct type, and this is an example where we try to influence the reader to use good coding style, so I think we should keep it 😊

src/components/text-editor/partial-styles/_prosemirror-menu.scss
adrianschmidt: Yeah, I understood. I just couldn't resist poking you because it felt so _utterly_ hackish 😂 🤗

src/components/text-editor/examples/text-editor-basic.tsx
adrianschmidt: ⚡

adrianschmidt (approved):

PR #2844, opened 2024-03-14
Menu search less loading

vicgeralds:
Removes additional debouncing of search in `limel-menu` since we already have debounced change events from the text input.

This means we avoid rendering a loading animation for more than half a second when the searcher resolves instantly.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu/menu.tsx
+++ b/src/components/menu/menu.tsx
@@ -190,6 +190,7 @@ export class Menu {
     private searchInput: HTMLLimelInputFieldElement;
     private portalId: string;
     private triggerElement: HTMLSlotElement;
+    private selectedMenuItem?: MenuItem;
 
     constructor() {
         this.portalId = createRandomString();
@@ -601,6 +602,7 @@ export class Menu {
         selectOnEmptyChildren: boolean = true,
     ) => {
         if (Array.isArray(menuItem?.items) && menuItem.items.length > 0) {
+            this.selectedMenuItem = menuItem;
             this.clearSearch();
             this.currentSubMenu = menuItem;
             this.navigateMenu.emit(menuItem);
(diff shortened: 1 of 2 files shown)


src/components/menu/menu.tsx
```diff
@@ -443,13 +423,18 @@ export class Menu {
...
 
             return;
         }
 
         this.loadingSubItems = true;
 
-        const result = await this.debouncedSearch(query);
+        const result = await this.searcher(query);
+
+        if (this.searchValue !== query) {
```
mjlime, line 435: We set `this.searchValue` to `query` a few lines above, how is this ever supposed to be `true`? The early return from this if-statement also forgets to reset `this.loadingSubItems` to `false`?

src/components/menu/menu.tsx
```diff
@@ -627,6 +615,11 @@ export class Menu {
             const menuLoader = menuItem.items as MenuLoader;
             this.loadingSubItems = true;
             const subItems = await menuLoader(menuItem);
+
+            if (this.selectedMenuItem !== menuItem) {
```
mjlime, line 619: Same as the comment above about this probably never going to return true and thus not work as expected, as well as missing the reset of `this.loadingSubItems` to `false`.

src/components/menu/menu.tsx
vicgeralds: We're awaiting a search request that could take seconds. Meanwhile, other searches that complete faster could be made, or the search is cleared for some reason (`this.clearSearch()` is called from six different places in this component), for example because the component is rendered with new items.

vicgeralds: Not setting `this.loadingSubItems` to `false` is intentional, because we're no longer loading the same search. We expect `this.loadingSubItems` to be set to `false` when the other search completes or when search is cleared.

src/components/menu/menu.tsx
vicgeralds: Right, was missing a reset b9c0f16c9e9af8fde0f342920cc5afe019f25a15

We're setting `this.loadingSubItems` to `false` in `this.clearSearch()` so I think it's covered.

vicgeralds: c13d5483183ce15226cb6e39e0ccfcb54c9d3141

mjlime (dismissed):

jgroth (approved):

PR #2847, opened 2024-03-14
fix(file-dropzone): display the component as a block element

Kiarokh:
Previously, it was a unnecessarily a `grid`,
which used to set certain expectations on its children. A block element is generic, and the consumer
can change it to whatever is suitable where they use it.

required for https://github.com/Lundalogik/lime-crm-components/pull/2187/

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/file-dropzone/examples/file-dropzone-type-filtering.scss
+++ b/src/components/file-dropzone/examples/file-dropzone-type-filtering.scss
@@ -4,16 +4,11 @@
     gap: 1.5rem;
 }
 
-limel-input-field,
-div {
-    min-height: 10rem;
-    background-color: rgb(var(--contrast-100));
-    border-radius: 0.5rem;
-}
-
 div {
     display: flex;
     align-items: center;
     justify-content: center;
     border: 1px solid rgb(var(--contrast-800));
+    border-radius: 0.5rem;
+    min-height: 10rem;
 }
--- a/src/components/file-dropzone/file-dropzone.scss
+++ b/src/components/file-dropzone/file-dropzone.scss
@@ -1,5 +1,5 @@
 :host(limel-file-dropzone) {
-    display: grid;
+    display: block;
     position: relative;
 }
 


src/components/file-dropzone/examples/file-dropzone-type-filtering.scss
```diff
@@ -4,16 +4,15 @@
     gap: 1.5rem;
 }
 
-limel-input-field,
-div {
-    min-height: 10rem;
-    background-color: rgb(var(--contrast-100));
-    border-radius: 0.5rem;
+limel-input-field {
+    --textarea-height: 6rem;
```
LucyChyzhova, line 8: I don't think we should lock the resizing by setting strict height.  It's good if consultants can resize the input field and see that the dropzone also growing.

src/components/file-dropzone/file-dropzone.scss
```diff
@@ -1,5 +1,5 @@
 :host(limel-file-dropzone) {
-    display: grid;
+    display: block;
```
LucyChyzhova, line 2: What if we use display: flex here?

Kiarokh: A "block element" is an HTML element that typically starts on a new line and takes up the full width available to it. This way, it creates a block-like structure on the web page. 

Examples of block-level elements include `<div>`, `<p>`, `<header>`, `<footer>`, `<section>`, and `<article>`. These elements are often used for structuring and organizing content on a webpage. 

Since this component is a "wrapper", it shouldn't automatically affect the layout of its children. It should be neutral. The only neutral thing we have is `block`, which will turn it into a block element. It's essentially telling the web-browser that it should treat `limel-file-dropzone` like a `div`.

Both `gird` and `flex` affect how their children are distributed in the layout. Grid forces the children to be rendered in rows (vertically), and Flex forces its children to be rendered in columns (horizontally). If we use this, we will have to document this behavior, and ask the consumers to override this or think about it every time they use this component.

Sometimes, the consumers should still override the layout, buy setting `display: something-else;` on `<limel-file-dropzone />`. This is what we did in the web-client, since there was 3 children there already and they were set-up to be flex children. So we turned `limel-file-dropzone` to a `flex` there.

src/components/file-dropzone/examples/file-dropzone-type-filtering.scss
Kiarokh: good idea!

src/components/file-dropzone/file-dropzone.scss
LucyChyzhova: Thanks, Kia! I was not sure since we are overriding it in the webclient. However, your explanations sound reasonable, let me test it on both sides one more time. Yesterday it broke the webclient layout.

Kiarokh: thanks Lucy. Just make sure you're on the latest on the web client and crm components. That problem is already fixed and merged there, and is independent of this change here.

LucyChyzhova (approved):

PR #2853, opened 2024-03-19
Readonly boolean 2

jgroth:
Suggestions for improvements for #2747 

- The private component is public, renamed to `limel-label` to make it reusable in other scenarios
- Changed the interface of the component, can now take other values and not only `booleans`
- Changed the interface of the `ReadonlyProps`, the `Label` interface now represents a single label with a `value`, `text` and `icon`


src/components/label/examples/label.scss
```diff
@@ -1,4 +1,4 @@
-:host(limel-example-readonly-boolean) {
+:host(limel-example-label) {
```
Kiarokh, line 1: we should also remember to reword the commit title

src/components/checkbox/examples/checkbox-readonly.tsx
```diff
@@ -45,15 +44,21 @@ export class CheckboxReadonlyExample {
...
-                readonlyProps={{
-                    trueIcon: 'news',
-                    falseIcon: {
-                        name: 'cancel_subscription',
-                        color: 'rgb(var(--color-orange-default))',
+                readonlyLabels={[
+                    {
+                        value: true,
+                        icon: 'news',
+                        text: 'Is subscribed to receive newsletters',
```
Kiarokh, line 51: 🤔 these don't seem to work for `true`

src/components/switch/examples/switch-readonly.tsx
```diff
@@ -42,15 +41,21 @@ export class SwitchReadonlyExample {
...
+                        icon: 'news',
+                        text: 'Is subscribed to receive newsletters',
+                    },
+                    {
+                        value: false,
+                        icon: {
+                            name: 'cancel_subscription',
+                            color: 'rgb(var(--color-orange-default))',
+                        },
+                        text: 'Is unsubscribed from newsletters',
```
Kiarokh, line 56: https://github.com/Lundalogik/lime-elements/assets/35954987/9deb035f-ad49-4f6c-9ab2-24cbbc228563

these seem to work in a weird way for limel-switch. You have to toggle the selected state for them to work

src/components/label/examples/label.tsx
Kiarokh, line 1: Uploading Screen Recording 2024-03-26 at 11.57.24.mov…

src/components/label/label.tsx
```diff
@@ -0,0 +1,87 @@
...
+ * The reason we offer this component is that the default styling
+ * of the Checkbox or Toggle switch in the `readonly` state may not always
+ * provide the best way of _visualizing information_, potentially leading to
+ * confusion and negatively affecting the end-users' experience.
+ *
+ * @exampleComponent limel-example-label
+ * @beta
+ */
+@Component({
+    tag: 'limel-label',
```
Kiarokh, line 23:
```suggestion
    tag: 'limel-dynamic-label',
```

src/components/label/label.tsx
```diff
@@ -0,0 +1,87 @@
+import { getIconName } from '../icon/get-icon-props';
+import { Component, Prop, h } from '@stencil/core';
+import { Label, LabelValue } from './label.types';
+import { Icon } from '../../interface';
+
+/**
+ * This components displays a different label depending on the current given
+ * value. A label can consist of a text and an icon. If no matching label is
+ * found, the default label will be displayed.
```
Kiarokh, line 9: To me this description is unclear:

> If no matching label is found, the default label will be displayed.

src/components/label/label.tsx
```diff
@@ -0,0 +1,87 @@
...
+     * Default label to display if no label with corresponding value is found
+     */
+    @Prop({ reflect: true })
+    public defaultLabel: Omit<Label, 'value'> = {};
+
+    /**
+     * Available labels
+     */
+    @Prop()
+    public labels: Label[] = [];
```
Kiarokh, line 47: I can't say that I read these docs and understand what these props do 😅 
Even when I look at the code, I am still confused. But the consumer only sees the props' description in the generated docs. So the docs should make it clear what these props do.

I tried to improve the docs like below:

```suggestion
    /**
     * The current value of the component; 
     * which is used to match with the `labels` array to determine the label to display. 
     * If no matching label is found, the `defaultLabel` is displayed.
     */
    @Prop({ reflect: true })
    public value: LabelValue;

    /**
     * The label to display when no matching value is found in the `labels` array. 
     * This is a fallback label that ensures there's always a label displayed for the component.
     */
    @Prop({ reflect: true })
    public defaultLabel: Omit<Label, 'value'> = {};

    /**
     * An array of available labels. Each label in this array has a corresponding value. 
     * When the `value` prop matches a value in this array, the corresponding label is displayed.
     */
    @Prop()
    public labels: Label[] = [];
```

However, I am still unsatisfied and confused 😕 .

What if we rename the component to `limel-dynamic-label` and then we can rename the props like these:


```suggestion
    /**
     * The current value of the component; 
     * which is used to match with the `dynamicLabels` array to determine the label to display. 
     * If no matching label is found, the `label` is displayed.
     */
    @Prop({ reflect: true })
    public value: LabelValue;

    /**
     * The label to display when no matching value is found in the `dynamicLabels` array. 
     * This is a fallback label that ensures there's always a label displayed for the component.
     */
    @Prop({ reflect: true })
    public label: Omit<Label, 'value'> = {};

    /**
     * An array of available labels which can be dynamically applied based on the given value. 
     * Each label in this array will have to have a corresponding value. 
     * When the `value` prop matches a value in this array, 
     * the corresponding label is dynamically displayed.
     */
    @Prop()
    public dynamicLabels: Label[] = [];
```

A bit better. But what do you think?

src/components/label/examples/label.scss
jgroth: Everything is fixups to your commits and targets your branch :smile:

src/components/checkbox/examples/checkbox-readonly.tsx
jgroth: cc42437a8c21ff9e60fa86f940011d4ce5f90d81

src/components/switch/examples/switch-readonly.tsx
jgroth: cc42437a8c21ff9e60fa86f940011d4ce5f90d81

src/components/label/examples/label.tsx
jgroth: cc42437a8c21ff9e60fa86f940011d4ce5f90d81

src/components/label/label.tsx
jgroth: 055a5006361279989495db177ec9e7249e43956b
c8208209df90fe84f067d179ba7d8556f81a5935
ef70fe8f2f34c6cddff8ff37d47b7c7736f99e19

src/components/label/label.tsx
jgroth: 4acef8ba1e5df8df25d68bb4b0b8476246d42dcf

src/components/label/label.tsx
jgroth: I updated the descriptions but did not change the names

bc895e20bdc592e497d6de86744a4aaf7be96be0

Kiarokh: I'm OK if you wanna squash all these and merge them into my branch. I think we can continue from there more easily.

jgroth: We can't squash them here easily, lets just merge it to the other branch first

Kiarokh: sure. I just mean things are OK from my perspective

Issue #2874, opened 2024-04-02
limel-icon: add support for displaying an image by supplying a url


adrianschmidt:
## New feature motivation

There are loads of places in our existing components where it would be nice to be able to show an image, for example a user avatar. What these places have in common is that most, if not all, already support showing an icon, via `limel-icon`. By adding support for rendering an image element with a supplied url to limel-icon, we would get support from showing images in all these places, at a very low cost.

Re: https://github.com/Lundalogik/hack-tuesday/issues/345

## New feature description

Let `limel-icon` detect if the string supplied as the `name` parameter is a url. If it is, render an `image` element, with the `src` attribute set to the url. Otherwise, render the icon as usual.

adrianschmidt:
I guess this issue was vetoed in the PR discussion, so I'll close this.

PR #2875, opened 2024-04-02
limel-icon: add support for displaying images

adrianschmidt:
fix: #2874
fix: https://github.com/Lundalogik/hack-tuesday/issues/345

image

Result of playing around with this in the limel-chip-set composite example.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/icon/icon.tsx
```diff
@@ -46,7 +46,7 @@ export class Icon {
     public size: IconSize;
 
     /**
-     * Name of the icon
+     * Name of the icon, or URL of image to display
```
adrianschmidt, line 49: I can hear you sharpening your pitchforks from here… 😬 🙈

But seriously, if we add support for displaying images as a separate component, we'll have to add that _everywhere_. By just making the small changes made to this file*, we get image support everywhere that we already have support for icons.

*) The other changes are mostly adding tests and docs, and a little bit of refactoring.

jgroth: We could also add a new `src` attribute that would take precedence over `name`

adrianschmidt: Yeah, true, that would separate concerns a _little_ bit at least…

But I think we would still have to touch every single component that uses `limel-icon` to implement support for it in them, right?

I guess it might be worth doing that… 🤔 It would be a simple change in each place at least. Smaller than having to add code for checking if we have a name or a url and switching between rendering a limel-icon or a "limel-image" or something.

I can check out a separate branch, make that change in limel-icon, and then make the corresponding change in one or two components using limel-icon, to see what it would look like 👍

src/components/icon/icon.tsx
```diff
@@ -101,4 +102,11 @@ export class Icon {
             container.innerHTML = svgData;
         }
     }
+
+    private renderImg(src: string) {
+        const container = this.host.shadowRoot.querySelector('div.container');
+        if (container) {
+            container.innerHTML = `<img src="${src}" />`;
```
jgroth, line 109: I don't think we need this hack for regular images, we should be able to just render the element as usual :smile: We had to do this for the icons to fix broken SVG etc.

Also, `img` tags "require" the `alt` attribute

adrianschmidt: Yeah, I know, but it felt like the two might interfere with each other if I used different methods… 🤔… maybe they wouldn't? I'll have a look at it 👍

adrianschmidt: Nah, it's being annoying. If you first render an icon, then add an src to render an image instead, the icon isn't removed. I'm sure that's solvable, but I think it's likely easier to just use the same hack for both.

src/components/icon/icon.tsx
adrianschmidt: I've updated it to take a separate `src` attribute instead. It's definitely cleaner. But we don't have support "for free" in any of the components that are using limel-icon. I'll add support to limel-chip or something, so we can see what it takes.

adrianschmidt: It's fairly simple, but still quite a lot of work…

Time for me to log off now though. Will continue tomorrow.

src/components/icon/examples/icon-image.tsx
```diff
@@ -0,0 +1,148 @@
...
+/**
+ * Displaying an image instead of an icon
+ *
+ * Because the `limel-icon` component is used by several other components
+ * already, we have added the ability to display an image instead of an icon.
+ * This makes it possible to display a user avatar, for example, anywhere that
+ * an icon is used.
+ *
+ * To render an image, simply provide the URL to the image as the `name`
+ * property.
```
Kiarokh: this explanation is great for the PR or for this commit, but feels a bit weird for the consumers in the docs I'd say.
Maybe better to say something along these lines:

> If you can't find a proper icon, feel free to add the path to your icon, instead the icon name. You can do the same in any other component that uses the `Icon` interface.


Also, I assume that in this case, the `color` and `backgroundColor` properties won't work anymore, even if the path was pointing to an SVG file, since we're rendering an `<img ` tag. So maybe worth to mention it somewhere, or make sure that svgs are treated as inlined svgs.

src/components/icon/examples/icon-image.tsx
```diff
@@ -0,0 +1,148 @@
...
+ *
+ * Supported url formats are:
+ * - Absolute URL (beginning with `https://` or `http://`)
+ * - Relative URL (beginning with a single slash)
+ * - Protocol-relative URL (beginning with double slashes)
+ *
```
Kiarokh, line 14: do we need anything regarding file-size, canvas size, etc?

Kiarokh (approved):

src/components/icon/icon.tsx
Kiarokh: > `img` tags "require" the alt attribute

the `svg`s also require that to be honest.

src/components/icon/examples/icon-image.tsx
adrianschmidt: Yeah, I was a little conflicted about the description. I kind of wanted to defend adding this somewhat weird ability to `limel-icon`… 😅 Or, if not defend, then at least explain why it's there.

I've actually had the thought that we could check if the string provided ends in .svg, and if it does, inject it in the page the same way we do with our own icons. But before doing that, we'd have to check that doing so doesn't create any vulnerabilities that wouldn't be created anyway by letting people load external SVGs in an `img` tag. (If the latter also creates vulnerabilities, we might want to filter what types of images we allow.)

For now, I'm pretty sure I did mention somewhere that it's possible to give a URL to an SVG, but setting the colour through CSS won't work when you do that.

src/components/icon/examples/icon-image.tsx
adrianschmidt: Yeah, we might want to add something about file size.

By "canvas size", I assume you mean the image's dimensions? The image is resized to fit the size of `limel-icon`, and border-radius is applied to make it round. I guess we might want to add a line telling people to use square images 👍

src/components/icon/icon.tsx
adrianschmidt: > I actually liked using the `name` more than adding a new `src` prop or similar.

Yeah, I'm not sure which way I prefer myself yet. The component does become nicer when the properties are separate, but just being able to start using it anywhere that you can configure which icon to use, regardless of whether any layers in-between you and `limel-icon` have been updated, is a very powerful argument.

Ultimately, I don't think "the component is a little nicer this way" is a strong enough argument for separating it into two properties. What I'm more worried about is if we would be painting ourselves into a corner that we later come to regret if we use the `name` property for both.

jgroth: The consumer has to update their code regardless if they want to use an image, so changing `name` to `src` would be done at the same time. I guess it's only more work for us to make sure that the `src` attribute is compatible everywhere if we go down that road instead of relying on `name` for URLs

src/components/icon/examples/icon-image.tsx
adrianschmidt: Regarding treating external SVGs the same way we treat our own icons though, I realised we _absolutely cannot_ do that. If we did, an attacker could inject any content they wanted by simply adding it to a file ending in `.svg`, and we'd happily inject it for them! 😱

We could probably validate that the content is actually SVG and stuff, but that would make the whole thing a lot more complex. Being able to load external SVG icons was not in the scope for the task here, so I'm not going to focus on that.

src/components/icon/icon.tsx
adrianschmidt: > The consumer has to update their code regardless if they want to use an image

Well, yes and no. If a direct consumer wants to change their code so it uses an image, then yes, there's no difference. But there are two cases where there _is_ a difference:

1. If there is a consumer in-between, and that consumer hasn't updated their code, the end-consumer has to wait for them to do that.
2. If the icon is set in configuration, and the user can enter the name of the icon as free text, then they could immediately start using images without waiting for any code to be updated.

But you might be right. While we certainly have a lot of case 1, most of those instances are perhaps inside lime-elements anyway. And while case 2 remains true, that's not the primary target case for this feature anyway, since that is to allow displaying user avatars in place of icons, and that's programmatic use, not config. (Or at least it's not going to be config only.)

Alright, I'll continue on adapting our own uses of `limel-icon` then 👍 😅

src/components/icon/icon.tsx
jgroth: > > `img` tags "require" the `alt` attribute
> 
> the `svg`s also require that to be honest.

There's no `alt` attribute for the `svg` element in the spec

src/components/icon/icon.tsx
Kiarokh: If you wanna go that path, to update things everywhere, I'd say adding a new feature in those places is more proper. In the sense that you either say 
```tsx
icon: {
	name: '',
	color: '',
	etc: …
}
```

or use the `Image` interface and add 
image: {
	src: '',
	alt: '',
	etc: …
}
```

Why should be integrated in limel-icon then?

Kiarokh: I mean the need wasn't really "_being able to replace any icon with any image_". The need was "We wanna show an image instead of icon on the chip."

I'm not sure if this implementation is correct.

jgroth:
This PR is already quite big, so I just wanted to mention that we don't have to fix _everything_ in this PR to make it easier to review :smile:

src/components/icon/icon.tsx
adrianschmidt: > I'm not sure if this implementation is correct. In many places, you might still want to keep the icon, even if there is an image. This is what we did for instance for the feed items:
> image
> where the icon goes down since it still give meaning, and the image is replaced with the icon, visually.

Ok, sure, I see your point there. I would consider that a bit of a special case though. I'm sure we can still make those even if we have what's proposed here.

I'm a little unsure how to interpret you at this point though. Do you no longer see any usefulness from this feature then?

Kiarokh: Sorry if I sound harsh… I'm sure or I hope you understand that I am not criticizing your creative and great work and smart approach. I'm just trying to let you know what scares me if we go this way. This has benefits certainly, but I'm worried that it will kick us back very quickly and very hard.

We can still complete what you started, but by adding support for images in limel-chip and consequently in the consumers of  limel-chip. Maybe adding an `Image` interface that offers things that concern images, and using it in a few places is a better approach.

adrianschmidt:
> This PR is already quite big, so I just wanted to mention that we don't have to fix _everything_ in this PR to make it easier to review 😄

Good point 👍

I've started by breaking out the two first commits that are just doing some general boyscouting:
https://github.com/Lundalogik/lime-elements/pull/2881
https://github.com/Lundalogik/lime-elements/pull/2882

Those should be fine to review and merge regardless of what happens to the rest of this.

src/components/icon/icon.tsx
Kiarokh: I made this alternative PR. What do you think? https://github.com/Lundalogik/lime-elements/pull/2885

hannahu:
So this could be closed? Just to clarify when someone will review 😄

Kiarokh:
I think so. Reopen if you disagree

PR #2885, opened 2024-04-04
enable adding images to chips

Kiarokh:
fix https://github.com/Lundalogik/hack-tuesday/issues/345

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -73,6 +73,7 @@ export interface Chip<T = any> {
     // @deprecated
     iconTitle?: string;
     id: number | string;
+    image?: Image_2;
     removable?: boolean;
     selected?: boolean;
     text: string;
@@ -203,6 +204,7 @@ export namespace Components {
         "disabled": boolean;
         "icon"?: string | Icon;
         "identifier"?: number | string;
+        "image"?: Image_2;
         "invalid": boolean;
         "language": Languages;
         "link"?: Omit<Link, 'text'>;
@@ -1095,6 +1097,7 @@ namespace JSX_2 {
         "disabled"?: boolean;
         "icon"?: string | Icon;
         "identifier"?: number | string;
+        "image"?: Image_2;
         "invalid"?: boolean;
         "language"?: Languages;
         "link"?: Omit<Link, 'text'>;
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -861,6 +861,14 @@ export interface Icon {
 // @public (undocumented)
 export type IconSize = 'x-small' | 'small' | 'medium' | 'large';
 
+// @public
+interface Image_2 {
+    alt: string;
+    loading?: 'lazy' | 'eager';
+    src: string;
+}
+export { Image_2 as Image }
+
 // @public (undocumented)
 export interface InfoTileProgress {
     displayPercentageColors?: boolean;
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -59,6 +59,7 @@ const INPUT_FIELD_TABINDEX = 1;
  * @exampleComponent limel-example-chip-set-input-type-text
  * @exampleComponent limel-example-chip-set-input-type-search
  * @exampleComponent limel-example-chip-icon-color
+ * @exampleComponent limel-example-chip-set-image
  * @exampleComponent limel-example-chip-set-composite
  */
 @Component({
@@ -631,6 +632,7 @@ export class ChipSet {
             identifier: chip.id,
             text: chip.text,
             icon: chip.icon,
+            image: chip.image,
             badge: chip.badge,
             selected: chip.selected,
             disabled: this.disabled,
--- a/src/components/chip-set/chip.types.ts
+++ b/src/components/chip-set/chip.types.ts
@@ -1,3 +1,4 @@
+import { Image } from '../../global/shared-types/image.types';
 import { Icon } from '../../global/shared-types/icon.types';
 
 /**
@@ -18,6 +19,11 @@ export interface Chip<T = any> {
      * Name of the icon to use. Not valid for `filter`.
      */
     icon?: string | Icon;
+
+    /**
+     * A picture to be displayed instead of the icon on the chip.
+     */
+    image?: Image;
 
     /**
      * Color of the icon. Overrides `--icon-color`.
--- a/src/components/chip-set/examples/chip-image.tsx
+++ b/src/components/chip-set/examples/chip-image.tsx
@@ -1,0 +1,53 @@
+import { Component, h } from '@stencil/core';
+
+/**
+ * Chips with images
+ *
+ * You can use images instead of icons on chips.
+ *
+ * :::note
+ * The image will be displayed instead of the icon, if both are provided.
+ * :::
+ */
+@Component({
+    tag: 'limel-example-chip-set-image',
+    shadow: true,
+})
+export class ChipSetImageExample {
+    public render() {
+        return [
+            <limel-chip-set
+                value={[
+                    {
+                        id: 1,
+                        text: 'Kiarokh',
+                        icon: {
+                            name: 'honey_badger',
+                            color: 'rgb(var(--color-yellow-default))',
+                        },
+                        image: {
+                            src: 'https://lundalogik.github.io/lime-elements/2e86c284-d190-4c41-8da2-4de50103a0cd.png',
+                            alt: 'A picture of Kiarokh Moattar, Product Designer at Lime Technologies',
+                        },
+                    },
+                    {
+                        id: 2,
+                        text: 'Lucy',
+                        image: {
+                            src: 'https://lundalogik.github.io/lime-elements/780af2a6-d3d1-4593-8642-f03210d09271.png',
+                            alt: 'A picture of Lucy Chyzhova, UX Designer at Lime Technologies',
+                        },
+                    },
+                    {
+                        id: 3,
+                        text: 'Adrian',
+                        image: {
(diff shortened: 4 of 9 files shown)


src/global/shared-types/img.types.ts
```diff
@@ -0,0 +1,23 @@
+/**
+ * This interface is used to specify a path to an image,
+ * along with related properties, like alt text.
+ * @public
+ */
+export interface Img {
```
Kiarokh, line 6: should the interface be called `Img` or `Image`?

The HTML element would be an `` so in this sense, the other attributes such as `src` and `alt` would be more meaningful for the consumer perhaps.

src/components/chip-set/chip.types.ts
```diff
@@ -19,6 +20,11 @@ export interface Chip<T = any> {
      */
     icon?: string | Icon;
 
+    /**
+     * A picture to be displayed instead of the icon on the chip.
+     */
+    image?: Img;
+
```
Kiarokh, line 27: I never quite understood why we are repeating some of these props in `limel-chip` again. But just like `icon`, I repeated this too.

src/interface.ts
```diff
@@ -50,3 +50,4 @@ export * from './components/tab-panel/tab-panel.types';
 export * from './components/table/table.types';
 export * from './global/shared-types/separator.types';
 export * from './global/shared-types/icon.types';
+export * from './global/shared-types/img.types';
```
Kiarokh, line 53: also not sure if we need to export this

src/components/chip/chip.tsx
```diff
@@ -251,13 +259,19 @@ export class Chip implements ChipInterface {
...
         const icon = getIconName(this.icon);
 
-        if (!icon) {
+        if (!icon && !this.image) {
             return;
         }
 
+        if (this.image) {
+            return (
+                <img src={this.image.src} alt={this.image.alt} loading="lazy" />
```
Kiarokh, line 271: Based on what I understand from the cryptic descriptions on [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#referrerpolicy), it's better/safer if we add `referrerpolicy="no-referrer"` to the image.
```tsx
<img src={this.image.src} alt={this.image.alt} loading="lazy"  referrerpolicy="no-referrer"/>
```

However, adding it like that in the code will show a ts error:
```
Type '{ src: string; alt: string; loading: "lazy"; referrerpolicy: string; }' is not assignable to type 'ImgHTMLAttributes<HTMLImageElement>'.
  Property 'referrerpolicy' does not exist on type 'ImgHTMLAttributes<HTMLImageElement>'.ts(2322)
```

src/components/chip-set/examples/chip-image.tsx
```diff
@@ -0,0 +1,53 @@
...
+                            src: 'https://lundalogik.github.io/lime-elements/780af2a6-d3d1-4593-8642-f03210d09271.png',
+                            alt: 'A picture of Lucy Chyzhova, UX Designer at Lime Technologies',
+                        },
+                    },
+                    {
+                        id: 3,
+                        text: 'Adrian',
+                        image: {
+                            src: 'https://lundalogik.github.io/lime-elements/0e6f74c0-11d9-465b-aac6-44f33da3cb7c.png',
+                            alt: 'A picture of Adrian Schmidt, Head of Smooth Operations at Lime Technologies',
```
Kiarokh, line 46: 😜

src/components/chip-set/chip-set.tsx
```diff
@@ -628,6 +629,7 @@ export class ChipSet {
             identifier: chip.id,
             text: chip.text,
             icon: chip.icon,
+            image: chip.image,
```
Kiarokh, line 632: not sure why this broke the composite example 
image

src/components/chip-set/chip-set.tsx
```diff
@@ -59,6 +59,7 @@ const INPUT_FIELD_TABINDEX = 1;
  * @exampleComponent limel-example-chip-set-input-type-text
  * @exampleComponent limel-example-chip-set-input-type-search
  * @exampleComponent limel-example-chip-icon-color
+ * @exampleComponent limel-example-chip-set-image
```
Kiarokh, line 62: in a followup PR, we can use the new `Img` interface and enable images in `limel-list` as well, which will enable us to add pictures in the list of suggestions of `limel-picker`

src/components/chip/chip.tsx
Kiarokh, line 271: are we or should we be concerned about svgs with embedded scripts, and generally all the bad things people can do with images?

src/global/shared-types/img.types.ts
jgroth: It should be `Image` :smile: We should avoid abbreviations

src/global/shared-types/img.types.ts
```diff
@@ -0,0 +1,23 @@
...
+    /**
+     * The path to the image file.
+     */
+    src: string;
+
+    /**
+     * The alternative text of the image, used to improve accessibility.
+     */
+    alt?: string;
```
jgroth, line 15: Should this not be required?

src/components/chip/chip.tsx
```diff
@@ -104,6 +106,12 @@ export class Chip implements ChipInterface {
     @Prop()
     public icon?: string | Icon;
 
+    /**
+     * A picture to be displayed instead of the icon on the chip.
+     */
+    @Prop()
+    public image?: Img;
```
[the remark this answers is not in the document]
adrianschmidt: `image`

src/components/chip-set/chip.types.ts
adrianschmidt: I'm not sure I understand what repetition you are referring to. Are you talking about defining `image?: Img;` here and defining `@Prop() public image?: Img;` in `src/components/chip/chip.tsx`? Or something else? 😊

src/components/chip/chip.tsx
adrianschmidt: If loading an external image in an `img` element can be an attack vector, I guess we need to consider it at least. I didn't get so far as to investigate that. I would hope that embedded scripts in svgs wouldn't be executed, or at least not executed in a way that gains any access to the rest of the page…? 😬

src/components/chip-set/chip.types.ts
Kiarokh: yes exactly.

src/global/shared-types/img.types.ts
jgroth: It's required on the `img` element. Making it required would be to "highly recommend" it since if a user dont want it they will simply put an empty string in there :smile:

src/components/chip-set/chip.types.ts
adrianschmidt: That's because the definition here defines what the "data object" representing a chip should look like, while the definition in `src/components/chip/chip.tsx` defines what properties the component _rendering_ a chip accepts.

It's no coincidence that there's an overlap there… 😄

src/interface.ts
adrianschmidt: Yes we do. Otherwise people can't use the type in the consuming code.

src/components/chip-set/chip.types.ts
Kiarokh: I still don't understand why we just didn't `extend` the Chip types from chipset, inside `limel-chip`.

We do similar things for instance here:

https://github.com/Lundalogik/lime-elements/blob/7fe1fbbd5af68cb064118af75e3c6985b881831e/src/components/progress-flow/progress-flow.types.ts#L7

jgroth: If I understand you correctly, you want to do something like this inside the component?
```ts
export class Chip extends ChipInterface
```

That will not work since the component is a class and the chip type is an interface. Classes can only extend other classes, and instead need to implement an interface or a type. We could in theory create a class for the chip as well, but Stencil does not allow components to extend other classes and we would still have to repeat all the properties so we could decorate them with `@Prop`

src/components/chip/chip.tsx
Befkadu1: The `img` interface doesn't have a property called `referrerpolicy`. It doesn't seem we can implement that. 

```
   interface ImgHTMLAttributes<T> extends HTMLAttributes<T> {
        alt?: string;
        decoding?: 'async' | 'auto' | 'sync';
        importance?: 'low' | 'auto' | 'high';
        height?: number | string;
        loading?: 'lazy' | 'auto' | 'eager';
        sizes?: string;
        src?: string;
        srcSet?: string;
        srcset?: string;
        useMap?: string;
        usemap?: string;
        width?: number | string;
    }

**Review by Befkadu1 — APPROVED** [review]
Works as it should!

**Comment by Befkadu1** [comment]
Maintainers approval needed @adrianschmidt @jgroth

File: etc/lime-elements.api.md [hunk 10]
```diff
@@ -861,6 +861,14 @@ export interface Icon {
 // @public (undocumented)
 export type IconSize = 'x-small' | 'small' | 'medium' | 'large';
 
+// @public
+interface Image_2 {
+    alt: string;
+    loading?: 'lazy' | 'eager';
+    src: string;
+}
+export { Image_2 as Image }
```
adrianschmidt, line 872: This file wasn't updated in the first commit, so the commits weren't atomic. I fixed that for you this time 😊

adrianschmidt (approved):
Changes to external API approved.

Issue #1858, opened 2022-10-10
improve accessibility of limel-tooltip by displaying it when receiving keyboard focus


Kiarokh:
## Current behavior
The tooltip becomes visible in response to a mouse hover. 
But unfortunately not after the owning element receives keyboard focus, currently.

## Expected behavior
The tip should become visible when owning element receives keyboard focus (in addition to hover). To make the component keyboard accessible.

Read more: https://www.w3.org/TR/wai-aria-1.1/#tooltip

## Environment

- **lime-elements** version: 
- Framework used: 
- Logs:

Kiarokh:
@reviewer you may be interested in reading this too:

https://github.com/Lundalogik/crm-feature/issues/3502#issuecomment-2037258158

PR #2888, opened 2024-04-04
fix(tooltip): display the tooltip, when the element is focused using …

Kiarokh:
…keyboard

Previously the tooltip was shown only when the element was hovered using mouse. This is an undesirable accessibility issue for sighted users who use keyboard to navigate the interface.

fix https://github.com/Lundalogik/lime-elements/issues/1858

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/tooltip/tooltip.tsx
+++ b/src/components/tooltip/tooltip.tsx
@@ -153,13 +153,15 @@ export class Tooltip {
     private addListeners() {
         this.ownerElement?.addEventListener('mouseover', this.showTooltip);
         this.ownerElement?.addEventListener('mouseout', this.hideTooltip);
-        this.ownerElement?.addEventListener('click', this.hideTooltip);
+        this.ownerElement?.addEventListener('focus', this.showTooltip);
+        this.ownerElement?.addEventListener('blur', this.hideTooltip);
     }
 
     private removeListeners() {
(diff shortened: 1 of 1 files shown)


src/components/tooltip/tooltip.tsx
```diff
@@ -153,13 +153,15 @@ export class Tooltip {
     private addListeners() {
         this.ownerElement?.addEventListener('mouseover', this.showTooltip);
         this.ownerElement?.addEventListener('mouseout', this.hideTooltip);
-        this.ownerElement?.addEventListener('click', this.hideTooltip);
```
Kiarokh, line 156: The `click` event listener was removed because it might interfere with the desired behavior of showing the tooltip on `focus` and hiding it on `blur`.

If a user clicks on the trigger element, it will also receive `focus`. If the tooltip is set to hide on `click`, it would immediately disappear after appearing, which is probably not what we want.

Befkadu1 (approved):

PR #2896, opened 2024-04-09
Make text editor form component

LawrenceBorst:
Make the text editor a form component

Use plugins to add readonly, disabled, and helpertext modes

fix: Lundalogik/crm-feature#4039

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


etc/lime-elements.api.md
```diff
@@ -1551,6 +1563,15 @@ namespace JSX_2 {
         "onChangeTab"?: (event: LimelTabPanelCustomEvent<Tab>) => void;
         "tabs"?: Tab[];
     }
+    // @beta
+    interface LimelTextEditor {
+        "disabled"?: boolean;
+        "helperText"?: string;
```
LawrenceBorst, line 1569: Talked to Lucy, at present `helpertext` has been confused with `placeholder`

LawrenceBorst: Wherever it says `helpertext` read `placeholder` instead 😅

Kiarokh: not anymore. i changed them all

src/components/text-editor/text-editor.tsx
```diff
@@ -41,12 +100,43 @@ export class TextEditor {
...
+            this.view.state.tr.setMeta(this.placeholderKey, this.placeholder),
+        );
+    }
+
+    @Watch('readonly')
+    public watchReadonly() {
+        this.view.dispatch(
+            this.view.state.tr.setMeta(this.readonlyKey, this.readonly),
+        );
+    }
```
Kiarokh, line 128: Not sure why you are doing this. But I think it's better to just return `limel-markdown` when `readonly="true"`

src/components/text-editor/examples/text-editor-basic.tsx
```diff
@@ -1,4 +1,8 @@
-import { LimelTextEditorCustomEvent } from '@limetech/lime-elements';
+import {
+    LimelCheckboxCustomEvent,
+    LimelInputFieldCustomEvent,
```
Kiarokh, line 3: 🤔 we never do this. Are these needed?

adrianschmidt: I don't know why you think we never use the actual type for events…? I searched the lime-elements code for `CustomEvent` in VSCode, and while we do use `CustomEvent` in some examples, from the ten-or-so I checked it looks like we import the actual type in the majority of cases.

Is there any particular reason you would prefer that we didn't? 🙂

src/components/text-editor/text-editor.tsx
```diff
@@ -55,22 +99,51 @@ export class TextEditor {
...
+        if (this.value) {
+            this.view.dom.innerHTML = this.value.html;
+        }
     }
+
+    private getHTML = (): string => {
+        if (
+            this.view.dom.textContent === '' ||
+            (this.view.dom.textContent === this.placeholder &&
+                this.view.dom.innerHTML !== `<p>${this.placeholder}</p>`) // TODO: this is a little too coupled to the placeholder plugin
```
adrianschmidt, line 133: So, there is a plugin for adding a placeholder, but we still have to do stuff like this and `this.view.dispatch( this.view.state.tr.setMeta(this.placeholderKey, this.placeholder) )`?

I think this is way too much bending-over-backwards to support a placeholder. If we can't do it in a neater way, I think we should skip it.

LawrenceBorst: I think we should always do a dispatch since I think it's the only way we can communicate state changes from the web component to the editor (plugin) state 🤔.

But I also don't really like the `textContent` and `innerHTML` logic here. The last condition avoids emitting an empty string value when the value happens to be the same as the placeholder. I think there is probably a better way to avoid coupling the plugin to the generic transaction handler.

src/components/text-editor/text-editor.tsx
LawrenceBorst: I'm not sure, but I worry that we then need complexity in some other place to hide and unhide the editor. I like this 
separation of concerns, as here we encapsulate editor-specific behaviour as an editor plugin, instead of doing it from within the consumer which I think should mostly just mediate between editor state changes and render the editor. Any thoughts @adrianschmidt ?

src/components/text-editor/examples/text-editor-basic.tsx
Kiarokh: No. Just checked other examples. If you say it's better, then we do

PR #2902, opened 2024-04-11
fix(markdown): better heading styles

Kiarokh:
These styles are used when markdown is renderer.
Previously headings were being rendered too large.
The result was not intended for rendering a web-page or
a presentation, where large headings improves
 the presentation.
It was rather small blocks of text, for instance
representing or rendering content of "email",
or an "article".

fix https://github.com/Lundalogik/crm-feature/issues/4033

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/markdown/partial-styles/_body-text.scss
+++ b/src/components/markdown/partial-styles/_body-text.scss
@@ -9,12 +9,15 @@
 }
 
 p,
-a,
 li {
     font-size: 0.875rem;
     word-break: break-word;
     hyphens: auto;
     -webkit-hyphens: auto;
+}
+
+a {
+    word-break: break-all;
 }
 
 p {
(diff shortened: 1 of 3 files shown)


src/components/markdown/partial-styles/_body-text.scss
```diff
@@ -12,6 +12,9 @@ p,
 a,
 li {
     font-size: 0.875rem;
+    word-break: break-word;
```
LucyChyzhova, line 14: Doesn't seems to work when it's a Heading

https://github.com/Lundalogik/lime-elements/assets/50618208/c98d97ee-1323-492f-9dd5-9b0d4d49eeed

Kiarokh: 🤗 great catch! I added a fixup!

LucyChyzhova (approved):
Works good!

PR #2906, opened 2024-04-11
Text editor as form component

LawrenceBorst:
Fixes Lundalogik/crm-feature#4039

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


etc/lime-elements.api.md
Kiarokh, line 1: I think the commit title can be change from 
`feat(text-editor-form-component): add form wrapper for text editor`

to 
```
feat(text-editor): enable using it as a standard input field

By offering the same standard input field props,
as well as integrating it into limel-form.
```

The consumers don't need to know internal names and such.
Also, all of these changes are ultimately about the `limel-text-editor`.

src/components/text-editor/text-editor.tsx
Kiarokh, line 1: commit title should change to:
`feat(text-editor): add `value` prop`

src/components/text-editor/text-editor.tsx
```diff
@@ -37,60 +13,84 @@ import { getFilteredMenu } from './menu/menu-filter';
     shadow: true,
     styleUrl: 'text-editor.scss',
 })
-export class TextEditor {
+export class TextEditor implements FormComponent<{ html: string }> {
```
jgroth, line 16: I'm not sure this is going to play nice if you want to place this component on a card etc. It would basically require you to write a wrapper component. I think we should change this to `FormComponent<string>` to make it easier to use

src/components/text-editor/text-editor.tsx
```diff
@@ -37,60 +13,84 @@ import { getFilteredMenu } from './menu/menu-filter';
...
+     */
+    @Event()
+    public change: EventEmitter<{ html: string }>;
 
-                    this.change.emit({ html: this.view.dom.innerHTML });
-                },
-            },
+    public render() {
+        return (
+            <limel-prosemirror-adapter
```
jgroth, line 84: Will this always emit HTML? What about markdown? Markdown takes considerably less space when stored in the database and I think we should default to that instead of HTML. We could have a prop that decides what mode to use if someone really wants HTML

src/components/text-editor/text-editor.tsx
```diff
@@ -1,34 +1,10 @@
...
+import { EditorButton } from './menu/types';
 /**
- * This editor offers a rich text editing experience with markdown support,
- * in the sense that you can easily type markdown syntax and see the rendered
- * result as rich text in real-time. For instance, you can type `# Hello, world!`
- * and see it directly turning to a heading 1 (an `<h1>` HTML element).
- *
- * Naturally, you can use standard keyboard hotkeys such as <kbd>Ctrl</kbd> + <kbd>B</kbd>
- * to toggle bold text, <kbd>Ctrl</kbd> + <kbd>I</kbd> to toggle italic text, and so on.
+ * See `prosemirror-adapter` for more information
```
jgroth, line 5: The adapter component is private so a user will never be able to read anything for more information :smile:

src/components/text-editor/text-editor.tsx
LawrenceBorst: @civing What do you think?

src/components/text-editor/text-editor.tsx
LawrenceBorst: Fixed in another PR, set to string and implicitly emits HTML. But there is still the question of whether we should emit HTML or markdown 🤔

jgroth: Ah, I got notified on so many PRs so maybe I didn't start in the right end :smile:

jgroth: Markdown supports HTML as well, but not the other way around, so I think markdown is better. That way you could have both and still only have one mode

LawrenceBorst: I will take care of this in a followup PR

src/components/text-editor/text-editor.tsx
civing: It would be nice if we stick to one or the other, if we stick to markdown then the consumer could transform the output to HTML of they have a specific need for that? I'm thinking about emails.

jgroth: If we emit markdown, it can be used with the markdown component out of the box to render the content.
If we emit HTML, we don't really have a similar component to use. It should work with the markdown component I think, but we might also unconsciously encourage consumers to use `element.innerHTML` which is really bad

civing (approved):

src/components/text-editor/text-editor.tsx
civing: Let's take care of this in a follow up PR

src/components/text-editor/text-editor.tsx
```diff
@@ -34,54 +13,77 @@ import { getFilteredMenu } from './menu/menu-filter';
...
+    public invalid?: boolean;
 
-        const menu: MenuElement[][] = buildFullMenu(mySchema)
-            .map((items) => getFilteredMenu(items, undefined))
-            .filter((items) => items.length);
+    /**
+     * Description of the text inside the editor
+     */
+    @Prop({ reflect: true })
+    public value: { html: string };
```
adrianschmidt, line 68: Time to change this one?

```suggestion
    public value: string;
```

civing: It's planned in this follow up PR https://github.com/Lundalogik/lime-elements/pull/2918

src/components/text-editor/text-editor.tsx
adrianschmidt: The limel-markdown component ought to work just fine for sanitising html. But yes, we will probably need to educate people about why they need to run any user generated html through limel-markdown, and not just use `element.innerHTML`.

adrianschmidt (approved):
Changes to the public API approved.

PR #2917, opened 2024-04-17
Export helper functions `getIconForFile`, `getIconFillColorForFile`, and `getIconBackgroundColorForFile`

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/file/icons.ts
+++ b/src/components/file/icons.ts
@@ -124,6 +124,15 @@ const filetypeIconTable: Record<string, string> = {
     dbf: DATA_ICON,
 };
 
+/**
+ * Takes a file extension as argument and returns the corresponding icon name.
+ * Uses the names from the private icon package `lime-icons8`. If you are using
+ * a different icon package, this functions is probably not useful for you.
+ *
+ * @param extension - The file extension (without the dot).
+ * @returns The name of the icon to use.
+ * @public
+ */
 export function getIconForFile(extension: string): string {
     return filetypeIconTable[extension.toLowerCase()] || DEFAULT_ICON;
 }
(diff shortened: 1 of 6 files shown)


src/components/file/icons.ts
```diff
@@ -124,6 +124,15 @@ const filetypeIconTable: Record<string, string> = {
...
+/**
+ * Takes a file extension as argument and returns the corresponding icon name.
+ * Uses the names from the private icon package `lime-icons8`. If you are using
+ * a different icon package, this functions is probably not useful for you.
+ *
+ * @param extension - The file extension (without the dot).
+ * @returns The name of the icon to use.
+ * @public
+ */
 export function getIconForFile(extension: string): string {
```
adrianschmidt, line 136: @Kiarokh @jgroth:

@alexwero wants to use this function in an add-on. I don't see any reason we shouldn't export these helpers. Do you?

jgroth (commented):
So, does this actually work now? :smile: I know it has been an issue in the past that you cannot export anything else than components and types from a Stencil project, but perhaps that has been fixed now? :thinking:

adrianschmidt:
> > So, does this actually work now? 😄 I know it has been an issue in the past that you cannot export anything else than components and types from a Stencil project, but perhaps that has been fixed now? 🤔
> 
> Oh… hehe… good point… maybe I should try that 😆

I'm not sure if the problem is with exporting it or with importing it, but yes, stuff seems to break when I try 😭

image

I guess we'd have to extract those helper functions to yet another package to solve this then? 😣

PR #2922, opened 2024-04-22
feat(markdown): allow `height` and `width` in inline `style` attribute

adrianschmidt:
fix: Lundalogik/limepkg-email#695

This can be tested by adding the following snippet to the composite example for `limel-markdown` in the docs for this PR and compare to the result when using the docs for the latest release.

```html
<img
    border="0"
    style="width: 0.3229in; height: 0.3229in;"
    src="https://lundalogik.github.io/lime-elements/versions/latest/favicon.svg"
    alt="signature_2123812728"
/>
<br>
<br>
<img
    border="0"
    width="31"
    height="31"
    src="https://lundalogik.github.io/lime-elements/versions/latest/favicon.svg"
    alt="signature_2123812728"
/>
```

The result should look like this:

Both images in the screenshot are approximately 31 by 31 pixels

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/markdown/allowed-css-properties.ts
+++ b/src/components/markdown/allowed-css-properties.ts
@@ -3,10 +3,12 @@ export const allowedCssProperties = [
     'color',
     'font-style',
     'font-weight',
+    'height',
     'text-decoration-color',
     'text-decoration-line',
     'text-decoration-skip-ink',
     'text-decoration-style',
     'text-decoration-thickness',
     'text-decoration',
+    'width',
 ];
--- a/src/components/markdown/markdown-parser.ts
+++ b/src/components/markdown/markdown-parser.ts
@@ -42,7 +42,7 @@ export async function markdownToHTML(
             // Allow the `style` attribute on all elements
             attributes: {
                 ...defaultSchema.attributes,
-                '*': ['style'],
+                '*': ['height', 'style', 'width'],
             },
         })
         .use(() => {


src/components/markdown/allowed-css-properties.ts
```diff
@@ -3,10 +3,12 @@ export const allowedCssProperties = [
...
     'font-style',
     'font-weight',
+    'height',
     'text-decoration-color',
     'text-decoration-line',
     'text-decoration-skip-ink',
     'text-decoration-style',
     'text-decoration-thickness',
     'text-decoration',
+    'width',
```
Kiarokh, line 13: I didn't test. But does it mean it works on both of these, inside the `style` tag and if they are directly tags?
```
    width="31"
    height="31"
    style="width: 31px; height: 31px;"
```

adrianschmidt: Ah, good point!

No, it doesn't. This change only adds them as allowed properties inside the `style` attribute. I just checked, and `height` and `width` added as properties are filtered out. I'll add a fixup!

src/components/markdown/markdown-parser.ts
```diff
@@ -42,7 +42,7 @@ export async function markdownToHTML(
             // Allow the `style` attribute on all elements
             attributes: {
                 ...defaultSchema.attributes,
-                '*': ['style'],
+                '*': ['height', 'style', 'width'],
```
adrianschmidt, line 45: Here's the fixup @Kiarokh. Since the fix was in a separate file, and the whole PR is so small, I squashed and force-pushed, rather than pushing a separate fixup.

src/components/markdown/allowed-css-properties.ts
adrianschmidt: ⚡

adrianschmidt:
@FredrikWallstrom Nice to hear!

FYI: I was considering whitelisting `max-height` and `max-width` too, but then thought that it's probably better if we can keep those for our own use. If a css property is set in an inline `style` attribute, that overrides any value set in a stylesheet. So by not allowing `max-height` and `max-width`, it should be possible for us to set those values ourselves, and thereby enforcing certain limitations on the content of the activity items. I just want to put that info in the back of your mind, so if you were to run into trouble with image sizes in emails, you might have an idea of why. My hope is that you won't need those values whitelisted, but if it turns out you do, it's simple to add 👍

adrianschmidt:
Anyone up for approving this PR then? 😊

FredrikWallstrom (approved):

PR #2926, opened 2024-04-23
Text editor emit value as markdown

LawrenceBorst:
Fixes Lundalogik/crm-feature#4075

How to test:
I would, in the `limel-example-prosemirror-adapter-basic` example, pass in a content type and see what value gets emitted. I would also pass in a value there to see if it gets parsed correctly depending on the content type.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/text-editor/utils/html-converter.ts
```diff
@@ -0,0 +1,21 @@
...
+        return new Promise((resolve) => {
+            resolve(text);
+        });
+    };
+
+    public serialize = (view: EditorView): string => {
+        if (view.dom.textContent === '') {
+            return '';
+        } else {
+            return view.dom.innerHTML;
```
civing, line 18: I don't think it's good practice to read directly from the `innerHTML` of prosemirror.

civing (approved):

civing (approved):

jgroth (approved):

PR #2933, opened 2024-04-25
Make text editor work in webclient

LawrenceBorst:
Fixes Lundalogik/crm-feature#4056 and https://github.com/Lundalogik/crm-feature/issues/4112

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
+++ b/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
@@ -152,7 +152,7 @@ export class ProsemirrorAdapter {
         this.menuCommandFactory = new MenuCommandFactory(mySchema);
 
         if (this.value) {
-            this.view.dom.innerHTML = this.value;
+            this.updateView(this.value);
         }
     };
 
@@ -169,14 +169,6 @@ export class ProsemirrorAdapter {
 
         this.view.dispatch(tr);
     }
-
-    private getHTML = (): string => {
-        if (this.view.dom.textContent === '') {
-            return '';
-        } else {
-            return this.view.dom.innerHTML;
-        }
-    };
 
     private handleActionBarItem = (event: CustomEvent<ActionBarItem>) => {
         event.preventDefault();
--- a/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
+++ b/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
@@ -5,6 +5,7 @@ import {
     EventEmitter,
     Prop,
     State,
+    Watch,
     h,
 } from '@stencil/core';
 import { EditorState } from 'prosemirror-state';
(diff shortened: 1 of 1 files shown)


src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -50,6 +51,13 @@ export class ProsemirrorAdapter {
     @Event()
     private change: EventEmitter<string>;
 
+    @Watch('value')
+    public watchValue(newValue: string) {
+        if (newValue === '') {
+            this.view.dom.innerHTML = '';
```
adrianschmidt, line 57: I think we might want to leave a note here, for our future selves 😊

```suggestion
            // Note!
            // This is only safe because we are setting `innerHTML` to an empty
            // string. If refactoring, please make sure this isn't accidentally
            // turned into something like `this.view.dom.innerHTML = newValue`.
            this.view.dom.innerHTML = '';
```

`fixup! feat(text-editor): add watcher for value prop`

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
adrianschmidt, line 1: I'd probably make this a `fix` myself, rather than a `feat`, but I guess it's a matter of opinion.

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -50,6 +51,17 @@ export class ProsemirrorAdapter {
     @Event()
     private change: EventEmitter<string>;
 
+    @Watch('value')
+    public watchValue(newValue: string) {
+        if (newValue === '') {
```
civing, line 56: I think we should handle not this specific case but all the cases when the consumer might want to change the value of the component. It can be loading in a new template or do other things programatically in the consumer that changes the content of the text editor outside of prosemirror.

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -67,6 +68,15 @@ export class ProsemirrorAdapter {
...
     private change: EventEmitter<string>;
 
+    @Watch('value')
+    protected watchValue(newValue: string) {
+        if (!this.view) {
+            return;
+        }
+
+        this.updateView(newValue);
+    }
```
john-traas, line 78:
```suggestion
    @Watch('value')
    protected watchValue(newValue: string) {
        if (!this.view) {
            return;
        }

        if (this.view.state.doc.textContent === newValue) {
            return;
        }

        if (newValue === '') {
            this.updateView(newValue);
        }
    }
```

If we simply update the view each time the value is changed we get some odd behaviours. One of them is that the `spacebar` no longer works, or rather that it does work but we reset the editor view and the space is not added.

I think we need to iterate on this, but for now we need to be able to clear the text editor by passing in an empty string.

john-traas (dismissed):
It works with the latest fixup. 

We still might need to handle various other cases, but the current implementation will update the view with the value (if passed in) when the component loads, and if an empty string is passed in the component will also clear.

john-traas:
@LawrenceBorst @civing is there anything currently holding up the approval of this PR?

LawrenceBorst:
> @LawrenceBorst @civing is there anything currently holding up the approval of this PR?

As far as I know not :)

john-traas (approved):

PR #2935, opened 2024-04-25
text editor height

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/text-editor/examples/text-editor-height.tsx
```diff
@@ -0,0 +1,42 @@
...
+ * Depending on the `overflow` property of the container,
+ * you may get undesired rendering, or parts of the content
+ * can become invisible for the end-users.
+ *
+ * You can test this by resizing this example (which will give it a hardcoded
+ * `width` and `height`), and then typing several lines of text!
+ *
+ * Therefore, you need to make sure that `--text-editor-max-height`
+ * is set to a value which correlates with the hardcoded height of the container.
+ * :::
```
Kiarokh, line 31: I tried really hard to force the editor to inherit its height from its parent, and never grow larger than its parent. However, it was a huge 😖  in the 🍑  and I couldn’t make it behave that way. There are too many nested containers and many ifs and buts. So the best way I could come up with is allowing the consumer to set a max-height in case they want to stop the editor from resizing after a certain point.

I think it’s a fair solution, but it should also be tested in the product, to make sure we get the expected behavior, for instance when the container is a grid child, or flex child, or has overflow, etc….

Kiarokh:
Don't approve! I might have solved this

src/components/text-editor/examples/text-editor-height.tsx
Kiarokh: Don't approve! I might have solved this

src/components/text-editor/examples/text-editor-height.tsx
```diff
@@ -7,28 +7,14 @@ import { Component, h } from '@stencil/core';
...
- * Depending on the `overflow` property of the container,
- * you may get undesired rendering, or parts of the content
- * can become invisible for the end-users.
- *
- * You can test this by resizing this example (which will give it a hardcoded
- * `width` and `height`), and then typing several lines of text!
- *
- * Therefore, you need to make sure that `--text-editor-max-height`
- * is set to a value which correlates with the hardcoded height of the container.
- * :::
```
Kiarokh: @adrianschmidt as you can see in this last commit, I have removed a bunch of stuff. 

What is left of it is what is proper implementation according to me. However, there are scenarios in which this fails, and there is no logically explanation for it. Like the recording below:

https://github.com/Lundalogik/lime-elements/assets/35954987/5d3fbcd4-5daf-4bb1-90ad-6e4e464dbbb1

Another weird scenario is where you:
1. remove the `resize` and `overflow` possibilities from the example, 
2. and instead give the host in the example a `max-height`.
3. type multi lines of text in the editor, and it grows/overflows out of its own container and there is no way to get it fixed.

My best guess is that ProesMirror is doing something to the container using javascript. I don't see any inlined-styles, dynamically applied. However, I'm quite sure that something is going on in the background, aligning or expanding the divs, that we can't see.

So if you also can't find a logic, we can drop the last commit and go with what I did up-to that point.
What do you say?

adrianschmidt: Ah, I looked at the video, and it looks like the editor adapts to the container element again when you resize the container element, right? That's interesting… I wonder if that can be used in some way 🤔

src/components/text-editor/text-editor.tsx
adrianschmidt, line 1: Oh… hmm… the menu bar scrolls away with the text content of the editor! 😲

https://github.com/Lundalogik/lime-elements/assets/154725/be9b8c7e-8275-4b3a-a3f2-715495dfb0e9

src/components/text-editor/examples/text-editor-height.tsx
adrianschmidt: > This also makes me wonder whether we should enable end users to resize the height of the text editor or not?

Yes, we should. I _HATE_ it when some bloody developer thinks they know better than me how much space I'll need, and decides to prevent me from resizing the textarea I'm trying to write something in! 😤 I can't count the number of times I've used the dev-tools to change that behaviour on some web page I was using… 🤮

adrianschmidt: If faced with the choice, I'd rather have the height adjustable by the user, and _not_ have it auto-expand based on content, then have it auto-expand, but then not be able to resize it manually.

src/components/text-editor/text-editor.tsx
Kiarokh: That's the actionbar that @john-traas  was working on. Did that got merged? :O 

Then I'll have to remove a bunch of styles form this branch.

Kiarokh: I'll have to do this next week, and then get back to the discussion here.

src/components/text-editor/examples/text-editor-height.tsx
Kiarokh: if the text-editor too should be resizable, then we will probably have even more problems with the UI 😅 
I'll update the branch now that their strange behaving menubar is gone, and see what surprises that one can bring

Kiarokh: > Ah, I looked at the video, and it looks like the editor adapts to the container element again when you resize the container element, right? That's interesting… I wonder if that can be used in some way 🤔

Yes, this is why I thought something more dynamic should be going on in the background

Kiarokh:
@LucyChyzhova I think I know why the issue appears now. I will have to refactor some parts, and remove the wrapping `<fieldset />`. That was seemingly the problem. It behaves in an illogical way, which made it so hard for me to understand what the reason is. Additionally, the new change with `limel-action-bar` made things slightly harder to track, since now I had to refactor more stuff, before investigating what's wrong.

The `<fieldset />` also makes it impossible to add "resizability" to the component, which might be a desired thing in some scenarios.

So I decided to drop the last commit that allowed the consumers to add a desired `max-height`, _(even though that was the only way to make this work)_, 

So I decided to leave this for review up to this stage. Then refactor the component, and get rid of the `<fieldset />` and make the component fully expand and shrink according to its container instead. The amount of change will big considerable. So it's better to do it in a follow up PR.


Please review this, and I'm aware of this existing issue:

https://github.com/Lundalogik/lime-elements/assets/35954987/16c41c57-037b-4e81-b399-2941d6eb6d5e

this will be fixed in the next PR

Befkadu1:
It looks we don't have min width and min height. Is it intended? The other Issue I was about to say was mentioned in these [conversations](https://github.com/Lundalogik/lime-elements/pull/2935#discussion_r1581000378)


https://github.com/Lundalogik/lime-elements/assets/22968497/61365fb6-29cc-4e54-ae1a-499327890c95

Kiarokh:
> It looks we don't have min width and min height. Is it intended? The other Issue I was about to say was mentioned in these [conversations](https://github.com/Lundalogik/lime-elements/pull/2935#discussion_r1581000378)


Good catch. I'll fix these in the next PR. I've changed this so many times, so I don't wanna go back there. I'm gonna refactor it all. So no point in fixing it here

Befkadu1 (approved):

Befkadu1 (approved):

PR #2937, opened 2024-04-26
feat(text editor): Add custom menu

john-traas:
With this PR we use our action bar to create the menu items.

Commands for the various actions tied to those menu items are created and dispatched to the Editor via transactions. 

This decouples us from the built-in limitations of Prose-Mirror's menu enabling us to modify or add to the menu when we want.

A good example of a use-case for this would be add-ons wanting to add functionality to allow the user to pre-populate the text editor with a pre-built template. 

We can provide the foundation to implement that type of functionality from the consumer without worrying about navigating Prose-mirror restrictions. 

Further iteration on the action bar will be taken to distinguish menu items that will be used for prose-mirror(editor) functions but that won't be required for the intents and purposes of this PR.

Fixes https://github.com/Lundalogik/crm-feature/issues/4077

### DIFF
--- a/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
+++ b/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
@@ -13,9 +13,10 @@ import { Schema, DOMParser } from 'prosemirror-model';
 import { schema } from 'prosemirror-schema-basic';
 import { addListNodes } from 'prosemirror-schema-list';
 import { exampleSetup } from 'prosemirror-example-setup';
-import { MenuElement, MenuItem } from 'prosemirror-menu';
-import { buildFullMenu } from './menu/full-menu';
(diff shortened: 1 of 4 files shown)


src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -44,28 +45,39 @@ export class ProsemirrorAdapter {
     @State()
     private view: EditorView;
 
+    @State()
+    private actionBarItems: Array<ActionBarItem | ListSeparator> = [];
+
+    private MenuCommandFactory: MenuCommandFactory;
```
civing, line 51:
```suggestion
    private menuCommandFactory: MenuCommandFactory;
```

civing (approved):
We have some styling and dead code to remove regarding the old menu. Let's do that in a follow up commit

PR #2958, opened 2024-05-02
chore(prosemirror-adaptor): remove unused files

civing:
Remove unused files related to the old menu that was replaced with the `limel-action-bar`.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


john-traas:
Looks good. We add the types.ts file back in another PR, but none of the current code in it will persist so easy to just remove it here, and then add it in the other PR.

PR #2960, opened 2024-05-02
feat(text-editor): enable `disabled` prop

Kiarokh:
We were adding specificity of "if not `readonly`" then these
are the colors. But further down, we actually specified colors for the
readonly mode. So this was an overkill. Also would complicate styles
for next commit, for the `disabled` state.
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/text-editor/text-editor.tsx
+++ b/src/components/text-editor/text-editor.tsx
@@ -40,7 +40,7 @@ export class TextEditor implements FormComponent<string> {
      * requirements are met, the field may become enabled again.
      */
     @Prop({ reflect: true })
-    public disabled?: boolean;
+    public disabled?: boolean = false;
 
     /**
      * Set to `true` to make the component read-only.
@@ -164,6 +164,8 @@ export class TextEditor implements FormComponent<string> {
                 value={this.value}
                 aria-controls={this.helperTextId}
                 id={this.editorId}
+                tabindex={this.disabled ? -1 : 0}
+                aria-disabled={this.disabled}
             />,
             this.renderPlaceholder(),
             this.renderHelperLine(),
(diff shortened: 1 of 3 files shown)


src/components/text-editor/text-editor.tsx
```diff
@@ -140,6 +140,8 @@ export class TextEditor implements FormComponent<string> {
                 onChange={this.handleChange}
                 value={this.value}
                 aria-controls={this.helperTextId}
+                tabindex={this.disabled ? -1 : 0}
+                aria-disabled={this.disabled}
```
Kiarokh, line 144: these, together with the newly added styles ☝️ 
```scss
    limel-prosemirror-adapter {
        pointer-events: none;
    }
```
will prevent the ProseMirror to become ever focusable using the keyboard or mouse. 

While individual buttons in `limel-action-bar` can accept a `disabled` prop, I don't think we need to do more than this and add that prop to them. I think the screen-readers will also take care of the rest, when they see `aria-disabled`.

Kiarokh:
I just convert this to DRAFT, so that we don't accidentally merge it. If we do so, all editors in forms will become disabled due to the bug we have right now. However, you can review this anyways.

src/components/text-editor/text-editor.tsx
LucyChyzhova: agree 👍

LucyChyzhova (approved):
Ready to be merged when `invalid` bug will be fixed. Otherwise the text-editor will be disabled everywhere

Befkadu1 (approved):

PR #2965, opened 2024-05-03
feat(text editor): set focus when clicking within the text editor

john-traas:
We still need to handle when the editor is resized. 

fixes https://github.com/Lundalogik/crm-feature/issues/4086


Kiarokh:
I close this, since I've added a PR that handles this properly.
This was a pure CSS issue.

PR #2973, opened 2024-05-13
Fix disabled trailing icon remains disabled until input field changes

vicgeralds:
Makes sure the mdc textfield disabled state is updated correctly. This required changing how the invalid icon is rendered as well.

Rendering an input field as disabled with a trailing icon and later setting `disabled={false}` did not work for some reason. When initializing `MDCTextField`, it will disable the trailing icon by setting tabindex="-1" which makes it unclickable. The problem can be seen here: 

https://github.com/Lundalogik/lime-elements/assets/5563232/fd3efad7-2c38-49d8-b9a1-55796f756cd0

It is possible to reproduce by changing the date picker composite example, adding `disabled: true` to the initial state `props` and removing ``key={`updateOnFormChange-${this.key}`}`` when rendering `<limel-date-picker>`.

When updating the disabled state, the initial invalid state of the input field also matters. Since `MDCTextField` expects the trailing icon to be a button that should be clickable when the input field is not disabled, rendering the invalid icon with class `mdc-text-field__icon--trailing` causes unwanted behaviour. If the input field is initially invalid, the saved tabstate for the trailing icon will be -1, and it will stay unclickable when enabling the input field.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -286,6 +286,8 @@ export class InputField {
         if (this.invalid) {
             this.mdcTextField.valid = false;
         }
+
+        this.mdcTextField.disabled = this.disabled || this.readonly;
     }
 
     public render() {
@@ -627,17 +629,13 @@ export class InputField {
             return;
         }
 
-        const html = [];
-
         const trailingIcon = this.getTrailingIcon();
 
         if (!this.isInvalid() && this.hasLink()) {
-            html.push(this.renderLinkIcon(this.getLink(), trailingIcon));
+            return this.renderLinkIcon(this.getLink(), trailingIcon);
         } else if (trailingIcon) {
-            html.push(this.renderTrailingIcon(trailingIcon));
+            return this.renderTrailingIcon(trailingIcon);
         }
-
-        return html;
     };
 
     private hasLink = () => {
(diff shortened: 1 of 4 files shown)


src/components/input-field/input-field.tsx
```diff
@@ -683,19 +681,21 @@ export class InputField {
...
-        const props: any = {
-            tabIndex: this.isInvalid() ? '-1' : '0',
-        };
-        if (!this.isInvalid()) {
-            props.onKeyPress = this.handleIconKeyPress;
-            props.onClick = this.handleIconClick;
-            props.role = 'button';
+        if (this.isInvalid()) {
+            return (
+                <i class="material-icons mdc-text-field__icon invalid-icon">
```
mjlime, line 686: Dropping the `mdc-text-field__icon--trailing` class will render [this styling](https://github.com/Lundalogik/lime-elements/pull/2973/files#diff-5b95f2a51dfa52377acef55a68454a5e298014dcd32b529ba8db9c8916f870b1R259-R273) unused and won't that make the icon appear strange?

vicgeralds: No, since that only resets unwanted styling that's only applied when using the `mdc-text-field__icon--trailing` class. Maybe we should remove that styling as well. Just need to double check that it's not needed somewhere else then.

mjlime: Just a small thing, not sure if it's relevant. When the invalid icon is shown the cursor is still a pointer and thus a bit confusing since it's not clickable at the moment.

vicgeralds: Yes, that's expected/current behaviour when the field is disabled or invalid I think.

vicgeralds: Hovering and clicking the icon behaves the same way as when hovering or clicking the edges of the field. Which makes the label animate in an odd way, but at least it's consistent 😄 

https://github.com/Lundalogik/lime-elements/assets/5563232/0b29583e-74ab-490d-87a0-357fe84bf2b4

vicgeralds: Sorry, I misunderstood. The cursor was indeed still a pointer in some cases when invalid and a trailing icon was also passed to the input field. Fixed here: 8380470cd99fad8dcae57a1036fd8f38d304b4fa

mjlime (approved):

PR #2974, opened 2024-05-13
feat(text-editor): enable end users to resize the editor

Kiarokh:
Also enable consumers to turn this resizing possibility off.

fix https://github.com/Lundalogik/crm-feature/issues/4097

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/text-editor/examples/text-editor-allow-resize.tsx
+++ b/src/components/text-editor/examples/text-editor-allow-resize.tsx
@@ -1,0 +1,60 @@
+import { Component, h, State } from '@stencil/core';
+/**
+ * Allow resize
+ * The text editor automatically adjusts its own height to fit the content inside.
+ * So as the user types, the editor will grow taller, potentially resizing its own
+ * container element.
+ *
+ * By default, the user can also manually change the height of the text editor
+ * by dragging its bottom right corner.
+ *
+ * As soon as the user has changed the height, this will override the automatic
+ * resizing, and the editor will no longer adjust its height to fit the content inside.
+ *
+ * By setting `allowResize` to `false`, you can disable the end user
+ * to resize the text editor vertically.
+ *
+ * :::tip
+ * Using `max-height` and `min-height` CSS properties, you can limit the
+ * resizing to a specific range.
+ * :::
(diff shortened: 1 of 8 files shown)


LawrenceBorst (approved):
Looks good to me. We could constrain the min width and height of the example but this is a minor issue.

src/components/text-editor/examples/text-editor-allow-resize.tsx
```diff
@@ -0,0 +1,60 @@
...
+ * container element.
+ *
+ * By default, the user can also manually change the height of the text editor
+ * by dragging its bottom right corner.
+ *
+ * As soon as the user has changed the height, this will override the automatic
+ * resizing, and the editor will no longer adjust its height to fit the content inside.
+ *
+ * By setting `allowResize` to `false`, you can disable the end user
+ * to resize the text editor vertically.
```
adrianschmidt, line 15: Is there a legitimate use case for preventing the user from resizing the editor?

adrianschmidt: Or, in other words:

I suggest that we remove the `allowResize` property, and always allow manual resizing. If you already have a strong argument for why this option will be needed, then feel free to convince me 😊

Kiarokh: yes, the legimiate case is shown in the next example, where you want to container to resize the text-editor. In that case, due to `overflow` shenanigans and CSS complexities, these two cannot coexist. So you need to turn OFF the resize for the text editor, for it to work properly.

adrianschmidt: Ah, ok… that kinda sucks… 😢

adrianschmidt (approved):
Changes to the external API approved.

PR #2982, opened 2024-05-17
perf(dock): avoid re-creating observer whenever menu is opened

jgroth:
The intersection observer was re-created every time the button in the menu was clicked, causing the callback to be invoked multiple times

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dock/dock-button/dock-button.tsx
+++ b/src/components/dock/dock-button/dock-button.tsx
@@ -65,6 +65,7 @@ export class DockButton {
 
     private tooltipId: string;
     private customComponentElement: HTMLElement;
+    private intersectionObserver?: IntersectionObserver;
 
     constructor() {
         this.tooltipId = createRandomString();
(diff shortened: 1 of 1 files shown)


hannahu:
Looks good! How can I test it?

jgroth:
> Looks good! How can I test it?

You can add a logpoint in the callback and see that it's only called once when a many is opened :upside_down_face: 
It will be invoked one more time for each time it's clicked without this fix

hannahu:
> > Looks good! How can I test it?
> 
> You can add a logpoint in the callback and see that it's only called once when a many is opened 🙃 It will be invoked one more time for each time it's clicked without this fix

Yeah I was thinking if there was another way to test if it caused more trouble.

hannahu (approved):

hannahu:
Some docs are failing?

PR #2985, opened 2024-05-17
Add translations for menu

john-traas:
Fixes https://github.com/Lundalogik/crm-feature/issues/4089


src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -53,6 +55,13 @@ export class ProsemirrorAdapter {
     @Prop()
     public value: string;
 
+    /**
+     * Defines the language for translations.
+     * Will translate the translatable strings on the components.
+     */
+    @Prop({ reflect: true })
+    public language: Languages = 'en';
```
adrianschmidt, line 63: Summary from conversation:

We have a [config-component](https://github.com/Lundalogik/lime-elements/blob/main/src/components/config/config.tsx), which can be used to set `defaultLocale`:

https://github.com/Lundalogik/lime-elements/blob/3649f10bd316c20c0bc8e8045d44a0de2454b435/src/global/config.ts#L15

That is set to `navigator.language` in the ConfigClass:

https://github.com/Lundalogik/lime-elements/blob/3649f10bd316c20c0bc8e8045d44a0de2454b435/src/global/config.ts#L25

…but I assume that value can be overwritten by using the `<limel-config>` custom element.

So, if we default this `language` prop to `defaultLocale`, like we do in limel-input-field, for example:

https://github.com/Lundalogik/lime-elements/blob/3649f10bd316c20c0bc8e8045d44a0de2454b435/src/components/input-field/input-field.tsx#L228

…we should be able to set the default language to whatever has been set in the Lime CRM user preferences when we load lime-elements in the web client.

Then we can hopefully avoid having to create a wrapper in lime-crm-building-blocks just to get the right language.

Issue #2978, opened 2024-05-16
Snackbar: Improve UX of dismissible snackbars


LucyChyzhova:
## Current behavior
Currently users gets wrong feeling that they can dismiss or undo the action that was done previously by pressing `x` while circular progress bar in place. However hitting the `x` button  just closes the snack bar and do not undo an action.

https://github.com/Lundalogik/lime-elements/assets/50618208/16167be3-8c8d-427c-a1b9-46117546c31d

1. …

## Environment

- **lime-elements** version: 
- Framework used: 
- Logs:

PR #2986, opened 2024-05-20
fix(snackbar): improve UX of dismissible snackbar

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/2978

Screenshot 2024-05-20 at 10 06 08

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/snackbar/snackbar.tsx
+++ b/src/components/snackbar/snackbar.tsx
@@ -160,7 +160,8 @@ export class Snackbar {
                     aria-relevant="additions"
                 >
                     <div class="mdc-snackbar__label" aria-atomic="false"></div>
-                    {this.renderActions(this.actionText, this.dismissible)}
+                    {this.renderActions(this.actionText)}
+                    {this.renderDismissButton(this.dismissible)}
                 </div>
             </aside>
         );
@@ -174,15 +175,14 @@ export class Snackbar {
         }
     }
 
-    private renderActions(actionText: string, dismissible: boolean) {
-        if (!actionText && !dismissible) {
+    private renderActions(actionText: string) {
+        if (!actionText) {
             return;
         }
 
         return (
             <div class="mdc-snackbar__actions" aria-atomic="true">
                 {this.renderActionButton(actionText)}
-                {this.renderDismissButton(dismissible)}
             </div>
         );
     }
--- a/src/components/snackbar/snackbar.scss
+++ b/src/components/snackbar/snackbar.scss
@@ -38,15 +38,15 @@
     }
 
     .mdc-snackbar__surface {
-        transition:
-            opacity 0.2s cubic-bezier(0, 0, 0.2, 1),
-            transform 0.2s cubic-bezier(0, 0, 0.2, 1);
         transform: scale(1);
         opacity: 1;
     }
 }
 
 .mdc-snackbar__surface {
+    transition:
+        opacity 0.2s cubic-bezier(0, 0, 0.2, 1),
+        transform 0.2s cubic-bezier(0, 0, 0.2, 1);
     padding: 0.5rem;
     display: flex;
     align-items: center;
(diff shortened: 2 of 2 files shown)


src/translations/en.ts
```diff
@@ -10,7 +10,7 @@ export default {
     'date-picker.year.heading': 'Year',
     'chip-set.clear-all': 'Clear all',
     'chip-set.remove-chip': 'Remove chip',
-    'snackbar.dismiss': 'Dismiss',
```
Kiarokh, line 13: dismiss is the correct word

src/components/snackbar/snackbar.tsx
```diff
@@ -174,61 +172,46 @@ export class Snackbar {
...
-    private renderTimeoutVisualization() {
-        return (
-            <svg width="36" height="36" viewBox="0 0 36 36">
-                <circle r="18" cx="18" cy="18" fill="var(--track-color)" />
-                <path
-                    class="track"
-                    d="M 18,18 m -16,0 a 16,16 0 1,0 32,0 a 16,16 0 1,0 -32,0"
-                    stroke="var(--fill-color)"
-                />
-            </svg>
```
Kiarokh, line 231: it seems you removed the countdown timer too. was this causing the confusion you mean?

src/translations/en.ts
LucyChyzhova: no 😅

src/components/snackbar/snackbar.tsx
LucyChyzhova: yes, exactly.

src/components/snackbar/examples/snackbar-positioning.scss
```diff
@@ -1,5 +1,5 @@
 limel-snackbar {
-    --snackbar-top: 0;
+    --snackbar-top: 0.5rem;
```
Kiarokh, line 2: this means we should also update the code in the webclient, as we're rendering the snackbars on the top-right location

Kiarokh: I made some changes in the styles. So we now don't have to update the web-client this anymore.

Kiarokh (approved):

Kiarokh:
Lucy I did a lot of refactoring in your PR for the Snackbar, divided it into several commit, and also improved styles. Got rid of the MDC styles. However, after a lot of thinking back and forth, I decided to keep the countdown timer animation on the dismiss button for now. 

Because:

1. I don’t think the user experiences the same thing now, since the button is moved to the top. I think the reason they experienced that they can “undo” some action was because the button was displayed in an irrelevant location. That location is always used to display “actions that users can take”, such as Undoing, or Moving forward. This was the confusing thing. Now we move the close button where all close buttons are conventionally located. 
1. We need to do a lot of refactoring if we wanna move the count down animation to the primary action button on the bottom-right corner of the snackbar. It takes time to do it right, which I don’t have right now.
    1. Even then, we don’t always have a primary action button. So sometimes we will end up not showing any countdown. I’d argue that this is still a useful information to display to the user (which was why we added it). 
    1. We will have a persistent snackbar soon, which won’t have that countdown. This will help us differentiate between these two types. But also a persistent snackbar doesn’t need to have a primary action button.
    1. We also will need to add more actions (secondary actions) to the snackbar.
1. We will soon work on notifications. I’m sure we will focus on the snackbar and its countless problems. There is so much I want to refactor in this component. So I think we need to wait with these kinds of changes.

Let’s live with the new position of the button and see how it feels. Maybe we think it’s fine now after all, or that we really need to change it. I just don’t wanna rush it.

PR #2994, opened 2024-05-22
feat(text-editor): add link to menu

civing:
* Adds a new link menu component to allow users to add text and a link
* Adds a link menu item to the Rich Text Editor (`limel-text-editor`)
* Mod + Click on a link will open link in a new tab
* Double Click on a link will pop the link menu component open allowing the user to edit

fix: https://github.com/Lundalogik/crm-feature/issues/4125
Fixes https://github.com/Lundalogik/crm-feature/issues/4036

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/text-editor/link-menu/link-menu.tsx
```diff
@@ -0,0 +1,97 @@
+import { Component, Prop, h, Event, EventEmitter } from '@stencil/core';
+import { EditorTextLink } from '../prosemirror-adapter/menu/types';
+
+/**
+ *
```
Kiarokh, line 5: hoops

src/components/text-editor/link-menu/link-menu.tsx
```diff
@@ -0,0 +1,97 @@
...
+ */
+@Component({
+    tag: 'limel-link-menu',
```
Kiarokh, line 8: is this a good tag name?

src/components/text-editor/link-menu/link-menu.tsx
```diff
@@ -0,0 +1,97 @@
...
+                onChange={this.handleLinkTitleChange}
+            />,
+            <limel-input-field
+                label="Link"
+                value={this.link?.href || ''}
+                onChange={this.handleLinkValueChange}
+            />,
+            <div class="actions">
+                <limel-button
+                    label="Cancel" // TODO: use translations
```
Kiarokh, line 58: translations are merged now, you can use them in this component

src/components/text-editor/link-menu/link-menu.scss
adrianschmidt, line 1: How is this PR going? Should I take over?

(Also, we are going to check with Add-ons what their order of priority is for features, and what they consider MVP for them to be able to use the editor in the email feature in production. And then we will focus very strictly on that feature set. So depending on how they prio links, this might be put on the back burner for a while. Just FYI.)

src/components/text-editor/link-menu/link-menu.tsx
LucyChyzhova: Great! I will add a fixup

src/components/text-editor/link-menu/link-menu.scss
LucyChyzhova: I had a day off yesterday, checking what's left right now. @adrianschmidt

src/components/text-editor/link-menu/link-menu.tsx
LucyChyzhova: fixed

src/components/text-editor/link-menu/link-menu.tsx
LucyChyzhova: fixed

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,150 @@
...
+     */
+    @Event()
+    private linkChange: EventEmitter<EditorTextLink>;
+
+    public render() {
+        return [
+            <limel-input-field
+                label="Text"
+                value={this.link?.text || ''}
+                leadingIcon="open_book"
```
Kiarokh, line 63:
```suggestion
                leadingIcon="text_cursor"
```

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,150 @@
...
+                onChange={this.handleLinkTitleChange}
+            />,
+            <limel-input-field
+                label="Link"
+                value={this.link?.href || ''}
+                leadingIcon="globe_earth"
```
Kiarokh, line 69:
```suggestion
                leadingIcon="-lime-text-link"
```

etc/lime-elements.api.md
adrianschmidt, line 1: Updating this file should never be a separate commit, because that means some other commit isn't atomic (because the changes in it should have updated this file, but didn't).

Don't worry about it for now though, I'll fix it this time 👍

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,146 @@
...
+    public language: Languages = 'en';
+
+    /**
+     * Open state of the dialog
+     */
+    @Prop()
+    public isOpen: boolean = false;
+
+    @State()
+    private invalidLink: boolean = false;
```
adrianschmidt, line 37: State-props are more internal than events, so we place state props after the event emitter properties. (I'll fix this, just wanted to comment to spread the knowledge 😊)

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Emitted when the user inputs new values for the link
+     */
+    @Event()
+    private linkChange: EventEmitter<EditorTextLink>;
+
+    public render() {
+        return [
+            <limel-input-field
+                label="Text"
```
adrianschmidt, line 61: We need translations here.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,146 @@
...
+                value={this.link?.text || ''}
+                leadingIcon="open_book"
+                onChange={this.handleLinkTitleChange}
+            />,
+            <limel-input-field
+                label="Link"
```
adrianschmidt, line 67: We need translations here too.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
adrianschmidt: ⚡ 

(@john-traas You can use ⚡ to indicate that you've addressed the comment, without having to resolve the thread. In this case though, I'm pretty sure Kia doesn't mind me resolving the thread even if he hasn't had time to double check that you made the change that he meant, because there really isn't much room for misunderstanding in this case 😊)

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
adrianschmidt: I'm resolving this too. The icon has been updated as requested.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
john-traas: Ah, good to know. Thanks 😊

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,146 @@
...
+     * Emitted when the menu is saved from inside the component.
+     */
+    @Event()
+    private save: EventEmitter<void>;
+
+    /**
+     * Emitted when the user inputs new values for the link
+     */
+    @Event()
+    private linkChange: EventEmitter<EditorTextLink>;
```
adrianschmidt, line 56: I guess this might be a quirk of prosemirror, and nothing we can do anything about, but I find it slightly weird that we emit updated values for the link as the user is typing, but do _not_ attach the current value to the `save` event.

Is that indeed due to prosemirror, or is it something we control, and could potentially change?

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,150 @@
...
+
+    private isValidLink = (href: string): boolean => {
+        try {
+            const url = new URL(href);
+
+            return url.protocol === 'http:' || url.protocol === 'https:';
+        } catch {
+            return false;
+        }
+    };
```
adrianschmidt, line 120: We are doing almost the same thing here:

image

We also have a whole heap of similar stuff in https://github.com/Lundalogik/lime-elements/blob/main/src/util/link-helper.ts

I don't think we will want to limit the possible protocols to `http` and `https`. We've limited the possible protocols in https://github.com/Lundalogik/lime-elements/blob/main/src/util/link-helper.ts for some reason that I can't quite remember right now, and it's already become a problem because there are all kinds of app-specific protocols people want to be able to use.

So, we should reuse what's already in https://github.com/Lundalogik/lime-elements/blob/main/src/util/link-helper.ts, and adapt it if needed. The functions in there that aren't already exported can certain be exported so they are possible to import and use here and in `menu-commands.ts`.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,150 @@
...
+            const formattedLink = 'http://' + href;
+            if (this.isValidLink(formattedLink)) {
+                this.link = { ...this.link, href: formattedLink };
+
+                return true;
+            }
+        }
+
+        return false;
+    };
```
adrianschmidt, line 110: We might want to put this in https://github.com/Lundalogik/lime-elements/blob/main/src/util/link-helper.ts?rgh-link-date=2024-05-29T13%3A15%3A22Z too, so it can be reused. The stuff that's already there doesn't allow links that are missing the protocol, but even if we don't change the existing behaviour of what's there, we could add this alongside, so it can be reused elsewhere too.

I guess we might want to wait until we're done with the text-editor MVP though.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,150 @@
...
+                    disabled={!this.link?.href || this.invalidLink}
+                    onClick={this.handleSave}
+                    slot="button"
+                />
+            </div>,
+        ];
+    }
+
+    private getTranslation(key: string) {
+        return translate.get(`${key}`, this.language);
```
adrianschmidt, line 91: Correct me if I'm missing something, but…

```suggestion
        return translate.get(key, this.language);
```

?

🙂

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,150 @@
...
+    private validateLink = (href: string): boolean => {
+        if (this.isValidLink(href)) {
+            return true;
+        }
+
+        // If the link doesn't have a scheme, add 'http://' and validate again
+        if (!/^[a-zA-Z][a-zA-Z\d+\-.]*:/.test(href)) {
+            const formattedLink = 'http://' + href;
+            if (this.isValidLink(formattedLink)) {
+                this.link = { ...this.link, href: formattedLink };
```
adrianschmidt, line 103: Ah, hang on! We're actually _updating_ the input value here?

Should we really do that? If we do, I think we should use `https` rather than `http`.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
Kiarokh: 😊 Personally I don't mind if you just mark as resolved neither. And if you use the <kbd>Commit suggestion</kbd> button directly in GitHub, it will mark the thread as "resolved" automatically. But when <kbd>Commit suggestion</kbd>  becomes disabled, it also means you've adressed the thing. So do whatever works for you. But commonly we use ⚡, specially if we address the threads it very quickly 😄

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
civing: You are absolutely right 👍.
My intention was to consolidate that before putting it in review. There are more things like this since I had to hand over this change mid-process 🙈

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
civing: I did add that and I'm not sure if this is what we want. If the user would insert a link to `www.google.com` they would expect that link to be absolute and not relative, I would assume.

So, `http://www.google.com`
Rather than, `https://lundalogik.github.io/lime-elements/versions/PR-2994/www.google.com`

But if they enter `/client/object/company/1033` the user might actually expect the link to be relative.

I don't know what we should support here and how clever we should try to be.

civing: I'm not so sure that we are ultimately prepending the schema to the link if missing. This is only the link validation method.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
john-traas: I have no idea on this one. It most certainly has no need for a string literal.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
john-traas: ⚡

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
john-traas: ⚡

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
john-traas: I don't think we should update the input value here. It should only handle link validation, nothing more. I'll change this in a fixup.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
john-traas: From what I could tell, this doesn't really do anything that the link-helper functions don't already do. I've used the functions in link-helper now to validate the urls. I think if we end up needing further functionality we can add that when needed.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
john-traas: I've gone ahead and used the functions in `link-helper.ts` for validation.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,150 @@
...
+                leadingIcon="text_cursor"
+                onChange={this.handleLinkTitleChange}
+            />,
+            <limel-input-field
+                label="Link"
+                value={this.link?.href || ''}
+                leadingIcon="-lime-text-link"
+                invalid={this.invalidLink}
+                onChange={this.handleLinkValueChange}
+            />,
```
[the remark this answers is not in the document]
john-traas: It is an issue that pretty much any garbage string returns as a `valid` url, and that's with the `link-helper.ts` functions as well. I'm not quite sure how we can be smart about this to be honest.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
john-traas: I've removed the update to the link as it's only used for validation. We also might not necessarily want to prepend a protocol to what a user might paste as a link?

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -211,6 +211,8 @@ export class ProsemirrorAdapter {
                 dispatchTransaction: this.handleTransaction,
             },
         );
+        // eslint-disable-next-line no-underscore-dangle
+        (this.view.dom as any).__adapterInstance = this;
```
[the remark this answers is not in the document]
john-traas: You're right we don't want this exposed. I've pushed a fixup and replaced this instead with an event emitter that we listen for to open the link-menu.

adrianschmidt:
You can ignore that the SonarCloud Scan job failed. It was green before, I just re-ran it because it had failed on another PR, but it looked like it was probably something wrong with SonarCloud rather than with the PR. So I re-ran it here to check if this would also fail, and it did.

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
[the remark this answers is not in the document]
john-traas: ⚡

src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
```diff
@@ -0,0 +1,214 @@
...
+        const node = view.state.doc.nodeAt(pos.pos);
+        if (
+            node &&
+            node.marks.some((mark) => mark.type.name === EditorMenuTypes.Link)
+        ) {
+            const linkMark = node.marks.find(
+                (mark) => mark.type.name === EditorMenuTypes.Link,
+            );
+            const href = linkMark.attrs.href;
+            const from = findStart(view.state.doc, pos.pos);
```
LawrenceBorst, line 117: For these two functions, could we use the `resolve` function to get more information on this `node`? (Because we are trying to find the start and end of the link node I believe)

src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
```diff
@@ -0,0 +1,214 @@
...
+            bubbles: true,
+            composed: true,
+        },
+    );
+    view.dom.dispatchEvent(event);
+};
+
+let lastClickTime = 0;
+const processClickEvent = (view: EditorView, event: MouseEvent): boolean => {
+    const DOUBLE_CLICK_DELAY = 500;
```
LawrenceBorst, line 158: We can place this constant outside the function

src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
```diff
@@ -0,0 +1,214 @@
...
+            handleDOMEvents: {
+                mousedown: (view, event) => {
+                    if (
+                        (event.metaKey || event.ctrlKey) &&
+                        event.button === 0
+                    ) {
+                        return processModClickEvent(view, event);
+                    }
+
+                    return processClickEvent(view, event);
```
LawrenceBorst, line 57: Should we call this `processDoubleClickEvent` in line with the `processModClickEvent` function?

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+        return !!getHref(href);
+    };
+
+    private handleCancel = () => {
+        this.cancel.emit();
+    };
+
+    private handleSave = () => {
+        this.save.emit();
```
LawrenceBorst, line 113: I wonder if we should stop the propagation for the events, since we have had trouble in the past with multiple events flying around under the same name

LawrenceBorst (commented):
Looked over the code, will scan once more after testing it out manually. It is easy to follow. I only struggled with the `getStart` and `getEnd` functions, but I left a comment regarding those.

I would say that nesting could be reduced quite a bit in the plugin file by using more early returns

src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
john-traas: Is resolve a built in prosemirror function? I can look at that.

civing (commented):
I'm not done reviewing but posting some interim comments.

src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
john-traas: ⚡

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+    private validateLink = (href: string): boolean => {
+        if (!hasKnownProtocol(href)) {
+            const formattedLink = prependProtocol(href);
+
+            return !!getHref(formattedLink);
+        }
+
+        return !!getHref(href);
+    };
```
civing, line 106: This validation seems to accept almost anything as valid links.

These ones will be considered valid:
http://www.exam ple.com
htttp://www.example.com
/path/with space

I don't think we should fix this in this PR, it's way too big already.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
civing: good catch!

src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
```diff
@@ -0,0 +1,214 @@
...
+
+let lastClickTime = 0;
+const processClickEvent = (view: EditorView, event: MouseEvent): boolean => {
+    const DOUBLE_CLICK_DELAY = 500;
+    const now = Date.now();
+
+    if (now - lastClickTime < DOUBLE_CLICK_DELAY) {
+        const linkData = getLinkDataAtPosition(view, event);
+        if (linkData) {
+            const { href, text, from, to } = linkData;
```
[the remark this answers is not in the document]
john-traas: Ah, I didn't check that the href attribute was the same one that is in the original mark. I'll update the functions.

src/components/text-editor/prosemirror-adapter/menu/menu-commands.ts
```diff
@@ -138,6 +183,16 @@ const toggleBlockType = (schema, type, attrs = {}, wrap = false) => {
     };
 };
 
+export const isValidUrl = (text: string): boolean => {
```
civing, line 186: This should probably be the same as "validateLink". But if we use that one I think almost anything will be considered a link 😅

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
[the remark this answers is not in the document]
john-traas: I've removed the limel-text from the file in a fixup.

src/components/text-editor/link-menu/limel-text-editor-link-menu.scss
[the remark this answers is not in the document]
john-traas: I've removed the limel-text from the file in a fixup.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
```diff
@@ -0,0 +1,146 @@
+import { Component, Prop, h, Event, EventEmitter, State } from '@stencil/core';
+import { EditorTextLink } from '../prosemirror-adapter/menu/types';
+import { Languages } from '../../date-picker/date.types';
+import translate from '../../../global/translations';
+
+/**
+ * This component is a menu for editing a link in the text editor.
+ * It allows the user to input a title of a link and a link itself.
+ */
+
```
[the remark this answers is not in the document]
john-traas: ⚡

src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
john-traas: ⚡

src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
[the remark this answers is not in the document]
john-traas: ⚡

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
john-traas: I left another comment on this above saying that I wasn't sure how we should handle this. I guess some clever Regex might be the way to go.

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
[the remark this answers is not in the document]
john-traas: ⚡

src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
john-traas: ⚡

src/components/text-editor/link-menu/limel-text-editor-link-menu.tsx
john-traas: ⚡

jgroth (approved):

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -110,23 +132,57 @@ export class ProsemirrorAdapter {
...
+            this.handleOpenLinkMenu,
+        );
+        this.view.destroy();
     }
 
     public render() {
         return [
             <div id="editor" />,
             <limel-action-bar
+                slot="trigger"
```
Kiarokh, line 154: why was this needed? 🤔

PR #2998, opened 2024-05-27
Extend menu commands 

john-traas:
Updates the menu-command factory to make it possible to toggle menu commands. 

Fixes https://github.com/Lundalogik/crm-feature/issues/4130

### DIFF
--- a/src/components/text-editor/prosemirror-adapter/menu/menu-command.spec.tsx
+++ b/src/components/text-editor/prosemirror-adapter/menu/menu-command.spec.tsx
@@ -8,6 +8,7 @@ import { exampleSetup } from 'prosemirror-example-setup';
 import { EditorView } from 'prosemirror-view';
 import { MenuCommandFactory } from './menu-commands';
 import { EditorMenuTypes } from './types';
+import { strikethrough } from './menu-schema-extender';
 
 describe('MenuCommandFactory', () => {
     let mySchema: Schema;
@@ -23,7 +24,9 @@ describe('MenuCommandFactory', () => {
                 'paragraph block*',
                 'block',
             ),
-            marks: basicSchema.spec.marks,
+            marks: basicSchema.spec.marks.append({
+                strikethrough: strikethrough,
+            }),
         });
 
         factory = new MenuCommandFactory(mySchema);
--- a/src/components/text-editor/prosemirror-adapter/menu/menu-commands.ts
+++ b/src/components/text-editor/prosemirror-adapter/menu/menu-commands.ts
@@ -1,14 +1,14 @@
 import { toggleMark, setBlockType, wrapIn } from 'prosemirror-commands';
 import { Schema, MarkType, NodeType, Attrs } from 'prosemirror-model';
-import { wrapInList } from 'prosemirror-schema-list';
-import { Command, EditorState } from 'prosemirror-state';
+import { findWrapping, liftTarget } from 'prosemirror-transform';
+import { Command, EditorState, TextSelection } from 'prosemirror-state';
 import { EditorMenuTypes, LevelMapping } from './types';
 
 type CommandFunction = (
     schema: Schema,
     mark: EditorMenuTypes,
     url?: string,
-) => Command;
+) => CommandWithActive;
 
 interface CommandMapping {
     [key: string]: CommandFunction;
@@ -42,7 +42,7 @@ const setActiveMethodForNode = (
         const node = $from.node($from.depth);
 
         if (node && node.type.name === nodeType.name) {
-            if (nodeType.name === 'heading') {
+            if (nodeType.name === LevelMapping.Heading && level) {
                 return node.attrs.level === level;
             }
 
@@ -107,6 +107,37 @@ const isExternalLink = (url: string): boolean => {
     return !url.startsWith(window.location.origin);
 };
 
+const toggleBlockType = (schema, type, attrs = {}, wrap = false) => {
+    const blockType = schema.nodes[type];
+    const paragraphType = schema.nodes.paragraph;
+
+    return (state, dispatch) => {
+        const { $from, to } = state.selection;
+        if (
+            state.selection instanceof TextSelection &&
+            $from.sameParent($from.doc.resolve(to))
+        ) {
+            if ($from.parent.type === blockType) {
+                if (dispatch) {
+                    dispatch(
+                        state.tr.setBlockType($from.pos, to, paragraphType),
+                    );
+                }
+
+                return true;
+            } else {
+                if (wrap) {
+                    return wrapIn(blockType, attrs)(state, dispatch);
+                } else {
+                    return setBlockType(blockType, attrs)(state, dispatch);
+                }
+            }
+        }
+
+        return false;
+    };
+};
+
 const createSetNodeTypeCommand = (
     schema: Schema,
     nodeType: string,
@@ -118,8 +149,10 @@ const createSetNodeTypeCommand = (
     }
 
     let command: CommandWithActive;
-    if (nodeType === 'heading' && level) {
-        command = setBlockType(type, { level: level });
+    if (nodeType === LevelMapping.Heading && level) {
+        command = toggleBlockType(schema, LevelMapping.Heading, {
+            level: level,
+        });
     } else {
         command = setBlockType(type);
     }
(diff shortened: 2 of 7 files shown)


src/components/text-editor/prosemirror-adapter/menu/menu-commands.ts
```diff
@@ -42,7 +42,7 @@ const setActiveMethodForNode = (
         const node = $from.node($from.depth);
 
         if (node && node.type.name === nodeType.name) {
-            if (nodeType.name === 'heading') {
+            if (nodeType.name === LevelMapping.Heading && level) {
                 return node.attrs.level === level;
```
adrianschmidt, line 46: Ok, so:

- if `node && node.type.name === nodeType.name`, but `nodeType.name` is not `'heading'`, return `true`
- if `node && node.type.name === nodeType.name`, and `nodeType.name` _is_ `'heading'`, but `level` is falsy, return `false`
- if `node && node.type.name === nodeType.name`, and `nodeType.name` _is_ `'heading'`, and `level` is truthy, return `true` if `node.attrs.level` and `level` are the same

Correct? 😅

john-traas: Yes. 😅 

This set of conditions should never happen as a `heading` must have a `level`:
if `node && node.type.name === nodeType.name`, and `nodeType.name` is `'heading'`, but `level` is falsy, return false

john-traas: We determine the `node.type` of the `current node` to show it is `active`. If the `node.type` is a `'heading'`, the `level` must be the same as the one on the menu item that was clicked so the relevant menu item shows as selected (active).

src/components/text-editor/prosemirror-adapter/menu/types.ts
```diff
@@ -15,6 +15,10 @@ export const EditorMenuTypes = {
     Link: 'link',
     OrderedList: 'ordered_list',
     BulletList: 'bullet_list',
+    Underline: 'underline',
+    Strikethrough: 'strikethrough',
```
adrianschmidt, line 19: I'm getting errors like these when using underline and strikethrough:

image

Underline and strikethrough looks like they are working in the editor, but as soon as you have any part of the text formatted with either of them, the output stops updating.

I don't think we should spend time debugging this until we know whether underline and strikethrough are in the set of MVP features though.

src/components/text-editor/prosemirror-adapter/menu/menu-commands.ts
adrianschmidt: Ah, yeah, "setActiveMethodForNode". Ok, that makes sense then 👍 🙂

src/components/text-editor/prosemirror-adapter/menu/menu-commands.ts
```diff
@@ -138,12 +173,52 @@ const createWrapInCommand = (
...
+    } else {
+        command = wrapIn(type);
+    }
+
     setActiveMethodForWrap(command, type);
 
     return command;
 };
 
+const toggleList = (listType) => {
```
adrianschmidt, line 188: I'm not sure if markdown supports nested lists. The current behaviour seems to be fairly similar to how the list buttons work in the editor here on GitHub. In the future, I imagine we might want more advanced behaviour, at least when the output mode is html, but if so, that's for another time.

👍

src/components/text-editor/prosemirror-adapter/menu/types.ts
```diff
@@ -15,6 +15,10 @@ export const EditorMenuTypes = {
...
+    Code: 'code',
+    CodeBlock: 'code_block',
```
adrianschmidt, line 21: This PR is broken into commits in quite a weird way, imo. I'd rather see that any changes that were already needed before adding new types here were made first, in one or more commits. (I'm guessing one commit, since it's basically fixing one problem: "be able to toggle stuff on and off".) Then any new features, like support for underline and strikethrough, should be added in their own commits.

But, unless underline and strikethrough are considered MVP, we shouldn't add them now.

src/components/text-editor/prosemirror-adapter/menu/menu-commands.ts
```diff
@@ -118,8 +149,12 @@ const createSetNodeTypeCommand = (
     }
 
     let command: CommandWithActive;
-    if (nodeType === 'heading' && level) {
-        command = setBlockType(type, { level: level });
+    if (nodeType === LevelMapping.Heading && level) {
+        command = toggleBlockType(schema, LevelMapping.Heading, {
+            level: level,
+        });
```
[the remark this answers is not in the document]
john-traas: I agree that it is a bit annoying, and it's the same for the lists in the current state. I have some thoughts about how to implement a better experience but felt based on the bet for the cycle, that it wasn't a priority at the moment.

src/components/text-editor/prosemirror-adapter/menu/types.ts
john-traas: It's a very small fix for these two. The marks for underline and strikethrough aren't in the default markdown serialiser that prosemirror provides.

adrianschmidt: OK, if you know what's needed, and it doesn't turn into a rabbit hole (aka, you can fix it before the meeting at 16:00), go ahead 😊

src/components/text-editor/prosemirror-adapter/menu/types.ts
john-traas: I can re-arrange the commits in an order that makes more sense.

src/components/text-editor/prosemirror-adapter/menu/types.ts
adrianschmidt: (To clarify, I don't mean to take the role of some ultimate dictator, deciding what you guys can and cannot do. My intention is just to provide a firm guiding hand 😊)

src/components/text-editor/utils/markdown-converter.ts
```diff
@@ -1,8 +1,32 @@
...
+            open: '<u>',
+            close: '</u>',
+            mixable: true,
+            expelEnclosingWhitespace: true,
+        },
+        strikethrough: {
+            open: '~~',
+            close: '~~',
+            mixable: true,
+            expelEnclosingWhitespace: true,
```
adrianschmidt, line 25: Have we decided anything on the markdown vs html output question?

I'm asking since we seem to be mixing html (`<u>undeline</u>`) and markdown (`~~strikethrough~~`) here…? I know vanilla markdown doesn't actually support underline, while some flavours support interpreting enclosing single underscores as underline (which I've always thought makes sense 🤷‍♂️).

adrianschmidt: Oh… `customMarkdownSerializer`… so I guess this is when outputting markdown then?

john-traas: This is when `markdown` is set as the `contentType`, but your question does have me thinking further about this.

adrianschmidt: Are we outputting markdown or html by default?

I would assume the main focus right now is html anyway, since that's presumably what the email feature needs?

john-traas: We are outputting markdown by default but that can be changed by modifying one line in the `prosemirror-adapter`

adrianschmidt: Should we hold off on including the underline button?

I mean, it will work if the output is formatted with `limel-markdown`, as that can handle mixed markdown and html. But we're not actually outputting pure markdown when there's something that's been underlined. So I'm thinking we might not want to include the underline button in the default set of menu buttons when the output mode is markdown…

But we don't want to spend any time on including it when in html mode and excluding it in markdown mode. Not right now.

adrianschmidt: I'm thinking I can just move that commit to a separate branch, and you can open a PR for that commit only, and then we can leave that to rest until we're ready to look at it.

john-traas: I think that sounds like a good solution.

src/components/text-editor/prosemirror-adapter/menu/menu-command.spec.tsx
adrianschmidt, line 1: I've tested it a little bit in Safari on an iPhone via BrowserStack, and that seems to work well enough. There seemed to be some odd behaviour around adding new lines inside a code block, but honestly, I can't see people adding a whole lot of code blocks via their phones…

adrianschmidt (approved):

src/components/text-editor/prosemirror-adapter/menu/menu-commands.ts
```diff
@@ -107,6 +107,37 @@ const isExternalLink = (url: string): boolean => {
...
+                    return wrapIn(blockType, attrs)(state, dispatch);
+                } else {
+                    return setBlockType(blockType, attrs)(state, dispatch);
+                }
+            }
+        }
+
+        return false;
+    };
+};
```
adrianschmidt, line 139: Oh, yeah, I didn't want to sidetrack the current cycle into a massive refactoring effort, but once we have achieved MVP, I think we will need to go over the whole text-editor component and refactor a bunch of parts, just to break stuff out into named functions, because following stuff like this is not easy, and a few months down the line, when there's no one who has it fresh in their memory that you can ask, it's going to be a nightmare.

(Not just a comment about this code specifically, there seems to be a lot of deeply nestled code since earlier too.)

john-traas: I definitely agree!

PR #2999, opened 2024-05-27
Fix text editor gets dirty

LawrenceBorst:
Fixes Lundalogik/crm-feature#4099

**How to test:**

I've done my best to test it. Here's what to look for.

Thinking of edge cases, there's only one, which is when the initial value is empty. Try and check if this has any strange behaviour.

Also, the save button should disappear when the value goes back to what it originally was. So if your text editor says "some sentence", and you edit it to something else, and then back to "some sentence", we should not see the save button.

We believe the cause of the original bug was the fact that the default content type was changed from "html" to "markdown". To test, try and change the content type in the card component in lime admin and see if the bug persists.

Make sure saving continues to work as expected.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
+++ b/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
@@ -8,7 +8,7 @@ import {
     Watch,
     h,
 } from '@stencil/core';
-import { EditorState, Plugin, PluginKey } from 'prosemirror-state';
+import { EditorState, Plugin, PluginKey, Transaction } from 'prosemirror-state';
 import { EditorView } from 'prosemirror-view';
 import { Schema, DOMParser } from 'prosemirror-model';
 import { schema } from 'prosemirror-schema-basic';
@@ -226,9 +226,14 @@ export class ProsemirrorAdapter {
         this.view.dispatch(tr);
     }
 
-    private handleTransaction = (transaction) => {
+    private handleTransaction = (transaction: Transaction) => {
         const newState = this.view.state.apply(transaction);
         this.view.updateState(newState);
+
+        if (transaction.getMeta('pointer')) {
+            return;
+        }
+
         this.change.emit(
             this.contentConverter.serialize(this.view, this.schema),
         );
--- a/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
+++ b/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
@@ -78,6 +78,7 @@ export class ProsemirrorAdapter {
     private menuCommandFactory: MenuCommandFactory;
     private schema: Schema;
     private contentConverter: ContentTypeConverter;
+    private suppressChangeEvent = false;
 
     /**
      * Dispatched when a change is made to the editor
@@ -211,6 +212,7 @@ export class ProsemirrorAdapter {
     }
 
     private async updateView(content: string) {
+        this.suppressChangeEvent = true;
         const html = await this.contentConverter.parseAsHTML(
             content,
             this.schema,
@@ -224,11 +226,16 @@ export class ProsemirrorAdapter {
         const tr = this.view.state.tr;
         tr.replaceWith(0, tr.doc.content.size, prosemirrorDoc.content);
         this.view.dispatch(tr);
+        this.suppressChangeEvent = false;
     }
 
     private handleTransaction = (transaction: Transaction) => {
         const newState = this.view.state.apply(transaction);
         this.view.updateState(newState);
+
+        if (this.suppressChangeEvent) {
+            return;
+        }
 
         if (transaction.getMeta('pointer')) {
             return;


src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -218,6 +227,13 @@ export class ProsemirrorAdapter {
     private handleTransaction = (transaction) => {
         const newState = this.view.state.apply(transaction);
         this.view.updateState(newState);
+
+        if (isEqual(this.prosemirrorDoc, transaction.doc) && this.shouldEmit) {
+            this.shouldEmit = false;
```
LawrenceBorst, line 238: Oh I seem to have the naming backwards 😅

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -127,6 +127,10 @@ export class ProsemirrorAdapter {
         ];
     }
 
+    public disconnectedCallback() {
+        this.view.destroy();
```
adrianschmidt, line 127: Whoa! Yeah, good catch!

I guess this means we have a memory leak in all apps that have the activities feed enabled with the text-editor… 😬

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -122,12 +122,11 @@ export class ProsemirrorAdapter {
...
-        /* eslint-disable multiline-ternary */
-        this.contentConverter =
-            this.contentType === 'markdown'
-                ? new markdownConverter()
-                : new HTMLConverter();
-        /* eslint-enable multiline-ternary */
+        if (this.contentType === 'markdown') {
+            this.contentConverter = new markdownConverter();
+        } else if (this.contentType === 'html') {
+            this.contentConverter = new HTMLConverter();
```
adrianschmidt, line 134: Nitpick: This isn't purely a refactor or code style change, since the existing behaviour would default to `HTMLConverter` if `this.contentType` was anything other than `'markdown'`, while this new behaviour is defaulting to `undefined` if `this.contentType` is neither `'markdown'` nor `'html'`.

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -122,12 +122,11 @@ export class ProsemirrorAdapter {
     }
 
     private setupContentConverter() {
-        /* eslint-disable multiline-ternary */
-        this.contentConverter =
-            this.contentType === 'markdown'
-                ? new markdownConverter()
-                : new HTMLConverter();
-        /* eslint-enable multiline-ternary */
```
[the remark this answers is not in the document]
john-traas: 😆 
<iframe src="https://giphy.com/embed/jUwpNzg9IcyrK" width="480" height="360" style="" frameBorder="0" class="giphy-embed" allowFullScreen></iframe><p><a href="https://giphy.com/gifs/the-simpsons-scared-homer-simpson-jUwpNzg9IcyrK">via GIPHY</a></p>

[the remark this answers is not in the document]
john-traas: 😆

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
adrianschmidt, line 238: I think the naming is a little bit confusing here. We seem to only be avoiding emitting something if `this.shouldEmit` is _true_, and then we set it to false, so it will never be true again…?

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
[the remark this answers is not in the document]
adrianschmidt: # 🤗

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -216,9 +226,20 @@ export class ProsemirrorAdapter {
...
-    private handleTransaction = (transaction) => {
+    private handleTransaction = (transaction: Transaction) => {
         const newState = this.view.state.apply(transaction);
         this.view.updateState(newState);
+
+        if (transaction.getMeta('pointer')) {
+            return;
+        }
+
+        if (isEqual(this.prosemirrorDoc, transaction.doc) && this.shouldEmit) {
```
adrianschmidt, line 237: Logical operators are evaluated left-to-right. Since `this.shouldEmit` will _almost always_ be `false`, we can save a lot of work by not recursively comparing the two data structures of unknown size just to then throw the result away when `this.shouldEmit` is evaluated to false.

```suggestion
        if (this.shouldEmit && isEqual(this.prosemirrorDoc, transaction.doc)) {
```

By putting the simple boolean check first, we will never enter the deep equality check if the boolean is `false`.

adrianschmidt:
FYI: I squashed the existing fixup, and rebased on what's currently the latest `main`.

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -123,6 +114,15 @@ export class ProsemirrorAdapter {
...
     }
 
+    public componentDidLoad() {
+        // Stencil complains loudly about triggering rerenders in
+        // componentDidLoad, but we have to, so we're using setTimeout to
+        // suppress the warning. /Ads
+        setTimeout(() => {
+            this.initializeTextEditor();
+        }, 0);
+    }
```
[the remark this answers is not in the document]
adrianschmidt: (I removed this commit.)

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
adrianschmidt, line 1: Oh, btw, the prosemirror-adapter is meant to be an internal component, I assume? Or is it something the consumer would be aware of?

I'm asking, because if it's an internal component that the consumer doesn't know about, we shouldn't use `prosemirror-adapter` as the "scope" for commits that are of a type that show up in the changelog, like `fix`, `feat`, and `perf`. Those should have a scope that is meaningful to the consumer, so the consumer knows which component is affected. So `text-editor` in this case (again, unless I'm mistaken about prosemirror-adapter).

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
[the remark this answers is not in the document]
LawrenceBorst: Ah you are right, the style guide always put `render` last :)

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
LawrenceBorst: This is good to know. I didn't think of this. I'll rename the commits

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
LawrenceBorst: Fixed

adrianschmidt: I'm not completely at peace with the fixup you made for this, because we now instead have a flag that says `this.shouldEmit = false` and we're still emitting stuff unless a specific circumstance happens. So I think I'll remove that fixup while I'm fiddling around with some stuff, and then try to find a different name for the property or something.

I'm going to fiddle around with the commits a little bit, so there may be merge conflicts if you are making changes at the same time.

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -123,12 +123,11 @@ export class ProsemirrorAdapter {
...
-        this.contentConverter =
-            this.contentType === 'markdown'
-                ? new markdownConverter()
-                : new HTMLConverter();
-        /* eslint-enable multiline-ternary */
+        if (this.contentType === 'markdown') {
+            this.contentConverter = new markdownConverter();
+        } else {
+            this.contentConverter = new HTMLConverter();
+        }
```
adrianschmidt: I updated this commit so that there is no change in functionality, and instead changed the `else` into an `else if` in the commit where you added the error logging 👍

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
adrianschmidt: I'm not sure I completely understand how this if statement works 😅

This part specifically:

```ts
isEqual(this.prosemirrorDoc, transaction.doc)
```

As far as I understand, we want to avoid emitting a `change` event if the changes are only syntactic changes that prosemirror makes automatically, stuff like replacing `<b>` with `<strong>`, for example. But I don't understand what `this.prosemirrorDoc` and `transaction.doc` are, or how their values being equal to each other tells us what we need to know? 🤔

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
adrianschmidt: This was merged in #3009

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -224,11 +226,21 @@ export class ProsemirrorAdapter {
         const tr = this.view.state.tr;
         tr.replaceWith(0, tr.doc.content.size, prosemirrorDoc.content);
         this.view.dispatch(tr);
+        this.suppressChangeEvent = false;
```
civing, line 229: Should this really be set here? It seem like a better fit to clear this boolean on row 237. I can't see any relevant events here that would make `handleTransaction` to be called before flipping this boolean

civing: But since this seems to have solved it for you guys I'm probably missing somehting 😅

adrianschmidt: I've looked at it, and I _think_ I disagree. It feels to me like `handleTransaction` shouldn't just assume that because we suppressed one change event, we should let the next one trough. I think it's better that the function that sets the value also clears it.

With the current state of the code, it shouldn't matter in practice, but I think it feels cleaner for potential future changes to not have `handleTransaction` making assumptions like that.

john-traas (approved):

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
civing: I agree with this. For some reason I assumed that the dispatch somehow was hiding some async code but it must not then.

PR #3000, opened 2024-05-27
Convert menu handling to plugins

john-traas:
There's a risk of the prosemirror adapter becoming too heavy with different parts of the text editor being too tightly coupled. 

We should adopt the same approach for extending the adaptor as we do for menu-state tracking, via plugins. This is the way prosemirror expects us to do it and it enables us to keep a good separation of concerns allowing for future development of this tool as and when it's needed. 

This PR extracts the menu-state-tracking to a standalone plugin and does the same for handling interactions with the action bar items. 
- the `menu-state-tracking-plugin` tracks the state of the editor view to determine if a markType or nodeType is active within the active commands and returns an object of the active menu types `Record<EditorMenuTypes, boolean>`
- the `prosemirror-adapter` updates the `ActionBarItems` setting `selected` to true if the type corresponds with the active types returned by the` menu-state-tracking-plugin`
- the `menu-action-interaction-plugin` tracks user interactions with the action bar and handles getting the relevant command, creating a transaction, and dispatching the transaction.  

Fixes https://github.com/Lundalogik/crm-feature/issues/4135

How to test 

- Run: npm start
- Open the limel-text-editor and test activating different menu items. 
- For example, placing the cursor over and clicking into some text that is `Bold` should cause the `B` menu item to switch to an active state.

### DIFF
--- a/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
+++ b/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
@@ -8,7 +8,7 @@ import {
     Watch,
     h,
 } from '@stencil/core';
-import { EditorState, Plugin, PluginKey, Transaction } from 'prosemirror-state';
+import { EditorState, Transaction } from 'prosemirror-state';
 import { EditorView } from 'prosemirror-view';
 import { Schema, DOMParser } from 'prosemirror-model';
 import { schema } from 'prosemirror-schema-basic';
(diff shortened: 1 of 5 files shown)


src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -263,12 +267,34 @@ export class ProsemirrorAdapter {
             plugins: [
                 ...exampleSetup({ schema: this.schema, menuBar: false }),
                 keymap(this.menuCommandFactory.buildKeymap()),
-                this.createMenuStateTrackingPlugin(this.actionBarItems),
                 createLinkPlugin(this.handleNewLinkSelection),
+                createMenuStateTrackingPlugin(
+                    editorMenuTypesArray,
+                    this.menuCommandFactory,
+                    this.updateActiveActionBarItems,
```
LawrenceBorst, line 274: NIT. The `updateCallback` that we use now would be hard to test as it explicitly relies on `this`. We could make the callback inside it pure, returning the action bar items and use something like 

```
createMenuStateTrackingPlugin(
    editorMenuTypesArray,
    this.menuCommandFactory,
    (activeTypes: Record<EditorMenuTypes, boolean>) => {
        this.actionBarItems = this.updateActiveActionBarItems(activeTypes, this.actionBarItems);
    },
)
```
And even put the callback outside the class in a separate file

The same can be said for `this.handleNewLinkSelection`

But my main reason for pointing this out is that the adapter file is growing quite large

LawrenceBorst: On that note, I would probably like to play around with the code for the link and see how much we can move into the plugin

LawrenceBorst (commented):
I have gone through the code and compared it with the previous code. It all looks good to me, although I think it is a good idea to have docstrings in the different plugin factories that we have now.

Everything still works as it used to.

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
civing: We will anyways be dependent on this on your anonymous arrow function, so I'm not so sure what we gain?

civing (approved):

PR #3002, opened 2024-05-28
Fix text editor typing fast bug

LawrenceBorst:
Fixes Lundalogik/crm-feature#4142

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/text-editor/text-editor.tsx
+++ b/src/components/text-editor/text-editor.tsx
@@ -2,6 +2,10 @@ import { Component, Event, EventEmitter, Host, Prop, h } from '@stencil/core';
 import { FormComponent } from '../form/form.types';
 import { createRandomString } from 'src/util/random-string';
 import { Languages } from '../date-picker/date.types';
+import { debounce } from 'lodash-es';
+
+const DEBOUNCE_TIMEOUT = 300;
+
 /**
  * A rich text editor that offers a rich text editing experience with markdown support,
  * in the sense that you can easily type markdown syntax and see the rendered
@@ -231,6 +235,10 @@ export class TextEditor implements FormComponent<string> {
 
     private handleChange = (event: CustomEvent<string>) => {
         event.stopPropagation();
-        this.change.emit(event.detail);
+        this.changeEmitter(event.detail);
     };
+
+    private changeEmitter = debounce((value: string) => {
+        this.change.emit(value);
+    }, DEBOUNCE_TIMEOUT);
 }
--- a/src/components/text-editor/text-editor.tsx
+++ b/src/components/text-editor/text-editor.tsx
@@ -229,7 +229,7 @@ export class TextEditor implements FormComponent<string> {
         }
     };
 
-    private handleChange = () => (event: CustomEvent<string>) => {
+    private handleChange = (event: CustomEvent<string>) => {
         event.stopPropagation();
         this.change.emit(event.detail);
     };


src/components/text-editor/text-editor.tsx
```diff
@@ -229,8 +233,12 @@ export class TextEditor implements FormComponent<string> {
...
-    private handleChange = () => (event: CustomEvent<string>) => {
+    private handleChange = (event: CustomEvent<string>) => {
         event.stopPropagation();
-        this.change.emit(event.detail);
+        this.changeEmitter(event.detail);
     };
+
+    private changeEmitter = debounce((value: string) => {
+        this.change.emit(value);
+    }, DEBOUNCE_TIMEOUT);
```
adrianschmidt, line 243: Since we've had problems with limel-input-field which we think are _caused_ by the debounce, it's feels a little worrying that we are here _adding_ debounce to get rid of what sounds like very similar problems… 🤔

I'm not saying I have any ideas on how else to solve it, I'm just worried that the underlying problem here is a race-condition or something, and that debounce is really just a band-aid rather than a proper fix.

I'm not sure if I'm right, but my _instinct_ says that debounce is not a fix for anything but performance issues. If things aren't working properly (apart from how fast it is) without the debounce, adding debounce is probably just hiding the problem by making it occur less often, or much less likely to occur.

I'd be happy to have a look at this, if you'd like me to.

LawrenceBorst:
Not written here, but Pelle mentioned we should flush the debounce on blur

src/components/text-editor/text-editor.tsx
LawrenceBorst: I would like that a lot, if you find the time. I agree that debounce isn't really tackling the issue fundamentally. But should we consider adding the debounce as a temporary fix? Add a `// TODO`?

adrianschmidt: TODO's never get addressed… 😞

I'll see if I can take a stab at it, but yeah, if this is working, you don't have to wait for that 👍

adrianschmidt (changes requested):
I don't think this is needed after all. I _think_ I just fixed this problem while fixing the always-dirty bug…

adrianschmidt:
With what's currently in https://github.com/Lundalogik/lime-elements/pull/2999, I can only replicate the "disappearing" empty lines, and I say "disappearing" in quotes, because that's only happening because the editor is set to use markdown, and in the markdown mode, blank lines are not preserved. You could rather argue that the bug is that it's possible to get the wysiwyg editor to preserve the blank lines, instead of immediately removing them as soon as they are added.

If the editor is used in html mode, blank lines do not disappear, but more than than, if you are in markdown mode, and add a code block, and then add blank lines in there, they don't disappear, because inside code block, empty lines should be preserved.

john-traas:
We'll look into it further. I have a suspicion as well that this won't be needed after the `always-dirty` fix.

john-traas:
Closing this as the issue was fixed [here](https://github.com/Lundalogik/lime-elements/pull/2999)

PR #3005, opened 2024-05-28
feat(text editor): block pasted inline images

john-traas:
Updates the link plugin which handles pasting to purge images from content pasted into the text editor. 

Fixes https://github.com/Lundalogik/crm-feature/issues/4152

### DIFF
--- a/src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
+++ b/src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
@@ -188,13 +188,14 @@ const processPasteEvent = (
 
     const text = clipboardData.getData('text/plain');
 
-    if (!isValidUrl(text)) {
-        return false;
+    // Process as a link if the text is a valid URL
+    if (isValidUrl(text)) {
+        pasteAsLink(view, text);
+
+        return true;
     }
 
-    pasteAsLink(view, text);
-
-    return true;
+    return false;
 };
 
 const pasteAsLink = (view: EditorView, href: string) => {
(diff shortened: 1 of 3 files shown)


src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
civing, line 1: I'm thinking that we maybe should create a new plugin file for images? 🤔 
As long as we register that one before the link one it should be fine. What do you think?

src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
```diff
@@ -177,9 +177,66 @@ const processDoubleClickEvent = (
...
+        if (node.type.name === 'image') {
+            return true;
+        }
+
+        let found = false;
+        node.content.forEach((child) => {
+            if (isImageNode(child)) {
+                found = true;
+            }
+        });
```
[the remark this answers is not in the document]
john-traas: I think I was in a hurry and defaulted to using the variable. We could just return true within the forEach for the children and return false at the end instead.

src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
civing: What do you think about this @john-traas ? https://github.com/Lundalogik/lime-elements/pull/3005/commits/c650433b75e9a2831c6aa885c2426c04c85f3e2f

civing (approved):

civing (approved):

src/components/text-editor/prosemirror-adapter/plugins/link-plugin.ts
john-traas: It looks really good. Much cleaner as a separate plugin!

PR #3009, opened 2024-05-31
Text editor quality improvements

LawrenceBorst:
Fixes Lundalogik/crm-feature#4099

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
+++ b/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
@@ -125,8 +125,12 @@ export class ProsemirrorAdapter {
     private setupContentConverter() {
         if (this.contentType === 'markdown') {
             this.contentConverter = new markdownConverter();
-        } else {
+        } else if (this.contentType === 'html') {
             this.contentConverter = new HTMLConverter();
+        } else {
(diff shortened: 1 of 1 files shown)


src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -122,6 +122,10 @@ export class ProsemirrorAdapter {
         ];
     }
 
+    public disconnectedCallback() {
+        this.view.destroy();
```
adrianschmidt, line 126: Ah. I just realised that we need to re-initialise the editor in `connectedCallback`, in case someone disconnects it, and then connects it again (typically if you move the editor around in the DOM). We never do that ourselves, but since it's lime-elements, we need to cover such use-cases anyway.

However, I think we can push that off until after the MVP, so I'll just create an issue for it now.

adrianschmidt (approved):

PR #3017, opened 2024-06-04
feat(text editor): handle keyboard navigation for link menu

john-traas:
- Tabbing within the link menu no longer closes the menu
- Hitting enter will save the link
- Hitting escape will close the link menu

Fixes https://github.com/Lundalogik/crm-feature/issues/4164

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -5944,11 +5944,11 @@ export namespace Components {
      */
     interface LimelTextEditorLinkMenu {
         /**
-          * Open state of the dialog
+          * Open state of the link-menu dialog
          */
         "isOpen": boolean;
         /**
-          * Defines the language for translations. Will translate the translatable strings on the components.
+          * Defines the language for translations.
          */
         "language": Languages;
         /**
@@ -18822,11 +18822,11 @@ declare namespace LocalJSX {
      */
     interface LimelTextEditorLinkMenu {
         /**
-          * Open state of the dialog
+          * Open state of the link-menu dialog
          */
         "isOpen"?: boolean;
         /**
-          * Defines the language for translations. Will translate the translatable strings on the components.
+          * Defines the language for translations.
          */
         "language"?: Languages;
         /**
--- a/src/components/text-editor/link-menu/editor-link-menu.tsx
+++ b/src/components/text-editor/link-menu/editor-link-menu.tsx
@@ -1,4 +1,12 @@
-import { Component, Prop, h, Event, EventEmitter, State } from '@stencil/core';
+import {
+    Component,
+    Prop,
+    h,
+    Event,
+    EventEmitter,
+    State,
+    Element,
+} from '@stencil/core';
 import { EditorTextLink } from '../prosemirror-adapter/menu/types';
 import { Languages } from '../../date-picker/date.types';
 import translate from '../../../global/translations';
@@ -8,6 +16,7 @@ import {
     prependProtocol,
 } from '../../../util/link-helper';
 import { LimelInputFieldCustomEvent } from '../../../components';
+import { ENTER } from '../../../util/keycodes';
 
 /**
  * This component is a menu for editing a link in the text editor.
@@ -29,13 +38,12 @@ export class TextEditorLinkMenu {
 
     /**
      * Defines the language for translations.
-     * Will translate the translatable strings on the components.
      */
     @Prop({ reflect: true })
     public language: Languages = 'en';
 
     /**
-     * Open state of the dialog
+     * Open state of the link-menu dialog
      */
     @Prop({ reflect: true })
     public isOpen: boolean = false;
@@ -62,6 +70,18 @@ export class TextEditorLinkMenu {
     @State()
     private invalidLink: boolean = false;
 
+    private textInput: HTMLLimelInputFieldElement;
+    private saveButton: HTMLLimelButtonElement;
+
+    public componentDidLoad() {
+        if (this.textInput) {
+            const inputField = this.textInput.shadowRoot.querySelector('input');
+            requestAnimationFrame(() => {
+                inputField.focus();
+            });
+        }
+    }
+
     public render() {
         return [
             <limel-input-field
@@ -69,6 +89,10 @@ export class TextEditorLinkMenu {
                 value={this.link?.text || ''}
                 leadingIcon="text_cursor"
                 onChange={this.handleLinkTitleChange}
+                onKeyDown={this.handleKeyDown}
+                ref={(el) =>
+                    (this.textInput = el as HTMLLimelInputFieldElement)
+                }
             />,
             <limel-input-field
                 label={this.getTranslation('editor-link-menu.link')}
@@ -79,6 +103,7 @@ export class TextEditorLinkMenu {
                 invalid={this.invalidLink}
                 onChange={this.handleLinkValueChange}
                 onAction={this.handleLinkInputAction}
+                onKeyDown={this.handleKeyDown}
             />,
             <div class="actions">
                 <limel-button
@@ -91,6 +116,9 @@ export class TextEditorLinkMenu {
                     label={this.getTranslation('save')}
                     disabled={!this.link?.href || this.invalidLink}
                     onClick={this.handleSave}
+                    ref={(el) =>
+                        (this.saveButton = el as HTMLLimelButtonElement)
+                    }
                     slot="button"
                 />
             </div>,
@@ -111,14 +139,32 @@ export class TextEditorLinkMenu {
         return !!getHref(href);
     };
 
+    private handleKeyDown = (event: KeyboardEvent) => {
+        if (event.key !== ENTER) {
+            return;
+        }
+
+        if (this.saveButton) {
+            this.saveButton.focus();
+        }
+
+        event.preventDefault();
+        if (this.link?.href && !this.invalidLink) {
+            this.handleSave(event);
+        }
+    };
+
     private handleCancel = (event: MouseEvent) => {
         this.cancel.emit();
         event.stopPropagation();
     };
 
     private handleSave = (event: MouseEvent | KeyboardEvent) => {
-        this.save.emit();
         event.stopPropagation();
+
+        requestAnimationFrame(() => {
+            this.save.emit();
+        });
     };
 
     private handleLinkInputAction = (
--- a/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
+++ b/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
@@ -182,13 +182,16 @@ export class ProsemirrorAdapter {
                         'max-height': 'inherit',
                     }}
                 >
-                    <limel-text-editor-link-menu
-                        link={this.link}
-                        isOpen={this.isLinkMenuOpen}
-                        onLinkChange={this.handleLinkChange}
-                        onCancel={this.handleCancelLinkMenu}
-                        onSave={this.handleSaveLinkMenu}
-                    />
+                    {this.isLinkMenuOpen && (
+                        <limel-text-editor-link-menu
+                            link={this.link}
+                            isOpen={this.isLinkMenuOpen}
+                            onLinkChange={this.handleLinkChange}
+                            onCancel={this.handleCancelLinkMenu}
+                            onSave={this.handleSaveLinkMenu}
+                        />
+                    )}
+                    ,
                 </limel-menu-surface>
             </limel-portal>,
         ];
@@ -346,7 +349,9 @@ export class ProsemirrorAdapter {
         const { value } = event.detail;
 
         if (value === EditorMenuTypes.Link) {
-            this.isLinkMenuOpen = true;
+            requestAnimationFrame(() => {
+                this.isLinkMenuOpen = true;
+            });
 
             return;
         }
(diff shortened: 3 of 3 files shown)


src/components.d.ts
```diff
@@ -7007,11 +7007,11 @@ declare namespace LocalJSX {
      */
     interface LimelTextEditorLinkMenu {
         /**
-          * Open state of the dialog
+          * Open state of the link-menu dialog
```
Kiarokh, line 7010: Preferably all these documentation changes that have nothing to do with the commit, should be separately commited with a `docs` type. 😊  

This is because: 
- The reviewer will have an easier time figuring out what has been done
- if we ever want to revet a commit in future, we shouldn't lose the changes, just because they were together with another change that altered the functionality.

john-traas: I don't disagree, however those changes need to be in this commit. They are auto-generated and reflect changes made to the code. If we reverted any changes in this commit the documentation would need to be updated again and be part of the same commit.

etc/lime-elements.api.md
Kiarokh, line 1: please change the commit title from:
`feat(text editor): handle keyboard navigation for link menu`
to
`fix(text-editor): enable keyboard navigation using Tab, for the link menu`

This is not a `feature` for the "consumer". We always write these change logs, for the consumer of this package, not the end-users. Nothing is possible now after this commit, for consumers (in terms of features that the code offers), compared to before this commit. They're not able to further customize anything that they couldn't before.

john-traas: Ah I see. ⚡

src/components.d.ts
john-traas: I'm assuming that some changes made in a commit in another PR didn't make their way into the docs in that PR and that's why some of the updates to the docs aren't relevant to this PR.

src/components/menu-surface/menu-surface.tsx
Kiarokh, line 1: these changes for the menu surface should preferably be separately committed and explained. Since this component is a `@private` component, and the consumers don't even know that it exists. Thus the changes is normally mentioned as `chore` or `refactor` or similar, unless what you have done directly `fix`es an issue for another component that is in turn using `limel-menu-surface`, then they'll be `fix(that-component)`.

But I guess this is a 
```
chore(menu-surface): fixed this thing that was problematic blabla

some further clarification here...
for us, the owners and reviewers.
```

Change log doesn't show `chore` or `refactor` commits, so consumers won't see these ☝️ 
but we need to know what happened, both for the history and the review. 😊

src/components.d.ts
Kiarokh: Hm... ah, ok... that's weird that they didn't come here. 🤔 the rebase should update this file properly.


Sorry I commented in this auto-generated file. I just wanted a place to start a thread. 😅 
I was mainly trying to say if there is a docs change, which is not in direct relation to the `fix` or `feat`, then we need to commit it separately.

And also if the docs are auto generated and have nothing to do with your changes, they should also be marked as `docs` 👍

src/components/text-editor/link-menu/editor-link-menu.tsx
```diff
@@ -17,7 +26,7 @@ import { LimelInputFieldCustomEvent } from '../../../components';
  */
 @Component({
     tag: 'limel-text-editor-link-menu',
-    shadow: true,
+    shadow: { delegatesFocus: true },
```
Kiarokh, line 29: if this change didn't work or wasn't helping for the rest of what you did, just remove it. Otherwise it'll be misleading for the reviewer who will be trying to connect the dots.

src/components/text-editor/link-menu/editor-link-menu.tsx
```diff
@@ -62,13 +76,23 @@ export class TextEditorLinkMenu {
...
+    }
+
     public render() {
         return [
             <limel-input-field
                 label={this.getTranslation('editor-link-menu.text')}
                 value={this.link?.text || ''}
                 leadingIcon="text_cursor"
                 onChange={this.handleLinkTitleChange}
+                tabIndex={0}
```
Kiarokh, line 94: see https://github.com/Lundalogik/lime-elements/pull/3017/files#r1627222797

src/components/text-editor/link-menu/editor-link-menu.tsx
Kiarokh, line 29: I think this line, **or perhaps something else**, might be causing a problem:

https://github.com/Lundalogik/lime-elements/assets/35954987/8391c9a6-9ab1-4db1-a165-909bb83920fa

when you type in the second input field and pause for a second, the focus jumps to the first input field (perhaps as the component re-renders). If it's this line, the `delegateFocus` and `tabindex` are working. If it's not these, then the problem might be coming from `limel-menu-surface`.

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -325,7 +333,14 @@ export class ProsemirrorAdapter {
...
 
-    private handleCancelLinkMenu = () => {
+    private handleCancelLinkMenu = (event: CustomEvent<string>) => {
+        event.preventDefault();
+        event.stopPropagation();
+
+        if (event.detail === TAB) {
+            return;
+        }
+
```
Kiarokh, line 343: another viable solution might be simply getting rid of `limel-menu-surface` and having `limel-text-editor-link-menu` directly in `limel-portal` as the only child. If we do so, we lose the inbuilt dismiss fuctionalities, like scrim click (clicking outside to dismiss) or pressing <kbd>esc</kbd> to dismiss. 

However, give how much things we have changed in this PR, I believe that's easier to just get rid of  `limel-menu-surface` here and re-implement these functionalities decently for the link menu instead.

src/components/text-editor/link-menu/editor-link-menu.tsx
```diff
@@ -45,14 +53,20 @@ export class TextEditorLinkMenu {
      * (*Not* emitted when the consumer sets the `open`-property to `false`.)
      */
     @Event()
-    private cancel: EventEmitter<void>;
+    private cancel: EventEmitter<string>;
```
LawrenceBorst, line 56: But this still emits `void` I believe

LawrenceBorst (commented):
Code looks good. I found one issue, when clicking on a link, backspacing, and pressing enter, sometimes it does not alter the link at all

https://github.com/Lundalogik/lime-elements/assets/70023170/31f43c0e-c0d5-45d8-8608-fbda567d4e38

src/components/text-editor/link-menu/editor-link-menu.tsx
LawrenceBorst: Are you referring to the input field being shown as invalid? I think that is on main too 😅

https://github.com/orgs/Lundalogik/projects/149/views/56?pane=issue&itemId=66149636 there's an issue for it

john-traas: I think he's referring both to that and the focus being set on the first input. I think it's due to the menu-surface being re-rendered which triggers a re-render of the input fields, and this in turn then sets the focus to the first input field.

src/components/text-editor/link-menu/editor-link-menu.tsx
```diff
@@ -45,14 +53,20 @@ export class TextEditorLinkMenu {
...
      * Emitted when the menu is saved from inside the component.
      */
     @Event()
     private save: EventEmitter<void>;
 
+    /**
+     * Emitted when the menu should stay open.
+     */
+    @Event()
+    public stayOpen: EventEmitter<void>;
```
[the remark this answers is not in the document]
john-traas: Hmmm, in looking over it again I think I removed the code that utilised this. I'll have to do some fixups.

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -195,6 +197,12 @@ export class ProsemirrorAdapter {
         }
     }
 
+    private keepLinkMenuOpen = () => {
+        requestAnimationFrame(() => {
```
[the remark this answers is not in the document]
john-traas: I'm going to test removing this as I'm not sure if we need it after I made some changes to the `editor-link-menu.tsx` file.

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
john-traas: I'm willing to try different approaches, however if we remove the limel-menu-surface we would have to add a lot more code to get back that functionality.

src/components/text-editor/link-menu/editor-link-menu.tsx
john-traas: We need this to pass the event.key from the dismiss event. I'm not sure exactly how they are tied in but this error occurs if we don't add the `<string>` to the `EventEmitter`.

The Error:
`Type '(event: CustomEvent<string>) => void' is not assignable to type '(event: LimelTextEditorLinkMenuCustomEvent<void>) => void'.Types of parameters 'event' and 'event' are incompatible.Type 'LimelTextEditorLinkMenuCustomEvent<void>' is not assignable to type 'CustomEvent<string>'.Types of property 'detail' are incompatible.Type 'void' is not assignable to type 'string'.`

src/components.d.ts
john-traas: There's a PR in that makes a change to the workflow so that we won't have to update this file like this anymore.

src/components/menu-surface/menu-surface.tsx
john-traas: I will make this change into a separate commit once this PR has been approved and I squash the fixups.

src/components/text-editor/link-menu/editor-link-menu.tsx
john-traas: ⚡

john-traas: I've fixed the focus issue in a subsequent fixup.

john-traas:
> Code looks good. I found one issue, when clicking on a link, backspacing, and pressing enter, sometimes it does not alter the link at all
> 
>  Enregistrement.de.l.ecran.2024-06-05.a.10.52.29.mov

It certainly seems to be tied to a race condition. I think it has to do with the way changes to the link are handled. 
- changes to the link are emitted by the link-menu
- those changes are caught by the prosemirror-adapter and passed back to the link-menu
- the link-menu updates 
- the handle save function can emit the save event before the above chain of events has completed

src/components/text-editor/link-menu/editor-link-menu.tsx
Kiarokh: this shouldn't be needed right?

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -171,13 +172,16 @@ export class ProsemirrorAdapter {
...
+                            link={this.link}
+                            isOpen={this.isLinkMenuOpen}
+                            onLinkChange={this.handleLinkChange}
+                            onCancel={this.handleCancelLinkMenu}
+                            onSave={this.handleSaveLinkMenu}
+                        />
+                    )}
+                    ,
                 </limel-menu-surface>
             </limel-portal>,
```
Kiarokh, line 186: for the records, I tired getting rid of both `portal` and `menu-surface`, in the hope that `delegateFocus` would work, and that we may not need too complex scripts. This was the code:
```tsx
    public render() {
        return [
            <div id="editor" />,
            <limel-action-bar
                accessibleLabel="Toolbar"
                actions={this.actionBarItems}
                onItemSelected={this.handleActionBarItem}
            />,
            <limel-text-editor-link-menu
                style={{
                    display: this.isLinkMenuOpen ? 'flex' : 'none',
                }}
                link={this.link}
                isOpen={this.isLinkMenuOpen}
                onLinkChange={this.handleLinkChange}
                onCancel={this.handleCancelLinkMenu}
                onSave={this.handleSaveLinkMenu}
            />,
        ];
    }
```
and 
```scss
:host(limel-text-editor-link-menu) {
    position: absolute;
    z-index: 2;
    inset: 1.5rem 0 0 0;
    margin: auto;

    // display: flex; // this is set as an inline style in the component
    flex-direction: column;
    gap: 0.5rem;
    padding: 0.5rem;

    width: calc(100% - 2rem);
    height: fit-content;
    overflow: hidden auto;

    border-radius: 0.5rem;
    background-color: var(--lime-elevated-surface-background-color);
    box-shadow: var(--shadow-depth-16);
}

.actions {
    display: flex;
    justify-content: end;
    gap: 0.5rem;
}
```
image

However, this didn't really help. Because:
1. the link popover was always center aligned, but it would stretch out of the container, resulting in bad placement in the rest of the UI (e.g. when the editor is not so tall and is place at the bottom of the screen).
2. didn't solve the `delegateFocus` issue

Kiarokh: Now I'm gonna try to see if we can make it without `limel-menu-surface` but with the `portal`

john-traas: Thanks for looking into that @Kiarokh

LawrenceBorst (approved):

Kiarokh:
@john-traas I made this one a draft, in favor of https://github.com/Lundalogik/lime-elements/pull/3023 
I think we can bring the necessary parts from this PR to the other one.
I think we should also ignore that `delegateFocus` problem for now to be able to move on. This now offers the basic keyboard navigation needs. We can delegate focus later.

PR #3019, opened 2024-06-05
fix(input-field): improve synchronisation with invalid property

LawrenceBorst:
Fixes Lundalogik/crm-feature#4166

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/input-field/input-field.tsx
```diff
@@ -287,6 +287,8 @@ export class InputField {
     public componentDidUpdate() {
         if (this.invalid) {
             this.mdcTextField.valid = false;
+        } else {
+            this.mdcTextField.valid = true;
```
LawrenceBorst, line 291:
```suggestion
        this.mdcTextField.valid = !this.invalid
```

src/components/input-field/input-field.tsx
```diff
@@ -391,6 +393,8 @@ export class InputField {
 
         if (this.invalid) {
             this.mdcTextField.valid = false;
+        } else {
+            this.mdcTextField.valid = true;
```
LawrenceBorst, line 397:
```suggestion
        this.mdcTextField.valid = !this.invalid;
```

civing (approved):

vicgeralds:
This needs to be reverted as it completely breaks native validation, for example, when passing `type="number"` or `required`

src/components/input-field/input-field.tsx
vicgeralds: Just because invalid isn't set to true doesn't mean the field should be invalid.

If there's a need to clear the invalid state, we need to be sure that's indeed the intention, otherwise we break existing behaviour.

PR #3021, opened 2024-06-07
Stop tracking components.d.ts

civing:
Stop tracking components.d.ts since it's disturbing as *** when it keeps changing with incremental builds.
Since we do this change we also need to fix the quirks with the incremental builds when verifying API. So add another build step in the workflow.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


jgroth (approved):

adrianschmidt (approved):

adrianschmidt (changes requested):
Wait… don't we need to remove the file from the index too? 🤔

EDIT: Fixed.

adrianschmidt (approved):

PR #3022, opened 2024-06-07
fix(text-editor): ensure link popover opens in a decent location

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/4169

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components.d.ts
+++ b/src/components.d.ts
@@ -5350,21 +5350,33 @@ export namespace Components {
     /**
      * The portal component provides a way to render children into a DOM node that
      * exist outside the DOM hierarchy of the parent component.
+     * When the limel-portal component is used, it creates a new DOM node (a div element)
+     * and appends it to a parent element (by default, the body of the document).
+     * The child elements of the limel-portal are then moved from
+     * their original location in the DOM to this new div element.
+     * This technique is often used to overcome CSS stacking context issues,
+     * or to render UI elements like modals, dropdowns, tooltips, etc.,
+     * that need to visually "break out" of their container.
+     * Using this component, we ensure that the content is always rendered in the
+     * correct position, and never covers its own trigger, or another component
+     * that is opened in the stacking layer. This way, we don't need to worry about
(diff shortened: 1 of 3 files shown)


john-traas:
This seems to introduce some strange behaviour to the editor-link-menu. 

https://github.com/Lundalogik/lime-elements/assets/122509723/2d23ba64-ac9d-4114-9fb9-63e86868b828

LucyChyzhova (approved):

LucyChyzhova:
> This seems to introduce some strange behaviour to the editor-link-menu.
> 
>  Screen.Recording.2024-06-07.at.13.16.11.mov

yes, it what I noticed as well check the [comment](https://github.com/Lundalogik/crm-feature/issues/4169#issuecomment-2154587879), but actually that behaviour was not introduced here, this bug is a portal bug which we already had.

Kiarokh:
> This seems to introduce some strange behaviour to the editor-link-menu.
> 
>  Screen.Recording.2024-06-07.at.13.16.11.mov

Yes. Gonna solve this separately. We have the same issue on main. Seems to be a `limel-portal` but https://github.com/Lundalogik/crm-feature/issues/4170

Kiarokh:
additional not:

In the real app, this will "_almost_" never happen. This is because the text editor will never _(or should never)_ be as tall as the screen itself. However, we need to mitigate this issue from the component as well:

https://github.com/Lundalogik/crm-feature/issues/4171

john-traas:
Emails in a chain of replies can get quite long. I think in the email story this could happen quite easily?

LucyChyzhova:
yes, for sure. We are aware this case

PR #3024, opened 2024-06-09
fix(markdown): don't always force line break

civing:
If the newline character are preceeded with a backslash, don't force a hard line break since that combination already symbolizes a line break with the commonmark specification.

The markdown viewer are defaulting to force hard line breaks by replacing new lines (either \n or \r) with "  \n". Two spaces according to markdown syntax followed by a new line characters will render as a new line withing a block. The purpose of this is to replace new line breaks with a markdown specific new line.

However, when done in combination with an actual line break using a backslash things get weird. '\\\n' will be replaced by '\\  \n' which will then be rendered as both a backslash and a new line in markdown. This is not the intention of the original markdown.

fixes: Lundalogik/crm-feature#4175

You can test this by entering the follow in the limel markdown [example](https://lundalogik.github.io/lime-elements/versions/37.49.2/#/component/limel-markdown/).

```
# Hello, world!

This is **markdown**!\
\
\
\
testing
```

This should render as new lines and remove the backslash.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


LawrenceBorst (approved):
Tested with different values. Regex looks correct

adrianschmidt:
I've tried to test this in the _Text editor in markdown mode_ example in the PR docs, but I can't see how the PR fixes anything?

If I enter the text suggested, it of course renders backslashes, as I've written backslashes… If I just enter linebreaks, those are removed from the output… 🙁

LawrenceBorst:
> I've tried to test this in the _Text editor in markdown mode_ example in the PR docs, but I can't see how the PR fixes anything?
> 
> If I enter the text suggested, it of course renders backslashes, as I've written backslashes… If I just enter line breaks, those are removed from the output… 🙁

I had to add double slashes, since Javascript simplifies strings like
`test
\
\
\test
`
to `testtest`, using `\` to show newlines only aesthetically in the code, but not the string itself.

So I changed it to

`test
\\
\\
\test
`

But was this the problem? I didn't fully understand your comment.

adrianschmidt:
> But was this the problem? I didn't fully understand your comment.

I guess what I meant to say was that I am missing steps to reproduce the original bug, and I can't figure out how to reproduce it on my own.

civing:
What's fixed here is to honor the hard brakes with a backward slash in markdown. This is one of the alternatives of doing a hard break for a new line very similar to two spaces. https://spec.commonmark.org/0.31.2/#hard-line-breaks

You can reproduce this if you go to either `dev-feature` or the markdown prosemirror [example](https://prosemirror.net/examples/markdown/) and pressing `Shift-Enter` in between two lines. Prosemirror markdown will then insert `\\\n` to represent a hard line break similar to the html `<br>`

This is one example I just created with `Shift-Enter`: https://dev-feature.lime-crm.com/client/object/company/16194?tab=_activities

civing:
Sorry for the confusion regarding reproducing this. I believe I forgot to save the video explaining this to the linked issue but it should be there now. I hope things are clearer now.

civing:
> I've tried to test this in the _Text editor in markdown mode_ example in the PR docs, but I can't see how the PR fixes anything?
> 
> If I enter the text suggested, it of course renders backslashes, as I've written backslashes… If I just enter line breaks, those are removed from the output… 🙁

Are you not getting the same line breaks as I do? 🤔 
image

adrianschmidt:
I'm not sure if "don't always force line break" is a very helpful description of what this change does, but I didn't want to spend too much time on coming up with something better and then waiting for it to build again. For next time though 😉

PR #3027, opened 2024-06-10
fix(input-field): revert synchronisation with invalid property

vicgeralds:
This reverts commit e33141ad1c954cb116897f2359700da8b11217f8.

When rendering an input field with `invalid` unset or set to `false`, we would set `this.mdcTextField.valid = true` and ignore the native validation state, but might still render an error icon.

For example, it might look like this:

image

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -287,8 +287,6 @@ export class InputField {
     public componentDidUpdate() {
         if (this.invalid) {
             this.mdcTextField.valid = false;
-        } else {
-            this.mdcTextField.valid = true;
         }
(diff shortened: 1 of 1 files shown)


LawrenceBorst (approved):
Reverting to a known stable state. We will find another way to fix the associated bug

PR #3023, opened 2024-06-07
handle keyboard navigation problems and more

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/4164
fix https://github.com/Lundalogik/crm-feature/issues/4172

## Review:
- [ ] Commits are [atomic](https://seesparkbox.com/foundry/atomic_commits_with_git)
- [ ] Commits have the correct *type* for the changes made
- [ ] Commits with *breaking changes* are marked as such

### Browsers tested:
(Check any that applies, it's ok to leave boxes unchecked if testing something didn't seem relevant.)

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

PR #3030, opened 2024-06-11
feat(limel-text-editor): remove native url validation from the link input field

LawrenceBorst:
Fixes Lundalogik/crm-feature#4166

## How to Test

Go into one of the limel prosemirror adapter examples and add a link. Now try writing "test" -> hit tab -> writing "test". Then click outside the input field and note that it does not turn red (before the problem was that a phrase like "test" was not considered a valid URL).

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/text-editor/link-menu/editor-link-menu.tsx
```diff
@@ -113,7 +113,7 @@ export class TextEditorLinkMenu {
             <limel-input-field
                 label={this.getTranslation('editor-link-menu.link')}
                 value={this.link?.href || ''}
-                type="url"
```
Kiarokh, line 116: I don't know what problems you experienced, but this could have some accessibility benefits, as well as mobile-friendly on-screen keyboards. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url 

Maybe we should allow custom link-validations possibilities for the component, instead of removing it and changing the type to `text`.

Kiarokh: Also, i noticed that you cannot press <kbd>Enter</kbd> when the focus is on this "open link" button. Not sure if it's related, or of it's because of the changes introduced here: [`2da6dba` (#3023)](https://github.com/Lundalogik/lime-elements/pull/3023/commits/2da6dbac7aa4ec89a7e14d5304ce83743ecf30fd)
image

LawrenceBorst: > I don't know what problems you experienced, but this...

It's unfortunate that the input field serves double duties as a relative and absolute url validator. For relative urls, native url validation will tell us that the input field is invalid.

> Maybe we should allow custom link-validations possibilities for the component, instead...

I don't fully understand. Do you mean native client side form validation as we have now?

> Also, i noticed that...

I'll look into the open link button, it might be a small fix.

vicgeralds (approved):

src/components/text-editor/link-menu/editor-link-menu.tsx
LawrenceBorst: Or we could use an aria label 🤷🏽

LawrenceBorst: > Also, i noticed that you cannot press Enter when the focus is on this "open link" button. Not sure...

I couldn't reproduce this in latest main or on this branch 😅

vicgeralds: https://github.com/Lundalogik/lime-elements/assets/5563232/978ad15c-ecf7-462d-b3b4-abe6cc65bc64

vicgeralds: > > Maybe we should allow custom link-validations possibilities for the component, instead of removing it and changing the type to text.
> 
> There's already custom link validation in place.

But I'm unable to produce an invalid url. 😅 
And when I'm looking at the implementation of the method `validateLink` it only validates that it's a non-empty string really.

Still, I think it's quite difficult to produce an invalid link when accepting relative uri:s.

LawrenceBorst:
Merging now as we intend to release the latest editor ASAP due to it being publicly available in quick note.

PR #3035, opened 2024-06-13
adapt to the 2024 brand colors

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/style/color-palette-extended-light-mode-only.css
```diff
@@ -7,14 +7,24 @@
  */
 
 :root {
-    /* Lime Technologies Brand Colors (Do not have dark/light mode variants) */
+    /* Lime Technologies Brand Colors */
+    --lime-brand-color-lime-green: 190, 224, 52; /* #BEE034 */
+    --lime-brand-color-ocean-teal: 111, 205, 182; /* #6FCDB6 */
+    --lime-brand-color-aqua: 	166, 239, 255; /* #A6EFFF */
+    --lime-brand-color-bubble-gum: 	255, 166, 234; /* #FFA6EA */
+    --lime-brand-color-sunny-orange: 254, 176, 0; /* FEB000 */
```
adrianschmidt, line 15: Consistency nitpick 😄

```suggestion
    --lime-brand-color-sunny-orange: 254, 176, 0; /* #FEB000 */
```

`fixup! feat(style): add CSS variables for new Lime brand colors`

src/style/color-palette-extended.css
```diff
@@ -8,14 +8,25 @@
...
 :root {
     --limel-chip-progress-mix-blend-mode: multiply;
-    /* Lime Technologies Brand Colors (Do not have dark/light mode variants) */
+
+    /* Lime Technologies Brand Colors */
+    --lime-brand-color-lime-green: 190, 224, 52; /* #BEE034 */
+    --lime-brand-color-ocean-teal: 111, 205, 182; /* #6FCDB6 */
+    --lime-brand-color-aqua: 	166, 239, 255; /* #A6EFFF */
+    --lime-brand-color-bubble-gum: 	255, 166, 234; /* #FFA6EA */
+    --lime-brand-color-sunny-orange: 254, 176, 0; /* FEB000 */
```
adrianschmidt, line 17: Consistency nitpick 😄

```suggestion
    --lime-brand-color-sunny-orange: 254, 176, 0; /* #FEB000 */
```

`fixup! feat(style): add CSS variables for new Lime brand colors`

src/style/color-palette-extended.css
```diff
@@ -17,7 +25,6 @@
     --lime-brand-color-sellable-orange: 255, 112, 67; /* #ff7043 */
     --lime-brand-color-orange: 255, 176, 59; /* #ffb03b */
     --lime-brand-color-yellow: 255, 207, 61; /* #ffcf3d */
-    --lime-brand-color-lime-green: 102, 187, 106; /* #66bb6a */
```
adrianschmidt: Urgh…

I can see arguments for keeping the old colours under the old names, and I can see arguments for updating the old colour variables to use the new colours instead too. But keeping all but one, which instead is updated, is kind of the worst of both options… 😣

Also, if we learn from this, and assume that every few years we'll get a new "brand refresh" with slightly tweaked colours… should we perhaps include a year in the colour names or something?

Kiarokh: I don't think we need the years. If the same name changes its value may be it.
if a Lime CRM Configurator person has chosen a brand color to paint something, they have to leave with their choice I'd say. In the CRM, we have a color picker that doesn't show any of these brand colors, and they cannot and should not use them really. If they do, they're to blame, not us. They have explicitly in that case said that they want whatever that brand color gives them, and don't care about UI or accessibility, etc…

adrianschmidt: Ok 👍 🤷‍♂️ 😄

adrianschmidt (approved):

src/style/color-palette-extended.css
adrianschmidt: Although, to be fair, what you said above sounds like a pretty good argument for just not including the brand colours to begin with… 🤔

Issue #2485, opened 2023-09-18
Animate menu, picker, or slider surfaces


Kiarokh:
## New feature motivation

Currently, when you open a dropdown menu or a menu list, the content instantly pops-in.

Generally, instant changes in the UI are really hard for users to notice and result in a bad experience. If you aren't actively paying attention, you may miss what happened, or what initiated this new element that poped up in the UI. You may wonder, what's the anchor point of this dropdown list. Did I click/tap correctly? Did it come from the icon which was beside where I clicked? 

These subtle things affect the perception of the users.

To avoid these perception problems, good design systems usually take advantage of UI clues such as: 

1. those little arrows which are pointing to the trigger element, like the one in popper.js 
	<img width="300px" alt="image" src="https://github.com/Lundalogik/crm-feature/assets/35954987/f27bbf55-c6df-4bc6-afa4-08fe09f1e815">

2. And also slightly animate the menu or popover content when it appears:
	

https://github.com/Lundalogik/lime-elements/assets/35954987/84f23c2a-210c-4eb3-a196-e6d70af7e193

## New feature implementation

For some of our components like popover, and tooltip, we could handle the animation ourselves. Because they are built from scratch by us.

But since our Menu and Select components are based on MDC, it might be better to use their inbuilt features? (or not, if we're gonna get rid of them at some point)

### If we decide to rely on MDC's styles:
1. We need to use the same timings and styles on our own components. Making a `mixin` for this would be necessary then.
2. MDC [offers CSS classes](https://github.com/material-components/material-components-web/blob/master/packages/mdc-menu-surface/README.md#css-classes) that can be added to `limel-menu-surface` that will help animating the appearance and disappearance of the menu content. 
<img width="1028" alt="image" src="https://github.com/Lundalogik/lime-elements/assets/35954987/9ed7ed80-d4c5-4e36-bab5-730faea72989">
Currently we are missing these classes.
⚠️ But the important thing is that these class names should be removed once the animation is over. The animation duration is set in the CSS for `height` and `transform` (and also `opacity`) properties on `mdc-menu-surface` div.
<img width="411" alt="Screenshot 2023-09-18 at 11 30 37" src="https://github.com/Lundalogik/lime-elements/assets/35954987/f908746c-c08c-4bfd-b31c-6739ac9ea8a4">
We should take the longest one of them as a reference for when to remove the CSS classes. (I guess that'd be enough).
3. We should set the correct anchor corner for the animations to play correctly. 

[...]

PR #3037, opened 2024-06-14
animate portal when it appears and disappears

Kiarokh:
fix https://github.com/Lundalogik/lime-elements/issues/2485

From this PR:

https://github.com/user-attachments/assets/1cca01dc-943e-4b5d-aab2-2448477bc0ad

https://github.com/user-attachments/assets/cd8b01b2-6f35-49d1-b66d-4416b29bfad2

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/portal/partial-styles/animate-portal.scss
```diff
@@ -10,6 +10,6 @@
 
     &.is-visible {
         opacity: 1;
-        display: block;
+        display: inline-block;
```
Kiarokh, line 13: image

src/components/portal/partial-styles/animate-portal.scss
```diff
@@ -10,6 +10,6 @@
 
     &.is-visible {
         opacity: 1;
-        display: block;
```
Kiarokh: image

Kiarokh:
```scss
// TODO: add these once the hiding animation starts working
// &[data-popper-placement='left-start'] {
//     transform-origin: right top;
// }
// &[data-popper-placement='left'] {
//     transform-origin: right center;
// }
// &[data-popper-placement='left-end'] {
//     transform-origin: right bottom;
// }
// &[data-popper-placement='top-start'] {
//     transform-origin: left bottom;
// }
// &[data-popper-placement='top'] {
//     transform-origin: center bottom;
// }
// &[data-popper-placement='top-end'] {
//     transform-origin: right bottom;
// }
// &[data-popper-placement='right-start'] {
//     transform-origin: left top;
// }
// &[data-popper-placement='right'] {
//     transform-origin: left center;
// }
// &[data-popper-placement='right-end'] {
//     transform-origin: left bottom;
// }
// &[data-popper-placement='bottom-start'] {
//     transform-origin: left top;
// }
// &[data-popper-placement='bottom'] {
//     transform-origin: center top;
// }
// &[data-popper-placement='bottom-end'] {
//     transform-origin: right top;
// }

```

src/components/portal/portal.tsx
```diff
@@ -175,9 +176,7 @@ export class Portal {
     @Watch('visible')
     protected onVisible() {
         if (!this.visible) {
-            this.hideContainer();
-            this.styleContainer();
```
Kiarokh, line 177: There are some inline styles on `<div class="limel-portal--container"` that control the placement (such as `position`, `transform`, `inset`, and `z-index`) in relation to the trigger element. They are being removed immediately when the portal gets hidden (e.g. scrim-clicking the menu-surface). 

Since I need these styles to persist during the "closing" animation (when the `is-closing` class is temporarily applied), I need to ensure that those inline styles remain intact until the animation finishes. Otherwise, the surface (portal-container) will move out of the viewport, as soon as the portal is closed.

So I had to make sure to keep all placement-related styles intact during the closing animation.
And only after the is-hiding animation completes, clear the inline styles.

src/components/portal/portal.tsx
```diff
@@ -212,24 +211,6 @@ export class Portal {
...
-        }
-
-        Array.from(this.container.children).forEach((element: HTMLElement) => {
-            const parent = this.parents.get(element);
-            if (!parent) {
-                return;
-            }
-
-            parent.appendChild(element);
-        });
```
Kiarokh, line 227: Most of the code changes in this commit is done after consultation with Chat GPT, after a long back and forth discussion about what I exactly wanted to happen. So I'm not sure about this code removal. When I asked why:

> The section you removed from the original removeContainer function was responsible for moving the elements inside the portal (children) back to their original parent in the DOM before the portal container itself was removed. Specifically, this part of the code:
> 
> **What it did:**
> - ` this.parents.get(element)`: It retrieves the original parent of the element, stored earlier in a `WeakMap`.
> - `parent.appendChild(element)`: It re-attaches each child back to its original parent in the DOM.
> **Consequences of removing it:**
> - If you are no longer moving elements out of their original position in the DOM (e.g., not manipulating their original DOM tree, just hiding and showing the portal), then this removal is safe.
> - If you are still moving elements out of the DOM (e.g., you are attaching elements into the portal container from other parts of the DOM), you might now be "losing" those elements when the container is removed, as they won’t be re-attached to their original parents.
> **When it's safe to remove:**
> - If your current code only manipulates the portal container (showing and hiding) without moving its children around in the DOM, then this removal is safe and aligns with your refactor.
> **When it's not safe:**
> - If you still need to move DOM elements into the portal container and expect them to return to their original place in the DOM when the portal is closed, this removal could lead to issues such as:
> - Elements disappearing permanently from the DOM when the container is removed.
> - Elements not being properly returned to their original positions.
> - If your current portal logic does not involve moving children from the original DOM tree, then you're fine. Otherwise, it may require revisiting this mechanism.


@jgroth this sounded similar or related to what you tried to do in this PR https://github.com/Lundalogik/lime-elements/pull/3182/ which is why I wanted us to wait with that one.

src/components/portal/portal.tsx
```diff
@@ -212,41 +211,21 @@ export class Portal {
         this.parent.appendChild(this.container);
     }
 
-    private removeContainer() {
```
jgroth, line 215: I don't think we should remove this function and just rename it to `hideContainer` Those two are doing different things. This one will be called when the element is disconnected from the DOM, and `hideContainer` will be called when visibility changes

src/components/portal/portal.tsx
jgroth: I don't see why we should wait with the other one. It took 30 seconds to open the admin page due to 2000+ portals moving elements around in the DOM. It seems more important to fix that than adding animations. Don't get me wrong, animations are nice but kind of pointless if you cant even load the page in the first place :smile:

src/components/portal/partial-styles/portal-container.scss
```diff
@@ -13,3 +13,59 @@
...
+        > * {
+            transform-origin: $origin;
+        }
+    }
+}
+
+.limel-portal--container {
+    > * {
+        // The element that the consumer component puts inside the portal
+        animation: fade-in 0.18s cubic-bezier(0.25, 0.8, 0.25, 1) forwards;
```
Kiarokh, line 50: #### Why animating the element inside the container, instead of the container itself?

This is because the container has some inlined-styles such as `transform` that are dynamically set by portal. I am trying to animate the same thing, which interferes with those inlined styles. So the next best thing is to animate the content of the container instead.

src/components/portal/portal.tsx
Kiarokh: ok. so merge that one, and I'll try to remake this based on the updated code.

src/components/portal/portal.tsx
```diff
@@ -252,6 +251,21 @@ export class Portal {
...
+        }
+
+        this.container.classList.add(IS_HIDING_CLASS);
+        this.styleContainer();
+
+        setTimeout(() => {
+            this.container.classList.remove(IS_HIDING_CLASS);
+            this.container.classList.remove(IS_VISIBLE_CLASS);
+            this.destroyPopper();
+        }, hideAnimationDuration);
```
civing, line 266: These timeouts can prove tricky. I'm not sure if this is a valid use case and what should be expected. But if you click it three times it ends up in a strange state of being opened but it's now showing.

https://github.com/user-attachments/assets/bc359aad-a44a-4725-bd5c-1e4d04812131

src/components/portal/portal.tsx
```diff
@@ -236,7 +236,6 @@ export class Portal {
             parent.appendChild(element);
         });
 
-        this.hideContainer();
```
civing, line 239: Hmm, why are we not hiding it any more when removing the container?

src/components/portal/portal.tsx
civing: Just to be clear, I'm not sure we should consider this scenario.

src/components/portal/portal.tsx
Kiarokh: I wrote some text as commit message about why. But here is how I understand it. Does it make sense?
---

What is the purpose of `removeContainer()`?
1. It detaches the elements inside the portal from the container and reattaches them to their original parents.
2. It removes the portal container itself from the DOM.

I felt it doesn't make sense to hide the container in `removeContainer()`. If we’re removing the container from the DOM, the visibility (IS_VISIBLE_CLASS) of the container no longer matters. Once removed from the DOM, the `hideContainer()` method doesn’t need to run.

Am I wrong?

src/components/portal/portal.tsx
Kiarokh: good point. 🤔 but we need the timeout for letting the animations run. Is there a better way to do this?

Maybe I can come up with a more "native" way, using this: https://developer.mozilla.org/en-US/docs/Web/CSS/transition-behavior

But I have to check it again. It can be a bit tricky sometimes.

Kiarokh: Oh I actually can't. I tried this before, but ended up adding this timeout. The thing is that I need to keep those dynamically applied inlined styles on the container while the animation is running. The javascript removes them and everything gets meaningless.

Kiarokh: I tied to fix it using a state variable. But I couldn't.

src/components/portal/portal.tsx
civing: It makes sense 👍

civing (approved):

PR #3041, opened 2024-06-20
Fix date picker custom formatter called with extra args

vicgeralds:
The date formatter function used with the date picker was called lots of times, and sometimes directly by flatpickr that passed its extra args `format` and `locale`.

Now we only call the custom formatter to format the value of the date picker, passing only the value argument, and not for formatting aria labels that flatpickr sometimes renders for pickable dates.

Also updates the value of a closed date picker when set programmatically 

fixes: Lundalogik/crm-feature/issues/3182

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/date-picker/date-picker.tsx
+++ b/src/components/date-picker/date-picker.tsx
@@ -6,7 +6,6 @@ import {
     Element,
     EventEmitter,
     Event,
-    Watch,
 } from '@stencil/core';
 import { createRandomString } from '../../util/random-string';
 import { isAndroidDevice, isIOSDevice } from '../../util/device';
@@ -150,9 +149,6 @@ export class DatePicker {
     private host: HTMLLimelDatePickerElement;
 
     @State()
-    private formattedValue: string;
-
-    @State()
     private internalFormat: string;
     @State()
     private showPortal = false;
@@ -183,8 +179,6 @@ export class DatePicker {
         this.useNative = !this.readonly && (isIOSDevice() || isAndroidDevice());
 
         this.updateInternalFormatAndType();
-
-        this.formattedValue = this.formatValue(this.value);
     }
 
     public componentWillUpdate() {
@@ -209,7 +203,7 @@ export class DatePicker {
                     label={this.label}
                     helperText={this.helperText}
                     required={this.required}
-                    value={this.formattedValue}
+                    value={this.formatValue(this.value)}
                     type={this.nativeType}
                     onChange={this.nativeChangeHandler}
                 />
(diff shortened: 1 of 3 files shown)


src/components/date-picker/pickers/Picker.ts
```diff
@@ -115,6 +117,24 @@ export abstract class Picker {
...
+        });
+
+        return (date: Date | null, format: string): string => {
+            if (!date) {
+                return '';
+            }
+
+            if (format === ARIA_DATE_FORMAT) {
+                return longDateFormat.format(date);
+            }
```
adrianschmidt, line 130: I think think might need an explanatory comment in the code… 😅

adrianschmidt: Ah… if I understand the code correctly, flatpickr is supplying the `format` argument, and when it tries to format a value in order to use it for an ARIA attribute, the `format` argument's value will be `'F j, Y'`. Correct?

Do you know if the "long" format we return is equivalent to `'F j, Y'`?

src/components/date-picker/date-picker.tsx
```diff
@@ -149,9 +148,6 @@ export class DatePicker {
     @Element()
     private host: HTMLLimelDatePickerElement;
 
-    @State()
-    private formattedValue: string;
```
adrianschmidt, line 153: Yeah, it seems a lot more resonable to only format when rendering. Good change! 👍

adrianschmidt (approved):

src/components/date-picker/pickers/Picker.ts
adrianschmidt: I felt like this was a relatively small thing, so I merged the fix anyway. But if you feel like you can write a helpful comment, or refactor with a named function in a useful way, please feel encouraged to do so! 😊

vicgeralds: > Ah… if I understand the code correctly, flatpickr is supplying the `format` argument, and when it tries to format a value in order to use it for an ARIA attribute, the `format` argument's value will be `'F j, Y'`. Correct?
> 
> Do you know if the "long" format we return is equivalent to `'F j, Y'`?

Correct. Yep, the long date style corresponds to "January 1, 2000" but localized. So the idea is that this will be more fit as an accessible label than a numeric date, and including time is also not meaningful here.

From what I could see those labels are only used when picking the dates of a month when the type of pickers is either "date" or "datetime".

PR #3045, opened 2024-06-20
fix(date-picker): avoid closing before input click

vicgeralds:
When the date picker is already open, clicking the input field (or tabbing to focus the clear button) would first hide the date picker on mouse down and then show it again on mouse up.

This is fixed by ignoring mousedown events from the input field. We also don't need to listen to keydown events, since we're already hiding the date picker on blur.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/date-picker/date-picker.tsx
+++ b/src/components/date-picker/date-picker.tsx
@@ -299,9 +299,6 @@ export class DatePicker {
         document.addEventListener('mousedown', this.documentClickListener, {
             passive: true,
         });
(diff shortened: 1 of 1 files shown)


LawrenceBorst (approved):
Tested. Looks good to me.

PR #3050, opened 2024-06-26
Fix table update data

vicgeralds:
Avoids scrolling the table when updating existing rows with new data.

If no rows are removed, we can call `tabulator.updateOrAddData` which apparently works better than `replaceData` when we want don't the scrolling position to change.

To make this work with active row and selection, rows are tracked by id when available. Tabulator supports indexing data by an id property, which by default is `id`: https://tabulator.info/docs/6.2/data#row-index

Fixes: https://github.com/Lundalogik/crm-feature/issues/3925

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


jgroth (approved):

src/components/table/table.tsx
```diff
@@ -263,8 +263,12 @@ export class Table {
     }
 
     @Watch('data')
-    protected updateData(newData = [], oldData = []) {
-        if (isEqual(newData, oldData)) {
+    protected updateData(newData: RowData[] = [], oldData: RowData[] = []) {
```
hannahu, line 266: This commit makes the sorting buggy when you press a column name. I guess we can't use the `updateOrAddData` since it has no new rows in the case of re-ordering ASC/DESC.

hannahu: See https://github.com/Lundalogik/crm-feature/issues/4247 for bug

PR #3056, opened 2024-07-01
fix(editor menu): allow block quotes to be toggled

john-traas:
Certain node types, in this case blockQuote, wrap the parent node, in this case a paragraph. We need to check if the blockType is active and then lift the selection out of the wrap. 

Not allowing for block quotes to be created across paragraphs is intentional (eg: write multiple paragraphs, try and select them all and wrap all in a block quote). 

Fixes https://github.com/Lundalogik/crm-feature/issues/4214


LawrenceBorst (approved):
Looks good to me. Did some manual testing as well

PR #3059, opened 2024-07-04
fix(text editor): ensure the link menu is correctly displayed

john-traas:
The z-index being set on the lime portal rendered from the prosemirror adaptor was incorrect and causing the link menu to be rendered behind other elements. 

Fixes https://github.com/Lundalogik/crm-feature/issues/4215

### DIFF
--- a/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
+++ b/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
@@ -184,7 +184,6 @@ export class ProsemirrorAdapter {
                 openDirection="top"
                 inheritParentWidth={true}
                 anchor={this.actionBarElement}
-                containerStyle={{ 'z-index': 1 }}
             >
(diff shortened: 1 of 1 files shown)


Kiarokh:
a nitpicky thing, but please change the commit message: `fix(link menu): remove errant z-index style`

to something like: `fix(text-editor): ensure the link menu is correctly displayed`

`link menu` (or more correctly written as a scope in a commit message: `text-editor-link-menu`) is a private component. It means consumers shouldn't even know or care it exists. So the scope for the consumer is `text-editor`, because that's where they see and experience the fix. We never use `@private` components as the scope.

Also, unless the change is `refactor` or `chore`, we should try to explain what kind of buggy behavior has been fixed, or what kind of feature-y thing has been added.

Then the commit message can have the details, such as it was about a crazy `z-index` or whatever, which caused X or Y.

Kiarokh (approved):

PR #3063, opened 2024-07-12
Add docs and design guide for text editor

john-traas:
This PR adds two relevant code comments and some documentation/design-guidelines for the `text-editor`.

The goal is to give an overview of our approach to using prosemirror and enable someone who hasn't worked with it before to understand how we're using it. There are plenty of relevant links to actual prosemirror documentation.

If there are things that are unclear or don't seem to make sense or that seem to be un-necessary additions, let me know. 😅 

Fixes https://github.com/Lundalogik/crm-feature/issues/4261

The generated docs link is outdated. You can view the files [here](https://github.com/Lundalogik/lime-elements/blob/add-docs-and-design-guide-for-text-editor/src/design-guidelines/text-editor/editor-development.md)


civing:
Noob question: how do I view these documentations? I can't see them in the generated docs linked in this PR

john-traas:
Oops, you can see them rendered as markdown [here](https://github.com/Lundalogik/lime-elements/blob/add-docs-and-design-guide-for-text-editor/src/design-guidelines/text-editor/editor-development.md)

civing (commented):
Awesome! I guess you are intentially not putting any guides here for them to show up under "Design Guidelines"?
https://github.com/Lundalogik/lime-elements/blob/9c428533af7452864f05bfef9ae6e630ada7896b/guides.ts#L10

john-traas:
Actually I wasn't certain about adding it there but it probably should be. Would make it easy for someone to read them.

civing:
> Actually I wasn't certain about adding it there because it didn't seem to fit with the other design-guidelines. But adding it there would make it easy for someone to read them.

@Kiarokh or @adrianschmidt what is your opinion about these design guidelines? They look to me more like notes to developers rather than design guidelines for users...? Should we have yet another section like "developer guidelines"

Kiarokh:
I haven't read the guidelines you wrote. But generally we never publicly expose what external libraries we use in our components. So it's better that developer guidelines are kept inside the components themselves. Sometimes we have `@private` (internal) components that are just there for us/developers. Example: `limel-helper-line`. They are shown in the docs with examples, if you go there. I think in those cases (components that will never be public), it can be fine to expose such info in the published docs.

But in general, public docs are just for consumers, not for us or contributors

civing:
> I haven't read the guidelines you wrote. But generally we never publicly expose what external libraries we use in our components. So it's better that developer guidelines are kept inside the components themselves. Sometimes we have `@private` (internal) components that are just there for us/developers. Example: `limel-helper-line`. They are shown in the docs with examples, if you go there. I think in those cases (components that will never be public), it can be fine to expose such info in the published docs.
> 
> But in general, public docs are just for consumers, not for us or contributors

Maybe we can keep them as is for now but not expose them in any form. I don't wanna throw away all this nice information. I will approve it, let's decide later on what to do with it.

Issue #1268, opened 2021-06-02
limel-select with `multiple=true`: selecting an option always scrolls the list of options to the top


adrianschmidt:
## Current behavior

https://user-images.githubusercontent.com/154725/120448819-395a7780-c38c-11eb-843f-dd9e47d39392.mov

1. Use a `limel-select` with `multiple=true` and enough options to make the dropdown scrollable.
2. Open the dropdown.
3. Scroll down a bit.
4. Select or deselect an option.

- The list of options scrolls to the top.

## Expected behavior

The list of options should stay at whatever scroll position it was on when the selection was made.

## Environment

- **lime-elements** version: 32.2.0
- Framework used: Reproducible in docs if enough options are added to the _Select multiple values_ example
- Logs:

PR #3065, opened 2024-07-15
fix(select): keep the scrolling position when multiple select true

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/1268
fix: https://github.com/Lundalogik/crm-feature/issues/2808

https://github.com/user-attachments/assets/199408d0-dbb4-4a6b-a9dd-261422b5d37b

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [x] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [x] Safari
- [x] Edge

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/select/select.tsx
```diff
@@ -255,11 +255,30 @@ export class Select {
     ) {
         event.stopPropagation();
 
+        // Construct the selector for the list element
```
hannahu, line 258: Remove this comment, not necessary and decreases readability ☺️

src/components/select/select.tsx
```diff
@@ -255,11 +255,30 @@ export class Select {
...
+        const list: HTMLElement = document.querySelector(selector);
+
+        if (!list) {
+            return;
+        }
+
+        const scrollPosition = list.scrollTop;
+        const delay = 8;
+
+        // Verify if we can manually manipulate the scroll position
```
hannahu, line 269: Same here. If you need to clarify something to the reviewer, just add a comment here on GH.

src/components/select/select.tsx
```diff
@@ -255,11 +255,28 @@ export class Select {
     ) {
         event.stopPropagation();
 
+        const selector = `#${this.portalId} limel-menu-surface limel-list`;
```
john-traas, line 258:
```suggestion
        const selector = `#${this.portalId} limel-menu-surface limel-list ul`;
```
I believe it's the actual ul component we should be getting the scroll position for.

src/components/select/select.tsx
```diff
@@ -255,11 +255,28 @@ export class Select {
...
         if (isMultiple(event.detail)) {
             const listItems: ListItem[] = event.detail;
             const options: Option[] = listItems.map((item) => item.value);
             this.change.emit(options);
 
+            // Use setTimeout to restore scroll position after a slight delay
+            setTimeout(() => {
+                list.scrollTop = scrollPosition;
+            }, delay);
+
```
john-traas, line 279:
```suggestion
            requestAnimationFrame(() => {
                list.scrollTop = scrollPosition;
            });
```
We just set the scroll position on the next animation frame.

LucyChyzhova: ah!!! so cool, I tried to use `requestAnimationFrame` instead but it didn't work for me, it could be because I missed `ul`..🤔

john-traas: When I first tried `requestAnimationFrame` it didn't work. I think it was adding the `ul` to the selector.

LucyChyzhova:
I tested and unfortunately the new changes brought "the jump" back, so the scrolling position got reset. I will keep my initial commit.

src/components/select/select.tsx
```diff
@@ -255,11 +255,28 @@ export class Select {
...
         if (isMultiple(event.detail)) {
             const listItems: ListItem[] = event.detail;
             const options: Option[] = listItems.map((item) => item.value);
             this.change.emit(options);
 
+            // Use setTimeout to restore scroll position after a slight delay
+            setTimeout(() => {
+                list.scrollTop = scrollPosition;
+            }, delay);
```
adrianschmidt, line 278: Using any delay other than 0 introduces a race condition. Either this works with a delay of 0, or we can't rely on this solution as a fix, because it depends on a race condition. (Although, in this case, if it usually works, and only sometimes doesn't, that's a better behaviour for the user than the current behaviour.)

I assume you've already tried this with a delay of 0, but I'll try it myself just in case, and if it doesn't work, I might play around and see if I can get it to work in some other way.

src/components/select/select.tsx
adrianschmidt, line 1: > The current solution doesn't really fix the issue. In my testing it seemed like at times there were two separate elements being scrolled at the same time.

Hm… I'm not seeing anything like that on my machine, and it's working well for me (although with a slight flicker sometimes, but that's not the end of the world).

I wonder if the behaviour you are seeing might be related to the previously mentioned race condition created by setTimeout, or if it's something else… 🤔

src/components/select/select.tsx
LucyChyzhova: > (Although, in this case, if it usually works, and only sometimes doesn't, that's a better behaviour for the user than the current behaviour.)

I agree!

> I assume you've already tried this with a delay of 0, but I'll try it myself just in case, and if it doesn't work, I might play around and see if I can get it to work in some other way.

Thank you, Adrian! 🙏

src/components/select/select.tsx
LucyChyzhova: > Hm… I'm not seeing anything like that on my machine, and it's working well for me (although with a slight flicker sometimes, but that's not the end of the world).

Same for me 🤔

src/components/select/select.tsx
adrianschmidt: ⚡

adrianschmidt: Please help test my fixup.

It's a pity that John isn't here to help test it, since he's the one who was having problems with the previous implementation… 😕

src/components/select/select.tsx
adrianschmidt: Ah! Yeah, there is clearly something weird going on in Edge on Windows:

https://github.com/user-attachments/assets/7732864b-3216-44a7-86ca-89eeee1f5f13

(I have tried in Chrome on macOS, where I'm not having any problems. I haven't tried any other browsers yet.)

src/components/select/select.tsx
john-traas: I am actually reading the thread because I'm curious to see if your fixup will solve it. When I still had the issues, it was when I had selected multiple items, but it wasn't consistent. I tried the requestAnimationFrame but didn't try the setTimeout within the rAF as you did. :D 


I'll test it now. It's too hot to be outside so I'm in the living room with a fan a full blast and reading through issues in the cooldown board 😆

src/components/select/select.tsx
john-traas: In Edge on macOS it also has odd behaviours.

https://github.com/user-attachments/assets/18e40e3a-1bbd-474e-bf24-32f0a81288be

adrianschmidt: I think the problem is that we've added the overflow stuff here: https://github.com/Lundalogik/lime-elements/pull/3065/files#diff-817817636ecff6902e33d23a4566f3ce0c76c4b8c1604695c95b7acc3c40b066R30-R33

I'm pretty sure that the drop down could already overflow and become scrollable. And since we haven't changed anything there, that's still in effect. For some reason, that is getting triggered in Edge. My initial suspicion would be that it happens on systems where scroll bars are taking up space, like they do on Windows. I'm not having any problems in Edge Version 127.0.2651.74 (Official build) (arm64) on my Mac, so I'm going to venture a guess that you have your system set to always show scroll bars, @john-traas?

adrianschmidt: Ah, OK, now I saw the screen recording. Hmm… that's odd. It doesn't seem like your scrollbars are always showing, but once there shows up a different scroll bar, so there seems like the problem is the same: two nested elements both being scrollable.

LucyChyzhova: and mine 🤔 - all good

https://github.com/user-attachments/assets/194eb2d6-2b0d-4ebb-87ba-b95ce2fe3df1

john-traas: I have it set to 'Automatically based on mouse or trackpad'

adrianschmidt: Ah! (Yes, again 😄)

I can recreate the problem in Chrome on my Mac as well. It's just less obvious that there are actually two separate elements being scrolled, since only one of the scrollbars are visible at a time, and both look the same (apart from usually being of different length).

LucyChyzhova: and Chrome: 

https://github.com/user-attachments/assets/92de7bf6-47f5-4bd5-a7ad-c6efd341f901

LucyChyzhova: however, in Safari it does jump 

https://github.com/user-attachments/assets/2dae1db0-773a-41ea-bcd2-7d8adde8dfdc

adrianschmidt: Try scrolling the component to roughly the middle of the screen before you open the drop down. Notice in my screen recording that I scroll all the way down, then pause a little bit, then scroll down again. The second time I get _further_ down. If I've scrolled that second time, and then select an option, the element that is scrolled that second time gets scrolled to the top, but not the element that is scrolled the first time.

https://github.com/user-attachments/assets/c3631e3a-375f-44e2-94f6-f4796afa6e65

adrianschmidt: I'm looking into finding the older code that made the dropdown scrollable, unless you already know where it is, @LucyChyzhova?

LucyChyzhova: no, but I'm trying to find as well.

adrianschmidt: It seems to be the mdc-menu-surface:

Screenshot 2024-07-29 at 15 36 18

adrianschmidt: I'd guess that MDC expects `.mdc-menu-surface` to be the element that is scrolling. I don't know if that actually matters or not… 🤷‍♂️

Although, I can't see why we couldn't just find that element and work on that instead of the `ul`? 🤔

adrianschmidt: I _think_ I've solved it. Pushing a fixup in a minute.

LucyChyzhova: testing now!

LucyChyzhova: YESS! Works everywhere for me!  💥

src/components/select/select.tsx
```diff
@@ -255,14 +255,14 @@ export class Select {
...
+        const selector = `#${this.portalId} .mdc-menu-surface`;
+        const menuSurface: HTMLElement = document.querySelector(selector);
 
-        if (!list) {
+        if (!menuSurface) {
             return;
         }
 
-        const scrollPosition = list.scrollTop;
+        const scrollPosition = menuSurface.scrollTop;
```
LucyChyzhova, line 265: ah!! I should not use list there! but `menuSurface`

adrianschmidt: Look at the whole fixup. That's "the same" variable, I just renamed it, because I changed so that it wasn't holding the list, but the menu-surface.

src/components/select/select.tsx
adrianschmidt: I'll squash and push 👍

src/components/select/select.tsx
adrianschmidt: Here's what the fixup looked like before I squashed it:

image

LucyChyzhova: right, missed the top part.

adrianschmidt (approved):

PR #3067, opened 2024-07-16
fix(table): make sorting work again

hannahu:
The sorting broke in [this commit](https://github.com/Lundalogik/lime-elements/pull/3050/commits/3472118224e917a01a34aa163c846da0578236df), since we can't call `this.tabulator.updateOrAddData()` when only the sorting has changed. Then we need to do `this.tabulator.replaceData()`. We still should use `this.tabulator.updateOrAddData()` for Victor's case, so we still wanna use that when no rows have been removed.

In other words:
- _Update_ data when only the content of rows changes, and the order remains the same, and otherwise _replace_ data.

I also put `this.pool.releaseAll` before the checks since it clears the state in the pool before making data updates, and should not only be when doing replace.

fix: Lundalogik/crm-feature#4247

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


jgroth (approved):
I could not reproduce it consistently, but at least I could not reproduce it at all with the fix :smile:

Issue #3070, opened 2024-07-19
Limel-button: loading result needs color


LucyChyzhova:
## Current behavior
when a limel-button is not primary, and you want to use the loading feature, the tick icon (success)  is not visible

https://github.com/user-attachments/assets/43bf9e7f-e0bb-4bb5-a3a8-148c2fe00803

1. …

## Environment

- **lime-elements** version: 
- Framework used: 
- Logs:

PR #3071, opened 2024-07-19
fix(button):  make loading feedback visible

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/3070

## Review:

https://github.com/user-attachments/assets/b1094d36-df37-4408-9b03-bd2e4034e7ab

go to `limel-example-button-composite` example 
1. uncheck primary
2. check/uncheck loading state

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -106,7 +106,7 @@ svg {
     width: functions.pxToRem(30);
 
     line {
-        stroke: rgb(var(--color-white));
+        stroke: rgb(var(--contrast-600));
         stroke-width: 2;
     }
 }


src/components/button/button.scss
```diff
@@ -106,7 +106,7 @@ svg {
     width: functions.pxToRem(30);
 
     line {
-        stroke: rgb(var(--color-white));
+        stroke: rgb(var(--contrast-700));
```
jgroth, line 109: When the button is primary, the checkbox is now grey instead of white, is this intended?
For non-primary buttons it looks nice though :slightly_smiling_face:

LucyChyzhova: yes, it was intended, otherwise - hard to see

jgroth: For the white button yes, but the green button has a grey checkmark now instead of white :smile:

LucyChyzhova: yes... true!! let's have it as 600 instead.

https://github.com/user-attachments/assets/454b19d7-b393-4089-b89c-6ee80af1ef17

jgroth (approved):

PR #3084, opened 2024-08-05
Allow snackbars to be stacked

jgroth:
related to https://github.com/Lundalogik/crm-feature/issues/3355
fix https://github.com/Lundalogik/crm-feature/issues/3356

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [x] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [x] Chrome on Android
- [x] iOS


src/components/snackbar/snackbar.scss
```diff
@@ -9,44 +9,46 @@
...
+    background: none;
+    border: none;
+    inset: unset;
+    top: 0;
+    right: 0;
+    // margin: unset;
+    // top: var(--snackbar-top, auto);
+    // right: var(--snackbar-right, 0);
+    // bottom: var(--snackbar-bottom, 0);
+    // left: var(--snackbar-left, 0);
```
Kiarokh, line 23: we shouldn't remove this feature

jgroth: How should we implement it? What is the need of the consumer?

src/components/snackbar/snackbar.scss
```diff
@@ -9,44 +9,46 @@
     box-sizing: border-box;
 }
 
-.mdc-snackbar {
-    top: var(--snackbar-top, auto);
-    right: var(--snackbar-right, 0);
-    bottom: var(--snackbar-bottom, 0);
-    left: var(--snackbar-left, 0);
```
Kiarokh, line 16: nobody has used these except ourselves in the webclient.

src/components/snackbar/container.ts
```diff
@@ -0,0 +1,59 @@
...
+     * Add a new element to the container
+     *
+     * @param snackbar - element to add
+     */
+    public add(snackbar: HTMLLimelSnackbarElement) {
+        const popover = this.getPopover(snackbar);
+
+        // Stencil does not seem to recognise the existance of showPopover
+        // @ts-ignore
+        popover?.showPopover();
```
adrianschmidt, line 20: 😲

Yeah, the Stencil version we are using is over a year old, and `popover` is marked as "Baseline 2024 - Newly Available" on MDN:

image

Myself, I went looking for how we define `showPopover` on the element, and after not finding _anything_, I finally googled for "js popover", and then I went 🤯… 😂

src/components/snackbar/snackbar.tsx
```diff
@@ -1,15 +1,20 @@
 import { Languages } from '../date-picker/date.types';
-import { MDCSnackbar, MDCSnackbarCloseEvent } from '@material/snackbar';
```
adrianschmidt, line 2: # 🥳

src/components/snackbar/container.ts
jgroth: Yeah it would be nice to use this for `limel-popover` and `limel-tooltip` as well in the future, but I think we need anchor positioning to be baseline as well first 
https://developer.chrome.com/blog/anchor-positioning-api

Kiarokh: and limel-modal before all of the above. that one doesn't even use portal

src/components/snackbar/snackbar.tsx
```diff
@@ -99,89 +102,92 @@ export class Snackbar {
...
+
+        this.closing = true;
+
+        setTimeout(() => {
+            this.open = false;
+            container.remove(this.host);
+            this.hide.emit();
+            this.offset = 0;
+        }, hideAnimationDuration);
+    };
```
adrianschmidt, line 161: These timeouts can conflict. If you open a snackbar in the docs, then click the button to close it, then open it again somewhat quickly, the first closing timeout is still active, and will close the snackbar really quickly after it opens the second time.

If we store the timeout handle from the `show` method, we can cancel that timeout in the `close` function. Then we'd avoid this problem.

I'm not sure if there can be some broken interaction if the `show` method is called during the timeout set in the `close` function as well, but it might be best to make sure any active timeout gets cancelled in that case too.

src/components/snackbar/snackbar.tsx
```diff
@@ -99,89 +102,92 @@ export class Snackbar {
...
-        this.mdcSnackbar.open();
+        if (this.timeout) {
+            setTimeout(
+                this.close,
+                Math.max(this.timeout - hideAnimationDuration, 0),
+            );
+        }
     }
 
+    private close = () => {
```
adrianschmidt, line 148: Since this is private, I'd prefer to move it below the render function.

Although, for the sake of symmetry, we might want to make this a public `@method` too?

src/components/snackbar/snackbar.tsx
```diff
@@ -99,89 +103,99 @@ export class Snackbar {
...
                     '--snackbar-timeout': `${this.timeout}ms`,
+                    '--snackbar-distance-to-top-edge': `${this.offset}px`,
+                }}
+                class={{
+                    open: this.open,
+                    'is-closing': this.closing,
                 }}
+                id={this.snackbarId}
+                role={this.open ? 'status' : undefined}
+                aria-relevant={this.open ? 'additions' : undefined}
```
Kiarokh, line 177: it seems we have always had a bad implementation of MDC, as these attributes remained static in the DOM in the previous versions.

Information about these accessibility things is generally really confusing when I read about them online. But I guess this is the correct implementation when I check this one, which has some example: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA22.html

They say:
> This is done by adding `role="status"` to the element that contains the [status message](https://www.w3.org/TR/WCAG21/#dfn-status-messages).

src/components/snackbar/snackbar.tsx
```diff
@@ -99,89 +103,99 @@ export class Snackbar {
...
             >
-                <div
-                    class="mdc-snackbar__surface"
-                    role="status"
-                    aria-relevant="additions"
-                >
-                    <div class="mdc-snackbar__label" aria-atomic="false"></div>
+                <div class="surface" aria-atomic="false">
```
Kiarokh, line 179: However, I don't really understand why MDC has `aria-atomic` explicitly set to `false`. So I left it as is.

w3 says:
> The role of status also has a default aria-atomic value of true, so that updates to the container marked with a role of status will result in the AT presenting the entire contents of the container to the user, including any author-defined labels (or additional nested elements).

src/components/snackbar/container.ts
jgroth: Yeah the dialog we can do already today :smiley:

Kiarokh:
@jgroth I changed `fix https://github.com/Lundalogik/crm-feature/issues/3355`. Setting timeout to `0` to make the snackbar "persistent" was already possible in the previous MDC-based component as long as I remember, since we were working with this PR [`afd00a3` (#2403)](https://github.com/Lundalogik/lime-elements/pull/2403/commits/afd00a37fdf58c1b556dd7fbad7d5e3d063e5f34) . 

After you current PR is merged, I'll go back to that one, which has a new example for Showcasing a persistent snackbar. Right now I don't remember why we got stuck with it then and never merged it. But I'll have a look again, when this one is approved merged.

src/components/snackbar/snackbar.tsx
jgroth: I would prefer not to have any `@Method` at all and instead have a property such as `open`, but that would be a breaking change so I left it out for now

src/components/snackbar/snackbar.tsx
```diff
@@ -37,9 +43,7 @@ import translate from '../../global/translations';
  * @exampleComponent limel-example-snackbar-dismissible
  * @exampleComponent limel-example-snackbar-with-action
  * @exampleComponent limel-example-snackbar-with-changing-messages
- * @exampleComponent limel-example-snackbar-positioning
  */
-
 @Component({
```
Kiarokh, line 47: can't comment on the relevant prop, but I just realized that `public multiline: boolean;` doesn't really do anything. I don't think it ever did, even in the MDC version. What do you say? Remove it?

Kiarokh: oh, had a look at this old PR which was never merged: [`1ee660b` (#2403)](https://github.com/Lundalogik/lime-elements/pull/2403/commits/1ee660b1b2c6c66021b3b9b8e0584e8baab36c97) and saw that I added this description for the prop:
```md
     * When `true`, the layout of the Snackbar will change.
     * The action and the dismiss buttons will be rendered in a
     * row below the text, instead of being rendered as trailing elements.
     * This might be useful for Snackbars with very long messages.
     ```

Maybe this was something we got out of the box from MDC? I think we should update the new styles to accommodate for the `multiline` prop. This will be helpful when we allow having multiple actions in the snackbar too.

File: src/components/snackbar/snackbar.tsx (hunk 5 above)
Comment (jgroth) [reply src/components/snackbar/snackbar.tsx]: d776e6e47

File: src/components/snackbar/snackbar.tsx (hunk 9 above)
Comment (jgroth) [reply src/components/snackbar/snackbar.tsx:47]: I think I remember that `multiline` was a property on the MDC snackbar, but maybe they have removed it in later versions? :thinking:

Comment (Kiarokh) [reply src/components/snackbar/snackbar.tsx:47]: It's weird to keep it, as we can achieve the same results dynamically with CSS only to be honest. I think we can handle this in another PR, get rid of this prop, and make a breaking change or something

File: src/components/snackbar/container.ts (hunk 3 above)
Comment (Kiarokh) [reply src/components/snackbar/container.ts:20]: let me know if you wanna start on it ;)

File: src/components/snackbar/snackbar.tsx (hunk 9 above)
Comment (jgroth) [reply src/components/snackbar/snackbar.tsx:47]: Seems it was only used to set this class

image

File: src/components/snackbar/snackbar.scss (hunk 2 above)
Comment (adrianschmidt) [reply src/components/snackbar/snackbar.scss:16]: Well… it _is_ an open source component library, so there may be people outside the company that has.

But, it won't lead to build errors if they set these variables. It will simply not have any effect. Instead, we provide this nice new functionality where multiple snackbars can be shown at once. I think that's acceptable to do without a major version bump.

File: src/components/snackbar/snackbar.tsx (hunk 7 above)
Comment (adrianschmidt) [reply src/components/snackbar/snackbar.tsx:183]: Yeah, I've noticed that these things often seem broken. MDC is supposed to set the correct values as you interact with different components, but I'm guessing the problem might be that as you interact with things, we re-render them, and then the values get re-set to what is only intended to be starting values.

File: src/components/snackbar/snackbar.tsx (hunk 9 above)
Comment (adrianschmidt) [reply src/components/snackbar/snackbar.tsx:47]: Let's mark it as deprecated.

Comment (jgroth) [reply src/components/snackbar/snackbar.tsx:47]: c9babb7f4

File: src/components/snackbar/snackbar.tsx [hunk 10]
```diff
@@ -74,6 +78,7 @@ export class Snackbar {
 
     /**
      * Whether to show the snackbar with space for multiple lines of text
+     * @deprecated
```
adrianschmidt, line 81:
```suggestion
     * @deprecated Setting this property no longer has any effect. The property will be removed in a future major version.
```

adrianschmidt (dismissed):

adrianschmidt:
Very nice work! 🥳 🙌

adrianschmidt (approved):

PR #3088, opened 2024-08-07
refactor(portal): use Popover API instead of moving elements in the DOM

jgroth:
This makes the portal component work without moving elements in the DOM, instead it now uses the Popover API native to the browser

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/portal/portal.tsx
```diff
@@ -67,26 +57,23 @@ export class Portal {
...
     /**
-     * The `parent` property specifies the parent element where the content
-     * of the portal will be moved to.
-     * By default, it is set to `document.body`, meaning the content
-     * will be appended as a child of the body element in the DOM.
-     * If you want the content to be appended to a different element,
-     * you can specify that element by setting this property.
-     * Please note that the specified parent element should exist
-     * in the DOM at the time of rendering the portal.
+     * @deprecated no longer needed or used, will be removed in a future release
```
Kiarokh, line 76: it's only us using limel-portal inside lime elements. Nobody knows about this `@private` component. You don't need to `deprecate` it. ;)

jgroth: We can deprecate it anyway :smile:

src/components/portal/portal.tsx
```diff
@@ -67,26 +57,23 @@ export class Portal {
     public position: 'fixed' | 'absolute' = 'absolute';
 
     /**
-     * A unique ID.
+     * @deprecated no longer needed or used, will be removed in a future release
      */
     @Prop({ reflect: true })
     public containerId: string;
```
Kiarokh, line 63: I think we should keep the unique `id`, and use it with `aria-controls` attribute on the trigger element. We have been doing this manually in many components. But it's better if portal takes care of this _if possible_, instead of us remembering and doing it where we use portal.

Anyhow, requiring an `id` is good I would say.

But we have had a problem in the previous implementation of the id. The id used to be put both on `<limel-portal id="uniqueID">` element itself (which wraps the trigger element), and also on the `<div class="limel-portal--container" id="uniqueID">` which is the content of the top layer. Then the trigger element would have the `aria-controls` referring to that `id`, like `<limel-button aria-controls="uniqueID">`. This was wrong. Two elements can't have the same id. Only the "surface" or top-layer should have the id. This is something we can fix now here.

src/index.html
```diff
@@ -17,7 +17,6 @@
             --kompendium-example-font-family: ui-sans-serif, system-ui, sans-serif;
             --kompendium-example-font-size: 0.875rem;
             --kompendium-example-line-height: normal;
-            --limel-portal-font-family: var(--kompendium-example-font-family);
```
Kiarokh, line 20: see [`fba69a2` (#3088)](https://github.com/Lundalogik/lime-elements/pull/3088/commits/fba69a24503dfb872de467f76992c4005da829dd#r1713300185)

src/components/popover/popover.tsx
```diff
@@ -148,8 +148,8 @@ export class Popover {
     }
```
Kiarokh, line 149: shouldn't we remove things like 
```tsx
        const popoverZIndex = getComputedStyle(this.host).getPropertyValue(
            '--popover-z-index',
        );
```
already in this PR?

jgroth: We probably can

src/components/portal/portal.tsx
jgroth: > I think we should keep the unique id, and use it with aria-controls attribute on the trigger element. We have been doing this manually in many components. But it's better if portal takes care of this if possible, instead of us remembering and doing it where we use portal.

I want to remove this component entirely eventually so I think it's better to not add more functionality to it now :slightly_smiling_face:

src/components/popover/popover.tsx
Kiarokh: ⚡

src/components/portal/portal.tsx
```diff
@@ -67,26 +57,23 @@ export class Portal {
     public position: 'fixed' | 'absolute' = 'absolute';
 
     /**
-     * A unique ID.
+     * @deprecated no longer needed or used, will be removed in a future release
```
Kiarokh, line 60:
```suggestion
     * A unique `id` that you should use together with `aria-controls` attribute 
     * on the trigger element to make the portal accessible.
```

src/components/portal/portal.tsx
Kiarokh: I still think we shouldn't deprecate this, as it forces the consumer to do the right thing. See my suggestion: https://github.com/Lundalogik/lime-elements/pull/3088/files#r1713476180

I think when we throw away this component, we can think about how to solve this later, not to make an accessibility bug.

jgroth: But it's not used anywhere in the component :thinking:

Kiarokh: ok, fair enough. Then what about the last two fixups I made?

src/components/portal/portal.tsx
```diff
@@ -120,6 +120,13 @@ export class Portal {
             return;
         }
 
+        if (this.containerId) {
+            // eslint-disable-next-line no-console
+            console.warn(
+                'The `containerId` prop is no longer needed. It is deprecated and will be removed in a future release.',
```
Kiarokh, line 126: there is only [one instance](https://github.com/search?q=org%3ALundalogik+%3Climel-portal&type=code) that someone uses `limel-portal`. I'd love to remove this `@private` prop, instead of deprecating it. But I guess this is more by the book.

jgroth: I'm not sure what you mean, you want to make the component public?
Not sure how they found out about this private component :smile:

src/components/portal/portal.tsx
jgroth: It looks great! :partying_face:

src/components/portal/portal.tsx
Kiarokh: No no. I just mean when components are private, we don't need to deprecate anything. We can just fix the whatever we have removed, so that the consumer components of that deprecated thing keep working as they should.

Now in this case, with my last fixup I'd have loved to remove the `containerID` prop. But since someone is using `limel-portal` in their project, then we will break their shit. So we can't just remove the prop. However, we can add a console warning instead and hope they see it. 😅

jgroth: I was just lazy and deprecated it instead of updating all the code that used it :smile: If you now have removed all those usages, we can just remove it instead :smile:

Kiarokh: I see 😝 
Yes. I've fixed them all. Pushed an update fixup

Kiarokh (approved):

civing (approved):

Issue #2893, opened 2024-04-09
limel-dialog: dialog should be rendered in a portal


adrianschmidt:
This issue was first reported as "when using a dialog inside another dialog, the whole dialog goes black". However, that is just a symptom of the underlying problem. The underlying problem is that the dialog created by `limel-dialog` is not rendered in a portal. It's frankly quite surprising that this hasn't been causing many problems, but I think the mitigating factor is that most dialogs is created via the Lime CRM platform's `DialogRenderer`, rather than by using `limel-dialog` directly, as using the `DialogRenderer` service means the Lime CRM Web Client will create the dialog at the top level of the app.

**_Below is the original description of this issue._**

## Current behavior

When using a dialog inside another dialog, the whole dialog goes black.

<img width="992" alt="image" src="https://github.com/Lundalogik/lime-elements/assets/154725/8619d2a2-c20c-4713-b112-89fa46caf3b9">

This isn't really a "normal" use case, but we have had a few instances of people having problems with it.

The cause of the problem, if I've understood it correctly, is the shadow that is used to indicate that the dialog content is overflowing, and that it's possible to scroll.

Reproduced in https://github.com/Lundalogik/lime-elements/pull/2600

```tsx
import { Component, h, State } from '@stencil/core';
@Component({
    tag: 'limel-example-dialog',
    shadow: true,
})
export class DialogExample {
    @State()
    private isOpen = false;
    public render() {
        return [
            <limel-button
                primary={true}
                label="Open"
                onClick={this.openDialog}
            />,
            <limel-dialog open={this.isOpen} onClose={this.closeDialog}>
                <p>This is a simple alert-dialog.</p>
                <limel-dialog open={this.isOpen} onClose={this.closeDialog}>
                    <p>This is a dialog, inside another dialog!</p>
                    <limel-button
                        label="Oh no..."
                        onClick={this.closeDialog}
                        slot="button"
                    />
                </limel-dialog>
                ,
                <limel-button
                    label="Ok"
                    onClick={this.closeDialog}
                    slot="button"
                />
            </limel-dialog>,
        ];
    }
    private openDialog = () => {
        this.isOpen = true;
    };
    private closeDialog = () => {
        this.isOpen = false;
    };
}
```

## Expected behavior

Both dialogs should render normally.

[...]

PR #3093, opened 2024-08-09
 refactor(dialog): use native dialog element and remove MDC

jgroth:
fix https://github.com/Lundalogik/lime-elements/issues/2893

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dialog/dialog.tsx
+++ b/src/components/dialog/dialog.tsx
@@ -1,16 +1,14 @@
 import { DialogHeading, ClosingActions } from './dialog.types';
-import { MDCDialog } from '@material/dialog';
 import {
     Component,
     Element,
     Event,
     EventEmitter,
     h,
+    Listen,
     Prop,
     Watch,
 } from '@stencil/core';
-import { isEqual } from 'lodash-es';
-import { dispatchResizeEvent } from '../../util/dispatch-resize-event';
 import { createRandomString } from '../../util/random-string';
 
 /**
@@ -91,17 +89,9 @@ export class Dialog {
     @Element()
     private host: HTMLLimelDialogElement;
 
-    private mdcDialog: MDCDialog;
-
     private id: string;
 
     private showFooter = true;
-
-    constructor() {
-        this.handleMdcOpened = this.handleMdcOpened.bind(this);
-        this.handleMdcClosed = this.handleMdcClosed.bind(this);
-        this.handleMdcClosing = this.handleMdcClosing.bind(this);
-    }
 
     public connectedCallback() {
         this.initialize();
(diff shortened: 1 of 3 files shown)


src/components/dialog/partial-styles/show-hide-animations.scss
```diff
@@ -0,0 +1,34 @@
...
+    transform: $initial-transform;
+
+    &[open] {
+        opacity: 1;
+        transform: translate3d(0, 0, 0) scale(1);
+
+        @starting-style {
+            opacity: 0;
+            transform: $initial-transform;
+        }
```
Kiarokh, line 18: this is supposed to make sure that the animation applies both when the dialog is opened and closed, but for some reason, every time I have tried this new thing, it only works one-way. 

a good tutorial about this can be found here: 

https://www.youtube.com/watch?v=vmDEHAzj2XE&t=231s

src/components/dialog/partial-styles/show-hide-animations.scss
Kiarokh, line 1: these animations work fine. But as long as you close the dialog by keyboard `Esc` or by pressing a close button inside the dialog. Scrim-clicking causes a glitch in the close animation, and the dialog jumps down first, and then transitions away when closing. It seems as if an element is added to the DOM for a fraction of a second while the dialog is being closed, which pushes the dialog down. Or when scrim-clicking, some sort of temporary style is added just before the dialog is closed.

I'm investigating to see why this happens.

src/components/dialog/dialog.tsx
```diff
@@ -1,16 +1,14 @@
 import { DialogHeading, ClosingActions } from './dialog.types';
-import { MDCDialog } from '@material/dialog';
```
Kiarokh, line 2: I found an edge case which will force us to fix things later, as soon as we have merged this PR.

Lime consultants have put many dialogs inside dialogs unfortunately. So basically, you can open a dialog on top of an existing dialog, by pressing a trigger element which is inside the first dialog. This has traditionally worked (with a buggy rendering in Chromium: https://github.com/Lundalogik/lime-elements/issues/2893).

to fake/test this behavior, paste this code in one of the example files:
```tsx
@Component({
    tag: 'limel-example-dialog',
    shadow: true,
})
export class DialogExample {
    @State()
    private isOpen = false;
    public render() {
        return [
            <limel-button
                primary={true}
                label="Open"
                onClick={this.openDialog}
            />,
            <limel-dialog open={this.isOpen} onClose={this.closeDialog}>
                <p>This is a simple alert-dialog.</p>
                <limel-dialog open={this.isOpen} onClose={this.closeDialog}>
                    <p>This is a dialog, inside another dialog!</p>
                    <limel-button
                        label="Oh no..."
                        onClick={this.closeDialog}
                        slot="button"
                    />
                </limel-dialog>
                <limel-button
                    label="Ok"
                    onClick={this.closeDialog}
                    slot="button"
                />
            </limel-dialog>,
        ];
    }
    private openDialog = () => {
        this.isOpen = true;
    };
    private closeDialog = () => {
        this.isOpen = false;
    };
}
``` 

However, the problem now in this PR is that, as soon as you click the first trigger, it will open the nested dialog too, and the result will be like this:

image


Previously, those buggy dialogs would still be usable if users used Firefox or Safari. However, with this PR, we will break their nested dialogs, if it's merged in the current state.

jgroth: > However, the problem now in this PR is that, as soon as you click the first trigger, it will open the nested dialog too, and the result will be like this:

That depends on how you have implemented your code :smile: In your example you are using `isOpen` to control both dialogs. If you have multiple dialogs you need a separate state for each one. Nothing in this PR changes that-

> Previously, those buggy dialogs would still be usable if users used Firefox or Safari. However, with this PR, we will break their nested dialogs, if it's merged in the current state.

How will they break?

src/components/dialog/partial-styles/show-hide-animations.scss
```diff
@@ -0,0 +1,48 @@
...
+        transition-behavior: allow-discrete;
+        background-color: rgb(var(--color-black), 0);
+    }
+
+    &[open] {
+        --limel-dialog-display-speed: 0.4s;
+        opacity: 1;
+        transform: translate3d(0, 0, 0) scale(1);
+
+        @starting-style {
```
jgroth, line 22: SonarCloud is complaining about `transition-behavior` and `@starting-style`, it seems they are not implemented in Firefox yet. Will it work acceptable in Firefox anyway?

src/components/dialog/dialog.tsx
Kiarokh: Oh silly me. I just copy/pasted some example code which I did for visual testing in the other issue like 10 years ago: https://github.com/Lundalogik/lime-elements/issues/2893 assuming that it's fine 🙈 😅 

Anyhow, it seems to be working fine in this PR now.

src/components/dialog/partial-styles/show-hide-animations.scss
Kiarokh: no it's still [not fully supported yet](https://caniuse.com/?search=starting-style), but the transition will be fine. we basically don't get any transition at all in firefox at the moment. But it's ok I'd say. The alternative would be dynamically adding and removing classes, when the portal closes. Like what we did for the snackbars. But I'm a bigger fan of the browsers' builtin features, not such scripts that try to compensate for them. I think we can let this go. What do you say?

jgroth: It's fine, I just wanted to make sure you were aware :smile: 
I can silence the warnings from SonarCloud

Kiarokh (approved):

PR #3097, opened 2024-08-13
feat(text-editor): add `ui` prop

Kiarokh:
fix https://github.com/Lundalogik/crm-feature/issues/4281

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [x] iOS


src/components/text-editor/text-editor.tsx
```diff
@@ -111,6 +112,16 @@ export class TextEditor implements FormComponent<string> {
...
 
+    /**
+     * Defines how the editor should look
+     * - `default`: The default look of the editor
+     * - `compact`: A more compact look of the editor, useful for
+     * displaying on mobile devices for instance space is limited.
+     * It will remove the toolbar, unless the editor is in focus.
+     */
+    @Prop({ reflect: true })
+    public ui?: 'default' | 'compact' = 'default';
```
jgroth, line 123: Does this need to be a `@Prop`? It seems to only affect styling, can we not do the same thing with a class?

Kiarokh: We could for now. But I am not a big fan of that approach, unless we want to hide it from consumers.
Because:
1. if we add more functionality or complexities in future, having a prop makes it way easier to control these things. 
2. it's hard to communicated and test these things in the docs. Also it's harder for consumers to add classes. I'd prefer that they play around with props instead of concepts that are "styles-related".

If there is something that we only internally want to use, I think we can do that with the class.

src/components/text-editor/text-editor.scss
```diff
@@ -48,10 +50,21 @@
 }
 
 :host(limel-text-editor:focus-within),
-:host(limel-text-editor:hover) {
+:host(limel-text-editor:not([ui='compact']):hover) {
```
LucyChyzhova, line 53: should we say 'default' instead of not([ui='compact'])?

LucyChyzhova (dismissed):
Works as expected! great job! 💥 (I just left a small comment, otherwise let's merge it! 😉

src/components/text-editor/text-editor.scss
Kiarokh: no. because we want this when the ui is not compact but is hovered. there is something for that when it is compact and hovered further down

src/components/text-editor/text-editor.tsx
```diff
@@ -111,6 +112,16 @@ export class TextEditor implements FormComponent<string> {
...
 
+    /**
+     * Defines how the editor should look
+     * - `standard`: The default look of the editor
+     * - `minimal`: A more compact look of the editor, useful for
+     * displaying on mobile devices for instance space is limited.
+     * It will remove the toolbar, unless the editor is in focus.
+     */
+    @Prop({ reflect: true })
+    public ui?: 'standard' | 'minimal' = 'standard';
```
Kiarokh, line 123: I renamed these to keep them in sync with feed-items' `ui` types

src/components/text-editor/text-editor.scss
LucyChyzhova: I meant like this: 
:host(limel-text-editor:([ui='default']):hover)

src/components/text-editor/text-editor.tsx
LucyChyzhova: that's a good idea!

src/components/text-editor/text-editor.scss
Kiarokh: I know what you mean :) . I think this is more correct. If we ever make other `ui` types, we don't want to mention their names every time here, and keep track of this. This way we're just saying for anything that is not the `minimal`, we consider this styles. It doesn't matter what.. but just _not_ the `minimal` ui

LucyChyzhova: hm, yes, from this point of you, it makes sense. agree 😊

adrianschmidt:
Tip:

If you run `npm run api:update`, and commit the result, the maintainers team should be requested for review automatically 🙂

src/components/text-editor/text-editor.tsx
```diff
@@ -111,6 +112,16 @@ export class TextEditor implements FormComponent<string> {
     @Prop({ reflect: true })
     public allowResize: boolean = true;
 
+    /**
+     * Defines how the editor should look
+     * - `standard`: The default look of the editor
+     * - `minimal`: A more compact look of the editor, useful for
+     * displaying on mobile devices for instance space is limited.
+     * It will remove the toolbar, unless the editor is in focus.
+     */
```
adrianschmidt, line 121:
```suggestion
    /**
     * Specifies the visual appearance of the editor.
     *
     * - `standard`: The default editor appearance with a full toolbar and
     *    standard layout.
     * - `minimal`: A compact editor appearance, ideal for limited space
     *    scenarios such as mobile devices. In this mode, the toolbar is hidden
     *    until the editor is focused.
     */
```

adrianschmidt: I mainly wanted to fix the grammatical issue in "useful for displaying on mobile devices for instance space is limited", but I liked the overall improvements ChatGPT suggested 😊

Kiarokh: nice. added it as a fixup

adrianschmidt (approved):

PR #3107, opened 2024-08-15
fix(file): display a better clue of about the dropzone

LucyChyzhova:
## Review:

https://github.com/user-attachments/assets/393d410c-ff26-46ca-87e1-8b429be51e55

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/file/file.tsx
```diff
@@ -119,17 +119,35 @@ export class File {
...
+    private renderDragAndDropTip() {
+        if (this.value || this.disabled || this.readonly) {
+            return;
+        }
+
+        return (
+            <span
+                class="drag-and-drop-tip"
+                dropzone-tip={this.dropZoneTip()}
+            ></span>
```
Kiarokh, line 143:
```suggestion
            />
```
you know me 😅

Kiarokh, line 143:
```suggestion
           return <span class="drag-and-drop-tip">{this.dropZoneTip()}</span>;
```

Here we don't need to do it the hacky way we used to do in the previous versions, where the solution was inside `limel-chip-set`. I don't even remember why we did it that we back then. I guess it must have been because of the lack of my typescript knowledge, or because we didn't want to mess around too much in MDC components, and not make updating them harder.

But right now, there is no reason to do this using pseudo elements and `content: attr(dropzone-tip);` really. We can just write this as the content of the `<span>`. I mean, we added the span itself. So we can add its content too. 

This of course means updating the CSS respectively.

src/components/file/file.scss
Kiarokh, line 1: Great! `Also make sure to remove the `_file-picker.scss` and the `:host(.is-file-picker.shows-dropzone[readonly]) {
` styles block from `_readonly.scss` in `chip-set`.

src/components/file/file.tsx
LucyChyzhova: That's how I started, - with just text in span 😀 , then checked how we did it before. 🙃
Sure! I will add a fixup.

PR #3109, opened 2024-08-16
fix(file-picker): improve UI for readOnly

LucyChyzhova:
#
Screenshot 2024-08-16 at 10 24 43

# Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/style/internal/shared_input-select-picker.scss
+++ b/src/style/internal/shared_input-select-picker.scss
@@ -270,6 +270,13 @@ $cropped-label-hack--font-size: 0.875rem; //14px
             color: $input-text-color;
         }
 
+        &.mdc-text-field--with-leading-icon,
+        &.has-leading-icon {
+            .lime-looks-like-input-value {
+                padding-left: 1rem;
+            }
+        }
+
         .lime-looks-like-input-value {
             color: $input-text-color;
             opacity: 1;


src/components/chip-set/partial-styles/_file-picker.scss
```diff
@@ -79,3 +79,8 @@ $max-width-of-label-when-not-readonly: 40%;
         }
     }
 }
+
+.mdc-text-field.mdc-text-field--disabled.lime-text-field--readonly
+    .lime-looks-like-input-value {
+    padding-left: 1rem;
```
Kiarokh, line 85: this wouldn't work if there is no leading icon.
in file-picker there is a hardcoded leading icon. But chip-set can be with or without the leading icon.

LucyChyzhova: ahhh! right! thanks! will check the icon

Kiarokh (approved):

PR #3128, opened 2024-08-26
Add popup plugin

john-traas:
Fixes https://github.com/Lundalogik/crm-feature/issues/4301

### DIFF
--- a/src/components/text-editor/prosemirror-adapter/mentions/node-schema-extender.ts
+++ b/src/components/text-editor/prosemirror-adapter/mentions/node-schema-extender.ts
@@ -1,0 +1,36 @@
+import { NodeSpec } from 'prosemirror-model';
+
+export const mention: NodeSpec = {
+    inline: true,
+    group: 'inline',
+    selectable: true,
+    atom: true,
+
+    attrs: {
+        type: {},
+        id: {},
+        name: {},
+    },
+
+    toDOM: (node) => [
+        'span',
+        {
+            class: 'mention',
+            style: 'color: blue',
+            'data-mention-name': node.attrs.name,
+            'data-mention-type': node.attrs.type,
+            'data-mention-id': node.attrs.id,
+        },
+        '@' + node.attrs.name,
+    ],
+    parseDOM: [
+        {
+            tag: 'span[data-mention-type][data-mention-id][data-mention-name]',
+            getAttrs: (dom) => ({
+                name: dom.getAttribute('data-mention-name'),
+                type: dom.getAttribute('data-mention-type'),
+                id: dom.getAttribute('data-mention-id'),
+            }),
+        },
+    ],
+};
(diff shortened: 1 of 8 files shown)


src/components/text-editor/prosemirror-adapter/mentions/node-schema-extender.ts
```diff
@@ -0,0 +1,35 @@
+import { NodeSpec } from 'prosemirror-model';
+
+export const mention: NodeSpec = {
```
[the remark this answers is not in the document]
john-traas: Many aspects of this will be agnostic and generic. However for NodeSpec's they will have to be unique and specific to what they hold. 

If we're traversing the Editor State the current Node or group of Nodes needs to tell us exactly what it or they are. So for tags we would create a separate NodeSpec to specify tags.

src/components/text-editor/prosemirror-adapter/plugins/trigger-popup-plugin.ts
```diff
@@ -0,0 +1,125 @@
...
+
+type Trigger = '@' | '#';
+
+const triggers: Trigger[] = ['@', '#'];
+
+const isTrigger = (key: string): key is Trigger => {
+    return triggers.includes(key as Trigger);
+};
+
+const shouldTrigger = (state: EditorState): boolean => {
```
Kiarokh, line 15: these short command trigger should also work if the previous character is a line-break. So when `@` or `#` are the first letters of a new paragraph.

src/components/text-editor/prosemirror-adapter/plugins/trigger-popup-plugin.ts
```diff
@@ -0,0 +1,125 @@
...
+    const { $from, $to } = state.selection;
+
+    const prevPos = $from.pos - 1;
+
+    if (!prevPos) {
```
Kiarokh, line 20: I think we generally try to avoid abbreviations. So maybe reconsider `pos` or `char` or `prev` and such 😊

src/components/text-editor/prosemirror-adapter/mentions/node-schema-extender.ts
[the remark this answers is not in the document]
Kiarokh: Not sure if I entirely understand the technical aspects that you mentioned. But what I was trying to say is, should this _@-ing_ or _#-ing_ be called "mentioning" and "tagging"? Or can we find more generic names and functionality-agnostic designs for them, such as `at-command`, `hashtag-command` etc… 

If we call it "mention" and create an entire set of user interactions and components for handling mentions (picking a person from a searchable list while typing a piece of text), our work will be easy and straightforward. The consumer's expectations will also be straightforward. They will only use the `@` trigger to implement mentions, where they use the text editor component. 

However, in another context for other consumers, the _@-ing_ may be used to link things together maybe… Or to add _locations_, or _dates_ or whatever the 3rd-party consumers' use-case may be. 

Maybe I'm over complicating… but this was just a question that poped in my head.

src/components/text-editor/prosemirror-adapter/plugins/trigger-popup-plugin.ts
Kiarokh: seems they already do that ;)

Issue #1628, opened 2022-02-18
limel-form: calculation of columns only on resize wrong initial value


TommyLindh2:
## Current behavior

When having a `limel-form` with layout grid the calculation of number of columns seems off and is "corrected" if the size of the form is resized.

It looks like it's [here](https://github.com/Lundalogik/lime-elements/blob/e5cff57c076efe832f60e0af960ae9a1474fd8d0/src/components/form/templates/grid-layout.ts#L33-L51)

The initial width is `35rem` and then it's single column view which makes my two fields on separate rows.
But when changing the width to `34rem` and then back to `35rem` the two fields are on the same row (Which is the wanted behavior)

See how it behaves below:

https://user-images.githubusercontent.com/154725/156997308-d52e60e6-5e5e-4c13-bf96-ad6784da951b.mov

(I made a video of the gif, [original gif here](https://user-images.githubusercontent.com/1814702/154678468-858e5d55-5fea-4e85-b9ea-2051881be17c.gif) / @adrianschmidt)

FormSchema:
```typescript
{
    type: 'object',
    required: ['note'],
    lime: {
        layout: {
            type: 'grid',
            columns: 2,
        },
    },
    properties: {
        person: {
            title: 'Person',
            readOnly: true,
            lime: {
                layout: {
                    colSpan: 1,
                },
            },
        },
        company: {
            title: 'Company',
            readOnly: true,
            lime: {
                layout: {
                    colSpan: 1,
                },
            },
        },
        note: {
            title: 'Note',
            lime: {
                layout: {
                    colSpan: 'all',
                }
            }
        }    
    }
};
```

Render it inside a `limel-dialog` and set the `--dialog-width` to for instance `35rem`

## Expected behavior

It should get the same number of columns initially and after a resize-event.
In this case 2 columns.

## Environment

- **lime-elements** version: 35.0.0-next.9

TommyLindh2:
@BjornOstberg if you want to follow this.

adrianschmidt:
Hmm… that's odd… 🤔

@Kiarokh Do you have any idea what could cause this?

@TommyLindh2 FYI, nowadays, GitHub supports uploading videos, as well as gifs, and videos have the benefit of having playback controls 🙂

BjornOstberg:
@Kiarokh I think I can share and show on monday if that fits your schedule?

Kiarokh:
Absolutely

Kiarokh:
[...]

PR #3129, opened 2024-08-26
fix(form): correctly determine the correct number of columns when form is used inside dialog

adrianschmidt:
fix: Lundalogik/lime-elements#1628

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


adrianschmidt:
@TommyLindh2 Can you try this and check that it resolves the problem?

TommyLindh2:
It seems to work, I've also updated the `Dialog with form and header` example so it uses `limel-form` so that we can verify that it works there as well 👍

adrianschmidt:
> I've also updated the `Dialog with form and header` example so it uses `limel-form` so that we can verify that it works there as well 👍

Oh, good idea! Thank you!

> First you see it "wrong" and then it updates, but I guess that's a bit hard to fix?

Yeah, it's not trivial to get around that unfortunately 😒

PR #3139, opened 2024-09-01
feat(text editor mentions): add plugin factory

john-traas:
Add's a new custom node spec to use for mentions in the Text Editor.

Fixes https://github.com/Lundalogik/crm-feature/issues/4311


src/components/text-editor/utils/markdown-converter.ts
```diff
@@ -1,14 +1,25 @@
...
     MarkdownSerializer,
+    MarkdownSerializerState,
     defaultMarkdownSerializer,
 } from 'prosemirror-markdown';
 import { markdownToHTML } from '../../markdown/markdown-parser';
 
+const mentionSerializer = {
+    mention: (state: MarkdownSerializerState, node: ProseMirrorNode) => {
+        state.write(
+            `<limebb-mention type="${node.attrs.type}" objectid="${node.attrs.objectid}" descriptive="${node.attrs.descriptive}"></limebb-mention>`,
```
jgroth, line 14: I don't think we should hardcode this element here like this. For this to be reusable, we would either have to provide some config or a factory to the the text editor component that will create the element that the consumer needs

john-traas: Absolutely

src/components/text-editor/types.ts
```diff
@@ -0,0 +1,16 @@
+import { NodeSpec } from 'prosemirror-model';
+
+/**
+ * @private
+ * set to private to avoid usage while under development
+ */
+export type TextEditorPlugin = {
+    node: NodeConfig[];
```
jgroth, line 8: We can't expose types from a third party like this. If there's a breaking change in these types those would be immediately exposed from lime elements without us necessarily being aware of it. We need to create our own type and then map that type to the `NodeSpec` type

src/components/text-editor/text-editor.tsx
```diff
@@ -93,6 +95,9 @@ export class TextEditor implements FormComponent<string> {
     @Prop({ reflect: true })
     public value: string;
 
+    @Prop()
```
jgroth, line 98: Perhaps this should be marked as private as well? :slightly_smiling_face:

src/components/text-editor/types.ts
civing: I think you might have conflated NodeConfig with NodeSpec? :)

src/components/text-editor/types.ts
```diff
@@ -0,0 +1,16 @@
...
+};
+
+export interface NodeConfig {
+    tagName: string;
+    attrs: string[];
+}
+
+export type NodeSpecFactory = (config: NodeConfig) => NodeSpec;
```
civing, line 16: I'm not sure we need to export this to consumers

src/components/text-editor/types.ts
```diff
@@ -0,0 +1,16 @@
...
+ * set to private to avoid usage while under development
+ */
+export type TextEditorPlugin = {
+    node: NodeConfig[];
+};
+
+export interface NodeConfig {
+    tagName: string;
+    attrs: string[];
+}
```
civing, line 14: Can we use the newly added `CustomComponent` used for the whitelist in the markdown reader? That was my intention at least when I created that one.

civing (commented):
I'm yet to review all of it. But here are some initial comments and thoughts

src/components/text-editor/types.ts
jgroth: Yes :slightly_smiling_face:

src/components/text-editor/text-editor.tsx
john-traas: Yes :)

src/components/text-editor/types.ts
john-traas: Good point. We can house it directly within the nodespec factory

src/components/text-editor/types.ts
john-traas: Sounds good

civing (dismissed):
Great!

civing (approved):

etc/lime-elements.api.md
```diff
@@ -564,6 +564,8 @@ export namespace Components {
     export interface LimelProsemirrorAdapter {
         "contentType": 'markdown' | 'html';
         "language": Languages;
+        // Warning: (ae-incompatible-release-tags) The symbol "plugins" is marked as @beta, but its signature references "CustomElement" which is marked as @alpha
+        "plugins": CustomElement[];
```
adrianschmidt, line 568: This warning means that we should mark the `plugins` prop as `@alpha`. I doesn't have an explicit `@beta` tag right now, but it, presumably, inherits the `@beta` tag from the class.

I'll see if adding `@alpha` to the prop removes this warning.

adrianschmidt: #3263

Issue #2042, opened 2023-01-16
Dock: shadow effect of hovered buttons gets cut off


Kiarokh:
## Current behavior

<img width="138" alt="image" src="https://user-images.githubusercontent.com/35954987/212692169-77793940-d4f8-46ce-8904-1541fe2b0134.png">

## Expected behavior
- either on hover, we need a `z-index`
- or a transparent background color for buttons by default, for non hovered buttons ⬅️  my choice

paulinewahle:
Changed the padding-bottom

PR #3141, opened 2024-09-02
fix(dock): prevent button shadow from cut-off

paulinewahle:
fix: https://github.com/Lundalogik/lime-elements/issues/2042
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/dock/dock.scss
```diff
@@ -70,7 +70,7 @@ nav {
     flex-grow: 1;
```
Kiarokh, line 71: First of all, congrats and great job on your first PR! It's so exciting every time, even for me! 🤩

src/components/dock/dock.scss
```diff
@@ -70,7 +70,7 @@ nav {
...
 
     padding: var(--dock-padding); // needed for the focus effect
-
+    padding-bottom: calc((var(--dock-padding) * 2.4));
```
Kiarokh, line 73: This approach is logical. But it has a few problems.
1. It adds some extra "unnecessary" distance between the buttons. 
    image 
    It's not a a big deal. But in general, we try to **save** as much space as possible. So if we can achieve the same result without adding the extra space, it'd be absolutely desirable.
2. In the mobile layout (scroll further down in the documentations examples), there is also the same padding-bottom. But it is not needed there.
    image So if we want to go with this approach, we need to make sure that the mobile layout is fine.

src/components/dock/dock.scss
```diff
@@ -70,7 +70,7 @@ nav {
...
     overflow-y: auto;
```
Kiarokh, line 74: The shadow cut-off appears due to the `overflow` on the scrollable area of the Dock. The overflow is needed, because there could be many items in this list. However, a better fix might be doing something about that. Now I don't want to give you the solution, because this is still about practicing and understanding how everything is connected, and what concerns exist in each component. So, I'm curious to see what ideas you can come up with.

paulinewahle: For a little clarification: the dock should remain 100vh and overflow for more possible items but not cut anything off when it's just the two default in the footer? Or will the footer never grow higher?

civing:
Look at you! First PR

src/components/dock/dock.scss
Kiarokh: I'm not sure if I understood your question. Could you please describe more?

Issue #3140, opened 2024-09-02
Spinner: Replace the old Lime brand colors with the new ones


Kiarokh:
The spinner shuffles through Lime's brand colors. 

https://github.com/user-attachments/assets/f1c1e2c4-7752-4e7f-aec8-38fae84979a8

But they have now changed and we need to update them. 

There are new css variables for these new colors, now available in the code.

PR #3145, opened 2024-09-03
Spinner brandcolors

paulinewahle:
fix: https://github.com/Lundalogik/lime-elements/issues/3140
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/spinner/examples/spinner-color.scss
+++ b/src/components/spinner/examples/spinner-color.scss
@@ -1,13 +1,13 @@
 :host {
-    --spinner-color-1: rgb(var(--color-cyan-darker));
-    --spinner-color-2: rgb(var(--color-cyan-dark));
-    --spinner-color-3: rgb(var(--color-cyan-default));
-    --spinner-color-4: rgb(var(--color-cyan-light));
-    --spinner-color-4: rgb(var(--color-cyan-lighter));
-    --spinner-color-5: rgb(var(--color-gray-lighter));
-    --spinner-color-6: rgb(var(--color-cyan-lighter));
-    --spinner-color-7: rgb(var(--color-cyan-light));
-    --spinner-color-8: rgb(var(--color-cyan-default));
-    --spinner-color-9: rgb(var(--color-cyan-dark));
-    --spinner-color-10: rgb(var(--color-cyan-darker));
+    --spinner-color-1: rgb(var(--lime-brand-color-lime-green));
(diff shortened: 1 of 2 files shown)


src/components/dock/dock.scss
```diff
@@ -70,7 +70,7 @@ nav {
     flex-grow: 1;
 
     padding: var(--dock-padding); // needed for the focus effect
-
+    padding-bottom: calc((var(--dock-padding) * 2.4));
```
LucyChyzhova, line 73: I think this is unrelated change. Could you please remove this commit.

paulinewahle: ⚡

LucyChyzhova:
Nice! We should also update the `Custom colors` example, since we are saying that we use Lime Technologies' brand colors there.

LucyChyzhova (dismissed):
Looks really nice! 😍

Kiarokh:
really great job! 
I just noticed that there is a stroke that is constantly orange, and another stroke that keeps changing 😅 

https://github.com/user-attachments/assets/c5a12b00-7761-423c-b628-c560babeec3a

paulinewahle:
Good catch @Kiarokh! I fixed the unbranded spinner colors. The grey brand color looks too dark in the branded spinner, I used green now but could use a non brand light grey instead if you'd like it better;)

Kiarokh:
tnx 😊 i think a neutral color should be fine. it doesn't have to be from the brand colors. you can pick from the `--contrast-xxx` series ;)

adrianschmidt:
> tnx 😊 i think a neutral color should be fine. it doesn't have to be from the brand colors. you can pick from the `--contrast-xxx` series ;)

+1

I was looking at the new spinner and comparing it to the old one before reading the comments here, and with green as the "neutral" colour, the spinner looks a lot "busier" than before.

(I see that you've already updated it 😊)

adrianschmidt:
Looks good to me 🙂

https://github.com/user-attachments/assets/f932c601-b365-4288-a39d-c65033f40337

LucyChyzhova (approved):
Looks good! Great job!

Issue #3149, opened 2024-09-05
limel-markdown: component crashes if supplied value contains malformed html


adrianschmidt:
Reported in [Open Hours in Teams](https://teams.microsoft.com/l/message/19:8eb8d446f40d4696ba03bee0d3b47a56@thread.tacv2/1709122481382?tenantId=15f5bb21-07a4-42c6-ada2-08965bd62081&groupId=2f699c45-fe12-4487-81ae-829eb08a675a&parentMessageId=1709122481382&teamName=Lime%20CRM&channelName=Dev%20Open%20Hours%20(Tues%20and%20Thurs%2010.00%20-%2011.00)&createdTime=1709122481382), and [in the forum](https://forum.lime-crm.com/t/communcation-comes-as-code-and-wont-send-replys-suddenly/6100).

## Current behavior

If the supplied string contains an html tag with a blank line (2 consecutive line breaks) between the opening (`<`) and closing (`>`) brackets, the component crashes while parsing, and no output is displayed at all.

1. Supply the following input to a limel-markdown instance, for example via the composite example in the docs:
	```
	<p>this paragraph is correct</p>
	<p
	
	>this paragraph is broken</p>
	<p>here is another correct paragraph</p>
	```

- Nothing is rendered.

## Expected behavior

While this isn't really wellformed html, browsers and email clients _do_ handle this, and therefore we kind of _have_ to in order for the component to be useful in Lime CRM. The output should ideally be rendered as if there was no line breaks inside the tag, but it should at least not crash completely and not output _anything_.

## Environment

- **lime-elements** version: confirmed in [v37.57.5](https://github.com/Lundalogik/lime-elements/releases/tag/v37.57.5), but likely present in all previous versions that has the `limel-markdown` component.
- Framework used: Can be reproduced in the docs.
- Logs:

TommyLindh2:
@adrianschmidt I think this is related, we have still reports about "faulty" rendering in communication flow since we started using `limel-markdown`.


Here's an example that will render incorrectly:

It will render the `</style></span>` as text, do you think it's possible to fix?

```
<div><span><style>

</style></span>

Hello2

</div>
```

adrianschmidt:
Great! As long as there are no downsides where we eat unreasonable amounts of performance, or have to make trade-offs where we need to break some other cases to make something work, it's lovely if we can make the component more forgiving 🙂

[...]

PR #3150, opened 2024-09-05
fix(markdown): do not crash when html tag contains blank line

adrianschmidt:
-CHANGELOG-
When the markdown component is fed html in which there is a blank line (2 consecutive line breaks) anywhere between the opening bracket and the closing bracket, the components crashes. Update the sub-dependency `hast-util-raw` to avoid this. The parser now switches to text-mode when this happens, which will output the rest of the html after the blank line, up until the closing bracket, as text. So it still will not render the html element correctly, but at least the parser does not crash, and the rest of the content after the element's closing bracket will be rendered correctly.

fix: #3149

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5652,10 +5652,11 @@
       }
     },
     "node_modules/hast-util-raw": {
-      "version": "9.0.1",
-      "resolved": "https://registry.npmjs.org/hast-util-raw/-/hast-util-raw-9.0.1.tgz",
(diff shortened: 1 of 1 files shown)


TommyLindh2 (approved):
I can verify that it solves the issue

PR #3157, opened 2024-09-10
feat(markdown): add component whitelist

civing:
Fixes: Lundalogik/crm-feature#4356
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/markdown/markdown-parser.ts
+++ b/src/components/markdown/markdown-parser.ts
@@ -9,6 +9,8 @@ import rehypeRaw from 'rehype-raw';
 import { visit } from 'unist-util-visit';
 import { sanitizeStyle } from './sanitize-style';
 import { Node } from 'unist';
+import { Schema } from 'rehype-sanitize/lib';
+import { CustomElement } from '../../global/shared-types/custom-element.types';
 
 /**
  * Takes a string as input and returns a new string
@@ -39,12 +41,7 @@ export async function markdownToHTML(
         .use(rehypeExternalLinks, { target: '_blank' })
         .use(rehypeRaw)
         .use(rehypeSanitize, {
-            ...defaultSchema,
-            attributes: {
-                ...defaultSchema.attributes,
-                p: [['className', 'MsoNormal']], // Allow the class 'MsoNormal' on <p> elements
-                '*': ['style', 'width'], // Allow `style` and 'width' attribute on all elements
-            },
+            ...getWhiteList(options?.whitelist ?? []),
         })
         .use(() => {
             return (tree: Node) => {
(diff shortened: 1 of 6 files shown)


src/components/markdown/markdown.tsx
```diff
@@ -32,11 +33,15 @@ export class Markdown {
     @Prop()
     public value: string;
 
+    @Prop()
+    public whitelist?: CustomComponent[];
+
```
Kiarokh, line 38: Don't forget to add an example, showcasing this new prop in the docs

src/global/shared-types/component.types.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Allowed component
+ */
+export interface CustomComponent {
```
jgroth, line 4: I think we need another name for this. Even though no type named `CustomCompoent` is being exported from this package today, the term is used in several other places and has another meaning so it might be confusing. Also, you can in theory allow regular HTML elements here as well, so maybe the term "element" is better than "component". `AllowedElement` perhaps?

civing: I was hoping to reuse this interface in the text editor part when we add custom nodes. Can we call it `CustomElement` ?

jgroth: I we make it private for now you can call it whatever you want until we know more about what it will be used for :slightly_smiling_face:

src/components/markdown/markdown.tsx
civing: done!

src/global/shared-types/component.types.ts
civing: see https://github.com/Lundalogik/lime-elements/pull/3157/commits/f72651ea26500dae156fad30e1a4cc6f7a877214

src/components/markdown/markdown.tsx
```diff
@@ -32,11 +34,15 @@ export class Markdown {
     @Prop()
     public value: string;
 
+    @Prop()
```
jgroth, line 37: Since the type is private we need to set this as private as well

src/components/markdown/markdown-parser.ts
```diff
@@ -67,4 +85,5 @@ export interface markdownToHTMLOptions {
      * Set to `true` to convert all soft line breaks to hard line breaks.
      */
     forceHardLineBreaks?: boolean;
+    whitelist?: CustomElement[];
```
jgroth, line 88: Is this interface exposed? If so, this needs to be private as well

src/components/markdown/markdown.tsx
civing: I made it private (or rather @alpha)

jgroth (dismissed):

jgroth (approved):

PR #3160, opened 2024-09-12
refactor(linear-progress): get rid of MDC dependecy

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/linear-progress/examples/linear-progress-accessible-label.tsx
+++ b/src/components/linear-progress/examples/linear-progress-accessible-label.tsx
@@ -1,0 +1,25 @@
+import { Component, h } from '@stencil/core';
+
+/**
+ * Using an accessible label
+ * The component tells the assistive technologies that it is a "Progress bar".
+ * However, instead of relying on such a generic label,
+ * you can provide a more specific label that describes what the progress bar is for,
+ * or what kind of progress it is visualizing.
+ *
+ * This is achieved, using the `accessibleLabel` property.
+ */
+@Component({
+    shadow: true,
+    tag: 'limel-example-linear-progress-accessible-label',
+})
+export class LinearProgressAccessibleLabelExample {
+    public render() {
+        return (
+            <limel-linear-progress
+                value={0.87}
+                accessibleLabel="Percentage of today's progress"
+            />
+        );
+    }
+}
--- a/src/components/linear-progress/linear-progress.tsx
+++ b/src/components/linear-progress/linear-progress.tsx
@@ -10,6 +10,7 @@ const PERCENT = 100;
  *
  * @exampleComponent limel-example-linear-progress
  * @exampleComponent limel-example-linear-progress-indeterminate
+ * @exampleComponent limel-example-linear-progress-accessible-label
  * @exampleComponent limel-example-linear-progress-color
  */
 @Component({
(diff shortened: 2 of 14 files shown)


src/components/linear-progress/examples/linear-progress-accessible-label.tsx
```diff
@@ -0,0 +1,25 @@
+import { Component, h } from '@stencil/core';
+
+/**
+ * Using an accessible label
+ * The component tells the assistive technologies that it is a "Progress bar".
+ * However, instead relying on such a generic label,
```
adrianschmidt, line 6:
```suggestion
 * However, instead of relying on such a generic label,
```

src/components/linear-progress/linear-progress.tsx
```diff
@@ -24,76 +38,56 @@ export class LinearProgress {
...
-                    <div class="mdc-linear-progress__buffer-bar"></div>
-                </div>
-                <div class="mdc-linear-progress__bar mdc-linear-progress__primary-bar">
-                    <span class="mdc-linear-progress__bar-inner" />
-                </div>
-                <div class="mdc-linear-progress__bar mdc-linear-progress__secondary-bar">
-                    <span class="mdc-linear-progress__bar-inner" />
-                </div>
-            </div>
+                <div class="progress" />
```
jgroth, line 73: Perhaps we should use the `meter` element here :thinking: 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meter

jgroth: Or `progress` :smile: 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress

Kiarokh: hmm… the `meter` element looks really interesting and feature-rich. But I don't actually know which one is better to use to be honest. 

Also, they seem to be hard to style, and styling support is inconsistent across browsers. 

If we decide to go that route, I'd say we can do a new refactor PR again later.

jgroth: Yeah I have no idea if it's possible to style them :smile: Just thought it could be good to have it for semantic reasons and to maybe assist screen readers etc

src/components/linear-progress/linear-progress.tsx
```diff
@@ -15,6 +18,13 @@ const PERCENT = 100;
     styleUrl: 'linear-progress.scss',
 })
 export class LinearProgress {
+    /**
+     * Defines the language for translations.
+     * Will translate the translatable strings on the components.
+     */
+    @Prop({ reflect: true })
+    public language: Languages = 'en';
```
adrianschmidt, line 27: It feels like we ought to default to the language requested by the browser here, if it's one we support, and otherwise use English. But I guess this PR is big enough without that addition.

adrianschmidt (dismissed):

src/components/linear-progress/linear-progress.tsx
Kiarokh: I totally agree. But I have added so many accessibility things here just for that ;) So we're safer than before

src/components/linear-progress/linear-progress.tsx
adrianschmidt: Yeah, true 👍

adrianschmidt (approved):

Issue #3158, opened 2024-09-11
`limel-form` with schema required fields breaks validations from being displayed correctly


simonedeblasio:
## Current behavior
If a field is required the part that displays errors in a form breaks. 

![image](https://github.com/user-attachments/assets/516527c9-031b-47eb-bc20-eecf565268bf)
Both fields here need to be larger than today, but since field 1 is set to required the error is not displayed.

If I click around like few times it is then finally displayed:
![image](https://github.com/user-attachments/assets/133ac23e-95a1-4096-bd95-59d737898c45)

If I just remove the "required" part in the schema `required: ['startdate']` it works and shows the error immediately. 

To clarify, for the field 2, it always shows the error directly and as expected. The only difference is the required part.

1. Create a dialog with form + schema. 
2. Add two fields in the schema, set one to required in the schema.
3. Add logic in the "onChange" for the form that updates the errors. For example date must be larger than today.

Like this:
``` typescript
<limel-form
    onChange={this.handleFormChange}
    value={this.formData}
    schema={getSchema(this.limetypes, this.limeobject)}
    errors={this.errors}
/>,
```

This bug might have a releated issue which is that if you pass your own errors to a form, it will still thinking something is empty even though you have filled it in:
![image](https://github.com/user-attachments/assets/7d15d821-84c4-4f3a-9c80-086a365b0724)
Even when switching between fields, still keeps saying that its required. I need type more/again in it for it to refresh the validation.

## Expected behavior

The error do be displayed right after the value was input

## Environment

- **lime-elements** version: 36.2.0

simonedeblasio:
Ping maybe @adrianschmidt or @vicgeralds

vicgeralds:
> schema={getSchema(this.limetypes, this.limeobject)}

What's happening in `getSchema`? If you create a new schema on every render, I expect that will cause problems.

Other than that, I'm not sure what might cause this. Will need to debug a bit when I get the time.

simonedeblasio:
The only logic in it is populating default values and setting readonly depning on the default values.
``` typescript
export const getSchema = (
    limetypes: Record<string, LimeType>,
    limeobject: LimeObject
) => {
    const today = new Date();
    const swedishStartdate = new Date(limeobject.startdate);
    const swedishStartdateString = getSwedishDate(limeobject.startdate);
    const swedishGraduationdateString = getSwedishDate(
        limeobject.graduationdate
    );

    return {
        type: 'object',
        properties: {
[...]

PR #3165, opened 2024-09-16
fix(form): consider any Date value as non-empty

vicgeralds:
This makes sure we don't hide validation errors for form fields that are set to a `Date` value or use a date picker. 

When we determine if a field should be shown as invalid, we check if a field is dirty or has a value, and that check didn't consider `Date` objects, detecting them as empty.

The date picker widget passes a date value rather than a string, so it wasn't handled properly. Also, while `Date` values are not considered valid form data (rather strings of format "date-time" or "date" are expected), setting a `Date` value in form data is an easy programming mistake to make, especially since limel-date-picker expects and emits `Date` values. Not showing a field with such a value as invalid gives the programmer the impression the value is ok when it's not.

fix: #3158

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


vicgeralds:
Ok, so I finally was able to reproduce and understand the problem. The fix that I did before solves it: https://github.com/Lundalogik/lime-elements/pull/3165

So it was a problem with detecting a `Date` value as empty, but the date values are converted from strings internally by the date picker widget. In other words, it was clearly a bug lime elements that caused this. 😊

adrianschmidt (approved):
Looks good to me. Thank you for the fix! 🙏

PR #3167, opened 2024-09-16
feat(chip): remove `@beta` tag

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,8 @@ For a full list of components, along with live examples, please visit the [docum
 
 [![Version](https://img.shields.io/npm/v/@limetech/lime-elements.svg)](https://www.npmjs.com/package/@limetech/lime-elements) [![semantic-release](https://img.shields.io/badge/%20%20%F0%9F%93%A6%F0%9F%9A%80-semantic--release-e10079.svg)](https://github.com/semantic-release/semantic-release)
 
+---
+
 ## Getting Started
 
 -   To install, run `npm install @limetech/lime-elements`.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -74,10 +76,28 @@ Providing your own icons is crucial as many of our components use an `Icon` inte
...
+
+Some components in Lime Elements are marked as `@beta`. These components are experimental and may still be in development. They could contain bugs and their APIs and interfaces might change without notice. This means that, while you can use these components, you should be aware that they are subject to changes and may not yet be fully stable.
+
+Key Points:
+
+1. **Experimental Status:** Components marked as `@beta` are not final and may evolve based on feedback and further development.
+1. **Potential Issues:** Be prepared for potential bugs or incomplete features.
+1. **Interface Changes:** We may alter their interfaces without formally documenting these changes in the change log. This allows us to refine the components based on real-world usage and feedback.
+
+Use `@beta` components with caution and at your own risk. They are available for early testing and feedback, helping us improve the library for future releases.
```
adrianschmidt, line 91: I don't think this explanation of the `@beta` tag is good enough. What needs to be clear is that we may make _breaking changes_ to a component marked as `@beta`, without bumping the major version number.

No components are final, all components may evolve based on feedback and further development.

Kiarokh, line 91:
```suggestion
Components marked as `@beta` are in a transitional state and may undergo significant changes. Here’s what you need to know:

1. **Breaking changes:** We may introduce breaking changes to `@beta` components without incrementing the major version number, or announcing a "Breaking change" in the change log. This means you should be prepared for potential interface changes that could affect your implementation of such components.
1. **Evolving nature:** While all components in Lime Elements are subject to evolution based on feedback and development, `@beta` components are specifically noted for their fluidity. Unlike fully released components, which are expected to maintain stability, `@beta` components can be modified more freely as we refine them.
1. **Use at your own risk:**  By using `@beta` components, you acknowledge that they are not fully stable and may still contain bugs. They are offered for early testing and feedback to help us improve the overall library.

In summary, while no component is final, `@beta` components are more prone to changes and should be used with caution. Your feedback is invaluable in shaping the future of these components.
```

Kiarokh: what about this one 👇

adrianschmidt, line 91:
```suggestion
## Using @beta components

Components marked as `@beta` are not yet stable. Here's what you need to know:

1. **Breaking changes:** We may change `@beta` components in ways that break your code. We won't increase the major version number when we do this.

2. **Unstable:** `@beta` components may have more bugs than other components. Use them at your own risk.

3. **Feedback wanted:** We offer `@beta` components to get your input. Your feedback helps us improve them.

Remember: All components can change, but `@beta` components change more often and without warning.
```

adrianschmidt (approved):

PR #3173, opened 2024-09-17
fix(table): only release element pool if should replace is true

john-traas:
The element pool should only be released if all the data should be replaced. 

Fixes https://github.com/Lundalogik/crm-feature/issues/4377


eketorp:
Fixes https://github.com/Lundalogik/crm-feature/issues/4344?

john-traas:
> Fixes [Lundalogik/crm-feature#4344](https://github.com/Lundalogik/crm-feature/issues/4344)?

Ah I didn't test that, but I think it will.

civing (approved):

PR #3182, opened 2024-09-18
perf(portal): do DOM manipulation when portal becomes visible

jgroth:
If many portals are rendered on a page there might be a heavy performance hit when all the elements are moved around in the DOM at the same time. We only need to actually move the elements once the portal becomes visible.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/portal/portal.scss
+++ b/src/components/portal/portal.scss
@@ -14,3 +14,7 @@
 :host([hidden]) {
     display: none;
 }
+
+slot {
+    display: none;
+}
--- a/src/components/portal/portal.tsx
+++ b/src/components/portal/portal.tsx
@@ -140,6 +140,17 @@ export class Portal {
             return;
         }
 
+        if (this.visible) {
+            this.init();
+        }
+    }
+
+    public componentDidLoad() {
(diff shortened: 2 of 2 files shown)


Kiarokh:
can we wait a little bit, before merging this one? I'm working on another PR that touches the same areas and logics of the code, but in an improved way. It'd be good to see if they're not conflicting too much with your changes.

jgroth:
> can we wait a little bit, before merging this one? I'm working on another PR that touches the same areas and logics of the code, but in an improved way. It'd be good to see if they're not conflicting too much with your changes.
> 
> I'll link it as soon as I have it ready

This is to solve an SOS ticket so we should probably not wait too long

Kiarokh (approved):
All instances that use `<limel-portal` seem to work fine after this change.

PR #3189, opened 2024-09-20
Fix jumping cursor in text-editor

adrianschmidt:
See commit messages for details.

Credit to @vicgeralds, because I basically lifted the solution straight from 01ce9be4e 🙏

fix: https://github.com/Lundalogik/crm-feature/issues/4312

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
+++ b/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
@@ -30,7 +30,7 @@ import {
 import translate from 'src/global/translations';
 import { createRandomString } from 'src/util/random-string';
 import { isItem } from 'src/components/action-bar/isItem';
-import { cloneDeep } from 'lodash-es';
+import { cloneDeep, debounce } from 'lodash-es';
 import { Languages } from '../../date-picker/date.types';
 import { strikethrough } from './menu/menu-schema-extender';
 import {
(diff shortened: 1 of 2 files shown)


src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
```diff
@@ -161,6 +177,7 @@ export class ProsemirrorAdapter {
             'open-editor-link-menu',
             this.handleOpenLinkMenu,
         );
+        this.view.dom.removeEventListener('blur', this.handleBlur);
```
vicgeralds, line 180: For good measure, we should check that `this.view` isn't undefined here, since it's technically possible (though unlikely) that the component is disconnected before we call `this.initializeTextEditor()` inside `componentDidLoad()`

vicgeralds (commented):
Haven't tested, but looks good to me!

vicgeralds (dismissed):

src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
adrianschmidt: ⚡ bc22455434a572b8930b5a8a7fa7e0bd58222daf

adrianschmidt: I'll squash it immediately.

adrianschmidt: Thanks for catching that by the way! 😊

adrianschmidt:
> Haven't tested, but looks good to me!

I've tested it fairly well, and I think our dear colleagues in Product will let us now on the double if there's anything wrong 😊

vicgeralds (approved):

Issue #3172, opened 2024-09-17
button: when `loading` add `aria-busy="true"`


Kiarokh:
The Button component has a loading state. 

So when it is loading, we need to add `aria-busy="true"`.

Read more https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-busy

PR #3193, opened 2024-09-23
Aria busy loading button

paulinewahle:
fix: https://github.com/Lundalogik/lime-elements/issues/3172

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -94,6 +94,8 @@ export class Button {
                         outlined: this.outlined,
                     }}
                     disabled={this.disabled || this.loading}
+                    aria-busy={this.loading ? 'true' : 'false'}
+                    aria-live="polite"
                 >
                     {this.renderIcon()}
                     {this.renderLabel()}


src/components/button/button.tsx
```diff
@@ -94,6 +94,8 @@ export class Button {
                         outlined: this.outlined,
                     }}
                     disabled={this.disabled || this.loading}
+                    aria-busy={this.loading ? 'true' : 'false'}
```
Befkadu1, line 97: If I am allowed to suggest, here is my suggestion. I think we can simplify it by removing the ternary expression. What do you think?

```suggestion
                    aria-busy={String(this.loading)}
```

Befkadu1:
I can't see your 1st commit in the "Files Changed" tab, is it removed?

src/components/button/button.tsx
adrianschmidt: I think using `String(this.loading)` makes this code harder to read than using the ternary. But that's a matter of personal preference of course.

But, do we even have to translate the boolean to a string?

adrianschmidt:
> I can't see your 1st commit in the "Files Changed" tab, is it removed? Maybe you can remove the 1st commit if you don't need it

That's looking extremely weird, yes. But the info-tile fix has already been merged to main, that's why it looks weird. I'll update the branch with a rebase instead, and it will resolve itself.

src/components/button/button.tsx
Befkadu1: > But, do we even have to translate the boolean to a string?

I don't know if we have to translate it to string, maybe not. `this.loading` might be enough then

Kiarokh: I've already approved this 
```tsx
aria-busy={this.loading ? 'true' : 'false'}
```
in an earlier PR on info-tile: https://github.com/Lundalogik/lime-elements/pull/3185/files


I think this is better for screen readers, to explicitly set aria-busy="false" when the loading state is complete. This ensures that the element consistently communicates its status to assistive technologies, avoiding any confusion.

if we go for "code simplicity" and write:
```tsx
aria-busy={this.loading}
```
the `aria-busy` attribute will disappear when loading is false, which could lead to screen readers not detecting the end of the loading state as clearly as when it's explicitly set to false. This would make the experience more predictable and accessible.

Kiarokh (approved):

src/components/button/button.tsx
adrianschmidt: Yeah, no, I didn't mean we should prioritise code simplicity over the quality of the _output_. I only meant if the result was the same 😅

Issue #3170, opened 2024-09-17
chip: when `loading` add `aria-busy="true"`


Kiarokh:
The Chip component has a loading state. 

So when it is loading, we need to add `aria-busy="true"`.

Read more https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-busy

adrianschmidt:
@paulinewahle When there's an issue that a PR fixes, we usually put the issue on the board (and move it to review when that time comes), rather then just the PR. Just an FYI 🙂

(I've removed the PR from the board and added the issue instead.)

PR #3194, opened 2024-09-23
fix(button): increase accessibility for loading button

paulinewahle:
fix: https://github.com/Lundalogik/lime-elements/issues/3170
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -94,6 +94,8 @@ export class Button {
                         outlined: this.outlined,
                     }}
                     disabled={this.disabled || this.loading}
+                    aria-busy={this.loading ? 'true' : 'false'}
+                    aria-live="polite"
                 >
                     {this.renderIcon()}
                     {this.renderLabel()}


src/components/button/button.tsx
```diff
@@ -94,6 +94,8 @@ export class Button {
                         outlined: this.outlined,
                     }}
                     disabled={this.disabled || this.loading}
+                    aria-busy={this.loading ? 'true' : 'false'}
+                    aria-live="polite"
```
adrianschmidt, line 98: It seems like this might be the wrong commit? This is only changing `limel-button`, not `limel-chip`, and this same  change seems to also be present in #3193?

Kiarokh: I think you accidentally made a duplicated branch from your other branch where you were working on `limel-button`.

Please make a new branch for chip 😄 and re-link it to the issue.

PR #3196, opened 2024-09-23
feat(chip-set): add `autocomplete` as a prop for `input` type

Kiarokh:
In the Chip set and Picker, the autocomplete creates issues. In `limel-picker` (which is a consumer of `limel-chip-set` this behavior is more annoying.
But since the `autocomplete` in the Chip set is set to `'off'` by default, the annoying problem should automatically disappear from Picker.

fix https://github.com/Lundalogik/crm-feature/issues/4386

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -178,6 +178,14 @@ export class ChipSet {
     public delimiter: string = null;
 
     /**
+     * For chip-set of type `input`, defines whether the input field should have autocomplete enabled.
+     * Read more about the `autocomplete` attribute
+     * [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete).
+     */
+    @Prop({ reflect: true })
+    public autocomplete: string = 'off';
+
+    /**
      * Defines the language for translations.
      * Will translate the translatable strings on the components. For example, the clear all chips label.
      */
@@ -424,6 +432,7 @@ export class ChipSet {
                     onChange={this.inputFieldOnChange}
                     placeholder={this.isFull() ? '' : this.searchLabel}
                     readonly={this.isFull()}
+                    autocomplete={this.autocomplete}
                 />
                 <div
                     class={{
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -221,6 +221,7 @@ export namespace Components {
         "type"?: ChipType;
     }
     export interface LimelChipSet {
+        "autocomplete": string;
         "clearAllButton": boolean;
         "delimiter": string;
         "disabled": boolean;
@@ -1153,6 +1154,7 @@ namespace JSX_2 {
         "type"?: ChipType;
     }
     interface LimelChipSet {
+        "autocomplete"?: string;
         "clearAllButton"?: boolean;
         "delimiter"?: string;
         "disabled"?: boolean;


Befkadu1:
Maybe you forgot to run `npm run api:verify` when API changes?

Kiarokh:
@Befkadu1 I updated the API. There is a bunch of discussion on the issue itself. Please make sure to read it before approving this one.

src/components/picker/picker.tsx
```diff
@@ -275,6 +275,7 @@ export class Picker {
                 onStopEdit={this.handleStopEditAndBlur}
                 emptyInputOnBlur={false}
                 clearAllButton={this.multiple && !this.chipSetEditMode}
+                autocomplete="nope"
```
jgroth, line 278:
```suggestion
                autocomplete="off nope"
```
We should probably still use the `off` keyword

adrianschmidt: You might want to read up on the issue 😂

adrianschmidt: Or, well… no, you probably don't want to do that to be honest…

At least keep something hug-friendly nearby if you do. It's not exactly life-affirming reading… 😣

adrianschmidt: Oh! Hang on, your suggestion is to _also_ use the `off` keyword. Oh, yeah, that makes sense! Sorry, just ignore me, apparently I'm having one of those days again 🙈 😅

adrianschmidt: Why `nope` though @Kiarokh? Didn't using a random string work?

jgroth: I did read up on the issue before making the comment :smile: I also read some of those links on the bug reports. I didn't read everything in detail though so maybe this suggestion won't work, but it seems like the `autocomplete` attribute accepted a space separated list of keywords :smile:

jgroth: Also, using `nope` here, will there be any risk that it start suggesting stuff from all the places where `nope` has been used? :smile:

adrianschmidt: > Also, using `nope` here, will there be any risk that it start suggesting stuff from all the places where `nope` has been used? 😄

Yeah, that's exactly my worry over using a static value rather than a random string. That's how I imagine I would have built the feature…

Kiarokh: > Why `nope` though @Kiarokh? Didn't using a random string work?

No I was just working on other stuff. I'll replace it now

Kiarokh: ⚡

src/components/chip-set/chip-set.tsx
```diff
@@ -248,6 +258,15 @@ export class ChipSet {
         this.handleDeleteAllIconClick =
             this.handleDeleteAllIconClick.bind(this);
         this.renderDelimiter = this.renderDelimiter.bind(this);
+        if (!this.autocomplete) {
+            this.autocomplete = this.getAutocompleteValue();
```
jgroth, line 262: Props are immutable so we can not assign a new value like this. We can however assign it when its declared

src/components/chip-set/chip-set.tsx
```diff
@@ -248,6 +258,15 @@ export class ChipSet {
...
+        }
+    }
+
+    componentWillLoad() {
+        if (!this.autocomplete) {
+            this.autocomplete = this.getAutocompleteValue();
```
jgroth, line 268: Same thing here, props are immutable

src/components/chip-set/chip-set.tsx
```diff
@@ -177,6 +177,16 @@ export class ChipSet {
...
+     * For chip-set of type `input`, defines whether the input field should have autocomplete enabled.
+     * Read more about the `autocomplete` attribute
+     * [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete).
+     *
+     * If no value is set, a random value is generated to avoid browser suggestions.
+     */
+    @Prop({ reflect: true })
+    public autocomplete: string;
+
     /**
```
Kiarokh, line 190: In the Chip set and Picker (which is a consumer of `limel-chip-set`) the default `autocomplete` suggestions of some browsers like Microsoft Edge, create user experience issues.

The experience is more annoying in the Picker, as it has its own list of suggestions. There is no point in overlaying another list on top of this list. 

By generating a random `autocomplete` value and defaulting to that in the Chip set component, we prevent browsers from offering autocomplete suggestions, while still using "off" to indicate that autocomplete is not desired. 

While there might be scenarios that the Chip set can benefit from auto suggestions. Therefore, I kept the `autocomplete` as a `prop` here. But I am not sure if this is the right way. Maybe we should just not use any prop at all, and just directly default to ``off ${crypto.randomUUID()}`;`

adrianschmidt: Yeah, I think we should avoid forcing autocomplete to be disabled without at least leaving the consumer the option to enable it again.

> Maybe we should just not use any prop at all, and just directly default to `off ${crypto.randomUUID()};`

Just to clarify some terminology here:
- If we would have no prop at all, that would be "hardcoding" the value.
- If we use `off ${crypto.randomUUID()}` as the default value, it would look like this:
  ```ts
  @Prop({ reflect: true })
  public autocomplete?: string = `off ${crypto.randomUUID()}`;
  ```

We should use the latter 🙂

Kiarokh: OK. I change how I default the value. My thought was we don't have to explain to the consumers why we use the random value. That's why I took it further down in the code. Anyhow, this one is also good. Not sure we need to explain why we default to what we default to. But it also doesn't hurt.

jgroth:
I still get autocompletion :disappointed: 

image
image

jgroth:
Did some quick hacking and it seems if I set `autocomplete="off"` on the form element it works for components that are in a form!

image

jgroth:
Also found this https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Turning_off_form_autocompletion

> If a browser keeps on making suggestions even after setting autocomplete to off, then you have to change the `name` attribute of the `<input>` element.

jgroth:
After some more hacking I think I got rid of the autocompletions by setting a new `name` attribute on the input field. It feels a bit sketchy to set this to a random value though, but we also don't set anything at all today it seems so :man_shrugging:

Kiarokh:
> After some more hacking I think I got rid of the autocompletions by setting a new `name` attribute on the input field. It feels a bit sketchy to set this to a random value though, but we also don't set anything at all today it seems so 🤷‍♂️

Maybe we should also do the same for the `name`. Some prop, followed by some random value? From what I understand, a `name` is needed for form submissions, and how the data is collected.

> The name attribute on an input element in HTML is used to specify a key that represents the input field when the form is submitted. When a form is sent to a server, each input field's value is paired with its name attribute to create key-value pairs that are sent as part of the form data.

adrianschmidt:
> After some more hacking I think I got rid of the autocompletions by setting a new `name` attribute on the input field. It feels a bit sketchy to set this to a random value though, but we also don't set anything at all today it seems so 🤷‍♂️

Ugh… that feels very sketchy yes… 😬

Don't get me wrong, I don't _know_ that it would be bad to do so, but my gut-feeling ain't good man 😅

Worth reading this first at least: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#name

Also, we want to make _very_ sure that the value of `name` isn't something screen readers read out before we do that 🙉 😄

adrianschmidt:
> Also found this [developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Turning_off_form_autocompletion](https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Turning_off_form_autocompletion)
> 
> > If a browser keeps on making suggestions even after setting autocomplete to off, then you have to change the `name` attribute of the `<input>` element.

Ah, sorry, I missed this comment! Then that strategy at least has some fairly authoritative backing! 😄

adrianschmidt:
> I still get autocompletion 😞
> 
> image image

What browser was that, and have you tried setting `autocomplete` to only a uuid, without including `off`?

jgroth:
> Maybe we should also do the same for the `name`. Some prop, followed by some random value? From what I understand, a `name` is needed for form submissions, and how the data is collected.

Yes, `name` is used for form submissions. However, we do not submit our forms in the traditional sense so for us it does not really matter if we set a random value or not to the `name` field. Also, since it's today not possible for consumer to set their own `name` field, they can't really use the components if they put them in their own `form` element (not using `limel-form`  (we should probably fix this :smile: ))

jgroth:
> > I still get autocompletion 😞
> > image image
> 
> What browser was that, and have you tried setting `autocomplete` to only a uuid, without including `off`?

I'm using Chrome on Ubuntu, and I was just using the PR so I did not do anything on my own (except from when I edited the DOM when I added the `name` attribute and `autocomplete` on the `form` element)

jgroth:
> > I still get autocompletion 😞
> > image image
> 
> What browser was that, and have you tried setting `autocomplete` to only a uuid, without including `off`?

Setting to only a UUID did not work, but it seems only setting it to `off` did work! :eyes:

Befkadu1:
It looks setting it to "off" on Microsoft Edge looks good. I tested it on the published link provided in the PR. 

https://github.com/user-attachments/assets/cda5c57e-3efe-4233-8df9-754d9381ad2c

Befkadu1 (approved):

jgroth (approved):

PR #3206, opened 2024-09-26
fix(markdown): fix broken default config - allow target attribute on links and more

adrianschmidt:
This was broken back in 43e89bfc4e936e4d1c2c300820ac87888d3aa6ec, which was only partially fixed in c452f3076e1c7dbe36539adb83d684c046f86062.

fix: Lundalogik/crm-feature#4348

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/markdown/markdown-parser.ts
+++ b/src/components/markdown/markdown-parser.ts
@@ -65,8 +65,11 @@ function getWhiteList(allowedComponents: CustomElement[]): Schema {
         ],
         attributes: {
             ...defaultSchema.attributes,
-            p: [['className', 'MsoNormal']], // Allow the class 'MsoNormal' on <p> elements
-            '*': ['style', 'width'], // Allow `style` and 'width' attribute on all elements
+            p: [
+                ...(defaultSchema.attributes.p ?? []),
+                ['className', 'MsoNormal'],
+            ], // Allow the class 'MsoNormal' on <p> elements
+            '*': [...(defaultSchema.attributes['*'] ?? []), 'style'], // Allow `style` attribute on all elements
         },
     };
 


adrianschmidt:
I've tested this with the following in the composite example:

```html
# Hello, world!

This is **markdown**!

<a href="https://www.google.com" target="_blank" rel="noreferrer">Google</a>

<img src="https://www.alleycat.org/wp-content/uploads/2019/03/FELV-cat.jpg" width="50%">
<img src="https://www.alleycat.org/wp-content/uploads/2019/03/FELV-cat.jpg" style="width: 60%">

<p style="color: red">test</p>
```

adrianschmidt:
Thanks @Befkadu1 for finding where the problem was introduced! I dug around a little, and that made me realise that this had happened before we had overridden the default config, and that the proper fix was to make sure we included the default config and _added_ our own values to that, and that we had to do that on every level.

So this PR fixes the problem at its root, so to speak.

adrianschmidt:
> Now internal links opens in new tab. Is it what we want?

Hmm… no, it's not… I guess we've got some code somewhere that's adding `target="_blank"` even though it's an internal link… 🤔

Can you open an issue for that?

Befkadu1:
> Can you open an issue for that? 🙏

I have opened another issue [here](https://github.com/Lundalogik/crm-feature/issues/4397).

PR #3210, opened 2024-09-27
Fix/3209 render partial html

### LINKED ISSUE #3209: limel-markdown: component renders partial HTML if linebreaks separates some elements
## Current behavior

This is a continued conversation from this issue: https://github.com/Lundalogik/lime-elements/issues/3149

Faulty rendering of HTML if line breaks separate opening and closing of elements
Here's an example that will render incorrectly:

It will render the `</style></span>` as text

```
<div><span><style>

</style></span>

Hello2
</div>
```

Image from [Lime elements docs - v37.61.8](https://lundalogik.github.io/lime-elements/versions/37.61.8/#/component/limel-markdown/)

image

## Expected behavior

It shouldn't render the HTML elements, but instead render it as a complete HTML.
The above example should just state the text `Hello2`

## Suggested solution by John from previous issue

> We should be able to do that. I did a quick test using Parse5 and added a small plugin to the pipeline. 
> 
> ```ts
> .use(() => {
>     return (tree: Node) => {
>         visit(tree, 'html', (node: any) => {
>             if (node.type === 'html') {
>                 const documentFragment = parse5.parseFragment(
>                     node.value,
>                 );
>                 node.value = parse5.serialize(documentFragment);
>             }
>         });
>     };
> });
> ```
> 
> This proved to at least fix the specific example you gave above but we'd probably want to take a bit more time investigating and solving this. I'll create an issue for it on our cooldown board. 

## Environment

- **lime-elements** version: 37.61.8

TommyLindh2:
Fixes #3209 

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/markdown/markdown-parser.ts
+++ b/src/components/markdown/markdown-parser.ts
@@ -1,6 +1,7 @@
 import { unified } from 'unified';
 import remarkParse from 'remark-parse';
 import remarkRehype from 'remark-rehype';
+import * as parse5 from 'parse5';
 import remarkGfm from 'remark-gfm';
 import rehypeExternalLinks from 'rehype-external-links';
 import rehypeSanitize, { defaultSchema } from 'rehype-sanitize';
@@ -37,6 +38,17 @@ export async function markdownToHTML(
     const file = await unified()
         .use(remarkParse)
         .use(remarkGfm)
+        .use(() => {
+            return (tree: Node) => {
+                // Run the parse5.serialize function on all html elements,
+                // to minify it to get rid of extra whitespace that causes rendering issues
+                // when elements are opened/closed with empty lines between them.
+                visit(tree, 'html', (node: any) => {
+                    const documentFragment = parse5.parseFragment(node.value);
+                    node.value = parse5.serialize(documentFragment);
+                });
+            };
+        })
         .use(remarkRehype, { allowDangerousHtml: true })
         .use(rehypeExternalLinks, { target: '_blank' })
         .use(rehypeRaw)
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,6 +51,7 @@
         "moment": "^2.30.1",
         "number-abbreviate": "^2.0.0",
         "parse-css-color": "^0.2.1",
+        "parse5": "^7.1.2",
         "prettier": "^3.3.3",
         "prosemirror-example-setup": "^1.2.3",
         "prosemirror-markdown": "^1.13.1",
@@ -5775,18 +5776,6 @@
         "url": "https://github.com/sponsors/wooorm"
       }
     },
-    "node_modules/hast-util-raw/node_modules/parse5": {
-      "version": "7.1.2",
-      "resolved": "https://registry.npmjs.org/parse5/-/parse5-7.1.2.tgz",
-      "integrity": "sha512-Czj1WaSVpaoj0wbhMzLmWD69anp2WH7FXMB9n1Sy8/ZFF9jolSQVMu1Ij5WIyGmcBmhk7EOndpO4mIpihVqAXw==",
-      "dev": true,
-      "dependencies": {
-        "entities": "^4.4.0"
-      },
-      "funding": {
-        "url": "https://github.com/inikulin/parse5?sponsor=1"
-      }
-    },
     "node_modules/hast-util-raw/node_modules/property-information": {
       "version": "6.3.0",
       "resolved": "https://registry.npmjs.org/property-information/-/property-information-6.3.0.tgz",
@@ -8332,6 +8321,12 @@
         "url": "https://opencollective.com/unified"
       }
     },
+    "node_modules/kompendium/node_modules/parse5": {
+      "version": "5.1.1",
+      "resolved": "https://registry.npmjs.org/parse5/-/parse5-5.1.1.tgz",
+      "integrity": "sha512-ugq4DFI0Ptb+WWjAdOK16+u/nHfiIrcE+sh8kZMaM0WllQKLI9rOUq6c2b7cwPkXdzfQESqvoqK6ug7U/Yyzug==",
+      "dev": true
+    },
     "node_modules/kompendium/node_modules/rehype-raw": {
       "version": "4.0.2",
       "resolved": "https://registry.npmjs.org/rehype-raw/-/rehype-raw-4.0.2.tgz",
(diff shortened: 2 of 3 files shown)


TommyLindh2:
Now the style tags are removed/rendered as expected but it seems like the old issue with line breaks in alt texts is back after this fix.

```html
<img alt="Lime elements logo

But with line breaks in it"
width="50px"
height="50px"
src="https://lundalogik.github.io/lime-elements/versions/latest/favicon.svg">
```

**With limebreaks in alt text**
image

**Without limebreaks in alt text**
image

TommyLindh2:
I just realized that the example above doesn't work in the latest release either, so it wasn't introduced in this PR.

@john-traas @adrianschmidt any ideas?

src/components/markdown/markdown-parser.ts
```diff
@@ -35,6 +36,17 @@ export async function markdownToHTML(
...
+                // to minify it to get rid of extra whitespace that causes rendering issues
+                // when elements are opened/closed with empty lines between them.
+                visit(tree, 'html', (node: any) => {
+                    const documentFragment = parse5.parseFragment(node.value);
+                    node.value = parse5.serialize(documentFragment);
+                });
+            };
+        })
         .use(remarkParse)
         .use(remarkGfm)
```
john-traas, line 51:
```suggestion
    const file = await unified()
        .use(remarkParse)
        .use(remarkGfm)
        .use(() => {
            return (tree: Node) => {
                // Run the parse5.serialize function on all html elements,
                // to minify it to get rid of extra whitespace that causes rendering issues
                // when elements are opened/closed with empty lines between them.
                visit(tree, 'html', (node: any) => {
                    const documentFragment = parse5.parseFragment(node.value);
                    node.value = parse5.serialize(documentFragment);
                });
            };
        })
```

john-traas, line 51: This should be placed after the first two functions in the pipeline:
        
```suggestion
    const file = await unified()
        .use(remarkParse)
        .use(remarkGfm)
        .use(() => {
            return (tree: Node) => {
                // Run the parse5.serialize function on all html elements,
                // to minify it to get rid of extra whitespace that causes rendering issues
                // when elements are opened/closed with empty lines between them.
                visit(tree, 'html', (node: any) => {
                    const documentFragment = parse5.parseFragment(node.value);
                    node.value = parse5.serialize(documentFragment);
                });
            };
        })
```

john-traas:
>  it seems like the old issue with line breaks in alt texts is back 

Odd, I've never seen that issue before.

src/components/markdown/markdown-parser.ts
TommyLindh2: Fixed, I did it together with a rebase with `main`.

_I was trying to move it around to see if it helped the issue 🤷‍♂️😅_

TommyLindh2:
> > it seems like the old issue with line breaks in alt texts is back
> 
> Odd, I've never seen that issue before.

Do you think we should ignore it in this PR because it's not the same issue or try to fix it at the same time?

john-traas:
> > > it seems like the old issue with line breaks in alt texts is back
> > 
> > 
> > Odd, I've never seen that issue before.
> 
> Do you think we should ignore it in this PR because it's not the same issue or try to fix it at the same time? Because it I understand it correctly having line breaks in an attribute is not a good practice because some HTML parsers allow it while some don't. So maybe we shouldn't even allow it?

I'm of the opinion that we shouldn't allow it. This is also meant to be a component that parses Markdown, not that fixes broken HTML. 😅 
From my perspective the initial issue that this PR fixes is one that seems reasonable for a user to assume, that closing tags would be recognised. 
Attribute values should not have line breaks within them. If a user inputs HTML like that, the component won't crash and I think the user seeing that it doesn't render correctly is fine. They can edit the content to fix the issue.

TommyLindh2:
> > > > it seems like the old issue with line breaks in alt texts is back
> > > 
> > > 
> > > Odd, I've never seen that issue before.
> > 
> > 
> > Do you think we should ignore it in this PR because it's not the same issue or try to fix it at the same time? Because it I understand it correctly having line breaks in an attribute is not a good practice because some HTML parsers allow it while some don't. So maybe we shouldn't even allow it?
> 
> I'm of the opinion that we shouldn't be concerned about it. This is also meant to be a component that parses Markdown, not that fixes broken HTML. 😅 From my perspective the initial issue that this PR fixes is one that seems reasonable for a user to assume, that closing tags would be recognised. Attribute values should not have line breaks within them. If a user inputs HTML like that, the component won't crash and I think the user seeing that it doesn't render correctly is fine. They can edit the content to fix the issue.
> 
> Do you have any thoughts on this @adrianschmidt?

I totally agree! 👍

blombergedvin:
I just want to bump this a bit more. It would be awesome if some sort of fix comes out during cooldown. A lot of customers (especially bigger ones) are experiencing HTML in their Communication Flow feed, and we should try to help them.

src/components/markdown/markdown-parser.ts
```diff
@@ -37,6 +38,17 @@ export async function markdownToHTML(
...
+        .use(() => {
+            return (tree: Node) => {
+                // Run the parse5.serialize function on all html elements,
+                // to minify it to get rid of extra whitespace that causes rendering issues
+                // when elements are opened/closed with empty lines between them.
+                visit(tree, 'html', (node: any) => {
+                    const documentFragment = parse5.parseFragment(node.value);
+                    node.value = parse5.serialize(documentFragment);
```
adrianschmidt, line 48: This doesn't quite handle html properly.

Look at the "HTML" example on this branch vs latest release for example:

**This PR:**
image

**Main:**
image

Also, the inline `<code>` element in the "Code" example isn't handled properly:

**This PR:**
image

**Main:**
image

adrianschmidt (changes requested):
It seems like markdown is handled properly, but html is sometimes mangled…

src/components/markdown/markdown-parser.ts
TommyLindh2: @john-traas is this something you have time to take a look at?

john-traas: I've been out sick for a few days but will be back from tomorrow. I'll look further into this and try and figure out what's going on. 
As I mentions when I first suggested something like this, I hadn't tested more use cases and it was just an initial exploration into how we might fix it.

john-traas: On looking further into this I think we might need to delegate responsibility to the consumer.

adrianschmidt: So, @TommyLindh2, either the consumer using the `limel-markdown` component needs to pre-process the data before sending it to `limel-markdown`, or, and I would say preferably, the component responsible for storing the data in the first place should process it and make sure the data isn't broken.

TommyLindh2: Yeah, I think it makes a lot of sense as well.
In this case we'll try to sanitize the content before passing it to limel-markdown.

TommyLindh2:
The component is not meant to render complex HTML, it is meant to render markdown and some simple HTML for formatting content more precise. So we will close this PR/Issue and it's up to the consumer to actually pass a valid content

PR #3240, opened 2024-10-14
feat(chart): add new component

Kiarokh:
while keeping the hidden element accessible to assistive technologies

fix https://github.com/Lundalogik/hack-tuesday/issues/361

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chart/chart.tsx
+++ b/src/components/chart/chart.tsx
@@ -1,0 +1,425 @@
+import { Component, h, Prop, Watch } from '@stencil/core';
+import { Languages } from '../date-picker/date.types';
+import translate from '../../global/translations';
+import { createRandomString } from '../../util/random-string';
+import { ChartItem } from './chart.types';
+
+const PERCENT = 100;
+const DEFAULT_INCREMENT_SIZE = 10;
+
+/**
+ * A chart is a graphical representation of data, in which
+ * visual symbols such as such bars, dots, lines, or slices, represent
+ * each data point, in comparison to others.
+ *
+ * @exampleComponent limel-example-chart-stacked-bar
+ * @exampleComponent limel-example-chart-orientation
+ * @exampleComponent limel-example-chart-max-value
+ * @exampleComponent limel-example-chart-type-bar
+ * @exampleComponent limel-example-chart-type-dot
+ * @exampleComponent limel-example-chart-type-area
+ * @exampleComponent limel-example-chart-type-line
+ * @exampleComponent limel-example-chart-type-pie
+ * @exampleComponent limel-example-chart-type-doughnut
+ * @exampleComponent limel-example-chart-type-ring
+ * @exampleComponent limel-example-chart-type-gantt
+ * @exampleComponent limel-example-chart-type-nps
+ * @exampleComponent limel-example-chart-multi-axis
+ * @exampleComponent limel-example-chart-multi-axis-with-negative-start-values
+ * @exampleComponent limel-example-chart-multi-axis-area-with-negative-start-values
+ * @exampleComponent limel-example-chart-axis-increment
+ * @exampleComponent limel-example-chart-accessibility
+ * @exampleComponent limel-example-chart-styling
+ * @exampleComponent limel-example-chart-creative-styling
+ * @beta
+ */
+
+@Component({
+    tag: 'limel-chart',
+    shadow: true,
+    styleUrl: 'chart.scss',
+})
+export class Chart {
+    /**
+     * Defines the language for translations.
+     * Will translate the translatable strings on the components.
+     */
+    @Prop({ reflect: true })
+    public language: Languages = 'en';
+
+    /**
+     * Helps users of assistive technologies to understand
+     * the context of the chart, and what is being displayed.
+     */
+    @Prop({ reflect: true })
+    public accessibleLabel?: string;
+
+    /**
+     * Helps users of assistive technologies to understand
+     * what the items in the chart represent.
+     */
+    @Prop({ reflect: true })
+    public accessibleItemsLabel?: string;
+
+    /**
+     * List of items in the chart,
+     * each representing a data point.
+     */
+    @Prop()
+    public items!: ChartItem[];
+
+    /**
+     * Defines how items are visualized in the chart.
+     */
+    @Prop({ reflect: true })
+    public type?:
+        | 'area'
+        | 'bar'
+        | 'doughnut'
+        | 'line'
+        | 'nps'
+        | 'pie'
+        | 'ring'
+        | 'dot'
+        | 'stacked-bar' = 'stacked-bar';
+
+    /**
+     * Defines whether the chart is intended to be displayed wide or tall.
+     * Does not have any effect on chart types which generate circular forms.
+     */
+    @Prop({ reflect: true })
+    public orientation?: 'landscape' | 'portrait' = 'landscape';
+
+    /**
+     * Specifies the range that items' values could be in.
+     * This is used in calculation of the size of the items in the chart.
+     * When not provided, the sum of all values in the items will be considered as the range.
+     */
+    @Prop({ reflect: true })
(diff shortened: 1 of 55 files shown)


src/components/chart/examples/chart-resizable-container.scss
```diff
@@ -0,0 +1,64 @@
...
+    position: relative;
+    display: flex;
+    flex-direction: column;
+    gap: 1rem;
+
+    resize: both;
+    overflow: auto;
+
+    box-sizing: border-box;
+    position: relative;
```
jgroth, line 15: Sonarcloud is complaining about this since it's already declared on line 6 :slightly_smiling_face:

src/components/chart/chart.tsx
```diff
@@ -0,0 +1,450 @@
...
+        minRange: number;
+        maxRange: number;
+        totalRange: number;
+    };
+
+    private cachedRange: {
+        minRange: number;
+        maxRange: number;
+        totalRange: number;
+    } | null = null;
```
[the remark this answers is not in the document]
Kiarokh: After the component is rendered, new items might get added to the dataset, which requires recalculations to redraw the items in the chart. Also the axis lines increments may change after the render. 

However, the overlord thought that:

> The function `calculateRange()` is being triggered whenever `axisIncrement`, `items`, or `maxValue` changes. Since this involves iterating through the items and performing calculations, consider memoizing the results using something like a cache to prevent recalculating unnecessarily if the values haven’t changed.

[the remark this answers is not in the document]
jgroth: Yes, I'm not arguing against the calculation itself. However, `rangeData` is only ever set to a new value inside `recalculateRangeData`, like so
```ts
    private recalculateRangeData() {
        this.rangeData = this.calculateRange();
    }
```
and `cachedRange` is only ever set to a new value inside `calculateRange`, which is then returned, so they will always be the same. I don't see why we need both

src/components/chart/chart.tsx
```diff
@@ -0,0 +1,450 @@
...
+    @Prop({ reflect: true })
+    public orientation?: 'landscape' | 'portrait' = 'landscape';
+
+    /**
+     * Specifies the range that items' values could be in.
+     * This is used in calculation of the size of the items in the chart.
+     * When not provided, the sum of all values in the items will be considered as the range.
+     */
+    @Prop({ reflect: true })
+    public maxValue?: number;
```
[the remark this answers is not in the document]
jgroth: I didn't fix this, we can add it later if we need it

src/components/chart/chart.tsx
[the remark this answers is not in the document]
Kiarokh: 🙈

src/components/chart/examples/chart-items-creative-styling.ts
```diff
@@ -26,26 +26,23 @@ export const stackedBarChartItems: ChartItem[] = [
...
 ];
 
-export const ganttChartItems: ChartItem[] = [
+export const ganttChartItems: Array<ChartItem<[number, number]>> = [
     {
         text: 'Week 1',
         prefix: '%',
-        startValue: -7,
-        value: 10,
+        value: [-7, 10],
```
Kiarokh, line 33: 👍 
there are places in the docs where I have mentioned `minValue`, which we should also change

src/components/chart/chart.types.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    /**
+     * A prefix shown before the value
+     */
+    prefix?: string;
+
+    /**
+     * A suffix shown after the value.
+     */
+    suffix?: string;
```
jgroth, line 26: I think we should remove these from this interface and find a better way to format the values

Kiarokh: I don't know what you mean by "find a better way". Can you explain a bit more?

But your comments reminds me of what you did in the info tiles, using standard browser-formatting of values. Is that what you mean?

Also keep in mind that these are used to improve accessibility.

src/components/chart/chart.tsx
```diff
@@ -0,0 +1,377 @@
...
+     */
+    @Prop({ reflect: true })
+    public type?:
+        | 'area'
+        | 'bar'
+        | 'doughnut'
+        | 'line'
+        | 'pie'
+        | 'ring'
+        | 'scatter'
```
jgroth, line 81: If we can get a real example for the `scatter` type to work we can keep this, otherwise we should remove it for now

Kiarokh: I'll rename this current `scatter` to `dot`, and will add scatter separately

src/components/chart/chart.types.ts
jgroth: It's better to let the consumer handle the formatting of the values, since we do not know anything about them. I added a `formattedValue` prop now instead that will be used if provided

jgroth: We could also have added some formatter function that the consumer could provide, but this seemed like the easiest approach right now

src/components/chart/chart.tsx
Kiarokh: I think this is the correct term, or at least the best that I find.

shot from the same page: 
image

 We could also call them `point chart` but that' just a "meh…😑 "

src/components/chart/chart.tsx
```diff
@@ -0,0 +1,381 @@
...
+        | 'stacked-bar' = 'stacked-bar';
```
jgroth, line 82: When I hear "stacked bar" I think of something like this
image

Our example here only has one bar though so it seems more like a pie chart that has been flattened :smile: Perhaps we should call it something else unless we add support for multiple stacked bars

src/components/chart/chart.tsx
```diff
@@ -0,0 +1,381 @@
...
+     */
+    @Prop()
+    public items!: ChartItem[];
+
+    /**
+     * Defines how items are visualized in the chart.
+     */
+    @Prop({ reflect: true })
+    public type?:
+        | 'area'
```
jgroth, line 75: As with the stacked bar example, I also often see this having multiple values on top of each other, e.g.
image

src/components/chart/chart.tsx
Kiarokh: I'd say these are chart-sets and require sending multiple arrays of items into the component. We should either make another component and call it for instance `chart-set`, or move the current one into a private sub-component. I don't know. 

Currently the consumer can make these chart-sets on their own, knowing a little CSS and having multiple `limel-charts`. But if we wanna handle them, we can of course. We (or the consumer) should just make sure that the min and max values in the datasets are the same, and the data has the same range. This means we need to add the `minValue` prop too, otherwise for those charts with X and Y axis, we will get the zero-line in different locations.

Also, such chart-sets (e.g. area or lines or bars) will be overlayed on each other. It means we will absolutely position each chart on the previous chart. There is no way in normal CSS and HTML to hover on the layers which are placed below the top-most layer. (we already have this issue with the pie chart, you can't hover the items since they are all layers on top of each other). This will be a downside of having such charts. We need to solve it then by proviing an interactive "Legend" where items inside the chart can be highlighted if you hover on their equivalents inside the legend.

src/components/chart/chart.types.ts
Kiarokh: first I thought it's like `limel-info-tile` where we have suffix and prefix. But I think the story is completely different there, since we render them in different places in the UI. So this approach is perhaps better. But on the other hand, it forces the consumer to also come up with a way to create their formatted value (which is OK)

jgroth (approved):

PR #3243, opened 2024-10-15
[WIP] prosemirror tables

DavidAkkerman:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/text-editor/examples/text-editor-with-tables.tsx
+++ b/src/components/text-editor/examples/text-editor-with-tables.tsx
@@ -14,7 +14,7 @@ import { Component, h, State } from '@stencil/core';
 export class TextEditorWithTablesExample {
     @State()
     private value: string =
-        '<table><tbody><tr><td style="background-color: rgb(25, 107, 36);color: white;"><p><strong>Column1</strong></p></td><td style="background-color: rgb(25, 107, 36);color: white;"><p><strong>Column2</strong></p></td></tr><tr><td style="background-color: rgb(193, 240, 200);color: black;"><p>Cell A1</p></td><td style="background-color: rgb(193, 240, 200);color: black;"><p>Cell B1</p></td></tr><tr><td style="color: black;"><p>Cell A2</p></td><td style="background-color: yellow;color: red;"><p>Cell B2</p></td></tr></tbody></table>';
+        '<table><tbody><tr><td style="background-color: rgb(25, 107, 36);color: white;"><p><strong>Column1</strong></p></td><td style="background-color: rgb(25, 107, 36);color: white;"><p><strong>Column2</strong></p></td></tr><tr><td style="background-color: rgb(193, 240, 200);color: black;"><p>Cell A1</p></td><td style="background-color: rgb(193, 240, 200);color: black;"><p>Cell B1</p></td></tr><tr><td style="color: green;"><p>Cell A2</p></td><td style="background-color: yellow;color: red;"><p>Cell B2</p></td></tr></tbody></table>';
 
     @State()
     private readonly = false;
--- a/src/components/text-editor/examples/text-editor-with-tables.tsx
+++ b/src/components/text-editor/examples/text-editor-with-tables.tsx
@@ -1,0 +1,50 @@
+import { Component, h, State } from '@stencil/core';
+/**
+ * Text editor with tables (HTML mode only).
+ *
+ * When using the text editor in HTML mode, it is possible to paste and
+ * display tables in the text editor.
+ * Basic interaction with the table is supported, but you cannot do
+ * complex operations
+ */
+@Component({
+    tag: 'limel-example-text-editor-with-tables',
+    shadow: true,
+})
+export class TextEditorWithTablesExample {
+    @State()
+    private value: string =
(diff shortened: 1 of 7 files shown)


adrianschmidt:
@DavidAkkerman, just an FYI: I've assigned you to this, because it's on the Feature Team board, and we have the workflow that the person that made the PR remains assigned when the PR is put in review, in order to keep some sense of ownership from the developer who made the PR. Then the reviewer _also_ assigns themself to the issue.

So, in the review column (on our board), one assignee means it's ready for review, two assignees means it's being reviewed 😊

src/components/text-editor/prosemirror-adapter/plugins/table-plugin.ts
```diff
@@ -0,0 +1,35 @@
...
+        cellAttributes: {
+            background: {
+                default: null,
+                getFromDOM: (dom) => {
+                    return dom.style.backgroundColor || null;
+                },
+                setDOMAttr: (value: string, attrs: Record<string, string>) => {
+                    if (value) {
+                        attrs.style =
+                            (attrs.style || '') + `background-color: ${value};`;
```
adrianschmidt, line 25: We probably need to check that we aren't opening up a vulnerability to script injection here. We might need to sanitise the value before using it.

src/components/text-editor/prosemirror-adapter/plugins/table-plugin.ts
```diff
@@ -0,0 +1,42 @@
...
```
adrianschmidt, line 25:
```suggestion
                            (attrs.style || '') + ` background-color: ${value};`;
```

I think we need to add a space, like that ☝️, right? 🤔

adrianschmidt: Ah, oh… no, of course, as long as the existing value ends with a semi-colon, we don't need the space 🤦‍♂️

src/components/text-editor/prosemirror-adapter/plugins/table-plugin.ts
```diff
@@ -0,0 +1,42 @@
...
+                },
+            },
+            color: {
+                default: null,
+                getFromDOM: (dom) => {
+                    return dom.style.color || null;
+                },
+                setDOMAttr: (value: string, attrs: Record<string, string>) => {
+                    if (value) {
+                        attrs.style = (attrs.style || '') + `color: ${value};`;
```
adrianschmidt, line 36: Same here, we need to consider injection attacks.

src/components/text-editor/examples/text-editor-with-tables.tsx
```diff
@@ -26,6 +26,7 @@ export class TextEditorWithTablesExample {
                 onChange={this.handleChange}
                 readonly={this.readonly}
                 contentType="html"
+                enableTables={true}
```
adrianschmidt, line 29: Unless I've missed something, it seems like this last commit (`fix: configure table handling`) should just be a part of the preceding commit, rather than a commit of its own?

DavidAkkerman: It should, but I didn't want to squash in case it's preferred to put the configuration in a different way

adrianschmidt: Alright, I'll change it to a fixup commit then 👍

src/components/text-editor/prosemirror-adapter/plugins/table-plugin.ts
DavidAkkerman: I did a quick test, and it seems like any tags are encoded if not explicitly allowed
image

Also tried to research a bit, and it seems that the prosemirror developers are generally quite on top of protecting against these kind of things, such as discussed in this forum post https://discuss.prosemirror.net/t/heads-up-xss-risk-in-domserializer/6572.

I do feel a bit out of my comfort zone with this topic though, so I would appreciate a second opinion :)

adrianschmidt: Yeah, it sounds very reasonable that prosemirror would come with good protections built-in, or at least possible to easily enable (in which case I hope we've done so). What I'm concerned about is if directly manipulating DOM attributes like this, in a plugin, bypasses the built-in protections? I'm a bit unsure where the value of the `value` parameter comes from? It from the pasted content originally, of course, but has that all first gone through sanitising first, or are we reading it straight from the raw input?

You don't have to be the one who can answer these questions @DavidAkkerman. I'm guessing that @john-traas may have already looked into stuff like this in his previous work on the text editor, and if not, I, John, or someone else from our team can look it up 🙂

anderssonjohan:
FYI - SonarCloud Scan will fail as long as the PR is created in a fork of this repo. We don't allow that (sharing the SONARCLOUD_TOKEN with the fork). I suggest recreating the branch in this repo instead.

Screenshot 2024-11-14 at 11 43 17

DavidAkkerman:
@john-traas I don't have permission to create a branch in this repo, could you do this? Then we can merge to that branch instead

src/components/text-editor/prosemirror-adapter/plugins/table-plugin.ts
john-traas: Just adding this for clarity in case anyone is reading in future. We have updated prosemirror-model to version 1.22.1 as per this [thread here](https://discuss.prosemirror.net/t/heads-up-xss-risk-in-domserializer/6572)

PR #3245, opened 2024-10-16
fix(checkbox): teardown and re-initialize component when readonly is changed

john-traas:
When the readonly value was changed the checkbox was not being torn down or re-initialized. This caused the element to return as undefined. 

fixes https://github.com/Lundalogik/crm-feature/issues/4376

To test link lime-elements locally and run a database with a limeobject which has a yesno property and is configured with variants. One variant should modify the yesno property to be readonly. 

Switch between variants causing the yesno field to switch between readonly and editable. 

After switching between variants the yesno field should be editable and if you change the value, save and reload it, the value should remain with the value it was saved with.


src/components/checkbox/checkbox.tsx
```diff
@@ -136,6 +136,14 @@ export class Checkbox {
         this.mdcCheckbox.indeterminate = newValue;
     }
 
+    @Watch('readonly')
+    protected handleReadonlyChange() {
+        this.destroyMDCInstances();
+        requestAnimationFrame(() => {
+            this.initialize();
+        });
+    }
```
civing, line 145: I'm a little bit wary about the cognetive complexity of the `requestAnimationFrame` and what's the most canonical way of achieving the intent here 😄

I think I might prefer this
```ts
  @Watch('readonly')
  protected handleReadonlyChange() {
    this.destroyMDCInstances();
    this.shouldReinitialize = true;
  }

  componentDidRender() {
    if (this.shouldReinitialize) {
      this.initialize();
      this.shouldReinitialize = false;
    }
  }
```
To me it's more intuitive that the intent is to initialize the MDC component after a render. But I'm open to the fact that the current implementation is the most sound one for all of you. Can I ask your opinion @jgroth?

jgroth: This seems more stable, I'm not sure if it could take more than one frame to render it and if so, the original approach would fail

civing (approved):

PR #3262, opened 2024-10-23
fix(markdown): better image resizing

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/4422

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


FredrikWallstrom (approved):

src/components/markdown/markdown-parser.ts
```diff
@@ -57,6 +57,12 @@ export async function markdownToHTML(
 }
 
 function getWhiteList(allowedComponents: CustomElement[]): Schema {
+    const defaultSchemaClone = [...(defaultSchema.attributes['*'] ?? [])];
+    const asteriskAttributeWhitelist = defaultSchemaClone.filter((attr) => {
+        return attr !== 'height';
+    });
+    asteriskAttributeWhitelist.push('style');
```
adrianschmidt, line 64: Huh… merged by me? 😄

Looking closer at this, I'm not sure I love that we're building this into limel-markdown.

It's good that we are consistent, because we were already removing `height` from the `style` attribute, and now we also remove the `height` attribute.

But it feels hacky, considering that we're making it built-in in core part of `limel-markdown`. This feels like something we might want to apply in a layer outside of lime-elements… 🤔

I don't know. I just wanted to write down my thoughts somewhere 😆

PR #3267, opened 2024-10-28
Add trigger plugin interface

john-traas:
This PR is redundant


john-traas:
This was a prototype implementation.

PR #3268, opened 2024-10-28
Trigger plugin

civing:
Adds triggers to be used when detecting a certain string and starting a trigger session. A trigger session begins when the magic string sequence is inserted, for instance an @ and will keep on sending user input back to the consumer until certain stop conditions occur.

fixes: Lundalogik/crm-feature#4430
Co-authored-by: John Traas <john.traas@lime.tech>

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/text-editor/examples/text-editor-custom-triggers.tsx
+++ b/src/components/text-editor/examples/text-editor-custom-triggers.tsx
@@ -1,0 +1,270 @@
+import {
+    Button,
+    LimelListCustomEvent,
+    ListItem,
+} from '@limetech/lime-elements';
+import {
+    Component,
+    h,
+    State,
+    Element,
+    Event,
+    EventEmitter,
+    Watch,
+} from '@stencil/core';
+import { createRandomString } from 'src/util/random-string';
+import { portalContains } from '../../portal/contains';
+import { ESCAPE } from '../../../util/keycodes';
+import { TextEditor, TriggerEventDetail } from '../text-editor.types';
+
+/**
+ * Custom triggers
+ *
+ * A trigger is a character or sequence of characters that if typed in the text editor
+ * will initiate a trigger session. The session is initialized with a `triggerStart`
+ * event. Subsequent characters written after the trigger sequence will be sent in a
+ * `triggerChange` event. When the focus is removed from the trigger a `triggerStop`
+ * event will be sent.
+ *
+ * The `triggerStart` event contains a `TextEditorInserter` object containing functions
+ * to manipulate the state of the text editor around the trigger. Using any of the
+ * supplied methods will effectivly replace the trigger content in the text editor with
+ * the content of choice.
+ *
+ * In this example we pass either a text or a `limel-chip` representing some chosen user
+ * in a mention like situation.
+ */
+@Component({
+    tag: 'limel-example-text-editor-triggers',
+    shadow: true,
+    styleUrl: 'text-editor-custom-triggers.scss',
+})
+export class TextEditorCustomTriggersExample {
+    constructor() {
+        this.portalId = createRandomString();
+        this.globalClickListener = this.globalClickListener.bind(this);
+    }
+    @State()
+    private value: string = '';
+
+    @State()
+    private triggerState: string = '';
+
+    @State()
+    private inputText: string = '';
+
+    @State()
+    private isPickerOpen: boolean = false;
+
+    @State()
+    private textEditorElement: HTMLElement;
+
+    @State()
+    private insertMode: 'text' | 'chip' = 'text';
+
+    @Element()
+    private host: HTMLLimelPopoverElement;
+
+    /**
+     * Emits an event when the component is closing
+     */
+    @Event()
+    private close: EventEmitter<void>;
+
+    @Watch('isPickerOpen')
+    protected watchOpen() {
+        this.setupGlobalHandlers();
+    }
+
+    public componentWillLoad() {
+        this.setupGlobalHandlers();
+    }
+
+    private setupGlobalHandlers() {
+        if (this.isPickerOpen) {
+            document.addEventListener('click', this.globalClickListener, {
+                capture: true,
+            });
+            document.addEventListener('keyup', this.handleGlobalKeyPress);
+        } else {
+            document.removeEventListener('click', this.globalClickListener);
+            document.removeEventListener('keyup', this.handleGlobalKeyPress);
+        }
+    }
+
+    private handleGlobalKeyPress = (event: KeyboardEvent) => {
+        if (event.key !== ESCAPE) {
+            return;
+        }
+
+        event.stopPropagation();
+        event.preventDefault();
+        this.close.emit();
+    };
+
+    private insertModeButtons: Button[] = [
+        {
+            id: '1',
+            title: 'text',
+            selected: true,
+        },
+        {
+            id: '2',
+            title: 'chip',
+        },
+    ];
+
+    private portalId: string;
+    private items: Array<ListItem<number>> = [
+        { text: 'Wolverine', value: 1, icon: 'wolf' },
+        { text: 'Captain America', value: 2, icon: 'captain_america' },
+        { text: 'Superman', value: 3, icon: 'superman' },
+        { text: 'Tony Stark', value: 4, icon: 'iron_man' },
+        { text: 'Batman', value: 5, icon: 'batman_old' },
+    ];
+
+    private triggerFunction?: TextEditor;
+
+    public render() {
+        return [
+            <limel-text-editor
+                style={{ display: 'block' }}
+                ref={(el) => (this.textEditorElement = el)}
+                value={this.value}
+                plugins={[
+                    { tagName: 'limel-chip', attributes: ['text', 'icon'] },
+                ]}
+                triggers={['@']}
+                onTriggerStart={this.handleTriggerStart}
+                onTriggerStop={this.handleTriggerStop}
+                onTriggerChange={this.handleTriggerChange}
+                onChange={this.handleChange}
+            />,
+            <limel-example-controls>
+                Insert mode:
+                <limel-button-group
+                    class="mode"
+                    value={this.insertModeButtons}
+                    onChange={this.handleInsertModeChange}
+                />
+                <div class="value">
+                    <limel-example-value
+                        label="Action"
+                        value={this.triggerState}
+                    />
+                    <limel-example-value
+                        label="Tag value"
+                        value={this.inputText}
+                    />
+                </div>
+            </limel-example-controls>,
+            this.renderPicker(),
+            <limel-example-value value={this.value} />,
+        ];
+    }
+
+    private handleTriggerStart = (event: CustomEvent<TriggerEventDetail>) => {
+        this.triggerState = 'start';
+        this.isPickerOpen = true;
+        this.triggerFunction = event.detail.textEditor;
+    };
+
+    private handleTriggerStop = () => {
+        this.triggerState = 'stop';
+        this.inputText = '';
+        this.isPickerOpen = false;
+    };
+
+    private handleTriggerChange = (event: CustomEvent<TriggerEventDetail>) => {
+        this.inputText = event.detail.value.toLowerCase();
+    };
+
+    private handleChange = (event: CustomEvent<string>) => {
+        this.value = event.detail;
+    };
+
+    private renderPicker = () => {
+        if (!this.isPickerOpen) {
+            return;
+        }
+
+        const items = this.items.filter((item: ListItem<number>) =>
+            item.text.toLowerCase().includes(this.inputText),
+        );
+
+        const dropdownZIndex = getComputedStyle(this.host).getPropertyValue(
+            '--dropdown-z-index',
+        );
+
+        return [
+            <limel-portal
+                containerStyle={{
+                    'background-color': 'rgb(var(--contrast-100))',
+                    'border-radius': '0.5rem',
+                    'box-shadow': 'var(--shadow-depth-16)',
+                    'z-index': dropdownZIndex,
+                }}
+                containerId={this.portalId}
+                visible={this.isPickerOpen}
+                openDirection="bottom-start"
+                inheritParentWidth={true}
+                anchor={this.textEditorElement}
+            >
+                {this.renderList(items)}
+            </limel-portal>,
+        ];
+    };
+
+    private globalClickListener(event: MouseEvent) {
+        const element: HTMLElement = event.target as HTMLElement;
+        const clickedInside = portalContains(this.host, element);
+        if (this.isPickerOpen && !clickedInside) {
+            event.stopPropagation();
+            event.preventDefault();
+            this.isPickerOpen = false;
+            this.close.emit();
+        }
+    }
+
+    private renderList = (items: Array<ListItem<number>>) => {
+        if (items.length === 0) {
+            return (
+                <div style={{ padding: '0.5rem' }}>
+                    Couldn't find. Not a hero yet! 🥲
+                </div>
+            );
+        }
+
+        return (
+            <limel-list
+                items={items}
+                onChange={this.handleListChange}
+                type="selectable"
+            />
+        );
+    };
+
+    private handleListChange = (
+        event: LimelListCustomEvent<ListItem<number>>,
+    ) => {
+        if (this.insertMode === 'text') {
+            this.triggerFunction.insert('@' + event.detail.text);
+
+            return;
+        }
+
+        this.triggerFunction.insert({
+            node: {
+                tagName: 'limel-chip',
+                attributes: {
+                    icon: event.detail.icon,
+                    text: event.detail.text,
+                },
+            },
+        });
+    };
+
+    private handleInsertModeChange = (event: CustomEvent<Button>) => {
+        this.insertMode = event.detail.title as any;
+    };
+}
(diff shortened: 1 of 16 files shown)


src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
```diff
@@ -0,0 +1,303 @@
...
+    | ')'
+    | "'";
+
+export type CustomElementInstance = Omit<CustomElement, 'attributes'> & {
+    attributes: Record<string, any>;
+};
+
+export type TextEditorNode = {
+    node: CustomElementInstance | string;
+    childs?: Array<TextEditorNode | string>;
```
jgroth, line 49:
```suggestion
    children?: Array<TextEditorNode | string>;
```
:sweat_smile:

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
```diff
@@ -0,0 +1,303 @@
...
+};
+
+export interface TextEditorInserter {
```
jgroth, line 52: Maybe this should just be named `TextEditor` if we want it to be able to add more methods to it in the future

```suggestion
export interface TextEditor {
```

src/components/text-editor/text-editor.tsx
```diff
@@ -105,6 +110,9 @@ export class TextEditor implements FormComponent<string> {
     @Prop()
     public plugins: CustomElement[] = [];
```
jgroth, line 111: Not part of this PR, but I think we should rename this to something like `customElements` instead. Calling it `plugins` is a bit confusing for the consumer

src/components/text-editor/text-editor.tsx
```diff
@@ -142,6 +150,15 @@ export class TextEditor implements FormComponent<string> {
...
     public change: EventEmitter<string>;
 
+    @Event()
+    public triggerStart: EventEmitter<TextEditorInserter>;
+
+    @Event()
+    public triggerStop: EventEmitter<void>;
+
+    @Event()
+    public triggerChange: EventEmitter<string>;
```
jgroth, line 160: I think these should be a bit more symmetric, now all of them have different event details. All of them should probably include the `TextEditor` somehow, maybe we need to wrap it in something like this
```ts
interface TriggerEventDetail {
    /**
     * The trigger that triggered this event
     */
    trigger: string;

    /**
     * The text editor
     */
    textEditor: TextEditor;

    /**
     * Current value of the trigger
     */
    value: string
}
```

`trigger` is needed in case you have multiple trigger characters, otherwise you would not know what trigger emitted the event.

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
```diff
@@ -0,0 +1,303 @@
...
+    | ']'
+    | '{'
+    | '}'
+    | '('
+    | ')'
+    | "'";
+
+export type CustomElementInstance = Omit<CustomElement, 'attributes'> & {
+    attributes: Record<string, any>;
+};
```
jgroth, line 45: Why do we need to re-type this? :thinking:

src/components/text-editor/text-editor.tsx
civing: 100% agree. This is the plan as well

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
civing: In defining a `CustomElement` you will have the attributes as a list of strings. When you are actually "instantiating" it you want the attributes as a record. :)

src/components/text-editor/text-editor.tsx
civing: Yes! I like this a lot.

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
civing: The name is maybe a bit generic?

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
civing: Fixed in latest commits

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
civing: Fixed in latest commit

src/components/text-editor/text-editor.tsx
civing: Fixed in latest commit

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
```diff
@@ -0,0 +1,352 @@
...
+     */
+    textEditor: TextEditor;
+
+    /**
+     * Current value of the trigger
+     */
+    value: string;
+}
+
+export const inserterFactory = (view: EditorView): TextEditor => {
```
hannahu, line 73: Have we investigated if there are any existing trigger functionality we can import instead of re-inventing the wheel?

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
```diff
@@ -0,0 +1,352 @@
...
+    view.dom.dispatchEvent(event);
+};
+
+export const createTriggerPlugin = (triggerCharacters: TriggerCharacter[]) => {
+    let activeTrigger: TriggerCharacter | null = null;
+    let triggerText = '';
+    let pluginView: EditorView | null = null;
+    let triggerPosition: number | null = null;
+
+    return new Plugin({
```
hannahu, line 229: Have we investigated if there are any existing trigger functionality we can import instead of re-inventing the wheel?

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
john-traas: Will this possibly become a bit confusing? We have the `text-editor` in lime-elements, and will have the `limebb-text-editor` as well (though perhaps this limebb component should be named differently).

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
john-traas: There's no bundled trigger functionality with ProseMirror. There are two lightweight libraries out there that implement very basic trigger functionality, but they weren't robust enough and didn't suit our needs.

src/components/text-editor/examples/text-editor-custom-triggers.tsx
```diff
@@ -0,0 +1,267 @@
...
+ */
+@Component({
+    tag: 'limel-example-text-editor-triggers',
+    shadow: true,
+    styleUrl: 'text-editor-custom-triggers.scss',
+})
+export class TextEditorCustomTriggersExample {
+    constructor() {
+        this.portalId = createRandomString();
+        this.globalClickListener = this.globalClickListener.bind(this);
```
jgroth, line 44: We don't need explicit binding like this if we use an arrow function instead

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
```diff
@@ -0,0 +1,352 @@
...
+) => {
+    const event = new CustomEvent<TriggerEventDetail>(type, {
+        detail: getTriggerEventDetail(view, trigger, value),
+        bubbles: true,
+        composed: true,
+    });
+    view.dom.dispatchEvent(event);
+};
+
+export const createTriggerPlugin = (triggerCharacters: TriggerCharacter[]) => {
```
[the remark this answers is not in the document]
jgroth: Yes, please split it up :smile:

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
jgroth: I see. I don't like the name `CustomElementInstance` :smile: Why does it need the `Instance` suffix?

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
```diff
@@ -0,0 +1,352 @@
...
+
+import { Node, NodeType, Schema } from 'prosemirror-model';
+import { Plugin, PluginKey, EditorState } from 'prosemirror-state';
+import { ReplaceStep, ReplaceAroundStep } from 'prosemirror-transform';
+import { EditorView } from 'prosemirror-view';
+import { CustomElement } from 'src/interface';
+
+export const triggerPluginKey = new PluginKey('triggerPlugin');
+
+export type TriggerCharacter =
```
jgroth, line 11: What did we say about this? Should we still keep it or do we just have to keep it for now?

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
```diff
@@ -0,0 +1,352 @@
+// import { ListItem } from 'src/components/list/list-item.types';
+
+import { Node, NodeType, Schema } from 'prosemirror-model';
+import { Plugin, PluginKey, EditorState } from 'prosemirror-state';
+import { ReplaceStep, ReplaceAroundStep } from 'prosemirror-transform';
+import { EditorView } from 'prosemirror-view';
+import { CustomElement } from 'src/interface';
+
+export const triggerPluginKey = new PluginKey('triggerPlugin');
```
jgroth, line 9: Do we need to export this?

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
jgroth, line 73: Why not make this into a class? :slightly_smiling_face: Perhaps move it to a separate file?

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
```diff
@@ -0,0 +1,352 @@
...
+    trigger: TriggerCharacter,
+    value: string,
+) => {
+    const event = new CustomEvent<TriggerEventDetail>(type, {
+        detail: getTriggerEventDetail(view, trigger, value),
+        bubbles: true,
+        composed: true,
+    });
+    view.dom.dispatchEvent(event);
+};
```
jgroth, line 221: This might be fine, but I think it would be better if we emit the events from the component instead. We can have a callback on the component that we invoke from here that emits the events, e.g.

```ts
class TextEditor {
    @Event()
    public triggerStart: EventEmitter<TriggerEventDetail>;

    private emitTriggerStart(data) {
        this.triggerStart.emit(...);
    }
}
```

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
civing: Yeah, I had the same thought

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
civing: But wheels are so fun to invent! 🛞

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
```diff
@@ -0,0 +1,352 @@
...
+                                const toPos = step.to;
+
+                                if (slice && slice.size) {
+                                    // Text added
+                                    textAdded += slice.content.textBetween(
+                                        0,
+                                        slice.size,
+                                    );
+                                } else if (fromPos !== toPos) {
+                                    // Text removed
```
jgroth, line 317: image

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
john-traas: We need to import it into the prosemirror adapter so that we can instantiate the plugin when the text editor view state is created.

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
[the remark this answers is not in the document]
civing: See https://github.com/Lundalogik/lime-elements/pull/3268/commits/b50ca82479b1ebda1b610441e2f660ea802b7b27

src/components/text-editor/examples/text-editor-custom-triggers.tsx
```diff
@@ -0,0 +1,267 @@
...
+        const items = this.items.filter((item: ListItem<number>) =>
+            item.text.toLowerCase().includes(this.inputText),
+        );
+
+        const dropdownZIndex = getComputedStyle(this.host).getPropertyValue(
+            '--dropdown-z-index',
+        );
+
+        return [
+            <limel-portal
```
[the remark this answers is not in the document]
civing: : (

src/components/text-editor/examples/text-editor-custom-triggers.tsx
```diff
@@ -0,0 +1,267 @@
...
+ *
+ * In this example we pass either a text or a `limel-chip` representing some chosen user
+ * in a mention like situation.
+ */
+@Component({
+    tag: 'limel-example-text-editor-triggers',
+    shadow: true,
+    styleUrl: 'text-editor-custom-triggers.scss',
+})
+export class TextEditorCustomTriggersExample {
```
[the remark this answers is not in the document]
civing: @LucyChyzhova was adding a lot of good things but yeah, maybe we can simplify

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
civing: I can change it in a follow up PR, there are some more things to do than simply changing the type to string

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
civing: I'm not really sure we need to export it? See https://github.com/Lundalogik/lime-elements/pull/3268/commits/58d429f27b22620d17184f27041dc0c8f2bac6c6

john-traas: Ah, I should have looked more closely 👀 
I was referring to the create plugin function. 🙈

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
civing: See https://github.com/Lundalogik/lime-elements/pull/3268/commits/0c355c81a6abaa86ba36262f648d16d93693539a 😅

src/components/text-editor/examples/text-editor-custom-triggers.tsx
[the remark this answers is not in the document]
LucyChyzhova: sure! will check what we can do

src/components/text-editor/prosemirror-adapter/plugins/trigger/factory.ts
```diff
@@ -0,0 +1,224 @@
...
+    transactions: Transaction[],
+    oldState: EditorState,
+): string => {
+    let textAdded = '';
+    let textRemoved = '';
+
+    transactions.forEach((transaction) => {
+        if (transaction.docChanged) {
+            transaction.steps.forEach((step) => {
+                if (
```
hannahu, line 96: Still some nested code, and also unnecessary inline code comments, but I'm ok merging it and get it in if it works as expected 😄

src/components/text-editor/prosemirror-adapter/plugins/trigger-plugin.ts
civing: I will change the name of `CustomElement` to `CustomElementDefinition` and simply call the new on `CustomElement`

jgroth (approved):

src/components/text-editor/prosemirror-adapter/plugins/trigger/factory.ts
```diff
@@ -0,0 +1,227 @@
...
+            stopTrigger();
+
+            return true;
+        }
+
+        if (isTrigger(event.key, triggerCharacters) && shouldTrigger(state)) {
+            activeTrigger = event.key;
+            triggerText = '';
+            triggerPosition = state.selection.$from.pos - triggerText.length;
+            sendTriggerEvent('triggerStart', view, activeTrigger, triggerText);
```
adrianschmidt, line 158: I came across this when working on https://github.com/Lundalogik/lime-elements/pull/3307. What's `triggerText` for, and is it supposed to ever be anything other than an empty string? If not, we can make the following change:

```diff
             triggerText = '';
-            triggerPosition = state.selection.$from.pos - triggerText.length;
+            triggerPosition = state.selection.$from.pos;
             sendTriggerEvent('triggerStart', view, activeTrigger, triggerText);
```

On the other hand, since it's a variable everywhere, it kinda looks like the intention is that it's not always going to remain an empty string? 🙂

civing: We used it on line 193 in this file. :) 

````ts
        if (updatedText !== triggerText) {
            triggerText = updatedText;
            sendTriggerEvent(
                'triggerChange',
                pluginView,
                activeTrigger,
                triggerText.slice(1),
            );
        }
````

PR #3275, opened 2024-10-30
Fix resize observer webkit warning

jgroth:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/table/table.tsx
```diff
@@ -451,9 +451,12 @@ export class Table {
...
 
         const observer = new ResizeObserver(() => {
-            this.tabulator = new TabulatorTable(table, this.getOptions());
-            this.setSelection();
-            observer.unobserve(table);
+            requestAnimationFrame(() => {
+                this.tabulator = new TabulatorTable(table, this.getOptions());
+                this.setSelection();
+                observer.unobserve(table);
+                observer.disconnect();
```
adrianschmidt, line 458: Oh! This line was missing before… so this also fixes a memory leak I suspect?

adrianschmidt (approved):
I was not able to reproduce the warnings, but the changes look solid 👍

src/components/table/table.tsx
adrianschmidt: Doh! I'm so stupid. Webkit is Safari 🤦‍♂️

I'll have a look there too 😄

PR #3277, opened 2024-10-31
UI improvements in dynamic-label, checkbox, and switch

Kiarokh:
This is important specially because this component is
used to visualize the readonly state of the boolean elements
(checkbox and switch) which don't truncate their text by deafult
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/dynamic-label/dynamic-label.scss
+++ b/src/components/dynamic-label/dynamic-label.scss
@@ -22,7 +22,7 @@ limel-icon {
     ); // The default `false` color. Will be overwritten by `Icon`
 }
 
-span {
+label {
     flex-grow: 1;
     font-size: 0.8125rem; // `13px`, Like Checkbox & Switch
     color: var(--mdc-theme-on-surface);
--- a/src/components/dynamic-label/dynamic-label.scss
+++ b/src/components/dynamic-label/dynamic-label.scss
@@ -1,5 +1,3 @@
-@use '../../style/mixins.scss';
-
 * {
     box-sizing: border-box;
     min-width: 0;
@@ -25,7 +23,6 @@ limel-icon {
 }
 
 span {
-    @include mixins.truncate-text;
     flex-grow: 1;
     font-size: 0.8125rem; // `13px`, Like Checkbox & Switch
     color: var(--mdc-theme-on-surface);
--- a/src/components/dynamic-label/dynamic-label.scss
+++ b/src/components/dynamic-label/dynamic-label.scss
@@ -7,7 +7,7 @@
     --limel-dynamic-label-min-height: 1.75rem;
     display: flex;
     gap: 0.5rem;
-    align-items: center;
+    align-items: flex-start;
     border-radius: 0.5rem;
     min-width: 0;
 }
@@ -26,4 +26,5 @@ span {
     flex-grow: 1;
     font-size: 0.8125rem; // `13px`, Like Checkbox & Switch
     color: var(--mdc-theme-on-surface);
+    padding-top: 0.375rem;
 }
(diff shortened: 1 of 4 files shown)


TommyLindh2:
@Kiarokh 
It still "jump" and doesn't have the same padding when readonly compared to a regular checkbox
Recorded from the [lundalogik.github.io/lime-elements/versions/PR-3277](https://lundalogik.github.io/lime-elements/versions/PR-3277/#/component/limel-checkbox/)
https://github.com/user-attachments/assets/026c1725-7d4b-456d-ac18-95ace636a830

Kiarokh:
Thanks @TommyLindh2. Yes, the last commit broke the results of the first commit 😅  Now fixed

LucyChyzhova:
another jump 🙈 - left/right. Or what do you think?

https://github.com/user-attachments/assets/787fd39f-5b99-4c94-b84d-a839724bf437

src/components/dynamic-label/dynamic-label.scss
```diff
@@ -25,8 +23,9 @@ limel-icon {
 }
 
 span {
-    @include mixins.truncate-text;
     flex-grow: 1;
     font-size: 0.8125rem; // `13px`, Like Checkbox & Switch
+    letter-spacing: 0.02em;
     color: var(--mdc-theme-on-surface);
```
TommyLindh2, line 29: This row: `color: var(--mdc-theme-on-surface);` makes the label another color than the checkbox label.
Not sure it should be removed or not, but it is slightly different I think

src/components/dynamic-label/dynamic-label.scss
```diff
@@ -25,8 +23,9 @@ limel-icon {
 }
 
 span {
-    @include mixins.truncate-text;
     flex-grow: 1;
     font-size: 0.8125rem; // `13px`, Like Checkbox & Switch
+    letter-spacing: 0.02em;
```
Kiarokh, line 28: @TommyLindh2 @LucyChyzhova I'm really glad that you guys also noticed these details. 🤗 

I'm aware of them, which is why I wrote "minimize" "[checkbox: minimize the layout shifts in readonly mode](https://github.com/Lundalogik/lime-elements/pull/3277/commits/f337ef1ccf6a94cbef73ade9133ce0a893bb7c6d)" for the commit message. The thing is that we are gonna remove all MDC stuff from Lime Elements. Switch and Checkbox are two of those components that should also be de-MDCfied, as they come with bugs and limitations. Many styles, specially typographic styles including the letter spacing is coming from the MDC, which cause different renderings.

I didn't want to change these styles, to make the future work a bit simpler. But @LucyChyzhova if you want to change the typographic styles such as the letter spacing, please do it in the Checkbox and Switch instead of inside Dynamic Label. 🙏

LucyChyzhova: > I didn't want to change these styles, to make the future work a bit simpler. But @LucyChyzhova if you want to change the typographic styles such as the letter spacing, please do it in the Checkbox and Switch instead of inside Dynamic Label. 

So you mean we should not apply it to span, but to the checkbox and switch labels instead?

Kiarokh: yes exactly. thanks :)

LucyChyzhova (approved):

PR #3279, opened 2024-10-31
Autofix: Snackbar never shown when timeout is -1

lattaai1:
I have identified the issue in the `limel-snackbar` component where setting the `timeout` to -1 causes the snackbar to hide immediately. To fix this, I've updated the `handleOpen` method in the `Snackbar` class to treat negative timeout values as persistent (no timeout). Here's a summary of the changes:

1. Modified the `handleOpen` method to check if the timeout is greater than 0 before setting the timeout.
2. Updated the JSDoc comment for the `timeout` property to clarify the behavior with different values.

These changes ensure that setting `timeout` to -1 (or any negative value) will make the snackbar persistent, matching the expected behavior described in the problem statement. 
> [!CAUTION]  
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

### DIFF
--- a/src/components/snackbar/snackbar.tsx
+++ b/src/components/snackbar/snackbar.tsx
@@ -67,9 +67,9 @@ export class Snackbar {
 
     /**
      * The amount of time in milliseconds to show the snackbar.
-     * If set to `null`, the snackbar will be persistent.
+     * If set to `null` or any value less than or equal to 0, the snackbar will be persistent.
      * This means:
-     * - either the end user will need to close is manually,
+     * - either the end user will need to close it manually,
      * which requires the `dismissible` property to be set to `true`.
      * - or the snackbar needs to be closed programmatically.
      */
@@ -181,7 +181,7 @@ export class Snackbar {
         this.closing = false;
         container.add(this.host);
 
-        if (this.timeout) {
+        if (this.timeout && this.timeout > 0) {
             this.timeoutId = window.setTimeout(
                 this.handleClose,
                 Math.max(this.timeout - hideAnimationDuration, 0),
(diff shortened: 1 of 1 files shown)


vicgeralds:
This looks quite good, but I'm going to close it since I'm already working on a fix.

adrianschmidt:
Maybe if we merge it instead of closing it, the Latta will help us with more bugs?

adrianschmidt:
I'm gonna merge it into a branch at least 😊

adrianschmidt (approved):

src/components/snackbar/snackbar.tsx
adrianschmidt, line 1: Does @lattaai1 read this?

The commit message should follow the conventional commits spec.

src/components/snackbar/snackbar.tsx
```diff
@@ -67,9 +67,9 @@ export class Snackbar {
 
     /**
      * The amount of time in milliseconds to show the snackbar.
-     * If set to `null`, the snackbar will be persistent.
+     * If set to `null` or any value less than or equal to 0, the snackbar will be persistent.
      * This means:
-     * - either the end user will need to close is manually,
+     * - either the end user will need to close it manually,
```
adrianschmidt, line 72: @lattaai1 We use atomic commits, so this typo fix should be placed in its own commit.

adrianschmidt (approved):
@lattaai1 I had two comments, but I'm going to approve and merge this to a branch so I can fix those issues for you this time. I'd love it if you were to fix some more bugs in our repo though! Great work!

src/components/snackbar/snackbar.tsx
lattaai1: Latta reads this. We are happy to see we helped you. Let´s do this - we will give you Latta for some your repos and you will help us to make that better via feedbacks. Are you in?

Issue #3278, opened 2024-10-31
Snackbar never shown when timeout is -1


vicgeralds:
## Current behavior

If limel-snackbar is rendered with `timeout` set to -1, it becomes hidden immediately.

## Expected behavior

Even if not documented, this used to mean disabling the timeout https://www.npmjs.com/package/@material/snackbar#javascript-api

Lime CRM currently relies on this behaviour when notifying "Failed to save object" error.

## Environment

- **lime-elements** version: 37.64.2

lattaai1:
Hello, I've fixed this issue!

This is what I've done:
I have identified the issue in the `limel-snackbar` component where setting the `timeout` to -1 causes the snackbar to hide immediately. To fix this, I've updated the `handleOpen` method in the `Snackbar` class to treat negative timeout values as persistent (no timeout). Here's a summary of the changes:

1. Modified the `handleOpen` method to check if the timeout is greater than 0 before setting the timeout.
2. Updated the JSDoc comment for the `timeout` property to clarify the behavior with different values.

These changes ensure that setting `timeout` to -1 (or any negative value) will make the snackbar persistent, matching the expected behavior described in the problem statement.

I also created pull request: https://github.com/Lundalogik/lime-elements/pull/3279


> [!CAUTION]  
> Disclaimer: This fix was created by Latta AI and you should never merge before you check the correctness of generated code!

---
This bug was fixed for free by Latta AI - https://latta.ai/ourmission

PR #3280, opened 2024-11-01
fix(snackbar): treat timeout -1 as meaning no timeout

vicgeralds:
Handles `timeout` set to -1 as meaning no timeout (snackbar becomes persistent). This restores previous behaviour.

fix: #3278

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/snackbar/examples/snackbar-persistent.tsx
```diff
@@ -31,7 +31,7 @@ export class SnackbarPersistentExample {
                 message="We use cookies to deliver services and analyze traffic."
                 actionText="Okay, I got it!"
                 open={this.isOpen}
-                timeout={null}
+                timeout={-1}
```
mjlime, line 34: The comment above doesn't match this value, it mentions `null` as before this change, but nothing about `-1`.

mjlime: Does `null` and `-1` yield the same result?

vicgeralds: Yes, both work. I didn't see the comment mentioning `null` 😊

vicgeralds: c4f0b4e41168c884ed1f92382bb83fb863ae0557

Kiarokh: @jgroth had some opinions regarding `null` vs `-1´. I don't remember

jgroth: I did? :sweat_smile: I don't remember us talking about `-1` :sweat_smile:

Kiarokh: yes you did 😄 I don't exactly remember when, but perhaps it was when we were working on this issue https://github.com/Lundalogik/lime-elements/pull/2403/ or we were working on something about Snackbars; I remember that you thought it was more correct to have the timeout on `null`, instead of MDC's default value for getting a persistent timeout which was `-1`. Since it wasn't something that I personally had any opinions about, I forgot the discussion or your reasons… But I remember we were talking about it 😅

mjlime (approved):
If no one else disapproves 😁

Kiarokh:
was this a breaking change only for you? couldn't you use `null` where it was breaking 😅 ?

vicgeralds:
> was this a breaking change only for you? couldn't you use null where it was breaking 😅 ?

I don't know. It's a breaking change for lime-crm-components notifying that saving a limeobject fails. Changing that to `timeout: null` either means ignoring strict null checks or updating the notification service api.

Also, limel-snackbar still strictly doesn't accept `null` as `timeout`. So if someone wants that change from `-1` to `null` to happen, they should present an argument why it's better (and I still don't see any reason to break compatibility).

I see passing `null` as potentially confusing, as it can be mistaken for meaning the same thing as passing `undefined`. While it might not be obvious what `timeout={-1}` means, at least it looks like an intentionally set value.

Kiarokh:
OK. So hopefully this doesn't break things that might have been done after the previous change :blush

adrianschmidt:
> I see passing `null` as potentially confusing, as it can be mistaken for meaning the same thing as passing `undefined`. While it might not be obvious what `timeout={-1}` means, at least it looks like an intentionally set value.

I agree with this argument. I'd say `timeout=-1` (or `timeout=0` for that matter) is a pretty clear indication that there is no timeout.

> OK. So hopefully this doesn't break things that might have been done after the previous change 😊

I looked at the earlier change—which consisted of [updating the docs](https://github.com/Lundalogik/lime-elements/pull/2403/files#diff-19fd7d69caf31c59fd3ad08c6a1e2adfc03a13cbc8bef86e38a407de743828a8R70) and [adding an if](https://github.com/Lundalogik/lime-elements/pull/2403/files#diff-19fd7d69caf31c59fd3ad08c6a1e2adfc03a13cbc8bef86e38a407de743828a8R294)—and that doesn't really look to me like the type-checking would allow `null`. If it does, I guess it's because of the `?` designating the property as optional. Regardless, the changes in this commit look like they are backwards compatible with anyone passing in `null`.

Issue #3285, opened 2024-11-05
Snackbar shows tooltip behind it


vicgeralds:
## Current behavior

<img width="344" alt="image" src="https://github.com/user-attachments/assets/402d4c29-3808-43db-9543-2e07f10cfd6f">

## Expected behavior

The tooltip should be shown above the snackbar.

## Environment

- **lime-elements** version: 37.65.1

PR #3286, opened 2024-11-05
fix(snackbar): show tooltip correctly

vicgeralds:
Fixes tooltip that was partially hidden behind the snackbar popover by finding a better portal parent than `document.body`. The snackbar also needed to include the portal container styling for this to work.

Removes the `parent` prop of the portal component, since it's private and wasn't used anywhere as far as I could see. Also, it wasn't possible to use if the parent element was in a shadow dom (without including portal container styling).

fix: #3285

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/portal/portal.tsx
```diff
@@ -406,4 +393,21 @@ export class Portal {
...
+            const parent = element.closest('.limel-portal--parent');
+            if (parent) {
+                return parent;
+            }
+
+            element = (element.getRootNode() as ShadowRoot).host;
+        }
+
+        return document.body;
+    }
```
adrianschmidt, line 412: Neat! 🙌

hannahu:
How can we verify that this works? Tests? How to test it?

vicgeralds:
> Tests? 

I'm not in the mood for writing e2e tests for this.

To me, it's enough that it's trivial to reproduce the error, and it's trivial to manually test that tooltips still work in lots of places.

If a bug was introduced, it's likely dependent on race conditions that are difficult to account for in a test. I would also guess it's then caused by a problem that already exists in limel-portal under specific conditions (since it does tricky dom manipulations).

vicgeralds:
After fix:

https://github.com/user-attachments/assets/1c21a989-56ef-4623-9098-96767ce26f21

LucyChyzhova:
I tested in web client - works as expected! 
Screenshot 2024-11-06 at 14 25 13

jgroth (approved):

Issue #3288, opened 2024-11-07
When adding new item section is collapsed


vicgeralds:
## Current behavior

https://github.com/user-attachments/assets/fef41f1b-3c87-45de-a35c-a68469ab7484

## Expected behavior

https://github.com/user-attachments/assets/b09b69a2-0f5b-4d54-a6ef-a797ea69b8a8

Looks like this was caused by https://github.com/Lundalogik/lime-elements/pull/3183

PR #3289, opened 2024-11-07
fix(form): set collapsible item initial state correctly

vicgeralds:
Sets the initial state `isOpen` of a collapsible item in a form correctly.

State shouldn't be updated in the constructor of a React component using `this.setState`. It should just be assigned an initial value.

fix: #3288

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


FredrikWallstrom (approved):
LGTM!

I can't talk about whether this impacts anything else, such as performance that was addressed in the earlier commit that caused this issue.

PR #3294, opened 2024-11-08
input-field: do not close completions dropdown when onscreen keyboard opens

adrianschmidt:
fix: https://github.com/Lundalogik/crm-feature/issues/4408

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu-surface/menu-surface.tsx
+++ b/src/components/menu-surface/menu-surface.tsx
@@ -90,9 +90,6 @@ export class MenuSurface {
             capture: true,
         });
         this.host.addEventListener('keydown', this.handleKeyDown);
-        window.addEventListener('resize', this.handleResize, {
-            passive: true,
-        });
     };
 
     private teardown = () => {
@@ -101,7 +98,6 @@ export class MenuSurface {
             capture: true,
         });
         this.host.removeEventListener('keydown', this.handleKeyDown);
-        window.removeEventListener('resize', this.handleResize);
     };
 
     private handleDocumentClick = (event) => {
@@ -127,12 +123,6 @@ export class MenuSurface {
 
         this.dismiss.emit();
         this.preventClickEventPropagation();
-    };
-
-    private handleResize = () => {
-        if (this.open) {
-            this.dismiss.emit();
-        }
     };
 
     private preventClickEventPropagation = () => {
--- a/src/components/input-field/input-field.tsx
+++ b/src/components/input-field/input-field.tsx
@@ -37,7 +37,8 @@ interface LinkProperties {
     target?: string;
 }
 
-const DEBOUNCE_TIMEOUT = 300;
+const CHANGE_EVENT_DEBOUNCE_TIMEOUT = 300;
+const RESIZE_HANDLER_DEBOUNCE_TIMEOUT = 100;
 
 /**
  * @exampleComponent limel-example-input-field-text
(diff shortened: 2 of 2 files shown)


src/components/menu-surface/menu-surface.tsx
```diff
@@ -129,12 +125,6 @@ export class MenuSurface {
         this.preventClickEventPropagation();
     };
 
-    private handleResize = () => {
-        if (this.open) {
-            this.dismiss.emit();
```
adrianschmidt, line 134: This change means the components using `limel-menu-surface` can no longer rely on the menu being closed and therefore being restyled when the user opens the menu again. Instead, the components have to restyle their menus themselves.

This needs to be checked and potentially fixed in:

- [x] limel-input-field
- [ ] limel-menu
- [ ] limel-picker
- [ ] limel-select

adrianschmidt: I've kludged a fix in `limel-input-field`, which closes the menu and then reopens it, to make the menu restyle itself.

`limel-menu` just worked, with no fixes needed.

I tried the same fix I used for `limel-input-field` on `limel-picker`, but that didn't work for some reason. The code that controls whether the menu should be shown is a bit more complex in the picker than in the text-field. I had a look at `limel-select`, but that was even more complex, using a `FunctionalComponent` and stuff… 😖 

If you make the window significantly smaller while having a dropdown open, it looks pretty bad, but you just need to close and open the menu again, and it's the correct size. The bug where the dropdown closes whenever the on-screen keyboard is opened, on the other hand, makes the component very difficult to use. In some cases, the component has also automatically blurred, closing the on-screen keyboard again. In that case, the component actually becomes completely useless.

@Kiarokh I don't think I'll find a fix for the picker and select today, and I think we should prioritise getting the fix for input-field in, since we have an SOS waiting for it. What do you say, can we live with the non-resizing dropdowns for `limel-picker` and `limel-select` for a while?

Kiarokh: I think I'm too tired right now 😅 , and I'm not sure if I follow everything. I wish you could screen-share and explain. Reading this comment and the comment on the issue rings also another bell for me: https://github.com/Lundalogik/crm-feature/issues/4170

Also, I'm a bit confused. On phones, aren't we using native dropdown menus for `limel-select`? :/ 

I'll read again next week, or better if we have a call, if you need my input on the matter. Otherwise, I think what you wrote at the end sounds like the right thing to do now

adrianschmidt: Yes, let's huddle about it. I'm also tired now, so let's do it on Monday! 👍

> On phones, aren't we using native dropdown menus for limel-select? :/

Yes, the bug that this PR fixeds isn't in limel-select. It's in limel-input-field. But the fix includes a change in `limel-menu-surface`, which is also used by these other components.

Don't worry, I'll show you on Monday! 🤗

Kiarokh: Le me know when you have time ;)

Kiarokh (approved):

PR #3300, opened 2024-11-14
chore(text editor): bump prosemirror-model to 1.22.1

john-traas:
As per this [post here](https://discuss.prosemirror.net/t/heads-up-xss-risk-in-domserializer/6572) we should update to prosemirror-model 1.22.1 as it contains a fix for a vulnerability to xss attacks in the DOMSerialiser.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -54,6 +54,7 @@
         "prettier": "^3.3.3",
         "prosemirror-example-setup": "^1.2.3",
         "prosemirror-markdown": "^1.13.1",
+        "prosemirror-model": ">=1.22.1",
         "prosemirror-schema-basic": "^1.2.3",
         "puppeteer": "^19.11.1",
         "react": "^18.3.1",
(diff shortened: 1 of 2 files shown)


adrianschmidt (dismissed):
Seems reasonable.

I just did a very quick smoke test, making sure it was possible to write stuff in the basic example, and that we got the updated value output.

adrianschmidt (changes requested):
This should be a `fix`, so it creates a new release, and it's good if the commit message says what the benefit to the consumer is. Something like:

```
fix(text-editor): update prosemirror-model to 1.22.1 to fix XSS vulnerability
```

john-traas:
> This should be a `fix`, so it creates a new release, and it's good if the commit message says what the benefit to the consumer is. Something like:
> 
> ```
> fix(text-editor): update prosemirror-model to 1.22.1 to fix XSS vulnerability
> ```

⚡ Updated

adrianschmidt (approved):

PR #3304, opened 2024-11-18
Delegate the focus inside the text editor

john-traas:
Fixes: https://github.com/Lundalogik/crm-feature/issues/4476

This uses ref forwarding to override the default focus method allowing a consumer to set focus on the prosemirror editor view.

### DIFF
--- a/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
+++ b/src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx
@@ -3,6 +3,7 @@ import {
     Element,
     Event,
     EventEmitter,
+    Host,
     Prop,
     State,
     Watch,
@@ -205,18 +206,20 @@ export class ProsemirrorAdapter {
     }
 
     public render() {
-        return [
-            <div id="editor" />,
-            <div class="toolbar">
-                <limel-action-bar
-                    ref={(el) => (this.actionBarElement = el)}
-                    accessibleLabel="Toolbar"
-                    actions={this.actionBarItems}
-                    onItemSelected={this.handleActionBarItem}
-                />
-            </div>,
-            this.renderLinkMenu(),
-        ];
+        return (
+            <Host onFocus={this.handleFocus}>
+                <div id="editor" />,
+                <div class="toolbar">
+                    <limel-action-bar
+                        ref={(el) => (this.actionBarElement = el)}
+                        accessibleLabel="Toolbar"
+                        actions={this.actionBarItems}
+                        onItemSelected={this.handleActionBarItem}
+                    />
+                </div>
+                {this.renderLinkMenu()}
+            </Host>
+        );
     }
 
     renderLinkMenu() {
@@ -452,9 +455,9 @@ export class ProsemirrorAdapter {
         this.link = event.detail;
     };
 
-    public setFocus() {
+    private handleFocus = () => {
         this.view?.focus();
-    }
+    };
 
     private handleNewLinkSelection = (text: string, href: string) => {
         this.link.text = text;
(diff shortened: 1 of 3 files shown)


civing (commented):
I think it's better to let the consumer decide when to set focus and how. I'm not sure we always want to set the focus back to the text editor when we have a `stopTrigger` event. It should be up to the consumer

john-traas:
> I think it's better to let the consumer decide when to set focus and how. I'm not sure we always want to set the focus back to the text editor when we have a `stopTrigger` event. It should be up to the consumer

Should we instead have a specific `setFocus` event that the text editor can listen for?

civing:
> > I think it's better to let the consumer decide when to set focus and how. I'm not sure we always want to set the focus back to the text editor when we have a `stopTrigger` event. It should be up to the consumer
> 
> Should we instead have a specific `setFocus` event that the text editor can listen for?

Does it work to override the focus() method in `limel-prosemirror-adatpor` and call `this.view.focus()` ?

john-traas:
> > > I think it's better to let the consumer decide when to set focus and how. I'm not sure we always want to set the focus back to the text editor when we have a `stopTrigger` event. It should be up to the consumer
> > 
> > 
> > Should we instead have a specific `setFocus` event that the text editor can listen for?
> 
> Does it work to override the focus() method in `limel-prosemirror-adatpor` and call `this.view.focus()` ?

The setFocus function on the `limel-prosemirror-adapter` does that:

```
public setFocus() {
    this.view?.focus();
}
```

I could try ref forwarding from the `limebb-text-editor` -> `limel-text-editor` -> `prosemirror-adapter` to call the `setFocus` function from the consumer.

civing:
> > > > I think it's better to let the consumer decide when to set focus and how. I'm not sure we always want to set the focus back to the text editor when we have a `stopTrigger` event. It should be up to the consumer
> > > 
> > > 
> > > Should we instead have a specific `setFocus` event that the text editor can listen for?
> > 
> > 
> > Does it work to override the focus() method in `limel-prosemirror-adatpor` and call `this.view.focus()` ?
> 
> The setFocus function on the `limel-prosemirror-adapter` does that:
> 
> ```
> public setFocus() {
>     this.view?.focus();
> }
> ```
> 
> I could try ref forwarding from the `limebb-text-editor` -> `limel-text-editor` -> `prosemirror-adapter` to call the `setFocus` function from the consumer.

I tried to override the public focus method and it works great. I'm not sure if this is the most canonical way of doing it though. But it is simple and works 😄

https://github.com/user-attachments/assets/ebf95125-8875-4971-9eae-cf86253feff0

### limel-text-editor
´´´ts
    connectedCallback() {
        this.host.focus = this.focus.bind(this);
    }
    
...

        return [
            <limel-prosemirror-adapter
                ref={(el) =>
                    (this.prosemirrorAdapter =
                        el as HTMLLimelProsemirrorAdapterElement)
                }
                aria-placeholder={this.placeholder}
                contentType={this.contentType}
                onChange={this.handleChange}
                customElements={this.customElements}
                value={this.value}
                aria-controls={this.helperTextId}
                id={this.editorId}
                tabindex={this.disabled ? -1 : undefined}
                aria-disabled={this.disabled}
                language={this.language}
                triggerCharacters={this.triggers}
            />,
            this.renderPlaceholder(),
            this.renderHelperLine(),
        ];
´´´

### limel-prosemirror-adatpor
´´´ts
    public connectedCallback() {
        if (this.view) {
            this.initializeTextEditor();
        }

        this.host.addEventListener(
            'open-editor-link-menu',
            this.handleOpenLinkMenu,
        );
        this.host.focus = this.focus.bind(this);
    }

    public focus() {
        console.log('Focus is great!');
        this.view?.focus();
    }
´´´

john-traas:
@civing I've dropped the stopTrigger function completely for now - it's easy to implement if we need it and just added ref forwarding so we can access the focus function on the `prosemirror-adapter`. 

I'm going to add a small PR to `lime-crm-building-block`s implementing this in the `limebb-text-editor`.

civing (approved):

PR #3308, opened 2024-11-20
feat(card): add new component

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -198,6 +198,17 @@ export namespace Components {
         "type"?: CalloutType;
     }
     // @beta
+    export interface LimelCard {
+        "actions"?: Array<ActionBarItem | ListSeparator>;
+        "clickable": boolean;
+        "heading"?: string;
+        "icon"?: string | Icon;
+        "image"?: Image_2;
+        "orientation": 'landscape' | 'portrait';
+        "subheading"?: string;
+        "value"?: string;
+    }
+    // @beta
     export interface LimelChart {
         "accessibleItemsLabel"?: string;
         "accessibleLabel"?: string;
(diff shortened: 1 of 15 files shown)


jgroth (dismissed):
Interfaces approved!

etc/lime-elements.api.md
adrianschmidt, line 1: I'm fixing the errors…

adrianschmidt (changes requested):

src/components/card/card.tsx
```diff
@@ -0,0 +1,192 @@
...
+        }
+    };
+
+    private renderActionBar() {
+        if (!this.actions.length) {
+            return;
+        }
+
+        return (
+            <limel-action-bar
```
Kiarokh, line 186: I must say that I'm not super happy about using `limel-action-bar` for rendering actions, as in most scenarios in card components, there is a "Primary" action, and actions are more highlighted. 
1. We don't have a way to make one of the actions primary in the action bar. 
2. The primary action is always at the right side, however, it might be so that it ends up in the overflow menu, when the card becomes too small. That also, I don't know how to handle yet. 

These are not easy to fix, without overcomplicating the APIs of `limel-action-bar`.

But I think for now, this can work. For future, we should probably make a custom actions sections inside the `limel-card`, instead.

src/components/card/card.tsx
```diff
@@ -0,0 +1,192 @@
+import { Component, h, Prop, Event, EventEmitter } from '@stencil/core';
+import { Image } from '../../global/shared-types/image.types';
+import { Icon } from '../../global/shared-types/icon.types';
+import { ActionBarItem, ListSeparator } from 'src/interface';
+import { isItem } from '../action-bar/isItem';
```
Kiarokh, line 5: I'm not sure if we need this

adrianschmidt (approved):

PR #3311, opened 2024-11-21
🤩fy card & info-tile

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [x] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/card/card.tsx
+++ b/src/components/card/card.tsx
@@ -1,10 +1,19 @@
-import { Component, h, Prop, Event, EventEmitter } from '@stencil/core';
+import {
+    Component,
+    h,
+    Prop,
+    Event,
+    EventEmitter,
+    Element,
+    Host,
+} from '@stencil/core';
 import { Image } from '../../global/shared-types/image.types';
 import { Icon } from '../../global/shared-types/icon.types';
 import { isItem } from '../action-bar/isItem';
 import { getIconName } from '../icon/get-icon-props';
 import { ListSeparator } from '../../global/shared-types/separator.types';
 import { ActionBarItem } from '../action-bar/action-bar.types';
+import { getMouseEventHandlers } from '../../util/3d-tilt-hover-effect';
 
 /**
  * Card is a component that displays content about a single topic,
@@ -87,17 +96,37 @@ export class Card {
     @Event()
     public actionSelected: EventEmitter<ActionBarItem>;
 
+    @Element()
+    private host: HTMLElement;
+
+    private handleMouseEnter: () => void;
+    private handleMouseLeave: () => void;
+
+    public componentWillLoad() {
+        const { handleMouseEnter, handleMouseLeave } = getMouseEventHandlers(
+            this.host,
+        );
+        this.handleMouseEnter = handleMouseEnter;
+        this.handleMouseLeave = handleMouseLeave;
+    }
+
     public render() {
         return (
-            <section tabindex={this.clickable ? 0 : ''}>
-                {this.renderImage()}
-                <div class="body">
-                    {this.renderHeader()}
-                    {this.renderSlot()}
-                    {this.renderValue()}
-                    {this.renderActionBar()}
-                </div>
-            </section>
+            <Host
+                onMouseEnter={this.handleMouseEnter}
+                onMouseLeave={this.handleMouseLeave}
+            >
+                <section tabindex={this.clickable ? 0 : ''}>
+                    {this.renderImage()}
+                    <div class="body">
+                        {this.renderHeader()}
+                        {this.renderSlot()}
+                        {this.renderValue()}
+                        {this.renderActionBar()}
+                    </div>
+                    <div class="limel-3d-hover-effect-glow" />
+                </section>
+            </Host>
         );
     }
 
--- a/src/components/info-tile/info-tile.tsx
+++ b/src/components/info-tile/info-tile.tsx
@@ -1,6 +1,7 @@
-import { Component, Prop, h } from '@stencil/core';
+import { Component, Prop, h, Element, Host } from '@stencil/core';
 import { InfoTileProgress } from '../info-tile/info-tile.types';
 import { Link } from '../../global/shared-types/link.types';
+import { getMouseEventHandlers } from '../../util/3d-tilt-hover-effect';
 
 /**
  * This component can be used on places such as a start page or a dashboard.
@@ -101,6 +102,20 @@ export class InfoTile {
     @Prop()
     public progress?: InfoTileProgress;
 
+    @Element()
+    private host: HTMLElement;
+
+    private handleMouseEnter: () => void;
+    private handleMouseLeave: () => void;
+
+    public componentWillLoad() {
+        const { handleMouseEnter, handleMouseLeave } = getMouseEventHandlers(
+            this.host,
+        );
+        this.handleMouseEnter = handleMouseEnter;
+        this.handleMouseLeave = handleMouseLeave;
+    }
+
     public render() {
         const extendedAriaLabel =
             this.checkProps(this?.prefix) +
@@ -116,36 +131,43 @@ export class InfoTile {
 
         const link = !this.disabled ? this.link?.href : '#';
 
-        return [
-            <a
-                title={this.link?.title}
-                href={link}
-                target={this.link?.target}
-                tabindex="0"
-                aria-label={extendedAriaLabel}
-                aria-disabled={this.disabled}
-                aria-busy={this.loading ? 'true' : 'false'}
-                aria-live="polite"
-                class={{
-                    'is-clickable': !!this.link?.href && !this.disabled,
-                    'has-circular-progress':
-                        !!this.progress?.value || this.progress?.value === 0,
-                }}
+        return (
+            <Host
+                onMouseEnter={this.handleMouseEnter}
+                onMouseLeave={this.handleMouseLeave}
             >
-                {this.renderIcon()}
-                {this.renderProgress()}
-                <div class="value-group">
-                    {this.renderPrefix()}
-                    <div class="value-and-suffix">
-                        {this.renderValue()}
-                        {this.renderSuffix()}
+                <a
+                    title={this.link?.title}
+                    href={link}
+                    target={this.link?.target}
+                    tabindex="0"
+                    aria-label={extendedAriaLabel}
+                    aria-disabled={this.disabled}
+                    aria-busy={this.loading ? 'true' : 'false'}
+                    aria-live="polite"
+                    class={{
+                        'is-clickable': !!this.link?.href && !this.disabled,
+                        'has-circular-progress':
+                            !!this.progress?.value ||
+                            this.progress?.value === 0,
+                    }}
+                >
+                    {this.renderIcon()}
+                    {this.renderProgress()}
+                    <div class="value-group">
+                        {this.renderPrefix()}
+                        <div class="value-and-suffix">
+                            {this.renderValue()}
+                            {this.renderSuffix()}
+                        </div>
+                        {this.renderSpinner()}
                     </div>
-                    {this.renderSpinner()}
-                </div>
-                {this.renderLabel()}
-            </a>,
-            this.renderNotification(),
-        ];
+                    {this.renderLabel()}
+                </a>
+                {this.renderNotification()}
+                <div class="limel-3d-hover-effect-glow" />
+            </Host>
+        );
     }
 
     private checkProps(propValue) {
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -456,3 +456,117 @@ $clickable-normal-state-transitions: (
     clip-path: inset(50%);
     white-space: nowrap;
 }
+
+// This mixin designed to enhance the visual effects,
+// when the `tiltFollowingTheCursor` utility function from `3d-tilt-hover-effect.ts`
+// is implemented in a component.
+// This adds styles to a `<div class"limel-3d-hover-effect-glow" />`, needed to create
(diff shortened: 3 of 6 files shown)


src/components/card/card.tsx
```diff
@@ -87,16 +99,30 @@ export class Card {
...
     public actionSelected: EventEmitter<ActionBarItem>;
 
+    @Element() private element: HTMLElement;
+    private the3dElementBounds: DOMRect;
+
+    public componentDidLoad() {
+        const the3dElement = this.element.shadowRoot.querySelector('section');
+        the3dElement.addEventListener('mouseenter', this.handleMouseEnter);
+        the3dElement.addEventListener('mouseleave', this.handleMouseLeave);
+    }
```
adrianschmidt, line 109: Hmm… I wonder if these event listeners need to be removed manually, or if they are removed automatically when the component is destroyed because they are applied to an internal element… 🤔

We need to not leak memory. The web client leaks _way_ too much memory already. I don't feel I can take the time to look that up right now though.

src/style/mixins.scss
adrianschmidt, line 1: Very nice effect! But, please make sure we only apply it when `prefers-reduced-motion` is not set 🙏

adrianschmidt (changes requested):
Must respect `prefers-reduced-motion`.

src/components/card/card.tsx
Kiarokh: > Hmm… I wonder if these event listeners need to be removed manually, or if they are removed automatically when the component is destroyed because they are applied to an internal element… 🤔

The overlord said:
> To ensure that there are no memory leaks, it's a good practice to manually remove event listeners when the component is destroyed. You can achieve this by adding a `disconnectedCallback` method to your component and removing the event listeners there.

So I added a fixup

src/style/mixins.scss
Kiarokh: I actually don't believe that is what they mean by `prefers-reduced-motion`. It's for _reducing_ motions, and crazy animations, not _removing_ them totally. 

This animation is not really that intrusive, and is already quite soft and decent. However, I can check if I can do anything about it.

Kiarokh: ⚡

adrianschmidt: > The prefers-reduced-motion [CSS](https://developer.mozilla.org/en-US/docs/Web/CSS) [media feature](https://developer.mozilla.org/en-US/docs/Web/CSS/@media#media_features) is used to detect if a user has enabled a setting on their device to minimize the amount of non-essential motion. The setting is used to convey to the browser on the device that the user prefers an interface that removes, reduces, or replaces motion-based animations.
>
> Such animations can trigger discomfort for those with [vestibular motion disorders](https://www.a11yproject.com/posts/understanding-vestibular-disorders/). Animations such as scaling or panning large objects can be vestibular motion triggers.

https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion

This animation is basically as non-essential as animations get. I think we should turn it off when `prefers-reduced-motion` is set.

The [vestibular motion disorders](https://www.a11yproject.com/posts/understanding-vestibular-disorders/) link in the quote above is very good. It's a very short read, explains the problem well, and gives concise guidance.

You can enabled reduced motion in macOS here:

> In macOS: System Preferences > Accessibility > Display > Reduce motion

adrianschmidt: I tested the current version just now, after writing the above quote, and the reduced motion one actually made me feel, not ill, but just ever so slightly queezy. I don't have any problems with the full animation myself, at least not on a single card[^1]. I think it's because the animation is large enough that my brain identifies it as the card being rotated in 3D space. But with the reduced version, the effect is so subtle that I think it doesn't register as a rotation, and instead I just see it as a weird distortion that I can't really place. Even though I _know_ what effect is being applied.

(I don't have any vestibular motion disorder though.)

[^1]: However, I'm a little worried about what the effect is going to be when moving my mouse pointer over a field of info-tiles like many customers have.

Kiarokh: I understand what you mean. I'm not an expert, nor can I judge how someone with such disorders will suffer from these types of motions. 

However, I _highly_ suspect that they won't suffer! If such motions would be so problematic, then the users wouldn’t be able to perform simple scroll on any web-page. Because the amount of motion in a scrolling action is way higher than these delicate and quite slow, and _user controllable_ 3D tilts that I apply here.

I don't think this will be an issue really. Or not an issue that we need to handle right here and right now. So I suggest we drop the commit with media queries.

### However…
There is a problem with how we calculate the rotation. The problem is that the larger the element is in the UI, the more it rotates, when the user moves the cursor. This sometimes creates unpleasant visual effects. Even if the intention is not creating ginormous cards or info-tiles, I want to reduce the amount of rotation smartly and dynamically, based on the visual size of the element.

I will implement some scaling factor that adjusts the rotation degree based on the element's dimensions. This way, larger elements will have a smaller rotation effect. I'll add a fixup.

Kiarokh: I also found out that there is another CSS issue with info-tiles, which contributed to your experience. I'll fix that too.

Kiarokh: ⚡ 

Also another alternative could be allowing the consumer to turn this effect off. But all further improvements I'd like to do in another PR

adrianschmidt: > I also found out that there is another CSS issue with info-tiles, which contributed to your experience.

I was looking at the effect on cards actually 😄

adrianschmidt: > Also another alternative could be allowing the consumer to turn this effect off. But all further improvements I'd like to do in another PR

Yes, I agree, both that we probably will want to let the consumer disable it, and that we can do that improvement in a later PR.

I really don't see why we wouldn't just disable the effect for `prefers-reduced-motion` though? I can't imagine there's a whole lot of users that have that enabled, but if someone _does_ have it enabled, why should we say "screw you, I'm gonna force my animations on you whether you like them or not"?

Regardless, let's remove the fixup that just makes the effect smaller. I think that's worse than not doing anything.

Kiarokh: can I squash and we go ahead? :)

adrianschmidt: For this yes, I wanna check the event-listeners thing too though.

src/components/card/card.tsx
adrianschmidt: I don't understand why we need to register these in `componentDidLoad` when we already attach them using `onMouseEnter` and `onMouseLeave` in the render function?

src/components/card/card.tsx
```diff
@@ -87,17 +96,37 @@ export class Card {
...
+            >
+                <section tabindex={this.clickable ? 0 : ''}>
+                    {this.renderImage()}
+                    <div class="body">
+                        {this.renderHeader()}
+                        {this.renderSlot()}
+                        {this.renderValue()}
+                        {this.renderActionBar()}
+                    </div>
+                    <div class="limel-3d-hover-effect-glow" />
```
adrianschmidt, line 127: I think there are more things I can clean up, and built into the helper-function. Like this element for example.

Kiarokh: I'm sure you can make this way better. Especially, if we can export this helper function to our other repos, it would sound like a good idea. But we can always improve these things further, by creating new issues; if moving forwards is more important at the moment. 

This is an internal feature right now, and styling the "glow" div (inside the consumer component) needs some CSS considerations as well. We can even turn this div into a `private` component which can carry its own styles, to avoid unnecessary documentations and usage of mixins, etc… and also to add further flexibility using props, e.g. allowing consumer to control the intensity or color of the glow, etc…

I think all of these can be done later though. At the moment, I'm not sure if we will use this effect in many other components. This effect is only useful in very particular UI components.

src/util/3d-tilt-hover-effect.ts
```diff
@@ -1,151 +1,143 @@
...
- * } from './path/to/3d-tilt-hover-effect';
+ * import { getMouseEventHandlers } from './path/to/3d-tilt-hover-effect';
  * ```
  *
- * 2. In your component, define the necessary properties:
+ * 2. **Initialize in your component**:
+ *
+ * Call `getMouseEventHandlers()` in your component to retrieve the `handleMouseEnter`
+ * and `handleMouseLeave` functions, and attach them to the target element.
+ * For example, if your interactive element is the host itself:
```
Kiarokh, line 20: > if your interactive element is the host itself:

The interactive element can be the `host`, but its parent must have these styles:
```scss
@mixin parent-of-the-3d-element {
    isolation: isolate;
    transform-style: preserve-3d;

    perspective: 1000px;
        perspective: 2000px;
    }
}
```

For the 3D effect to appear properly. For components like Card or Info tile, the only way is to use a nested element, and use it as the interactive element. Not sure if it's enough that we mention which mixins they should use, or if we should document this further.

src/components/card/card.tsx
Kiarokh: @adrianschmidt If we wanted to make `limel-3d-hover-effect-glow` its own `@private` component, where would be the correct place to add such a component? In the `util` folder? And what would be a good name for it?

adrianschmidt: All components, private or not, should go in their own folder in the `components` folder, according to Stencil's style guide. We've departed from that in some cases, but I think we can put it there. And `limel-3d-hover-effect-glow` seems like a good name to me 👍

It would be great if we could contain the styling in a component! Or at least part of the styling.

src/components/3d-hover-effect-glow/3d-hover-effect-glow.tsx
```diff
@@ -0,0 +1,24 @@
+import { Component, h, Host } from '@stencil/core';
+/**
+ * This component enhances the visual effects, when the `tiltFollowingTheCursor`
+ * utility function from `3d-tilt-hover-effect.ts` is implemented in a component.
+ *
+ * This component should be added to the HTML structure of the consumer component.
+ *
+ * Styles of this component are needed to create a glow effect on the
+ * 3D element within the when the parent element is hovered.
+ * when the `tiltFollowingTheCursor` utility function from `3d-tilt-hover-effect.ts`
```
adrianschmidt, line 10: This line seems to have been left from an earlier iteration? Or something else is wrong. It doesn't quite fit in, at any rate 🙂

src/components/3d-hover-effect-glow/3d-hover-effect-glow.scss
```diff
@@ -0,0 +1,21 @@
+:host(limel-3d-hover-effect-glow) {
+    transition: background, opacity;
+    transition-timing-function: ease-out; // Must be the same as the 3D Element itself
+    transition-duration: 0.8s; // Must be the same as the 3D Element itself
+
+    display: block;
+    pointer-events: none;
```
adrianschmidt, line 7: Ah, yeah, of course… this means that `:hover` will never be applied to this element… such a _tiny_ little annoyance that the opacity change on `:hover` can't be styled in here… 😑

But this is still much better than having to add both a div and a mixin 👍

src/components/3d-hover-effect-glow/3d-hover-effect-glow.tsx
Kiarokh: perhaps this could be better formulated. I was trying to say if you want the glow effect, you're gonna need this component (and the styles that it carries) kind of 😓

adrianschmidt: Ah, maybe all that's wrong with it is that the period is at the end of the second line instead of at the end of the third?

Kiarokh: Feel free to update it as you wish bro. My head is everywhere now, and I am trying to catch up with a deadline. 🙈

adrianschmidt:
I realised I've been wanting to gold plate this way too much, and haven't managed to pull it off in a reasonable time. So I've pushed the not-yet-finished gold plated version on the branch `card-3d-gold-plate`, and reverted this to the state I had saved on `card-3d-bak`, and then squashed the `chore(util): complete the tiltFollowingTheCursor utility function` commit into the `chore(util): add a utility function to create an interactive 3D effect` commit.

Let's get this merged! 😊

adrianschmidt (approved):

src/components/info-tile/info-tile.tsx
```diff
@@ -101,6 +102,20 @@ export class InfoTile {
...
 
+    @Element()
+    private host: HTMLElement;
+
+    private handleMouseEnter: () => void;
+    private handleMouseLeave: () => void;
+
+    public componentWillLoad() {
+        const { handleMouseEnter, handleMouseLeave } = getMouseEventHandlers(
+            this.host,
```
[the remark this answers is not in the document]
Kiarokh: 🙏

Kiarokh:
> I realised I've been wanting to gold plate this way too much, and haven't managed to pull it off in a reasonable time. So I've pushed the not-yet-finished gold plated version on the branch `card-3d-gold-plate`, and reverted this to the state I had saved on `card-3d-bak`, and then squashed the `chore(util): complete the tiltFollowingTheCursor utility function` commit into the `chore(util): add a utility function to create an interactive 3D effect` commit.
> 
> Let's get this merged! 😊

Fair enough! 🙏 

I made a new PR https://github.com/Lundalogik/lime-elements/pull/3330 from a part of that `card-3d-gold-plate` to simplify this process even more, and make that other PR a bit smaller

PR #3317, opened 2024-11-25
docs(text-editor): use limel-popover instead of limel-portal

FredrikWallstrom:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/text-editor/examples/text-editor-custom-triggers.tsx
```diff
@@ -163,9 +127,7 @@ export class TextEditorCustomTriggersExample {
         }
 
         if (event.key === ESCAPE) {
-            this.close.emit();
             this.isPickerOpen = false;
-            // @TODO: close the trigger session
```
FredrikWallstrom, line 168: What do we mean with this TODO?

civing: I've fixed it. We should call the this.triggerFunction?.stopTrigger() so that we actually terminate the trigger session from within lime-element

civing: Fixed it as it will come with a commit in another PR

civing (approved):
Great job! 🎉

PR #3318, opened 2024-11-25
fix(text-editor): close XSS vulnerability

adrianschmidt:
fix: Lundalogik/crm-feature#4490

There is also a temporary commit with two new examples, with XSS injected in the initial value. Those are only for testing during the review, and that commit should be removed before we merge.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/text-editor/utils/html-converter.ts
```diff
@@ -1,14 +1,27 @@
...
-            resolve(text);
-        });
+        const whitelist: CustomElementDefinition[] = this.customNodes.map(
+            (nodeConfig: CustomElementDefinition) => ({
+                tagName: nodeConfig.tagName,
+                attributes: nodeConfig.attributes,
+            }),
+        );
+
+        return markdownToHTML(text, { whitelist: whitelist });
```
civing, line 24: Why do we need the map here? Maybe I'm missing something? :)
```suggestion
        return markdownToHTML(text, { whitelist: this.customNodes });
```

civing (commented):
Seems to work from my perspective! The html content provided by `this.value` seems to be properly sanitized before assigned to `initialContentElement.innerHTML`!

Waiting for @jgroth to approve.

src/components/text-editor/utils/html-converter.ts
```diff
@@ -1,14 +1,27 @@
 import { ContentTypeConverter } from './content-type-converter';
 import { EditorView } from 'prosemirror-view';
+import { markdownToHTML } from '../../markdown/markdown-parser';
+import { CustomElementDefinition } from 'src/interface';
```
jgroth, line 4: This import might be problematic if I remember correctly, it might mess up the generated exported types. A relative import should work, e.g. `../../../interface` (I might have gotten the number of levels incorrect :smile: )

jgroth (commented):
Looks good, just a minor comment that might not be a problem

src/components/text-editor/utils/html-converter.ts
adrianschmidt: ⚡ https://github.com/Lundalogik/lime-elements/commit/5e0a2c42030cf48ee8b9477a026bf95a0f06f78a

src/components/text-editor/utils/html-converter.ts
adrianschmidt: Good catch! I just took it from `markdownConverter`, assuming it did something. But it doesn't seem to be doing anything at all 😅

civing (approved):

PR #3322, opened 2024-11-25
fix: input email icon uses hardcoded RGB coming from MDC

NourHany:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


Kiarokh:
Thank you for the PR @NourHany ! 🙏 
I've added a "fix link" to the description section of your PR, and we will review it soon.

Kiarokh (changes requested):
Great work, and quick fix on a very fresh issue! 👏 

Could you please just change the commit title to something like this:
`fix(input-field): ensure email icon displays well in dark-mode`
or similar

For the sake of our "change log", and making it easy to read and understand for the consumers of Lime Element, we formulate the commit messages in a certain way:

- we try to explain the problem in a non-technical way, describing the problem which was solved or removed, or the feature that was added
- we do not expose the internal 3rd-party libraries that might be used, for example in this case MDC
- also, at least for the `fix` and `feat` commits, the scope is the name of the component, so `input-field` instead of `input`

Cheers

NourHany:
i've changed the title and commit, let me know if there is anything i can do!

adrianschmidt:
The SonarCloud Scan PR check is failing because it requires an access token from us, and when workflows run on third-party PRs they do not get access to our secrets.

PR #3330, opened 2024-11-28
Add `limel-3d-hover-effect-glow` as a `@private` internal component & use it

Kiarokh:
A `@private` component that is required for making
a 3D glow effect on some components. This
can be replaced by some internal elements and
styles, to further simplify things.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/3d-hover-effect-glow/3d-hover-effect-glow.scss
+++ b/src/components/3d-hover-effect-glow/3d-hover-effect-glow.scss
@@ -1,0 +1,21 @@
+:host(limel-3d-hover-effect-glow) {
+    transition: background, opacity;
+    transition-timing-function: ease-out; // Must be the same as the 3D Element itself
+    transition-duration: 0.8s; // Must be the same as the 3D Element itself
+
+    display: block;
+    pointer-events: none;
+
+    position: absolute;
+    inset: 0;
+
+    opacity: 0.1;
+
+    background-image: radial-gradient(
+        circle at var(--limel-3d-hover-effect-glow-position, 50% -20%),
+        rgb(var(--color-white), 0.3),
+        rgb(var(--color-white), 0)
+    );
+
+    mix-blend-mode: plus-lighter;
+}
--- a/src/components/3d-hover-effect-glow/3d-hover-effect-glow.tsx
+++ b/src/components/3d-hover-effect-glow/3d-hover-effect-glow.tsx
@@ -1,0 +1,26 @@
+import { Component } from '@stencil/core';
+/**
+ * This component enhances the visual effects, when the `tiltFollowingTheCursor`
+ * utility function from `3d-tilt-hover-effect.ts` is implemented in a component.
+ *
+ * This component should be added to the HTML structure of the consumer component.
+ *
+ * This component carries its own styles which are needed to create a glow effect on the
+ * 3D element within the parent element, when the parent is hovered.
+ *
+ * The parent element must be using the `tiltFollowingTheCursor` utility function
+ * imported from `3d-tilt-hover-effect.ts`. This function will dynamically
+ * affect parts of the styles of this 3D glow effect.
+ *
+ * @private
+ */
+@Component({
+    tag: 'limel-3d-hover-effect-glow',
+    shadow: true,
+    styleUrl: '3d-hover-effect-glow.scss',
+})
+export class HoverEffectGlowComponent {
+    public render() {
+        return null;
+    }
+}
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -136,10 +136,9 @@ limel-action-bar {
 }
 
 // The 3D effect
-@include mixins.limel-3d-hover-effect-glow(
-    section,
-    var(--card-border-radius, $default-border-radius)
-);
+limel-3d-hover-effect-glow {
+    border-radius: var(--card-border-radius, $default-border-radius);
+}
 
 :host(limel-card) {
     @include mixins.parent-of-the-3d-element;
--- a/src/components/card/card.tsx
+++ b/src/components/card/card.tsx
@@ -124,7 +124,7 @@ export class Card {
                         {this.renderValue()}
                         {this.renderActionBar()}
                     </div>
-                    <div class="limel-3d-hover-effect-glow" />
+                    <limel-3d-hover-effect-glow />
                 </section>
             </Host>
         );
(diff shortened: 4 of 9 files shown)


src/components/3d-hover-effect-glow/3d-hover-effect-glow.tsx
Kiarokh, line 1: the good thing with this component is that: 
1. we can add more CSS props for it (like the color of the highlight/glow). 
2. it can be more easily used in other repos by us, if needed. It's private, but still available.
3. together with mixins, which are also available via importing the `mixins.scss` from node modules folder, we can achieve the same effect way more easily this way

src/components/3d-hover-effect-glow/3d-hover-effect-glow.scss
```diff
@@ -0,0 +1,21 @@
...
+    transition-timing-function: ease-out; // Must be the same as the 3D Element itself
+    transition-duration: 0.8s; // Must be the same as the 3D Element itself
+
+    display: block;
+    pointer-events: none;
+
+    position: absolute;
+    inset: 0;
+
+    opacity: 0.1;
```
adrianschmidt, line 12: The problem with putting the styling here is that the component will be rendered before the styling applies. It won't be visible then, since it's unstyled, but it will have a default opacity of 1. Then the styling applies, which set's the opacity to 0.1. But there's also been a transition applied to the opacity, so there will be this little "lightning flash" when the component is loaded.

adrianschmidt: https://github.com/user-attachments/assets/dfc087da-b7d2-411a-b7f0-bea5fd338fc7

Kiarokh: 😄 On my slow machine things load in a totally different way and jump around so much until the page is completely loaded, so that I can't even see this visual effect that you screen recorded.

Bu I think I can fix it. I'll adda. fixup

Kiarokh: ### A note for future:
There are several ways of mitigating the effect that you mentioned.

The best one would be using [the `@property` rule](https://developer.mozilla.org/en-US/docs/Web/CSS/@property) to animate the `x` and `y`, and opacity values of the radial gradient. I tried it, but couldn't really make it work. I don't exactly know why. Due to lack of time, I decided to keep a copy of the code here, and do it in a simpler but dirtier way. 

What I did was like below…
In `3d-hover-effect-glow.scss` tried to `transition` the CSS custom properties, combining them with `@property` rule.

```scss
:host(limel-3d-hover-effect-glow) {
    @property --limel-3d-hover-effect-glow-opacity {
        syntax: '<number>';
        inherits: true;
        initial-value: 0.1;
    }

    @property --limel-3d-hover-effect-glow-position-x {
        syntax: '<percentage>';
        inherits: true;
        initial-value: 50%;
    }

    @property --limel-3d-hover-effect-glow-position-y {
        syntax: '<percentage>';
        inherits: true;
        initial-value: -20%;
    }

    transition-property: --limel-3d-hover-effect-glow-opacity,
        --limel-3d-hover-effect-glow-position-x,
        --limel-3d-hover-effect-glow-position-y;
    transition-timing-function: ease-out; // Must be the same as the 3D Element itself
    transition-duration: 0.8s; // Must be the same as the 3D Element itself

    display: block;
    pointer-events: none;

    position: absolute;
    inset: 0;

    background-image: radial-gradient(
        circle at var(--limel-3d-hover-effect-glow-position-x, 50%)
            var(--limel-3d-hover-effect-glow-position-y, -20%),
        rgb(var(--color-white), var(--limel-3d-hover-effect-glow-opacity, 0.1)),
        rgb(var(--color-white), 0)
    );

    mix-blend-mode: plus-lighter;
}
```

and then in the `mixins.scss`, I would do:
```scss

@mixin the-3d-element {
    
   // …

    &:hover {
        limel-3d-hover-effect-glow {
            --limel-3d-hover-effect-glow-opacity: 0.5;
                --limel-3d-hover-effect-glow-opacity: 0.2;
            }
        }
    }
}
```

and of course the JS in `3d-tilt-hover-effect.ts` was updated too:

```tsx
        const glowPositionX = `
    ${center.x * GLOW_POSITION_MULTIPLIER + the3dElementBounds.width / CENTER_DIVISOR}px
`;
        const glowPositionY = `
    ${center.y * GLOW_POSITION_MULTIPLIER + the3dElementBounds.height / CENTER_DIVISOR}px
`;
        element.style.setProperty(
            '--limel-3d-hover-effect-glow-position-x',
            glowPositionX,
        );
        element.style.setProperty(
            '--limel-3d-hover-effect-glow-position-y',
            glowPositionY,
        );
    };
```
and
```tsx
    const handleMouseLeave = () => {
        document.removeEventListener('mousemove', tiltCallback);
        element.style.removeProperty('--limel-3d-hover-effect-rotate3d');
        element.style.removeProperty('--limel-3d-hover-effect-glow-position-x');
        element.style.removeProperty('--limel-3d-hover-effect-glow-position-y');
    };

```

adrianschmidt: Yeah, I've noticed there's a bit of a jump too.

But, can we put this PR on ice until cooldown? (Or if you can convince someone appropriate that I should spend cycle time on this, then I'm happy to help out and review further 😅)

Kiarokh: Yes, all is done on this PR for now. And this can go to cooldown.

But now I don't know which cooldown board 😄, since we have now several cooldown boards 🙈

src/components/3d-hover-effect-glow/3d-hover-effect-glow.tsx
adrianschmidt: > together with mixins, which are also available via importing the mixins.scss from node modules folder

Just be aware that if you import stuff like that, and we ever change something internally in lime-elements, so that file is no longer included in the same place, all repos where such an import is made will break if they update lime-elements.

So if we really want to be able to do that, we should document where that file is supposed to be in the published package, and preferably also add some kind of test for that, so we can't accidentally break it.

adrianschmidt (approved):

PR #3332, opened 2024-12-02
Clicked item event in list

civing:
Adds a specific event for interaction in `limel-list`. With the current change event it can be hard to sort out which element was actually interacted with.

If one item is already selected and a new one is clicked then one could see which one is selected to know the interaction. But if the same selected item is clicked it's not possible to know that that element was interacted with since the events come one by one by design.

With the new interaction event it's easy to know which item was either clicked or selected by other means.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/list/list.tsx
+++ b/src/components/list/list.tsx
@@ -99,6 +99,13 @@ export class List {
     @Event()
     protected select: EventEmitter<ListItem | ListItem[]>;
 
+    /**
+     * Fires when a user interacts with an item in the list (e.g., click,
+     * keyboard select).
+     */
+    @Event()
+    interact: EventEmitter<ListItem>;
+
     public connectedCallback() {
         this.setup();
     }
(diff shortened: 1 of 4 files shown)


src/components/list/list.tsx
```diff
@@ -99,6 +99,12 @@ export class List {
     @Event()
     protected select: EventEmitter<ListItem | ListItem[]>;
 
+    /**
+     * Fires when a user interacts with an item in the list (e.g., click,
+     * keyboard select).
+     */
+    @Event() interaction: EventEmitter<ListItem>;
```
jgroth, line 106: We have a similar named event in a few other places, so for consistency we should rename this
```suggestion
    @Event() interact: EventEmitter<ListItem>;
```

Kiarokh: Also you should add the same event to `limel-menu-list`!

civing: Thanks Johan, I'll change the naming so it's more consistent throughout the package.

civing: See https://github.com/Lundalogik/lime-elements/pull/3332/commits/6a992ca2b9495958b83ba23d82a1fee84e3e4674

Kiarokh (approved):

jgroth (approved):

Issue #3338, opened 2024-12-02
List: make clicking on list items trackable by analytics apps like Heap


Kiarokh:
## New feature motivation
Consumers want to be able to track clicking on list items clickable, for example through [Heap](https://www.heap.io/). However, today, the list items do not have any identifier attribute, except a `data-index`. The problem is that this identifier is a variable, depending on the number and order of the items in the list.
<img width="300" alt="image" src="https://github.com/user-attachments/assets/388238dd-75f1-417f-9a9e-74a1fbcc4e55"> <img width="267" alt="image" src="https://github.com/user-attachments/assets/34c05485-c8dc-4e4d-b03a-4be290b955c3">
For example, here in this menu, the number and orders of items change, depending on the user role.

## New feature description
There should be something on the `<li />` that makes it identifyable.

## New feature implementation
See the PR.

PR #3339, opened 2024-12-02
fix(list): make clicking on list items trackable by analytics apps

Kiarokh:
fix https://github.com/Lundalogik/lime-elements/issues/3338

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/menu-list/menu-list-renderer.tsx
```diff
@@ -139,6 +139,8 @@ export class MenuListRenderer {
                 aria-disabled={item.disabled ? 'true' : 'false'}
                 aria-selected={item.selected ? 'true' : 'false'}
                 data-index={index}
+                data-text={item.text}
+                data-secondary-text={item.secondaryText}
```
Kiarokh, line 143: @magnusfagerlund this change reflects that `text` and `secondaryText` props of the list items on the `<li />` element, like below:
image

>[!IMPORTANT]
>These could be different based on the language. Is it OK for you and your Heap? Or do you need some static thing like `id` which is the same for each language. I should just say that if we go that way, you don't get any information  from Heap, unless we update all lists everywhere, and add some id to their items manually.
> That we could still do probably in future. But I think this should be enough for now. Right?

magnusfagerlund: It would be great if we could use `id`. It will be too much admin work in Heap to keep track of the different languages otherwise.

Kiarokh: @jgroth also raised a concern about leaking personal info, if we use the data itself. While that could be the case for anything clickable. I still think it's a valid concern, and an important thing to mitigate if we can.

Kiarokh: @adrianschmidt has some idea using some `key`s, but don't do it during the cycle if that takes too much time Adrian

magnusfagerlund: If there is a risk of leaking personal information using ´text´and ´secondaryText´, then we shouldnt do it. Then we should only use ´íd´.

jgroth: As it looks right now, we would send all the text visible in the list item straight to Heap, i.e. if you click on an item in the global search or in a relation picker, we would send personal information.

I think we need another property on the `ListItem` interface that is used specifically for this. There is no `id` property right now, so maybe that is what we need

adrianschmidt: This somewhat overlaps with the need to have a unique and non-changing `key` attribute for components that are rendered in a loop (like when we render a list by looping over an array and rendering a `limel-list-item` component for each item in the array).

I won't go into all the technical detail on why we need that here, but it's got to do with performance and not mixing up the internal state between components. (You might remember that we had a problem in Lime Admin, where the open/closed state of the collapsible-sections would not follow the item as you moved it in the list, but would instead stay in the same place. That was caused by us not having a `key` attribute on the collapsible-sections.)

The overlap is that the value of the `key` attribute also has to be tied to the actual item being rendered. It can't just be the index in the array, because if you then move around items in the array, the internal component state will stay with the index, not with the item.

I think compounding the `text` and `secondaryText` values to use for uniquely identifying the items is a very reasonable default. I don't think the lime-elements component can really be fully responsible for what could potentially end up in the wrong place, but it might have to be accommodating. It's impossible to anonymise the values of `text` and `secondaryText` without also anonymising the text we are actually showing the user, which obviously isn't an option. And I don't think we should avoid reflecting the value of properties just because those values could potentially end up in some tracking system. Filtering that out feels to me like it has to be the responsibility of the tracking system (whether that is Heap's code or some custom code of ours that we run as part of the analytics add-on).

But, at least for the sake of putting a uniquely identifiable value in the `key` property, we can certainly hash the value. That's still not super helpful for the purpose of tracking, but perhaps we can work something out for what to put in the `id` attribute too. For example, we could default to using the `text` and `secondaryText` in cleartext, but add a property to `listItem` that let's the consumer tell us we should anonymise the value (by hashing). We could also add a property that let's the consumer provide a custom value to use. Then we can write `'userProfile` there, instead of the name of the user, in the case of the user menu.

I guess we could put a `gdpr` prop on `limel-config`, and if that's set to `'strict'`, we default to hashing everything we can, in all components… 🤷‍♂️

Kiarokh: Yeah, for me this change I did in this PR now is exactly like using `reflect: true;`. Let me explain why:

I have always wanted to refactor the list items and make them a private internal component, instead of repeating the spaghetti code that we have today from MDC, in both limel-list, and limel-menu-list. If I do that refactoring, and create a new component that renders list items, what I would end up doing is using `reflect: true` for these props, just like we do any other place in Lime Elements.

We can't suddenly decide not to reflect them, since in CRM, lists can sometimes contain personally identifiable information. Right?

adrianschmidt: Yeah, I'm with you on that line.

On the other hand, the only value we have of this specific PR at this specific moment is that it becomes possible to better track clicks on list- and menu-items, right? But if we add this, we basically have to also add something in lime-webclient or the analytics package that makes sure we're not sending loads and loads of personally identifiable data to Heap. Which in turn would mean we get no benefit, right?

So I would vote for simply not adding this at the moment.

---

The reason companies take GDPR so seriously, is that the possible sanctions for infringements are _extremely_ severe.

When it comes to fines, there are two levels, depending on the severity of the infringement:

> 1. Up to €10 million, or 2% annual global turnover – whichever is higher.
> 2. Up to €20 million, or 4% annual global turnover – whichever is higher.

Note "whichever is _higher_". Lime Technologies Sweden AB had a turnover of 437,723,000 SEK in 2023. 2% of that is 8,754,460 SEK, or approximately €760,000. €10,000,000 is quite significantly higher than €760,000, _so the €10,000,000 maximum fine applies._ That's more than 25% of the company's 2023 entire _turnover_. I'm no economist, but I'm guessing that's not really something any normal company would survive. And that's for the _less_ serious level of infringements.

There are also other sanctions that can be applied, like "Imposing a temporary or permanent ban on data processing". Something like that would definitely kill the company dead.

https://www.itgovernance.eu/sv-se/dpa-and-gdpr-penalties-se

adrianschmidt: I've had a look at `list-renderer.tsx` and `menu-list-renderer.tsx`, and the individual list-items aren't actually represented by their own custom component. They're just old-school html elements. So there's no internal state that they keep track of. So the need for a `key` attribute don't actually apply in this case, as far as I understand.

I've unassigned myself from this issue. There's nothing wrong with the code, so you can consider it approved on a technical level. I feel like we very much do _not_ have approval on a gdpr level though, especially since @magnusfagerlund said the following earlier in the thread:

> If there is a risk of leaking personal information using `text` and `secondaryText`, then we shouldn't do it. Then we should only use `id`.

magnusfagerlund: Now, I'm out on thin ice so take it for what it is and don't shoot me please :)... But Lime Go are applying 
image
`data-intercom-target` in some places of the app. In this case to make it easier to user Intercom when creating Product Tours and such, but could we apply something similar?

Kiarokh: We can limit this to `limel-menu-list` and only display `data-text` (skipping `data-secondary-text`) to only cover menus and dropdowns, etc… which don't have personal info (at least in our product). 

But I think this is overkill… Let's talk about the details in a private chat instead, since this is a public repo

adrianschmidt: Perfect! I've updated the commit 👌

adrianschmidt (approved):

src/components/menu-list/menu-list-renderer.tsx
adrianschmidt: > But Lime Go are applying `data-intercom-target` in some places of the app. In this case to make it easier to user Intercom when creating Product Tours and such, but could we apply something similar?

That's basically what we are doing here, but the problem is knowing what value to put in there. Go has probably hard-coded a few values on their main menu items specifically. We could also do that for the select few places where we know exactly what the menu option is going to be, because it's not configurable. But for any other place, where configuration, user role, or anything else, can affect what options are available, we need some automatic way.

Going with only the main text, and only on menu items, not on list items, seems like a very good solution for now 👍

adrianschmidt: Ah… now I think I understood the comments about using `id`. We don't have a property called `id` on `menuItem`, but we do have a property called `value`, which I think is the one that should be used if you need to identify the menu option in a language independent way. So perhaps we should simply use that one instead?

jgroth: `value` can be anything so it's hard to use that. If an object is used it will just say `[Object object]` everywhere. You could of course implement `toString()` to get something else, but it also makes it a bit unclear that it will be used for this, and if you do it on let's say a `LimeObject` we are back at sending personal information since we have no control over what `value` is. 
Using something like `id` is probably a lot safer since it should be used to identify things :smile:

adrianschmidt: Ah, crap. Good point!

Then I think using only `text` and only for menus is the way to go for now.

adrianschmidt (approved):

PR #3351, opened 2024-12-06
fix(text-editor): send stop trigger when outside

civing:
Sends a stopTrigger event when the cursor goes outside the trigger.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


john-traas:
I will approve this one. In an upcoming PR there may be changes to the way we look for and handle active triggers, but for now this solves the current issue.

john-traas (approved):

PR #3365, opened 2024-12-12
chore(spinner): do not set `limeBranded` to `true` by default

Kiarokh:
There is lime's logo everywhere the spinner is being used. In a public library, it is better that this specific design is not default.
In Lime's products, there is no point of repeating this logo everywhere in the UI, and a better approach is to
more purposefully set it to `true`, where appropriate in the ui.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


LucyChyzhova (approved):
Finally! 😀

PR #3374, opened 2024-12-18
limel-picker: add default searcher function for simple client-side filtering

adrianschmidt:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/picker/examples/picker-empty-suggestions.tsx
+++ b/src/components/picker/examples/picker-empty-suggestions.tsx
@@ -4,7 +4,10 @@ import { Component, h, State } from '@stencil/core';
 const NETWORK_DELAY = 500;
 
 /**
- * With no suggestions and a message for empty search results
+ * With a custom search function
+ *
+ * The custom search function returns two suggestions if the query is empty.
+ * Otherwise, it filters the items based on the query.
  *
  * :::important
  * This example simulates that searching is done on the server. Because these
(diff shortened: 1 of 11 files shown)


src/components/picker/examples/picker-empty-suggestions.tsx
```diff
@@ -56,8 +59,11 @@ export class PickerExample {
             if (query === '') {
                 // Simulate some network delay
                 setTimeout(() => {
-                    resolve([]);
+                    const result = this.allItems.slice(8, 10);
+                    resolve(result);
                 }, NETWORK_DELAY);
+
+                return;
```
jgroth, line 66: We could probably improve this as well if we want to get rid of the `new Promise()` syntax, similar to the other comment I had. The simulated network delay can be done something like this
```ts
await new Promise(resolve => setTimeout(resolve, NETWORK_DELAY));
```

src/components/picker/picker.tsx
```diff
@@ -556,13 +570,44 @@ export class Picker {
...
+
+                return searchText.includes(query.toLowerCase());
+            });
+
+            return resolve(
+                filteredItems.slice(0, DEFAULT_SEARCHER_MAX_RESULTS),
+            );
+        });
+    };
+
```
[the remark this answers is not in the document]
adrianschmidt: Yeah, I took it completely from the basic example.

Maybe they had to be more complicated when we used the mocked network delay, and could have been simplified when we removed that?

Thanks for this improvement!

src/components/picker/examples/picker-empty-suggestions.tsx
adrianschmidt: ⚡ 4d4e4e46b5dd2f45eb94a7f6516a5cf21c47a363

jgroth (approved):

PR #3376, opened 2024-12-19
fix(info-tile): simplify value and loading state checks

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/4526

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/info-tile/info-tile.tsx
+++ b/src/components/info-tile/info-tile.tsx
@@ -183,8 +183,12 @@ export class InfoTile {
     private renderValue = () => {
         const characterCount = (this.value ?? '').toString().length;
 
-        if (!this.value && this.loading) {
+        if ((this.value === 'undefined' || !this.value) && this.loading) {
             return <span class="value">···</span>;
+        }
+
+        if (!this.value) {
+            return;
         }
 
         if (this.value) {


src/components/info-tile/info-tile.tsx
```diff
@@ -183,10 +183,14 @@ export class InfoTile {
     private renderValue = () => {
         const characterCount = (this.value ?? '').toString().length;
 
-        if (!this.value && this.loading) {
+        if ((this.value === 'undefined' || !this.value) && this.loading) {
```
jgroth, line 186: I don't think we should have this check. `this.value` should not get the value `undefined` in the first place so the bug is somewhere else :slightly_smiling_face:

LucyChyzhova:
I made a pr in BB repo - https://github.com/Lundalogik/lime-crm-building-blocks/pull/866 , therefore closing this one

PR #3392, opened 2025-01-08
add support for displaying pictures in the list items

Befkadu1:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -2294,6 +2294,7 @@ export interface ListItem<T = any> {
     icon?: string | Icon;
     // @deprecated
     iconColor?: string;
+    image?: Image_2;
     primaryComponent?: ListComponent;
     secondaryText?: string;
     selected?: boolean;
--- a/src/components/list/examples/list-pictures.tsx
+++ b/src/components/list/examples/list-pictures.tsx
@@ -1,0 +1,109 @@
+import { ListItem } from '@limetech/lime-elements';
+import { Component, h, State } from '@stencil/core';
+
+/**
+ * List with Pictures and Icons
+ *
+ * :::note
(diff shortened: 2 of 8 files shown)


etc/lime-elements.api.md
```diff
@@ -2294,6 +2294,7 @@ export interface ListItem<T = any> {
     icon?: string | Icon;
     // @deprecated
     iconColor?: string;
+    image?: Image_2;
```
Kiarokh, line 2297: i know this is auto generated, but it just felt weird that it says `Image_2`. Did we do something wrong, or is this correct?

Kiarokh (approved):

etc/lime-elements.api.md
Befkadu1: No idea it is auto generated running the command `npm run api:update`

adrianschmidt: I don't know. I'm not sure if I've seen that in this file before. I know I've seen it somewhere, but I'm fairly sure that's been in `components.d.ts`. In this case, I find `Image_2` referenced in `dist/lime-elements.d.ts`, but I can't seem to find where it's actually coming from.

I've tried removing the existing `dist` folder and other temporary folders, and run the build and `api:update` script again, but just got the same result anyway.

It's exported as `Image` anyway, so it shouldn't be a problem:

image
image

adrianschmidt (approved):
Changes to external interface approved 👍

PR #3395, opened 2025-01-10
feat(chart): allow making items clickable

Kiarokh:
## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chart/chart.tsx
+++ b/src/components/chart/chart.tsx
@@ -448,7 +448,7 @@ export class Chart {
         return item;
     }
 
-    private handleClick = (event: MouseEvent) => {
+    private readonly handleClick = (event: MouseEvent) => {
         const item = this.getClickableItem(event.currentTarget as HTMLElement);
         if (!item) {
             return;
@@ -458,7 +458,7 @@ export class Chart {
         this.interact.emit(item);
     };
 
-    private handleKeyDown = (event: KeyboardEvent) => {
+    private readonly handleKeyDown = (event: KeyboardEvent) => {
         if (event.key !== 'Enter' && event.key !== ' ') {
             return;
         }
--- a/src/components/chart/chart.tsx
+++ b/src/components/chart/chart.tsx
@@ -1,4 +1,4 @@
-import { Component, h, Prop, Watch } from '@stencil/core';
+import { Component, Event, EventEmitter, h, Prop, Watch } from '@stencil/core';
 import { Languages } from '../date-picker/date.types';
 import translate from '../../global/translations';
 import { createRandomString } from '../../util/random-string';
@@ -28,6 +28,7 @@ const DEFAULT_INCREMENT_SIZE = 10;
  * @exampleComponent limel-example-chart-multi-axis-with-negative-start-values
  * @exampleComponent limel-example-chart-multi-axis-area-with-negative-start-values
  * @exampleComponent limel-example-chart-axis-increment
+ * @exampleComponent limel-example-chart-clickable-items
  * @exampleComponent limel-example-chart-accessibility
  * @exampleComponent limel-example-chart-styling
  * @exampleComponent limel-example-chart-creative-styling
@@ -115,6 +116,12 @@ export class Chart {
         maxValue: number;
         totalRange: number;
     };
+
+    /**
+     * Fired when a chart item with `clickable` set to `true` is clicked
+     */
+    @Event()
+    public interact: EventEmitter<ChartItem>;
 
     public componentWillLoad() {
         this.recalculateRangeData();
@@ -222,7 +229,11 @@ export class Chart {
                     class={this.getItemClass(item)}
                     key={itemId}
                     id={itemId}
+                    data-index={index}
                     tabIndex={0}
+                    role={item.clickable ? 'button' : null}
+                    onClick={this.handleClick}
+                    onKeyDown={this.handleKeyDown}
                 >
                     <th>{this.getItemText(item)}</th>
                     <td>{this.getFormattedValue(item)}</td>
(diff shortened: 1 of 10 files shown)


src/components/chart/examples/chart-items-clickable.tsx
```diff
@@ -0,0 +1,31 @@
+import { ChartItem } from '@limetech/lime-elements';
```
adrianschmidt, line 1: This shouldn't be a `.tsx` file. `tsx` is for files that output jsx. Basically, only files that contain a Stencil `@Component` need to be `.tsx`. Otherwise it's a `.ts` file.

src/components/chart/chart.tsx
```diff
@@ -422,4 +433,38 @@ export class Chart {
...
+        }
+
+        const item = this.items[Number(index)];
+        if (!item.clickable) {
+            return;
+        }
+
+        event.stopPropagation();
+        this.itemSelected.emit(item);
+    };
```
DavidAkkerman, line 451: Would it work to define this method as 
```suggestion
    private handleClick = (item: any) => () => {
        if (!item.clickable) {
            return;
        }

        event.stopPropagation();
        this.itemSelected.emit(item);
    };
```

Then you could inject the item straight away in line 235 with `onClick={this.handleClick(item)}`. You could do the same for the keyDown event, and you don't need to search the item based on the index anymore.

src/components/chart/examples/chart-items-clickable.tsx
Kiarokh: I squashed this fixup. If there are no more comments @adrianschmidt @DavidAkkerman we can merge this now. Note that the component is still `@beta`. 

My next plan for this component is to add the "Chart Legend" feature.

src/components/chart/chart.tsx
```diff
@@ -116,6 +117,12 @@ export class Chart {
         totalRange: number;
     };
 
+    /**
+     * Fired when a feed item with `clickable` set to `true` is clicked
+     */
+    @Event()
+    public itemSelected: EventEmitter<ChartItem>;
```
adrianschmidt, line 124: "selected" implies that we are keeping track of selection state. Are we?

If not, I think we should call this something else. We use `interact` in several components already.

adrianschmidt (changes requested):

src/components/chart/chart.tsx
```diff
@@ -422,4 +433,38 @@ export class Chart {
...
+        }
+
+        const item = this.items[Number(index)];
+        if (!item.clickable) {
+            return;
+        }
+
+        if (event.key === 'Enter' || event.key === ' ') {
+            event.preventDefault();
+            this.handleClick(event as unknown as MouseEvent);
```
adrianschmidt, line 467: This forced recasting of a `KeyboardEvent` to a `MouseEvent` means that someone might later make changes in the `handleClick` function, which causes no compiler warnings, and works just fine when tested with clicks, but which throws an uncaught error bringing the app, or parts of it, down at runtime.

So I'm opposed 😝

Either break out the parts that you need into a separate function which are called by both `handleKeyDown` and `handleClick`, or update `handleClick` to accept either kind of event:

```ts
private handleClick = (event: MouseEvent | KeyboardEvent) => {
```

Although, in the latter case, it does look kind of weird that `handleClick` accepts a keyboard event, so we might consider renaming the function in that case.

src/components/chart/examples/chart-items-clickable.tsx
adrianschmidt: I haven't reviewed this yet, but I'm looking at it now.

src/components/chart/examples/chart-clickable-items.tsx
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * This provides visual and accessible feedback when the hovered, or focused,
+ * and also results in emitting an event when clicked.
+ *
+ * :::note
+ * Items in Doughnut and Pie charts support `clickable` property as well.
+ *
+ * However—due to technical reasons—, the only way to interact with their items
+ * is using the keyboard (by tabbing on them and pressing the <kbd>Enter</kbd> or
+ * <kbd>Space</kbd> keys.
```
adrianschmidt, line 18: 🤨

Can you elaborate on the technical reasons to a fellow nerd? 😄

adrianschmidt: (Just because I'm interested, not as a requirement for approval 😊)

Kiarokh: Because layers are stacked on top of each other and cover each other. Therefore only the top-most one is therefore clickable. This is however something that I have blocked with CSS. So you can't click no item in these two types.

Without using SVGs to draw each slice of the chart, or without using a very complicated logic to create a `clip-path: polygon`, I don't know any better way to solve this issue. Maybe there could be a way to measure where the cursor is and whether that region is exactly where an item was... I don't know. But I can't and don't want to solve this "technical problem". 😅

src/components/chart/chart.tsx
Kiarokh: Good point! I'll fixup!

src/components/chart/chart.tsx
Kiarokh: I swear to my imaginary entity that I had done this, but forgot to push it. 🙈

Now ⚡

adrianschmidt (approved):

PR #3398, opened 2025-01-14
React 19

vicgeralds:
- Updates to React 19 which comes with better support for rendering custom elements
- Removes `LimeElementsAdapter` that's used for rendering custom elements in rjsf forms

This is a fix, because LimeElementsAdapter set some properties later depending on their type, meaning they were set later than expected in some situations or in unexpected order.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/form/adapters/widget-adapter.ts
```diff
@@ -1,15 +1,14 @@
...
 import { getHelpComponent } from '../help';
 
-interface WidgetAdapterProps {
+export interface WidgetAdapterProps {
     name: string;
     value?: any;
     widgetProps: WidgetProps;
     extraProps?: any;
-    events?: { [key: string]: (event: any) => void };
+    events: { change: (event: CustomEvent) => void };
```
adrianschmidt, line 11: Hmm… this changes the object from allowing any number of properties (`[key: string]`), to only allowing a single property called `event`. Was that really intentional?

How about the following?

```ts
    events: {
        change: (event: CustomEvent) => void,
        [key: string]: (event: any) => void,
    };
```

This makes the object required, and it has to have the `change` property, but it can also have any number of extra properties.

Although, come to think of it, this commit doesn't change `lime-elements.api.md`, so I guess this is a purely internal change anyway?

adrianschmidt (dismissed):
This seems to work flawlessly… 😅

Nice work! 💪

src/components/form/adapters/widget-adapter.ts
vicgeralds: > Was that really intentional?

Yes. That’s how it’s actually used, and if we need to handle more events, we can add them. It’s clearer what events need to be handled and easier to implement than allowing any event.

> so I guess this is a purely internal change anyway?

Correct

adrianschmidt (approved):

Issue #3404, opened 2025-01-17
card: body text overflows out of a card with fixed a `height`


Kiarokh:
<img width="629" alt="Image" src="https://github.com/user-attachments/assets/dd5497ca-fef2-4144-9d07-ecc53a9add6f" />

Kiarokh:
related to https://github.com/Lundalogik/crm-feature/issues/1313

PR #3405, opened 2025-01-17
fix(card): make the UI more responsive

Kiarokh:
There is a good chance that the `value` which will be rendered in markdown is very large.
In such cases, the consumer may limit the `height` or `max-height` of the component, which will result in the content to overflow out of the card.
The component should have a more responsive UI
when its size is more limited by the consumer.

fix https://github.com/Lundalogik/lime-elements/issues/3404

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -55,7 +55,6 @@ section {
 }
 
 img {
-    flex-shrink: 0;
     transition: filter 0.6s ease;
     object-fit: cover;
     border-radius: calc(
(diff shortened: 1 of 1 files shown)


LucyChyzhova (approved):
👏 💫 ✨

PR #3421, opened 2025-01-30
feat(header): properly support the `Icon` interface

LucyChyzhova:
Now the consumer can specify `color` and `backgroundColor` directly on the `icon`. The styles code are however backwards compatible, and still support the old custom CSS props of `--header-icon-color` and `--header-icon-background-color`. In case any consumer has specified the CSS props, they will get the same result as before. But we removed the props from the documentations, since we want the consumers to use the `Icon` interface instead of adding styles.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/header/examples/header-colors.scss
+++ b/src/components/header/examples/header-colors.scss
@@ -1,6 +1,5 @@
 :host(limel-example-header-colors) {
     --header-background-color: rgb(var(--color-yellow-dark));
-    --header-icon-color: rgb(var(--color-white));
     --header-heading-color: rgb(var(--color-brown-default));
     --header-subheading-color: rgb(var(--color-brown-darker));
 }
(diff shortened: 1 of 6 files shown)


Kiarokh (dismissed):
I think this is the correct way to move forward, for all components, and get rid of all those silly and old CSS props. But I cannot merge this PR, it needs some blessings from @adrianschmidt @jgroth

adrianschmidt:
I'll have a look 👍

adrianschmidt:
The changes to `etc/lime-elements.api.md` completely disappeared when I ran `npm run api:update`, so I don't know what weird state you were in to get all the changes you did.

It seems like once that file has been touched in a PR, the PR requires a review from a maintainer, even if those changes are then removed from the PR. So if you want to avoid having to wait for a maintainer for no good reason, keep an eye on what you push 😊

adrianschmidt (approved):
Rubberstamp

PR #3449, opened 2025-02-17
fix(chip): stop propagation of cancel

civing:
When closing the menu for the action menu the cancel event was never stopped. This cancel event eventually bubbled all the way up to the object creation dialog effectively closing it upon a scrim click of the action menu.

This fix stops the propagation of the cancel event following a close of the action menu dialog.

Fixes: Lundalogik/crm-feature#4613

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/chip/chip.tsx
```diff
@@ -351,6 +351,7 @@ export class Chip implements ChipInterface {
                 items={menuItems}
                 onSelect={this.handleActionMenuSelect}
                 openDirection="bottom-end"
+                onCancel={this.handleActionMenuCancel}
```
jgroth, line 354: Is this the correct place to fix this bug? :thinking: 
Maybe it should be handled in the dialog instead?

civing: I thought about that as well. I'm not sure how that menu are suposed to work, the event was there for a reason I suppose so if we block that event maybe it will be considered a braking change? I will walk around that could and try to decipher intent. But even if it's a "bug" it can be considered a "feature" I guess if people are using that event for some reason.

jgroth (approved):

src/components/chip/chip.tsx
civing: Okay after talking to you I realised I understood you incorrectly. On the chip itself the only source of  a `cancel` event is that dialog for the action menu and thus no documented behaviour for that event on the chip itself. And if someone is for some reason listening for that I think they need to really explain why. 🙂

The intent for that event us purely internal and only for that menu. I don't think it makes sense for someone to rely on that event. If we would extrapolate that we can hardly change anything 😄

Just my 2 cents.

PR #3455, opened 2025-02-19
Revert "chore(mixins): enable passing extra transitionable props to some mixins"

Kiarokh:
Reverts Lundalogik/lime-elements#3430

Since this var is undefined, it breaks all the other transitions caused by this mixin. So changes happen instantly instead. Right now I do not know how to fix this issue. But maybe later I can come up with a solution.


adrianschmidt (approved):
I'm trusting your judgement here 😄

PR #3458, opened 2025-02-21
feat(text editor): conditionally allow commands

john-traas:
This changes the `ActiveMenuItems` from a type to an interface.

Rather than handling some extremely complex actions that may or may not be part of a good UX, we should instead disable certain commands based on certain conditions. 

This change will allow us to do that both visually (not show the button in the menu) and within the command factory (ensure key bindings aren't used to trigger a command that's not allowed. 

## Summary by CodeRabbit

- **New Features**
	- The text editor’s toolbar now displays only the actions permitted by the current context, ensuring a more relevant and streamlined interface.
- **Refactor**
	- Underlying menu state handling has been improved to differentiate between active and permissible commands, further enhancing the consistency of available actions.
	- Action bar items are now filtered based on their allowed status, improving control over visibility.
- **Tests**
	- Introduced unit tests for the menu state tracking functionality to ensure robustness and correctness in various scenarios.

### DIFF
--- a/src/components/text-editor/prosemirror-adapter/plugins/menu-state-tracking-plugin.ts
+++ b/src/components/text-editor/prosemirror-adapter/plugins/menu-state-tracking-plugin.ts
@@ -7,24 +7,31 @@ import { EditorMenuTypes } from '../menu/types';
 
 export const actionBarPluginKey = new PluginKey('actionBarPlugin');
 
-export type ActiveMenuItems = Record<EditorMenuTypes, boolean>;
+export interface ActiveMenuItems {
+    active: Record<EditorMenuTypes, boolean>;
+    allowed: Record<EditorMenuTypes, boolean>;
+}
 
-export type UpdateMenuItemsCallBack = (activeTypes: ActiveMenuItems) => void;
+export type UpdateMenuItemsCallBack = (
+    activeTypes: Record<EditorMenuTypes, boolean>,
+    allowedTypes: Record<EditorMenuTypes, boolean>,
+) => void;
 
 const getMenuItemStates = (
     menuTypes: EditorMenuTypes[],
     menuCommandFactory: MenuCommandFactory,
     view: EditorView,
 ): ActiveMenuItems => {
-    const activeTypes: ActiveMenuItems = {};
+    const activeTypes: Record<EditorMenuTypes, boolean> = {};
+    const allowedTypes: Record<EditorMenuTypes, boolean> = {};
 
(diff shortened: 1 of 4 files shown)


src/components/text-editor/prosemirror-adapter/plugins/menu-state-tracking-plugin.ts
```diff
@@ -7,24 +7,31 @@ import { EditorMenuTypes } from '../menu/types';
...
 ): ActiveMenuItems => {
-    const activeTypes: ActiveMenuItems = {};
+    const activeTypes: Record<EditorMenuTypes, boolean> = {};
+    const allowedTypes: Record<EditorMenuTypes, boolean> = {};
 
     menuTypes.forEach((type) => {
         const command: CommandWithActive = menuCommandFactory.getCommand(type);
-        activeTypes[type] =
-            command && command.active && command.active(view.state);
+        activeTypes[type] = command?.active?.(view.state) || false;
```
adrianschmidt, line 30:
```suggestion
        activeTypes[type] = !!command?.active?.(view.state) || false;
```

This way, if `command` or `active` is null or undefined, we still set the value to `false`, rather than `null` or `undefined`.

src/components/text-editor/prosemirror-adapter/plugins/menu-state-tracking-plugin.ts
```diff
@@ -7,24 +7,31 @@ import { EditorMenuTypes } from '../menu/types';
...
+        allowedTypes[type] = command?.allowed?.(view.state) ?? true;
```
adrianschmidt, line 31:
```suggestion
        allowedTypes[type] = !!(command?.allowed?.(view.state) ?? true);
```

I think we need the parenthesis in this case, or any potential falsy value would first be converted to `false`, and _then_ compared again `??`, which would prevent the right side of the `??` from ever being executed.

adrianschmidt (approved):
Two minor suggestions. Otherwise this looks good to me. I mean, I have to admit that my eyes did gloss over  at times, so I may have missed things, but to the best of my ability to judge, this looks good 😅

src/components/text-editor/prosemirror-adapter/plugins/menu-state-tracking-plugin.ts
john-traas: ⚡ c17bc67664488c4eff60cd9fb4c7bf50cd198554

adrianschmidt (approved):

PR #1712, opened 2022-04-04
chore: bump tabulator to 5.x

anderssonjohan:
Work for upgrading limel-table component to use tabulator 5.x where a lot of things in the API have been changed.

Upgrade guide:
http://tabulator.info/docs/5.0/upgrade

Identified tasks:

- [x] callbacks moved to events
- [x] update sort events to use "output" interface (missing in `@types/tabulator-tables` and added here https://github.com/DefinitelyTyped/DefinitelyTyped/pull/58857)
- [x] rename ajax-related props according to upgrade guide
- [x] make remote load work with the new data loaders
- [ ] turn off invalidOption warnings when being sure all renamed options are migrated/taken care of (http://tabulator.info/docs/5.0/upgrade#debug)

## Review:
- [ ] Commits are [atomic](https://seesparkbox.com/foundry/atomic_commits_with_git)
- [ ] Commits have the correct *type* for the changes made
- [ ] Commits with *breaking changes* are marked as such

### Browsers tested:
(Check any that applies, it's ok to leave boxes unchecked if testing something didn't seem relevant.)

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS
adjust imports

PR #3478, opened 2025-03-06
chore(table): update tabulator to 6

civing:
@coderabbitai summary

fixes: Lundalogik/crm-feature#4449

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,7 @@
     "@types/lodash-es": "^4.17.12",
     "@types/react": "^19.0.10",
     "@types/react-dom": "^19.0.4",
-    "@types/tabulator-tables": "^4.9.4",
+    "@types/tabulator-tables": "^6.2.4",
     "@typescript-eslint/eslint-plugin": "^8.24.0",
     "@typescript-eslint/parser": "^8.24.0",
     "codemirror": "^5.65.9",
@@ -107,7 +107,7 @@
     "shelljs": "0.8.5",
     "shx": "^0.3.3",
     "style-to-object": "^1.0.8",
-    "tabulator-tables": "^4.9.3",
+    "tabulator-tables": "^6.3.1",
     "typescript": "^4.9.5",
     "unified": "^11.0.5",
     "unist-util-visit": "^5.0.0"
--- a/src/components/table/table-selection.ts
+++ b/src/components/table/table-selection.ts
@@ -1,5 +1,10 @@
 import { EventEmitter } from '@stencil/core';
-import Tabulator from 'tabulator-tables';
+import {
+    Tabulator,
+    ColumnDefinition as TabulatorColumnDefinition,
+    CellComponent as TabulatorCellComponent,
+    RowComponent as TabulatorRowComponent,
+} from 'tabulator-tables';
 import { setElementProperties } from './columns';
 import { ElementPool } from './element-pool';
 import { Selection, SelectionChangeSet } from './selection';
@@ -79,12 +84,12 @@ export class TableSelection {
      * @returns The column definitions with the checkbox column prepended to it
      */
     public getColumnDefinitions(
-        columnDefinitions: Tabulator.ColumnDefinition[],
-    ): Tabulator.ColumnDefinition[] {
+        columnDefinitions: TabulatorColumnDefinition[],
+    ): TabulatorColumnDefinition[] {
         return [this.getRowSelectorColumnDefinition(), ...columnDefinitions];
     }
 
-    private getRowSelectorColumnDefinition(): Tabulator.ColumnDefinition {
+    private getRowSelectorColumnDefinition(): TabulatorColumnDefinition {
         return {
             title: '',
             formatter: this.getRowSelectorFormatter(),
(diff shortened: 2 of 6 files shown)


Kiarokh:
Not sure if anything from this old PR is useful here. But might be worth to double check https://github.com/Lundalogik/lime-elements/pull/1712

src/components/table/table.tsx
```diff
@@ -8,7 +8,20 @@ import {
...
+    Tabulator,
+    Options as TabulatorOptions,
+    Sorter as TabulatorSorter,
+    SorterFromTable as TabulatorSorterFromTable,
+    SortDirection as TabulatorSortDirection,
+    ColumnComponent as TabulatorColumnComponent,
+    ColumnDefinition as TabulatorColumnDefinition,
+    RowComponent as TabulatorRowComponent,
+    OptionsData as TabulatorOptionsData,
+    OptionsPagination as TabulatorOptionsPagination,
```
Kiarokh, line 22: One of the styling issues that we face in the Webclient is because the pagination bar is not rendered anymore (in this updated version). 
image

jgroth:
> Not sure if anything from this old PR is useful here. But might be worth to double check #1712

I checked the old PR and made some updates. All examples seems to work now

Kiarokh:
> > Not sure if anything from this old PR is useful here. But might be worth to double check #1712
> 
> I checked the old PR and made some updates. All examples seems to work now

I had a quick look. It seems the only thing that is different is our customization for the sorting icon in column headers which is gone now. But maybe we can live with that (if the directions are as we intended them to be)

src/components/table/table-selection.ts
```diff
@@ -170,7 +179,7 @@ export class TableSelection {
         }
     };
 
-    private getActiveRows: () => Tabulator.RowComponent[] = () => {
+    private getActiveRows: () => TabulatorRowComponent[] = () => {
```
Kiarokh, line 182: https://github.com/user-attachments/assets/124bd8f0-fd18-48f0-93af-ea13cc3ad7f9

in the web client, checking the boxes will also open the sidepanel. They are considered to be a click on the row

jgroth: Fixed in f52d3d474abd5a93d3bac4889040dfb92dbcde9b

package.json
```diff
@@ -58,7 +58,7 @@
     "@types/lodash-es": "^4.17.12",
     "@types/react": "^19.0.12",
     "@types/react-dom": "^19.0.4",
-    "@types/tabulator-tables": "^4.9.4",
+    "@types/tabulator-tables": "^6.2.4",
```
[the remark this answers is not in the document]
Kiarokh: @coderabbitai this file `@import '../../../node_modules/tabulator-tables/scss/tabulator.scss';` doesn't seem to exists. What do you mean?

src/components/table/table.tsx
```diff
@@ -331,7 +331,9 @@ export class Table {
                 return;
             }
 
-            this.tabulator.updateOrAddData(newData);
+            if (newData.length > 0) {
+                this.tabulator.updateOrAddData(newData);
+            }
```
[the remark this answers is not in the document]
Kiarokh: Yes. sorry I squashed the change. so you don't see what was before. There was no `if`

before:
```tsx
            // if (newData.length > 0) {
                this.tabulator.updateOrAddData(newData);
            // }
```

The intelligent said:

> The e2e tests were failing because of a change in behavior in the newer version of Tabulator. The issue was:
> 
> When `updateData` was called with empty data or identical data, the code would fall through to calling `this.tabulator.updateOrAddData(newData)`
> In the new Tabulator version, calling `updateOrAddData([]) `with an empty array triggers a "No data provided" warning/error
> This caused the tests to fail with "Protocol error: Target closed" errors
> The fix: Added a check before calling `updateOrAddData` to ensure we only call it when there's actually data to add

Kiarokh:
> There seems to be some issues with this. The first time I try to paginate I get an error in the webclient. It goes away after a few seconds, but the data is still the old data even if it says it is displaying the second page. Pagination after the error seems to work (even the page that initially did not work)
> 

I don't know… 🙈 I think I'll leave this to the engineers

LucyChyzhova:
here: 

https://github.com/user-attachments/assets/2d06206d-82b2-4631-b1e6-b7b96441c9bb

jgroth:
It seems to work on the PR link
image

Kiarokh:
> It seems to work on the PR link

Sorry for the fake news. Yes it work. I think my local branch wasn't updated yet.

Kiarokh:
I've tested this in the Webclient and Admin, as well as Docs. Things seem to work fine. Let's merge

PR #3482, opened 2025-03-09
fix(table): apply new columns without a full init

civing:
Don't do a full initialization of the tabulator just because the column definition changed. Instead, set the columns and do a sorting when the rendering is complete.

fix: Lundalogik/crm-feature#4449

@coderabbitai summary

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -200,10 +200,12 @@ export class Table {
     private firstRequest: boolean;
     private currentSorting: ColumnSorter[];
     private tableSelection: TableSelection;
+    private shouldSort: boolean = false;
 
     constructor() {
         this.handleDataSorting = this.handleDataSorting.bind(this);
         this.handlePageLoaded = this.handlePageLoaded.bind(this);
+        this.handleRenderComplete = this.handleRenderComplete.bind(this);
         this.handleAjaxRequesting = this.handleAjaxRequesting.bind(this);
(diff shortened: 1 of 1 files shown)


john-traas (dismissed):
Looks good! However, there are some tests failing.

src/components/table/table.tsx
```diff
@@ -200,10 +200,12 @@ export class Table {
     private firstRequest: boolean;
     private currentSorting: ColumnSorter[];
     private tableSelection: TableSelection;
+    private shouldSort = false;
 
     constructor() {
         this.handleDataSorting = this.handleDataSorting.bind(this);
         this.handlePageLoaded = this.handlePageLoaded.bind(this);
+        this.handleRenderComplete = this.handleRenderComplete.bind(this);
```
Befkadu1, line 208: I think it is better to remove all these `bind` stuff in the cool down and convert those function to arrow functions instead.

civing: Yeah I had the same thought

john-traas (approved):

PR #3487, opened 2025-03-12
fix(snackbar): add missing translations for the dismiss button

Kiarokh:
@coderabbitai summary

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/translations/en.ts
```diff
@@ -23,6 +23,7 @@ export default {
     'file-viewer.more-actions': 'More…',
     'editor-menu.bold': 'Bold',
     'editor-menu.italic': 'Italic',
+    'editor-menu.strikethrough': 'Strikethrough',
```
Kiarokh, line 26: before
image


after
image

jgroth (dismissed):

src/global/translations.ts
```diff
@@ -21,7 +21,11 @@ const allTranslations = {
 const REGEX = /\{\s*(\w+)\s*\}/g;
 
 export class Translations {
-    public get(key: string, language = 'en', params?: object): string {
+    public get(
+        key: string,
+        language = 'en',
+        params?: object,
+    ): string | undefined {
```
Kiarokh, line 28: @jgroth I rebased on main and there was a change in `translations.st` file which was conflicting with my change. This is how the conflict looked like:

image

The `string` type was already added in this PR
https://github.com/Lundalogik/lime-elements/pull/3464/files#diff-8a5b1230d00b72dede2900072c8f3d7fae328c9488f7ed5aba33b6f3d19385f4

So I just added the `| undefined`.

You need to reapprove this PR

jgroth: But the new implementation can no longer return `undefined` so this change is no longer needed :smile:

Kiarokh: ⚡

jgroth (approved):

PR #3511, opened 2025-03-26
feat(action-bar): add collapsible functionality

LucyChyzhova:
fix: https://github.com/Lundalogik/crm-feature/issues/3926

@coderabbitai summary

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/action-bar/action-bar.tsx
```diff
@@ -132,13 +139,17 @@ export class ActionBar {
                 class={{
                     'is-full-width': this.layout === 'fullWidth',
                     'is-floating': this.layout === 'floating',
+                    'is-shrunk':
+                        !!this.actionBarIsShrunk && !!this.isExpandable,
+                    'can-be-shrunk': !!this.isExpandable,
                 }}
                 role="grid"
             >
                 <div class="items" role="rowgroup">
```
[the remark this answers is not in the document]
LucyChyzhova: noooooooo

src/components/action-bar/action-bar.tsx
```diff
@@ -187,6 +237,17 @@ export class ActionBar {
         }
     };
 
+    private get tooltipLabel() {
+        const expand = this.getTranslation('action-bar.expand');
+        const collapse = this.getTranslation('action-bar.collapse');
+
+        return this.actionBarIsShrunk ? expand : collapse;
+    }
```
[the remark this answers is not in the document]
LucyChyzhova: Lint doesn't like it!

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
```diff
@@ -28,6 +30,15 @@ export class ActionBarOverflowMenu {
     @Prop({ reflect: true })
     public openDirection: OpenDirection = 'bottom-end';
 
+    @Prop()
+    public actionBarIsShrunk = false;
```
civing, line 34: To make this even more generic, could you instead add a `Icon` to the interface?
```suggestion
    @Prop()
    public overFlowIcon?: Icon;
```

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
```diff
@@ -48,6 +59,15 @@ export class ActionBarOverflowMenu {
...
 
     private countOverflowedItems = () => {
+        if (this.actionBarIsShrunk) {
+            return (
+                <limel-icon
+                    name="more"
+                    title={this.getTranslation('action-bar.actions')}
+                />
+            );
+        }
```
civing, line 69: Here we could render `overFlowIcon` if not undefined

src/components/action-bar/action-bar.tsx
```diff
@@ -75,6 +84,12 @@ export class ActionBar {
     @Prop({ reflect: true })
     public layout?: 'fullWidth' | 'floating';
 
+    @Prop({ reflect: true })
+    public actionBarIsShrunk = false;
+
+    @Prop({ reflect: true })
+    public isExpandable = false;
```
civing, line 91: Collapsible?
```suggestion
    @Prop({ reflect: true })
    public collapsible = false;
```

src/components/action-bar/action-bar.tsx
```diff
@@ -75,6 +84,12 @@ export class ActionBar {
     @Prop({ reflect: true })
     public layout?: 'fullWidth' | 'floating';
 
+    @Prop({ reflect: true })
+    public actionBarIsShrunk = false;
```
civing, line 88: Why do we have this as a prop? Only to reflect it?

src/components/action-bar/action-bar.tsx
```diff
@@ -170,10 +189,41 @@ export class ActionBar {
...
+                <limel-tooltip
+                    label={this.tooltipLabel}
+                    elementId="tooltip-expand-shrink-button"
+                />
+            </button>
+        );
+    }
+
+    private handleCollapseExpandClick = () => {
+        this.actionBarIsShrunk = !this.actionBarIsShrunk;
```
civing, line 224: You can't mutate a prop! :)

src/components/action-bar/examples/action-bar-floating-with-expand-button.tsx
```diff
@@ -0,0 +1,65 @@
...
+ *
+ */
+@Component({
+    tag: 'limel-example-action-bar-floating-expand',
+    shadow: true,
+    styleUrl: 'action-bar-floating.scss',
+})
+export class ActionBarFloatingExpandExample {
+    @State()
+    private actionBarItems: Array<ActionBarItem | ListSeparator> = [
```
Befkadu1, line 19: move above the `@Component({`, this can't be a state since there is nothing to update. I see that we have mentioned this in many components and better to create a new file called `action-bar-items` and reuse it. Of course this can be a refactor work. What do you think? 

```suggestion
    const actionBarItems: Array<ActionBarItem | ListSeparator> = [
```

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
Befkadu1: I think we don't need to pass the icon from the parent component since the icon is hard coded. I added my suggestion [here](https://github.com/Lundalogik/lime-elements/pull/3511#discussion_r2018526558)

src/components/action-bar/action-bar.tsx
LucyChyzhova: I have changed it to `@State()`

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
LucyChyzhova: I went with Pelle's approach here.  
@civing Could you check Bef's comment here.

src/components/action-bar/action-bar.scss
```diff
@@ -4,8 +4,12 @@
 * @prop --action-bar-item-max-width: Maximum width of a button in the action bar. Defaults to `10rem`. Keep in mind that the buttons should not appear too big.
 * @prop --action-bar-border-radius: Defines the roundness of the corners of the action bar. Defaults to `0`.
 * @prop --action-bar-background-color: Background color of the whole component. Defaults to `--contrast-100`.
+* @prop --action-bar-shrink-icon-color: Color of the optional icons used in each dock item. Only affects inactive dock items, defaults to text colors for default or selected states.
```
Befkadu1, line 7: What is this `dock items` ?

LucyChyzhova: Good you read it! fixed 😀

LucyChyzhova:
I need help with tests related to `translations` 🙏

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
```diff
@@ -48,6 +56,18 @@ export class ActionBarOverflowMenu {
...
+                <limel-icon
+                    style={{
+                        color: `${this.overFlowIcon.color}`,
+                    }}
+                    name={this.overFlowIcon.name}
+                    aria-label={this.overFlowIcon.title}
+                />
+            );
+        }
+
```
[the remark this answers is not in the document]
civing: I don't see why we need to leak the concept of the action bar being shrunk down to this component. :)

It's much easier to read this component if the consumer explicitly sets a overFlowIcon and if that's set it's rendered instead of the default +4 or whatever number of overflow actions is present.

So, I prefer the current way since it decouples the two components. And as a bonus, if we find more reasons to change the overflow icon from the action bar we have a clear interface.

src/components/action-bar/action-bar.tsx
```diff
@@ -164,16 +186,56 @@ export class ActionBar {
...
+
         return (
             <limel-action-bar-overflow-menu
                 openDirection={this.openDirection}
                 items={items}
                 onSelect={this.handleSelect}
                 role="gridcell"
+                overFlowIcon={
+                    this.actionBarIsShrunk ? shrunkOverFlowIcon : undefined
+                }
```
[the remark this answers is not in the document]
Befkadu1: As far I see the consumer of `limel-action-bar-overflow-menu` is only `action-bar.tsx` and I see no use to pass the icon.  The name, `limel-action-bar-overflow-menu` also indicates it would be useful only in `action-bar` because of that we can rule out that `limel-action-bar-overflow-menu`  can be used by other components I think.

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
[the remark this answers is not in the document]
adrianschmidt: What is the purpose of adding the new property? (Regardless of what we want to call it.)

Is the purpose to let the consumer control whether the action bar is collapsed, or is the purpose to let the consumer set which icon is used to indicate that there are hidden items? Or something else?

src/components/action-bar/action-bar.tsx
adrianschmidt: Yes, we have done it in other places. That doesn't mean it's a good practice. We should avoid it, unless we have very good reason to use it.

Befkadu1: I know @adrianschmidt I am just spreading the knowledges that it is possible to mutate

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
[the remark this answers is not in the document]
Befkadu1: The purpose of the new property is when to show `more` icon `...`  and a button displaying a number `+1`. 
As you see in the attached video if the action bar is shrunk show `more` icon `...` else the number of hidden actions like this `+1`.

src/components/action-bar/action-bar.tsx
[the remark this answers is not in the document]
adrianschmidt: Thank you for explaining!

I think I agree with @civing in this case. The menu doesn't need to know anything about whether the action bar is shrunk or not. So by using `actionBarIsShrunk`, we are essentially leaking implementation details into the sub-component. That's not as bad as leaking to outside components, but it's still unnecessary.

On the contrary, I think it makes sense that the action-bar controls what the menu button is showing.

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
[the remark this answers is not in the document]
adrianschmidt: Thank you for explaining!

I think I agree with @civing in this case. The menu doesn't need to know anything about whether the action bar is shrunk or not. So by using `actionBarIsShrunk`, we are essentially leaking implementation details into the sub-component. That's not as bad as leaking to outside components, but it's still unnecessary.

On the contrary, I think it makes sense that the action-bar controls what the menu button is showing.

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
```diff
@@ -48,6 +56,18 @@ export class ActionBarOverflowMenu {
     }
 
     private countOverflowedItems = () => {
+        if (this.overFlowIcon) {
```
Befkadu1, line 59: The function name `countOverflowedItems` is no more descriptive for what it does. I would suggest this instead `getOverflowMenuTriggerContent` @LucyChyzhova

Befkadu1 (dismissed):

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
LucyChyzhova: let's simplify it to getOverflowTriggerContent

src/components/action-bar/action-bar-item/action-bar-overflow-menu.tsx
```diff
@@ -42,12 +50,26 @@ export class ActionBarOverflowMenu {
...
+                <limel-icon
+                    style={{
+                        color: `${this.overFlowIcon.color}`,
+                    }}
+                    name={this.overFlowIcon.name}
+                    aria-label={this.overFlowIcon.title}
+                />
+            );
+        }
+
```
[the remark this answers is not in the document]
Befkadu1: I will fix this one

Befkadu1 (approved):

Befkadu1 (approved):

src/components/action-bar/action-bar.tsx
civing: Yes I know it's possible but that decorators extra parameter "mutable" should really be used with caution.

src/components/action-bar/action-bar.tsx
```diff
@@ -132,13 +147,17 @@ export class ActionBar {
                 class={{
                     'is-full-width': this.layout === 'fullWidth',
                     'is-floating': this.layout === 'floating',
+                    'is-shrunk':
+                        !!this.actionBarIsShrunk && !!this.isExpandable,
+                    'can-be-shrunk': !!this.isExpandable,
                 }}
                 role="grid"
             >
```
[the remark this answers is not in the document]
adrianschmidt: @LucyChyzhova (and @Kiarokh): SonarQube complains a _lot_ about us using ARIA table roles instead of actual tables, and other cases where it says we are not using the correct combination of elements and roles, like loads of places where we give elements that normally aren't interactive a `role=button`.

Since CodeRabbit also has started complaining about stuff like that, I thought I'd point it out.

jgroth (approved):

adrianschmidt (approved):
The changes to the external API look reasonable, and internal stuff we can always update 🙂

adrianschmidt:
Haha, 🥷'd…

😝

PR #3516, opened 2025-04-02
feat(table): introduce recent active markers

paulinewahle:
fix: https://github.com/Lundalogik/crm-feature/issues/1294

## Summary by CodeRabbit

- **New Features**
  - Recently active table rows are now visually indicated, allowing users to see not only the currently selected row but also the previous one.  
- **Style**
  - Enhanced table row styling with a consistent left border indicator for selected and recently active rows.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/partial-styles/tabulator-custom-styles.scss
+++ b/src/components/table/partial-styles/tabulator-custom-styles.scss
@@ -1,6 +1,22 @@
 @use '../../../style/mixins';
 @use '../../../style/functions';
+@use '../../../style/theme-color-variables';
+@use '../../../style/color-palette-extended';
 
+@mixin table-row-selected-indicator($border-color, $z-index) {
+    &:before {
+        $width-of-selected-row-indicator: 0.2rem;
+        content: '';
+        display: inline-block;
+        box-sizing: border-box;
+        position: sticky;
+        z-index: $z-index;
+        inset: 0 0 auto 0;
+        border: $width-of-selected-row-indicator solid $border-color;
+        border-radius: 1rem;
+        margin-right: -$width-of-selected-row-indicator * 2;
+    }
+}
 .interactive-feedback {
     // This is a div, injected by us into all rows.
     // We use it to visualize interactive feedback.
@@ -45,19 +61,10 @@
                 box-shadow: var(--button-shadow-inset-pressed);
             }
 
-            &:before {
-                $width-of-selected-row-indicator: 0.2rem;
-                content: '';
-                display: inline-block;
-                box-sizing: border-box;
-                position: sticky;
-                z-index: $table--has-interactive-rows--selectable-row--hover + 1;
-                inset: 0 0 auto 0;
-                border: $width-of-selected-row-indicator solid
-                    var(--mdc-theme-primary);
-                border-radius: 1rem;
-                margin-right: -$width-of-selected-row-indicator * 2;
-            }
+            @include table-row-selected-indicator(
+                var(--mdc-theme-primary),
+                $table--has-interactive-rows--selectable-row--hover + 1
+            );
         }
     }
 }
@@ -88,6 +95,12 @@
                 }
             }
         }
+    }
+    .recent-active {
+        @include table-row-selected-indicator(
+            rgb(var(--contrast-700)),
+            $table--has-interactive-rows--selectable-row--hover + 1
+        );
     }
 }
 
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -41,11 +41,7 @@ const FIRST_PAGE = 1;
  * @exampleComponent limel-example-table-layout-low-density
  * @exampleComponent limel-example-table-interactive-rows
  */
-@Component({
-    tag: 'limel-table',
-    styleUrl: 'table.scss',
-    shadow: true,
-})
+@Component({ tag: 'limel-table', styleUrl: 'table.scss', shadow: true })
 export class Table {
     /**
      * Data to be displayed in the table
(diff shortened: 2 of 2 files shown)


src/components/table/table.tsx
```diff
@@ -41,11 +41,7 @@ const FIRST_PAGE = 1;
  * @exampleComponent limel-example-table-layout-low-density
  * @exampleComponent limel-example-table-interactive-rows
  */
-@Component({
-    tag: 'limel-table',
-    styleUrl: 'table.scss',
-    shadow: true,
-})
+@Component({ tag: 'limel-table', styleUrl: 'table.scss', shadow: true })
```
Kiarokh, line 44: 🤔 this looks like an auto lint on save; and shouldn't be committed with your changes in this PR. I guess you should
1. make sure your branch is rebased on the latest `main`
2. run `npm i` 
3. save this file again, and see if these changes are removed from your PR

src/components/table/table.tsx
```diff
@@ -810,11 +831,7 @@ export class Table {
 
     render() {
         return (
-            <Host
-                class={{
-                    'has-low-density': this.layout === 'lowDensity',
-                }}
-            >
+            <Host class={{ 'has-low-density': this.layout === 'lowDensity' }}>
```
Kiarokh, line 834: see https://github.com/Lundalogik/lime-elements/pull/3516/files#r2026352349

src/components/table/partial-styles/tabulator-custom-styles.scss
```diff
@@ -87,6 +87,22 @@
...
+                    position: sticky;
+                    z-index: $table--has-interactive-rows--selectable-row--hover +
+                        1;
+                    inset: 0 0 auto 0;
+                    border: $width-of-selected-row-indicator solid
+                        rgb(211, 211, 211);
+                    border-radius: 1rem;
+                    margin-right: -$width-of-selected-row-indicator * 2;
+                }
+            }
```
Kiarokh, line 105: this is mostly repetition of lines 48 to 60.
it's better to either make a `mixin` for these lines; or have a common CSS rule for both. The only difference is the colors of the dots, which can the be written in a separate CSS rule for each

src/components/table/table.tsx
paulinewahle: What do you mean?

Kiarokh: I meat to link it to the previous comment about auto linting :)

src/style/mixins.scss
```diff
@@ -602,3 +602,19 @@ $narrow-dialog-breakpoint: 760px;
...
+        box-sizing: border-box;
+        position: sticky;
+        //z-index: $table--has-interactive-rows--selectable-row--hover + 1;
+        z-index: $z-index;
+        inset: 0 0 auto 0;
+        border: $width-of-selected-row-indicator solid $border-color;
+        border-radius: 1rem;
+        margin-right: -$width-of-selected-row-indicator * 2;
+    }
+}
```
Kiarokh, line 620: this is great! However you can keep this mixin in the same file where it is used. This is because it is not a reusable mixin for any other component. We only use this file (`mixins.scss`) for shared mixins (both in components in this codebase, and in other consumers of Lime Elements). Local mixins are better to be kept where they are used.

src/components/table/partial-styles/tabulator-custom-styles.scss
```diff
@@ -87,24 +78,14 @@
...
-                        rgb(211, 211, 211);
-                    border-radius: 1rem;
-                    margin-right: -$width-of-selected-row-indicator * 2;
-                }
-            }
         }
     }
+    .recent-active {
+        @include mixins.table-row-indicator(
+            rgb(200, 200, 200),
```
Kiarokh, line 85: see [`0ffb03f` (#3516)](https://github.com/Lundalogik/lime-elements/pull/3516/commits/0ffb03f2598a4fa1b7da3df0036b65f651868830#r2088491986)

src/style/mixins.scss
```diff
@@ -602,3 +602,19 @@ $narrow-dialog-breakpoint: 760px;
         @content;
     }
 }
+
+@mixin table-row-indicator($border-color, $z-index) {
```
Kiarokh, line 606:
```suggestion
@mixin table-row-selected-indicator($border-color, $z-index) {
```

Since the mixin is only responsible for visualizing the row that is _selected_, it's better to name it this way. Otherwise, the reader can be confused about what this piece of code actually does.

Kiarokh:
two things I can think of:
1. Since we are upgrading Tabulator here https://github.com/Lundalogik/lime-elements/pull/3478 we should revisit this issue after the upgrade
2. We should also have a more global approach here. We have other views, such as List view, Kanban view, etc… All of these should have some sort of support for indicating the "previously selected item". We can't only implement this in the table view. How would we do this best? 🤔

john-traas:
> two things I can think of:
> 
> 1. Since we are upgrading Tabulator here [chore(table): update tabulator to 6 #3478](https://github.com/Lundalogik/lime-elements/pull/3478) we should revisit this issue after the upgrade
> 2. We should also have a more global approach here. We have other views, such as List view, Kanban view, etc… All of these should have some sort of support for indicating the "previously selected item". We can't only implement this in the table view. How would we do this best? 🤔

Is this a blocker to getting this PR in? Should we create issues for those other views?

Kiarokh:
> Is this a blocker to getting this PR in? Should we create issues for those other views?

I think it's just hard to remember to check and verify it in the https://github.com/Lundalogik/lime-elements/pull/3478 
I'd at least wait for that PR to be merged first.

And I think we can make an umbrella issue for all views, and connect this as a sub-task to that one.
We will have many different implementations in various repos, to address the issue of:

> - Hey, which item did I select before this one?
> - Oh, which item was selected, just before I deselected it?

PR #3518, opened 2025-04-04
get rid of MDC input fields (step 1)

Kiarokh:
This is a private component, used to render a notched outline around all input elements that can have a floating label. Inspired by Material Design's styles for input fields. We will use it in various components to unify styles and avoid repeating code.

fix https://github.com/Lundalogik/crm-feature/issues/4696

>[!IMPORTANT]
>- This PR updates `limel-input-field` which is used in several other components, such as date-picker, color-picker, file. A visual inspection of those components would be nice during the review.
>- it also updates `limel-chip-set` which is used in `limel-picker`. That one also needs to be tested.

## Summary by CodeRabbit

- **New Features**
  - Introduced an interactive notched outline component with a live example, enhancing the presentation of input fields.
  - Added a new basic example for the notched outline component to demonstrate its functionality.

- **Refactor**
  - Streamlined the layout and rendering of input fields and text editors for improved modularity and accessibility.
  - Enhanced the management of identifiers for input fields to improve accessibility.
  - Simplified the rendering logic in the `TextEditor` and `SelectTemplate` components for better structure.

- **Style**
  - Updated visual styling for input elements, icons, and various component states (e.g., focus, disabled) for consistent user experience.
  - Added new styles for the notched outline component and its states.
  - Removed unnecessary styles from the text editor and input field components to simplify the design.
  - Introduced a new border radius variable for select components.
  - Updated styles for the chip set component to reflect new design choices.
  - Enhanced styles for the `limel-notched-outline` component to improve visual hierarchy and responsiveness.

- **Tests**
  - Refined component tests to focus on key visual and accessibility attributes for more robust validation.
  - Updated end-to-end tests to align with the new structure of components, ensuring accurate verification of labels and outlines.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.scss
+++ b/src/components/chip-set/chip-set.scss
@@ -6,9 +6,6 @@
 
 @use '@material/textfield';
 @use '@material/textfield/icon';
-@use '@material/notched-outline/mdc-notched-outline';
-@use '@material/floating-label';
-@use '@material/floating-label/mdc-floating-label';
 
 /**
  * @prop --icon-background-color: Background color of the icon. Defaults to transparent.
@@ -21,20 +18,28 @@
 @include textfield.core-styles;
 @include icon.icon-core-styles;
 
-@include shared_input-select-picker.outlined-style-overrides;
-@include shared_input-select-picker.floating-label-overrides;
-@include shared_input-select-picker.cropped-label-hack;
-@include shared_input-select-picker.disabled-overrides;
-@include shared_input-select-picker.readonly-overrides;
 @include shared_input-select-picker.leading-icon;
-@include shared_input-select-picker.lime-empty-value-for-readonly;
-@include shared_input-select-picker.lime-looks-like-input-value;
 
 $height-of-chip-set-input: functions.pxToRem(36);
 $leading-icon-space: 1.5rem;
 
 :host(limel-chip-set) {
     isolation: isolate;
+}
+
+:host(limel-chip-set[type='input']) {
+    limel-notched-outline {
+        [slot='content'] {
+            min-height: shared_input-select-picker.$height-of-mdc-text-field;
+        }
+    }
+}
+
+:host(limel-chip-set:not([type='input'])) {
+    .limel-notched-outline {
+        --limel-notched-outline-border-color: transparent;
+        --limel-notched-outline-background-color: transparent;
+    }
 }
 
 .mdc-chip-set {
@@ -44,12 +49,6 @@ $leading-icon-space: 1.5rem;
     gap: 0.5rem;
     min-height: shared_input-select-picker.$height-of-mdc-text-field;
     position: relative;
-
-    &.chip-set--with-label {
-        .lime-floating-label--float-above {
-            padding-left: functions.pxToRem(4);
-        }
-    }
 
     &.mdc-chip-set--input {
         padding: 0.4rem 0.5rem;
@@ -75,7 +74,7 @@ $leading-icon-space: 1.5rem;
         @include shared_input-select-picker.input-field-placeholder;
 
         width: auto;
-        padding: 0 functions.pxToRem(12);
+        padding: 0 0.5rem;
 
         flex-grow: 1;
         flex-shrink: 0;
@@ -100,39 +99,6 @@ $leading-icon-space: 1.5rem;
     }
 }
 
-// used only in chipsets that do not have input
-.chip-set__label {
-    @include mixins.truncate-text;
-    width: 120%; // `120%` instead of `100%`,
-    // because this class is always together with `mdc-floating-label--float-above`,
-    // which scales the label down. So there is more horizontal space to display the label in.
-    top: functions.pxToRem(13);
-    padding-left: functions.pxToRem(20);
-}
-
-// Because MDC removes some classes in chipset, we add custom
-// classes with similar names and expected behavior & styles.
-// These class names start with `lime-`, instead of `mdc-`.
-.lime-notched-outline--notched {
-    .mdc-notched-outline__notch {
-        border-top: 1px solid transparent !important;
-
-        .lime-floating-label--float-above {
-            // This overrides MDC's original top value which is `top: 50%`.
-            // The reason is that a % value aligns the label in a wrong position
-            // vertically, when there are multiple rows of chips.
-            top: functions.pxToRem(27);
-
-            transform: translateY(-34.75px) scale(0.75) !important;
-            font-size: shared_input-select-picker.$cropped-label-hack--font-size;
-        }
-    }
-}
-
-.force-invalid {
-    @extend .mdc-text-field--invalid;
-}
-
 .clear-all-button {
     @include mixins.clear-all-button;
     @include mixins.visualize-keyboard-focus;
@@ -144,14 +110,14 @@ $leading-icon-space: 1.5rem;
     opacity: 0; // Is hidden, but can receive focus (such as when navigating through tab indexes).
 
     &:focus,
-    .has-chips:not(.mdc-text-field--disabled):hover &,
-    .has-chips:not(.mdc-text-field--disabled).mdc-text-field--focused & {
+    .has-chips:not(.disabled):hover &,
+    .has-chips:not(.disabled).mdc-text-field--focused & {
         opacity: 1;
         outline: none;
     }
 
     .mdc-chip-set:not(.has-chips) &,
-    .has-chips.mdc-text-field--disabled & {
+    .has-chips.disabled & {
         display: none; // Won't receive focus when disabled
     }
 }
@@ -160,15 +126,6 @@ $leading-icon-space: 1.5rem;
     &:not(.has-chips) {
         .mdc-text-field__input {
             padding-left: $leading-icon-space;
-        }
-
-        .mdc-floating-label {
-            &:not(.lime-floating-label--float-above) {
-                left: $leading-icon-space;
-            }
-            &.mdc-floating-label--float-above {
-                left: functions.pxToRem(4);
-            }
         }
     }
 
@@ -207,10 +164,3 @@ limel-chip {
 
 @import './partial-styles/_readonly';
 @import './partial-styles/_helper-text';
-
-// To make the input field render smaller than the default MDC UI
-.mdc-text-field {
-    &.mdc-text-field--outlined {
-        min-height: shared_input-select-picker.$height-of-mdc-text-field;
-    }
-}
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -18,6 +18,7 @@ import translate from '../../global/translations';
 import { getHref, getTarget } from '../../util/link-helper';
 import { isEqual } from 'lodash-es';
 import { LimelChipCustomEvent } from '../../components';
+import { createRandomString } from '../../util/random-string';
 
 const INPUT_FIELD_TABINDEX = 1;
 
@@ -243,8 +244,10 @@ export class ChipSet {
 
     private mdcTextField: MDCTextField;
     private readonly handleKeyDown = handleKeyboardEvent;
+    private labelId: string;
 
     constructor() {
+        this.labelId = createRandomString();
         this.renderChip = this.renderChip.bind(this);
         this.renderInputChip = this.renderInputChip.bind(this);
         this.isFull = this.isFull.bind(this);
@@ -419,7 +422,7 @@ export class ChipSet {
                     <input
                         tabIndex={INPUT_FIELD_TABINDEX}
                         type={this.inputType}
-                        id="input-element"
+                        id={this.labelId}
                         disabled={this.readonly || this.disabled}
                         class={{
                             'mdc-text-field__input': true,
@@ -483,7 +486,7 @@ export class ChipSet {
 
         return (
             <div class="mdc-notched-outline__notch">
-                <label class={labelClassList} htmlFor="input-element">
+                <label class={labelClassList} htmlFor={this.labelId}>
                     {this.label}
                 </label>
             </div>
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -7,7 +7,6 @@ import {
     Event,
     EventEmitter,
     h,
-    Host,
     Method,
     Prop,
     State,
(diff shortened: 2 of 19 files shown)


src/components/notched-outline/notched-outline.tsx
```diff
@@ -0,0 +1,140 @@
...
+    @Prop({ reflect: true })
+    public hasFloatingLabel = false;
+
+    public render() {
+        return (
+            <div class="limel-notched-outline">
+                <slot name="content" />
+                <span
+                    class="limel-notched-outline--outlines"
+                    aria-hidden="true"
```
Kiarokh, line 103: The `label` is visually inside the outlines container that has `aria-hidden="true"`. But screen readers will still associate the label with its input element due to the explicit `for` relationship. The label is correctly associated with its input via the `htmlFor` attribute

Kiarokh:
>[!NOTE]
>1. `limel-chip-set` is still left to do. But that component is over complicated and I think it is better to do it later in a separate PR.
>1. `limel-select` could also benefit from this new `limel-notched-outline` component. But I also indend to do it in a separate PR.

src/components/notched-outline/examples/notched-outline-basic.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+    private setHasLeadingIcon = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.hasLeadingIcon = event.detail;
+    };
+
+    private setHasFloatingLabel = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.hasFloatingLabel = event.detail;
+    };
```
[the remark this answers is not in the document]
Kiarokh: it would be nice if we did this in all docs examples actually…

src/components/notched-outline/notched-outline.tsx
```diff
@@ -0,0 +1,139 @@
...
+ * Inspired by Material Design's styles for input fields.
+ * We use it in various components to unify styles and avoid
+ * repeating code.
+ *
+ * :::note
+ * The component has `shadow: false`. This is to improve performance,
+ * and ensure that its internal elements are considered as a internal elements
+ * of the consumer's DOM. This way, the value `for` in `<label for="id-of-input-element">`
+ * would be correctly associated with the input element's `id`, in the consumer component.
+ * :::
```
Kiarokh, line 15: This was I think an important part of the strategy. However, there are minor risks of consumer styles leaking into the elements of this component. For example, if the consumer has styled a `div` element, then the `div`s within `limel-notched-outline` can inherit it. On the other hand, being a native part of the consumer's DOM is crucial for this setup to work. The consumer sometimes needs to be able to style something within the notch-group.

Maybe we could use other unconventional element names instead of `div`s to mitigate styles leaking unintentionally. For example instead of `<div class="limel-notched-outline"` we could write `<notched-outline class="limel-notched-outline"`? 🤔 

Maybe I'm being too careful. The consumer is always us, so we could be more careful in the implementation? What do you think?

src/components/notched-outline/examples/notched-outline-basic.tsx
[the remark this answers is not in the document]
adrianschmidt: @coderabbitai Thanks for the input! I have one question, and one point of feedback.

## Question
I assume this would be called using something like the following:

```
                <limel-checkbox
                    checked={this.disabled}
                    label="Disabled"
                    onChange={this.toggleState('disabled')}
                />
                <limel-checkbox
                    checked={this.readonly}
                    label="Readonly"
                    onChange={this.toggleState('readonly')}
                />
                <limel-checkbox
                    checked={this.required}
                    label="Required"
                    onChange={this.toggleState('required')}
                />
```

Stencil's documentation recommends that you avoid using `=>` or `.bind` in templates. As far as I've been led to understand, that's because they create a new copy of the function each time the render function is called. I'm not sure if that's the case or not, but if so, isn't that a problem here too?
(Even though this is just an example in the documentation, people often copy from these examples, so it's important that we use practices and coding patterns that are suitable for production code.)

## Feedback
The "Committable suggestion" that you provided is not actually committable. You used git diff syntax instead of GitHub's suggestion syntax.

src/components/notched-outline/notched-outline.tsx
adrianschmidt: > The consumer is always us

Yeah, I was a bit confused until I read this, because this component is going to be used by us, inside other components like `limel-input-field`, right? No-shadow is absolutely the way to go here 👍

src/components/notched-outline/examples/notched-outline-basic.tsx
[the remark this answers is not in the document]
Kiarokh: I think this could be a separate PR where we handle these in all other examples in the same time. For this PR, I think it's better to be consistent with the rest

src/components/notched-outline/notched-outline.scss
```diff
@@ -0,0 +1,246 @@
...
+    }
+
+    &[required]:not([required='false']) {
+        .limel-notched-outline--notch {
+            label {
+                padding-right: 0.75rem;
+
+                &:after {
+                    content: '*';
+                    scale: 1.5;
```
LucyChyzhova: It seems too intense now, check before and now:
before:
Screenshot 2025-04-08 at 14 06 30
Screenshot 2025-04-08 at 13 51 12

Now:
Screenshot 2025-04-08 at 13 50 30
Screenshot 2025-04-08 at 13 59 00
Screenshot 2025-04-08 at 13 58 56

Kiarokh: sharp eyes and good catch. I've commented out the `scale` for now, since there is a risk that we merge this PR and then some components that haven't implemented the notched-outline yet, would look different. Otherwise, I think it's nice to see a bigger asterisk in general :)

src/components/notched-outline/notched-outline.scss
```diff
@@ -0,0 +1,246 @@
...
+
+    &[readonly]:not([readonly='false']) {
+        --limel-notched-outline-border-color: transparent !important;
+        --limel-notched-outline-background-color: transparent !important;
+
+        .limel-notched-outline--notch {
+            label {
+                transition-duration: 0s;
+            }
+        }
```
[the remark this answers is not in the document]
Kiarokh: this is a conscious choice. Also currently we don't want readonly fields that are invalid to show any visual indication that makes them look like an input field. Users cannot edit invalid&readonly fields, so they can become really confused

src/components/input-field/input-field.tsx
```diff
@@ -321,21 +321,26 @@ export class InputField {
...
+                label={this.label}
+                required={this.required}
+                invalid={this.invalid || this.isInvalid()}
+                disabled={this.disabled}
+                readonly={this.readonly}
+                hasValue={!!this.value}
+                hasLeadingIcon={!!this.leadingIcon}
+            >
+                <label slot="content" class={this.getContainerClassList()}>
+                    {this.renderLeadingIcon()}
```
LucyChyzhova, line 335: Please check this video:

https://github.com/user-attachments/assets/e8a5931d-dd2e-4fdc-9c85-6a7d68c33c2b

When component has a leading icon floating label gets extra space to the left

src/components/select/select.scss
```diff
@@ -69,12 +71,6 @@
     flex-wrap: wrap;
     width: 100%;
 
-    &:not(.limel-select--readonly) {
```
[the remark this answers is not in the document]
Kiarokh: great find. I pushed a fixup! :)

src/components/notched-outline/notched-outline.scss
Kiarokh: Now that all components have implemented notched outline, I've commented the `scale` back

src/components/chip-set/chip-set.tsx
```diff
@@ -337,35 +340,60 @@ export class ChipSet {
...
+                hasValue={!!this.value}
+                hasLeadingIcon={!!this.leadingIcon}
+                hasFloatingLabel={this.floatLabelAbove()}
+                onClick={this.handleTextFieldFocus}
+            >
+                <div slot="content" class={classes}>
+                    {this.renderContent(value)}
+                </div>
+            </limel-notched-outline>,
+            this.renderHelperLine(),
```
Kiarokh, line 385: I've intentionally moved the helper-line here. Previously, only `type` `input` would show the helper text. But this way other types of `chip-set` can also show the helper text. I believe this is the correct and consistent way of doing it. There is no point for the component to have a prop, that only works for a certain type.

src/components/chip-set/chip-set.scss
```diff
@@ -45,12 +50,6 @@ $leading-icon-space: 1.5rem;
     min-height: shared_input-select-picker.$height-of-mdc-text-field;
     position: relative;
 
-    &.chip-set--with-label {
```
LucyChyzhova, line 48: not particularly about this line, but could you take a look at the before and after:
before:
Screenshot 2025-04-10 at 10 25 36
after:
Screenshot 2025-04-10 at 10 23 08

Do you want it to have it that close - label and chips? i like more with some space :)

src/components/chip-set/chip-set.scss
```diff
@@ -45,12 +50,6 @@ $leading-icon-space: 1.5rem;
...
-        .lime-floating-label--float-above {
```
LucyChyzhova, line 49: before:

https://github.com/user-attachments/assets/27f71da5-f3b6-4f93-8c23-b4395b9068eb

after:

https://github.com/user-attachments/assets/9de20001-19f4-40ae-9cc5-26c0a147e795

src/components/text-editor/text-editor.scss
```diff
@@ -6,7 +6,6 @@
  * @prop --text-editor-fade-out-background-color: the color of the fade-out effect at the top and bottom of the text editor, when the text-editor is in readonly state. Defaults to rgb(var(--contrast-100)).
  */
 
-@include shared_input-select-picker.lime-looks-like-input-value;
 .lime-looks-like-input-value {
     padding: var(--limel-text-editor-padding);
```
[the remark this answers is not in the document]
LucyChyzhova: in lime admin: some new hovering effect. intentionally? 

https://github.com/user-attachments/assets/35e4ccc1-e70f-4f4d-88a5-49be210df5da

src/components/chip-set/chip-set.scss
Kiarokh: 🤗 I love your review @LucyChyzhova ! Thanks for checking all these details and finding these issues. 
I'm gonna add fixups

Kiarokh: https://github.com/user-attachments/assets/48513da3-fcf0-4a45-b879-89fa759e274c

on my machine it clears the thing they have typed. 🤔 which I kind of like
maybe we should leave it as is :😆

LucyChyzhova: it's because you have a check for `Empty input on blur`

src/components/text-editor/text-editor.scss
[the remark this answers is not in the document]
Kiarokh: i couldn't figure out why lime admins does this. but fixed the first issue in the feed. I think we can ignore the admin page for now

src/components/chip-set/chip-set.scss
LucyChyzhova: looks fine in web-client though. Let's have it as an improvement 🥰
Screenshot 2025-04-10 at 13 55 38

LucyChyzhova:
I think I tested everything what I could. Thanks for resolving my comments Kia. 🤗  Looks good now. But I'm not sure if I'm the right person to approve it. Tell me how we should proceed with this PR 😉

src/components/chip-set/chip-set.scss
LucyChyzhova: i'm not sure if you decided to fix it or not 😀

LucyChyzhova (dismissed):
great job Kia! ❤️ Approving! 
just lint and build checks failed.

src/components/chip-set/chip-set.scss
Kiarokh: This new one is the "correct" placement of it 😊. So in case a readonly with `type` of `input` (like in the web client) is placed side-by-side of a `type` of `choice` for example, then the alignment of their labels and content will be correct.

src/components/chip-set/chip-set.scss
Kiarokh: ⚡ yes, fixed and pushed ;)

src/components/notched-outline/examples/notched-outline-basic.scss
```diff
@@ -0,0 +1,22 @@
...
+    // sake of this example.
+    border: none;
+    outline: none;
+
+    height: 2.5rem;
+    width: 100%;
+    padding: 0 1rem;
+
+    background-color: transparent;
+}
```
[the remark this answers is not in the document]
Kiarokh: @coderabbitai the examples in Lime Elements are already scoped. They all have `shadow: true,` and there is no reason to further scope them like your suggestion. That makes the styles more complicated, harder to read for consumers, and unnecessarily specified.

LucyChyzhova (dismissed):

src/components/chip-set/chip-set.tsx
```diff
@@ -337,35 +339,60 @@ export class ChipSet {
...
+                hasLeadingIcon={!!this.leadingIcon}
+                hasFloatingLabel={this.floatLabelAbove()}
+                onClick={this.handleTextFieldFocus}
+            >
+                <div slot="content" {...contentProps} class={classes}>
+                    {this.renderContent(value)}
+                </div>
+            </limel-notched-outline>,
+            this.renderHelperLine(),
+        ];
```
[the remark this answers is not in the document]
Kiarokh: @coderabbitai 🤔 oh actually I think this onClick should rather be on the `<div slot="content"` and be conditionally shown, only if the `type` is `input`. However, I don't think a corresponding keyboard event is needed, since users can tab into the `<input />` element which is within the div, and set the focus on it

LucyChyzhova (approved):

jgroth (approved):

adrianschmidt (approved):
Changes to external API approved 👍

adrianschmidt:
Ninja'd again 😆

src/components/chip-set/chip-set.tsx
```diff
@@ -385,113 +412,42 @@
...
+                // that event from propagating since we are emitting our own change event
+                onChange={this.inputFieldOnChange}
+                placeholder={this.isFull() ? '' : this.searchLabel}
+                readonly={this.isFull()}
+                autocomplete={this.autocomplete}
+            />,
+            this.renderLeadingIcon(),
+            this.renderClearAllChipsButton(),
+        ];
     }
```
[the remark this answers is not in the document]
adrianschmidt: @coderabbitai @Kiarokh This isn't about the keys in the chips (held in `this.value` and rendered by `this.renderInputChip`), this is about the array wrapping the outermost level of elements; the `return […]`. Since both CodeRabbit and SonarQube will complain about the lack of `key` properties on the elements in the array when this pattern is used, we should stop using it, and instead use `return (<Host>…</Host>)`. If you could switch to this pattern @Kiarokh, I might not have to bug CodeRabbit about it all the time 😅

PR #3521, opened 2025-04-08
fix(action-bar): restrict collapse button to floating layout only

LucyChyzhova:
## Summary by CodeRabbit

- **Bug Fixes**
	- Adjusted the action bar so that the collapse/expand control will no longer appear when using a floating layout. This change refines the interface, ensuring the control only displays when applicable based on the current design configuration.
  
- **New Features**
	- Updated documentation to reflect changes in component functionality from "expand" to "collapse" for the action bar, enhancing clarity on usage.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/action-bar/action-bar.tsx
+++ b/src/components/action-bar/action-bar.tsx
@@ -209,7 +209,11 @@ export class ActionBar {
     };
 
     private renderCollapseExpandButton() {
-        if (!this.collapsible || this.actions.length <= 1) {
+        if (
+            !this.collapsible ||
+            this.actions.length <= 1 ||
+            this.layout !== 'floating'
+        ) {
             return;
         }
 
--- a/src/components/action-bar/action-bar.tsx
+++ b/src/components/action-bar/action-bar.tsx
@@ -39,7 +39,7 @@ import { Icon } from '../../global/shared-types/icon.types';
  * @exampleComponent limel-example-action-bar-selected-item
  * @exampleComponent limel-example-action-bar-colors
  * @exampleComponent limel-example-action-bar-floating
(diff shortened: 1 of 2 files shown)


src/components/action-bar/action-bar.tsx
```diff
@@ -209,7 +209,11 @@ export class ActionBar {
     };
 
     private renderCollapseExpandButton() {
-        if (!this.collapsible || this.actions.length <= 1) {
+        if (
+            !this.collapsible ||
+            this.actions.length <= 1 ||
+            this.layout !== 'floating'
+        ) {
```
Kiarokh, line 216: I feel a bit unsure about having a prop, that only works when another prop is set to a certain state. This is what I was trying to communicate in the other issue as one of the reasons for reconsidering this design choice. 

If you wanna keep this as a temporary solution for now, please update the documentations as well as documenting the prop itself as alpha and even perhaps private.

Like below:
```tsx
    /**
     * When set to `true`, the action bar will be collapsible,
     * but only when the `layout` prop is also set to `floating`.
     * 
     * @private
     * @alpha
     */
    @Prop({ reflect: true })
    public collapsible = false;
```

As you see, the updated docs feels a bit strange. But if we make it `alpha`, it feels safer later to revert the changes.

Kiarokh:
closing in favor of https://github.com/Lundalogik/crm-client/issues/936

Issue #3537, opened 2025-04-23
Select: if an empty option which has a `text` label is chosen, the field's label will overlap it.


Kiarokh:
## Current behavior
`limel-select` allows users to select an empty option. This is to provide a way to override the native select behavior, where you cannot "unset" or deselect a selected choice.

Currently, if an empty option which has a `text` label is chosen, the field's label will overlap it.
<img width="1281" alt="Image" src="https://github.com/user-attachments/assets/9c2590c2-1573-410a-ab8b-080001912df4" />

## Expected behavior

When there is an empty option like `{ text: 'All', value: '' },` in this array: 

```tsx
    private options: Option[] = [
        { text: 'All', value: '' },
        { text: 'Luke Skywalker', value: 'luke' },
        { text: 'Han Solo', value: 'han' },
        { text: 'Leia Organo', value: 'leia' },
    ];
```

The user should be able to select the empty choice that has no `value`. 
1. If that empty option is selected and if the option has a `text`, the `limel-notched-outline` should get a floating label.
2. If there is no `value` and no `text`, then the label should not float above.

>[!note]
>In previous implementation, the label would float above, even when you had chosen the empty option. This _according to me_ is not good design. When the empty option is selected (which could also mean that the previously selected choice is now deselected), the select should be simply looking like its default mode, when nothing is selected.
> Note that when the field is required, the empty option is not displayed in the list of options. So users cannot continue, without selecting anything. This was fixed a while ago: https://github.com/Lundalogik/lime-elements/issues/1982 
>
> However, when the field is not required, we are simply providing a way for the user to unset their choices.

## Background stories
I think we tried to solve this issue once before: https://github.com/Lundalogik/crm-feature/issues/1839 but I don't remember where we ended up honestly. I think we didn't really change that behavior where the empty selected option was still visualized like below:

<img width="45%" alt="image" src="https://user-images.githubusercontent.com/35954987/201608317-f82d5753-0d99-4b6f-b641-fe1ad457fa3e.png" style="vertical-align: top;">  <img width="45%" alt="image" src="https://user-images.githubusercontent.com/35954987/201608201-e4c42b41-3ccf-46f8-b642-f71b356a1db2.png">

Now with new changes in implementing `limel-notch-outline` https://github.com/Lundalogik/lime-elements/pull/3518 we seemingly fixed the floating text problem for the empty field, but introduced a new issue. 

[...]

PR #3538, opened 2025-04-23
fix(select): ensure label floats, when a labeled empty option is selected

Kiarokh:
… and that the label does not float,
when an unlabeled empty option is selected.
fix https://github.com/Lundalogik/lime-elements/issues/3537

## Summary by CodeRabbit

- **New Features**
	- Enhanced the select input example to showcase two methods for handling empty options: an unlabeled empty option and a labeled "None" option, each with its own state and options.
- **Documentation**
	- Expanded example documentation to clarify the approaches to empty options and their behavior when the select is required.
- **Style**
	- Improved label floating behavior in select inputs, ensuring the label floats based on both open state and whether a meaningful value is selected.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/select/examples/select-with-empty-option.tsx
+++ b/src/components/select/examples/select-with-empty-option.tsx
@@ -5,10 +5,43 @@ import { Component, h, State } from '@stencil/core';
  * With Empty Option
  *
  * Adding an empty option makes it possible for the user to "unset"
- * the value. Try selecting a value below, and then selecting the empty
- * option again.
+ * any chosen value.
  *
- * If the component is set as required, the empty option is removed.
+ * This example demonstrates two different approaches to empty options in select components:
+ *
+ * **1. Unlabeled empty option**:
+ * A select with a completely empty option (both `text` and `value` are empty).
+ *
+ * Provides a way to clear a selection without any visible text.
+ * This works well when the select's purpose is clear from context
+ * and "no selection" feels intuitive.
+ *
+ * **2. Labeled empty option**:
+ * A select with a labeled empty option (using a label as `text`,
+ * but with an empty `value`).
+ *
+ * This approach uses descriptive labels like "All", "None", while still
+ * having an empty `value`. When good labels are used in the right context,
+ * this improves clarity for users by explicitly communicating what an empty
+ * selection means in this specific context.
+ *
+ * Try selecting a value below, and then selecting the empty
+ * option again. Notice how the select's UI resets to an empty default state,
+ * if no `text` is provided to label the empty option.
+ *
+ * :::important
+ * If the component is set as `required`, and the empty option is unlabeled,
+ * it will be removed from the list.
+ * This is to ensure that if users open the required dropdown,
+ * and close it without selecting anything, the component can make set
+ * itself to `invalid`, forcing the user to choose a choice,
+ * before being able to continue.
+ *
+ * However, when the empty option is labeled, it will remain in the list,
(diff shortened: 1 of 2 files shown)


src/components/select/examples/select-with-empty-option.tsx
```diff
@@ -30,7 +30,17 @@ import { Component, h, State } from '@stencil/core';
...
+ * it will be removed from the list.
+ * This is to ensure that if users open the required dropdown,
+ * and close it without selecting anything, the component can make set
+ * itself to `invalid`, forcing the user to choose a choice,
+ * before being able to continue.
+ *
+ * However, when the empty option is labeled, it will remain in the list,
+ * even if the component is set as `required`. This can be confusing for users,
+ * as they may not understand why the empty option is still available and selectable,
+ * but once selected, the component becomes `invalid`.
```
Kiarokh, line 43: It's funny. 

Why I did this PR is because @anderssonjohan reported a case which I didn't know consumers are doing.; which is having an optional `select`, with an empty but labeled option, [example here](https://github.com/Lundalogik/lime-crm-components/blob/b9542bbbc4745882c61ccf38e91ab04f22187c90/src/admin/ai-assistant/gateway-log/ai-gateway-log.tsx#L75-L118). 

To ensure that we will not miss such cases in future, I updated this example file and added a `limel-select` with a labeled empty option for testing. I also added a bunch of docs to explain things for consumers.

While testing, I faced this situation which could allow a user to select an empty option (with `text` but no `value`) 👇 :

image
_(I have a vague memory that this has been reported as a bug before…)_

Then after seeing this scenario, I added this part of the docs above ☝️ to make sure consumers won't do mistakes. 

Then I thought, wait a second, this sounds really stupid. Why do we show the empty option anyway? We shouldn't render the empty choice for the required option anyways. It doesn't matter if it is labeled or not. Let me fix it.

When trying to fix it, I saw this comment:
https://github.com/Lundalogik/lime-elements/blob/f8cdcea6afbd586aa27ef790dee813b21917c994/src/components/select/select.template.tsx#L307

> If the select component IS required, we keep only options that are not "empty". We only check the text property, because some systems use an "empty option" that will have a value for the `value` property, to signify "no option selected". Such an option should be treated as "empty".


This is where I gave up fixing this. 

This is not a part of this issue and has existed for a long time, since we [did this](https://github.com/Lundalogik/lime-elements/pull/2007) I believe. So we don't have to fix it here and now. But I still think it's weird. Do you remember anything about this comment @Befkadu1 @adrianschmidt?

Are we fine with just explaining the behavior in the docs, or do we need to do something about this choice?

anderssonjohan: @Kiarokh Yes, you are correct. The filters you highlighted (+ the third limel-select) have the first option set to `{ text: 'All', value: '' }`. 

Another example is this:

Screenshot 2025-04-23 at 17 23 16

The default option/empty value for this search filter is not "All" but "Successful or Failed".
These labels describes the default filter and is neither defining "Is empty" or "everything". The default filter may however be "show all", but it's not necessarily true for all cases.

anderssonjohan (approved):

anderssonjohan:
Works as expected in the PR release/docs! 🎉

PR #3545, opened 2025-04-29
Button: use the `Icon` interface

Kiarokh:
## Summary by CodeRabbit

- **New Features**
	- Button icons now support enhanced accessibility with explicit accessibility attributes and support for passing icon objects with customizable color.
- **Documentation**
	- Expanded button icon examples to illustrate best practices for accessibility, including proper use of icon titles and tooltips for screen reader support.
- **Style**
	- Updated button icon styles by shifting icon sizing to specific elements.
	- Simplified grid layouts in example components with adjusted column sizing and removal of variable column layout.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -152,7 +152,7 @@ export class Button {
             return;
         }
 
-        return <limel-icon class="icon" name={this.icon} />;
+        return <limel-icon name={this.icon} />;
     }
 
     private renderLabel() {
--- a/src/components/button/button.tsx
+++ b/src/components/button/button.tsx
@@ -3,6 +3,8 @@ import {
     makeEnterClickable,
     removeEnterClickable,
 } from '../../util/make-enter-clickable';
+import { Icon } from '../../global/shared-types/icon.types';
+import { getIconName, getIconTitle } from '../icon/get-icon-props';
 
 /**
  * @exampleComponent limel-example-button-basic
@@ -45,7 +47,7 @@ export class Button {
      * Set icon for the button
      */
     @Prop({ reflect: true })
-    public icon: string;
+    public icon: string | Icon;
 
     /**
      * Set to `true` to disable the button.
@@ -97,7 +99,7 @@ export class Button {
                     aria-busy={this.loading ? 'true' : 'false'}
                     aria-live="polite"
                 >
-                    {this.renderIcon()}
+                    {this.renderIcon(this.icon)}
                     {this.renderLabel()}
                     {this.renderSpinner()}
                     <svg viewBox="0 0 30 30">{this.renderLoadingIcons()}</svg>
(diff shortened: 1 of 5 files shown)


src/components/button/button.tsx
```diff
@@ -45,7 +47,7 @@ export class Button {
      * Set icon for the button
      */
     @Prop({ reflect: true })
-    public icon: string;
+    public icon: string | Omit<Icon, 'backgroundColor'>;
```
adrianschmidt, line 50: This seems unnecessarily annoying… 🤔 I think this means that if you have a variable or property typed to `Icon`, you can't pass that into this property without either casting it, or proving to the typescript compiler that the object doesn't contain a property called `backgroundColor`.

Also, this means that rather than the docs looking like this:
image

…they look like this:
image

I would suggest that we instead document that any `backgroundColor` value will be ignored.

src/components/button/button.tsx
```diff
@@ -147,12 +149,41 @@ export class Button {
...
+                'Please add either a `label`, a `title` to the `icon`, or use a `limel-tooltip`. ' +
+                'See https://lundalogik.github.io/lime-elements/versions/latest/#/component/limel-button/ ' +
+                'for more information.';
+            // eslint-disable-next-line no-console
+            console.warn(WARNING_MESSAGE);
+        }
+
+        if (typeof icon === 'object') {
+            iconColor = icon.color;
+        }
```
adrianschmidt, line 174:
```suggestion
        const title = getIconTitle(icon);

        if (!this.label && !title) {
            const WARNING_MESSAGE =
                '⚠️ Accessibility warning: `limel-button` has no `label` and its `icon` has no `title`. ' +
                'This creates an inaccessible button for screen reader users. ' +
                'Please add either a `label`, a `title` to the `icon`, or use a `limel-tooltip`. ' +
                'See https://lundalogik.github.io/lime-elements/versions/latest/#/component/limel-button/ ' +
                'for more information.';
            // eslint-disable-next-line no-console
            console.warn(WARNING_MESSAGE);
        }

        let iconColor: string | undefined;

        if (typeof icon === 'object') {
            iconColor = icon.color;
        }
```

(I've moved the `let iconColor: string | undefined;` down to just before the variable is used.)

src/components/button/button.tsx
Kiarokh: Aha. Ok. Makes sense. I intentionally didn't make docs about how the iconColor can be used. I have some ideas regarding that, and generally don't want to showcase too much coloring possibilities. So we can perhaps for now not document anything about colors, and not use `backgroundColor` in the code neither. What do you think?

Kiarokh: ⚡

adrianschmidt (approved):

PR #3552, opened 2025-05-09
fix(text-editor): prevent Excel tables from being processed as images when pasted

FredrikWallstrom:
Fixes https://github.com/Lundalogik/lime-elements/issues/3512
Fixes https://github.com/Lundalogik/limepkg-email/issues/1556

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of pasted content to prevent Excel tables from being incorrectly treated as images during paste events in the text editor.
  - Enhanced image paste functionality to better detect and process image files from the clipboard while filtering out unwanted image nodes.

## Reproduce

* Images should be pasted as images.
* Tables copied from Excel should not be pasted as images.
* If you format your Cells as a Table in Excel, they should include CSS styling.
* [If you only copy *Cells* from Excel, the will lack CSS styling.](https://github.com/Lundalogik/crm-client/issues/51)

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/text-editor/prosemirror-adapter/plugins/image/inserter.ts
+++ b/src/components/text-editor/prosemirror-adapter/plugins/image/inserter.ts
@@ -194,9 +194,43 @@ const processPasteEvent = (
         return false;
     }
 
+    const isImageFilePasted = handlePastedImages(view, clipboardData);
+
+    const filteredSlice = new Slice(
+        filterImageNodes(slice.content),
+        slice.openStart,
+        slice.openEnd,
+    );
+
+    if (filteredSlice.content.childCount < slice.content.childCount) {
+        const { state, dispatch } = view;
+        const tr = state.tr.replaceSelection(filteredSlice);
+        dispatch(tr);
+
+        return true;
+    }
+
+    return isImageFilePasted;
+};
+
+/**
+ * Processes any image files found in the clipboard data and dispatches an imagePasted event.
+ *
+ * @param view - The ProseMirror editor view
+ * @param clipboardData - The clipboard data transfer object containing potential image files
+ * @returns True if at least one valid image file was found and processed, false otherwise
+ */
+function handlePastedImages(
+    view: EditorView,
+    clipboardData: DataTransfer,
+): boolean {
+    let isImageFilePasted = false;
     const files = Array.from(clipboardData.files || []);
+
     for (const file of files) {
-        if (file.type.startsWith('image/')) {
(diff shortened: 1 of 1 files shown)


src/components/text-editor/prosemirror-adapter/plugins/image/inserter.ts
```diff
@@ -214,19 +248,28 @@ const processPasteEvent = (
...
+function isImageFile(file: File, clipboardData: DataTransfer): boolean {
+    const isContentTypeImage = file.type.startsWith('image/');
+    if (!isContentTypeImage) {
+        return false;
     }
 
-    return files.length > 0;
-};
+    const html = clipboardData?.getData('text/html')?.toLowerCase() ?? '';
+    const isProbablyFromExcel = html.includes('<table');
```
john-traas, line 272: I had a thought about whether or not there's something more explicit we can check to determine if what is pasted originated from Excel, I imagine there might be some sort of <meta> tag in what is copied to the clipBoard. But this works and maybe we can update it in future.

john-traas (dismissed):
Nice! It looks good and worked well when tested.

src/components/text-editor/prosemirror-adapter/plugins/image/inserter.ts
FredrikWallstrom: Yeah that would be lovely. I didn't find such a flag/tag, did you check as well? Maybe I missed something.

john-traas: I tried this

```
function isHtmlFromExcel(html: string): boolean {
    if (!html) {
        return false;
    }

    return (
        html.includes('name=generator content="microsoft excel"') ||
        html.includes('xmlns:x="urn:schemas-microsoft-com:office:excel"')
    );
}
```

and in the `isImageFile` function just return `!isHtmlFromExcel(html)`;

That worked detecting content I pasted from excel. But now I had another thought, do we explicitly only want to allow tables from excel? Maybe just checking for the `<table` start tag is enough for our use case and it might not always originate from excel.

john-traas (approved):
Looks good and works well!

Issue #3561, opened 2025-05-20
Fix any remote command injection vulnerabilities in GitHub Actions workflows


adrianschmidt:
Quote from [original post in Slack](https://lime-technologies.slack.com/archives/C1HFBM67J/p1746539371623149):

> If you code workflows, remember to never use variable interpolation directly into scripts (shell, python, javascript, etc.)
> A short example of this is covered by this article and what to do instead: https://www.kenmuse.com/blog/github-actions-injection-attacks/
>
> Here's an example from a very old version of one of our workflows:
> https://github.com/Lundalogik/solution-sveriges-allmannytta/blob/6f6264093b2deea79f131f6bb1b529343c81e7ec/workflows/ci.yml#L33
>
> This is not about inputs though, but anyone gaining admin access to that repo could potentially use https://cli.github.com/manual/gh_secret_set to inject python code into the secret and execute that code. This is of course a stretch, because if you're admin and can overrule every single policy, then you can also modify the workflow itself.
> But even though this context would be safer to use with interpolation in workflow steps, it will still be less secure than passing the secret as an env variable.
>
> Using the interpolation will actually render the Python code with the secret like this:
> ```
> if not bool("fru1ty"):
> ```
> ..but if you always be consistent with passing all script inputs using the env vars, then the code would not change (and your secret can contain a double quote 😏 :
> ```
>         shell: python
>         env:
>           PYPI_PASSWORD: ${{secrets.PYPI_PASSWORD}}
>         run: |
>           import sys
>           import os
>           if not bool(os.environ["PYPI_PASSWORD"]):
> ```
>
> Tip: To avoid this kind of injection with Python CLI programs, using `envvar` with click options make it very easy and safe to use both interactively and with workflows. [Example](https://github.com/Lundalogik/lime-project/blob/2fd90f16a7ab203adf30cb5f6f98ace8753e79f7/lime_project/cli/commands/deploy.py#L459)

Check all workflows in the `.github/workflows` folder, and update as required.

PR #3564, opened 2025-05-22
ci(workflows): use environment variables in scripts to prevent command injection

adrianschmidt:
fix: #3561

@coderabbitai summary

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


publish-docs.js
```diff
@@ -385,19 +380,9 @@ function commit(message) {
...
-
-    if (argv.authorEmail) {
-        shell.echo('setting git config user.email');
-        shell.exec(`git config --local user.email '${argv.authorEmail}'`);
-    }
-
     shell.exec('git add -A --ignore-errors');
 
-    if (shell.exec(`git commit -m "${message}"`).code !== 0) {
+    if (shell.exec('git commit -m "$COMMIT_MESSAGE"', {env: {...process.env, COMMIT_MESSAGE: message}}).code !== 0) {
```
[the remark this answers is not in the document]
adrianschmidt: Gemini says the following in response to the above comment:

> The suggestion is to use an array-based exec call if shelljs supports it, or ensure proper escaping. shelljs.exec() does not directly support an array form like child_process.execFile in Node.js. The current approach of passing the message via an environment variable (COMMIT_MESSAGE) is already a robust way to handle special characters and quotes safely, as the shell will substitute the variable's content literally without re-interpreting it as part of the command structure. So, the existing fix is sound for shelljs.

.github/workflows/create-backport-branch.yml
adrianschmidt, line 1: This workflow is not in active use. It can be run manually, but would need some updating to be more user friendly.

Successful test run here: https://github.com/Lundalogik/lime-elements/actions/runs/15205209129/job/42766624340

anderssonjohan (approved):
LGTM!

.github/actions/upload-frontend-build-output-to-s3/action.yml
adrianschmidt, line 1: Successful run here: https://github.com/Lundalogik/lime-elements/actions/runs/15205375226/job/42767152936#step:9:1

Uploaded result here: https://cdn.lime-crm.com/packages/@limetech/lime-elements/38.13.2-dev.1/dist/lime-elements/lime-elements.esm.js

.github/workflows/pr-checks.yml
adrianschmidt, line 1: This workflow has obviously been run for this PR 😄

.github/workflows/publish-docs.yml
adrianschmidt, line 1: This workflow has also been run for this PR.

publish-docs.js
adrianschmidt, line 1: This script is called by the above workflow, so it has been run for this PR.

Result available here (until it's been cleaned up by `cleanup.yml` of course): https://lundalogik.github.io/lime-elements/versions/PR-3564/#/

anderssonjohan (approved):

anderssonjohan (approved):

PR #3577, opened 2025-06-09
fix(menu-list-render): show primary text on hover

LucyChyzhova:
If a list item is too long and gets truncated, it's impossible to read the full text. 
We can use this variable ` --menu-surface-width`  to make list wider, but we never know how many actions could be used and what title of them. 
. 
before:
Screenshot 2025-06-09 at 14 13 04
after:
Screenshot 2025-06-09 at 14 13 11

@coderabbitai summary

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu-list/menu-list-renderer.tsx
+++ b/src/components/menu-list/menu-list-renderer.tsx
@@ -169,14 +169,19 @@ export class MenuListRenderer {
     private renderText = (item: MenuItem) => {
         if (this.isSimpleItem(item)) {
             return (
-                <span class="mdc-deprecated-list-item__text">{item.text}</span>
+                <span class="mdc-deprecated-list-item__text" title={item.text}>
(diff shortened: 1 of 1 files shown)


Kiarokh (changes requested):
Using the `title` attribute is not good enough.

There is already a feature for this, implemented in `limel-menu`. We should just implement the same by using that prop (`surfaceWidth`) in `limel-split-button` as well.

image

LucyChyzhova:
related pr https://github.com/Lundalogik/limepkg-email/pull/1613

PR #3581, opened 2025-06-11
feat: menuitems to file chips

viktorSoftDev:
fix: #3583 

To be able to add commands in the file chips

image

## Summary by CodeRabbit

* **New Features**
  * Introduced a new example component demonstrating file input with custom menu items like "Download" and "Share".
  * Added support for displaying and interacting with custom menu items on file inputs.

* **Documentation**
  * Updated documentation examples to showcase the new file menu items feature.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -882,6 +882,7 @@ export interface FileInfo {
     iconColor?: Color;
     id: number | string;
     lastModified?: Date;
+    menuItems?: Array<MenuItem | ListSeparator>;
     size?: number;
 }
 
--- a/src/components/file/examples/file-menu-items.tsx
+++ b/src/components/file/examples/file-menu-items.tsx
@@ -1,0 +1,67 @@
+import {
+    FileInfo,
+    LimelMenuCustomEvent,
+    MenuItem,
+} from '@limetech/lime-elements';
+import { Component, h, Host, State } from '@stencil/core';
+/**
+ * Custom menu items
+ *
+ * By providing custom menu items, you can add additional actions.
+ *
+ */
+@Component({
+    tag: 'limel-example-file-menu-items',
+    shadow: true,
+})
+export class FileMenuItemsExample {
+    @State()
+    private value: FileInfo = {
+        filename: 'deal.pdf',
+        id: 123,
+        menuItems: [
+            {
+                text: 'Download',
+                icon: 'download',
+                value: 1,
+                selected: false,
+            },
+            {
+                text: 'Share',
+                icon: 'share',
+                value: 2,
+                selected: false,
+            },
+        ],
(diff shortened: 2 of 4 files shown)


src/components/chip-set/chip-set.tsx
```diff
@@ -617,6 +623,7 @@ export class ChipSet {
             type: chipType,
             removable: removable,
             menuItems: chip.menuItems,
+            onMenuItemSelected: this.handleMenuItemSelected(chip),
```
Kiarokh, line 626: image

Just wonder, aren't we already using this menu items in the web-client ☝️ , without these new changes? How is that working there? 🤔 

We should always add examples to the docs pages, for each new feature that we add. I think we've missed adding it for this particular feature of the chips, in the chip-set. But if something is different in this PR, I suggest that we add an example of this fresh feature, and explain its usage.

Kiarokh: @coderabbitai please remind future contributors of this repository who add new props to existing components, to consider adding docs examples, featuring the newly added prop and explaining how it is used.

src/components/chip-set/chip.types.ts
```diff
@@ -115,3 +115,8 @@ export interface Chip<T = any> {
  * @beta
  */
 export type ChipType = 'default' | 'filter';
+
+export interface MenuItemEvent {
+    chip: Chip;
+    menuItem: MenuItem;
+}
```
adrianschmidt, line 122: This doesn't seem to define an event, but an event payload.

src/components/chip-set/chip-set.tsx
adrianschmidt: Looking at the `Chip` type, it already has support for a property called `menuItems`: https://lundalogik.github.io/lime-elements/versions/latest/#/type/Chip/

Similarly, the `ListItem` type supports a property called `actions`, which amounts to the same thing: https://lundalogik.github.io/lime-elements/versions/latest/#/type/ListItem/

viktorSoftDev: Went with extending the FileInfo interface.

Kiarokh: I don't have a good overview, but I think the implementation should be similar to `limel-chip-set` if possible.

viktorSoftDev:
@adrianschmidt Are you the right person / do you have time to help me review this? 🙂

PR #3616, opened 2025-07-16
fix(popover): ensure content is added to portal container

jgroth:
When iterating over the HTMLCollection, elements were removed from the collection during iteration when appended to the portal container. This caused elements to be skipped and never added to the container

fix Lundalogik/crm-client#184

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for the PopoverSurface component to verify rendering with various child element configurations.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/popover-surface/__snapshots__/popover-surface.spec.tsx.snap
+++ b/src/components/popover-surface/__snapshots__/popover-surface.spec.tsx.snap
@@ -1,0 +1,44 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`render component with one child 1`] = `
+<limel-popover-surface>
+  <mock:shadow-root>
+    <div class="limel-popover-surface" tabindex="0">
+      <div>
+        Child 1
+      </div>
+    </div>
+  </mock:shadow-root>
+</limel-popover-surface>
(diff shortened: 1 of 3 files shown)


src/components/popover-surface/__snapshots__/popover-surface.spec.tsx.snap
FredrikWallstrom, line 1: Cool, I didn't know about Snapshot Testing. I have two questions.

I assume you have generated these snapshots manually? I read something that you can automatically update your snapshots with `jest --updateSnapshot`, I tried that but I got a bunch of errors 😄 Is it possible to generate or update these automatically?

When would you say it is useful to us Snapshot Testing? I can imagine it is not that easy to follow if the only thing you are doing is Snapshot Testing and you have quite complex components.

FredrikWallstrom (approved):

Issue #3632, opened 2025-08-05
checkbox: get rid of MDC styles


Kiarokh:
Step 1 of 2

Read more on the parent issue

PR #3634, opened 2025-08-05
Refactor checkbox component to remove MDC dependencies and update styles

Kiarokh:
As a first step of getting rid of MDC, we need to make sure that the component renders correctly, without relying on any styles applied by MDC.

fix https://github.com/Lundalogik/lime-elements/issues/3632

## Summary by CodeRabbit

* **Refactor**
  * Checkbox component styles were fully rewritten to use custom CSS, removing all dependencies on Material Design Components. Visual appearance and interaction states have been improved and are now fully custom.
  * Checkbox template markup was simplified, flattening the structure and centralizing state-based class management.

* **Style**
  * Improved checkbox alignment, sizing, and transitions for a smoother user experience.
  * Adjusted dynamic label spacing for a more compact layout.
  * Updated table row selector and list item styles for better checkbox integration.

* **Chores**
  * Removed unused helper text styles related to the checkbox component.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/partial-styles/_row-selection.scss
+++ b/src/components/table/partial-styles/_row-selection.scss
@@ -6,9 +6,7 @@ $checkbox-distance-to-left-edge: 0.5rem;
 
 .select-all,
 .limel-table--row-selector {
-    // prevent the checkbox affecting the row height
-    --mdc-checkbox-touch-target-size: 1.125rem;
-    --mdc-checkbox-ripple-size: 2.5rem;
+    --limel-checkbox-min-height: 1.25rem; // as big as the box itself
 }
 
 .select-all {
@@ -26,6 +24,7 @@ $checkbox-distance-to-left-edge: 0.5rem;
 
     text-overflow: unset !important; // Otherwise a `…` will be rendered besides the checkbox, since we truncate all cells
     padding-left: $checkbox-distance-to-left-edge !important;
+    padding-top: 0.25rem !important;
 
     limel-checkbox {
         margin-top: functions.pxToRem(1);
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -1,110 +1,212 @@
-@use '@material/checkbox';
-@use '@material/form-field';
+@use '../../style/mixins';
 
 /**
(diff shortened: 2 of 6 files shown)


src/components/table/partial-styles/_row-selection.scss
```diff
@@ -6,9 +6,7 @@ $checkbox-distance-to-left-edge: 0.5rem;
 
 .select-all,
```
LucyChyzhova, line 7: hm, i can't select all by clicking the top check box anymore

https://github.com/user-attachments/assets/bc13235a-084a-4111-83aa-29ca978fb9bd

Kiarokh: limel-checkbox is used the table component in 2 places. One is in the header row, to display a "select all" checkbox, and one in each individual row, for selecting individual rows.

In the current PR I removed MDC's styles (along with some classes) from `limel-checkbox` component. This change has made the "select all" checkbox unclickable. It is still possible to tab-focus the checkbox and toggle its state using keyboard. But clicking doesn't work.

When I inspect these two instances of `limel-checkbox` in `limel-table`, the only difference I see is presence of a (empty) `<label>` element on the other row checkboxes. But the one in the header doesn't have a label. I believe that is the reason is for the checkbox not being clickable. It seems that the `label` element is the only clickable part of the native checkbox element that I am using in CheckboxTemplate. The actual `<input type="checkbox"` element is hidden using CSS (due to styling needs). The `<div class="box"`  that visualizes the checkbox in a custom way is actually not clickable at all, both due to `pointer-events: none;` and also even without that style, clicking it doesn't do anything anyways. 

I'm currently investigating to see how to best solve this issue

LucyChyzhova: great! 💫  I will test once more

LucyChyzhova (approved):
great job, Kia! 👏

Issue #1437, opened 2021-11-15
New component - Radio button


TommyLindh2:
## New feature motivation

It is a common web element that is used in forms and similiar scenarios.
It just makes sense to have it

Plus we want it in the dynamic checklist addon

## New feature description

Something like [this](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/radio)

adrianschmidt:
Yeah, I absolutely agree.

This might be simple too, because we already have a radio-button template that is used by limel-list. It's not a full component, but I'd guess that converting it to one shouldn't be too difficult…

blombergedvin:
Totally agree! 
However, not something I feel we need to address the current cycle, rather focus on the issues we have. 
I Will put it into the cooldown board, since it feels like a good cooldown issue.

Kiarokh:
This will be a great addition. I've seen consultants using a checkbox and hacking it work like a radio button (so user can only select 1 choice). This could be very confusing for some users and can results in a really terrible user experience.

Kiarokh:
@adrianschmidt @jgroth 

Me and @LucyChyzhova started making this as a new limel-component, but eventually came to this conclusion that it'll be nice to offer the radio button not as a single and stand-alone component, but wrapped into a list of radio buttons. Just like the Button Group component, which is in its core a radio button list too.

The reason is that:
1. radio buttons can NEVER be 1, and they should be in a list with at least 2 radios.
2. radio buttons should have a `name` tag, which should be shared between them all (unlike checkboxes, or switches), 
2. and just like the `id`, the `name` should also be auto-generated for the consumer to do less work
2. it should b easy for consumers to send an array of labels and the list of radio buttons get generated for them (and they shouldn't care about putting one radio per row for instance, or adding `<br>`s after each button)


So, with all that said, it feels we already have this feature 😂 and that's the limel-list with radio buttons!


I think what @joelricci or similar folks need is having `limel-form` and schema support the limel-list with radio buttons! so that they can incorporate them in forms. 

What do you think?

adrianschmidt:
Yeah, sounds like a very reasonable conclusion to me 👍

Kiarokh:
what do you say @joelricci ?

Kiarokh:
And I think, in the "form" context, usually there is a question before the radio button or checkbox list. And that would be the `label` right?

Example:

What's your favorite fruit?
- [ ] Apple
- [ ] Samsung
- [ ] Huawei 

[...]

PR #3638, opened 2025-08-07
@coderabbitai

Kiarokh:
fix https://github.com/Lundalogik/lime-elements/issues/1437

@coderabbitai summary

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -0,0 +1,119 @@
...
+ * multiple choices, making them ideal for exclusive selections.
+ *
+ * :::note
+ * A single radio button is never useful in a UI. Radio buttons should always come in groups
+ * of at least 2 options where only one can be selected at a time.
+ * :::
+ *
+ * @exampleComponent limel-example-radio-button-group-basic
+ * @exampleComponent limel-example-radio-button-group-icons
+ * @exampleComponent limel-example-radio-button-group-multiple-lines
```
Kiarokh, line 27: I'd love to add some examples with striped rows (and perhaps those interactive styles) like what we have in `limel-list` 👇 
image
But since those are not applied using props, I can't. Or actually I can, but I'd rather not do it here, unless we add new props for these in `limel-list`. We can add these later in follow up PRs.

src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -0,0 +1,119 @@
...
+     * numbers we default to 1; and if they type decimals we round up.
+     */
+    @Prop({ reflect: true })
+    public maxLinesSecondaryText: number = 3;
+
+    /**
+     * Emitted when the selected value changes
+     */
+    @Event()
+    private change: EventEmitter<any>;
```
Befkadu1, line 72: no type here as well

src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -0,0 +1,119 @@
...
+     * Disables all radio buttons when `true`
+     */
+    @Prop({ reflect: true })
+    public disabled = false;
+
+    /**
+     * Set to `true` if the radio button group should display larger icons with a background
+     */
+    @Prop({ reflect: true })
+    public badgeIcons: boolean;
```
Befkadu1, line 57: I don't like this name, maybe would have been good to name it like `hasBadgeIcons` or something like that since it is a boolean. I saw this name in many places even in the interface `ListRendererConfig`, so we keep it like this. But I just thought that was not a name we chose from the beginning.

src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -0,0 +1,119 @@
...
+    private internalOptions: Array<ListItem> = [];
+
+    public componentWillLoad() {
+        this.updateInternalOptions();
+    }
+
+    @Watch('options')
+    @Watch('value')
+    @Watch('disabled')
+    @Watch('maxLinesSecondaryText')
```
Befkadu1, line 84: I don't understand why we watch all these props? In my world watching for the changes on `value` is enough

src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -0,0 +1,119 @@
...
+        this.internalOptions = this.options.map((option) => ({
+            ...option,
+            selected: option.value === this.value,
+            disabled: this.disabled || option.disabled,
+        }));
+    };
+
+    private handleChange = (event: LimelListCustomEvent<ListItem>) => {
+        event.stopPropagation();
+
```
Befkadu1, line 111: Maybe we need this as well
```
 if (!event.detail) {
            return;
        }
```

src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -0,0 +1,119 @@
...
+        this.updateInternalOptions();
+    }
+
+    @Watch('options')
+    @Watch('value')
+    @Watch('disabled')
+    @Watch('maxLinesSecondaryText')
+    protected handlePropsChange() {
+        this.updateInternalOptions();
+    }
```
Befkadu1, line 87: You can remove all these. We don't need to have a state for this we can simply have a function that you called it `updateInternalOptions` and return the items directly. Maybe call its name to `createListItems()`. You can use it like this
<limel-list
         items={this.createListItems()}
         type="radio"
         badgeIcons={this.badgeIcons}
         maxLinesSecondaryText={this.maxLinesSecondaryText}
         onChange={this.handleChange}
  />

src/components/radio-button-group/radio-button-group.tsx
Kiarokh: Yes, I totally agree with you. I just copied form existing naming. Since we're making just a wrapper for `limel-list`, it felt we could use the same name. But personally, I have no problems changing it.

src/components/radio-button-group/radio-button-group.tsx
Kiarokh: You’re right—since this is a thin wrapper around limel-list, we don’t need to watch all those props or keep internal state. I'll fix

src/components/radio-button-group/radio-button-group.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    public render() {
+        return (
+            <limel-list
```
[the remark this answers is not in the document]
Kiarokh: Update:

The "deselection behavior" depends on how we implement the `handleChange`.
So I'm adding an example to cover that case.

Befkadu1 (approved):
Awesome work ⚡

Issue #3679, opened 2025-09-19
Increase working area for code editor


paulinewahle:
## New feature motivation

https://product.lime-crm.com/client/object/feedback/14743

adrianschmidt:
@paulinewahle Can you add a description of what change is needed in lime-elements? I followed the link, but it wasn't clear to me what needed changing here in lime-elements.

PR #3680, opened 2025-09-22
feat(code-editor): expand to fullscreen view

paulinewahle:
fix: https://github.com/Lundalogik/lime-elements/issues/3679

## Summary by CodeRabbit

- New Features
  - Add expandable fullscreen mode for the code editor with an optional fullscreen button.
  - New prop: expandable (default: false) to enable expansion.
  - New event: expand emitted when users request fullscreen.
  - Editor content and state seamlessly persist between inline and fullscreen views.
- Style
  - Updated editor layout and focus handling; fullscreen button appears on focus and in dialogs.
- Documentation
  - New example showcasing the expandable editor with a JSON configuration.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/code-editor/code-editor.tsx
+++ b/src/components/code-editor/code-editor.tsx
@@ -153,31 +153,13 @@ export class CodeEditor {
     }
 
     public componentDidRender() {
-        if (!this.editor) {
-            this.editor = this.createEditor();
-            this.cmWrapper = this.editor.getWrapperElement() as HTMLElement;
-        }
+        this.ensureEditor();
+
         if (!this.expandable) {
             return;
         }
-        // Re-parent the existing CodeMirror DOM into dialog or back inline
-        if (this.openFullscreenMode) {
-            const dialogHost = this.host.shadowRoot.querySelector(
-                '.editor-dialog-host'
-            ) as HTMLElement;
-            if (dialogHost && this.cmWrapper?.parentElement !== dialogHost) {
-                dialogHost.append(this.cmWrapper);
-                requestAnimationFrame(() => this.editor.refresh());
-            }
-        } else {
-            const inlineHost = this.host.shadowRoot.querySelector(
-                '.editor-inline-host .editor'
-            ) as HTMLElement;
-            if (inlineHost && this.cmWrapper?.parentElement !== inlineHost) {
-                inlineHost.append(this.cmWrapper);
-                requestAnimationFrame(() => this.editor.refresh());
-            }
-        }
+
+        this.reparentEditor();
     }
 
     @Watch('value')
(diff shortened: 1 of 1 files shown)


src/components/code-editor/code-editor.tsx
```diff
@@ -108,18 +147,37 @@ export class CodeEditor {
...
         const editorElement = this.host.shadowRoot.querySelector('.editor');
-        // eslint-disable-next-line no-unsafe-optional-chaining
         for (const child of editorElement?.childNodes) {
             child.remove();
         }
     }
 
     public componentDidRender() {
-        if (this.editor) {
+        if (!this.editor) {
```
Befkadu1, line 156: Too many conditions and it is very complex.

paulinewahle: > Too many conditions and it is very complex.

Do you mean this condition is unnecessary or the overall structure is too complex? 
Because my idea was it to move the same editor between the two places, which is more complex, but that way we could save the repetitive code of placing the code editor twice.

src/components/code-editor/examples/code-editor-expandable.tsx
```diff
@@ -0,0 +1,37 @@
+import { Component, h, State } from '@stencil/core';
+import { data } from '../../table/examples/birds';
+
```
Kiarokh: I really like the idea of this feature. 
when it comes to the implementation, we may want to consider to have a look at the fullscreen API which we use in `limel-file-viewer` to enter some sort of fullscreen mode. That might make the implementation less complicated. But needs to be investigated.

But looking at the screenshot which was provided in the linked issue, I wonder if we are treating the problem in the right place. 🤔 It seems they just want a taller code editor in their UI, and the issue is on the consumer side, not the component. Entering fullscreen mode might be helpful. But in that context, they may need to see the code editors side-by-side, or see other info simultaneously. Therefore I am not sure if this new feature really fixes their issue. It might, but that's worth investigating too. 😊 

Image

paulinewahle: The limel-file-viewer absolutely seems to be the cleaner solution for a fullscreen mode. 
Regarding the problems users are having in the application configuration, yes they might want to see this side by side and @uyttyu mentioned that a change to the max height would solve this issue because the config code can be long and involves long scrolling right now. 
This PR isn't by far the perfect solution, but my quick attempt to acutely improve a really big pain. So, instead of leaving this laying around until we can make it perfect, can't we give them this instead of nothing for now?

Kiarokh:
let's close this PR for now. @paulinewahle let's have a talk at a later step, and investigate this again. We have now your new "COPY" button, and that means things should change in this PR to. We may re-open it again, or make a new one.

The underlying issue behind this PR is solve now, elsewhere.

I take the liberty to close it for now. I hope you don't misunderstand me @paulinewahle. The current state requires a lot of work, before it can get merged. Let's do it when the need is stronger, or when we have more time.

PR #3699, opened 2025-10-06
Form: Add hidden schema option and early null return in SchemaField

jecrapo:
## What?
Add support for `lime.hidden` property in form schemas to hide fields from the UI while preserving their values in form data.

Changes:
- Add `hidden?: boolean` to LimeSchemaOptions interface
- Values are still  preserved in form state for all types (objects, arrays, primitives)
- Hidden fields don't participate in CSS grid layout

Usage:
```python
field = fields.Str(
  metadata={
    "lime": {
      "hidden": True
    }
  }
)
```

Fixes: https://github.com/Lundalogik/crm-insights-and-intelligence/issues/95

## Why?
There was no straightforward way to hide form fields generated from schemas simply by adding a property to the schema. I need it so I added it. It think it can be useful for other people too.

## Summary by CodeRabbit

* **New Features**
  * You can now hide form fields from the UI while preserving their values in the form data via a new schema option.
  * Hidden fields are fully excluded from rendering, including any custom rendering paths, enabling cleaner conditional or advanced workflows.
  * Backwards compatible: existing forms behave unchanged unless the new hidden flag is used.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/fields/schema-field.ts
+++ b/src/components/form/fields/schema-field.ts
@@ -270,6 +270,10 @@ export class SchemaField extends React.Component<FieldProps> {
     }
 
     render() {
+        if (this.props.schema.lime?.hidden) {
+            return null;
+        }
+
         if (hasCustomComponent(this.props.schema)) {
             return this.renderCustomComponent(this.props);
         }
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -2559,6 +2559,7 @@ export interface LimeSchemaOptions {
     disabled?: boolean;
     // (undocumented)
     help?: string | Partial<Help>;
+    hidden?: boolean;
     layout?: LimeLayoutOptions;
 }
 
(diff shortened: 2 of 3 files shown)


src/components/form/fields/schema-field.ts
```diff
@@ -270,6 +270,10 @@ export class SchemaField extends React.Component<FieldProps> {
     }
 
     render() {
+        if (this.props.schema.lime?.hidden) {
+            return null;
```
adrianschmidt, line 274: > Hidden fields don't participate in CSS grid layout

Have you confirmed that that is the case?

The reason I'm asking is that I'm unsure if there will be some empty custom component rendered anyway. I see that this is a React component, so perhaps that doesn't happen? Had this been a Stencil component, returning `null` from the `render` method would still have left a `<limel-schema-field></limel-schema-field>` in the html, and that would have had to be hidden using CSS to void it taking up a space in the grid.

adrianschmidt (approved):
🥳

PR #3706, opened 2025-10-15
fix(list): prevent double `interact` event emits, when an item is clicked

Kiarokh:
fix https://github.com/Lundalogik/crm-client/issues/367
fix https://github.com/Lundalogik/crm-client/issues/364

## Summary by CodeRabbit

* **New Features**
  * Improved keyboard support: Enter/Space now toggle selection on focused list items.
  * Centralized interaction handling for more consistent selection across list variants.

* **Bug Fixes**
  * Clicking action menus no longer toggles items.
  * Disabled and aria-disabled items correctly ignore clicks and key presses.
  * More accurate selection state and feedback messages (Selected/Deselected).

* **Refactor**
  * Moved interaction handling from individual items to the list container for consistent behavior.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/list-item/list-item.tsx
+++ b/src/components/list-item/list-item.tsx
@@ -1,12 +1,4 @@
-import {
-    Component,
-    Host,
-    Prop,
-    h,
-    Event,
-    EventEmitter,
-    Element,
-} from '@stencil/core';
+import { Component, Host, Prop, h } from '@stencil/core';
 import { getIconName } from '../icon/get-icon-props';
 import type { IconSize } from '../icon/icon.types';
 import { createRandomString } from '../../util/random-string';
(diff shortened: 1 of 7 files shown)


devbymadde (approved):

src/components/list-item/list-item.tsx
```diff
@@ -143,18 +135,6 @@ export class ListItemComponent implements ListItem {
     public type: 'listitem' | 'menuitem' | 'option' | 'radio' | 'checkbox' =
         'listitem';
 
-    /**
-     * Emitted when the list item toggles selection (only for selectable types and not disabled).
-     */
-    @Event()
-    public interact: EventEmitter<{
```
adrianschmidt: This seems like a very breaking change, and ~must~ _might have to_ be marked as such.

EDIT: I guess it depends on whether there was ever a legitimate use-case for the stuff that was removed. If not, then this can perhaps just go as a bugfix.

adrianschmidt (changes requested):

src/components/list-item/list-item.tsx
adrianschmidt: > @adrianschmidt the component is `@private` and only other components internally are using it.

Ah, ok! Then it's fine, yes 😄

Issue #3718, opened 2025-10-31
Inconsistent text color in mobile UI reduces readability of read-only fields


LucyChyzhova:
## Current behavior
Safari (especially iOS Safari) recently started applying its own -webkit-text-fill-color and opacity styling to <input readonly> elements — treating them visually like disabled.
Desktop browsers often ignore this.

<img width="1170" height="744" alt="Image" src="https://github.com/user-attachments/assets/e7741284-9c9d-416a-9c87-0eea04c31d96" />

1. …

## Environment

- **lime-elements** version: 
- Framework used: 
- Logs:

PR #3719, opened 2025-10-31
fix(readonly): make readonly state readable on mobile devices

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/3718

Screenshot 2025-10-31 at 14 41 22

## Summary by CodeRabbit

* **Style**
  * Enhanced styling for readonly input fields to ensure proper text display and consistent visibility. Readonly form inputs now render with explicit color treatment and full opacity for improved usability.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/input-field/partial-styles/_readonly.scss
+++ b/src/components/input-field/partial-styles/_readonly.scss
@@ -39,3 +39,14 @@
         }
     }
 }
+
+:host([readonly]) {
+    .mdc-text-field__input {
+        // Override browser default styling for readonly inputs to ensure proper visibility
+        // - opacity: 1 prevents browsers from making readonly text appear faded/dimmed
+        // - -webkit-text-fill-color ensures consistent text color across WebKit browsers
+        //   (overrides browser autofill colors and disabled state styling)
+        opacity: 1;
+        -webkit-text-fill-color: shared_input-select-picker.$input-text-color;
+    }
+}


src/components/input-field/partial-styles/_readonly.scss
```diff
@@ -39,3 +39,13 @@
...
+
+:host(.lime-text-field--readonly),
+:host([readonly]) {
+    .mdc-text-field--disabled .mdc-text-field__input,
+    .mdc-text-field__input[readonly] {
+        opacity: 1;
+        color: shared_input-select-picker.$input-text-color;
+        -webkit-text-fill-color: shared_input-select-picker.$input-text-color;
+    }
+}
```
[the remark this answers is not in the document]
LucyChyzhova: Thanks Kia. I checked with browser stack but don't see it's overrides anything. Also this bug appears only in the input-field..

Kiarokh: Thanks for checking. But the specificity of this rule is really high. Why do we need to make it so specific? 

It feels we could just have something like:
```scss
    .mdc-text-field__input[readonly] {
        // opacity: 1;
        // color: shared_input-select-picker.$input-text-color;
        -webkit-text-fill-color: shared_input-select-picker.$input-text-color;
    }
```

Do we have to nest it in double selectors for the readonly:
```scss
:host(.lime-text-field--readonly),
:host([readonly]) {
…
}
```
and even further nested selector classes?

Are you sure we need the `opacity` and `color`? If this is an iOS/Safari fix, the `-webkit-text-fill-color: ` should be enough. If all the 3 rules are needed, maybe also consider adding a docs block together with this piece of code, so the next reader knows why this overrides are there. Because they look brutally specific and hard to understand why they should be kept there.

src/components/input-field/partial-styles/_readonly.scss
```diff
@@ -42,10 +42,12 @@
 
 :host(.lime-text-field--readonly),
 :host([readonly]) {
```
Kiarokh: Do we need both of these? I think onl this one is enough:
```scss
:host([readonly]) {
```

PR #3757, opened 2025-12-10
fix(menu): improve keyboard navigation

Kiarokh:
Since this component relies on MDC, there is already some inbuilt keyboard interactions support for arrow keys. We had some additional customized keyboard navigation support for the cases when there is a searcher. But they didn't work well-enough for keyboard users. For instance, the breadcrumbs were completely ignored. User could press the DOWN key to set the focus on menu items, but they could never set the focus back to the input field again by pressing UP for instance, to keep modifying their keywords.

## Expected interaction design
**In a menu with searcher and nested items** 
- when you open the menu, focus should be on the input field. It doesn't matter that you type a keyword or not, in both cases:
- when you press <kbd>↓</kbd>
  - if there is a breadcrumbs, focus should go to the first item of the bread crumbs component _(note that internal keyboard navigation of `limel-breadcrumbs` component is not handled here in this PR and requires another PR)_
  - if there is no breadcrumbs, focus should be set on the first search result (first menu item)
-   when you keep pressing down <kbd>↓</kbd> and reach the last list item, and press  <kbd>↓</kbd>  again, the focus should go back up to the input field. _(If there is no input field, the focus should go back up to the first menu item.)_
- When the focus is already in the input field, if you press  <kbd>↑</kbd> the focus should go to the last menu item.
- For menu items that have sub-menus (nested menu items), pressin  right <kbd>→</kbd> should navigate to the sub menus
- When you are in a sub menu, and the focus is not on the input field or on the bread crumbs (focus is on the list items), pressing  left <kbd>←</kbd> should go one step back to the previous step in the hierarchy.

https://github.com/user-attachments/assets/960da85a-20f7-4eb1-b6ce-7a5967c70473

Change summary
- Added refs and helper methods so the menu can focus the search field, breadcrumbs, or list items programmatically (e.g., `setBreadcrumbsElement`, `focusFirstListItem`, `isLastListItemFocused`).
- Enhanced keyboard handling: generated capture-phase listener to intercept arrow keys on the list; updated handlers for the input field and breadcrumbs so focus moves logically between search input, breadcrumbs, and menu items; restored wrap-around behaviour when no searcher exists.

@coderabbitai summary

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/menu/menu.e2e.ts
+++ b/src/components/menu/menu.e2e.ts
@@ -138,3 +138,192 @@ describe('limel-menu', () => {
         });
     });
 });
+
+describe('limel-menu keyboard navigation', () => {
+    describe('with nested items (sub-menus)', () => {
+        let page: E2EPage;
+        let limelMenu: HTMLLimelMenuElement & E2EElement;
+
+        beforeEach(async () => {
+            page = await newE2EPage({
+                html: `
+                    <limel-menu>
+                        <button slot="trigger">Open Menu</button>
+                    </limel-menu>
+                `,
+            });
+            limelMenu = (await page.find('limel-menu')) as any;
+            const items = [
+                {
+                    text: 'Parent Item',
+                    items: [{ text: 'Child Item 1' }, { text: 'Child Item 2' }],
+                },
+                { text: 'Another Item' },
+            ];
+            await limelMenu.setProperty('items', items);
+            await page.waitForChanges();
+        });
+
+        it('right arrow on item with sub-menu emits navigateMenu', async () => {
+            // Click the trigger button to open the menu (this sets up proper focus)
+            const trigger = await page.find('button');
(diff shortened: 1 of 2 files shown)


Kiarokh:
@coderabbitai review again

devbymadde:
I tested this locally and it seems like when I push arrow down, it's always the second to first item that gets focused first, and not the first one as I'd expect

https://github.com/user-attachments/assets/f30d840c-5c15-4345-9779-b247830e5f35

PR #3990, opened 2026-03-31
feat(select): add `hotkey` prop for items & improve keyboard navigation

Kiarokh:
fix: https://github.com/Lundalogik/crm-client/issues/755

## New Features
* Added keyboard shortcut support for select options—users can now assign and use hotkeys to quickly select items when the dropdown is open
* Hotkeys are only active when the dropdown menu is open and only for enabled options
* Includes support for platform-specific modifier key behavior

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/select/select.tsx
+++ b/src/components/select/select.tsx
@@ -14,6 +14,11 @@ import {
 } from '@stencil/core';
 import { isMobileDevice } from '../../util/device';
 import { ENTER, SPACE } from '../../util/keycodes';
+import {
+    hotkeyFromKeyboardEvent,
+    normalizeHotkeyString,
+    tokenizeHotkeyString,
+} from '../../util/hotkeys';
 import { isMultiple } from '../../util/multiple';
 import { createRandomString } from '../../util/random-string';
 import { SelectTemplate, triggerIconColorWarning } from './select.template';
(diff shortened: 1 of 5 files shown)


src/components/select/select.tsx
```diff
@@ -367,6 +398,105 @@ export class Select {
...
+
+            return;
+        }
+
+        this.change.emit(matchedOption);
+        this.menuOpen = false;
+        this.setTriggerFocus();
+    };
+
+    private isReservedSelectHotkey(hotkey: string): boolean {
```
adrianschmidt: @Kiarokh Small observation on `isReservedSelectHotkey` vs `isReservedMenuHotkey`:

In **menu**, the logic is:
```ts
if (hasModifiers) {
    return false; // any modified key is allowed
}
return key === 'arrowup' || ... || key === 'tab' || key === 'enter' || key === 'space' || key === 'escape';
```

In **select**, arrow keys and tab are reserved *regardless* of modifiers:
```ts
if (key === 'arrowup' || ... || key === 'arrowright') {
    return true; // always reserved, even with modifiers
}
if (key === 'tab') {
    return true; // always reserved, even with modifiers
}
const hasModifiers = tokens.length > 1;
return !hasModifiers && (key === 'enter' || key === 'space' || key === 'escape');
```

This means a hotkey like `ctrl+arrowdown` would work in a menu but be blocked in a select. Since unmodified arrows are used for dropdown navigation in both components equally, it seems like the logic should be the same — i.e., the select version should also do the early `if (hasModifiers) return false` check, matching the menu.

Was this divergence intentional, or should they be unified?

PR #4007, opened 2026-04-09
feat(form): upgrade @rjsf/core from v2 to v6

jgroth:
## Summary by CodeRabbit

* **New Features**
  * Drag-and-drop array reordering with emitted reorder events
  * Row layout context to suppress labels/helper text
  * Shared RJSF validator integrated for live validation

* **Improved UX**
  * Consolidated validation and helper-text behavior across fields
  * Better nested object/array rendering and custom-component handling
  * Code editor now reports changes with field path context
  * Form styling scoped to rjsf-field

* **Tests**
  * New end-to-end suite and additional unit tests for arrays and validation

* **Chores**
  * Dev dependency updates for RJSF validator packages

## Summary

Upgrades `@rjsf/core` from v2 to v6, along with the new required packages
`@rjsf/utils`, `@rjsf/validator-ajv8`, and `ajv` v8.

The v6 API restructures how validation works — validators are now separate
packages instead of built into core. This branch adapts all form fields,
templates, and widgets to the new API, replaces the custom `schema-cache`
with Ajv8-based validation, and adds `validator.ts` and
`validation-display.ts` modules.

The earlier commit on this branch caches compiled Ajv validators by schema
content hash to avoid recompilation on every render.

Closes Lundalogik/crm-client#856

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


src/components/form/fields/array-field.ts
```diff
@@ -30,28 +38,70 @@ import { resetDependentFields } from './field-helpers';
...
+
+        if (!Array.isArray(formData)) {
+            return;
+        }
+
+        const reordered = [...formData];
+        const [item] = reordered.splice(fromIndex, 1);
+        reordered.splice(toIndex, 0, item);
+
+        this.props.onChange(reordered, fieldPathId.path);
```
Befkadu1, line 86: `fieldPathId` is destructured from props here without a guard. `handleReorder` is driven by a DOM event rather than RJSF's own call flow, so if the field is ever mounted in a context where RJSF didn't supply `fieldPathId` (or unmount races a pending event), `fieldPathId.path` will throw. A cheap `if (!fieldPathId) return;` up top — or a spec covering the missing case — would prevent the runtime error.

src/components/form/form.e2e.tsx
```diff
@@ -0,0 +1,549 @@
...
+async function waitForReactRender(
+    root: HTMLElement,
+    waitForChanges: () => Promise<void>
+) {
+    for (let i = 0; i < 10; i++) {
+        await waitForChanges();
+        if (root.shadowRoot?.querySelector('.root *')) {
+            return;
+        }
+    }
```
Befkadu1, line 548: This helper returns silently after 10 iterations even when the shadow DOM content never rendered. A future regression that stops React from rendering would cause every test that calls this helper to pass anyway — because the subsequent `querySelector(...)` returns `null` and any `expect(...).toBeTruthy()` that follows is preceded by a `return`, not a throw. Suggest throwing at the end of the loop so real failures surface:

```ts
async function waitForReactRender(root, waitForChanges) {
    for (let i = 0; i < 10; i++) {
        await waitForChanges();
        if (root.shadowRoot?.querySelector('.root *')) {
            return;
        }
    }
    throw new Error('React did not render within 10 ticks');
}
```

Befkadu1 (approved):
⚡

PR #4009, opened 2026-04-09
fix(table): prevent local re-sort of server data in remote mode

### LINKED ISSUE #4008: fix(table): remote mode re-sorts server data locally using string comparison
## Summary

In remote mode (`mode="remote"`), `limel-table` re-sorts data locally after receiving it from the server, using Tabulator's default string comparison. This means integer columns (e.g. 1, 2, 10, 11, 20) are displayed in lexicographic order (1, 10, 11, 2, 20) instead of the server-returned order.

## Root cause

`getAjaxOptions()` sets `ajaxSorting: true`, which was the Tabulator 4/5 API for remote sorting. In Tabulator 6, the correct option is `sortMode: 'remote'`. Without it, Tabulator defaults to `sortMode: 'local'` and re-sorts data after every server response — overriding the correctly sorted data from the server.

## Steps to reproduce

1. Use `limel-table` with `mode="remote"` and paginated integer data
2. Click a column header to sort
3. Observe that the displayed order is lexicographic (string sort) instead of the server-returned order

## Expected behavior

Data is displayed in the order returned by the server, without any client-side re-sorting.

jensgustafsson:
## Summary

- Adds `sortMode: 'remote'` to Tabulator options when `mode="remote"`, which is the Tabulator 6 API for disabling client-side sorting — replacing the now-dead `ajaxSorting: true` (Tabulator 4/5 only)
- Fixes `requestData` to return a plain data array when pagination is not enabled, since `sortMode: 'remote'` causes Tabulator to call `requestData` on sort changes and the `{last_page, data}` wrapper requires the pagination module to unwrap it

## Test plan

- [ ] Unit test: `getAjaxOptions()` includes `sortMode: 'remote'` when mode is remote, absent otherwise (`table.spec.ts`)
- [ ] E2e test: table with integer data in reverse order — after sort click and server response with numerically sorted data, display matches server order and not string sort order (`table.e2e.tsx`)
- [ ] Existing remote sorting e2e test still passes

Closes #4008

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end and unit tests covering remote sorting behavior and aggregate-field updates.

* **Bug Fixes**
  * Ensured remote sorting returns rows in the same order provided by the server.
  * Fixed aggregate-change handling to update table columns without full reinitialization.

* **Improvements**
  * Made data responses consistent for paginated vs non-paginated requests.

### DIFF
--- a/src/components/table/table.e2e.tsx
+++ b/src/components/table/table.e2e.tsx
@@ -165,6 +165,64 @@ describe('limel-table', () => {
     });
 
     describe('remote sorting', () => {
+        it('displays data in the order returned by the server, without re-sorting locally', async () => {
+            // Initial data is in reverse numeric order (unsorted).
+            // After sorting, the server returns data in numeric ASC order.
+            // If Tabulator re-sorts locally as strings, it would produce
+            // "1","10","11","2","20" instead of 1, 2, 10, 11, 20.
+            const columns = [{ field: 'order', title: 'Order' }];
+            const initialData = [
+                { id: 5, order: 20 },
+                { id: 4, order: 11 },
+                { id: 3, order: 10 },
+                { id: 2, order: 2 },
+                { id: 1, order: 1 },
+            ];
+
+            const { root, waitForChanges, setProps } = await renderTable({
+                mode: 'remote',
+                data: initialData,
+                columns,
+                totalRows: 5,
+                pageSize: 5,
+            });
+
+            // Simulate server responding with numerically sorted data
+            const serverSortedData = [
+                { id: 1, order: 1 },
+                { id: 2, order: 2 },
+                { id: 3, order: 10 },
+                { id: 4, order: 11 },
+                { id: 5, order: 20 },
+            ];
+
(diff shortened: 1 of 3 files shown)


src/components/table/table.e2e.tsx
```diff
@@ -165,6 +165,64 @@ describe('limel-table', () => {
...
+                    setProps({ data: serverSortedData });
+                }
+            });
+
+            const container = getContainer(root);
+            const headers = container.querySelectorAll('[role="columnheader"]');
+
+            (headers[0] as HTMLElement).click();
+            await waitForChanges();
+            await new Promise((resolve) => setTimeout(resolve, 300));
```
devbymadde, line 210: I don't think it's optimal to set timeouts like this, it could be flaky when running in CI. But I noticed that we already have it in may places in this test file already, so let's go for it 🫣

devbymadde (approved):
I've tested it and it works well, both the sorting but also together with the other fixes that we've made. Well done Jens/Claude :D

PR #4027, opened 2026-04-15
feat(chip-set): expose per-chip invalid state

### LINKED ISSUE #4026: chip-set: expose invalid state per chip
## Description

The `limel-chip` component already supports an `invalid` prop that visualises
a chip in an error state (red background with a chessboard pattern, white text).
However, when chips are rendered via `limel-chip-set`, consumers can only mark
the chip-set as a whole as invalid — there is no way to flag specific entries
in the `value` array as invalid.

## Motivation

A common use case is an input chip-set of free-form values (such as email
addresses or tags) where some individual entries may fail validation while
others are valid. Today, consumers have to either:
- show a single container-level error (useless when most entries are valid), or
- swap the chip icon and hope that communicates the problem.

Neither gives a proper per-chip visual error state, even though `limel-chip`
itself already supports it.

## Proposed solution

Add an optional `invalid` field to the `Chip` interface and forward it through
`limel-chip-set` to the rendered `limel-chip`, so consumers can flag
individual chips as invalid via the `value` array:

```ts
value = [
  { id: 1, text: 'alice@example.com' },
  { id: 2, text: 'not-an-email', invalid: true, icon: 'error' },
  { id: 3, text: 'bob@example.com' },
];
```

This is independent of the existing chip-set-level `invalid` prop, which
continues to mark the whole field as invalid.

TommyLindh2:
Closes #4026

## Summary

- Add an optional `invalid` field to the `Chip` interface
- Forward `chip.invalid` through `limel-chip-set`'s `getChipProps` to the rendered `limel-chip`, which already supports the `invalid` prop
- Add a `limel-example-chip-set-invalid-chips` example demonstrating the feature with an email recipients input

Consumers can now flag individual chips as invalid via the `value` array:

```ts
value = [
  { id: 1, text: 'alice@example.com' },
  { id: 2, text: 'invalid-email@address', invalid: true, icon: 'error' },
  { id: 3, text: 'bob@example.com' },
];
```

This is independent of the chip-set-level `invalid` prop, which continues to mark the whole field as invalid.

## Test plan

- [x] `npm run build` passes
- [x] `npm run lint` passes
- [x] `npm run test` — all 1119 tests pass
- [ ] Manual: open the new `limel-example-chip-set-invalid-chips` example and confirm `invalid-email@address` renders with the invalid styling and error icon while the two email entries render normally
- [ ] Manual: type a new invalid entry (e.g. `foo`) and a valid entry (e.g. `carol@example.com`) and confirm each is flagged correctly

## Summary by CodeRabbit

* **New Features**
  * Chips can now be marked invalid individually so single chips show an error/validation appearance.
  * Chip objects accept an optional invalid flag to control per-chip error state.

* **Documentation**
  * Added an example demonstrating per‑chip invalid state, input handling, and dynamic chip updates.

### DIFF
--- a/src/components/chip-set/examples/chip-set-invalid-chips.tsx
+++ b/src/components/chip-set/examples/chip-set-invalid-chips.tsx
@@ -1,0 +1,173 @@
+import {
+    Chip,
+    LimelChipSetCustomEvent,
+    Option,
+    JSX,
+    LimelSelectCustomEvent,
+} from '@limetech/lime-elements';
+import { Component, h, Host, State } from '@stencil/core';
+import { ENTER } from '../../../util/keycodes';
+
+type ChipSetType = Exclude<JSX.LimelChipSet['type'], undefined>;
+
+/**
+ * Per-chip invalid state
+ *
+ * Set `invalid: true` on any chip in the `value` array to mark that
+ * specific chip as invalid. This is independent of the chip-set-level
+ * `invalid` prop, which is intended for signalling that the whole field
+ * is invalid. Per-chip `invalid` lets the consumer flag individual
+ * entries, for example an address that fails validation in a list of
+ * recipients.
+ *
+ * In this example, each entry is checked with a simple email regex when
+ * added. Invalid entries are rendered with `invalid: true` and an error
+ * icon.
+ *
+ * :::note
+ * Marking individual chips as `invalid` does **not** automatically set
+ * the invalid state of the chip-set as a whole. The consumer is
+ * responsible for deciding whether the field itself should be
+ * considered invalid, and for setting the chip-set-level `invalid` prop
+ * accordingly. This gives the consumer full control over the
+ * validity of the field.
+ * :::
+ */
+@Component({
+    tag: 'limel-example-chip-set-invalid-chips',
+    shadow: true,
+})
+export class ChipSetInvalidChipsExample {
+    @State()
+    private value: Chip[];
+
+    @State()
+    private textValue = '';
+
+    @State()
+    private selectedInputType?: Option<ChipSetType>;
+
+    private inputTypeOptions?: Array<Option<ChipSetType>>;
+
+    constructor() {
+        this.value = [
+            this.createChip('alice@example.com'),
+            this.createChip('invalid-email@address'),
+            this.createChip('bob@example.com'),
+        ];
+    }
+
+    public componentWillLoad() {
+        this.inputTypeOptions = [
+            { text: 'Input', value: 'input' },
+            { text: 'Choice', value: 'choice' },
+            { text: 'Filter', value: 'filter' },
+        ];
+        this.selectedInputType = this.inputTypeOptions[0];
+    }
+
+    public render() {
+        return (
+            <Host>
+                <limel-chip-set
+                    type={this.selectedInputType?.value}
+                    label="Recipients"
+                    helperText="Type an email address and press Enter. Invalid entries are flagged on the chip."
+                    searchLabel="Type an email address"
+                    value={this.value}
+                    onChange={this.handleChange}
+                    onInput={this.handleInput}
+                    onKeyUp={this.onKeyUp}
+                />
+                <limel-example-controls
+                    style={{ '--example-controls-max-columns-width': '9rem' }}
+                >
+                    <limel-select
+                        label="Chip set type"
+                        value={this.selectedInputType}
+                        options={this.inputTypeOptions}
+                        onChange={this.handleInputTypeChange}
+                    />
+                </limel-example-controls>
+                <limel-example-value value={this.value} />
+            </Host>
+        );
+    }
+
+    private handleInputTypeChange = (
+        event: LimelSelectCustomEvent<Option<ChipSetType>>
+    ) => {
+        this.selectedInputType = event.detail;
+        this.value = this.value.map((chip) => {
+            chip.selected = false;
+            return chip;
+        });
+    };
+
+    private handleInput = (
+        event: LimelChipSetCustomEvent<string> | InputEvent
+    ) => {
+        if (event instanceof CustomEvent) {
+            this.textValue = event.detail;
+        }
+    };
+
+    private onKeyUp = (event: KeyboardEvent) => {
+        if (event.key === ENTER && this.textValue.trim()) {
+            this.value = [
+                ...this.value,
+                this.createChip(this.textValue.trim()),
+            ];
+            this.textValue = '';
+        }
+    };
+
+    private handleChange = (event: LimelChipSetCustomEvent<Chip | Chip[]>) => {
+        if (Array.isArray(event.detail)) {
+            this.value = event.detail;
+        } else {
+            const interactedChip = event.detail as Chip;
+
+            if (this.selectedInputType?.value === 'filter') {
+                this.value = this.value.map((chip) => {
+                    if (chip.id === interactedChip.id) {
+                        return interactedChip;
+                    }
+
+                    return chip;
+                });
+            } else if (this.selectedInputType?.value === 'choice') {
+                this.value = this.value.map((chip) => {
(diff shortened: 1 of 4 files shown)


src/components/chip-set/examples/chip-set-invalid-chips.tsx
```diff
@@ -0,0 +1,98 @@
...
+    @State()
+    private value: Chip[];
+
+    @State()
+    private textValue = '';
+
+    constructor() {
+        this.value = [
+            this.createChip('alice@example.com'),
+            this.createChip('not-an-email'),
```
Kiarokh:
```suggestion
            this.createChip('invalid-email@address'),
```

src/components/chip-set/examples/chip-set-invalid-chips.tsx
```diff
@@ -0,0 +1,98 @@
...
+            this.createChip('bob@example.com'),
+        ];
+    }
+
+    public render() {
+        return [
+            <limel-chip-set
+                type="input"
```
Kiarokh: the invalid is not only related to the `type="input"` right? So please add a normal chip-set in this example as well, to make sure we cover both cases. 🙏 It's always nice to have a visual test. Even if we write a code-test for some stuff

Kiarokh: it's possible to test this in the composite example. But wouldn't hurt to have it visible in the example where we are talking about the feature

Image

src/components/chip-set/examples/chip-set-invalid-chips.tsx
```diff
@@ -0,0 +1,98 @@
...
+ * Set `invalid: true` on any chip in the `value` array to mark that
+ * specific chip as invalid. This is independent of the chip-set-level
+ * `invalid` prop, which is intended for signalling that the whole field
+ * is invalid. Per-chip `invalid` lets the consumer flag individual
+ * entries, for example an address that fails validation in a list of
+ * recipients.
+ *
+ * In this example, each entry is checked with a simple email regex when
+ * added. Invalid entries are rendered with `invalid: true` and an error
+ * icon.
```
Kiarokh, line 25: if the item already has an icon, it doesn't show an error icon. I'm also a bit uncertain about this. should we show an error icon? You know that the consumers don't have our icon package installed. So it'll be a challenge to have a list of icons that the component's UIs show, and make sure that consumers provide same icons with the same filenames. Also if we manage to generate such a list, we should never change an icon that the UI of a component is showing. That'd break the consumer's UIs.

I think we should skip the error icon. It's enough with how the chip is already showing its own invalid state

src/components/chip-set/examples/chip-set-invalid-chips.tsx
```diff
@@ -0,0 +1,98 @@
+import { Chip, LimelChipSetCustomEvent } from '@limetech/lime-elements';
+import { Component, h, State } from '@stencil/core';
+import { ENTER } from '../../../util/keycodes';
+
+/**
+ * Per-chip invalid state
+ *
+ * Set `invalid: true` on any chip in the `value` array to mark that
+ * specific chip as invalid. This is independent of the chip-set-level
+ * `invalid` prop, which is intended for signalling that the whole field
```
Kiarokh, line 18: an invalid chip should also make the input component invalid if you ask me. that's a better design. it explains why the whole chip-set is invalid.

src/components/chip-set/examples/chip-set-invalid-chips.tsx
```diff
@@ -0,0 +1,98 @@
+import { Chip, LimelChipSetCustomEvent } from '@limetech/lime-elements';
+import { Component, h, State } from '@stencil/core';
+import { ENTER } from '../../../util/keycodes';
```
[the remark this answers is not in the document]
TommyLindh2: We do that in a other examples, so I'll leave it as is

src/components/chip-set/examples/chip-set-invalid-chips.tsx
```diff
@@ -0,0 +1,98 @@
...
+                label="Recipients"
+                helperText="Type an email address and press Enter. Invalid entries are flagged on the chip."
+                searchLabel="Type an email address"
+                value={this.value}
+                onChange={this.handleChange}
+                onInput={this.handleInput}
+                onKeyUp={this.onKeyUp}
+            />,
+            <limel-example-value value={this.value} />,
+        ];
```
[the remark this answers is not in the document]
TommyLindh2: Addressed 8a52a7c54089e91ab268d19d5ab7ed8bc2ba0781

src/components/chip-set/examples/chip-set-invalid-chips.tsx
```diff
@@ -0,0 +1,98 @@
...
+    private createChip = (text: string): Chip => {
+        const isValid = this.looksLikeEmail(text);
+
+        return {
+            id: text,
+            text: text,
+            removable: true,
+            invalid: !isValid,
+            ...(!isValid && { icon: 'error' }),
+        };
```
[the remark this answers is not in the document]
TommyLindh2: It's just an example, it's not important. I'll leave it

src/components/chip-set/examples/chip-set-invalid-chips.tsx
TommyLindh2: Addressed: c9bc96e6fedd86346f51967cd5d1c2dfd8442f09

src/components/chip-set/examples/chip-set-invalid-chips.tsx
TommyLindh2: Fixed: 5c17769bde27037ce3532eaf65c7ccb4b390b34a

src/components/chip-set/examples/chip-set-invalid-chips.tsx
TommyLindh2: The icon is only in the example, not in the component. That way the consumer can choose wether or not to use an icon and in that case which icon to use.

Isn't that how we want it?

src/components/chip-set/examples/chip-set-invalid-chips.tsx
TommyLindh2: Personally I think it's up to the consumer to decide that.
Then they can do however they want. Maybe we can add it to the example?

Or do you think the invalid prop should be hard coded if any of the chips have invalid: true?

Kiarokh: Ok yes, it makes sense I guess. In many cases, the chip could be perhaps invalid, but the the chipset still valid. We could perhaps add a note about this

src/components/chip-set/examples/chip-set-invalid-chips.tsx
Kiarokh: oh yes. sorry. that's fine

Kiarokh (dismissed):

src/components/chip-set/examples/chip-set-invalid-chips.tsx
TommyLindh2: Because you've approved this now, I will squash and merge this, and this can be done later

TommyLindh2: Changed, my mind, I'll do a docs commit adding this as a note.
And let you approve this PR again. So we don't forget about it

TommyLindh2: Here's my fix for adding the note about invalid state: [diff change of force-push](https://github.com/Lundalogik/lime-elements/compare/f79e5c82f870dd9cc2f3f9306865985e4862f7e4..30e50beea55dae5a80dc925ad8ba425ad922a197)

I did a force push to the existing commit where I added the example

Kiarokh (approved):

PR #4032, opened 2026-04-16
fix(picker): reset search query when a value is selected or removed

### LINKED ISSUE #4031: Picker retains stale search query after selecting a value
## Description
When searching for something in the picker (e.g. "Ayla") and selecting a value, the search string is kept in the component's local state (`textValue`). 

If the user then empties the field and focuses it again, it visually appears as an empty search (`""`), but the picker internally still searches for the old string ("Ayla"). This causes confusing results where the dropdown shows filtered items that don't match the apparently empty input.

## Steps to reproduce
1. Focus the picker and type a search string, e.g. "Ayla"
2. Select a matching value from the dropdown
3. Remove the selected value (chip) so the field is empty again
4. Focus the picker input

**Expected:** The dropdown shows unfiltered results (as if searching for `""`)
**Actual:** The dropdown shows results filtered by the old search string "Ayla"

## Root cause
`handleChange` in `picker.tsx` does not reset `this.textValue` when a chip is removed. On the next focus, `handleInputFieldFocus` reads the stale `textValue` and passes it to the debounced search.

TommyLindh2:
## Summary
- Reset `textValue` to `''` in `handleChange` so the picker doesn't retain a stale search query after selecting or removing a value
- Fixes the bug where focusing an empty picker would still search using the previous query string

Closes #4031

https://github.com/user-attachments/assets/3ad19155-0b15-4232-b250-34c4f856eb49

## Test plan
- [ ] Open a picker, search for a value (e.g. "Ayla"), and select it
- [ ] Remove the selected chip so the field is empty
- [ ] Focus the picker again and verify the dropdown shows unfiltered results (not filtered by the old query)
- [ ] Verify the same works correctly for multi-select pickers

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where search text in the picker wasn't clearing when chip selections changed.

### DIFF
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -661,6 +661,7 @@ export class Picker {
 
     private handleChange(event: LimelChipSetCustomEvent<Chip | Chip[]>) {
         event.stopPropagation();
+        this.textValue = '';
 
         let newValue = null;
         if (this.multiple) {


ha2818an (approved):
✅ 👍

PR #4035, opened 2026-04-20
Add reset-button-user-agent-styles mixin and apply to buttons

Kiarokh:
fix https://github.com/Lundalogik/crm-client/issues/923

## Summary by CodeRabbit

* **Refactor**
  * Consolidated button styling reset logic across components using a shared standardized approach, improving consistency and maintainability of button styling throughout the application.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/style/mixins.scss
+++ b/src/style/mixins.scss
@@ -21,6 +21,26 @@
 }
 
 /**
+* Neutralizes the user-agent styles that browsers apply to `<button>` elements
+* (their default background, border, padding, font, etc.), without stripping
+* the useful defaults that `all: unset;` would also remove
+* (such as `display`, `cursor`, and focus behavior).
+*
+* Prefer this mixin over `all: unset;` on `<button>` elements, then layer
+* the styling you want on top — often one of the `is-*-clickable` mixins.
+*/
+@mixin reset-button-user-agent-styles {
+    appearance: none;
+    background: none;
+    border: none;
+    padding: 0;
+    margin: 0;
+    font: inherit;
+    color: inherit;
+    text-align: inherit;
+}
+
+/**
 * This can be used on a trigger element that opens a dropdown menu or a popover.
 */
 
--- a/src/components/action-bar/action-bar-item/action-bar-item.scss
+++ b/src/components/action-bar/action-bar-item/action-bar-item.scss
@@ -14,7 +14,7 @@ limel-action-bar-item {
 }
 
 button {
-    all: unset;
+    @include mixins.reset-button-user-agent-styles;
 
     &:not([disabled]) {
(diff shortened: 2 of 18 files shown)


src/style/mixins.scss
```diff
@@ -20,6 +20,26 @@
...
+*/
+@mixin reset-button-user-agent-styles {
+    appearance: none;
+    background: none;
+    border: none;
+    padding: 0;
+    margin: 0;
+    font: inherit;
+    color: inherit;
+    text-align: inherit;
```
LucyChyzhova, line 40: should we add :
line-height: inherit;

what do you think?

src/style/mixins.scss
```diff
@@ -20,6 +20,26 @@
...
+/**
+* Neutralizes the user-agent styles that browsers apply to `<button>` elements
+* (their default background, border, padding, font, etc.), without stripping
+* the useful defaults that `all: unset;` would also remove
+* (such as `display`, `cursor`, and focus behavior).
+*
+* Prefer this mixin over `all: unset;` on `<button>` elements, then layer
+* the styling you want on top — often one of the `is-*-clickable` mixins.
+*/
+@mixin reset-button-user-agent-styles {
```
LucyChyzhova, line 32: Should we document the focus-outline contract in the docblock? The mixin correctly leaves `:focus-visible` alone, but that only holds if consumers don't override it with `outline: none` afterward. A one-line note, - something like "preserves focus outline; do not suppress without a replacement"

what do you think?

src/style/mixins.scss
Kiarokh: Not sure if it is needed. Do you see a case for it?

src/style/mixins.scss
Kiarokh: we have the `visualize-keyboard-focus` mixin. I think the consumers are expected to know what mixins there are, what they do, and what to use & when. This's fine as is I'd say.

src/style/mixins.scss
LucyChyzhova: Just double check with Claude and : Pre-Chromium Edge and old Safari versions had bugs where the font shorthand would reset line-height to normal instead of inheriting it. That's why you'll still see defensive line-height: inherit in older reset stylesheets and library code.

but maybe you are right we don't have to specify it now.

LucyChyzhova (approved):

PR #4038, opened 2026-04-21
fix(chip-set): keep input visible when it contains text

### LINKED ISSUE #4037: chip-set input hides typed text when focus leaves with chips present
## Current behavior

When `limel-chip-set` has existing chips and the user types but hasn't yet committed their text (via Enter/delimiter), the input is hidden as soon as focus leaves it. The typed text visually disappears even though it's still held in `textValue`.

Steps to reproduce:
1. Render a `limel-chip-set` of type `input` with at least one chip in `value`.
2. Type something into the input without committing it (no Enter, no delimiter).
3. Tab or click to a sibling focusable element (e.g. a picker's static custom action like "Add new").
4. Observe: the typed text is no longer visible; the input has the `hidden` class applied.

## Expected behavior

If the input holds uncommitted text, it should remain visible regardless of focus state or chip presence, so the user can see what they typed.

## Related

Surfaced while implementing limepkg-email #1934 where users tab from the recipient input to a "Create new recipient" custom action and lose visibility of the email they had typed.

ha2818an:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the input field visibility in the chip selection component. The input field now correctly remains visible when there is uncommitted text, improving the editing experience.

## Summary
- `inputHidden()` now returns `false` whenever `textValue` is non-empty, so typed-but-uncommitted text stays visible when focus leaves the input.
- Fixes #4037.

Surfaced while implementing [limepkg-email #1934](https://github.com/Lundalogik/limepkg-email/pull/1995) where users tab from the recipient input to a "Create new recipient" static action and lost visibility of the email they had typed.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/chip-set/chip-set.tsx
+++ b/src/components/chip-set/chip-set.tsx
@@ -540,6 +540,13 @@ export class ChipSet {
             return this.isFull();
         }
 
+        // Keep the input visible while it holds uncommitted text, so typed
+        // content doesn't disappear when focus leaves (e.g. tabbing to a
+        // sibling action).
+        if (this.textValue) {
+            return false;
+        }
+
         // If there are chips in the picker, hide the input to avoid the input
         // being placed on a new line and adding ugly space beneath the chips.
         // If there are no chips, show the input, or the picker will look weird.


FredrikWallstrom (approved):
Looks good, now it works just like it works when you don't have any chip at all.

Question is whether we want it to work like this or if we should "erase" all text when you unfocus the picker via the "Add new" action. I can kinda escape the text cleanup through the action, see movie.

However, out of scope for this PR.

https://github.com/user-attachments/assets/069752ce-2407-46a3-9877-9a0520d78d2a

PR #4047, opened 2026-04-23
fix(action-bar): do not render expand/shrink button for few icon-only…

LucyChyzhova:
… actions

https://github.com/user-attachments/assets/0a66e358-812d-4094-b6c9-837309827677

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved action bar collapse/expand button visibility: the button is now suppressed when displaying fewer than 4 items that are all icon-only.

* **Documentation**
  * Added a new example demonstrating a floating action bar with icon-only actions and dynamic action toggling.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/action-bar/action-bar.tsx
+++ b/src/components/action-bar/action-bar.tsx
@@ -40,6 +40,7 @@ import { Icon } from '../../global/shared-types/icon.types';
  * @exampleComponent limel-example-action-bar-colors
  * @exampleComponent limel-example-action-bar-floating
  * @exampleComponent limel-example-action-bar-floating-expand
+ * @exampleComponent limel-example-action-bar-floating-few-icon-only
  * @exampleComponent limel-example-action-bar-styling
  * @exampleComponent limel-example-action-bar-as-primary-component
  * @exampleComponent limel-example-action-bar-icon-title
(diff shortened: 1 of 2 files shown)


src/components/action-bar/action-bar.tsx
```diff
@@ -213,6 +214,12 @@ export class ActionBar {
             return;
         }
 
+        const items = this.actions.filter(isItem);
+        const allIconOnly = items.every((item) => item.iconOnly);
+        if (items.length < 4 && allIconOnly) {
```
Kiarokh, line 219: Note that actions can have labels of any size. Or be icon-only. So the `length` (number of actions) isn't the problem. The width is the problem, and it will be hard to track properly. Because size is also relative to the container width. Handling all these scenarios is unnecessary in this component, and make it extremely complicated. None of these should be concern of the component.

PR #4048, opened 2026-04-23
Remove shrink/expand functionality from action bar

Kiarokh:
The `collapsible` prop let consumers hide the action bar behind a shrink/expand toggle inside the component itself. That was the wrong layer to solve a layout problem at: a primitive action bar should not offer users a way to hide itself to reveal what it covers. The toggle also caused confusion for `iconOnly` and `fullWidth` variants, and added accessibility and complexity cost to a component that `limel-action-bar` did not need.

Remove the implementation (state, render branch, styles, `overFlowIcon` trigger, translation keys, and the example) but keep `collapsible` as a no-op `@deprecated` prop so existing consumers do not break. Emit a `console.warn` only when a consumer explicitly sets `collapsible={true}`, pointing them to implement the behavior in their own UI instead.

fix: https://github.com/Lundalogik/crm-client/issues/914
fix: https://github.com/Lundalogik/crm-client/issues/936

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed action bar collapse/expand functionality.
  * Removed `language` and `overFlowIcon` props.
  * Overflow menu now displays only an item count label.

* **Deprecations**
  * The `collapsible` prop now triggers a deprecation warning.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


etc/lime-elements.api.md
```diff
@@ -159,8 +159,8 @@ export namespace Components {
     export interface LimelActionBar {
         "accessibleLabel"?: string;
         "actions": Array<ActionBarItem | ListSeparator>;
+        // @deprecated (undocumented)
         "collapsible": boolean;
-        "language": Languages;
```
jgroth, line 163: These changes seems like they are breaking, but there is no mention of breaking changes in the commit message and thus no new major release will be made. I also think we should deprecate properties first before just removing them

Kiarokh: good point. will check

Kiarokh: the `language` is breaking. But keep it, and I'll instead use it for defaulting the accessible label to something useful

etc/lime-elements.api.md
```diff
@@ -174,7 +175,6 @@ export namespace Components {
     export interface LimelActionBarOverflowMenu {
         "items": Array<MenuItem | ListSeparator>;
         "openDirection": OpenDirection;
-        "overFlowIcon"?: Icon;
```
devbymadde, line 177: Do we need to mark also this as deprecated? It's a private component so perhaps not necessary

devbymadde (dismissed):
LGTM, just wondering if we should also mark the `overFlowIcon` as deprecated before we remove it.

etc/lime-elements.api.md
adrianschmidt: No, since they are marked as `@internal`, it should be fine 👍

src/translations/no.ts
```diff
@@ -5,6 +5,7 @@ export default {
     loading: 'Laster…',
     value: 'Verdi',
     items: 'Elementer',
+    'action-bar': 'Handlingslinje',
```
adrianschmidt: "Handlingslinje"… 🤔

That sounds very much like "gulebøj" to me, but maybe that's what it is actually called…

Have these translations been verified with any native speakers?

Kiarokh: I'm innocent

Kiarokh: I double checked with Claude, and it thinks `Handlingsfelt` is better for norwegian:

> It could, but I'd lean against it. "Handlingsfelt" is more natural in Norwegian for this kind of UI element. "Felt" (field/area) maps better to the concept of a container holding multiple buttons, while "linje" (line) implies something thinner and more linear — which is why it works fine in Danish where "linje" is commonly used for UI bars, but in Norwegian "felt" is the more established convention for this type of component.
That said, if consistency between Danish and Norwegian matters more to you than idiomatic precision, using "Handlingslinje" in both wouldn't be wrong — just slightly less natural in Norwegian.

devbymadde (approved):

PR #4049, opened 2026-04-24
fix(form): handle deep-leaf changes inside array items correctly

jgroth:
## Summary by CodeRabbit

* **Bug Fixes**
  * Consistent clearing: cleared fields (including nested array/object items and custom-rendered fields) now emit undefined.
  * Improved deep-update handling: deep changes inside arrays and nested fields rebuild and emit the correct updated value while preserving dependent sibling state.
  * Null handling: conversion from null→undefined now follows schema nullability rules.

* **Tests**
  * Added end-to-end and schema tests covering clearing, deep array updates, dependency propagation, and custom-rendered fields.

## Summary
- Fix 1: custom string components (e.g. `limel-input-field`) inside array items could not
  be cleared — emitting `undefined` left the field stuck at its previous value.
- Fix 2: when a trigger field inside an array item changed branch (e.g. a
  `dependencies`/`oneOf` schema), siblings that became invalid under the new branch kept
  their stale values.
- Fix 3: nested-object leaf data inside an array item (e.g. a `config` object with
  `additionalProperties` rendered by a custom component) was lost when the enclosing item
  also had `dependencies` — the deep-leaf change bubbled through the ancestor
  `SchemaField` for the array, which ran `resetDependentFields` with the leaf value
  against the array schema, corrupting `formData` before `ArrayField` could rebuild the
  item.
- Shared root cause: RJSF v6's `ArrayField` funnels every descendant change through one
  shared `onChange` that coerces `undefined` to `null` and bypasses our item-level
  `resetDependentFields` pass.
- In `ArrayField.handleChange`, intercept deep-path changes before the array-shape logic:
  rebuild the affected item locally with `undefined` restored where the leaf schema does
  not permit `null`, run `resetDependentFields` on it, then propagate the new array up.
- In `SchemaField.handleChange`, mirror the same `isDeepLeafChange` guard so deep-leaf
  bubbles pass through untouched — `resetDependentFields` should only run at the level
  where schema and data shapes line up.
- Adds e2e coverage for clearing top-level and array-nested string custom components, for
  resetting obsolete dependent siblings on a trigger change, and for preserving
  nested-object leaf data inside array items with dependencies.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/form/fields/array-field.ts
+++ b/src/components/form/fields/array-field.ts
@@ -1,8 +1,9 @@
 import React from 'react';
 import { getDefaultRegistry } from '@rjsf/core';
 import { FieldProps } from '@rjsf/utils';
-import { resetDependentFields } from './field-helpers';
+import { resetDependentFields, schemaAllowsNull } from './field-helpers';
 import { ARRAY_REORDER_EVENT } from '../templates/array-field';
+import { FormSchema } from '../form.types';
 
 const { fields: defaultFields } = getDefaultRegistry();
 const BaseArrayField = defaultFields.ArrayField;
@@ -87,8 +88,31 @@ export class ArrayField extends React.Component<FieldProps> {
     }
 
     private handleChange(newData, path) {
-        const { formData: oldData, schema } = this.props;
+        const { formData: oldData, schema, fieldPathId } = this.props;
         const { rootSchema } = this.props.registry;
+
+        // RJSF v6's built-in ArrayField shares one onChange handler across all
+        // descendants, so changes to leaves deep inside an array item bubble
+        // through here with the leaf value (not the full array) and a path
+        // that is deeper than this field's own path. In that case the
+        // array-shape logic below does not apply and would crash on scalar
+        // `newData`. The built-in handler also coerces `undefined` to `null`
+        // for every child change, which breaks field clearing for custom
+        // components; restore `undefined` when the leaf schema does not
+        // include `'null'`, mirroring the ingress conversion in `schema-field`.
+        if (fieldPathId && path.length > fieldPathId.path.length) {
+            const leafSchema = getSchemaAtPath(
+                schema as FormSchema,
+                path.slice(fieldPathId.path.length)
+            );
+            const value =
+                newData === null && !schemaAllowsNull(leafSchema)
+                    ? undefined
+                    : newData;
+            this.props.onChange(value, path);
+
+            return;
+        }
 
         // This case handles when the first list item is added. When there are no
         // items we get undefined instead of []
@@ -145,4 +169,33 @@ export class ArrayField extends React.Component<FieldProps> {
             React.createElement(BaseArrayField as any, arrayProps)
         );
     }
+}
+
+function getSchemaAtPath(
+    schema: FormSchema,
+    pathSegments: Array<string | number>
+): FormSchema | undefined {
+    let current: FormSchema | undefined = schema;
+    for (const segment of pathSegments) {
+        if (!current) {
+            return;
+        }
+
+        current = resolveSchemaSegment(current, segment);
+    }
+
+    return current;
+}
+
+function resolveSchemaSegment(
+    schema: FormSchema,
+    segment: string | number
+): FormSchema | undefined {
+    if (typeof segment === 'number') {
+        const items = schema.items;
+
+        return Array.isArray(items) ? items[segment] : items;
+    }
+
+    return schema.properties?.[segment];
 }
--- a/src/components/form/fields/array-field.ts
+++ b/src/components/form/fields/array-field.ts
@@ -1,6 +1,7 @@
 import React from 'react';
 import { getDefaultRegistry } from '@rjsf/core';
-import { FieldProps } from '@rjsf/utils';
+import { FieldProps, FieldPathList } from '@rjsf/utils';
+import { cloneDeep, isPlainObject, set } from 'lodash-es';
 import { resetDependentFields, schemaAllowsNull } from './field-helpers';
 import { ARRAY_REORDER_EVENT } from '../templates/array-field';
 import { FormSchema } from '../form.types';
@@ -88,28 +89,11 @@ export class ArrayField extends React.Component<FieldProps> {
     }
 
     private handleChange(newData, path) {
-        const { formData: oldData, schema, fieldPathId } = this.props;
+        const { formData: oldData, schema } = this.props;
         const { rootSchema } = this.props.registry;
 
-        // RJSF v6's built-in ArrayField shares one onChange handler across all
-        // descendants, so changes to leaves deep inside an array item bubble
-        // through here with the leaf value (not the full array) and a path
-        // that is deeper than this field's own path. In that case the
-        // array-shape logic below does not apply and would crash on scalar
-        // `newData`. The built-in handler also coerces `undefined` to `null`
-        // for every child change, which breaks field clearing for custom
-        // components; restore `undefined` when the leaf schema does not
-        // include `'null'`, mirroring the ingress conversion in `schema-field`.
-        if (fieldPathId && path.length > fieldPathId.path.length) {
-            const leafSchema = getSchemaAtPath(
-                schema as FormSchema,
-                path.slice(fieldPathId.path.length)
-            );
-            const value =
-                newData === null && !schemaAllowsNull(leafSchema)
-                    ? undefined
-                    : newData;
-            this.props.onChange(value, path);
+        if (this.isDeepLeafChange(path)) {
+            this.handleDeepLeafChange(newData, path);
 
             return;
         }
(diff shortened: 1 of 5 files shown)


src/components/form/fields/array-field.ts
```diff
@@ -146,3 +217,75 @@ export class ArrayField extends React.Component<FieldProps> {
...
+            return;
+        }
+
+        current = resolveSchemaSegment(current, segment);
+    }
+
+    return current;
+}
+
+function resolveSchemaSegment(
```
FredrikWallstrom, line 237: **Schema walker only handles `properties`/`items` — conditional branches return `undefined`.**

`resolveSchemaSegment` does not traverse `dependencies`, `oneOf`, `anyOf`, `allOf`, or `if`/`then`/`else`. For a leaf that lives inside a conditional branch — e.g. `dependencies.X.oneOf[*].properties.Y` (exactly the shape used by `arrayItemWithDependenciesSchema` in the new tests) — `getSchemaAtPath` returns `undefined`.

For the *clearing* case this is harmless: `schemaAllowsNull(undefined) === false`, so `null` correctly becomes `undefined`. But if a leaf inside a conditional branch is declared as `type: ['string', 'null']` (legitimately nullable), this lookup will still report "null not allowed" and convert a real `null` into `undefined` — silently changing user data.

Given that Lime CRM forms rely heavily on `dependencies` + `oneOf`, and the whole point of this fix is conditional-schema handling inside array items, it's worth either (a) walking conditional branches here, or (b) leaving a comment that nullability inside conditional branches is not supported and any null will be coerced to undefined.

src/components/form/fields/array-field.ts
```diff
@@ -146,3 +217,75 @@ export class ArrayField extends React.Component<FieldProps> {
...
+    target: unknown,
+    pathSegments: Array<string | number>,
+    value: unknown
+): unknown {
+    if (pathSegments.length === 0) {
+        return value;
+    }
+
+    const clone = cloneContainer(target, pathSegments[0]);
+    set(clone, pathSegments, value);
```
FredrikWallstrom, line 273: **`lodash.set` writes `key: undefined`, it does not delete the key.**

When a custom component clears, `value` is `undefined` here, and `set(clone, ['icon'], undefined)` produces `{ icon: undefined }` rather than `{}`. After `resetDependentFields` the item becomes e.g. `{ icon: undefined, useA: true }`.

The new test `converts null to undefined via array-field when nested custom component clears` asserts only `latest.views[0].icon).toBeUndefined()`, which is true for both `{}` and `{ icon: undefined }` — so it does not catch this.

This matches RJSF's form-level `_set` behavior, so it is consistent with the rest of the pipeline. But downstream consumers that check `'icon' in item` (rather than `item.icon === undefined`) will see a key that's still present. Worth verifying Lime CRM's form-data consumers don't care, or switching to `_unset` when `value === undefined`.

src/components/form/fields/array-field.ts
```diff
@@ -133,6 +141,69 @@ export class ArrayField extends React.Component<FieldProps> {
...
+     * @param newData
+     * @param path
+     */
+    private handleDeepLeafChange(newData: unknown, path: FieldPathList): void {
+        const { formData: oldData, fieldPathId } = this.props;
+
+        const [indexSegment, ...pathInItem] = path.slice(
+            fieldPathId.path.length
+        );
+        const index = Number(indexSegment);
```
FredrikWallstrom, line 177: **`Number(indexSegment)` silently produces `NaN` for non-numeric segments.**

If `indexSegment` is anything other than a numeric string/number, `Number(...)` returns `NaN` and `newArray[NaN] = ...` writes a `NaN`-keyed property — the array's `length` does not change, no error is thrown, but the form data is silently corrupted.

In normal RJSF flow the segment immediately after the array's path is always a numeric index, so this should not happen. But because `isDeepLeafChange` is the only gate and it does not validate that segment, any malformed path (e.g. an unusual `setFieldValue` call) would corrupt the array invisibly. A `Number.isInteger(index)` guard with an early return would harden the path.

src/components/form/fields/schema-field.ts
```diff
@@ -171,6 +177,21 @@ export class SchemaField extends React.Component<FieldProps> {
...
+     * through this SchemaField with a path that is deeper than its own.
+     * The enclosing ArrayField rebuilds the affected item and runs
+     * `resetDependentFields` at the item level (see array-field.ts), so we
+     * must pass the partial leaf data through untouched here.
+     * @param path
+     */
+    private isDeepLeafChange(path: FieldPathList): boolean {
+        const { fieldPathId } = this.props;
+
+        return !!fieldPathId && path.length > fieldPathId.path.length;
```
FredrikWallstrom, line 192: **The bypass is broader than "inside an array" — it also applies to deep changes inside a nested non-array object, where there's no compensation layer.**

The `ArrayField` change compensates for the bypass by re-running `resetDependentFields` at item level. But this `isDeepLeafChange` triggers for *any* path deeper than the field's own — including deep changes inside a nested object that has nothing to do with arrays.

Example: `{ properties: { config: { properties: { trigger }, dependencies: { trigger: ... } } } }`. When `config.trigger` changes, the `config` SchemaField sees a deep-leaf path and skips `resetDependentFields`. Sibling fields under `config` that became invalid for the new branch are not cleared.

This is largely a pre-existing limitation (pre-PR, `resetDependentFields` early-returned for scalar leaves and *corrupted* data for object leaves — which is the bug being fixed), so this PR is strictly better than before. Still worth confirming Lime CRM doesn't rely on dep-resets for nested-object trigger changes outside arrays — if it does, an analogous fix is needed at the object level (e.g. an override on `ObjectField`).

src/components/form/fields/array-field.ts
```diff
@@ -133,6 +141,69 @@ export class ArrayField extends React.Component<FieldProps> {
...
+        newData: unknown,
+        index: number,
+        pathInItem: Array<string | number>
+    ): unknown {
+        const { formData: oldData, schema } = this.props;
+        const { rootSchema } = this.props.registry;
+
+        const itemSchema = (
+            Array.isArray(schema.items) ? schema.items[index] : schema.items
+        ) as FormSchema;
```
FredrikWallstrom, line 194: **`schema.items` may be `undefined` (or `false` in JSON Schema 2020) — no guard.**

The `as FormSchema` cast hides this. If `schema.items` is `undefined`, or a tuple shorter than `index` with no `additionalItems` handling, `itemSchema` becomes `undefined` and is passed into `resetDependentFields(oldItem, newItem, undefined, rootSchema)`, which then calls `retrieveSchema(validator, undefined, ...)` and is likely to throw inside RJSF.

Not a likely hit for Lime CRM's hand-authored schemas, but since the PR description calls this out as an essential component, an explicit early-return when `itemSchema` is undefined would harden it against pathological input.

src/components/form/form.e2e.tsx
```diff
@@ -500,6 +504,68 @@ test('renders nested objects inside array items without inheriting array context
...
+        value: { views: [{}] },
+        onChange,
+    });
+
+    await change('Icon', 'star');
+    await change('Icon', undefined);
+
+    const latest = onChange.mock.lastCall[0].detail;
+    expect(latest.views[0].icon).toBeUndefined();
+});
```
FredrikWallstrom, line 567: **Test gap: cascading deep-leaf changes through nested arrays, and tuple items.**

The new tests cover one array level. The fix's `handleDeepLeafChange` will fire twice for `array > array > leaf` (once at each `ArrayField`), and the inner array's `cloneDeep` + `resetDependentFields` will run before the outer one re-clones. I traced this and believe it works, but a regression test with two array-nesting levels (or `array > object > array > leaf`) would lock in the cascade behavior. Also no coverage for tuple items (`items: [s1, s2]`), which take a different branch in `buildResetItem`.

FredrikWallstrom (commented):
* I'm not sure what to test. I only have one scenario, which is when Props are lost when you hit the save button in Lime Admin. That seems to be solved with this PR ✅ 
   * What more can I test, in practice?
* Since this is not super easy to understand for me, I will post some Claude comments.  Looks like nitpicks/minor comments, feel free to take action on them or not.
* Probably not related to this PR, but I have severe performance issues when saving in Lime Admin, when the schema is big.

jgroth:
> I'm not sure what to test. I only have one scenario, which is when Props are lost when you hit the save button in Lime Admin. That seems to be solved with this PR ✅
> * What more can I test, in practice?

That issue should be solved by the 3rd commit. The initial issue I wanted to fix was in commit 1 (you could not remove chips in nested array fields). 

image


The second commit was a bug found by coderabbit during the review, I'm not sure how to reproduce it in lime-admin, but there's a test added for it.

FredrikWallstrom:
> > I'm not sure what to test. I only have one scenario, which is when Props are lost when you hit the save button in Lime Admin. That seems to be solved with this PR ✅
> > 
> > * What more can I test, in practice?
> 
> That issue should be solved by the 3rd commit. The initial issue I wanted to fix was in commit 1 (you could not remove chips in nested array fields).
> 
> image
> The second commit was a bug found by coderabbit during the review, I'm not sure how to reproduce it in lime-admin, but there's a test added for it.

That seems to be working now ✅

FredrikWallstrom (approved):

PR #4053, opened 2026-04-27
fix(card): don't show hover effects when card is neither clickable no…

### LINKED ISSUE #4052: limel-card: remove onHover when 3d-effect and clickable = false
https://github.com/user-attachments/assets/2a18cd55-62f3-4fa6-8859-a35461134882

LucyChyzhova:
…r has 3D effect
fix: https://github.com/Lundalogik/lime-elements/issues/4052

https://github.com/user-attachments/assets/4e4ee17e-7fe3-49e5-a134-3fcea51b97e4

## Summary by CodeRabbit

* **Style**
  * Refined 3D and hover visuals so effects only apply to cards explicitly marked interactive, preventing unintended styling on non-interactive cards.
  * Restricted image saturation and transition triggers to cards with the interactive/3D setting, improving visual clarity and reducing accidental hover effects.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -39,12 +39,23 @@ section {
     padding: 0.25rem;
     background-color: var(--card-background-color, rgb(var(--contrast-300)));
 
+    :host(limel-card[clickable]) &,
+    :host(limel-card[show-3d-effect]) & {
+        @include mixins.the-3d-element;
+    }
+
+    :host(limel-card[clickable]) & {
+        @include mixins.the-3d-element--clickable;
+    }
+
     &:hover {
-        border-color: transparent;
-        background-color: var(
-            --card-background-color--hovered,
-            var(--card-background-color, rgb(var(--contrast-200)))
-        );
+        :host(limel-card[clickable]) & {
+            border-color: transparent;
+            background-color: var(
+                --card-background-color--hovered,
+                var(--card-background-color, rgb(var(--contrast-200)))
+            );
+        }
     }
 }
 
@@ -88,8 +99,8 @@ img {
     max-width: 100%;
     max-height: 100%;
 
-    section:hover &,
-    section:focus-visible & {
+    :host(limel-card[show-3d-effect]) section:hover &,
+    :host(limel-card[show-3d-effect]) section:focus-visible & {
         transition-duration: 0.2s;
         filter: saturate(1.3);
     }
@@ -215,14 +226,6 @@ limel-3d-hover-effect-glow {
 
 :host(limel-card[show-3d-effect]) {
     @include mixins.parent-of-the-3d-element;
-}
-
-section {
-    @include mixins.the-3d-element;
-
-    :host(limel-card[clickable]) & {
-        @include mixins.the-3d-element--clickable;
-    }
 }
 
 :host(limel-card[selected]) section {


src/components/card/card.scss
```diff
@@ -38,14 +38,15 @@ section {
...
-        );
-    }
+:host(limel-card[clickable]) section:hover,
+:host(limel-card[show-3d-effect]) section:hover {
+    border-color: transparent;
+    background-color: var(
+        --card-background-color--hovered,
+        var(--card-background-color, rgb(var(--contrast-200)))
+    );
 }
```
Kiarokh:
```suggestion
	&:hover {
		:host(limel-card[clickable]) & {
		    border-color: transparent;
		    background-color: var(
		        --card-background-color--hovered,
		        var(--card-background-color, rgb(var(--contrast-200)))
		    );
	    }
    }
}
```
I might be missing something. But this change is only for the `clickable` prop, and must stay focused on that. Or is there a reason to mix them both?

Also, to stay consistent with how we write other components styles, and because these styles are particularly focusing on `section` I'd prefer to have a nested SCSS style, rather than closing the previous block and opening a new one. In this part of the code, we're still talking about the `section` and how it should behave and look like.

src/components/card/card.scss
```diff
@@ -88,7 +89,8 @@ img {
     max-width: 100%;
     max-height: 100%;
 
-    section:hover &,
+    :host(limel-card[clickable]) section:hover &,
+    :host(limel-card[show-3d-effect]) section:hover &,
     section:focus-visible & {
```
Kiarokh:
```suggestion
    :host(limel-card[show-3d-effect]) section:hover &,
    :host(limel-card[show-3d-effect]) section:focus-visible & {
```

This is only about the `show-3d-effect` not the `clickable`

src/components/card/card.scss
```diff
@@ -217,12 +219,13 @@ limel-3d-hover-effect-glow {
...
+:host(limel-card[show-3d-effect]) section {
     @include mixins.the-3d-element;
+}
 
-    :host(limel-card[clickable]) & {
-        @include mixins.the-3d-element--clickable;
-    }
+:host(limel-card[clickable]) section {
+    @include mixins.the-3d-element--clickable;
 }
```
Kiarokh: can you please move these up the the `section` block?

src/components/card/card.scss
LucyChyzhova: yes, seems to work like this as well

src/components/card/card.scss
```diff
@@ -88,7 +99,7 @@ img {
     max-width: 100%;
     max-height: 100%;
 
-    section:hover &,
+    :host(limel-card[show-3d-effect]) section:hover &,
     section:focus-visible & {
```
Kiarokh:
```suggestion
    :host(limel-card[show-3d-effect]) section:focus-visible & {
```

Kiarokh (approved):

PR #4056, opened 2026-04-30
feat(image): support referrerpolicy on rendered images

civing:
## Summary

- Adds an optional `referrerpolicy?: ReferrerPolicy` field to the shared `Image` interface (`src/global/shared-types/image.types.ts`) + the api report.
- Adds an internal `ImageTemplate` functional component (`src/util/image.template.tsx`, `@internal`, not exported) that renders an `` from an `Image` — forwarding `src`, `alt`, `loading` (`image.loading ?? 'lazy'`), and `referrerPolicy` via the **typed** Stencil prop (no cast/workaround). It also forwards optional `style` and `onError` for consumers that need to control presentation or react to a load failure.
- Routes `limel-card`, `limel-chip`, `limel-list-item`, `limel-file-viewer`, and `limel-profile-picture` through `ImageTemplate`, removing each component's hand-rolled ``.
- `limel-profile-picture` and `limel-file-viewer` set `referrerpolicy="no-referrer"` (their images point at third-party avatar / file hosts).
- Adds e2e coverage on `limel-list-item` (referrerpolicy set/absent, loading default/eager) and `limel-profile-picture` (no-referrer + object-fit style + image-error state).

When `referrerpolicy` is omitted, the attribute is not emitted — existing consumers are unaffected.

## Why

When an image `src` points at a third-party service, the browser otherwise sends the originating CRM page URL in the `Referer` request header — leaking customer-relationship metadata (which tenant viewed which record while researching which company) to whoever runs that service. `referrerpolicy="no-referrer"` removes the leak at no functional cost: the image still loads, the third party still receives the request, but no originating-page URL is attached. This PR gives `Image`-shaped props a first-class way to opt into that, and makes the affected components honor it.

## Context (downstream consumers)

Upstream prerequisite for [`lime-crm-components` PR #4122](https://github.com/Lundalogik/lime-crm-components/pull/4122) ("Company favicons"), which loads favicons from Google's S2 favicon service (a third party) and sets `referrerpolicy` via the `Image` objects it passes into `limel-list-item` / `limel-chip` / `limel-card`. Originating request: [`hack-tuesday` #195](https://github.com/Lundalogik/hack-tuesday/issues/195).

## Implementation notes

- **Helper naming.** `ImageTemplate` (a `FunctionalComponent`) is named distinctly from the existing private `renderImage` methods on individual components, so the shared, stateless render path is unambiguous. It is `@internal` and not exported from the public API.
- **No typing workaround.** `referrerPolicy` is the typed Stencil JSX prop, used directly with no cast — the installed `@stencil/core` (4.43.5) declares `referrerPolicy` on `ImgHTMLAttributes` (the fix for stenciljs/core#6692).
- **`limel-profile-picture`** renders through `ImageTemplate`; the helper forwards its `style` binding (the `--limel-profile-picture-object-fit` custom property) and its `onError` handler, so behavior is unchanged.
- **`limel-file-viewer`** switched its image branch to `ImageTemplate` (wrapped in a `<Fragment>`; passes `alt: this.alt ?? ''` since the component's `alt` is optional but `Image.alt` is required).

### A note on `limel-file-viewer` and `referrerpolicy`

`limel-file-viewer` now renders images with `referrerpolicy="no-referrer"` unconditionally. We considered exposing a `referrerPolicy` prop as an escape hatch but **skipped it for now**: the main consumer (Lime CRM) does not render SharePoint/third-party images through file-viewer's `` path — SharePoint Live Docs previews use an `<iframe>`, thumbnails render via building-blocks' own ``, and the file URLs file-viewer *does* render are same-origin CRM links, where `no-referrer` is a harmless no-op. We can add a `referrerPolicy` prop later if a consumer ever needs to override it.

### Out of scope / follow-up

`limel-card` / `limel-chip` / `limel-list-item` render `image.src` without URL sanitization (pre-existing; not a script-execution vector for ``, but inconsistent with `limel-file-viewer`). Tracked for a separate PR in #4143.

## Commit structure

Eight commits:

1. `feat(image): add referrerpolicy to the Image interface` — type + api report.
2. `chore(image): add ImageTemplate helper for shared image rendering` — internal helper.
3. `refactor(card,chip,list-item): render Image via ImageTemplate helper` — wires three components + e2e.
4. `refactor(file-viewer): render image via ImageTemplate helper`.
5. `fix(profile-picture): suppress referrer on rendered `.
6. `refactor(image): forward style and onError through ImageTemplate` — widens the helper.
7. `refactor(profile-picture): render avatar via ImageTemplate helper` — migrates onto the widened helper + e2e.
8. `fix(file-viewer): suppress referrer on rendered `.

## Test plan

- e2e on `limel-list-item`: `referrerpolicy` present when set / absent when unset; `loading` defaults to `lazy` and honors an explicit `eager`.
- e2e on `limel-profile-picture`: the avatar renders with `no-referrer`, lazy loading, and the object-fit style var; an image `error` flips the `has-image-error` host state.
- `npm run build` passes; `npm run api:update` produces no diff beyond the documented `referrerpolicy` addition.
- `limel-card` / `limel-chip` / `limel-file-viewer` are exercised through the same shared `ImageTemplate` path.

## Summary by CodeRabbit

* **New Features**
  * Added support for referrer policy configuration on images to enhance privacy.
  * Enabled customizable image loading behavior (lazy vs. eager).

* **Refactor**
  * Consolidated image rendering logic across components for improved consistency.

* **Tests**
  * Added comprehensive tests to verify image attribute handling and behavior.

### DIFF
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -1255,6 +1255,7 @@ export type IconSize = 'x-small' | 'small' | 'medium' | 'large';
 interface Image_2 {
     alt: string;
     loading?: 'lazy' | 'eager';
+    referrerpolicy?: ReferrerPolicy;
     src: string;
 }
 export { Image_2 as Image }
--- a/src/components/card/card.tsx
+++ b/src/components/card/card.tsx
@@ -9,6 +9,7 @@ import {
     State,
 } from '@stencil/core';
 import { Image } from '../../global/shared-types/image.types';
+import { ImageTemplate } from '../../util/image.template';
 import { Icon, IconName } from '../../global/shared-types/icon.types';
 import { isItem } from '../action-bar/is-item';
 import { getIconName } from '../icon/get-icon-props';
@@ -243,7 +244,7 @@ export class Card {
 
         return (
             <div class="image-wrapper">
-                <img src={this.image.src} alt={this.image.alt} loading="lazy" />
+                <ImageTemplate image={this.image} />
             </div>
         );
     }
--- a/src/global/shared-types/image.types.ts
+++ b/src/global/shared-types/image.types.ts
@@ -20,4 +20,13 @@ export interface Image {
      * - `eager` means that the image will be loaded as soon as possible.
      */
     loading?: 'lazy' | 'eager';
+
+    /**
+     * The `referrerpolicy` attribute of the image. Set to `'no-referrer'`
+     * when `src` points to a third-party service that should not receive
+     * the originating page URL in the `Referer` header (e.g. external
+     * favicon or avatar services). When omitted, the attribute is not
+     * emitted and the browser's default referrer policy applies.
+     */
+    referrerpolicy?: ReferrerPolicy;
 }
--- a/src/components/chip/chip.tsx
+++ b/src/components/chip/chip.tsx
@@ -8,6 +8,7 @@ import {
     Prop,
 } from '@stencil/core';
 import { Icon, IconName } from '../../global/shared-types/icon.types';
+import { ImageTemplate } from '../../util/image.template';
 import { Languages } from '../date-picker/date.types';
 import { Link } from '../../global/shared-types/link.types';
 import { getRel } from '../../util/link-helper';
@@ -295,9 +296,7 @@ export class Chip implements ChipInterface {
         }
 
         if (!isEmpty(this.image)) {
-            return (
-                <img src={this.image.src} alt={this.image.alt} loading="lazy" />
-            );
+            return <ImageTemplate image={this.image} />;
         }
 
         return (
--- a/src/components/file-viewer/file-viewer.tsx
+++ b/src/components/file-viewer/file-viewer.tsx
@@ -245,6 +245,7 @@ export class FileViewer {
                     image={{
                         src: this.sanitizeUrl(this.fileUrl),
                         alt: this.alt ?? '',
+                        referrerpolicy: 'no-referrer',
                     }}
                 />
             </Fragment>
--- a/src/components/file-viewer/file-viewer.tsx
+++ b/src/components/file-viewer/file-viewer.tsx
@@ -1,6 +1,7 @@
 import {
     Component,
     Element,
+    Fragment,
     h,
     Prop,
     State,
@@ -17,6 +18,7 @@ import { FileType, OfficeViewer } from './file-viewer.types';
 import { LimelMenuCustomEvent } from '../../components';
 import { Email } from '../email-viewer/email-viewer.types';
 import { loadEmail } from '../email-viewer/email-loader';
+import { ImageTemplate } from '../../util/image.template';
 
 /**
  * This is a smart component that automatically detects
@@ -236,14 +238,17 @@ export class FileViewer {
     };
 
     private renderImage = () => {
-        return [
-            this.renderButtons(),
-            <img
-                src={this.sanitizeUrl(this.fileUrl)}
-                alt={this.alt}
-                loading="lazy"
-            />,
-        ];
+        return (
+            <Fragment>
+                {this.renderButtons()}
+                <ImageTemplate
+                    image={{
+                        src: this.sanitizeUrl(this.fileUrl),
+                        alt: this.alt ?? '',
+                    }}
+                />
+            </Fragment>
+        );
     };
 
     private renderVideo = () => {
--- a/src/components/list-item/list-item.e2e.tsx
+++ b/src/components/list-item/list-item.e2e.tsx
@@ -108,5 +108,41 @@ describe('limel-list-item', () => {
 
         const imgEl = root.querySelector('img');
         expect(imgEl).not.toBeNull();
+        expect(imgEl?.hasAttribute('referrerpolicy')).toBe(false);
+        expect(imgEl?.getAttribute('loading')).toEqual('lazy');
+    });
+
+    it('forwards referrerpolicy to the rendered image when set', async () => {
+        const imgSrc =
+            'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///ywAAAAAAQABAAACAUwAOw==';
+        const { root, waitForChanges } = await render(
+            <limel-list-item
+                text="Pic"
+                image={{
+                    src: imgSrc,
+                    alt: 'a',
+                    referrerpolicy: 'no-referrer',
+                }}
+            ></limel-list-item>
+        );
+        await waitForChanges();
+
+        const imgEl = root.querySelector('img');
+        expect(imgEl?.getAttribute('referrerpolicy')).toEqual('no-referrer');
+    });
+
+    it('applies the image loading strategy when set', async () => {
+        const imgSrc =
+            'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///ywAAAAAAQABAAACAUwAOw==';
+        const { root, waitForChanges } = await render(
+            <limel-list-item
+                text="Pic"
+                image={{ src: imgSrc, alt: 'a', loading: 'eager' }}
+            ></limel-list-item>
+        );
+        await waitForChanges();
+
+        const imgEl = root.querySelector('img');
+        expect(imgEl?.getAttribute('loading')).toEqual('eager');
     });
 });
--- a/src/components/list-item/list-item.tsx
+++ b/src/components/list-item/list-item.tsx
@@ -9,6 +9,7 @@ import { ListSeparator } from '../../global/shared-types/separator.types';
 import { CheckboxTemplate } from '../checkbox/checkbox.template';
 import translate from '../../global/translations';
 import { Languages } from '../date-picker/date.types';
+import { ImageTemplate } from '../../util/image.template';
 
 /**
  * This components displays the list item.
@@ -253,7 +254,7 @@ export class ListItemComponent implements ListItem {
             return;
         }
 
-        return <img src={this.image.src} alt={this.image.alt} loading="lazy" />;
+        return <ImageTemplate image={this.image} />;
     };
 
     private renderActionMenu = (actions: Array<MenuItem | ListSeparator>) => {
--- a/src/components/profile-picture/profile-picture.e2e.tsx
+++ b/src/components/profile-picture/profile-picture.e2e.tsx
@@ -1,0 +1,38 @@
+import { render, h } from '@stencil/vitest';
+
+describe('limel-profile-picture', () => {
+    const imgSrc =
+        'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///ywAAAAAAQABAAACAUwAOw==';
+
+    it('renders the avatar with no-referrer, lazy loading and the object-fit style', async () => {
+        const { root, waitForChanges } = await render(
+            <limel-profile-picture
+                value={imgSrc}
+                imageFit="contain"
+            ></limel-profile-picture>
+        );
+        await waitForChanges();
+
+        const imgEl = root.shadowRoot.querySelector('img');
+        expect(imgEl).not.toBeNull();
+        expect(imgEl?.getAttribute('referrerpolicy')).toEqual('no-referrer');
+        expect(imgEl?.getAttribute('loading')).toEqual('lazy');
+        expect(
+            imgEl?.style.getPropertyValue('--limel-profile-picture-object-fit')
+        ).toEqual('contain');
+    });
+
+    it('flags the host with has-image-error when the image fails to load', async () => {
+        const { root, waitForChanges } = await render(
+            <limel-profile-picture value={imgSrc}></limel-profile-picture>
+        );
+        await waitForChanges();
+
+        expect(root.classList.contains('has-image-error')).toBe(false);
+
+        root.shadowRoot.querySelector('img')?.dispatchEvent(new Event('error'));
+        await waitForChanges();
+
+        expect(root.classList.contains('has-image-error')).toBe(true);
+    });
+});
--- a/src/components/profile-picture/profile-picture.tsx
+++ b/src/components/profile-picture/profile-picture.tsx
@@ -254,6 +254,7 @@ export class ProfilePicture {
                     }}
                     loading="lazy"
                     onError={this.onImageError}
+                    referrerPolicy="no-referrer"
                 />
             );
         }
--- a/src/components/profile-picture/profile-picture.tsx
+++ b/src/components/profile-picture/profile-picture.tsx
@@ -16,6 +16,7 @@ import translate from '../../global/translations';
 import { Languages } from '../date-picker/date.types';
 import { createRandomString } from '../../util/random-string';
 import { resizeImage, ResizeOptions } from '../../util/image-resize';
+import { ImageTemplate } from '../../util/image.template';
 
 /**
  * This component displays a profile picture, while allowing the user
@@ -246,15 +247,12 @@ export class ProfilePicture {
 
         if (src) {
             return (
-                <img
-                    src={src}
-                    alt=""
+                <ImageTemplate
+                    image={{ src, alt: '', referrerpolicy: 'no-referrer' }}
                     style={{
                         '--limel-profile-picture-object-fit': this.imageFit,
                     }}
-                    loading="lazy"
                     onError={this.onImageError}
-                    referrerPolicy="no-referrer"
                 />
             );
         }
--- a/src/util/image.template.tsx
+++ b/src/util/image.template.tsx
@@ -1,0 +1,29 @@
+import { FunctionalComponent, h } from '@stencil/core';
+import { Image } from '../global/shared-types/image.types';
+
+interface ImageTemplateProps {
+    image: Image;
+}
+
+/**
+ * Renders an `Image` as a plain `<img>` element. Centralises the
+ * attribute forwarding so consumer components do not have to
+ * reimplement it. Intended for internal use by lime-elements
+ * components that accept an `Image`-shaped prop.
+ *
+ * @param props
+ * @param props.image - the image to render
+ * @internal
+ */
+export const ImageTemplate: FunctionalComponent<ImageTemplateProps> = ({
+    image,
+}) => {
+    return (
+        <img
+            src={image.src}
+            alt={image.alt}
+            loading={image.loading ?? 'lazy'}
+            referrerPolicy={image.referrerpolicy}
+        />
+    );
+};
--- a/src/util/image.template.tsx
+++ b/src/util/image.template.tsx
@@ -3,20 +3,39 @@ import { Image } from '../global/shared-types/image.types';
 
 interface ImageTemplateProps {
     image: Image;
+
+    /**
+     * Inline styles to set on the rendered `<img>`, e.g. to drive
+     * `object-fit` through a custom property.
+     */
+    style?: { [key: string]: string | undefined };
+
+    /**
+     * Handler invoked when the image fails to load (the `<img>`
+     * `error` event), letting the host react to a broken or blocked
+     * source.
+     */
+    onError?: (event: Event) => void;
 }
 
 /**
  * Renders an `Image` as a plain `<img>` element. Centralises the
  * attribute forwarding so consumer components do not have to
- * reimplement it. Intended for internal use by lime-elements
- * components that accept an `Image`-shaped prop.
+ * reimplement it. Optional `style` and `onError` are
+ * forwarded to the element for consumers that also need to control
+ * its presentation or react to load failures. Intended for internal
+ * use by lime-elements components that accept an `Image`-shaped prop.
  *
  * @param props
  * @param props.image - the image to render
+ * @param props.style - inline styles for the `<img>`
+ * @param props.onError - handler for the `<img>` `error` event
  * @internal
  */
 export const ImageTemplate: FunctionalComponent<ImageTemplateProps> = ({
     image,
+    style,
+    onError,
 }) => {
     return (
         <img
@@ -24,6 +43,8 @@ export const ImageTemplate: FunctionalComponent<ImageTemplateProps> = ({
             alt={image.alt}
             loading={image.loading ?? 'lazy'}
             referrerPolicy={image.referrerpolicy}
+            style={style}
+            onError={onError}
         />
     );
 };


etc/lime-elements.api.md
adrianschmidt, line 1: The second commit seems like it should be a `chore`, and the third commit a `refactor`? Otherwise this looks very good. I'll post a full review in a comment.

adrianschmidt:
# Consolidated PR Review -- 6 Dimensions

> Generated by a 6-agent review (backward compatibility, code quality, architecture, security, observability, performance) running in parallel against this PR's branch.

## PR Summary

This PR adds an optional `referrerpolicy?: ReferrerPolicy` field to the shared `Image` type, introduces a `renderImg(image: Image)` helper centralising `` JSX (with a documented spread-cast workaround for Stencil's typing gap upstream-tracked at stenciljs/core#6692), and switches `limel-card`, `limel-chip`, and `limel-list-item` to render via the helper. 7 files changed, +66/-9. Two new e2e tests on `limel-list-item` cover the attribute-set and attribute-absent branches. Unblocks `lime-crm-components` PR #4122 (company favicons), where security review flagged that 3 of 4 `Image` consumer surfaces had no plumbing for `referrerpolicy`.

## Merge Readiness -- READY TO MERGE :white_check_mark:

The PR is strictly better than `main` along every reviewed dimension: it adds capability without regressing emitted DOM, public types, or runtime cost, and it's the upstream prerequisite for a real information-disclosure fix downstream. No `[High]` findings from any agent.

- **Blockers:** None.
- **Non-blocking but worth addressing:** Two `[Medium]` items in Architecture worth picking up — helper placement (`shared-types/` vs. `src/util/` or a `*.template.tsx` colocation), and the silent `image.loading` drop the helper inherits from the inline JSX. See Top Recommendations.

---

## 1. Backward Compatibility -- SAFE :white_check_mark:

Adding an optional field to a public interface and replacing inline `` JSX with an internal helper is a clean, additive change. Emitted DOM is identical when `referrerpolicy` is unset; helper is intentionally not re-exported from `src/interface.ts:63`, so it stays internal and absent from `etc/lime-elements.api.md`.

- **[Low]** `renderImg` has no `@internal` TSDoc tag (`src/global/shared-types/image.tsx:13`). It's not surfaced today because `interface.ts` only re-exports `image.types`, but a future `export *` over `shared-types/` would leak it. Adding `@internal` would make the contract explicit.
- **[Low]** Co-locating `image.tsx` next to `image.types.ts` creates two same-base-name files differing only by extension (`card.tsx:12`, `chip.tsx:11`, `list-item.tsx:11` import via the bare specifier `'../../global/shared-types/image'`). Resolves unambiguously today, but mildly confusing for maintainers.
- **[Low]** `Image.loading` was unused pre-PR and remains unused (helper hardcodes `loading="lazy"`). Pre-existing; not a regression. (Also flagged stronger by Architecture.)

**Positives:** Strictly additive public-type change. `card`'s `<div class="image-wrapper">` preserved at `card.tsx:245` so existing `card.scss:69` selectors keep matching. The "absent" assertion at `list-item.e2e.tsx:111` guards the no-emission contract that protects existing consumers. `ReferrerPolicy` is a global from `lib.dom.d.ts` (in tsconfig `lib`), so no downstream typing setup needed. Three atomic commits make any future bisect/revert clean.

## 2. Code Quality -- GOOD :white_check_mark:

Solid, focused change. Helper is small but justified by the spread-cast workaround; e2e tests are real (one per branch); JSDoc on the new type field explains *why* a consumer would want `'no-referrer'`.

- **[Low]** `image.tsx:27` casts to `Record<string, string>`, which is wider than necessary and discards the `ReferrerPolicy` union typing. A precise `Record<'referrerpolicy', ReferrerPolicy>` (or a typed local) would suppress the Stencil typing mismatch without pretending the value is arbitrary.
- **[Low]** Truthy check on `image.referrerpolicy` (`image.tsx:18`) skips emission for `''`, which is technically a valid `ReferrerPolicy` literal meaning "use default". Functionally equivalent to omitting the attribute, so not a bug — but the JSDoc on `image.types.ts:31` could note it.
- **[Low]** Helper-level test coverage lives in `list-item.e2e.tsx` only; `card` and `chip` got no new tests. A co-located `src/global/shared-types/image.spec.tsx` (matching the `src/util/*.spec.ts` convention) would more accurately reflect where the logic lives. Current coverage is adequate.
- **[Low]** Helper location: `shared-types/` previously held only `*.types.ts`. Putting a JSX helper there blurs the directory's purpose.
- **[Low]** Naming: `renderImg` resolves the real collision with private `renderImage` methods in `card.tsx:240` and `list-item.tsx:259`, but reads slightly less consistently than the rest of the codebase. `renderImageElement` or an alias-at-import would also work.

**Positives:** Eliminates real triplication (`card.tsx`, `chip.tsx`, `list-item.tsx` previously each had their own `` JSX). The Stencil-typing escape hatch is isolated to one call site and clearly comment-flagged with the upstream issue link, so it's removable when typing lands. The "absent" e2e assertion is a useful regression guard. API report (`etc/lime-elements.api.md:1240`) updated, making the type extension a deliberate public-API change rather than drift.

## 3. Architecture -- ACCEPTABLE :warning:

The refactor is sound for the three call sites (verified to be the only `Image`-typed surfaces in this repo). Main concerns are the helper's directory placement and one missed opportunity to fix a pre-existing type/behavior mismatch.

- **[Medium]** Helper placement violates the directory's intent. `src/global/shared-types/image.tsx` is the only `.tsx`/JSX-emitting file in `shared-types/` (every sibling — `color.types.ts`, `icon.types.ts`, `file.types.ts`, etc. — is a pure type module) and the first file there to import `@stencil/core`, turning a type-only module into a runtime-bearing one. Two existing conventions fit better: `src/components/<x>/<x>.template.tsx` exporting a `FunctionalComponent` (see `checkbox/checkbox.template.tsx`, `select/select.template.tsx`) or `src/util/` (already referenced from `interface.ts:68` via `image-resize`). A `FunctionalComponent` would also let consumers write `<ImageTemplate image={this.image} />` instead of `{renderImg(this.image)}`, which is more idiomatic Stencil.
- **[Medium]** `renderImg` silently drops `image.loading`. `Image.loading?: 'lazy' | 'eager'` is documented on the public type (`image.types.ts:22`), but `image.tsx:26` hardcodes `loading="lazy"`. Pre-existing in the inline JSX, so not a regression — but a centralising PR was the natural place to fix the type/behavior mismatch. Either honour `image.loading ?? 'lazy'` or remove `loading` from the interface.
- **[Low]** Spread-cast pattern won't scale gracefully — a fifth attribute (`crossorigin`, `decoding`, `fetchpriority`) needing the same workaround would duplicate. A `buildOptionalAttrs(image)` helper or single accumulating object would be more extensible. Not blocking.
- **[Low]** Inconsistent "is this image renderable?" guards: `card.tsx:241` uses `!this.image?.src`, `chip.tsx:298` uses `!isEmpty(this.image)` (lodash), `list-item.tsx:260` uses `!this.image`. Pre-existing; the helper could absorb the guard (`return null` when `!image?.src`) and normalize the contract.
- **[Low]** Public-vs-internal export depends on `interface.ts` enumerating files rather than a structural barrier. `src/util/` or `*.template.tsx` would make the boundary structural.

**Positives:** Correctly identifies all `Image`-typed surfaces in the repo (other `` uses in `profile-picture.tsx`, `file-viewer.tsx`, `email-viewer/*` are on non-`Image`-typed inputs and rightly excluded). Wrapping decisions stay at the call site (`card`'s `image-wrapper` is a card-specific layout concern). DRYs up triplication without taking over component-level conditional-render decisions.

## 4. Security -- SAFE :white_check_mark:

This PR is itself a security improvement: it adds the plumbing for downstream consumers to suppress `Referer` leakage for `` elements pointing at third-party services. No new vulnerabilities introduced.

- **[Low]** Runtime values of `image.referrerpolicy` are not validated (the `ReferrerPolicy` constraint is TS-only). Browsers ignore unknown tokens per spec — no XSS surface — but consumers passing untrusted strings should know an invalid value silently equals "no leak protection."
- **[Low]** Only `list-item` has e2e assertions for both branches; `card` and `chip` have none. A future change to `renderImg` could regress those silently. Defense-in-depth.

**Spread-cast attribute-injection check:** `referrerPolicyAttr` is a fresh object literal with one hard-coded key (`image.tsx:18-20`). JSX spread of a fresh object literal enumerates only own properties, so attacker-controlled `image.referrerpolicy` cannot inject sibling attributes (`onerror`, `src`, etc.) even under hypothetical prototype pollution. The `as Record<string, string>` cast is type-only. **Default-behavior verification:** When `referrerpolicy` is omitted, the helper produces `{}` and spreads nothing, so the `` is emitted with exactly the same attributes as before this PR. No change to default leakage potential — matching the PR's stated intent.

**Positives:** Centralizing forwarding gives one place for future security-relevant attributes (`crossorigin`, `decoding`) to land. Conservative emission avoids any change to existing rendered output. Truthy guard correctly handles `''`/`null`/`undefined`. Unblocks `lime-crm-components` #4122 to suppress `Referer` leakage to Google's favicon endpoint — a concrete information-disclosure improvement.

## 5. Observability -- SAFE :white_check_mark:

Pure UI plumbing: synchronous JSX-emitting helper with no failure modes, no I/O, no async behavior. Nothing meaningful to log, instrument, or correlate; no observability regression introduced. lime-elements has no project-wide logger/telemetry module the helper could be bypassing by living in `shared-types/`.

- **[Low]** None of the three call sites wire `onError`/`onLoad` on the rendered `` (`card.tsx:240-246`, `chip.tsx:291-300`, `list-item.tsx:259-265`), so a 404, CSP block, or `no-referrer` rejection still fails silently. Pre-existing — verified the inline JSX had the same shape — and `renderImg` is now the single right place to add broken-image telemetry if Lime ever wants it.
- **[Low]** `renderImg` doesn't defensively handle `image == null`. All current call sites guard, but a future caller forgetting the guard would get a `TypeError` at render time with no helper-level message. A one-line `if (!image?.src) return null;` would make misuse self-diagnosing.

**Positives:** New e2e tests at `list-item.e2e.tsx:108-131` give a clean regression signal in both directions. Inline comment on the spread-cast at `image.tsx:14-17` documents *why* and links the upstream Stencil issue.

## 6. Performance -- SAFE :white_check_mark:

Near-zero-cost refactor. Bundle, tree-shaking, and JSX equivalence all check out. Per-render cost is one extra small object allocation + spread per `` rendered, dwarfed by the vdom allocation Stencil already does for the same element.

- **[Low]** `renderImg` allocates a `{}` (or `{ referrerpolicy }`) per call (`image.tsx:18-27`). Each `limel-list-item` is its own Stencil component, so a list of N items produces N calls only when those items actually re-render — not a hot inner loop within a single parent render. Not measurable; don't optimize.
- **[Low]** The empty-object branch churns a fresh `{}` each call in the common case. A module-level `const EMPTY = {}` would reuse the literal, but this is textbook micro-optimization.

**Bundle/tree-shaking checks (all clean):** `import { h } from '@stencil/core'` is already in every Stencil component bundle, so zero new runtime dependency. The `Image` type is type-only and compiles away — no runtime import of `image.types.ts` is introduced. `renderImg` is a single named export with no side effects, so tree-shakers strip it for non-importers. JSX output is structurally equivalent to the previous inline JSX (`card`'s `<div class="image-wrapper">` is retained at `card.tsx:245`).

**Positives:** Centralisation gives one place to tune `loading`/`referrerpolicy` in the future. Helper is invoked after the early-return for missing `image`, so absent images cost nothing extra.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Code Quality | **GOOD** |
| Architecture | **ACCEPTABLE** |
| Security | **SAFE** |
| Observability | **SAFE** |
| Performance | **SAFE** |

## Top Recommendations

1. **[Medium from Architecture]** Move `renderImg` out of `src/global/shared-types/`. The directory previously held only `*.types.ts` files; introducing JSX + a `@stencil/core` runtime import there is the first crossing of that line. Two existing conventions fit: `src/util/render-image.tsx` (matches `image-resize` re-exported via `interface.ts:68`) or a `*.template.tsx` colocated near a consumer (matches `checkbox/checkbox.template.tsx`). The `*.template.tsx` form would also let it become a `FunctionalComponent` and be invoked as `<ImageTemplate image={this.image} />`, which is more idiomatic Stencil.
2. **[Medium from Architecture]** Decide what `Image.loading` means. The field is documented on the public type (`image.types.ts:22`) but `image.tsx:26` hardcodes `loading="lazy"` and ignores it (the inline JSX did the same — so this PR doesn't regress, but it's the natural place to fix). Either `loading={image.loading ?? 'lazy'}` in the helper, or remove `loading` from the public `Image` interface.
3. **[Low from Code Quality]** Tighten the cast at `image.tsx:27`: replace `as Record<string, string>` with `Record<'referrerpolicy', ReferrerPolicy>` (or a small typed local) so the `ReferrerPolicy` union typing isn't widened to arbitrary strings.
4. **[Low from Backward Compatibility]** Add an `@internal` TSDoc tag on `renderImg` (`image.tsx:13`) to make the public-vs-internal contract explicit and tracked under any future strict-trim rollup, since the boundary currently relies only on `interface.ts` not enumerating the file.
5. **[Low from Code Quality]** Mention in the JSDoc on `image.types.ts:31` that `referrerpolicy: ''` is treated the same as omitting the attribute, since the helper's truthy check skips emission for `''` (a technically-valid `ReferrerPolicy` literal). Alternatively, change the helper guard to `image.referrerpolicy !== undefined` for strict pass-through.
6. **[Low from Observability]** Optionally add `if (!image?.src) return null;` at the top of `renderImg` so a future caller forgetting the call-site guard gets a self-diagnosing no-op instead of a `TypeError`. All current callers already guard, so this is purely defensive.
7. **[Low from Security]** Consider mirroring the attribute-presence/absence assertions to `card` and `chip` e2e tests for defense-in-depth, so a future change to `renderImg` can't silently regress the no-emission contract on those two consumers.

---

:robot: Generated by [`/6-agent-review`](https://github.com/Lundalogik/lime-elements/pull/4056).

etc/lime-elements.api.md
Kiarokh: > The second commit seems like it should be a `chore`, and the third commit a `refactor`? Otherwise this looks very good. I'll post a full review in a comment.

I've noticed that Claude has a different understanding of the `feat` and `fix`. It considered only Bug fixes as fix, (and marks "_problems that are now fixed_" as feat)

Kiarokh:
we have a bunch of places we use the `<img` tag in the Building Blocks too. I think updating them would be a really nice next step. That'll also reduce the concerns (in the favicons PR) that initiated this PR 👍

src/global/shared-types/image.types.ts
```diff
@@ -20,4 +20,13 @@ export interface Image {
      * - `eager` means that the image will be loaded as soon as possible.
      */
     loading?: 'lazy' | 'eager';
+
+    /**
+     * The `referrerpolicy` attribute of the image. Set to `'no-referrer'`
+     * when `src` points to a third-party service that should not receive
+     * the originating page URL in the `Referer` header (e.g. external
+     * favicon or avatar services). When omitted, the attribute is not
+     * emitted and the browser's default referrer policy applies.
```
Kiarokh, line 29: shouldn't we always default to `'no-referrer'`?

src/components/card/card.tsx
Kiarokh, line 1: limel-file-viewer can also display images. perhaps it could use this feature too

etc/lime-elements.api.md
civing: Good call — applied:

- Commit 2 → `chore(image): add ImageTemplate helper for shared image rendering`
- Commit 3 → `refactor(card,chip,list-item): render Image via ImageTemplate helper`

Also added a 4th commit (`refactor(file-viewer)`) applying the same helper to `limel-file-viewer`.

civing:
Thanks for the thorough review. Per-finding writeup of what landed in this push:

### Addressed

- **Architecture 1 (helper placement)** — moved to `src/util/image.template.tsx`, converted to `ImageTemplate` `FunctionalComponent`, invoked as `<ImageTemplate image={...} />`.
- **Architecture 2 (`image.loading` dropped)** — helper now honors `image.loading ?? 'lazy'`. The field was on the public `Image` type but silently ignored across all three render sites before this PR.
- **Backward Compatibility 1 / `@internal` tag** — added on `ImageTemplate`.
- **Code Quality (cast tightening)** — variable typed as `{ referrerpolicy?: Image['referrerpolicy'] }`; spread-site cast kept as the minimum needed to satisfy Stencil's missing JSX typing.
- **Commit types (from your line comment)** — commit 2 → `chore`, commit 3 → `refactor`; helper rename `renderImg` → `ImageTemplate` reflected in messages.

### Extra in this push

- 4th commit: `refactor(file-viewer)` switches `limel-file-viewer`'s image branch to `ImageTemplate`. Same shape applies cleanly. `profile-picture` was considered but skipped — it carries `style`, `onError`, and custom logic that doesn't fit the helper's contract.

### Deferred (open to doing if you want)

- Card/chip e2e parity for the attribute-set/absent branches (`list-item` already has both; the helper is now the single rendering site so coverage is transitive).
- Defensive `if (!image?.src) return null;` in the helper.
- JSDoc note that `referrerpolicy: ''` is treated as omitted.

### Skipped (with rationale)

- Runtime `ReferrerPolicy` value validation — browsers ignore unknown tokens per spec; silent fallback is reasonable.
- `onError`/`onLoad` telemetry — pre-existing absence across all sites; the helper is the right place to add later if you ever want it.
- Allocation micro-optimization — per "don't optimize" guidance.
- `buildOptionalAttrs` factoring — single attribute today; revisit if a second needs the same workaround.

src/components/card/card.tsx
civing: Done!

src/global/shared-types/image.types.ts
civing: It would be the natural privacy-positive default, but it'd be observable across every `ImageTemplate` consumer (so a breaking change, not just a feature flip) — and lime-elements is general-purpose, where some consumer could be relying on `Referer` reaching the image host (hotlink-protected CDNs, server-side correlation).

Leaning toward keeping it opt-in here and letting consumers like lime-crm-components set `'no-referrer'` explicitly on the `Image` objects they construct (the favicon and avatar use cases that motivated this). Happy to revisit as a separate breaking-change PR if we want to make the opinionated call design-system-wide.

Kiarokh:
> 🤖 AI-generated review from 6 parallel agents (re-review, updated after the fixes pushed since). Treat as input, not a verdict — agents can be wrong or miss context.

This updates the earlier 6-agent re-review to reflect the changes pushed since. The branch is now 8 commits at the current tip.

All six dimensions are SAFE/GOOD with no `[High]`. Both `[Medium]` findings from the prior review are resolved; what remains is optional `[Low]` nits or a tracked follow-up.

### Resolved since the previous review
- 🟡 → ✅ **Unused `class` prop on `ImageTemplate`** (Architecture) — removed; the helper now exposes only `image` + `style` + `onError`.
- 🟡 → ✅ **file-viewer's unconditional `no-referrer`** (Backward Compatibility) — kept as a deliberate, privacy-positive default and documented in the PR body (including why Lime CRM's SharePoint paths are unaffected). An escape-hatch prop can be added later if a consumer needs it.
- 🔵 → ✅ **`style` prop type** widened to `{ [key: string]: string | undefined }` to match Stencil's `` type.
- 🔵 → ✅ **Test coverage** — added a `limel-profile-picture` e2e (no-referrer + object-fit style var + image-error state) that exercises the migration's `style`/`onError` forwarding, which previously had no direct test.

---

## Dimensions

**What works well:** the `Image` change is purely additive and matches the api report; `ImageTemplate` is confirmed `@internal` and unexported; the profile-picture migration is output-equivalent, and file-viewer's array→`<Fragment>` switch is DOM-equivalent.

**Issues:** None.

<details>

- 🔵 file-viewer's `renderImage` now returns a `<Fragment>` while sibling `renderPdf`/`renderText`/etc. still return arrays — cosmetic inconsistency, output unchanged.

</details>

### 2. Code Quality — GOOD ✅

**What works well:** the typed `referrerPolicy` (no cast) is correct for 4.43.5; `onError` matches Stencil's `DOMAttributes.onError`; the migration is leftover-free; the new tests cover the headline behavior on two surfaces.

**Issues:** None.

<details>

- 🔵 profile-picture relies on the helper's `loading ?? 'lazy'` default (it dropped its explicit `loading="lazy"`); now guarded by the new profile-picture e2e plus the list-item default-loading test.

</details>

### 3. Architecture — GOOD ✅

**What works well:** the abstraction is now complete across all five consumers, with a principled split — image *data* on `Image`, host *presentation* (`style`/`onError`) on the internal `ImageTemplateProps`; `src/util/` placement is justified for a shared helper.

**Issues:** None.

<details>

- 🔵 the `style` passthrough could be documented as "CSS custom properties only" to keep it a variable-injection seam rather than a generic style hatch.
- 🔵 extensibility is "add-a-prop"; a typed `Pick<JSXBase.ImgHTMLAttributes>` passthrough would scale better only if the forwarded-attribute list grows — future-only.

</details>

### 4. Security — SAFE ✅

**What works well:** the privacy mechanism is verified end-to-end by e2e (present/absent cases at runtime); file-viewer's leak is closed with `sanitizeUrl` still running before the `Image` is built; the only `style` ever passed is a `'cover'|'contain'` literal.

**Issues:** None.

<details>

- 🔵 the `style` passthrough is a *theoretical* CSS-exfiltration seam; not real today (internal helper, literal-only callers) — a one-line "trusted values only" comment would document the boundary.
- 🔵 `limel-card` / `limel-chip` / `limel-list-item` render `image.src` without URL sanitization (pre-existing; not a script-execution vector for ``, but inconsistent with `limel-file-viewer`). Tracked for a separate PR in #4143.

</details>

### 5. Observability — GOOD ✅

**What works well:** profile-picture's error observability is preserved through the `onError` passthrough (drives `has-image-error` / popover); no `console.*` added; `src`/page-URL never logged; the helper stays out of public types.

**Issues:** None.

<details>

- 🔵 image-load failures stay silent on card/chip/list-item/file-viewer (no `onError` passed) — acceptable; the hook now exists for consumers (e.g. #4122) and those surfaces have no error UI.

</details>

### 6. Performance — SAFE ✅

**What works well:** performance-neutral vs `main` — the functional component is inlined (no wrapper element, no re-mount that would re-fetch images), the N-items hot path passes the stable `this.image` prop, and lazy loading is preserved everywhere.

**Issues:** None.

<details>

- 🔵 profile-picture/file-viewer build a transient `image` literal per render — immaterial (each renders a single image, not on an N-items loop).

</details>

---

## Top Recommendations

None outstanding. The two `[Medium]` findings are resolved (see *Resolved since the previous review*), and the remaining `[Low]` items are optional or tracked in #4143.

Kiarokh (approved):

Issue #590, opened 2020-02-10
limel-dialog doesn't respect --dialog-height


joelricci:
## Current behavior

Dialog adapts height to content. 

## Expected behavior

Height of the dialog should respect `--dialog-height`

## Environment

- **lime-elements** version: 24.6.2
- Framework used: Stencil 1.8.5
- Logs: n/a

adrianschmidt:
@magnusfagerlund Ping for prioritisation 🙂

justFredster:
@magnusfagerlund this is the reason behind the behavior that you reported as weird in ACD. So, it is super important then right? ;)

magnusfagerlund:
Not critical, but would of course be very helpful. Someone in the add-ons team that could look at it during cooldown?

PR #4066, opened 2026-05-07
fix(email-viewer, markdown): handle unreadable emails without manipulating their content

### LINKED ISSUE #4065: Emails with hardcoded text colors can become unreadable in Lime CRM
## The problem

When Lime CRM renders an email, the **text color** is whatever the sender chose when they wrote it — and the **background** comes from Lime CRM's own theme. Two pieces, two different sources. When they don't match, the result can be unreadable.

Image Image

Why this happens? For example, a sender wrote their email in a dark-mode email client. Or for some reason their email had a white text, and a dark background. You're using Lime CRM in light mode. → White text on a light background. Invisible. Or the opposite of above.

Handling this becomes even more complicated when the CRM user is "forcing" a dark or light mode (per individual user preference), but the email's content is carrying a `<meta name="color-scheme">` tag, or `prefers-color-scheme` media queries inside the email's CSS. These would conflict with the forced scheme, while the email is inheriting from the browser or OS.

This affects two places where email content shows up:

1. **The activities feed** — short snippets of incoming emails are mixed with internal comments via `<limel-markdown>`.
2. **The email viewer** — the full original `.eml` file is shown via `<limel-email-viewer>`, inside the file viewer.

Both need to be addressed, but the right fix is different in each.

---

## A bit of history

Back in March 2024 we got a feature request — [`Lundalogik/limepkg-email#590`](https://github.com/Lundalogik/limepkg-email/issues/590) — saying senders should be able to use color in their emails and have it show up in the feed. The classic example: someone writes "see my responses in red" and replies inside a previous email in red.

We addressed that by allowing inline `color` and `background-color` to pass through `<limel-markdown>`'s sanitizer. It worked. Senders' colors started showing up.

But it created the readability problem above. And — more importantly — it pushed an email-shaped responsibility onto a generic component. `<limel-markdown>` is used all over Lime CRM and Lime Elements, not just for email snippets. Asking a generic Markdown/HTML renderer to also "remember sender colors so emails look right" is the kind of responsibility creep that's hard to spot up front but expensive over time.

---

### Part 1 — `<limel-markdown>`: stop preserving inline colors

Remove `color` and `background-color` from the inline-style allowlist. The component goes back to rendering structure and inheriting its text color from the surrounding theme. That makes feed snippets readable in any theme — no more invisible text, and it makes the component's job simpler again.

**But what about the "see my responses in red" use case from #590?**

That use case still matters, just not at this layer. Now that the new `<limel-email-viewer>` exists, the original email is always one click away. I propose handling the rare case where a sender uses color for emphasis as a **promoted action on the feed item itself** ("View original email"), inside `limeb-feed` (the foundation of the activity feed in `lime-crm-building-blocks`). 

Image
💡 We can also display the promoted action for a better discoverability 👇 
Image

That's where edge-case fidelity belongs — it's a product-level affordance, not a rendering concern. Most feed items don't need to reproduce sender colors, and the few that do are best handled by opening the original.

### Part 2 — `<limel-email-viewer>`: theme-following surface, with a one-click escape

The email viewer sets a sensible **default background** based on the app theme:

- App in **light mode** → email body has a light surface (off-white, with a gentle bluish-grey tone).
- App in **dark mode** → email body has a dark surface (off-black, with a gentle bluish tone).

>[!note]
>The chosen tones aren't pure white or pure black — they have a faint hue. That means even when an email's hardcoded text color is exactly `#fff` or `#000`, there's still a sliver of contrast. There's never a complete white-on-white or black-on-black wash-out.

For the cases that still will happen: (e.g. a dark-mode-composed email viewed in light mode), there's a small button labeled **"Hard to read?"** sitting at the top of the email body. One click flips the email's background to the opposite tone. **The email content itself is not modified** — only the surface behind it changes.

A tooltip on the button explains the why in plain language, so users encountering the button for the first time understand what it does.

https://github.com/user-attachments/assets/580c7c82-61e9-4a78-86fe-3e5965b15166

That gives the user a clean escape for any combination of (app theme × email text color), with one click, and without us touching the sender's bytes.

💡 In future, we can extend this functionality and make it a "rememberable" thing per email. And save it in the database.
---

### Strip inline colors from the email viewer too

The same trick we used for `<limel-markdown>`: just delete inline `color` and `background-color` declarations from email HTML. That works for readability — but it destroys the very thing the email viewer is for. The whole point of the email viewer is **fidelity** — letting the user see what the sender actually sent. Stripping declarations would break that promise.

### Auto-detect the email's "intended" color scheme

Some emails carry hints about their intended display mode (a `<meta name="color-scheme">` tag, or `prefers-color-scheme` media queries inside the email's CSS). We could parse those and pick a surface accordingly. But:

- Most emails include no hint at all.
- When a hint *is* present, it usually means "I support both light and dark" rather than "I am dark." Acting on that as if it were certainty produces wrong guesses.
- Even if we did detect well, we'd still need a manual override for the cases we got wrong — so we'd be paying complexity *and* still need the manual button. As explain earlier, things will get complicated when user forces dar/light mode, but browser or OS are saying something else, and we have custom CSS vars for colors which are automatically switching according to user's scheme preference in the app.

### Walk the rendered DOM and adjust low-contrast colors per element

This was actually attempted earlier in [PR #4060](https://github.com/Lundalogik/lime-elements/pull/4060). The idea: after the email renders, walk every text element, measure its color's contrast against the resolved background, and strip declarations below a 3:1 threshold. Save the original color on a `data-` attribute so it can come back when the theme flips.

It works. But it brings a lot with it:

- Walking the entire rendered DOM on every email render.
- A WCAG contrast calculation, plus a parser for every CSS color format senders might use.
- A `MutationObserver` and a media-query listener to react to theme changes at runtime.
- Per-element bookkeeping so stripped colors can be restored.
- Long term maintenance and edge cases that we don't see

When that approach was reproposed as a `<limebb-markdown>` wrapper inside `lime-crm-building-blocks` (so the generic `<limel-markdown>` could stay simple), the *cost* didn't change — it just relocated. The complexity is the **behaviour**, not the address. And it still manipulates the email's content, even if the manipulation is reversible.

For a problem the user encounters occasionally, this is an expensive answer. The "Hard to read?" button does the same job for a fraction of the code, and it never touches the sender's content.

---

## Summary

- `<limel-markdown>` goes back to being a clean, theme-neutral renderer. Inline colors no longer survive sanitization, so feed snippets are always readable.
- The "preserve sender emphasis colors in the feed" need from `limepkg-email#590` is best handled at the product level — a "See the original email" promoted action on the feed item — not at the rendering layer.
- `<limel-email-viewer>` adapts its background to the app theme, and offers a small **"Hard to read?"** button for the rare cases where the email's hardcoded colors clash with the surface. The email content is never modified.
- We deliberately chose **not** to detect, calculate, or auto-adapt anything. The cost outweighs the benefit, and a single-click manual override solves the problem with almost no machinery.

Kiarokh:
fix: #4065

## Summary

Two atomic fixes that together solve the readability problem described in #4065 without ever modifying email content:

1. **`<limel-markdown>`** — drops `color` and `background-color` from the inline-style allowlist. The component goes back to being a clean, theme-neutral renderer; feed snippets stay readable in any theme. The "preserve sender emphasis colors in the feed" use case from `limepkg-email#590` will be solved at the product level — a "See the original email" promoted action on the feed item (in `limeb-feed`).

2. **`<limel-email-viewer>`** — sets a theme-following default background on the email body (using `--contrast-300` / `--contrast-1500` tokens that auto-invert with the app theme), and adds a small **"Hard to read?"** toggle that flips the surface to the opposite tone for emails whose hardcoded colors collide with the default. A `limel-tooltip` on the button explains the why. **The email content itself is never modified** — only the surface behind it changes.

See #4065 for the full reasoning, including the approaches we considered and rejected (auto-detection of color-scheme hints, contrast-aware DOM walking from #4060, etc.).

## Test plan

- [ ] In Lime CRM **light mode**, open the activity feed and verify email snippets render in the theme's text color (no longer invisible-on-light).
- [ ] In Lime CRM **dark mode**, open the activity feed and verify the same — text inherits the theme color.
- [ ] Open an `.eml` file via `<limel-file-viewer>` while the app is in **light mode**. Verify the email body has a light off-white surface.
- [ ] Repeat in **dark mode**. Verify the email body has a dark off-black surface.
- [ ] Open an email composed in dark mode (white text) while the app is in **light mode**. Verify the white text is hard to see, then click **"Hard to read?"** — the surface should flip to dark and the white text becomes readable.
- [ ] Open an email composed in light mode (black text) while the app is in **dark mode**. Verify the same in reverse.
- [ ] Hover/focus the toggle button and verify the tooltip appears with the longer helper text.
- [ ] Switch between two emails and verify the toggle resets to default for the new email.
- [ ] Verify the toggle button is sticky at the top of the body when scrolling a long email.
- [ ] Verify the toggle button doesn't block clicks on email content underneath it (`pointer-events: none` on the wrapper).

## Browsers tested

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

## Summary by CodeRabbit

* **New Features**
  * Added email surface background toggle to switch between light and dark backgrounds for improved readability.

* **Style**
  * Removed inline color and background-color styling support from markdown content for consistency.

* **Localization**
  * Added translations for the new email surface toggle across 8 languages: Danish, German, English, Finnish, French, Dutch, Norwegian, and Swedish.

### DIFF
--- a/src/components/email-viewer/email-viewer.scss
+++ b/src/components/email-viewer/email-viewer.scss
@@ -1,6 +1,8 @@
 @use '../../style/mixins';
 
 :host(limel-email-viewer) {
+    isolation: isolate;
+
     display: block;
     width: 100%;
     height: 100%;
@@ -72,6 +74,7 @@
 
         &.date {
             position: absolute;
+            z-index: 1;
             right: 0.25rem;
             bottom: 0;
             transform: translateY(50%);
@@ -156,7 +159,6 @@ section {
     flex-direction: column;
     border-top: 1px dashed rgba(var(--contrast-700));
     min-height: 2rem;
-    overflow-y: auto;
 }
 
 limel-collapsible-section {
@@ -198,7 +200,16 @@ limel-collapsible-section {
 .body {
     flex-grow: 1;
     max-width: 100%;
-    padding: 0.75rem;
+    padding: 1rem;
+    padding-bottom: 3rem;
+    overflow-y: auto;
+
+    background-color: rgb(var(--contrast-300));
+    color-scheme: light;
+    &.toggled-surface-background {
+        background-color: rgb(var(--contrast-1500));
+        color-scheme: dark;
+    }
 
     &.plain-text {
         white-space: pre-wrap;
@@ -209,5 +220,52 @@ limel-collapsible-section {
 
     img {
         max-width: 100% !important;
+    }
+}
+
+.toggle-dark-light {
+    display: flex;
+    align-items: center;
+    justify-content: flex-end;
+
+    position: sticky;
+    top: 0;
+    z-index: 1;
+
+    padding: 0.25rem 0.5rem;
+    $height: 2.5rem;
+    height: $height;
+    margin-bottom: -$height;
+
+    pointer-events: none;
+
+    button {
+        pointer-events: all;
+        @include mixins.reset-button-user-agent-styles;
+        @include mixins.is-elevated-clickable();
+        @include mixins.visualize-keyboard-focus();
+
+        display: flex;
+        align-items: center;
+        justify-content: center;
+        gap: 0.25rem;
+
+        padding: 0.125rem 0.25rem;
+        height: fit-content;
+
+        font-size: 0.75rem;
+        border-radius: 0.375rem;
+        border: 1px solid rgba(var(--contrast-700));
+
+        opacity: 0.6;
+
+        &:hover,
+        &:focus-visible {
+            opacity: 1;
+        }
+
+        limel-icon {
+            width: 1.125rem;
+        }
     }
 }
--- a/src/components/email-viewer/email-viewer.tsx
+++ b/src/components/email-viewer/email-viewer.tsx
@@ -14,6 +14,7 @@ import { Email, EmailAttachment, EmailHeaderType } from './email-viewer.types';
 import { applyRemoteImagesPolicy, containsRemoteImages } from './remote-images';
 import { splitEmailAddressList } from './split-email-address-list';
 import { formatBytes } from '../../util/format-bytes';
+import { createRandomString } from '../../util/random-string';
 
 /**
  * This is a private component, used to render `.eml` files inside
@@ -70,6 +71,11 @@ export class EmailViewer {
     @State()
     private allowRemoteImagesState = false;
 
+    @State()
+    private surfaceBackgroundState = false;
+
+    private toggleButtonId = createRandomString();
+
     /**
      * Emitted when the user requests remote images to be loaded.
      *
@@ -79,7 +85,11 @@ export class EmailViewer {
     public allowRemoteImagesChange: EventEmitter<boolean>;
 
     @Watch('email')
-    protected resetAllowRemoteImages(newEmail?: Email, oldEmail?: Email) {
+    protected handleEmailChange(newEmail?: Email, oldEmail?: Email) {
+        // Surface preference is per-email-view: a new message always opens
+        // with the default light background.
+        this.surfaceBackgroundState = false;
+
         if (!newEmail) {
             this.allowRemoteImagesState = false;
             return;
@@ -98,6 +108,7 @@ export class EmailViewer {
                     {this.renderRemoteImageBanner()}
                     <section>
                         {this.renderAttachments()}
+                        {this.renderSurfaceToggle()}
                         {this.renderBody()}
                     </section>
                 </div>
@@ -146,7 +157,13 @@ export class EmailViewer {
             this.getAllowRemoteImages()
         );
 
-        return <div class="body" innerHTML={innerHtml} part="email-body" />;
+        return (
+            <div
+                class={this.getBodyClassNames()}
+                innerHTML={innerHtml}
+                part="email-body"
+            />
+        );
     }
 
     private renderBodyText() {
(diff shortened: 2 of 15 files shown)


FredrikWallstrom:
Hey Kia, jumping in on the design direction here. The email-viewer half of this PR (theme-following surface + "Hard to read?" toggle) feels right to me, and I'd accept it independently of where the markdown discussion lands. What I want to question is part 1 — dropping `color` and `background-color` from `<limel-markdown>`'s sanitizer.

The framing of part 1 as "removing email-shaped responsibility from a generic component" is what I want to question. Inline `color` in HTML isn't an email-specific feature — it's a basic markdown/HTML feature. GitHub renders it. CommonMark with raw HTML enabled renders it. Any markdown surface that allows `<span style="...">` will respect colors. `<limel-markdown>` is a *public* component in `@limetech/lime-elements`, used outside Lime CRM by partners and other internal products. Removing colors from the inline-style allowlist isn't undoing a special case — it's removing a generic feature that any consumer might be relying on. Status messages with colored severity, formatted docs, inline emphasis, anything that uses `<span style="color: ...">` in their markdown will silently change when we ship this. That's a public API change that goes beyond fixing the email-readability issue, and we have no good visibility into who outside this repo depends on it.

The other part of the framing I'd question is treating color use as a deliberate, flagged thing — "see my responses in red." That's one shape of color use, but plenty of it isn't deliberate at all. People highlight a name in blue, color a date red, paste an action item in green, use a colored heading from a template — without thinking of it as "color emphasis." Senders don't consciously decide "I'm using color here," they just write that way, and recipients read it that way. That's what #590 was actually about, even if the phrasing made it sound niche — it's not really one feature request, it's a pattern in how people compose emails. When we strip color universally, we lose all of that. For most messages the loss is invisible. For some it's load-bearing in a way that only becomes obvious after it's gone. The "View original email" promoted action is a fine fallback when someone consciously *wants* to see the rendered original, but it doesn't help in the case where someone is just glancing through their feed and would naturally read the colored phrase inline if it were there.

Separate point — the writeup undersells how slimmed-down the contrast approach got. The cost it lists (DOM walk, contrast calc, `MutationObserver`, media-query listener, per-element bookkeeping for restoration) isn't what's in the merged code. Those got pruned earlier in the discussion. The live version is roughly: post-render, `querySelectorAll('[style*="color"]')`, compute contrast, strip declarations under 3:1. No observers, no listeners, no restoration data — theme change requires a page reload, and that's fine. That's a meaningfully smaller picture than the writeup paints, and the trade-off looks different when weighed against the actual size.

What I find compelling about the contrast approach is that it intervenes only in cases where leaving the content alone would already produce something unreadable. For the organic color use we've both seen — colored names, dates, highlights, headings — those mostly meet 3:1 against a normal CRM surface, so they pass through untouched. The walk kicks in only for the genuinely-broken combinations: a sender composing in dark mode whose `color: white` lands on our light feed, or Outlook's auto-injected near-black landing on our dark feed. Those are cases where doing nothing is *also* destroying the user's reading — silently, with no recourse short of clicking through to the viewer.

So I'd reframe the part 1 trade-off: it's not "preserve sender colors at the cost of architectural purity vs. drop them and keep the renderer clean." It's "drop a generic markdown feature from a public component (breaking external consumers as collateral) and lose the organic-color reading that #590 was actually about, vs. keep the public API as-is and add ~50 lines of CRM-specific logic in `lime-crm-building-blocks` to handle the cases where ignoring contrast would already be unreadable."

Kiarokh:
Thanks @FredrikWallstrom 

Regarding dropping or keeping colors in markdown. You're right that we have consumers that might be using it as is. I'm not sure how serious these use cases are, but perhaps that's a bad idea to suddenly remove the functionality. So I'm not insisting on doing so. But if I was gonna redesign this component again, I'd probably decide to exclude colors via sanitizer again 😄 It feels, for a markdown renderer component which is going to appear on Web Application UIs (as a natural part of the UI itself), that's an overkill. Imagine if Slack, Teams, Github, etc… all rendered text with color (with **_any_** user-chosen color) everywhere. That would open the Pandora's box, specially with dark mode / light mode support of most app UIs these days (as it has already done in our case). I think it would have been better to keep that door shut from the beginning. 

But if we drop this, and instead go with the contrast solution that you proposed, then the other commit in the current PR will also become meaningless. We should rather use your solution in both components in that case. That'll give a much better out of the box UX at the end. 

I'm not sure how much does it cost when it comes to performance. Probably nothing noticable if it is only one instance doing it. But should we consider to add it under a `prop`? In the web client, we could have lots of `limel-markdown` in the DOM simultaneously. For example, in the Table or Kanban view, each row or item has markdown inbuilt. They are also used in other parts of the UI. Imagine to have the feed in the sidepanel opend in the table view. So on a page, we could potentially have hundreds of them. I strongly feel this contrast checks are unnecessary in majority of cases, and the only place they make sense is in the feed and almost-ONLY for emails scenarios. So if we go that way, I'd like to have that under a prop for `limel-markdown`

PR #4067, opened 2026-05-08
fix(form): namespace schema $id by content to avoid stale validator

FredrikWallstrom:
## Summary

When a consumer renders `limel-form` with a schema that reuses the same `$id` while its content changes between renders (e.g. a server schema whose `properties` get filtered based on user input), the form's `validate` event would fire with errors that don't match the schema's actual content — most visibly, fields would be reported as "additional properties" even though they're declared in `properties`.

The root cause is that the validator caches compiled validators by `$id`. The first variant compiled under a given `$id` is reused for every later variant with the same `$id`, so swapping the schema's content silently has no effect on validation.

This fix wraps the schema in a copy whose `$id` becomes `<original-$id>-<fingerprint>`, where the fingerprint is a short hash of the stringified content (sans `$id`), before handing it to RJSF and the validator. Distinct content produces a distinct cache entry; identical content still hits the cache. The fingerprint — rather than the raw content — keeps `$id` short and URI-safe, since the validator parses it as a URI and chokes on JSON-shaped characters. The consumer's original schema is left untouched.

Fixes Lundalogik/crm-client#971.

## Test plan

- [x] New regression test in `form.e2e.tsx` covers the scenario (same `$id`, schema gains a required property, value updates to satisfy it).
- [x] Existing form e2e suite still passes (42/42).
- [x] Existing form spec suite still passes (56/56).
- [x] Manual verification: open the bulk-edit dialog, pick a text field, type — the Update button enables. Lime Admin loads schemas without URI-malformed errors and at expected speed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where form validation would not properly update when schema requirements changed dynamically while maintaining the same schema identifier. Forms now correctly re-validate against updated requirements.

* **Tests**
  * Added end-to-end validation test covering dynamic schema requirement changes at runtime.

### DIFF
--- a/src/components/form/form.tsx
+++ b/src/components/form/form.tsx
@@ -170,7 +170,7 @@ export class Form {
             React.createElement(
                 JSONSchemaForm,
                 {
-                    schema: this.schema,
+                    schema: this.prepareSchema(this.schema),
                     formData: this.value,
                     onChange: this.handleChange,
                     widgets: widgets,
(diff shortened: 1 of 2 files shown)


src/components/form/form.tsx
```diff
@@ -209,11 +209,42 @@ export class Form {
...
+        const { $id, ...rest } = schema;
+        const json = JSON.stringify(rest);
+
+        let fingerprint = 5381;
+        for (let i = 0; i < json.length; i++) {
+            fingerprint =
+                ((fingerprint << 5) + fingerprint + json.codePointAt(i)) &
+                0xff_ff_ff_ff;
+        }
+        const contentId = (fingerprint >>> 0).toString(36);
```
FredrikWallstrom, line 240: I tried to keep it simple and just use JSON.stringify (or URLEncode) as `$id`, but unfortunately the `$id` became too big, which caused Lime Admin to crash on big schemas.

jgroth: So the ID would be a JSON encoded string of the schema itself? 😄

jgroth (approved):

src/components/form/form.tsx
FredrikWallstrom: yeah, that actually worked (but not on big schemas though).

PR #4069, opened 2026-05-08
feat: adapt low-contrast inline colors to surface

FredrikWallstrom:
## Summary by CodeRabbit

* **New Features**
  * Markdown gains an optional adaptColorContrast mode to auto-adjust inline text colors for better WCAG contrast.
  * Email viewer now applies color-contrast adaptation during render for improved readability.

* **Documentation**
  * Added an example demonstrating markdown color-contrast adaptation.

* **Tests**
  * Added comprehensive unit and end-to-end tests covering color parsing, contrast calculations, and adaptive behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Lundalogik/lime-elements/pull/4069)

Fixes Lundalogik/crm-client#621
Fixes Lundalogik/crm-client#966

When externally-authored HTML lands inside CRM (imported email bodies, in particular), the sender's inline `color:` declarations can collide with the host theme — black-on-dark, white-on-white, etc. — leaving content unreadable.

This adds a small `adaptColorContrast` utility that walks a rendered subtree, computes WCAG contrast for each element with an inline `color`, and removes declarations below a 3:1 ratio so the surface's themed text color inherits through. Brand colors that already meet contrast pass through untouched.

Three atomic commits:

1. **`feat(util): add adaptColorContrast utility`** — pure DOM walk + WCAG luminance / contrast / sRGB compositing, plus spec and e2e tests. Lives in `src/util/` since it's not specific to any one component.
2. **`feat(markdown): add opt-in adaptColorContrast prop`** — new `adaptColorContrast` boolean prop on `<limel-markdown>`, default `false`. Off by default keeps the component a neutral renderer; consumers that mix authored content with themed surfaces (the activity feed) opt in.
3. **`feat(email-viewer): adapt color contrast in rendered body`** — runs unconditionally after each render in `<limel-email-viewer>`, since it always renders externally-authored content.

Replaces the closed PR #4060 (which had the same logic but ran unconditionally on every `<limel-markdown>` render) and supersedes the limbb wrapper approach in `Lundalogik/lime-crm-building-blocks#1423` (closed). Direction agreed offline with @Kiarokh — the algorithm lives in lime-elements, opt-in for `<limel-markdown>`, default-on for `<limel-email-viewer>`.

## Review:

## Test plan:
- [ ] Render an `.eml` whose body has inline `color: black` text in CRM dark mode via the file viewer. Verify the text is readable (inherits the surface's themed color) rather than rendering black-on-dark.
- [ ] Render an `.eml` whose body has inline `color: white` text in CRM light mode. Verify the text falls back to a readable color rather than being invisible.
- [ ] Render an `.eml` whose body uses brand colors (red, blue, green) on a normal CRM surface in both light and dark mode. Brand colors should still be visible.
- [ ] In a surface that uses `<limel-markdown>` *with* `adapt-color-contrast` set, confirm the same behaviour for inline-colored markdown content.
- [ ] In any other surface that uses `<limel-markdown>` *without* `adapt-color-contrast`, confirm rendering is unchanged from main.

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/email-viewer/email-viewer.tsx
+++ b/src/components/email-viewer/email-viewer.tsx
@@ -14,6 +14,7 @@ import { Email, EmailAttachment, EmailHeaderType } from './email-viewer.types';
 import { applyRemoteImagesPolicy, containsRemoteImages } from './remote-images';
 import { splitEmailAddressList } from './split-email-address-list';
 import { formatBytes } from '../../util/format-bytes';
+import { adaptColorContrast } from '../../util/adapt-color-contrast';
 
 /**
  * This is a private component, used to render `.eml` files inside
@@ -90,6 +91,14 @@ export class EmailViewer {
         }
     }
 
+    private bodyElement?: HTMLDivElement;
+
+    public componentDidRender() {
+        if (this.bodyElement?.isConnected) {
+            adaptColorContrast(this.bodyElement);
+        }
+    }
+
     public render() {
         return (
             <Host>
@@ -146,7 +155,14 @@ export class EmailViewer {
             this.getAllowRemoteImages()
         );
 
-        return <div class="body" innerHTML={innerHtml} part="email-body" />;
+        return (
+            <div
+                class="body"
+                innerHTML={innerHtml}
+                part="email-body"
+                ref={(el) => (this.bodyElement = el as HTMLDivElement)}
+            />
+        );
     }
 
     private renderBodyText() {
--- a/src/components/markdown/markdown.tsx
+++ b/src/components/markdown/markdown.tsx
@@ -6,6 +6,7 @@ import { ImageIntersectionObserver } from './image-intersection-observer';
 import { hydrateCustomElements } from './hydrate-custom-elements';
 import { morphChildren } from './morph-dom';
 import { DEFAULT_MARKDOWN_WHITELIST } from './default-whitelist';
+import { adaptColorContrast } from '../../util/adapt-color-contrast';
 
 /**
  * The Markdown component receives markdown syntax
@@ -36,6 +37,7 @@ import { DEFAULT_MARKDOWN_WHITELIST } from './default-whitelist';
  * @exampleComponent limel-example-markdown-custom-component
  * @exampleComponent limel-example-markdown-custom-component-with-json-props
  * @exampleComponent limel-example-markdown-remove-empty-paragraphs
+ * @exampleComponent limel-example-markdown-adapt-color-contrast
  * @exampleComponent limel-example-markdown-composite
  */
 @Component({
@@ -91,6 +93,22 @@ export class Markdown {
     @Prop({ reflect: true })
     public removeEmptyParagraphs = true;
 
+    /**
+     * Adapt rendered inline `color:` declarations to the surrounding
+     * surface. After each markdown re-render the component walks the
+     * rendered DOM and removes any inline `color` whose contrast against
+     * the resolved background falls below WCAG 3:1, letting the surface's
+     * themed text color inherit through. Brand colors that already meet
+     * contrast are left alone.
+     *
+     * Default `false` so the component remains a neutral renderer; turn
+     * this on for surfaces that render externally-authored content
+     * (e.g. imported email bodies) where the host application's theme
+     * drives the surrounding text color.
+     */
+    @Prop({ reflect: true })
+    public adaptColorContrast = false;
+
     @Watch('value')
     public async textChanged() {
         try {
@@ -126,6 +144,10 @@ export class Markdown {
             // rehype-sanitize can't inspect values inside JSON strings.
             hydrateCustomElements(this.rootElement, combinedWhitelist);
 
+            if (this.adaptColorContrast) {
+                adaptColorContrast(this.rootElement);
+            }
+
             this.setupImageIntersectionObserver();
         } catch (error) {
             console.error(error);
@@ -139,6 +161,11 @@ export class Markdown {
 
     @Watch('removeEmptyParagraphs')
     public handleRemoveEmptyParagraphsChange() {
+        return this.textChanged();
+    }
+
+    @Watch('adaptColorContrast')
+    public handleAdaptColorContrastChange() {
         return this.textChanged();
     }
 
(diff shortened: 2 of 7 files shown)


src/components/markdown/markdown.tsx
```diff
@@ -91,6 +92,22 @@ export class Markdown {
...
+     * color inherit through. Brand colors that already meet contrast are
+     * left alone.
+     *
+     * Default `false` so the component remains a neutral renderer; turn
+     * this on for surfaces that render externally-authored content
+     * (e.g. imported email bodies) where the host application's theme
+     * drives the surrounding text color.
+     */
+    @Prop({ reflect: true })
+    public adaptColorContrast = false;
```
adrianschmidt, line 110: Doing cleanup after every render sounds rather inefficient to me. Why can't it be done as pre-processing by the component providing the bad input in the first place? 🙂

FredrikWallstrom: Believe me, me and @Kiarokh has been discussing this back and forth 😄 

See https://github.com/Lundalogik/lime-crm-building-blocks/pull/1423, and https://github.com/Lundalogik/lime-elements/pull/4066

Pre-processing the "real field value" doesn't feel right too me since we will mutate users content which will then be lost on round-trips. I rather see a visual solution.

We would love to hear your thoughts how to solve this.

adrianschmidt: OK. I'm going to dig into your arguments, and see if there are really no better solutions. Please give me over the weekend before you merge this. This component unfortunately has a history of being the go to place for people to try to cram patches to fix their broken email HTML, and I _really_ want that to stop. But if this is truly the best solution in this particular case, I _will_ approve it.

src/components/email-viewer/email-viewer.tsx
```diff
@@ -90,6 +91,14 @@ export class EmailViewer {
         }
     }
 
+    private bodyElement?: HTMLDivElement;
+
+    public componentDidRender() {
+        if (this.bodyElement) {
+            adaptColorContrast(this.bodyElement);
+        }
+    }
```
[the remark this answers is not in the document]
FredrikWallstrom: Added the `?.isConnected` guard so we short-circuit cleanly when the body div has been unmounted.

src/components/markdown/markdown.tsx
```diff
@@ -91,6 +93,22 @@ export class Markdown {
...
```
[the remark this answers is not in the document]
FredrikWallstrom: Real bug. Added a `@Watch('adaptColorContrast')` that calls `textChanged()`, mirroring the pattern used for `whitelist` and `removeEmptyParagraphs`. As you noted, `morphChildren` re-injects the original markup, so a `true → false` toggle restores the previously-stripped colors for free.

src/components/markdown/markdown.tsx
```diff
@@ -91,6 +93,22 @@ export class Markdown {
...
+     * removes any inline `color` whose contrast against the resolved
+     * background falls below WCAG 3:1, letting the surface's themed text
+     * color inherit through. Brand colors that already meet contrast are
+     * left alone.
+     *
+     * Default `false` so the component remains a neutral renderer; turn
+     * this on for surfaces that render externally-authored content
+     * (e.g. imported email bodies) where the host application's theme
+     * drives the surrounding text color.
+     */
```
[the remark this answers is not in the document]
FredrikWallstrom: Tightened the docstring to "After each markdown re-render…" so the trigger semantics are explicit.

src/util/adapt-color-contrast.ts
```diff
@@ -0,0 +1,211 @@
+const MIN_CONTRAST_RATIO = 3;
+const STRIPPED_COLOR_ATTR = 'data-limel-stripped-color';
```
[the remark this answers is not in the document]
FredrikWallstrom: Removed `STRIPPED_COLOR_ATTR` and the `setAttribute` write. The restore-on-theme-change logic that originally justified it was dropped earlier in the discussion, so it's just dead state at this point.

src/components/markdown/markdown.tsx
adrianschmidt: OK, I've looked at you earlier discussion. It seems like this is a fairly reasonable trade-off.

Claude helped me summarise your conversations, and traced back the first "email band-aid" to be my commit 43e89bfc4 (`feat(markdown): support certain whitelisted inline css styles`, fix for Lundalogik/limepkg-email#590). It told me _the fact that you put the email-shaped feature in there in the first place is worth naming. Otherwise the precedent argument lands as "stop doing what I started."_ 😄

But yes, "stop doing what I started" is _exactly_ what I'm saying 😆

So, consider this approved on a conceptual level. I'm not setting an approved-stamp, since I have neither reviewed nor tested the code.

Thank you for holding off on merging this until I had a chance to look into it a bit 🙏

src/components/email-viewer/email-viewer.tsx
```diff
@@ -90,6 +91,14 @@ export class EmailViewer {
         }
     }
 
+    private bodyElement?: HTMLDivElement;
+
+    public componentDidRender() {
+        if (this.bodyElement?.isConnected) {
+            adaptColorContrast(this.bodyElement);
```
devbymadde, line 98: Does this have to run on every re-render, or could we check so that it only runs for example when innerHTML is updated? (Or maybe it's never re-rendered apart from when the body changes? 🤔)

FredrikWallstrom: We could skip to run this on every re-render by memoizing the produced innerHtml string. However, that means holding a full extra copy of the body HTML in memory, not just a pointer to the input.

Which approach do we think is best? Holding a copy of the innerHTML in-memory, or run adaption on each re-render?

(The reason we'd need to memoize is that `applyRemoteImagesPolicy` returns a fresh string on every call when remote images are on, so Stencil keeps re-setting innerHTML between renders, which re-parses the HTML and silently wipes our adaptation.)

devbymadde: Alright, let's keep it as it is :)

devbymadde (approved):

PR #4080, opened 2026-05-20
fix(file-viewer): guard Fullscreen against missing browser API

devbymadde:
## Fix: guard `Fullscreen` against missing browser API

iPhone Safari does not expose `requestFullscreen` (or any vendor-prefixed variant) on regular `HTMLElement`s. The `Fullscreen` constructor unconditionally called `.bind(element)` on the resolved method, which threw a `TypeError` when none of the variants existed. The throw propagated up through `limel-file-viewer`'s constructor, leaving the component with an empty shadow DOM whenever `allowFullscreen` was set.

This was particularly visible for images: `renderPdf` doesn't touch `this.fullscreen`, so PDFs rendered fine, but `renderImage` → `renderButtons` → `renderToggleFullscreenButton` calls `this.fullscreen.isSupported()`, which then crashed `render()` entirely — and the user saw nothing.

### Changes

- `fullscreen.ts`: guard `.bind` behind an `if (this.enter)` check; type `enter` as `(() => void) | undefined`. 
- `fullscreen.ts`: `isSupported()` now returns `!!this.enter` instead of `!!this.requestFullscreen`. The previous implementation checked the class's own arrow-function property, so it always returned `true` — which is why callers' `isSupported()` guards never prevented the crash.
- `fullscreen.spec.ts`: new unit tests covering supported, vendor-prefixed, and unsupported (iPhone Safari) cases.

### Impact

All consumers passing `allowFullscreen` to `limel-file-viewer` will now render images correctly on iPhone Safari. The fullscreen button is simply hidden where the API is unavailable.

## Summary by CodeRabbit

* **Tests**
  * Added tests covering fullscreen support detection and behavior across standard, vendor-prefixed, and unsupported environments.

* **Refactor**
  * Improved fullscreen stability: detection logic refined, vendor-prefixed APIs handled, and operations become safe no-ops when fullscreen isn’t available.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Lundalogik/lime-elements/pull/4080?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [x] Firefox
- [x] Safari

Mobile:
- [ ] Chrome on Android
- [x] iOS

### DIFF
--- a/src/components/file-viewer/fullscreen.spec.ts
+++ b/src/components/file-viewer/fullscreen.spec.ts
@@ -1,0 +1,59 @@
+import { Fullscreen } from './fullscreen';
+
+describe('Fullscreen', () => {
+    describe('when the browser supports the Fullscreen API', () => {
+        let element: any;
+        let requestFullscreenSpy: ReturnType<typeof vi.fn>;
+        let fullscreen: Fullscreen;
+
+        beforeEach(() => {
+            requestFullscreenSpy = vi.fn();
+            element = { requestFullscreen: requestFullscreenSpy };
+            fullscreen = new Fullscreen(element);
+        });
+
+        it('reports as supported', () => {
+            expect(fullscreen.isSupported()).toBe(true);
+        });
+
+        it('calls the native method with the element bound as `this`', () => {
+            fullscreen.requestFullscreen();
+
+            expect(requestFullscreenSpy).toHaveBeenCalledTimes(1);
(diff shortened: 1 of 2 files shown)


Kiarokh (approved):
works like a charm

PR #4081, opened 2026-05-22
feat(picker): support per-item removable flag on selected chips

Befkadu1:
fix https://github.com/Lundalogik/crm-client/issues/1036
## Summary
- Add `PickerItem` type (extends `ListItem`) with optional `removable` flag, and export it from the package entry.
- Honor `removable: false` when rendering picked chips so individual items can be locked from removal.
- Retype public `value`, `allItems`, `change`, and `interact` to use `PickerItem` (backward compatible — `removable` is optional).
- Add `limel-example-picker-non-removable` demonstrating the feature.

## Test plan
- [ ] Run `npm run watch` and open the new "Picker with non-removable values" example.
- [ ] Verify the pre-selected "You" chip has no remove button.
- [ ] Verify Backspace/Delete do not remove "You".
- [ ] Verify "Clear all" removes other chips but leaves "You" in place (and hides itself once only "You" remains).
- [ ] Verify clicking "You" still emits an `interact` event (visible via `<limel-example-value>`).
- [ ] Verify other participants can be added and removed normally.

## Summary by CodeRabbit

* **New Features**
  * Picker supports non-removable (locked) selected items so users can prevent specific chips from being removed.

* **Behavior Changes**
  * Picker selection and search now consistently handle items with removable flags, improving UX for locked items.

* **Documentation**
  * Added an example showing how to configure and use non-removable items in the picker.

[Review Change Stack](https://app.coderabbit.ai/change-stack/Lundalogik/lime-elements/pull/4081?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -681,7 +681,7 @@ export namespace Components {
         "actionPosition": ActionPosition;
         "actions": Array<ListItem<Action>>;
         "actionScrollBehavior": ActionScrollBehavior;
-        "allItems"?: Array<ListItem<PickerValue>>;
+        "allItems"?: PickerItem[];
         "badgeIcons": boolean;
         "delimiter": string;
         "disabled": boolean;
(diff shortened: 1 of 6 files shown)


Kiarokh (approved):
I added some fixups. If you like the changes too, let's merge this

PR #4084, opened 2026-05-25
fix(text-editor): prevent content loss when pasting tables into cells

john-traas:
## Summary

- Adds a `table-paste-plugin` registered before prosemirror-tables' `tableEditing()`. It reads the pasted slice the same way prosemirror-tables does, so anything that is not a cell grid, such as text copied from inside a cell, pastes normally.
- Unless cells were selected explicitly, a pasted grid never overwrites cells. The paste replaces the selection like any other paste, then adds the grid as rows below the caret's row when the column count matches and no cell spans that row boundary. Otherwise it becomes a new table directly below the enclosing table, or lands at the caret when the selection reached out of the table.
- A paste that mixes a table with other blocks is moved intact below the enclosing table.
- `CellSelection` pastes still use prosemirror-tables' cell-grid replace, so selecting cells first remains the explicit way to overwrite them.
- Tables get simple visible borders while editing (scoped to `limel-prosemirror-adapter`; `limel-markdown` rendering is unchanged)

Closes part of [lime-19624](https://product.lime-crm.com/client/explorer/bug/19624) / [Lundalogik/crm-client#1050](https://github.com/Lundalogik/crm-client/issues/1050). The reported bug is "pasting table rows in the Conversations composer deletes other rows" — confirmed in 4 environments. The composer opts into `contentType='html'` legitimately, so the fix lives here, not in `limepkg-email`.

## Test plan

- [x] Plugin spec: `npx vitest run src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.spec.ts --project spec` (19 tests, including the slice shape of text copied from a cell, a node selection inside a cell, and a ragged pasted table)
- [x] Real copy-and-paste round trips on a mounted view: `npx vitest run src/components/text-editor/prosemirror-adapter/editor-tables.e2e.ts --project e2e`
- [x] Full text-editor suites: 288 spec tests and 109 e2e tests pass
- [x] Verified in Chromium on the `text-editor-with-tables` docs example with the OS clipboard: text copied between cells pastes as text; four drag-selected cells paste as new rows; a two-column external table joins as a row below the caret's row; a three-column external table becomes a separate table below; a selection dragged from the paragraph below into a cell pastes without nesting a table in the cell
- [x] `CellSelection` cell-grid replace unchanged (select cells, paste a table, selected cells replaced as before)

## Summary by CodeRabbit

* **New Features**
  * Added support for pasting table rows, cells, and complete tables into the editor.
  * Compatible rows are inserted below the current table row without overwriting existing content.
  * Incompatible tables and block content are placed outside the current table.
  * Preserves surrounding content and updates the selection after handled pastes.

* **Style**
  * Improved table borders, spacing, padding, and cell alignment.

* **Tests**
  * Added comprehensive coverage for table paste scenarios and selection handling.


src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export const tablePastePluginKey = new PluginKey('tablePastePlugin');
+
+// Narrow gate: only intercept when the pasted slice's top-level is a
+// complete <table>. Bare <tr>/<td> slices would not fit at block level
+// (where the post-table insertion happens), so we let the default
+// prosemirror-tables flow handle them.
+const sliceStartsWithTable = (slice: Slice): boolean => {
+    return slice.content.firstChild?.type.spec.tableRole === 'table';
+};
```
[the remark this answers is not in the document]
john-traas: Reordered — `createTablePastePlugin` now sits above the private helper.

⚡ 9450b50b2f5b743f617d30c340d3ec9a76d66d1d

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
```diff
@@ -0,0 +1,80 @@
...
+                    for (let depth = $from.depth; depth > 0; depth--) {
+                        if ($from.node(depth).type.spec.tableRole === 'table') {
+                            tableDepth = depth;
+                            break;
+                        }
+                    }
+
+                    if (tableDepth < 0) {
+                        return false;
+                    }
```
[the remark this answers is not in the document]
john-traas: The inconsistency was real, but gating on `state.selection.empty` would have fallen through to prosemirror-tables' `handlePaste`, which calls `insertCells` for any table slice pasted inside a table — reintroducing the exact grid-overwrite content loss this PR prevents, just for range selections instead of carets.

Instead, the plugin now deletes the selected range in the same transaction before inserting the pasted table after the enclosing table, so paste-replaces-selection semantics hold. Since prosemirror-tables normalizes cross-cell ranges into `CellSelection`, a non-collapsed `TextSelection` here is always confined to a single cell, which keeps the deletion simple. Added the suggested test covering a range selection inside a cell.

⚡ 9450b50b2f5b743f617d30c340d3ec9a76d66d1d

Kiarokh:
Oh i just updated the branch. That explains the force push. But that was on mistake. I thought it was another PR I was looking at. Sorry

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
```diff
@@ -0,0 +1,63 @@
...
+    view.dispatch(tr);
+
+    return true;
+};
+
+// Narrow gate: only intercept when the pasted slice contains a
+// complete <table> at the top level. Bare <tr>/<td> slices would not
+// fit at block level (where the post-table insertion happens), so we
+// let the default prosemirror-tables flow handle them.
+const sliceContainsTable = (slice: Slice): boolean => {
```
jgroth, line 112: Rows copied inside the editor arrive as top-level `table_row` nodes, so this gate misses them and the old cell-grid replace still deletes rows.

Open a text editor example that has a table with three or more rows. Select cells that span two rows, copy, click into the first cell of the table, then paste. Rows one and two are overwritten instead of new rows being added.

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    if (!(state.selection instanceof TextSelection)) {
+        return false;
+    }
+
+    if (!sliceContainsTable(slice)) {
+        return false;
+    }
+
+    if (!isInTable(state)) {
```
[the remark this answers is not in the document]
jgroth: I think this should be fixed. Open a text editor example that has a table. Put the caret in the paragraph above the table, drag down and release in the middle of a cell in the first row, then paste a table copied from Excel. The first row disappears, and the pasted table lands below the whole table instead of where the selection was.

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    view.dispatch(tr);
+
+    return true;
+};
+
+// Narrow gate: only intercept when the pasted slice contains a
+// complete <table> at the top level. Bare <tr>/<td> slices would not
+// fit at block level (where the post-table insertion happens), so we
+// let the default prosemirror-tables flow handle them.
```
jgroth: I think there is too much AI word soup here that is hard to comprehend. Can we refactor this so the comments are not needed, or at least improve them a bit?

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+function paste(holder: FakeViewHolder, slice: Slice): boolean {
+    return createTablePastePlugin().props.handlePaste!(
+        holder.view,
+        new Event('paste') as ClipboardEvent,
+        slice
+    ) as boolean;
+}
+
+describe('table-paste-plugin', () => {
+    describe('caret inside a cell with TextSelection', () => {
```
jgroth: I advice against using describe blocks like this that adds no value and just indents the code, making it harder to read due to shorter line lengths

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
john-traas: Confirmed — cell-selection copies arrive as bare `table_row` nodes and fell through to the cell-grid replace. Rows now insert after the caret's row when their width matches the table; mismatched widths and tables using row/colspans fall back to a new table after the enclosing one (cell-selection copies are always rectangular, so that's always valid). Verified live with your repro: two copied rows paste as added rows, nothing overwritten.

⚡ 3f8474690c36c70b2e85f3faa64f769b3c7340cd

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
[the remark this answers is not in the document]
john-traas: Fixed — a selection reaching from outside the table into a cell now has the paste replace it in place as a block. Returning false instead would have handed the paste back to prosemirror-tables' cell-grid replace, so the plugin handles this case itself. Verified live with the drag-from-paragraph repro.

⚡ 3f8474690c36c70b2e85f3faa64f769b3c7340cd

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
john-traas: Rewrote the module — the prose blocks are replaced by named steps (`pasteRows`, `pasteAfterEnclosingTable`, `pasteInPlace`, `selectionStaysInsideOneTable`); the remaining comments are one-liners on non-obvious behavior only.

⚡ 3f8474690c36c70b2e85f3faa64f769b3c7340cd

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.spec.ts
john-traas: Flattened — one top-level describe, context folded into the test names.

⚡ 3f8474690c36c70b2e85f3faa64f769b3c7340cd

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
```diff
@@ -0,0 +1,176 @@
...
+};
+
+const pasteRows = (view: EditorView, rows: Fragment): boolean => {
+    const { state } = view;
+    const $cell = selectionCell(state);
+    const enclosingTable = $cell.node(-1);
+
+    // Splicing rows into a table with spanning cells can break its grid.
+    const spliceIsSafe =
+        rowWidth(rows.firstChild!) === TableMap.get(enclosingTable).width &&
```
jgroth: Not sure how this is supposed to work, but this might be a regression compared to main. This breaks copying a single cell and pasting it into another cell.

Reproduced in a table with two columns. Select text in one cell, copy, put the caret in another cell, paste.

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
```diff
@@ -0,0 +1,176 @@
...
+    if (!(state.selection instanceof TextSelection)) {
+        return false;
+    }
+
+    const pastedRows = sliceAsRows(slice);
+    if (!sliceContainsTable(slice) && !pastedRows) {
+        return false;
+    }
+
+    if (!isInTable(state)) {
```
jgroth: This gate makes the cross-boundary fix work in one direction only.

`isInTable(state)` reads `$head` alone (`prosemirror-tables` `util.ts`). My repro drags downward, so the head lands in a cell and the plugin takes over. Dragging the other way puts the head outside the table, the plugin returns false, and the default paste destroys the table just as before.

Repro: put the caret inside a cell in the second row, drag up into the paragraph before the table, then paste a table. Same result on `main` and on this branch:

`paragraph("befpasted")  TABLE{[r2c2]}  paragraph("after")` — the whole first row and one cell of the second row are gone, and the pasted table is flattened to plain text.

Dragging down out of the table is just as bad: the trailing paragraph disappears and the pasted table becomes text inside the cell.

Suggestion: this file already has `enclosingTablePos()`. Use it on both `$from` and `$to` instead of `isInTable(state)`, so a selection with either end in a table reaches `pasteInPlace`.

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
```diff
@@ -0,0 +1,176 @@
...
+        return false;
+    }
+
+    if (!selectionStaysInsideOneTable(state)) {
+        return pasteInPlace(view, slice.content, pastedRows);
```
jgroth: Not sure how this should work either, but dragging the other way round nests the pasted table inside the cell so there seems to be different behavior depending on how you make the selection.

Create a table with a paragraph after it. Put the mouse down in that paragraph, drag **up** into a cell, then paste a table from Excel.

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
john-traas: Fixed. The plugin now uses prosemirror-tables' own `pastedCells` check to decide what counts as a table paste, so copied cell text pastes as text again. Covered by a spec and a real copy/paste e2e.

⚡ 73a3e67fd784b92991ea54b7988fe225cf52b58c

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
john-traas: Fixed. The separate path for selections reaching out of the table is gone. The paste replaces the selection like Backspace would, then the rows land by the same rule as a caret paste, so nothing nests inside a cell in either drag direction. Covered in spec and e2e.

⚡ 73a3e67fd784b92991ea54b7988fe225cf52b58c

src/components/text-editor/prosemirror-adapter/plugins/table-paste-plugin.ts
```diff
@@ -0,0 +1,123 @@
+import { Plugin, PluginKey, Selection, Transaction } from 'prosemirror-state';
+import { Fragment, ResolvedPos, Slice } from 'prosemirror-model';
+import { EditorView } from 'prosemirror-view';
+import {
+    __Area as CellGrid,
+    __pastedCells as pastedCells,
```
jgroth: Is it safe to import and use these? The look suspiciously private to me 😄

john-traas: Oh my, I missed that... 🙈 
I'll look to see if there's any public entry point for this functionality instead.

john-traas: Fixed. The plugin now reads the pasted slice itself on public API only (`TableMap`, `tableNodeTypes`); `fixTables` in `tableEditing()` squares off ragged rows, with a spec for that case.

⚡ 3273d28e37800ef1454b2845d4597992253b016f

jgroth (approved):

PR #4085, opened 2026-05-25
fix(table): refresh remote paginator when totalRows or pageSize change

jensgustafsson:
## Summary by CodeRabbit

* **Bug Fixes**
  * Remote-mode pagination now refreshes when total rows or page size change: paginator UI updates, the table re-requests data, current scroll position is preserved after refresh, and transient failures during refresh are handled gracefully.

* **Tests**
  * Added tests covering remote paginator refresh, scroll restoration, and graceful failure handling.

## Summary
- In remote (ajax) mode, Tabulator's visible paginator buttons are driven by the `last_page` returned from `ajaxRequestFunc`, not by `setMaxPage`. As a result, changing `totalRows` or `pageSize` after init updated the internal max page but left the rendered paginator UI stale.
- The `totalRows` and `pageSize` watchers now also call a new `refreshRemotePaginator()`, which triggers `tabulator.replaceData()` so Tabulator re-invokes the ajax callback and the paginator picks up the new page count. Local mode is unaffected (guarded by `isRemoteMode()`).
- `replaceData()` rather than `setData()` because the latter calls `rowManager.resetScroll()` — visible as the table snapping to top-left on every totals refresh. `replaceData()` still rebuilds the row elements as a side-effect of going through the data pipeline (Tabulator's `renderInPosition` doesn't fully preserve vertical scroll across the rebuild), so the user's scroll position is saved and restored explicitly around the call.
- A small `getRowScrollContainer()` helper names the `.tabulator-tableholder` lookup so the magic selector lives in one place.

## Review:

## Test plan
- [ ] In remote mode, mount a table and update `totalRows` after init — paginator buttons reflect the new page count.
- [ ] In remote mode, change `pageSize` after init — paginator buttons reflect the new page count, current page is preserved (or clamped if it exceeds the new max).
- [ ] In remote mode, scroll the table vertically and horizontally, then update `totalRows` — scroll position is preserved (no snap to top-left).
- [ ] In local (non-remote) mode, changing `totalRows` / `pageSize` behaves as before (no extra data fetch).

### Browsers tested:
Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -318,11 +318,13 @@ export class Table {
     @Watch('totalRows')
     protected totalRowsChanged() {
         this.updateMaxPage();
+        this.refreshRemotePaginator();
     }
 
     @Watch('pageSize')
     protected pageSizeChanged() {
         this.updateMaxPage();
+        this.refreshRemotePaginator();
     }
 
     @Watch('page')
@@ -701,6 +703,56 @@ export class Table {
 
     private updateMaxPage() {
         this.tabulator?.setMaxPage(this.calculatePageCount());
+    }
+
+    /**
+     * In remote mode the visible paginator buttons are rendered from the
+     * `last_page` value returned by `ajaxRequestFunc`, not from
+     * `setMaxPage`. When `totalRows` or `pageSize` change after init, force
+     * Tabulator to re-invoke the ajax callback so the paginator UI picks up
+     * the new page count.
+     *
+     * `replaceData` rather than `setData` because the latter calls
+     * `rowManager.resetScroll()` (visible as the table snapping to
+     * top-left every refresh). `replaceData` still wipes and rebuilds
+     * the row elements as a side-effect of going through the data
+     * pipeline (causing a brief flicker), so save and restore the scroll
+     * position explicitly: Tabulator's `renderInPosition` path doesn't
+     * fully preserve vertical scroll across the row rebuild.
+     */
+    private async refreshRemotePaginator() {
+        if (!this.isRemoteMode() || !this.tabulator || !this.initialized) {
+            return;
+        }
+
+        const scrollContainer = this.getRowScrollContainer();
+        const scrollTop = scrollContainer?.scrollTop ?? 0;
+        const scrollLeft = scrollContainer?.scrollLeft ?? 0;
+
+        // `replaceData` resolves through `requestData`, which rejects when the
+        // component is destroyed, and Tabulator's ajax pipeline can reject too.
+        // Since the watchers call this without awaiting, swallow rejections here
+        // to avoid an unhandled promise rejection.
+        try {
+            await this.tabulator.replaceData();
+        } catch {
+            return;
+        }
+
+        if (scrollContainer) {
+            scrollContainer.scrollTop = scrollTop;
+            scrollContainer.scrollLeft = scrollLeft;
+        }
+    }
+
+    /**
+     * Tabulator's scrollable row container, inside `<limel-table>`'s
+     * shadow DOM. The user's vertical/horizontal scroll position within
+     * the table lives on this element's `scrollTop` / `scrollLeft`.
+     */
+    private getRowScrollContainer(): HTMLElement | null {
+        return (this.host.shadowRoot?.querySelector('.tabulator-tableholder') ??
+            null) as HTMLElement | null;
     }
 
     private getOptions(): TabulatorOptions {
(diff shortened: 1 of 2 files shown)


src/components/table/table.tsx
```diff
@@ -318,11 +318,13 @@ export class Table {
     @Watch('totalRows')
     protected totalRowsChanged() {
         this.updateMaxPage();
```
jgroth, line 320: `refreshRemotePaginator()` is async but called without `await` or `.catch()` from this synchronous watcher (same on line 325). `replaceData()` resolves through `requestData`, which returns `Promise.reject()` when `this.destroyed` is true, and Tabulator's own ajax pipeline can reject too. A rejection here becomes an unhandled promise rejection. Wrapping the body in a try/catch inside `refreshRemotePaginator`, or attaching `.catch()` at the call site, would contain it.

src/components/table/table.tsx
```diff
@@ -703,6 +705,46 @@ export class Table {
...
+     * Tabulator to re-invoke the ajax callback so the paginator UI picks up
+     * the new page count.
+     *
+     * `replaceData` rather than `setData` because the latter calls
+     * `rowManager.resetScroll()` (visible as the table snapping to
+     * top-left every refresh). `replaceData` still wipes and rebuilds
+     * the row elements as a side-effect of going through the data
+     * pipeline (causing a brief flicker), so save and restore the scroll
+     * position explicitly: Tabulator's `renderInPosition` path doesn't
+     * fully preserve vertical scroll across the row rebuild.
```
jgroth, line 721: table.spec.ts already drives the watchers against a mocked `tabulator.replaceData` (lines 27-37), so a test for this branch is cheap: assert that a `totalRows`/`pageSize` change in remote mode calls `replaceData()` with no args and that local mode does not, plus that scrollTop/scrollLeft are restored. None of the new paths are exercised right now.

src/components/table/table.tsx
```diff
@@ -703,6 +705,46 @@ export class Table {
...
+     */
+    private async refreshRemotePaginator() {
+        if (!this.isRemoteMode() || !this.tabulator || !this.initialized) {
+            return;
+        }
+
```
jgroth, line 727: Changing `pageSize` updates `last_page` via `calculatePageCount()`, but Tabulator's `paginationSize` is still the init value — the table keeps rendering the old number of rows per page even though the button count updates. This predates the PR rather than being introduced here, but the test plan claims `pageSize` changes are reflected, and a `setPageSize` call would make that true.

jgroth (commented):
Looks good, Claude found some stuff that should not be too hard to fix

src/components/table/table.tsx
jensgustafsson: Good point — wrapped the `await this.tabulator.replaceData()` in a try/catch inside `refreshRemotePaginator` that returns early on rejection (so the scroll-restore is also skipped, since the data wasn't replaced). Contains the rejection without it surfacing as an unhandled one. Fixed in the latest commit.

src/components/table/table.tsx
jensgustafsson: Added a `remote paginator refresh` suite driving the watchers against the mocked `tabulator`: asserts a `totalRows`/`pageSize` change in remote mode calls `replaceData()` with no args, that local mode does not refresh, that scrollTop/scrollLeft are restored across the rebuild, and that a `replaceData` rejection is swallowed. Thanks for pointing out it was cheap to cover.

src/components/table/table.tsx
jensgustafsson: Confirmed the local-mode `paginationSize` staleness is real. I started with a `setPageSize(this.pageSize)` call, but the only public API for it forces a `setPage(1)` afterwards (`userSetPageSize` → `setPageSize`; `setPage(1)`), which in local mode also emits a spurious `changePage(1)` via `handlePageLoaded`. That's a behavior change to the local-mode path this PR otherwise doesn't touch, so I'd rather not bundle it here and have it design the page-reset/event behavior deliberately.

Tracked separately in #4096. Keeping this PR scoped to the remote paginator.

jgroth (dismissed):

jgroth (approved):

PR #4100, opened 2026-06-02
test: smoke-test that documentation examples mount cleanly

adrianschmidt:
## What

Adds a browser e2e test (`src/examples-smoke.e2e.ts`) that mounts **every documented example component** and asserts it:

- upgrades (`customElements.get(tag)` is defined),
- renders non-empty output (shadow or light DOM), and
- logs no `console.error` (which is also where Stencil's `safeCall` routes lifecycle throws, and where the runner reports uncaught errors / unhandled rejections).

Examples are enumerated from the `@exampleComponent` doc tags in `components.d.ts`, so the list stays in sync as examples are added or removed. Runs inside the existing `e2e` Vitest project — no CI or config changes.

## Why

The ~440 example components shown in the docs had **zero automated coverage**. They silently bit-rot when a component's API changes, when an example throws on load, or when it renders nothing. This is a cheap, broad smoke layer over all of them.

## Notes

- **Already gating.** This runs inside the existing `e2e` Vitest project, so the standard `Test` CI job picks it up alongside the rest of the e2e suite — no separate job, and it gates like any other test.
- **8 composites skipped.** The `-composite` "playground" examples take a `schema` prop that the docs' `kompendium-debug` route injects before rendering them. Bare mounting can't reproduce that, so they're skipped here and covered instead by the docs-driven example tests, which render through the real `#/debug/` route (with the injected context).
- **One part of a multi-phase effort** to test the example components. Companion docs-driven suites load each example through the real docs/debug route to reach what in-process mounting can't — **runtime** (#4101) and **accessibility** (#4102) — with **visual** a natural future addition. Those run as separate, initially non-blocking CI jobs.

## Summary by CodeRabbit

* **Tests**
  * Added automated smoke and runtime end-to-end tests that load documented example components in a browser to verify they render without errors.

* **Chores**
  * Added test runner configuration and scripts, a lightweight local docs server for running example tests, CI job to run runtime example tests, and updated VCS ignore rules to exclude test attachment artifacts.


adrianschmidt:
# Consolidated PR Review

> 7-dimension multi-agent review (backward compatibility, code quality, architecture, security, observability, performance, Lime platform), synthesized into one report.

## PR Summary

PR #4100 adds a browser e2e smoke test (`src/examples-smoke.e2e.ts`) that mounts all ~440 documented example components (enumerated from `@exampleComponent` doc tags in the generated `components.d.ts` via a Vite `?raw` import) and asserts each one upgrades, renders non-empty output, and logs no `console.error`. 3 files changed, +130/−0 — test file, an ambient `*?raw` module declaration, and a `.gitignore` entry. Runs inside the existing gating `e2e` Vitest project.

The PR is strictly better than `main`: it adds broad coverage where there was none, introduces no production code paths, and the build/publish isolation was verified on every path (test file excluded via `tsconfig.dist.json`'s `**/*.e2e.*`, ambient declaration absent from the api-extractor rollup). No regressions found; the caveats are robustness improvements to the new suite itself.

- **Blockers:** None — no `[High]` or `[Critical]` findings across any dimension.
- **Non-blocking but worth addressing:** Five `[Medium]` follow-ups (async error-capture window, lossy `console.error` capture, `HARMLESS_MESSAGES` drift coupling, `EXPECTED_NOISY` granularity, gating-CI wall-clock) — see Top Recommendations.

---

Test-only PR with no API/export/prop/event changes. The published-build claim was verified: `.e2e.ts` is excluded by `tsconfig.dist.json` and the ambient declaration is absent from `etc/lime-elements.api.md`.

- **[Low]** `src/raw-import.d.ts:4` — the global ambient `declare module '*?raw'` is *not* matched by the dist excludes (`**/*.e2e.*`, `**/*.spec.*`) and is within the base `include: ["src"]`. If Stencil's `dist` target copies it into `dist/types/`, it would ship via `files: ["dist/"]` and globally augment consumers' `*?raw` resolution. Impact is minimal (only affects consumers who themselves use `?raw`; doesn't enter the api-extractor rollup), but verify once against a real `dist/types/` build, or move/rename it so the existing test excludes catch it.
- **Positives:** Correctly isolated from the published surface on every checked path; the `console.error` patch is restored per-test in `finally`.

## 2. Code Quality -- GOOD ✅

Core assumptions verified: `components.d.ts` contains exactly 440 `@exampleComponent` tags, the regex is sound, and all 8 skipped composites genuinely dereference `schema` on mount. Main concerns are robustness of the async error capture.

- **[Medium]** `examples-smoke.e2e.ts:98-101,110-115` — **async errors can escape the capture window and be misattributed.** The test restores `console.error` after `componentOnReady` + one `requestAnimationFrame`. Any lifecycle throw or rejection firing later (deferred microtask, `setTimeout`, late icon-load rejection) is either missed (false negative) or — since tests run sequentially sharing the real `console.error` — attributed to whichever test is running when it surfaces (flaky false positive on an unrelated example). Inherent to the bare-mount approach and acceptable for a smoke layer, but the most likely future flakiness source. (Performance's rAF-wait note is the same root cause.)
- **[Medium]** `examples-smoke.e2e.ts:44-64` vs `vitest-setup-browser.ts:6-13` — **`HARMLESS_MESSAGES` is likely dead code with a silent drift failure mode.** The setup file already `preventDefault()`s the Cache rejection, so in the normal case it never reaches `console.error` and the filter never fires. The magic string is duplicated across two files coupled only by a comment; if they drift, the rejection slips past `preventDefault`, no longer matches, and every test in the file fails. Extract a shared constant.
- **[Low]** `examples-smoke.e2e.ts:52-61` — `'reason' in value` is a structural duck-type, not a `PromiseRejectionEvent` check; any object with a `reason` property gets unwrapped. Affects only the diff message, not pass/fail.
- **[Low]** `examples-smoke.e2e.ts:71` — the `> 400` count guard catches "dropped to zero" but a silent loss of ~30 examples (440→415) would still pass.
- **Positives:** Enumeration-from-generated-source keeps the suite self-maintaining; the skip list is accurate (every entry verified to dereference `schema` on mount); the `console.error` swap is safe under the current sequential execution model.

## 3. Architecture -- GOOD ✅

Sound choices for a broad smoke layer — regex over `components.d.ts?raw` is the lighter-coupling option vs. kompendium metadata, and reusing the `e2e` project avoids new CI surface.

- **[Low]** `examples-smoke.e2e.ts:28-37` — the skip list is a hand-maintained denylist keyed on *mount-time dereference behavior*, invisible from the tag name. Verified correct today (9 composites declare `schema`; only the 8 that dereference it on mount are skipped — `button-group-composite` merely forwards it, so its inclusion is right). A future mount-time dereferencer fails loudly (fail-safe); a listed example that stops needing schema is silently skipped (low impact). Consider a comment pointing maintainers at the `componentWillLoad`/`this.schema.` pattern as the inclusion trigger.
- **[Low]** Composite-skip and harmless-error knowledge is now spread across three places (this file, `vitest-setup-browser.ts:9`, and #4101's HARMLESS list). The two layers are intentionally complementary (bare mount vs. real `#/debug/` route), so drift risk is real but minor; a shared constant would add cross-project coupling that may not pay for itself.
- **[Low]** `examples-smoke.e2e.ts:82-87,116-119` — the `console.error` monkey-patch is safe only under serial in-file execution; if the e2e project ever enables in-file parallelism, error attribution interleaves. Fine under Vitest browser-mode defaults.
- **Positives:** Genuinely low-coupling design — one file, no CI/config changes, fail-safe error capture, and a cheap coverage-drift guard.

## 4. Security -- SAFE ✅

No significant issues found. No supply-chain leakage (test file and ambient declaration verified unreachable from the published npm artifact on every path: `files` whitelist, `tsconfig.dist.json` excludes, api-extractor rollup), no exfiltration surface (`components.d.ts` is already public), and no new exposure paths.

- **[Low]** The `console.error` override + substring filter (`examples-smoke.e2e.ts:50,63-64,84-87`) is narrowly scoped to a documented test-env artifact and restored in `finally`; very unlikely to mask a security-relevant error.
- **Positives:** Build isolation is correct on every path checked.

## 5. Observability -- GOOD ✅

One named test per example means CI failures identify the offending component directly. The weaknesses are in how much diagnostic context survives a failure.

- **[Medium]** `examples-smoke.e2e.ts:84-86` — **lossy capture: only `args[0]` of each `console.error` call is kept.** Stencil and browsers often pass multi-arg errors (`console.error('Stencil render error', error, element)` or format strings); when `args[0]` is `'%s'`, the failure report shows a useless `["%s"]` while the real `Error` in `args[1]` is discarded. The full line reaches the live browser console, but Vitest's failure report shows the captured array. Capture all args (joined/stringified) to make failures self-diagnosing.
- **[Medium]** `examples-smoke.e2e.ts:42,110` — **`EXPECTED_NOISY` is a per-tag on/off switch, not message-level.** A tag added to silence one benign warning would also mask a genuine new mount crash in that component. Currently empty so no live blind spot, but prefer expected-*message* matching (as `HARMLESS_MESSAGES` already does) over blanket suppression.
- **[Low]** `examples-smoke.e2e.ts:74-76` — the 8 skips are visible in output but carry no reason string; an operator can't distinguish intentional-by-design from flaky-disable.
- **[Low]** No structured reporter for the e2e project — pre-existing, not a regression.
- **Positives:** Per-example named tests, tag-bearing assertion messages, and the `> 400` guard make most failures immediately actionable.

## 6. Performance -- ACCEPTABLE ✅

The dominant cost is gating-CI wall-clock, not memory: cleanup was verified clean (portals/popovers only attach to `document.body` on interaction the test never performs; `disconnectedCallback` teardown is proper; no `setInterval` anywhere in the examples).

- **[Medium]** `examples-smoke.e2e.ts:54` — 440 sequential tests in one file = one long-lived browser page, each paying per-`it` overhead plus a rAF tick; roughly **20–60s added to the gating Test job**. Linear and bounded, not pathological — a permanent cost to be aware of, not a regression.
- **[Low]** `examples-smoke.e2e.ts:73` — the single-frame rAF wait is a potential flake source for the heaviest examples, though verified low-risk (editors/tables render a shell synchronously; async-data examples like `table-remote` render a loading shell). Same root cause as Code Quality's `[Medium]` capture-window finding.
- **[Low]** `examples-smoke.e2e.ts:7` — the `?raw` import inlines the ~1.9 MB `components.d.ts` as a string in the test bundle; one `matchAll` pass, negligible.
- **Positives:** Verified-clean teardown across 440 mount/unmount cycles — no portal, overlay, timer, or observer accumulation.

**No LPID findings.** The lime-issue-detector catalog targets production runtime paths (hooks, endpoints, event handlers, Stencil components); this PR adds only test/dev-time artifacts. The plausibly-relevant frontend rules (LPID-FE-1/2 fetch loops, FE-6/7 state mutation, FE-8 cleanup, FE-10 XSS sink) were checked and confirmed not to apply — e.g. `shadowRoot.innerHTML` at `examples-smoke.e2e.ts:106` is a read, not a sink.

- **Positives:** No production runtime path introduced; proper monkey-patch restoration and DOM teardown.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Backward Compatibility | **GOOD** ✅ |
| Code Quality | **GOOD** ✅ |
| Architecture | **GOOD** ✅ |
| Security | **SAFE** ✅ |
| Observability | **GOOD** ✅ |
| Performance | **ACCEPTABLE** ✅ |
| Lime Platform Issues | **SAFE** ✅ |
| **Merge Readiness** | **MERGE WITH CAVEATS** ⚠️ |

## Top Recommendations

1. **[Medium from Observability]** Capture *all* `console.error` arguments (e.g. push the full `args` array or a stringified join) at `examples-smoke.e2e.ts:84-86`, so a failure report contains the actual error instead of potentially just a format string. Cheapest, highest-payoff fix.
2. **[Medium from Code Quality]** The post-`componentOnReady` single-rAF capture window can miss late async errors or misattribute them to the next test. Consider draining one more macrotask (`setTimeout(0)` after the rAF), and/or checking `consoleErrors` once more in a shared `afterEach` so stragglers fail the *right* test. If you accept the limitation, document it where `console.error` is restored.
3. **[Medium from Code Quality]** Extract the `"Failed to execute 'add' on 'Cache'"` string into a shared constant used by both `vitest-setup-browser.ts` and the test, eliminating the comment-coupled duplication — and clarify whether the test-side filter is load-bearing or defensive (the setup's `preventDefault()` likely makes it dead in the normal path).
4. **[Medium from Observability]** Change `EXPECTED_NOISY` from per-tag blanket suppression to per-tag *expected message* matching before the first entry is ever added, so a genuine new crash in a "noisy" component can't hide behind a known benign warning.
5. **[Medium from Performance]** Be aware the suite adds an estimated 20–60s to the gating Test job. No action required now; if it grows, splitting the file (Vitest parallelizes per-file) is the natural lever — but note that would require making the `console.error` swap parallel-safe first (Architecture's `[Low]`).
6. **[Low from Backward Compatibility]** Verify `dist/types/` from a real build doesn't include `raw-import.d.ts`; if it does, move the declaration so the existing `**/*.e2e.*` dist excludes cover it.

**Bottom line:** No blockers — a net-new, well-isolated smoke layer with verified build isolation and clean teardown. The caveats are all hardening of the suite's own error-capture machinery, best done before the suite starts gating merges at scale.

adrianschmidt:
## Review follow-up

Worked through the 7-dimension review. Two findings addressed in fixup `6edefc91a`; the rest deliberately skipped, with reasoning below.

### ✅ Addressed

**1. Capture all `console.error` arguments (Observability [Medium]).**
The capture kept only `args[0]`. Stencil and the browser often spread an error across several arguments (`console.error('render error', error, element)`) or pass a format string, so a failure report could show a useless `"%s"` while the real `Error` was discarded. Now the full `args` array is captured and joined (`formatErrorArgs`) for both the harmless-error filter and the failure message, so failures are self-diagnosing.

**6. Stop the `*?raw` ambient declaration shipping in `dist/types/` (Backward Compatibility [Low]).**
The review's suggested fix (rename so the `**/*.e2e.*` dist-exclude catches it) **doesn't actually work** — I tested it. Stencil's `dist` output copies *every* `.d.ts` under `src/` into `dist/types/` regardless of `tsconfig.dist.json`'s `exclude`, so both `raw-import.d.ts` and a renamed `raw-import.e2e.d.ts` still leaked into the published types (and would globally augment consumers' own `?raw` resolution). Real `.e2e.ts`/`.spec.ts` *source* files are excluded correctly — only standalone `.d.ts` files slip through.

The fix that works: inline `declare module '*?raw'` into `examples-smoke.e2e.ts` (which *is* excluded from `dist/types/`) and delete the standalone file. Verified: after a clean `npm run build`, `dist/types/` contains **zero** `raw-import` files, and the `?raw` import still type-resolves (transpile clean, e2e suite green: `441 tests | 8 skipped`).

### ⏭️ Skipped (with reasoning)

**2. Late-async-error capture window (Code Quality [Medium]).** Inherent to the bare-mount smoke approach. Adding a `setTimeout(0)` drain doesn't *fix* the race — it just shifts the window and adds flakiness surface. The review itself rates it "acceptable for a smoke layer." Left as-is; the existing comment notes the single-frame settle is intentional.

**3. Extract the shared `"Failed to execute 'add' on 'Cache'"` constant (Code Quality [Medium]).** The review's own Architecture [Low] argues the opposite — a shared constant "would add cross-project coupling that may not pay for itself." It's two short strings with a sync comment. Not worth the churn.

**4. `EXPECTED_NOISY` per-message instead of per-tag (Observability [Medium]).** The set is empty — no live blind spot. Designing message-level matching before a single entry exists is speculative; better designed when the first real entry appears.

**5. +20–60s on the gating Test job (Performance [Medium]).** Informational; the review says "no action required now." The sharded follow-ups (#4101–#4103) already address example-test wall-clock; this in-process suite intentionally rides the existing `e2e` job.

adrianschmidt:
# Consolidated PR Review (re-review)

> Re-review after fixup `6edefc91a`, which addressed two findings from the prior 7-dimension review (full `console.error` args capture; `*?raw` declaration inlined into the test file). 7-dimension multi-agent review, synthesized into one report.

## PR Summary

Adds a browser e2e smoke test (`src/examples-smoke.e2e.ts`) that mounts all ~440 documented example components (enumerated from `@exampleComponent` doc tags in the generated `components.d.ts` via Vite `?raw` import) and asserts each upgrades, renders non-empty output, and logs no `console.error`. 2 files changed, +146/−0 (test file + a `.vitest-attachments/` gitignore entry). Rides the existing gating `e2e` Vitest project; 8 schema-dependent composites are skipped and ceded to the docs-driven suites (#4101/#4102).

The PR is strictly better than `main` — broad coverage where there was none, no production code, and the fixup's two fixes were independently re-verified (dist isolation clean on all three publish paths; multi-arg capture correct for every realistic error shape). One new `[Medium]` surfaced in the fixup's own code: the error formatter keeps only `.message`, discarding the stack trace that a mount-failure report most needs.

- **Blockers:** None — no `[High]` or `[Critical]` findings in any dimension.
- **Non-blocking but worth addressing:** One `[Medium]` (stack-trace truncation, a one-line fix) and a handful of `[Low]`s — see Top Recommendations. The four previously-deferred caveats (async capture window, shared constant, `EXPECTED_NOISY` granularity, wall-clock) remain known and deferred with sound reasoning; none re-raised.

---

## 1. Backward Compatibility -- SAFE ✅

Test-only PR; published-artifact isolation after the fixup was verified mechanically, not assumed: `package.json` `files` is `["dist/", "index.scss", "CHANGELOG.md"]`; `tsconfig.dist.json` excludes `**/*.e2e.*`; the api-extractor rollup analyzes only `temp/types/index.d.ts` and `src/index.ts` never references the test or `?raw`; zero committed standalone `.d.ts` files remain under `src/`. Two pre-existing `.e2e.ts` files corroborate that `.e2e.ts` sources never reach `dist/types/`. The `.gitignore` entry is Vitest browser-mode's default attachments directory — no build/publish effect.

- **[Low]** The "inline `declare module '*?raw'` into an excluded `.e2e.*` file, never a standalone `.d.ts`" invariant is enforced only by the explanatory comment at `src/examples-smoke.e2e.ts:1-7`. A future contributor extracting it back to a shared `.d.ts` would silently reintroduce the `dist/types/` leak. Documentation-only residual risk.
- **Positives:** The previously-flagged leak vector is correctly closed by inlining into an excluded source — the rename suggested by the prior review would have failed, and the author tested that before choosing the working fix.

## 2. Code Quality -- GOOD ✅

The fixup's new code was traced through realistic argument shapes and holds up; the composite skip list was re-verified as correct and minimal (all 8 dereference `schema` during mount and would throw bare; non-skipped composites like `button-group-composite` only forward it). `components.d.ts` confirmed as a gitignored generated artifact, so enumeration stays in sync as claimed.

- **[Low]** `formatArg` (`examples-smoke.e2e.ts:74`) collapses plain objects without a `message` key to `"[object Object]"` (e.g. `{code: 500}` or a DOMException-like with only `.name`). Doesn't occur in the Stencil/runner error paths this suite actually sees; cosmetic only.
- **[Low]** `isHarmlessError` (`examples-smoke.e2e.ts:84-87`) substring-matches against the *join* of all args, so a single `console.error` call carrying both the harmless Cache message and a real error would be suppressed whole. Three agents converged on this; in practice the Cache rejection arrives as its own runner-emitted call (one `PromiseRejectionEvent` arg), never co-mingled, and the full line still reaches CI logs — only the assertion is affected. Speculative today; arg-level filtering would future-proof it. (Observability argued `[Medium]`; the negligible present likelihood and log-visible fallback keep it `[Low]`.)
- **[Low]** `HARMLESS_MESSAGES` is likely dead in the normal path — `vitest-setup-browser.ts:6-13` already `preventDefault()`s the icon-cache rejection. Defensible belt-and-suspenders (covers a race or future runner change); no action needed.
- **[Low]** A stale `@exampleComponent` doc tag naming an unregistered component would hang at `customElements.whenDefined` (`examples-smoke.e2e.ts:120`) and fail via test timeout rather than a clear assertion. Still a red test; acceptable.
- **Positives:** The skip list is precisely justified against real mount-time `schema` dereferences; the `> 400` count guard plus generated-artifact enumeration give a genuine coverage tripwire; the `console.error` swap is serial-safe under the project's single-chromium config and restored in `finally` even on assertion failure.

## 3. Architecture -- GOOD ✅

The fixup's mechanism is the right one: no cleaner hook exists in this repo — `vite/client` types aren't resolvable (Stencil bundles its own Vite), there's no test-only tsconfig in the Vitest browser project's type path, and a standalone `.d.ts` provably ships. Suite placement is a clean fit with the existing `e2e` project conventions; the coverage boundary with #4101/#4102 is explicit and non-overlapping.

- **[Low]** `REQUIRES_DOCS_SCHEMA` (`examples-smoke.e2e.ts:39-48`) is a hand-maintained denylist keyed on mount-time dereference behavior invisible from the tag name, and `kompendium-debug`'s schema injection lives outside this repo, so the rationale can't be verified from source here. The Architecture agent argued `[Medium]` because the list sits on the gating path — but the failure mode is fail-safe and self-attributing: a new schema-dereferencing composite fails loudly *in the PR that adds it*, with the test name pointing at the exact tag, and the fix is one line. That matches the prior review's `[Low]` assessment of the same unchanged code, so `[Low]` stands. Structural detection (`-composite` suffix + `schema` prop) is a reasonable follow-up if the list grows.
- **[Low]** The `e2e` project now spans two test intents: the two pre-existing `.e2e.ts` files are unit-style helpers needing only a DOM, while this is a true integration smoke depending on the built bundle loaded by `vitest-setup-browser.ts`. Taxonomy observation, not a regression.
- **Positives:** Single-source-of-truth enumeration with a zero-coverage tripwire; minimal-blast-radius type declaration that respects the verified Stencil copy-all-`.d.ts` constraint; no CI/config churn.

## 4. Security -- SAFE ✅

No significant issues found. Artifact isolation re-verified after the fixup via static config analysis (standalone `raw-import.d.ts` confirmed deleted; `tsconfig.dist.json` exclude; unchanged api-extractor surface); the only data read is the already-public generated `components.d.ts`; the `console.error` override mirrors to the original and is restored in `finally`, so nothing is suppressed from CI logs — the broadened join affects only this test's assertion, not security signal.

- **Positives:** Test-only change with no runtime, network, auth, or secret surface.

## 5. Observability -- GOOD ✅

Per-example named tests and tag-bearing assertion messages make failures attributable. The fixup fixed the multi-arg loss, but introduced a narrower truncation in the same code.

- **[Medium]** `formatArg` (`examples-smoke.e2e.ts:71-74`) returns `reason.message` alone for Errors, **discarding the stack trace** — the highest-value artifact for diagnosing which component/hook/file threw during a mount. A failure reading `["Cannot read properties of undefined (reading 'x')"]` says *that* something broke, not *where*. Since lifecycle errors routed through Stencil's `safeCall` arrive as Error objects, this hits the dominant failure mode. One-line fix: prefer `reason.stack` (V8 includes name+message) with `String(reason)` fallback. This is the direct continuation of prior finding #1, not a re-raise of a deferred item.
- **[Low]** Joined-args suppression blind spot — see Code Quality; same finding, cross-referenced.
- **[Low]** The 8 skips emit no reason in test output (`it.skip` at `examples-smoke.e2e.ts:98`); an operator must read the source to learn coverage is delegated to #4101.
- **[Low]** No structured (JUnit/JSON) reporter on the e2e project — pre-existing, flagged for completeness only.
- **Positives:** The count-guard test converts silent coverage collapse into a loud, self-explaining failure; the multi-arg capture is a real improvement over the prior single-arg loss.

## 6. Performance -- ACCEPTABLE ✅

Prior claims re-verified in the current worktree: portal `init()` (body-attached container + ResizeObserver) only runs on interaction, `disconnectedCallback` tears down fully, so 440 mount/unmount cycles accumulate nothing. The fixup introduced no new perf issue. The known ~20–60s gating wall-clock, single-rAF flake surface, and ~1.9 MB `?raw` string stand as previously assessed (deferred, informational).

- **[Low]** `formatErrorArgs` re-joins once per `HARMLESS_MESSAGES` entry plus once for the failure map — 1–2 redundant joins per test at current scale; negligible.
- **[Low]** Captured `args` arrays can hold DOM-element references, but only for one test's lifetime (local `const`, GC-eligible at `finally`); no cross-test accumulation.
- **Positives:** Symmetric per-test teardown verified clean against the actual portal implementation; no regression from the fixup.

**No LPID findings** (re-verified against the freshly fetched catalog, not carried over). The fixup's changes — error-arg formatting and the inlined ambient declaration — introduce no runtime anti-pattern; `shadowRoot.innerHTML` remains a read, not a sink, and cleanup/restore is disciplined.

- **Positives:** Test-only change with disciplined teardown and an ambient declaration deliberately scoped to keep it out of published consumer types.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Backward Compatibility | **SAFE** ✅ |
| Code Quality | **GOOD** ✅ |
| Architecture | **GOOD** ✅ |
| Security | **SAFE** ✅ |
| Observability | **GOOD** ✅ |
| Performance | **ACCEPTABLE** ✅ |
| Lime Platform Issues | **SAFE** ✅ |
| **Merge Readiness** | **MERGE WITH CAVEATS** ⚠️ |

## Top Recommendations

1. **[Medium from Observability]** In `formatArg` (`examples-smoke.e2e.ts:71-74`), return `reason.stack ?? String(reason)` instead of `reason.message` for Error instances, so a failed mount's report identifies *where* the throw happened. One-line change; directly serves the suite's core purpose.
2. **[Low from Code Quality]** Future-proof `isHarmlessError` by filtering at the argument level (drop only the harmless arg) rather than substring-matching the joined call — closes the theoretical both-in-one-call suppression path before a second `HARMLESS_MESSAGES` entry ever lands.
3. **[Low from Observability]** Attach a reason to the 8 `it.skip`s (e.g. a suffix in the test name pointing at #4101) so CI output self-documents the delegated coverage.
4. **[Low from Architecture]** If `REQUIRES_DOCS_SCHEMA` grows, replace instance enumeration with structural detection (`-composite` suffix + declared `schema` prop); for now a pointer comment suffices.

**Bottom line:** Both fixup fixes verified correct and complete; no regressions vs. `main`, no blockers. The single `[Medium]` is a one-line diagnosability improvement in the fixup's own formatter — worth folding in before merge, but not gating. Previously deferred caveats remain soundly deferred.

adrianschmidt:
## Re-review follow-up

Worked through the re-review. One finding addressed in fixup `901db25ac`; the rest skipped, with reasoning.

### ✅ Addressed

**1. Return the stack trace, not just `.message` (Observability [Medium]).**
`formatArg` returned `reason.message` for Errors, discarding the stack — the most useful artifact for diagnosing *where* a mount threw. It now returns `reason.stack ?? reason.message` (V8's `.stack` leads with the name and message, so it's a strict superset). Since Stencil routes lifecycle throws through `safeCall` as Error objects, this hits the dominant failure mode. The harmless-error filter is unaffected — the stack still contains the message it substring-matches on.

**3. Self-documenting skip reasons (Observability [Low]).**
The 8 schema-dependent composites now render in CI as `… (skipped: needs docs schema injection, covered by the docs-driven suite)` instead of a silent `skipped`, so an operator doesn't have to read the source to learn the coverage is delegated.

### ⏭️ Skipped (with reasoning)

**2. Argument-level filtering in `isHarmlessError` (Code Quality / Observability [Low]).** The theoretical gap — a single `console.error` call carrying *both* the harmless Cache message and a real error being suppressed whole — can't occur in practice: the icon-cache rejection is emitted by the runner as its own standalone `PromiseRejectionEvent` call, never co-mingled with a component error. The review downgraded this to `[Low]` ("negligible present likelihood", "speculative today") and the full line still reaches CI logs regardless. Pure future-proofing for a second `HARMLESS_MESSAGES` entry that doesn't exist; YAGNI.

**4. Structural detection for `REQUIRES_DOCS_SCHEMA` (Architecture [Low]).** The reviewer's own note: "for now a pointer comment suffices." The denylist is fail-safe (a new schema-dereferencing composite fails loudly *in the PR that adds it*, naming the exact tag) and the explanatory comment already documents the inclusion trigger. Worth revisiting only if the list grows.

### Carried-over caveats (still soundly deferred)
The four from the first review — async capture window, shared Cache-message constant, `EXPECTED_NOISY` granularity, and the ~20–60s gating wall-clock — remain deferred for the same reasons given previously; the re-review did not re-raise them.

adrianschmidt:
# Consolidated PR Review (re-review, round 3)

> Re-review after fixup `901db25ac` (stack-trace capture in `formatArg`; self-documenting skip reasons). 7-dimension multi-agent review, synthesized into one report.

## PR Summary

PR #4100 adds a browser e2e smoke test (`src/examples-smoke.e2e.ts`) that mounts all ~440 documented example components (enumerated from `@exampleComponent` doc tags in the generated `components.d.ts` via Vite `?raw` import) and asserts each upgrades, renders non-empty output, and logs no `console.error`. 2 files changed, +155/−0 (test file + a `.vitest-attachments/` gitignore entry). Rides the existing gating `e2e` Vitest project; 8 schema-dependent composites are skipped with self-documenting titles and ceded to the docs-driven suites (#4101/#4102).

The PR is strictly better than `main`, and round 3 surfaced **no new actionable findings**: both round-2 fixes were verified correctly implemented, every previously-deferred item was re-checked and its deferral reasoning still holds, and everything new this round is `[Low]`/speculative. The one remaining `[Medium]` (gating wall-clock) is an accepted, twice-adjudicated cost — informational, not a follow-up.

- **Blockers:** None — no `[High]` or `[Critical]` findings in any dimension, for the third consecutive round.
- **Non-blocking but worth addressing:** Nothing actionable remains. The handful of `[Low]`s below are cosmetic or speculative; the deferred caveats from rounds 1–2 stay soundly deferred.

---

## 1. Backward Compatibility -- SAFE ✅

No findings. Isolation re-verified mechanically at the current HEAD: `package.json` `files` = `["dist/", "index.scss", "CHANGELOG.md"]`; `tsconfig.dist.json` excludes `**/*.e2e.*`, so the inline `declare module '*?raw'` is never part of the dist TypeScript program — and because the file is a `.e2e.ts`, not a standalone `.d.ts`, even Stencil's copy-all-`.d.ts`-into-`dist/types/` behavior (the round-1 hazard) cannot apply. `src/index.ts` never references the test; `etc/lime-elements.api.md` contains no trace of it.

- **Positives:** The one plausible leak vector (the `?raw` module augmentation) is correctly contained inside an excluded `.e2e.ts` rather than a shippable `.d.ts`.

## 2. Code Quality -- GOOD ✅

The skip list was re-verified as precise: all 8 `REQUIRES_DOCS_SCHEMA` entries dereference `this.schema` in a lifecycle hook and would throw bare; `button-group-composite` only forwards `schema` and is correctly not skipped. The 440-tag count and regex were confirmed against source; the `> 400` guard fails loudly if the generated `components.d.ts` is absent.

- **[Low]** `examples-smoke.e2e.ts:77` — `formatArg` collapses a non-Error object lacking `.message` to `"[object Object]"`. Known from round 2; realistic payloads here are Errors or strings. Cosmetic.
- **[Low]** `examples-smoke.e2e.ts:124-126` — the hand-written `as HTMLElement & { componentOnReady?: ... }` cast; safe via optional chaining, acceptable for a test helper.
- **[Low]** `examples-smoke.e2e.ts:62` vs `vitest-setup-browser.ts:9` — the duplicated Cache-message literal; verified the setup file `preventDefault()`s the rejection, so `HARMLESS_MESSAGES` is dead belt-and-suspenders in the normal path. Deferral reasoning from round 1 re-checked and holds (see also Observability's reachability note — same root).
- **Positives:** Exemplary why-not-what comments on every non-obvious decision; small single-responsibility helpers with `unknown` typing; robust `try/finally` cleanup; the count-guard meta-test is a genuine coverage tripwire.

## 3. Architecture -- GOOD ✅

Placement and coupling re-verified: `vitest.config.ts` scopes the file to the `e2e` project via `include: ['src/**/*.e2e.{ts,tsx}']`; `components.d.ts` is confirmed gitignored/generated and regenerated before the e2e run; the denylist was mechanically validated against all 16 `*-composite` sources.

- **[Low]** `examples-smoke.e2e.ts:20` — residual coupling to the `@exampleComponent` doc-tag spelling emitted by Stencil into `components.d.ts`; inherent to the (correct) enumeration strategy, guarded by the count test. Completeness only.
- **[Low]** `examples-smoke.e2e.ts:40-49` — `REQUIRES_DOCS_SCHEMA` hand-curation: re-verified correct and fail-safe; the structural-detection alternative stays adjudicated `[Low]` (third concurring assessment).
- **[Low]** The `e2e` project now spans unit-style helpers and a true integration smoke — taxonomy note carried from round 2, not a regression.
- **Positives:** Zero config/CI surface change; the #4101/#4102 boundary is explicit, documented, and mechanically correct.

## 4. Security -- SAFE ✅

No significant issues. Isolation re-verified at three independent layers (files whitelist, dist tsconfig exclude, api-extractor rollup from `temp/types/`); the `console.error` override mirrors every call to the original, so nothing is suppressed from CI logs; the only data read (`components.d.ts`) is already public.

- **[Low]** `examples-smoke.e2e.ts:46-50` — substring allowlist could in theory filter a genuine error containing the Cache message from the *assertion* (never from logs). Same `[Low]` as round 1; completeness only.
- **Positives:** Test-only change with no new exposure path; all three isolation layers verified intact.

## 5. Observability -- GOOD ✅

Both round-2 fixes verified correctly implemented: `reason.stack ?? reason.message` at line 75 (and the harmless-filter `.includes()` remains correct since V8 stacks contain the message), and the 8 skips now render self-documenting titles pointing at the docs-driven suite.

- **[Low]** `examples-smoke.e2e.ts:142-147` — on failure, the tag is in the assertion message while the formatted errors are in the `toEqual([])` diff; the reader correlates two parts of the output. All information is present; minor.
- **[Low]** `examples-smoke.e2e.ts:62` + `vitest-setup-browser.ts:6-13` — whether the harmless-error filter is load-bearing or dead depends on whether the `preventDefault()`'d rejection still reaches `console.error`; the suite never verifies its own filter's reachability. Speculative; same root as Code Quality's duplication note — deferral stands.
- **[Low]** `formatArg`'s `"[object Object]"` fallback — see Code Quality; cross-referenced.
- **Positives:** Full-args capture, stack traces, rejection unwrapping, per-tag assertion context, self-documenting skips, and the count guard make CI failures directly diagnosable.

## 6. Performance -- ACCEPTABLE ✅

No new performance issues from the fixups; the happy path skips the formatting helpers entirely (they run only when errors were captured). Teardown re-verified clean across 440 mount/unmount cycles.

- **[Medium]** `examples-smoke.e2e.ts:88-153` — the known ~20–60s added to the gating Test job (440 sequential tests, single chromium instance). Verified **not** materially worse than the prior estimate; no sleeps/polling/retries added. Deferred twice with accepted reasoning (the sharded suites #4101–#4103 own example-test wall-clock; this suite intentionally rides the e2e job). Recorded for severity continuity, not re-raised as actionable.
- **[Low]** `examples-smoke.e2e.ts:139` — single-frame rAF wait remains the only timing assumption; prior verification (shells render synchronously) unchanged.
- **[Low]** `examples-smoke.e2e.ts:30-37` — one-time `matchAll` over the ~1.9 MB inlined string; negligible, unchanged.
- **Positives:** Per-test teardown accumulates no DOM, listeners, or memory; error-formatting cost is failure-path-only.

**No LPID findings** (re-verified against the freshly fetched catalog, not carried over). The only potentially relevant category (LPID-FE-*) does not match: no fetch loops, no state mutation, no lifecycle blocking; `shadowRoot.innerHTML` at `examples-smoke.e2e.ts:138` is a read, not a sink; monkey-patch and host node torn down in `finally`.

- **Positives:** No production runtime path; disciplined teardown throughout.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Backward Compatibility | **SAFE** ✅ |
| Code Quality | **GOOD** ✅ |
| Architecture | **GOOD** ✅ |
| Security | **SAFE** ✅ |
| Observability | **GOOD** ✅ |
| Performance | **ACCEPTABLE** ✅ |
| Lime Platform Issues | **SAFE** ✅ |
| **Merge Readiness** | **READY TO MERGE** ✅ |

## Top Recommendations

No actionable recommendations remain — there are no `[High]`/`[Critical]` findings, and the single `[Medium]`:

1. **[Medium from Performance]** — the ~20–60s gating wall-clock cost — is the deliberately-accepted, twice-deferred cost of riding the existing e2e job; re-verified this round as not worse than estimated. No action; revisit only if the suite grows enough to warrant file-splitting (which would first require making the `console.error` swap parallel-safe).

The remaining `[Low]`s (`"[object Object]"` fallback, filter-reachability self-verification, assertion-message/diff correlation) are cosmetic or speculative — fine to leave, or fold into a future hardening pass if the suite ever starts producing hard-to-read failures.

**Bottom line:** Third round, zero new actionable findings. Both round-2 fixes verified correct; every deferred caveat re-checked and still soundly deferred; all seven dimensions green. The PR is strictly better than `main` and ready to merge.

Kiarokh (approved):

PR #4110, opened 2026-06-04
fix(collapsible-section): keep collapsed content out of the tab order

### LINKED ISSUE #4109: Collapsible section: focusable elements inside it, can still be focused even when the section is closed
## Current behavior

https://github.com/user-attachments/assets/b4b1589e-557c-4ffd-a7df-26e1277b3f5d

## Expected behavior
A collapsed section is hidden only visually (the body animates via grid height + opacity, never display:none — which can't be animated), so its fields stayed keyboard-focusable while marked aria-hidden: a sighted keyboard user could Tab into invisible content, and a screen reader could land on a focused aria-hidden node. axe flags this as aria-hidden-focus.

## How to fix
Mark the body `inert` (see https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/inert )while collapsed, in sync with aria-hidden, so its content leaves the tab order and the a11y tree. inert is layout-neutral (animation unaffected) and propagates through the slot to projected content.

Kiarokh:
Fixes #4109

## Summary

A collapsed `limel-collapsible-section` is hidden only **visually** — the body animates via CSS grid height + opacity, never `display: none` (which can't be animated). Its fields therefore stayed **keyboard-focusable while marked `aria-hidden`**: a sighted keyboard user could Tab into invisible content, and a screen reader could land on a focused `aria-hidden` node. axe flags this as `aria-hidden-focus`.

## Fix

Mark the body `inert` while collapsed, in sync with `aria-hidden`:

- `inert` removes the subtree from the tab order **and** the accessibility tree.
- It's layout-neutral, so the open/close animation is unaffected.
- It propagates through the `<slot>` to projected (light-DOM) content.

## Tests

New e2e tests assert the body is `inert` when collapsed (and not when open), and — the real proof — that a **slotted** focusable element can't receive focus while collapsed but can once expanded, confirming `inert` reaches projected content in a real browser. 15/15 collapsible-section e2e pass.

## Accessibility baseline

Removed `aria-hidden-focus` from the six examples whose only source of it was a collapsed section: `builtin-field-types-form`, `collapsible-section-css-props`, `collapsible-section-with-slider`, `form-layout`, `nested-form`, `tooltip-composite`. Only that rule was touched — e.g. `collapsible-section-with-slider` keeps its `color-contrast` entry.

## Test plan

- [ ] Open a `limel-collapsible-section` (or any form with a collapsed section) and Tab through it — focus skips the collapsed content, and reaches it once expanded.
- [ ] `npm test` — collapsible-section + form suites pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved accessibility for collapsible sections — collapsed content is now non-interactive and cannot receive focus.

* **Tests**
  * Added keyboard/navigation coverage to verify `inert` and focus behavior for collapsible section content when collapsed by default and after opening.
  * Updated accessibility baseline expectations to match refreshed rule outputs.

### DIFF
--- a/example-tests/axe-baseline.json
+++ b/example-tests/axe-baseline.json
@@ -20,7 +20,7 @@
     "nested-interactive"
   ],
   "limel-code-diff": ["aria-required-children", "color-contrast"],
-  "limel-collapsible-section": ["aria-hidden-focus", "color-contrast"],
+  "limel-collapsible-section": ["color-contrast"],
   "limel-dock": ["button-name"],
   "limel-email-viewer": ["color-contrast", "image-alt"],
   "limel-example-file-viewer-basic": ["heading-order"],
--- a/src/components/collapsible-section/collapsible-section.e2e.tsx
+++ b/src/components/collapsible-section/collapsible-section.e2e.tsx
@@ -113,6 +113,36 @@ describe('limel-collapsible-section', () => {
                     body.getAttribute('id')
                 );
             });
+
+            it('marks the body as `inert` when collapsed', () => {
+                const body = root.shadowRoot.querySelector('.body');
+                expect(body.hasAttribute('inert')).toBe(true);
+            });
+
+            it('removes `inert` from the body when open', async () => {
+                await setProps({ isOpen: true });
+                await waitForChanges();
+
+                const body = root.shadowRoot.querySelector('.body');
+                expect(body.hasAttribute('inert')).toBe(false);
+            });
+
+            it('keeps slotted content out of the tab order when collapsed', () => {
+                const button = root.querySelector('button') as HTMLElement;
+                button.focus();
+
+                expect(document.activeElement).not.toBe(button);
+            });
+
+            it('lets slotted content receive focus when open', async () => {
+                await setProps({ isOpen: true });
+                await waitForChanges();
+
+                const button = root.querySelector('button') as HTMLElement;
+                button.focus();
+
+                expect(document.activeElement).toBe(button);
+            });
         });
 
         describe('when setting `isOpen` to `true`', () => {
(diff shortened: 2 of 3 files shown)


Kiarokh:
@coderabbitai ful review please

example-tests/axe-baseline.json
```diff
@@ -1,195 +1,97 @@
...
+  "limel-prosemirror-adapter": ["aria-valid-attr-value"],
+  "limel-shortcut": ["aria-prohibited-attr"],
+  "limel-switch": ["button-name"],
+  "limel-table": [
+    "aria-required-children",
+    "aria-required-parent",
+    "color-contrast",
+    "empty-table-header"
+  ]
 }
```
adrianschmidt, line 72: Was this linted? If your autolinter can't respect that this file is excluded from linting, we will need to include it in the forced linting, so that it doesn't keep flipping between different formatting in every other PR 😣

adrianschmidt: Ah, wait, no, it's probably `.editorconfig`!

I'll see what I can do about it. Either add special rules for this file to `.editorconfig`, or change how the file is generated.

adrianschmidt: No, I changed my mind again. `.editorconfig` doesn't change what's already written in a file, it only affects the editor's behaviour.

adrianschmidt: (I'm fixing this, so the generator outputs the file already prettified.)

adrianschmidt: Fixed in #4113

LucyChyzhova (approved):
great improvement!!!

PR #4112, opened 2026-06-05
test: key the axe baseline by component instead of example

adrianschmidt:
## What

Re-keys the axe baseline from **example tags** to the **components that cause the violations**, and makes the local example-test commands hermetic against stale doc builds.

## Why

With the accessibility checks now gating, adding a *new example* that demonstrates a component with already-accepted a11y debt failed the gate: the baseline was keyed by example tag, so the new example had no entry and every violation counted as "new" — even though the author introduced no new problem.

## How

- Each violating node is attributed to its **nearest enclosing `limel-*` component**, walking up through shadow roots over real DOM nodes (axe's `elementRef` option — no selector parsing). Nodes in an example's own markup fall back to the example tag.
- Only genuinely new **(component, rule)** pairs fail now, regardless of which example exposes them. A new example reusing components with known debt passes by construction — the example tag no longer participates in the lookup for component-attributed nodes.
- **The regenerated baseline collapses from 303 example entries to 52 keys:** 32 component entries (in effect a per-component a11y debt register) + 20 example-fallback entries (almost all `heading-order` in examples' own markup).
- Ratchet semantics unchanged (only shrinks; bootstrap when the file is absent), with one new guard: **regeneration requires a full run** — with component keys, a partial run can't prove a pair is gone anywhere, so the writer refuses incomplete observations.

## Tooling (first commit)

The example tests run against the built docs, and Stencil's incremental cache can serve a stale build after branch switches (it has silently skipped regenerating `kompendium.json`). Each `test:examples*` command now cleans `www/` + `.stencil/` and rebuilds before running, via new bare `:ci` variants that just run Playwright:

- CI's sharded jobs call the `:ci` variants directly (they download the prebuilt `example-docs` artifact and must not rebuild).
- `npm run test:examples:accessibility:update` added for baseline regeneration (previously a hand-typed env-var invocation).
- Composite commands end with a trailing `--`, so Playwright args still pass through (e.g. `npm run test:examples:accessibility -- -g "limel-example-list"`).

## Verification

Full local run against the regenerated baseline: **442 passed** (441 examples + the enumeration guard), baseline written by a complete single-worker update run.


adrianschmidt:
# Consolidated PR Review (6-agent + Lime platform)

## PR Summary

PR #4112 re-keys the axe accessibility baseline in lime-elements from example tags to the components that cause the violations (attributed via a shadow-root-piercing DOM walk using axe's `elementRef`), collapsing the baseline from 303 example entries to 52 keys. It also makes local `test:examples*` commands hermetic by cleaning `www/` + `.stencil/` and rebuilding before each run, with new bare `:ci` variants for CI's sharded jobs. 4 files changed, +224 −1157.

The PR is strictly better than the target branch: no regressions were found in any dimension, the attribution logic is verified correct, CI is correctly insulated from the new local rebuild, and the new write guards make baseline regeneration *safer* than before.

- **Blockers:** None
- **Non-blocking but worth addressing:** See Top Recommendations — all `[Low]` polish items (a stale eslint-config comment, silent success path of the baseline writer, guard-message clarity).

---

The PR renames/splits the example-test npm scripts (adding `:ci` bare variants), rewrites the a11y baseline format from example-keyed to component-keyed, and makes the a11y gate more permissive for already-accepted component debt. All consumers of the changed contracts (CI workflow, eslint ignore list, docs) were checked — no backward-compatibility regressions found.

- **[Low]** `eslint.config.mjs:15` — the comment still says the baseline is "Generated by `UPDATE_AXE_BASELINE=1`; not hand-edited", but the spec's own guidance (`accessibility.spec.ts:303-306`) now instructs developers to add deliberately-accepted `(component, rule)` debt to the file *by hand*. The file remains eslint-ignored, so nothing breaks — a stale comment, not a functional issue.
- *The gate-granularity change (example-keyed → component-keyed tolerance) is a deliberate, justified trade-off described in the PR body and the spec's header comment — not a defect. Covered in Code Quality (cross-reference).*
- **Positives:**
  - Both CI call sites correctly migrated to the no-rebuild `:ci` variants matching the artifact-download model (rebuilding would `rm -rf www` over the prebuilt `example-docs` artifact, so the switch is required for correctness); no orphaned references to renamed scripts anywhere in workflows, actions, or docs.
  - The format migration is fully self-contained (the only baseline consumer is the spec itself; eslint merely ignores it) and the new full-run write-guard closes the soundness hole that component-keying would otherwise open.

## 2. Code Quality -- GOOD ✅

The new attribution logic is well-structured, well-commented, and internally consistent; types inside `page.evaluate` are sound, the serialization boundary is respected (only strings cross the bridge), and the committed baseline (52 keys, sorted) matches what the writer would emit.

- **[Low]** Informational, not a defect: keying shared base components (e.g. `limel-button: color-contrast`) by (component, rule) means a new example exposing an *already-debted* pair passes silently even if that specific instance is a fresh, fixable problem. This is the explicit, documented intent of the PR (a per-component debt register that lets new examples reuse debted components without re-failing), and the trade-off — and its bounding to existing pairs only — is spelled out in the file comment and PR description. The reasoning is sound; noted only so reviewers are aware the gate granularity is now per-component rather than per-example.
- **[Low]** `componentOf` (`accessibility.spec.ts:210`) declares a default parameter (`node: Element | null = null`) but every call site passes an explicit argument. The default is dead; minor polish.
- *Retry double-counting of `scannedExamples` is covered in Observability (cross-reference) — fails safe (refuses to write), and the documented update workflow runs with `retries: 0`.*
- **Positives:**
  - Excellent self-documenting comments that state the design rationale, the full-run requirement, and the ratchet/bootstrap semantics inline where the logic lives.
  - The shadow-root walk correctly hops boundaries one at a time (`parentElement ?? rootNode.host`), terminates at `exampleHost`, and skips scaffolding; `elementRef: true` is the right axe option to avoid selector-string parsing.
  - Failure messages and the single-writer `beforeAll` guard are designed to fail loud and safe rather than silently corrupt the baseline.

## 3. Architecture -- GOOD ✅

The component-attribution model is sound and the architecturally correct unit (debt lives in components, not examples); module-level mutable state is safe under Playwright's worker model given the single-worker enforcement.

- **[Low]** Cross-component-boundary rules (e.g. `aria-required-parent`) are attributed to whichever single node axe flags, which may not be the component that owns the fix — the diff itself shows `limel-action-bar` carrying `aria-required-children` while `limel-action-bar-item` carries `aria-required-parent` for the same structural relationship. Behavior is correct and deterministic; only the "one component owns this rule" mental model is imperfect. At most a one-line caveat in the comment block.
- **[Low]** The full-run write guard (`accessibility.spec.ts:103-109`) couples regeneration to a fully green single-worker run: with 440+ examples and 0 local retries, one flaky example forces a complete re-run. Inherent tension of component-keying (a partial run can't prove a pair is gone), acknowledged in the PR — flagging the scaling cost, not asking for a fix.
- **[Low]** `scannedExamples` does double duty as "examples scanned" including clean ones — accurate, just slightly undersold by the name.
- **Positives:** The single-writer invariant is defended in depth: config pins workers to 1, `beforeAll` rejects a `--workers` override, and `afterAll` refuses incomplete observations — three independent guards against a clobbered baseline.

## 4. Security -- SAFE ✅

No significant issues found. Every changed path was traced for injection, exposure, and untrusted-input surfaces: `AXE_SOURCE` comes from the pinned `node_modules` dependency; the baseline writer uses a fixed path, gated on a local-only env var, writing only public data (axe rule IDs, tag names); the workflow changes add no new `${{ }}` interpolation; the npm scripts interpolate no untrusted input. The loosened gate governs only a11y debt and cannot mask anything security-relevant.

- **Positives:** `permissions: contents: read`, pinned action SHAs, and the env-var-gated writer keep the tooling's attack surface minimal; the `:ci` split reduces what runs in CI rather than adding surface.

## 5. Observability -- GOOD ✅

The new failure messages and guard errors are markedly more actionable than the target branch; the only gaps are on the success path and guard-message edge cases.

- **[Low]** The baseline writer (`accessibility.spec.ts:95-124`) emits no signal distinguishing "baseline written" from "nothing written" — no path, key count, or dropped-pair count. A developer must diff the file by hand. The one asymmetry in an otherwise well-instrumented flow.
- **[Low]** The incomplete-run guard's error message lists "a test filter, or examples that errored out" as causes but not sharding — combining `UPDATE_AXE_BASELINE=1` with `--shard` would throw a confusing count mismatch on every shard. Footgun-clarity nit; update mode isn't wired into any sharded command.
- **[Low]** `scannedExamples++` (`accessibility.spec.ts:277`) isn't retry-idempotent; Playwright retries would double-count and trip the exact-equality guard with a confusing "N of M" where N > M. Can't fire today (local retries are 0; it would also fail safe by refusing to write) — only relevant if retries are ever enabled for the update path.
- **Positives:** The reworked failure message names the component, rule, element target, and axe summary, then states the two valid remediations — a developer hitting a failure knows exactly what to do; every refuse-to-write path throws with the file path and a concrete next step.

## 6. Performance -- GOOD ✅

Nothing ships to the npm package; CI is correctly insulated from the new rebuild via the `:ci` variants consuming the shared prebuilt artifact. No CI time or artifact-size regression.

- **[Low]** `docs:rebuild` deletes the Stencil incremental cache on every local run, so each invocation pays a full cold rebuild even when nothing changed — eliminating the fast incremental path for developers iterating on tests against an unchanged tree. The stated point of the change, consciously chosen; a lighter incremental variant could be offered alongside.
- **[Low]** `componentOf` is O(DOM depth) per violating node with no `element → component` memoization across nodes sharing ancestors — trivial at actual data sizes, noted for completeness only.
- **[Low]** Per-node linear `Array.includes` over baseline rule lists (`accessibility.spec.ts:288`) — lists are ≤ ~10 entries; a `Set` would be O(1) but data sizes don't justify it.
- **Positives:** The re-keying shrinks the baseline 303 → 52 keys, and `elementRef` element handles never cross the `page.evaluate` serialization boundary — no leak.

No findings. The lime-issue-detector catalog scopes its rules to Lime runtime contexts (UoW hooks, endpoints, event handlers, Celery tasks, `lime_file` I/O, shipped Stencil components, Python test fixtures) — none exist in this diff. The in-browser spec code was checked specifically: only `axe.run` and bounded shadow-root traversal, no `innerHTML`/`eval` sinks, and the `componentOf` walk terminates at `exampleHost`/null. No CRITICAL/HIGH/MEDIUM LPID pattern is introduced.

- **Positives:** The partial-run and multi-worker guards make the test infrastructure more robust than the target branch.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Code Quality | **GOOD** |
| Architecture | **GOOD** |
| Security | **SAFE** |
| Observability | **GOOD** |
| Performance | **GOOD** |

## Top Recommendations

1. **[Low from Backward Compatibility]** Update the stale comment at `eslint.config.mjs:15` — the baseline is no longer strictly "not hand-edited" now that deliberately-accepted debt entries are meant to be added by hand.
2. **[Low from Observability]** Log a one-line success summary from the baseline writer ("wrote 52 keys to …, dropped N fixed pairs") so an update run's outcome is visible without diffing the file.
3. **[Low from Observability]** Mention sharding in the incomplete-run guard's error message as a possible cause of the count mismatch.
4. **[Low from Architecture]** Add a one-line caveat to the baseline comment that cross-component rules (`aria-required-parent`/`-children`) may split one structural issue across two keys.

The review found no regressions — the PR is a net improvement over `main` in every dimension, with notably strong defense-in-depth around baseline regeneration (single-worker pin, `--workers` override rejection, full-run write guard).

---

*Edited: the Backward Compatibility and Code Quality sections were re-synthesized from re-run review agents after a fix to the review skill. The previous `[Medium]` finding ("deliberate gate weakening … deserves a conscious sign-off") was retracted: it flagged a trade-off explicitly documented in the PR description, commit message, and code comments, and its supporting claim that the old per-example keying "would have caught" a new violation of a baselined pair did not hold up (a new example failed indiscriminately under the old keying, and the remediation would have absorbed the violation). The rerun also surfaced one new finding (the stale `eslint.config.mjs` comment, now Recommendation 1).*

🤖 Generated with [Claude Code](https://claude.com/claude-code) — 6-agent review

Kiarokh (approved):
much better 🙌

PR #4134, opened 2026-06-16
feat(code-diff): add interactive playground example

### LINKED ISSUE #4133: Add an interactive playground to the code-diff examples page
## New feature motivation

The `limel-code-diff` examples page demonstrates the component with hard-coded inputs. To explore the component, a reader has to edit example source or imagine how their own content would render. An interactive playground lets a reader paste in their own "before"/"after" content and immediately see the resulting diff, while toggling the available display options — a much faster way to evaluate the component.

## New feature description

Add an interactive example at the **top** of the code-diff examples page:

- Two editable `limel-code-editor` boxes (old value / new value), side by side.
- A live `limel-code-diff` below them that updates as the reader types.
- Controls to switch the `layout` between `unified` and `split`, and to toggle `lineWrapping` on/off.

## New feature implementation

- New example component `limel-example-code-diff-interactive`, registered as the first `@exampleComponent` on `limel-code-diff`.
- Editors feed the diff via their `change` event; the layout/line-wrapping controls reuse the existing `limel-example-controls` + `limel-select` + `limel-switch` pattern.
- Extract the `layout` prop's inline `'unified' | 'split'` union into a named, public `CodeDiffLayout` type so the example (and consumers) can reference it instead of repeating the literal.

viktorSoftDev:
Closes #4133

## Summary by CodeRabbit

* **New Features**
  * Added a new interactive code diff playground example to switch between unified and split layouts, toggle line wrapping, and preview changes live with customizable code samples.

* **Refactor**
  * Improved code diff layout typing by introducing a shared `CodeDiffLayout` type and updating component/example props and selections to use it.

* **Style**
  * Enhanced the interactive example layout responsiveness and editor sizing behavior for better viewing on different screen widths.

* **Documentation**
  * Updated component JSDoc to include an additional example reference.

Adds an interactive playground to the top of the `limel-code-diff` examples page: two editable `limel-code-editor` boxes feeding a live `limel-code-diff`, with controls for `unified`/`split` layout and `lineWrapping`. Also extracts the `layout` prop's inline union into a named, public `CodeDiffLayout` type.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/code-diff/code-diff.tsx
+++ b/src/components/code-diff/code-diff.tsx
@@ -1,6 +1,7 @@
 import { Component, Element, h, Prop, State, Watch, Host } from '@stencil/core';
 
 import { Languages } from '../date-picker/date.types';
+import { CodeDiffLayout } from './code-diff.types';
 import translate from '../../global/translations';
 import {
     DiffHunk,
@@ -74,7 +75,7 @@ export class CodeDiff {
      * - `split` — side-by-side comparison with old on left, new on right
      */
     @Prop({ reflect: true })
-    public layout: 'unified' | 'split' = 'unified';
+    public layout: CodeDiffLayout = 'unified';
 
     /**
      * Number of unchanged context lines to display around each change.
--- a/src/components/code-diff/code-diff.tsx
+++ b/src/components/code-diff/code-diff.tsx
@@ -34,6 +34,7 @@ import {
  * @exampleComponent limel-example-code-diff-split
  * @exampleComponent limel-example-code-diff-line-wrap
  * @exampleComponent limel-example-code-diff-expand
+ * @exampleComponent limel-example-code-diff-interactive
  */
 @Component({
     tag: 'limel-code-diff',
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -112,6 +112,9 @@ export interface ClosingActions {
     scrimClick: boolean;
 }
 
+// @public
+export type CodeDiffLayout = 'unified' | 'split';
+
 // Warning: (ae-incompatible-release-tags) The symbol "Color" is marked as @public, but its signature references "_Internal" which is marked as @internal
 //
 // @public
@@ -325,7 +328,7 @@ export namespace Components {
     export interface LimelCodeDiff {
         "contextLines": number;
         "language"?: string;
-        "layout": 'unified' | 'split';
+        "layout": CodeDiffLayout;
         "lineWrapping": boolean;
         "newHeading"?: string;
         "newValue": string | object;
@@ -1999,7 +2002,7 @@ export namespace JSX {
     export interface LimelCodeDiff {
         "contextLines"?: number;
         "language"?: string;
-        "layout"?: 'unified' | 'split';
+        "layout"?: CodeDiffLayout;
         "lineWrapping"?: boolean;
         "newHeading"?: string;
         "newValue"?: string | object;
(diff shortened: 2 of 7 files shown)


src/components/code-diff/code-diff.tsx
```diff
@@ -27,6 +28,7 @@ import {
  * and collapsible unchanged context sections.
  *
  * @beta
+ * @exampleComponent limel-example-code-diff-interactive
```
adrianschmidt: Putting this on top makes perfect sense while it's being reviewed, but I think we should put it further down before we merge. At least the basic example should come first, so the reader gets a gentle introduction and understands what what component `limel-code-diff` actually produces (not obvious in the interactive example, since the first thing you see is the two code editors).

viktorSoftDev: Yeah, that makes a lot of sense

john-traas (approved):

src/components/code-diff/code-diff.tsx
viktorSoftDev: Changed it around to the bottom 💪

adrianschmidt (approved):
## 6-Agent Review — READY TO MERGE ✅

Purely additive: a new interactive playground example for `limel-code-diff`, plus a non-breaking extraction of the `layout` union into a named, public `CodeDiffLayout` type. Strictly better than `main` — no regressions in any of the seven dimensions, no `[High]`/`[Critical]` findings. Only a couple of optional `[Low]` notes.

### Verdicts

| Dimension | Verdict |
|---|---|
| Code Quality | **GOOD** |
| Architecture | **SAFE** |
| Security | **SAFE** |
| Observability | **SAFE** |
| Performance | **GOOD** |

### What was checked

- **Backward compatibility** — `CodeDiffLayout` resolves to the byte-identical union, so `layout: 'unified' | 'split'` and `layout: CodeDiffLayout` are interchangeable for consumers. No duplicate-export collision, additive-only `.api.md` delta, unique example tag.
- **Code quality** — Prop/event APIs verified against the real components (`oldValue`/`newValue` are real props; `limel-code-editor` emits `change: EventEmitter<string>`). Extraction is complete — no inline unions remain anywhere.
- **Architecture** — `code-diff.types.ts` matches the established `*.types.ts` pattern and correctly stays distinct from the pre-existing *internal* `code-diff/types.ts`. Idiomatic one-way data flow, composition-only coupling.
- **Security** — Traced the only user-input path (typed values → `limel-code-diff`): no HTML-injection sinks, user text renders as escaped JSX text nodes, search term is regex-escaped, no secrets in the sample data.
- **Observability** — The handlers' `event.stopPropagation()` is the correct Stencil pattern (mirrors the sibling example); nothing is swallowed.
- **Performance** — Live re-diff per keystroke is the pre-existing, intended reactive contract of `limel-code-diff` on tiny payloads; the PR adds no new code paths into the core component. Handlers are class-field-bound, constants are module-level.
- **Lime platform (lime-issue-detector catalog)** — No LPID findings; state handling is correct, no I/O / timers / XSS sinks.

### Optional `[Low]` notes (non-blocking)

1. **Stability tag** — `CodeDiffLayout` is `@public` while `limel-code-diff` itself is `@beta` (sibling `chart.types` uses `@beta`). This is the PR's stated intent — just flagging that it commits the type to the stable contract ahead of the component.
2. **Example duplication** — `LAYOUT_OPTIONS` + handlers are duplicated across the two code-diff examples, which is consistent with the repo's "self-contained example" convention. Leave as-is, or optionally name the default layout instead of `LAYOUT_OPTIONS[1]`.

Approving. 🚢

<sub>Generated with a 6-agent review (backward compatibility, code quality, architecture, security, observability, performance, Lime platform).</sub>

PR #4136, opened 2026-06-17
fix(markdown): keep id/name attributes unprefixed in sanitizer

### LINKED ISSUE #4135: limel-markdown does not render limel-icon because the `name` attribute is rewritten by the sanitizer
## Description

`limel-icon` is whitelisted by default in `limel-markdown`, so it should be possible to render an icon directly from markdown content, e.g.:

```html
<limel-icon size="x-small" name="globe"></limel-icon>
```

However, the icon does not render. The `name` attribute — which `limel-icon` uses to select which icon to show — is silently rewritten by the HTML sanitizer.

## Root cause

`limel-markdown` uses `rehype-sanitize`, which by default guards against [DOM clobbering](https://github.com/syntax-tree/hast-util-sanitize#clobber) by prefixing the clobber-prone attributes (`id` and `name`) with `user-content-`.

As a result, `name="globe"` becomes `name="user-content-globe"`, and `limel-icon` then tries to look up an icon called `user-content-globe`, which does not exist.

## Expected behavior

A whitelisted `limel-icon` written in markdown should render with the `name` it was given.

## Notes

The same prefixing also affects `id`/`name` attributes more broadly, and notably **breaks GFM footnote navigation** (the footnote ids end up double-prefixed as `user-content-user-content-fn-1` while the links still point at `#user-content-fn-1`). Fixing this issue therefore also repairs footnotes as a positive side effect.

TommyLindh2:
## Description

`limel-markdown` uses `rehype-sanitize`, which by default guards against [DOM clobbering](https://github.com/syntax-tree/hast-util-sanitize#clobber) by prefixing the clobber-prone attributes (`id`, `name`, `aria-describedby`, `aria-labelledby`) with `user-content-`.

This broke whitelisted `limel-icon` elements written directly in markdown: `name="globe"` was rewritten to `name="user-content-globe"`, so `limel-icon` looked up an icon that does not exist and nothing rendered.

Setting `clobberPrefix: ''` disables this prefixing, so authored `id`/`name` attributes are preserved as-is.

Closes #4135

## Footnotes — a related improvement

GFM footnote ids and links are already namespaced with `user-content-` by `remark-rehype`. The sanitizer's own prefix was being applied *on top* of that, but only to the clobber attributes (not to hrefs), so the two desynced:

```
Before:  <li id="user-content-user-content-fn-1">   ← link points at #user-content-fn-1 (no match)
After:   <li id="user-content-fn-1">                ← matches the link
```

This change makes the footnote ids consistent with the links (and aria attributes) that reference them.

> [!NOTE]
> This only makes the ids/links **consistent** — it does not, by itself, make footnotes scroll. `limel-markdown` renders inside a shadow root, and native URL-fragment navigation does not resolve targets inside shadow DOM. Actually scrolling to a footnote requires a click handler that scrolls within the shadow root, tracked in #4137. The consistency fixed here is a prerequisite for that handler to find its target.

## Changes

- `markdown-parser.ts`: set `clobberPrefix: ''` on `rehype-sanitize`, with a comment explaining why.
- `markdown-parser.spec.ts`: tests for unprefixed `id`, in-sync `name`/fragment links, and footnote ids not being double-prefixed.
- New example `limel-example-markdown-built-in-component` demonstrating `limel-icon` rendered from markdown via the default whitelist.

## How to test

- Run the spec: `npm run test:spec -- markdown-parser`
- View the new "Built-in whitelisted component" example and confirm the globe icon renders.
- Inspect the footnotes example DOM and confirm the footnote `id` matches the reference `href` (no `user-content-user-content-` double prefix).

## Follow-up

- #4137 — make in-page anchor links (footnotes) actually scroll inside the shadow root.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed markdown output to properly handle footnote IDs and custom element attribute references, ensuring correct synchronization between generated anchors and their targets.

* **Documentation**
  * Added example component demonstrating markdown support for built-in elements.

### DIFF
--- a/src/components/markdown/markdown-parser.ts
+++ b/src/components/markdown/markdown-parser.ts
@@ -43,9 +43,7 @@ export async function markdownToHTML(
         .use(remarkRehype, { allowDangerousHtml: true })
         .use(rehypeRaw)
         .use(createLinksPlugin())
-        .use(rehypeSanitize, {
-            ...getWhiteList(options?.whitelist ?? []),
-        })
+        .use(rehypeSanitize, getWhiteList(options?.whitelist ?? []))
         .use(() => {
             return (tree: Node) => {
                 // Run the sanitizeStyle function on all elements, to sanitize
@@ -74,9 +72,7 @@ export async function sanitizeHTML(
 ): Promise<string> {
     const file = await unified()
         .use(rehypeParse)
-        .use(rehypeSanitize, {
-            ...getWhiteList(whitelist ?? []),
-        })
+        .use(rehypeSanitize, getWhiteList(whitelist ?? []))
         .use(() => {
             return (tree: Node) => {
                 // Run the sanitizeStyle function on all elements, to sanitize
@@ -99,6 +95,32 @@ function getWhiteList(allowedComponents: CustomElementDefinition[]): Schema {
 
     const whitelist: Schema = {
         ...defaultSchema,
+        // Disable rehype-sanitize's default `user-content-` prefix, which it
+        // otherwise prepends to the DOM-clobber attributes (`id`, `name`,
+        // `aria-describedby`, `aria-labelledby`).
+        //
+        // Without this, whitelisted custom elements break: e.g.
+        // `<limel-icon name="globe">` becomes `name="user-content-globe"` and
+        // the icon no longer resolves. It also keeps GFM footnote ids
+        // consistent with the links that reference them — remark-rehype already
+        // namespaces those with `user-content-`, and a second prefix here would
+        // desync the id (`user-content-user-content-fn-1`) from its link
+        // (`#user-content-fn-1`).
+        //
+        // Set here, in the shared schema, so that both `markdownToHTML` and
+        // `sanitizeHTML` behave identically — the same content must not be
+        // prefixed on one path and left unprefixed on the other.
+        //
+        // Safe to disable here because every current consumer renders this
+        // output inside a shadow root (`limel-markdown` and `limel-text-editor`
+        // are both `shadow: true`). Unprefixed `id`/`name` attributes inside a
+        // shadow root cannot clobber document-level globals
+        // (`document.getElementById`, `window.<name>`, `document.forms`), which
+        // is what the prefix guards against. If a future consumer renders this
+        // output into the light DOM — or these functions become public exports
+        // — reinstate the prefix or scope its removal to whitelisted custom
+        // elements only.
+        clobberPrefix: '',
         strip: [...(defaultSchema.strip ?? []), 'style'],
         tagNames: [
             ...(defaultSchema.tagNames || []),
--- a/src/components/markdown/examples/markdown-built-in-component.tsx
+++ b/src/components/markdown/examples/markdown-built-in-component.tsx
@@ -1,0 +1,27 @@
+import { Component, h } from '@stencil/core';
+
+const markdown = `
+## Built-in whitelisted components
+
+A set of lime-elements components is whitelisted by default, so they can be
+used directly in markdown content without configuring the \`whitelist\` prop.
+
+For example, \`limel-icon\` is whitelisted, so you can render an icon inline: <limel-icon size="small" name="globe"></limel-icon>
+`;
+
+/**
+ * Built-in whitelisted component
+ *
+ * Some lime-elements components, such as `limel-icon`, are whitelisted by
+ * default. They can be used directly in markdown content without specifying
+ * them in the `whitelist` prop.
+ */
+@Component({
+    tag: 'limel-example-markdown-built-in-component',
+    shadow: true,
+})
+export class MarkdownBuiltInComponentExample {
+    public render() {
+        return <limel-markdown value={markdown} />;
+    }
+}
--- a/src/components/markdown/markdown-parser.spec.ts
+++ b/src/components/markdown/markdown-parser.spec.ts
@@ -1,4 +1,5 @@
 import { markdownToHTML, sanitizeHTML } from './markdown-parser';
+import { DEFAULT_MARKDOWN_WHITELIST } from './default-whitelist';
 
 /**
  * Normalize HTML for comparison: collapse whitespace between tags,
@@ -431,6 +432,83 @@ describe('markdownToHTML', () => {
             expect(result).toEqualHtml(
                 '<p><my-component allowed="ok">Content</my-component></p>'
             );
+        });
+    });
+
+    describe('clobberPrefix (DOM-clobbering prefix)', () => {
+        it('should keep name attributes on whitelisted custom elements unprefixed', async () => {
+            // Regression test for #4135. `limel-icon` uses the `name`
+            // attribute to select which icon to render. `name` is a
+            // DOM-clobber attribute, so the sanitizer's default prefix
+            // rewrote `name="globe"` to `name="user-content-globe"`, and the
+            // icon then looked up an icon that does not exist.
+            const result = await markdownToHTML(
+                '<limel-icon size="small" name="globe"></limel-icon>',
+                {
+                    whitelist: [
+                        { tagName: 'limel-icon', attributes: ['name', 'size'] },
+                    ],
+                }
+            );
+
+            expect(result).not.toContain('user-content-');
+            expect(result).toEqualHtml(
+                '<p><limel-icon size="small" name="globe"></limel-icon></p>'
+            );
+        });
+
+        it('should render limel-icon from the default whitelist with its name intact', async () => {
+            // Complements the test above by exercising the real
+            // DEFAULT_MARKDOWN_WHITELIST rather than an inline whitelist, so
+            // that dropping `name` from `limel-icon` in default-whitelist.ts
+            // would re-break #4135 and be caught here.
+            const result = await markdownToHTML(
+                '<limel-icon size="small" name="globe"></limel-icon>',
+                { whitelist: DEFAULT_MARKDOWN_WHITELIST }
+            );
+
+            expect(result).not.toContain('user-content-');
+            expect(result).toEqualHtml(
+                '<p><limel-icon size="small" name="globe"></limel-icon></p>'
+            );
+        });
+
+        it('should keep id attributes unprefixed', async () => {
+            // By default rehype-sanitize prepends `user-content-` to `id`
+            // attributes to guard against DOM clobbering. We disable that so
+            // ids stay as authored.
+            const result = await markdownToHTML('<div id="my-section">x</div>');
+            expect(result).toEqualHtml('<div id="my-section">x</div>');
+        });
+
+        it('should keep name/fragment links in sync', async () => {
+            // The `name` anchor and the `#fragment` link must keep matching
+            // values, otherwise in-page navigation breaks.
+            const result = await markdownToHTML(
+                '<a name="anchor">target</a> <a href="#anchor">go</a>'
+            );
+            expect(result).toEqualHtml(
+                '<p><a name="anchor">target</a> <a href="#anchor">go</a></p>'
+            );
+        });
+
+        it('should not double-prefix footnote ids and references', async () => {
+            // remark-rehype already namespaces footnote ids/links with
+            // `user-content-`. Without disabling the sanitizer's own prefix,
+            // the id would become `user-content-user-content-fn-1` and no
+            // longer match the `#user-content-fn-1` link.
+            const result = await markdownToHTML(
+                'Here is a footnote[^1].\n\n[^1]: My reference.'
+            );
+
+            expect(result).not.toContain('user-content-user-content-');
+
+            // The reference link must be an anchor pointing at the footnote
+            // item, and the footnote item must be an `<li>` carrying the
+            // matching id — asserting the values sit on the expected nodes,
+            // not merely that the strings appear somewhere in the output.
+            expect(result).toMatch(/<a[^>]*\bhref="#user-content-fn-1"[^>]*>/);
+            expect(result).toMatch(/<li[^>]*\bid="user-content-fn-1"[^>]*>/);
         });
     });
 
--- a/src/components/markdown/markdown.tsx
+++ b/src/components/markdown/markdown.tsx
@@ -34,6 +34,7 @@ import { adaptColorContrast } from '../../util/adapt-color-contrast';
  * @exampleComponent limel-example-markdown-keys
  * @exampleComponent limel-example-markdown-blockquotes
  * @exampleComponent limel-example-markdown-horizontal-rule
+ * @exampleComponent limel-example-markdown-built-in-component
  * @exampleComponent limel-example-markdown-custom-component
  * @exampleComponent limel-example-markdown-custom-component-with-json-props
  * @exampleComponent limel-example-markdown-remove-empty-paragraphs


src/components/markdown/markdown-parser.ts
```diff
@@ -45,6 +45,20 @@ export async function markdownToHTML(
...
+            // Without this, whitelisted custom elements break: e.g.
+            // `<limel-icon name="globe">` becomes `name="user-content-globe"`
+            // and the icon no longer resolves.
+            //
+            // It also keeps GFM footnote ids consistent with the links that
+            // reference them — remark-rehype already namespaces those with
+            // `user-content-`, and a second prefix here would desync the id
+            // (`user-content-user-content-fn-1`) from its link
+            // (`#user-content-fn-1`).
+            clobberPrefix: '',
```
adrianschmidt: I want to look into what the effects of simply disabling this across all input is before we merge this. There's a reason the default doesn't allow passing in these values unaltered, so we might want to look into allowing it only on certain components or something. I'll have a chat with Claude, but I'm requesting changes now just so I don't have to feel stressed that someone will over-eagerly approve and merge this before I've had a chance.

adrianschmidt (dismissed):

adrianschmidt:
# Consolidated PR Review

_Multi-agent review across 7 dimensions (backward compatibility, code quality, architecture, security, observability, performance, Lime-platform issues)._

## PR Summary

This PR sets `clobberPrefix: ''` on the `rehype-sanitize` call in `markdownToHTML` (`src/components/markdown/markdown-parser.ts`), disabling the `user-content-` prefix that the sanitizer otherwise prepends to the four DOM-clobber attributes (`id`, `name`, `aria-describedby`, `aria-labelledby`). This fixes two real, pre-existing bugs: whitelisted `<limel-icon name="globe">` rendered from markdown (the `name` was being mangled to `user-content-globe`) and GFM footnote navigation (footnote ids were double-prefixed as `user-content-user-content-fn-1`, desynced from their links). It adds 3 spec tests, a new example component, and a JSDoc registration. +79/-0 across 4 files.

The PR is strictly better than `main` for the bug it targets — `limel-icon` and footnote navigation are both broken on `main` today and this fixes both. It introduces **no new exploitable vulnerability**: the sanitizer still strips scripts / event-handlers / dangerous URLs / forms / iframes, and `limel-markdown` renders inside a shadow root (`shadow: true`), which contains the document-global vectors (`document.getElementById`, `window.<name>`, `document.forms`) that make DOM clobbering dangerous. So on regression grounds it is not blocked.

The caveats below are non-blocking follow-ups — most notably that the change disables a defense-in-depth control at a **broader granularity than the bug strictly requires**.

- **Blockers:** None. No regression introduced — the two bugs fixed both pre-existed on `main`, no new XSS/injection vector is opened, and shadow-DOM encapsulation contains the DOM-clobbering surface that the removed prefix protected.
- **Non-blocking but worth addressing:** Three `[Medium]` items (global-vs-scoped granularity, missing regression test for the actual headline bug, `sanitizeHTML` left inconsistent) — see Top Recommendations.

---

## 1. Backward Compatibility — SAFE ✅

The change relaxes attribute rewriting on markdown *output* only; it cannot break consumers that don't depend on the prefix being present.

- **No significant issues found.** A full repo grep shows the only `user-content` references live in `markdown-parser.ts`/spec and the separate `email-viewer/sanitize-email-html.ts`. No SCSS selector, no `getElementById`/`querySelector`, and no anchor/scroll handler targets a prefixed id. The behavioral change makes output *more* compatible (anchor links now resolve; footnote ids match their links).
- Established in-repo precedent: `email-viewer/sanitize-email-html.ts` already sets `clobberPrefix: ''` for the same anchor-link reason, so this aligns markdown with an existing decision.
- **Positives:** Tightly scoped, follows existing precedent, ships focused regression tests, and no consumer in the repo depends on the removed prefix.

## 2. Code Quality — GOOD ✅

Clean, well-commented diff; the main gaps are around test targeting and an un-mirrored sibling function.

- **[Medium]** No test pins the actual headline bug. Issue #4135 is about a whitelisted custom element (`<limel-icon name="globe">`) keeping its `name`; the new tests (`markdown-parser.spec.ts:437-471`) cover plain `<div id>`, plain `<a name>`, and footnotes — but not a whitelisted custom element retaining `name=`/`size=` unprefixed. A future whitelist/sanitizer change could re-break `limel-icon` with all tests green. A ~4-line test (`markdownToHTML('<limel-icon name="globe"></limel-icon>', {whitelist: [...]})` asserting `name="globe"` and no `user-content-`) would directly guard the fix.
- **[Medium]** Fix not applied to the sibling parser `sanitizeHTML`. `clobberPrefix: ''` is added only to `markdownToHTML`; `sanitizeHTML` (`markdown-parser.ts:85-105`) keeps the default. The text-editor uses *both* (`markdown-converter.ts` for display, `html-converter.ts` for HTML conversion), so the same content is prefixed on one path and not the other. Either the divergence is intentional and should be documented, or the fix should be consistent.
- **[Low]** The footnote test (`spec.ts:466-470`) uses loose independent `toContain` assertions rather than asserting the id and link are on the expected nodes; the `not.toContain('user-content-user-content-')` guard covers the main regression.
- **Positives:** The explanatory comment block (`markdown-parser.ts:48-60`) clearly states what/why/failure-mode; the example component matches existing conventions exactly and `limel-icon` is genuinely in `default-whitelist.ts`.

## 3. Architecture — ACCEPTABLE ✅

The fix is correct but uses a blunter lever than the problem strictly requires — which is the granularity question raised in the review thread.

- **[Medium]** Global `clobberPrefix: ''` is broader than the two bugs need. Verified against the libraries: `hast-util-sanitize` exposes two levers — `clobber: ['ariaDescribedBy','ariaLabelledBy','id','name']` (which attributes) and `clobberPrefix` (the prefix string). Setting the prefix to `''` disables prefixing for *all four* attributes on *all* elements, including raw author `<div id>`/`<a name>`. Narrower alternatives exist:
  - Dropping only `name` from the `clobber` array would fix `limel-icon` but **not** footnotes — footnote ids are emitted as `id="user-content-fn-1"` (via `mdast-util-to-hast`), so they're re-prefixed through the `id` clobber, not `name`.
  - **Aligning the prefixes** is the cleaner targeted fix for footnotes: the desync is a *double-prefix collision* (`remark-rehype`/`mdast-util-to-hast` already applies `user-content-`, then the sanitizer applies it again). Letting a single layer own the prefix fixes footnotes without dropping clobber protection at all.
  - Net: the global disable is defensible and simplest, but it is a strictly larger blast radius than the named bugs require.
- **[Low]** `markdownToHTML`/`sanitizeHTML` asymmetry (see Code Quality `[Medium]`); the clobber decision lives as an inline option rather than a named/shared one — consider hoisting it into `getWhiteList`'s schema so the divergence is intentional and visible.
- No internal data shape leaks through the public `MarkdownToHTMLOptions` surface; the parser stays a pure transform and the whitelist merge stays in the orchestrating component. Not an exposure defect.
- **Positives:** Minimal, surgically-scoped, no public-API change; fixes two genuine pre-existing bugs; clean parser/component separation preserved.

## 4. Security — FAIR ⚠️

Removing `clobberPrefix` disables a deliberate DOM-clobbering defense, but shadow-DOM encapsulation contains the dangerous vectors — so this is a contained reduction of a defense-in-depth control, **not a new high-severity vulnerability**.

- **[Medium]** Defense-in-depth control removed on a path that renders untrusted content. `limel-markdown` is fed user-generated content across the platform (history/activity notes, chat messages, feed/notification bodies, table-cell record values). The `user-content-` prefix is the upstream mitigation against attacker-controlled `id`/`name` shadowing JS/DOM properties. **Why this is `[Medium]` and not `[High]`:** `limel-markdown` is `shadow: true`, so authored `id`/`name` inside the shadow root do **not** populate `document.getElementById`, `window.<name>`, or `document.forms` — the classic exploitable clobbering primitives are contained; and the sanitizer still strips scripts, event handlers, `javascript:`/`data:` URLs, forms, and iframes (only the prefix changed, so no new XSS path is opened). Residual, contained risks: within the shadow root, attacker `id`/`aria-*` can collide with the component's own ids or retarget aria references — but since the attacker authors the whole markdown blob, this is self-inflicted on their own content and grants no privilege crossing.
- **[Low]** No new injection vector beyond clobbering — confirmed only the prefix option changed; script/URL/handler stripping and the whitelist schema are untouched.
- **Recommendation:** Prefer the narrower fix (resolve the `limel-icon name` collision via the whitelist/hydration layer, or align footnote prefixes per Architecture) rather than disabling the prefix platform-wide. If the global disable is accepted, document shadow-DOM containment as the explicit justification and apply it consistently to `sanitizeHTML`.
- **Positives:** Prefix-only change with sanitizer stripping fully preserved; untrusted email bodies use a separate sanitizer that retains the prefix; reasoning documented inline; new tests lock in behavior.

## 5. Observability — GOOD ✅

- **No significant issues found.** A declarative config option in a pure transform; no new code path, error handling, or silent-failure surface. If anything it *removes* a class of confusing silent attribute rewrites (the "why did my icon name change?" surprise). This is a frontend component library, so service-style logging/metrics expectations don't apply.
- **Positives:** The inline comment documenting *why* the prefix is disabled directly aids future debuggability.

## 6. Performance — SAFE ✅

- **No significant issues found.** Setting an existing config option to `''` is performance-neutral to marginally cheaper (no per-attribute prefix concatenation); no new imports, traversals, or allocations. The example component is static and bundles into docs only — zero shipped-library bundle impact.
- **Positives:** Surgical change with no new dependencies and the consumer's whitelist memoization preserved.

## 7. Lime Platform Issues — SAFE ✅

- **No applicable LPID findings.** This is a frontend TS/Stencil PR with no Python/UoW/endpoint/handler/task/file-I/O surface, so the backend rule families do not apply. LPID-FE-10 (Frontend XSS Sink) was evaluated and does **not** match — `clobberPrefix` does not weaken XSS protection (the sanitizer still strips `<script>`/event handlers); DOM clobbering is a distinct class the catalog has no rule for. The DOM-clobbering defense-in-depth reduction is captured in Section 4.
- **Positives:** No new Lime-platform runtime hazard; no anti-pattern in the example or tests.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Code Quality | **GOOD** |
| Architecture | **ACCEPTABLE** |
| Security | **FAIR** |
| Observability | **GOOD** |
| Performance | **SAFE** |
| **Merge Readiness** | **MERGE WITH CAVEATS** |

## Top Recommendations

1. **[Medium from Architecture]** Decide the granularity deliberately rather than defaulting to a global disable. The narrower fix: **align the footnote prefixes** (let a single layer own `user-content-`) to fix footnotes *with* clobber protection intact, and resolve the `limel-icon name` collision via the whitelist/hydration layer. If the team consciously accepts the global disable, record shadow-DOM containment as the written justification.
2. **[Medium from Security]** Same root recommendation from the security angle: prefer scoping the relaxation over disabling the DOM-clobbering mitigation platform-wide on a path that renders untrusted content; shadow-DOM containment makes it acceptable but not free.
3. **[Medium from Code Quality]** Add a regression test for the actual headline bug — a whitelisted `<limel-icon name="globe">` rendered from markdown keeps `name="globe"` (no `user-content-`).
4. **[Medium from Code Quality]** Resolve the `markdownToHTML`/`sanitizeHTML` inconsistency: either apply the same `clobberPrefix` decision to `sanitizeHTML` or document why the two intentionally diverge.
5. **[Low from Code Quality]** Tighten the footnote test to assert the id and reference link sit on the expected nodes, not just that both strings appear somewhere in the output.
6. **[Low from Architecture]** Consider hoisting the clobber decision into the shared `getWhiteList` schema so it's a named, visible choice rather than an inline option on one of two call sites.

adrianschmidt:
## Follow-up: is it worth scoping the fix to only the whitelisted custom elements?

A natural alternative to the global `clobberPrefix: ''` is to keep the sanitizer's default prefix and strip it back off **only** on whitelisted custom elements (`limel-icon` etc.), preserving DOM-clobbering protection for ordinary author-written HTML. Having looked into it: **given that `limel-markdown` renders inside a shadow root, scoping isn't worth the extra complexity.** Here's the accounting.

**The exploitable residual after shadow DOM + the sanitizer is essentially nil:**

- The genuinely dangerous clobber carriers — `<iframe name>`, `<form>`/`<form target>`, `<object>`, `<embed>` — are **stripped by the sanitizer regardless** of the prefix. The prefix only ever guarded `id`/`name`/`aria-*` on surviving elements.
- Those surviving `id`/`name` values live **inside the shadow root**, so they never populate `document.getElementById`, `window.<name>`, or `document.forms` — the vectors that make clobbering exploitable. (Neither `limel-markdown` nor `limel-text-editor` does any `getElementById`/named-property lookup that could be clobbered.)
- The only residuals worth considering — same-root `aria`/`id` collisions — are **self-inflicted within a single untrusted blob**: the attacker authors all the text *and* the aria reference, so retargeting a label or duplicating an id gains them nothing they couldn't already do by writing the text directly. No privilege crossing.
- `name` on `<a>`/`<div>`/custom elements is the old fragment-anchor name, **not** a browsing-context name; the elements that actually create cross-boundary contexts are the stripped ones.

So the scoped plugin would defend a near-vacuous residual — at a permanent cost of ~20 lines of custom hast manipulation (including a subtle `user-content-user-content-` collapse to keep footnotes working), tests, mirroring into `sanitizeHTML`, and a per-render traversal. And the global flag is *also* the simpler footnote fix; a scoped-to-custom-elements version would re-break footnotes (they're plain `<li>`/`<a>`, not whitelisted). Complexity up, both bugs fixed less cleanly, security delta ≈ zero.

**What actually carries the safety, and the one real risk:** `markdownToHTML`/`sanitizeHTML` are **internal** (not package exports), and both consumers (`limel-markdown`, `limel-text-editor`) are `shadow: true`. So the "output always lands in a shadow root" invariant holds for every current caller. The only way the global disable becomes a real hole is a *future* change that adds a light-DOM consumer or makes these functions public — a maintainer footgun, not an attacker, and guardable far more cheaply than a plugin.

**Conclusion — keep `clobberPrefix: ''`, and instead do two cheap things:**

1. **Record the invariant as a comment** where the prefix is disabled: this is safe *because* output is always rendered in a shadow root; if a consumer ever renders to light DOM (or these become public), revisit. That converts the implicit cross-file dependency (safety lives in `markdown.tsx`, the relaxation in `markdown-parser.ts`) into an explicit guard against the one risk that's actually real here.
2. **Resolve the `markdownToHTML`/`sanitizeHTML` asymmetry** — the text-editor uses *both*, so right now the same content is mangled on one path and not the other. Apply the same `clobberPrefix` decision to `sanitizeHTML`.

I'll push these two as fixups.

adrianschmidt:
Pushed three fixups addressing the follow-ups from the review above. None change the headline approach (`clobberPrefix: ''` stays) — they make it consistent, documented, and tested.

**1. Consistency — `markdownToHTML` and `sanitizeHTML` now behave identically** (Top Rec #4)
Moved `clobberPrefix: ''` out of the inline `markdownToHTML` call and into the shared `getWhiteList` schema, so both sanitizer entry points pick it up. The text-editor uses *both* paths (`markdown-converter` for display, `html-converter` for HTML conversion), so previously the same content could be prefixed on one path and not the other. They can no longer drift.

**2. Documented why disabling the prefix is safe** (Top Rec #1)
Added a comment recording the invariant the safety rests on: every current consumer (`limel-markdown`, `limel-text-editor`) renders this output **inside a shadow root**, so unprefixed `id`/`name` can't clobber document-level globals (`document.getElementById`, `window.<name>`, `document.forms`) — which is what the prefix guards against. The comment also flags what to do if that ever stops being true (a light-DOM consumer, or these functions becoming public exports): reinstate the prefix or scope its removal to whitelisted custom elements.

> On scoping specifically: I looked into restricting the un-prefixing to whitelisted custom elements only, but given the shadow-root containment the exploitable residual is essentially nil, while the scoped version adds ~20 lines of custom tree-walking and would re-break footnotes. Details in the follow-up comment above. The documented invariant is the cheaper, equivalent safeguard.

**3. Regression test for the actual bug** (Top Rec #3)
The original tests covered `id`, `name`/fragment sync, and footnotes, but not the headline case from #4135 — a whitelisted custom element keeping its `name`. Added a test that renders `<limel-icon name="globe">` and asserts the output keeps `name="globe"` with no `user-content-` prefix. Without the fix this fails on `name="user-content-globe"`.

Spec is green (67 tests). The remaining `[Low]` items from the review (tightening the footnote assertions) I've left as-is — happy to fold them in if you'd like.

adrianschmidt:
# Consolidated PR Review

_Re-review across 7 dimensions (backward compatibility, code quality, architecture, security, observability, performance, Lime-platform issues) after the three fixups._

## PR Summary

The PR sets `clobberPrefix: ''` on the shared `rehype-sanitize` schema (`getWhiteList` in `markdown-parser.ts`), disabling the `user-content-` prefix the sanitizer otherwise prepends to the four DOM-clobber attributes (`id`, `name`, `aria-describedby`, `aria-labelledby`). This fixes two pre-existing bugs: whitelisted `<limel-icon name="globe">` rendered from markdown (the `name` was mangled to `user-content-globe`) and GFM footnote navigation (footnote ids were double-prefixed to `user-content-user-content-fn-1`, desynced from their links). +113/−3 across 4 files. Closes #4135.

The PR is strictly better than `main` — both bugs are broken on `main` today and both are fixed here — and it introduces no regression: the sanitizer still strips scripts, event handlers, `javascript:`/`data:` URLs, forms, iframes, objects, and embeds (verified against the existing security tests), and both consumers (`limel-markdown`, `limel-text-editor`) render inside a shadow root, which contains the document-global vectors DOM-clobbering protection guards against. The three `[Medium]` items from the prior review have each been discharged by the fixups.

- **Blockers:** None. No regression: the two bugs fixed both pre-existed on `main`, no new XSS/injection path is opened (only the `clobberPrefix` option changed, not the stripping rules), and shadow-DOM encapsulation contains the DOM-clobbering surface.
- **Prior caveats, now resolved:**
  1. *sanitizeHTML inconsistency* → fixed by hoisting `clobberPrefix: ''` into the shared `getWhiteList` schema so both entry points behave identically.
  2. *Missing regression test for the headline bug* → fixed by the `<limel-icon name="globe">` test, which fails on `main` (`name="user-content-globe"`) and passes here.
  3. *Global-vs-scoped granularity needing a deliberate decision* → the decision was made, justified publicly, and documented inline (shadow-root containment invariant + explicit reversal trigger).
- **Non-blocking but worth addressing:** three `[Low]` polish items — see Top Recommendations.

---

## 1. Backward Compatibility — SAFE ✅

The change relaxes attribute rewriting on markdown *output* only; nothing in the repo depends on the removed prefix.

- **[Low]** Cosmetic call-site asymmetry: `markdownToHTML` now passes `getWhiteList(...)` directly (`markdown-parser.ts:46`) while `sanitizeHTML` still wraps it in a redundant spread `{ ...getWhiteList(...) }` (`markdown-parser.ts:75-77`). Functionally identical — both pick up `clobberPrefix` from the schema — so the consistency intent is met; only tidiness.
- Verified: neither `markdownToHTML` nor `sanitizeHTML` is re-exported from `src/index.ts` or present in `etc/lime-elements.api.md`; a repo grep for `user-content-` matches only the new comment and the pre-existing `email-viewer` sanitizer. No props/events/types removed; the change makes output *more* compatible (anchor/footnote links now resolve).
- **Positives:** Tightly scoped to two internal functions, aligns markdown with the already-shipped `email-viewer` precedent, and documents the conditions under which the prefix must be reinstated.

## 2. Code Quality — GOOD ✅

The three fixups genuinely resolve the prior `[Medium]`s; remaining items are `[Low]` polish.

- **[Low]** Same cosmetic spread asymmetry as above (`sanitizeHTML` keeps `{ ...getWhiteList(...) }`); dropping the spread would make the two sibling functions read identically.
- **[Low]** The footnote test (`markdown-parser.spec.ts`) uses three independent `toContain` assertions rather than asserting the id and href sit on the expected nodes; it pins the double-prefix regression but a structural (`toEqualHtml`) assertion would be more robust. The author explicitly chose to leave this as-is and offered to fold it in.
- **[Low]** The `limel-icon` regression test passes an explicit `whitelist` rather than exercising the real `DEFAULT_MARKDOWN_WHITELIST`. Correctly isolates the clobber behavior under test, but doesn't also pin that `limel-icon`/`name` ships in the default whitelist (it does — `default-whitelist.ts:32-34`).
- **Positives:** Fix centralized in the single shared schema builder so both paths stay in lockstep; the headline-bug regression test reproduces #4135 exactly and fails pre-fix; the explanatory comment block states what/why/failure-mode/reversal-trigger clearly.

## 3. Architecture — GOOD ✅

The fix is correctly centralized and the global-vs-scoped decision holds up under scrutiny.

- **[Low]** Schema-construction duplication between `markdown-parser.ts` and `email-viewer/sanitize-email-html.ts` — both independently set `clobberPrefix: ''` and re-derive from `defaultSchema`. Pre-existing divergence (the email sanitizer was deliberately forked for fidelity), not introduced here; worth only a tracking note if a third sanitizer appears.
- **[Low]** The cosmetic spread asymmetry (cross-ref Code Quality) slightly undercuts the "both paths behave identically" comment, though they do.
- On the granularity question: the global disable is architecturally correct. The dangerous clobber *carriers* (`iframe`/`form`/`object`/`embed`) are absent from `tagNames` and stripped regardless; scoping to whitelisted custom elements would re-break footnotes (plain `<li>`/`<a>`) for no safety gain inside a shadow root; and consolidating `user-content-` ownership onto remark-rehype is the correct ownership boundary. No internal shape leaks through the public `MarkdownToHTMLOptions` surface.
- **Positives:** Eliminates a real path-asymmetry defect, documents the containment invariant precisely (condition + reversal trigger), and consolidates prefix ownership onto a single layer.

## 4. Security — ACCEPTABLE ✅

Removing `clobberPrefix` reduces a defense-in-depth control, but shadow-DOM encapsulation contains the dangerous vectors and all stripping rules are preserved — no new exploitable path vs `main`.

- **[Low]** The change disables DOM-clobbering prefixing for `id`/`name`/`aria-*` on both `markdownToHTML` and `sanitizeHTML`. The prior review rated this `[Medium]` and recommended: accept the global disable as a conscious decision, document shadow-DOM containment as the justification, and apply it consistently. **All three have been done** — so this is now an accurately-described, deliberately-chosen, documented trade-off, which the documented-intent rule places at `[Low]`/informational. Substantively: both consumers are `shadow: true` (`markdown.tsx:47`; `text-editor.tsx:49`), so unprefixed `id`/`name` never populate `document.getElementById`/`window.<name>`/`document.forms`; the dangerous carriers are stripped regardless (verified by the existing tests); residual same-root aria/id collisions are self-inflicted within a single attacker-authored blob (no privilege crossing). The only real risk — a *future* light-DOM consumer or these internal functions becoming public exports — is documented inline with an explicit escape hatch, and the functions are confirmed internal.
- Verified only the `clobberPrefix` option changed; script/handler/URL/tag stripping and the whitelist schema are untouched. The separate untrusted-email path (`sanitize-email-html.ts`) is unaffected and already sets `clobberPrefix: ''` independently.
- **Positives:** Prefix-only change with stripping fully preserved and proven by existing tests; containment reasoning documented inline with a remediation trigger; the fixup unified both paths so the same content can't be prefixed on one and not the other.

## 5. Observability — GOOD ✅

- **No significant issues found.** Pure declarative config change; no logging, error-handling, or control-flow added or removed. `unified().process()` rejections still propagate through the `async` functions — no new swallowed-error or silent-failure surface. Service/backend logging expectations don't apply to a frontend component library.
- **Positives:** The inline comment documenting *why* the prefix is disabled and when to revert is durable debugging context.

## 6. Performance — SAFE ✅

- **No significant issues found.** Setting `clobberPrefix: ''` is perf-neutral-to-marginally-cheaper (skips per-attribute prefix concatenation), and the `markdownToHTML` path drops one redundant schema spread allocation. `getWhiteList` was never memoized on `main` and its call frequency is unchanged; the consumer-side whitelist memoization (`markdown.tsx`) is preserved. The example component and spec are tree-shaken out of the shipped bundle.
- **Positives:** Surgical change, no new dependencies or allocations, consumer memoization intact.

## 7. Lime Platform Issues — SAFE ✅

- **No applicable LPID findings.** Frontend TS/Stencil PR with no Python/UoW/endpoint/handler/task/file-I/O surface, so backend rule families don't apply. LPID-FE-10 (Frontend XSS Sink) was evaluated against the fresh catalog and does **not** match — `clobberPrefix` governs DOM clobbering (a distinct class with no LPID rule), not XSS; the sanitizer still strips `<script>`/handlers/dangerous URLs, and output renders into a shadow root. The defense-in-depth reduction is captured in Section 4.
- **Positives:** No new Lime-platform runtime hazard; the static example introduces no anti-pattern.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Code Quality | **GOOD** |
| Architecture | **GOOD** |
| Security | **ACCEPTABLE** |
| Observability | **GOOD** |
| Performance | **SAFE** |

## Top Recommendations

All remaining items are `[Low]` polish; none block merge.

1. **[Low from Code Quality / Backward Compatibility / Architecture]** Drop the redundant `{ ...getWhiteList(...) }` spread in `sanitizeHTML` (`markdown-parser.ts:75-77`) so both call sites read identically.
2. **[Low from Code Quality]** Tighten the footnote test to assert the id and `href` sit on the expected nodes (structural assertion) rather than three independent `toContain` checks.
3. **[Low from Code Quality]** Add a test that exercises the real `DEFAULT_MARKDOWN_WHITELIST` so a future drop of `name` from `default-whitelist.ts` would be caught.

**Bottom line:** The fixups did exactly what the prior review asked — consistency, a headline-bug regression test, and a documented deliberate decision on granularity. No regressions, no blockers; the PR is strictly better than `main`. Ready to merge.

adrianschmidt:
Pushed three fixups addressing the `[Low]` polish items from the Top Recommendations of the re-review above. All three are non-functional cleanups — the headline approach (`clobberPrefix: ''` in the shared schema) is unchanged.

**1. Drop the redundant spread in `sanitizeHTML`** (Top Rec #1)
`markdownToHTML` already passed `getWhiteList(...)` directly while `sanitizeHTML` still wrapped it in `{ ...getWhiteList(...) }`. Since `getWhiteList` returns a fresh schema object, the spread was a redundant shallow copy and the only thing left making the two sibling call sites read differently. Both now call `getWhiteList(...)` identically.

**2. Tighten the footnote test** (Top Rec #2)
Replaced the two independent `toContain` checks with structural `toMatch` assertions that pin the `href` to an `<a>` and the matching `id` to the `<li>`, so the test verifies the values sit on the expected nodes rather than merely appearing somewhere in the output. The `not.toContain('user-content-user-content-')` double-prefix guard is retained.

**3. Add a default-whitelist regression test** (Top Rec #3)
The existing `limel-icon` test passes an inline whitelist, which isolates the clobber-prefix behavior but doesn't pin that `limel-icon`/`name` actually ships in `DEFAULT_MARKDOWN_WHITELIST`. Added a complementary test that renders `<limel-icon name="globe">` using the real `DEFAULT_MARKDOWN_WHITELIST`, so dropping `name` from `default-whitelist.ts` would re-break #4135 and be caught here.

adrianschmidt (approved):
# Consolidated PR Review

_Multi-agent review across 7 dimensions (backward compatibility, code quality, architecture, security, observability, performance, Lime-platform issues). This is a **third-pass review** — the PR has already absorbed six fixups across two prior review rounds._

## PR Summary

Sets `clobberPrefix: ''` in the shared `getWhiteList` `rehype-sanitize` schema (`src/components/markdown/markdown-parser.ts:123`), disabling the `user-content-` prefix the sanitizer otherwise prepends to the four DOM-clobber attributes (`id`, `name`, `aria-describedby`, `aria-labelledby`). This fixes two pre-existing bugs: whitelisted `<limel-icon name="globe">` rendered from markdown (the `name` was mangled to `user-content-globe`) and GFM footnote navigation (footnote ids were double-prefixed to `user-content-user-content-fn-1`, desynced from their links). Adds a `clobberPrefix` spec block (5 tests), a static example component, and a JSDoc registration. +134/−6 across 4 files. Closes #4135.

The PR is strictly better than `main` — both bugs are broken on `main` today and both are fixed here — and it introduces no regression. The fix is now centralized in the single shared schema (so `markdownToHTML` and `sanitizeHTML` behave identically), the containment invariant is documented inline with a reversal trigger, and the headline bug has a regression test exercising both an inline whitelist and the real `DEFAULT_MARKDOWN_WHITELIST`. All three `[Low]` polish items from the prior round (drop redundant spread, structural footnote assertions, default-whitelist test) are also in. Every agent independently verified the security claims against the current code.

- **Blockers:** None. No regression: both fixed bugs pre-existed on `main`, only the `clobberPrefix` option changed (script/handler/URL/tag stripping and the whitelist schema are untouched and still proven by the existing security tests), and both consumers (`limel-markdown`, `limel-text-editor`) render inside a shadow root, which contains the document-global vectors DOM-clobbering protection guards against.
- **Non-blocking but worth addressing:** Two optional `[Low]` polish items — see Top Recommendations.

---

## 1. Backward Compatibility — SAFE ✅

Relaxes attribute rewriting on markdown *output* only; nothing in the repo depends on the removed prefix.

- Verified: `markdownToHTML`/`sanitizeHTML`/`getWhiteList` are not re-exported from `src/index.ts` and absent from `etc/lime-elements.api.md`; no prop/event/type signatures changed. The behavioral change makes output *more* compatible (anchor/footnote links now resolve). The dropped `{ ...getWhiteList() }` spread is behavior-preserving (`getWhiteList` returns a fresh object; `rehype-sanitize` doesn't mutate its options).
- Counterfactual checked: the inline comment's "every current consumer renders inside a shadow root" claim holds — `email-viewer/sanitize-email-html.ts` references `sanitizeHTML` only in a comment and runs its own independent pipeline, so it is genuinely unaffected.
- **Positives:** Tightly scoped to two internal functions; aligns markdown with the already-shipped `email-viewer` `clobberPrefix: ''` precedent; strong regression tests including the real default whitelist.

## 2. Code Quality — GOOD ✅

The six fixups genuinely resolved every prior finding; nothing new of substance.

- **[Low]** Pre-existing duplication of the `clobberPrefix: ''` + `defaultSchema`-re-derivation pattern between `markdown-parser.ts` and `email-viewer/sanitize-email-html.ts:59-63`. Not introduced here; consolidating is out of scope for this fix (cross-ref Architecture).
- Verified: `clobberPrefix` is a valid `Schema` property (no casts/`any`); the headline-bug regression tests are meaningful (`DEFAULT_MARKDOWN_WHITELIST` genuinely lists `name` for `limel-icon` at `default-whitelist.ts:33`, so dropping it would fail the test); the footnote test now uses structural `toMatch` assertions binding `href`/`id` to the expected `<a>`/`<li>` nodes.
- **Positives:** Single-line root-cause fix in the shared schema with both call sites consistent; thorough inline rationale documenting the security invariant and its escape hatch; example follows existing conventions exactly.

## 3. Architecture — GOOD ✅

The fix is correctly centralized and the global-vs-scoped decision holds up under scrutiny.

- **[Low]** Same pre-existing `email-viewer` schema-construction duplication noted above — a latent maintenance hazard (a future change to the clobber model would need applying in two spots), not introduced by this PR and correctly out of scope.
- Verified against the real call graph: both consumers render into a shadow root (`markdown.tsx:47` `shadow: true`; `text-editor.tsx:49` `shadow: { delegatesFocus: true }`), the functions are confirmed internal (not in `src/index.ts`/`api.md`), and the documented reversal trigger (future light-DOM consumer or public export) is the complete set of conditions that would invalidate the containment argument. Centralizing into `getWhiteList` *removes* a prior path-asymmetry defect rather than adding coupling. The global-vs-scoped decision is documented and sound (scoping would re-break footnotes and add ~20 lines of tree-walking for ≈zero safety gain inside a shadow root) — not re-raised, per the documented-intent rule.
- **Positives:** Eliminates a real path-asymmetry defect; documents the containment invariant precisely with a reversal trigger; consolidates `user-content-` prefix ownership onto a single layer.

## 4. Security — ACCEPTABLE ✅

Removing `clobberPrefix` reduces a defense-in-depth control, but shadow-DOM encapsulation contains the dangerous vectors and all stripping rules are preserved — no new exploitable path vs `main`.

- **[Low / informational]** Disabling `user-content-` prefixing is a defense-in-depth reduction on a path that renders untrusted user content (history/notes/chat/feed). Independently re-verified on current code: (1) only the `clobberPrefix` option changed — `...defaultSchema` still supplies `protocols`, the `clobber` attr→property map, and the default allow-lists; the existing XSS/dangerous-tag/`javascript:`-URL/dangerous-CSS test suite (`markdown-parser.spec.ts:594-752`) is unchanged and in force; (2) both consumers are real shadow roots, so unprefixed `id`/`name` never populate `document.getElementById`/`window.<name>`/`document.forms`; (3) both functions are non-exported; (4) the dangerous clobber carriers (`iframe`/`form`/`object`/`embed`) are absent from `tagNames` and stripped regardless. The residual — same-shadow-root aria/id collisions within a single attacker-authored blob — involves no privilege crossing and no new XSS vector. This is the accurately-described, documented, deliberately-chosen trade-off of the PR, which the documented-intent rule places at informational.
- **Positives:** Prefix-only change with stripping fully preserved and proven by retained tests; containment reasoning and reversal conditions documented at the code site; new tests assert no `user-content-` leakage and footnote id/link sync on the actual nodes; the separate untrusted-email sanitizer is unaffected.

## 5. Observability — GOOD ✅

- **No significant issues found.** Pure declarative config change: one config key plus a comment, and a spread→direct-call refactor that is semantically identical. No `try`/`catch`/`.catch()` introduced — `unified().process()` rejections still propagate to the async caller; no new silent-failure surface; no logging added (no sensitive-data-in-logs concern). Service/backend logging expectations don't apply to a frontend component library.
- **Positives:** The 25-line inline comment recording the failure mode, the shadow-DOM safety boundary, and the revert conditions is durable debugging context.

## 6. Performance — SAFE ✅

- **No significant issues found.** `clobberPrefix: ''` is perf-neutral-to-marginally-cheaper (skips per-attribute prefix concatenation). Dropping the redundant `{ ...getWhiteList() }` spread at both call sites removes one object allocation + shallow copy per parse — a genuine (tiny) GC win, behavior identical. `getWhiteList` was never memoized on `main` and its call frequency is unchanged; consumer-side whitelist memoization (`markdown.tsx:121-130`) is preserved. The example component and spec tree-shake out of the shipped bundle.
- **Positives:** Surgical change; zero new work on render/watch/memoization paths.

## 7. Lime Platform Issues — SAFE ✅

- **No applicable LPID findings.** Frontend TS/Stencil PR with no Python/UoW/endpoint/handler/task/file-I/O surface, so backend rule families don't apply. The full frontend rule set (LPID-FE-1 … FE-11) was walked against the fresh catalog. LPID-FE-10 (Frontend XSS Sink) targets HTML/code-execution *sinks* fed non-constant input (`innerHTML`/`eval`/etc.) — this PR introduces no such sink; it reconfigures a sanitizer schema, and DOM clobbering is a distinct class with no LPID rule. The defense-in-depth reduction is captured in Section 4. No loops, un-awaited HTTP, lifecycle `await`, state mutation, timers, or secrets introduced.
- **Verification note (not a finding):** the inline comment names `limel-markdown` and `limel-text-editor` as the consumers; a third sanitizer in the tree (`email-viewer/sanitize-email-html.ts`) uses its **own** schema, not the shared `getWhiteList`, so it is genuinely unaffected — the comment's safety reasoning holds.
- **Positives:** No new Lime-platform runtime hazard; the static example introduces no anti-pattern.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Code Quality | **GOOD** |
| Architecture | **GOOD** |
| Security | **ACCEPTABLE** |
| Observability | **GOOD** |
| Performance | **SAFE** |

## Top Recommendations

All remaining items are `[Low]` and optional; none block merge.

1. **[Low from Architecture / Code Quality]** Pre-existing: `markdown-parser.ts` and `email-viewer/sanitize-email-html.ts` independently make the same `clobberPrefix: ''` + `defaultSchema`-re-derivation decision. Not in scope here, but worth a tracking note if a third sanitizer ever appears — a future change to the clobber model would need applying in two places.
2. **[Low from Lime Platform Issues]** Optional one-liner in the inline comment noting that `email-viewer` has its own separate schema, so the "every current consumer" claim stays airtight for future maintainers.

**Bottom line:** Three review passes and six fixups have left this PR with no blockers and no `[Medium]`+ findings — the two prior rounds' `[Medium]`s (sanitizeHTML consistency, missing headline test, deliberate granularity decision) and `[Low]`s (redundant spread, footnote assertions, default-whitelist test) are all genuinely discharged, independently re-verified here. The PR is strictly better than `main`. Ready to merge.

TommyLindh2:
Reviewed fixups and they look good also rebased on main + squashed fixups.

PR #4139, opened 2026-06-17
refactor(text-editor): extract editor-config to enable real-stack testing

john-traas:
## What

Extracts the text editor's schema and plugin assembly out of the `limel-prosemirror-adapter` component into a pure, importable module, and adds an integration test that exercises the **real** editor stack.

- **refactor**: `initializeSchema()` / `createEditorState()` (private methods on the adapter) → pure `buildEditorSchema()` / `buildEditorPlugins()` in `editor-config.ts`. The component delegates to them; plugin order and behaviour are preserved exactly.
- **test**: `editor-config.spec.ts` builds the production schema + the real ordered plugin list and asserts, using the official `prosemirror-test-builder`:
  - commands produce the expected transaction;
  - the link/image `handlePaste` ordering contract (ProseMirror is first-truthy-wins) — guarding against a new/changed plugin silently shadowing another on a shared event;
  - a transaction sequence keeps the document structurally valid.
- **build**: adds a `clean` script (`shx rm -rf .stencil dist www`) and the `prosemirror-test-builder` dev dependency.

## Why

The editor is now used across the CRM, so we need regression coverage for cross-plugin interactions (paste / keymap / transaction chains). Those only surface when the **real, full plugin set** runs against the **real schema** — previously impossible because the schema and plugin list were private to the component. This unblocks that, tests against source (immune to stale-`dist` issues), and needs no bespoke test-harness suite.

## Verification

- `npm run test:spec`: 884/884 pass (incl. the new spec)
- text-editor e2e: 14/14 pass
- Behaviour-preserving: no plugin order or logic changes.

## Summary by CodeRabbit

* **New Features**
  * Added a `clean` script to remove generated build/output directories.
  * Introduced shared ProseMirror editor schema and plugin builders, now used by the adapter.
* **Bug Fixes**
  * Improved clipboard paste handling so link handling is applied before image handling.
  * Ensured paste handlers correctly return no-op behavior for plain-text clipboard data.
* **Tests**
  * Added a real-stack integration test covering schema/plugin setup, Bold command behavior, paste-handler ordering, and document integrity across transactions.


src/components/text-editor/prosemirror-adapter/editor-config.ts
```diff
@@ -0,0 +1,127 @@
...
+import { createTriggerPlugin } from './plugins/trigger/factory';
+import { getTableNodes, getTableEditingPlugins } from './plugins/table-plugin';
+import { getImageNode } from './plugins/image/node';
+import { createNodeSpec } from '../utils/plugin-factory';
+import { ContentTypeConverter } from '../utils/content-type-converter';
+import { CustomElementDefinition } from '../../../global/shared-types/custom-element.types';
+import { Languages } from '../../date-picker/date.types';
+import { TriggerCharacter, InlineImages } from '../text-editor.types';
+
+type ContentType = 'markdown' | 'html';
```
rasmusflomen: Not blocking, just a thought: now that this file names the type, could the adapter import `ContentType` from here instead of repeating the `'markdown' | 'html'` literal at `prosemirror-adapter.tsx:70`? That way there's one definition and the two can't drift apart. Fine to leave since the literal predates this PR.

rasmusflomen (dismissed):
Just one small question, nothing blocking. Other than that it's a clean, behaviour-preserving extraction — I diffed the moved code against the old method bodies and node/mark assembly plus plugin order match exactly. Nice to finally have the link-before-image paste ordering locked down by a real-stack test. Great work, John!

src/components/text-editor/prosemirror-adapter/editor-config.ts
john-traas: Thx for the review and good suggestion.  ⚡ 034ce61ddd7c60289fe8c11f8c064420df9db78f

rasmusflomen (dismissed):
Nice follow-up on the ContentType extraction — pulling it into a named type means the two sites can't drift apart. Still a clean, behaviour-preserving refactor and the real-stack test is a good addition. Great work, John!

etc/lime-elements.api.md
```diff
@@ -772,7 +772,7 @@ export namespace Components {
     // @beta
     export interface LimelProsemirrorAdapter {
         "clear": () => Promise<void>;
-        "contentType": 'markdown' | 'html';
```
Kiarokh, line 775: this won't be a breaking change, right?

john-traas: No, this should not be a breaking change. The accepted values and runtime behavior remain exactly the same: `'markdown' | 'html'`. The only difference is that the inline union has been extracted into the shared `ContentType` type, giving us a single source of truth for the type definition.

rasmusflomen (approved):
Just confirming the ContentType question — extracting the inline union into a named type is structurally identical, so no breaking change there. Clean, behaviour-preserving extraction and the real-stack test locking down link-before-image paste order is a nice addition. Great work, John!

jgroth (approved):

PR #4140, opened 2026-06-17
fix(list): focus pressed row on mousedown to keep clicks on target

Befkadu1:
fix https://github.com/Lundalogik/crm-client/issues/795
## Summary
- A click on a row deep in a scrolled `limel-list` was lost: `delegatesFocus` redirected focus to the first (off-screen) row and scrolled it into view, moving the pressed row out from under the pointer so mouseup landed on empty `<ul>` space and MDC computed index `-1`. This is the cause of the object explorer widget "jumps to top instead of opening the clicked object" bug.
- Fix: on `mousedown`, suppress the default focus delegation and focus the pressed row with `{ preventScroll: true }`. Mouse-only (not `pointerdown`) so touch-drag scrolling isn't blocked; `preventDefault` for any row (so disabled rows don't scroll-jump either), but focus only moves to non-disabled rows.
- Added e2e tests for `radio`/`checkbox` toggling and pressed-row focus.

## Test plan
- [ ] `npx stencil-test --project e2e list/list.e2e` passes (15/15)
- [ ] Object explorer widget: scroll down → load more → click a deep row → it opens, list does not jump to top
- [ ] `radio` and `checkbox` list examples still toggle on click
- [ ] `limel-menu-list` still activates items (shares the pattern; not changed here)

Refs Lundalogik/crm-client#795

## Release Notes

* **Tests**
  * Expanded E2E test coverage for the list component, including verification of focus behavior for selectable, radio, and checkbox list types.

* **Bug Fixes**
  * Improved focus management when interacting with list items via mouse, ensuring the correct item receives focus.

### DIFF
--- a/src/components/list/list.e2e.tsx
+++ b/src/components/list/list.e2e.tsx
@@ -207,6 +207,89 @@ describe('limel-list', () => {
                     selected: true,
                 });
             });
+
+            it('focuses the pressed item, not the first one', async () => {
+                const items = [
+                    { text: 'item 1' },
+                    { text: 'item 2' },
+                    { text: 'item 3' },
+                ];
+                const { root, waitForChanges } = await render(
+                    <limel-list type="selectable" items={items}></limel-list>
+                );
+                await waitForChanges();
+
+                const itemEls =
+                    root.shadowRoot.querySelectorAll('limel-list-item');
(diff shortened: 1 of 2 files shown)


devbymadde (approved):
Great work, tested it locally and it works well! Code looks good, let's ship it :D

PR #4145, opened 2026-06-23
Fix/4060 download button

### LINKED ISSUE #4070: file-viewer: overlapping buttons and duplicate download in unsupported fallback
After enabling `allowDownload={true}` on `<limel-file-viewer>` ([BB PR #1432](https://github.com/Lundalogik/lime-crm-building-blocks/pull/1432)), two visual problems surfaced:

### Problem 1: Toolbar buttons overlap with consumer-positioned UI

`.buttons` in `file-viewer/partial-styles/ui-controls.scss` is `position: absolute; top: 0.25rem; right: 0.25rem`. Any consumer that places its own controls in the top-right (e.g. `limebb-summary-popover` puts its close button at `top: 0.125rem; right: 0.125rem`) will collide. With more `allow*` flags enabled, more buttons stack down and the collision area grows.

Image

There's no CSS `::part` exposed on `.buttons` and no CSS variable for offset, so consumers can't reposition the toolbar.

### Problem 2: Duplicate download button in renderText / renderEmail fallback

`renderText` (`file-viewer.tsx:265-274`) and `renderEmail` (`file-viewer.tsx:276-291`) render `renderButtons()` (the toolbar) AND wrap their `<object>` / `<limel-email-viewer>` with `renderNoFileSupportMessage()` as fallback. That message has its own inline download button (`file-viewer.tsx:343`). When the embedded view fails to render natively (observed for `.json` blob URLs in Chrome), both render → two download buttons.

Image

### Related

- BB PR that enabled allowDownload and surfaced these: https://github.com/Lundalogik/lime-crm-building-blocks/pull/1432
- The popover wrapping uses `limebb-summary-popover` via `limebb-document-chips`, which now renders attachment previews in the email composer.

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/4070

## Bug Fixes

- Prevented duplicate download buttons in the file viewer when unsupported content is shown within other UI that already includes a download button bar.
- Updated unsupported-file fallback rendering so the download button is only displayed when appropriate.
- Adjusted file viewer control button positioning for better layout alignment (moved vertical offset from top to bottom).

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/file-viewer/file-viewer.tsx
+++ b/src/components/file-viewer/file-viewer.tsx
@@ -269,7 +269,7 @@ export class FileViewer {
     };
 
     private renderText = () => {
-        const fallbackContent = [this.renderNoFileSupportMessage()];
+        const fallbackContent = [this.renderNoFileSupportMessage(false)];
 
         return [
             this.renderButtons(),
@@ -291,7 +291,9 @@ export class FileViewer {
                 fallbackUrl={this.sanitizeUrl(this.fileUrl)}
                 language={this.language}
             >
-                <div slot="fallback">{this.renderNoFileSupportMessage()}</div>
+                <div slot="fallback">
+                    {this.renderNoFileSupportMessage(false)}
+                </div>
             </limel-email-viewer>,
         ];
     };
@@ -332,7 +334,15 @@ export class FileViewer {
         return officeViewers[this.officeViewer];
     };
 
-    private renderNoFileSupportMessage = () => {
+    /**
+     * @param withDownloadButton whether to render a download button as a
+     * recovery option. Should be `false` when the message is shown as
+     * fallback content alongside the buttons bar, which already renders the
+     * download button when enabled, to avoid showing two download buttons.
+     */
+    private readonly renderNoFileSupportMessage = (
+        withDownloadButton = true
+    ) => {
         return (
             <div class="no-support" role="alert">
                 <h1>⚠️</h1>
@@ -346,7 +356,7 @@ export class FileViewer {
                         'file-viewer.message.try-other-options'
                     )}
                 </p>
-                {this.renderDownloadButton()}
+                {withDownloadButton && this.renderDownloadButton()}
             </div>
         );
     };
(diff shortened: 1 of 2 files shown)


LucyChyzhova:
`The PR title references issue #4060, but the PR is actually addressing issue #4070. This is misleading, as the title does not accurately reflect the actual issue being fixed.`

yeah, missed the number , should be 4070 instead...

src/components/file-viewer/file-viewer.tsx
```diff
@@ -332,7 +334,13 @@ export class FileViewer {
         return officeViewers[this.officeViewer];
     };
 
-    private renderNoFileSupportMessage = () => {
+    /**
+     * @param withDownloadButton whether to render a download button as a
+     * recovery option. Should be `false` when the message is shown as
+     * fallback content alongside the buttons bar, which already renders its
+     * own download button, to avoid showing two download buttons.
+     */
```
ha2818an, line 342: The JSDoc says the buttons bar "already renders its own download button," but it only does so when allowDownload is true. If we want to be precise maybe  "...which already renders the download button **when enabled**" so it doesn't imply the button is unconditional.

src/components/file-viewer/file-viewer.tsx
```diff
@@ -332,7 +334,13 @@ export class FileViewer {
...
+    private renderNoFileSupportMessage = (withDownloadButton = true) => {
```
ha2818an: Worth adressing to get it cleared

ha2818an (dismissed):
Verified by linking local lime-element to local lime-bb and checking the example docs. I'd say the comments are nits so approving for now.

hatchakka3 (approved):
Changes look good and have been verified by @ha2818an ✅

PR #4146, opened 2026-06-26
Scatter chart

Kiarokh:
Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

## Summary by CodeRabbit

* **New Features**
  * Added support for scatter charts with two-value points, coordinate-style value formatting, and updated tooltip labels.
  * Introduced new scatter chart examples, including a negative-values demo and orientation/dot-size controls.
* **Bug Fixes**
  * Improved scatter axis/grid rendering and zero-crossing line behavior across orientations.
  * Enhanced scatter item and label positioning for clearer “above/below” text layout.
* **Documentation**
  * Clarified how scatter chart point values are interpreted and how display options apply to scatter charts.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -262,6 +262,7 @@ export namespace Components {
         | 'pie'
         | 'ring'
         | 'dot'
+        | 'scatter'
         | 'stacked-bar';
     }
     export interface LimelCheckbox {
@@ -1818,6 +1819,7 @@ export namespace JSX {
         | 'pie'
         | 'ring'
         | 'dot'
+        | 'scatter'
         | 'stacked-bar';
     }
 
@@ -1854,6 +1856,7 @@ export namespace JSX {
         | 'pie'
         | 'ring'
         | 'dot'
+        | 'scatter'
         | 'stacked-bar';
     }
 
--- a/src/components/chart/chart.scss
+++ b/src/components/chart/chart.scss
@@ -7,6 +7,7 @@
 @forward './partial-styles/ring';
 @forward './partial-styles/stacked-bar';
 @forward './partial-styles/nps';
+@forward './partial-styles/scatter';
 @forward './partial-styles/axises';
 @use '../../style/mixins';
 
@@ -16,6 +17,7 @@
 * @prop --chart-item-divider-color: Defines the color that visually separates items in some charts, such as `stacked-bar` chart items. Defaults to `rgb(var(--color-white), 0.6)`.
 * @prop --chart-axis-line-color: Defines color of the axis lines. Defaults to `--contrast-900`. Note that lines have opacity as well, and get opaque on hover.
 * @prop --chart-item-border-radius: Defines the roundness of corners of items in a chart. Defaults to different values depending on the chart type. Does not have any effect on `pie` and `doughnut` types.
+* @prop --chart-scatter-item-size: Defines the diameter of the dots in a `scatter` chart. Defaults to `0.5rem`.
 */
 
 :host(limel-chart) {
--- a/src/components/chart/chart.tsx
+++ b/src/components/chart/chart.tsx
@@ -27,6 +27,7 @@ interface AxisRange {
  * @exampleComponent limel-example-chart-type-dot
  * @exampleComponent limel-example-chart-type-area
  * @exampleComponent limel-example-chart-type-line
+ * @exampleComponent limel-example-chart-type-scatter
  * @exampleComponent limel-example-chart-type-pie
  * @exampleComponent limel-example-chart-type-doughnut
  * @exampleComponent limel-example-chart-type-ring
@@ -35,6 +36,7 @@ interface AxisRange {
  * @exampleComponent limel-example-chart-multi-axis
  * @exampleComponent limel-example-chart-multi-axis-with-negative-start-values
  * @exampleComponent limel-example-chart-multi-axis-area-with-negative-start-values
+ * @exampleComponent limel-example-chart-type-scatter-negative
  * @exampleComponent limel-example-chart-axis-increment
  * @exampleComponent limel-example-chart-clickable-items
  * @exampleComponent limel-example-chart-accessibility
--- a/src/components/chart/chart.tsx
+++ b/src/components/chart/chart.tsx
@@ -11,6 +11,8 @@ interface AxisRange {
     minValue: number;
     maxValue: number;
     totalRange: number;
+    /** Step between adjacent axis lines, used to render the grid. */
+    increment: number;
 }
 
 /**
@@ -124,6 +126,7 @@ export class Chart {
         | 'pie'
         | 'ring'
         | 'dot'
+        | 'scatter'
         | 'stacked-bar' = 'stacked-bar';
 
     /**
@@ -153,7 +156,22 @@ export class Chart {
     @Prop({ reflect: true })
     public loading: boolean = false;
 
+    /**
+     * Range of the value axis. For most chart types this is the single value
+     * axis. For `scatter` it is the range of each point's second value
+     * (`value[1]`); the shared axis-line rendering always draws its grid and
+     * badges from this range. Whether it runs vertically (landscape) or
+     * horizontally (portrait) is decided by `orientation`.
+     */
     private range: AxisRange;
+
+    /**
+     * Range of each scatter point's first value (`value[0]`). Only used by the
+     * `scatter` type, where both values are plotted on value axes. It is laid
+     * out on the opposite screen edge to `range` — horizontally in landscape,
+     * vertically in portrait.
+     */
+    private secondaryAxisRange: AxisRange;
 
     /**
      * Fired when a chart item with `clickable` set to `true` is clicked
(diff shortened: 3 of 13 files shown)


Kiarokh:
> 🤖 AI-generated review from 6 parallel agents, re-synthesized against the current branch. Treat as input, not a verdict — agents can be wrong or miss context.

# Consolidated PR Review

## PR Summary

Adds a `scatter` chart type to `limel-chart`: each item is an absolutely-positioned dot at an `[x, y]` coordinate (reusing the existing `value: [number, number]` tuple), with both axes auto-ranged and negative-safe — each drawing its own grid, zero line, and badges — `orientation` transposing the plot, and the display-label props honored. 12 files, +764 / −43, in four atomic commits (refactor → feat → docs → docs).

Strictly better than `main` in every dimension the agents examined — no regressions, no new security/perf/compatibility breaks, and the negative-safe increment math even fixes a latent `NaN` that `main` produced for all-negative data. An initial pass flagged two `[Medium]` items (the `axisIncrement` behaviour and a zero-span `NaN`) and one documentation nit; **all three were fixed in-branch before this review** — see _Addressed during development_ below. No open High/Medium items remain.

- **Blockers:** None.
- **Non-blocking nits:** in the per-dimension collapsibles below.

---

## Dimensions

No backward-incompatible changes; everything is additive or behavior-equivalent for existing types.

**What works well:** the `'scatter'` union member only widens the public type, `ChartItem` is unchanged, and `--chart-scatter-item-size` is the only new public surface; the `renderAxises`→`renderAxisLines` and `calculateAxisIncrement(number[])` refactors were verified numerically identical for bar/dot/line/area/pie, and the `:not([type='scatter'])` axis-rule scoping plus the new `.horizontal-lines`/`.vertical-lines` rules can't match non-scatter charts. The `[x,y]`-vs-`[start,end]` tuple ambiguity is now documented on `ChartItem.value`.

**Issues:** None.

### 2. Code Quality — GOOD ✅

Cohesive helpers and correct edge handling; both previously-flagged `[Medium]` items are resolved, leaving only Low nits.

**What works well:** the new private helpers (`calculateNiceStep`/`axisSpan`/`calculateAxisRange`/`normalize`) are single-purpose and well-commented; `calculateNiceStep` fixed a `Math.log10(negative) → NaN` hazard; `normalize` now guards a zero-span axis.

**Issues:** None.

<details>
<summary>Minor nits (4)</summary>

- 🔵 `getXValue`/`getYValue` (`chart.tsx`) plot a scalar `value` at `(v, v)` on the diagonal; the docs now emphasize the `[x, y]` tuple but don't call out this scalar fallback, so it remains an undocumented surprise.
- 🔵 A point whose value equals the rounded axis max normalizes to exactly `100%`, so the centered dot half-overflows the plot edge (`chart.tsx` `normalize` + `_scatter.scss`) — cosmetic.
- 🔵 SCSS duplication: `.axises.horizontal-lines`/`.vertical-lines` repeat the landscape/portrait `.axis-line` bodies, and `_scatter.scss`'s axis-label placement repeats `_layout-for-charts-with-x-y-axises.scss` — a shared mixin would cut the copy-paste.
- 🔵 `limel-chart` has no spec/e2e tests, so the new math helpers ship untested (the component has none, so just a note).

</details>

### 3. Architecture — GOOD ✅

Scatter fits the existing CSS-variable + per-type-branch architecture coherently; the `axisIncrement`/two-axis concern from the first pass is resolved (scatter now auto-derives both axes and no longer mutates the prop field).

**What works well:** reusing the shared `range` field for `value[1]` so the existing axis-line machinery works unchanged is sound and well-documented; `AxisRange.increment` is the right home for a per-axis step; the structure extends cleanly to a future bubble-size or second two-axis type.

**Issues:** None.

<details>

- 🔵 The orientation→screen-axis flip is encoded in two places that must stay in lockstep — TS picks which range feeds `horizontal-lines`/`vertical-lines`, CSS maps `offset-x/-y` to edges (`chart.tsx` `renderScatterAxises`, `_scatter.scss`); consistent today, inherent to the CSS-variable design.
- 🔵 Scatter is special-cased across ~6 methods (consistent with how pie/doughnut/line already branch), worth watching if a third tuple-consuming type ever arrives.

</details>

### 4. Security — SAFE ✅

No new security issues; presentational component with consistently escaped output.

**What works well:** `item.text` and the `(x, y)` value render only as escaped JSX text nodes through `<td>`, `limel-badge`, and `limel-tooltip-content` (no `innerHTML` anywhere); the new offset/size CSS variables are component-derived numbers. The `item.color → --limel-chart-item-color` write is the same pre-existing pattern every chart type uses (scatter only adds an early-return in front of it), and custom-property values can't break out of the declaration via the DOM `style` API.

**Issues:** None.

### 5. Observability — GOOD ✅

For a presentational component with no logging by design, scatter improves accessible-DOM observability and never crashes.

**What works well:** scatter keeps the full accessible `<table>` (caption, `<th>` headers, per-item text/value cells), renders the value as the semantically-correct `(x, y)`, and deliberately keeps the value label visible at `y=0` by disabling `has-value-zero`; empty `items` and the all-at-origin degenerate case now both degrade to a stable, finite render.

**Issues:** None.

<details>

- 🔵 A non-tuple `value` on a scatter chart is silently misread as a diagonal point with no dev-time signal (same root as the Code Quality scalar nit); a dev-mode warning when a scatter item's `value` isn't a 2-tuple would close the gap.
- 🔵 Arbitrary `NaN`/`Infinity` coordinates still propagate to `NaN%` offsets and the dot silently disappears (the all-values-zero case is now guarded; literal `NaN`/`Infinity` data is not) — consistent with `main`'s unguarded handling of bad values.

</details>

### 6. Performance — GOOD ✅

No performance regressions; the scatter render path is, if anything, slightly cheaper per item than the line/area branches.

**What works well:** `getScatterItemStyle` skips the line/area next-item lookup; recompute only fires on `@Watch`/load, not per render; `mix-blend-mode: hard-light` matches the existing `dot` type; no new listeners, timers, or retained references.

**Issues:** None.

<details>

- 🔵 `recalculateScatterRanges` makes several O(n) passes and spreads the value arrays into `Math.max/min` — a pre-existing pattern, negligible at realistic sizes, and recompute is rare (`chart.tsx`).
- 🔵 `renderItems` still calls `calculateSizeAndOffset` for scatter items even though `getScatterItemStyle` discards the result (`chart.tsx`) — dead O(1) work per item, not worth optimizing.

</details>

---

## Addressed during development

Caught in an initial pass and already fixed in this branch — listed for the record, not as open work:

1. **✅ [was 🟡 Medium] `axisIncrement` on a two-axis chart** — scatter now auto-derives **both** axes independently and ignores `axisIncrement` (the same stance it takes on `maxValue`, since one scalar prop can't sensibly drive two axes of different scale), and no longer mutates `this.axisIncrement`. Documented in the scatter example. _(`chart.tsx` `recalculateScatterRanges`, `chart-type-scatter.tsx`)_
2. **✅ [was 🟡 Medium] Zero-span axis → `NaN` positions** — `normalize` now guards `range.totalRange === 0` and centers the point, so a lone point or all-at-origin data renders a visible dot instead of a blank `NaN`-positioned one. _(`chart.tsx` `normalize`)_
3. **✅ [was 🔵 Low] Tuple semantics footgun** — `ChartItem.value` and the scatter example now spell out that a `[number, number]` is `[x, y]` for `scatter` but `[start, end]` elsewhere. _(`chart.types.ts`, `chart-type-scatter.tsx`)_

## Top Recommendations

None open — the two `[Medium]` items above are already resolved in-branch. Remaining items are all `[Low]` nits in the per-dimension collapsibles; none block merge.

john-traas:
Really nice addition — the scatter type works well. One thing I ran into while reviewing, in **portrait** orientation: the points come out vertically mirrored against their gridlines. A point at max-X sits at the *bottom* even though its `40` badge is at the *top*. Landscape is spot on; it's only portrait's vertical axis that's flipped.

The cause is that portrait anchors the dots with `top:` while the gridlines use `column-reverse` (minimum at the bottom):
https://github.com/Lundalogik/lime-elements/blob/100cbcebb555207e04b80c86fd2b583dafa84c7e/src/components/chart/partial-styles/_scatter.scss#L40-L49

I've pushed a small fixup (`24a0a8e7`) that anchors from the bottom, the way landscape does — verified in a build that every point then lands on its gridlines in both orientations. There's also a second commit extracting the axis math into a little module with a spec (following the `tabs.ts` / `tabs.spec.ts` pattern). What do you think?

Kiarokh:
Thanks @john-traas I don't see your fixup. Maybe it was squashed already? But when I check the docs now, I see that the orientation doesn't work as expected. The chart's X and Y axises are supposed to flip, (see bar chart example, to know what I mean). So I think the fixup unfortunately broke the intended design.

john-traas (approved):

adrianschmidt (approved):
Changes to the external API approved.

Issue #409, opened 2019-09-05
Possibility to autofocus limel-input-field


joelricci:
## New feature motivation

It would be really neato if you, for instance, have a dialog with an input field and have the input field to autofocus when you open the dialog.

## New feature description

A new `autofocus` property on `limel-input-field` would do the trick. 

## New feature implementation

I tried
```
10 PRINT "AUTOFOCUS"
20 GOTO 10
```
but it didn't work.

Kiarokh:
This seems to be solve already. Because we're doing it on column filter popovers. I close it

PR #4151, opened 2026-07-01
fix(markdown): reset image height correctly

devbymadde:
We've gotten multiple bug reports that images in emails are scaled incorrectly and "squashed", i.e they take up the full width but has a set height causing them to render incorrectly. 

It was initially thought to be fixed in [#409](https://github.com/Lundalogik/crm-client/issues/409
) but didn't work for emails specifically when they have a fixed height. Due to css rules the inline styling will take precedence over our class `reset-img-height`, hence the `!important` to make sure that the auto height is always prioritized.

To reproduce this for emails you can do the following steps:

1. Build your local lime-elements
2. Link lime-elements to your local instance running email integration
3. Compose an email in Outlook and attach an inline image. Click the image and in the side panel that opens to the right, increase the height by 1 px, then decrease it again. Sounds confusing, but it's the only way I've managed to reproduce that the inline style height is actually sent.
4. Image should not be squashed in the conversation view, but use the correct aspect ratio.

Closes: [#625](https://github.com/Lundalogik/crm-client/issues/625)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved markdown image rendering so the reset styling more reliably preserves images’ natural height by increasing the CSS override precedence.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/markdown/markdown.scss
+++ b/src/components/markdown/markdown.scss
@@ -41,7 +41,7 @@ hr {
     // added fixed heights to images that appear in the activity feed.
     #markdown {
         img {
-            height: auto;
+            height: auto !important;
         }
     }
 }


FredrikWallstrom:
I'm not able to reproduce this 🤔

FredrikWallstrom (approved):
Great work!

adrianschmidt (dismissed):
Hmm… I don't think `!important` in CSS can override inline CSS, but I might be wrong. Just holding this PR here while I check.

adrianschmidt (commented):
I've checked, and yes, CSS with `!important` _does_ override inline CSS, as long as the inline CSS doesn't _also_ have `!important`.

_However,_ this is yet another PR that is essentially an attempt at fixing an problem specific to rendering emails with crappy or broken source input, by adding stuff to `limel-markdown`, which then risks creating problems for consumers that are _not_ providing crappy input.

We need to stop polluting limel-markdown with these "fixes" for emails, and either pre-process the source input, or create a copy of limel-markdown specifically for emails, in lime-crm-building-blocks for example.

Anyway, I'll remove my request for changes. If there's currently no better place to do this than in limel-markdown, I guess I won't stop you. But I _will_ make a big breaking change in limel-markdown to remove all these email fixups one day, and put them all in a special component instead 😄

devbymadde:
> However, this is yet another PR that is essentially an attempt at fixing an problem specific to rendering emails with crappy or broken source input, by adding stuff to limel-markdown, which then risks creating problems for consumers that are not providing crappy input.

I understand and kind of agree with you - however, since the fix is in an existing class (`.reset-img-height`) that is specifically created for this scenario but didn't work as intended, and is also only applied from another of our internal packages, I figured that it made sense to put it there.

adrianschmidt:
> > However, this is yet another PR that is essentially an attempt at fixing an problem specific to rendering emails with crappy or broken source input, by adding stuff to limel-markdown, which then risks creating problems for consumers that are not providing crappy input.
> 
> I understand and kind of agree with you - however, since the fix is in an existing class (`.reset-img-height`) that is specifically created for this scenario but didn't work as intended, and is also only applied from another of our internal packages, I figured that it made sense to put it there.

Yeah, absolutely! 👍

adrianschmidt (approved):

PR #4154, opened 2026-07-01
Add loading, progress, and invalid states to limel-file (component and per-file)

Kiarokh:
## Why

`limel-file` had no way to reflect that something is happening to the selected file — a known gap. Its consumer `limec-file-picker` can today only grey the field out via `disabled` during work, and there is no way to surface upload progress at all.

>[!important]
>Before starting the review, please note that some of the decisions in this PR may feel initially strange. This is due to two reasons:
>1. `limel-file` allows only selection of one single file. We know consumers have requested multiple file selection possibility. We want to enable this feature in future. To keep this component future-proof this means, there shouldn't be a one-on-one relationship between the file(chip) and the file picker. Each file can individually become `invalid` or be on `loading` state.
>2. An optimal `FileInfo` interface, should have a `status` prop or similar (name not decided yet). That would allow each file to reflect its own status. For example: "Uploading…", "Failed!", "43%" (progressed, loaded, or uploaded), etc… Right now in this PR, we have some accessibility issues with aria-live, and we can't properly explain to the assistive tech what is going on. But a future per-file `status` in the follow up PR will help solve this issue. Such statuses will also be reflected on the chip's badge itself, helping sighted users to understand the status of what is happening better

Two tiers of "busy" state, built up in atomic commits:

- **`feat(file): add a loading prop`** — a component-level `loading` prop: the component's own indeterminate busy state (spinner + `aria-busy`), controlled entirely by the consumer, shown whether or not a file is selected. Does not disable interactivity.
- **`feat(file): support a per-file loading state`** — `FileInfo.loading`, rendered as an indeterminate indicator on that file's chip. A loading file also puts the parent into its loading state. Independent of the component prop; the multi-file-ready per-file model.
- **`feat(chip-set): forward the progress property to the chip`** — `limel-chip` already had a determinate `progress` bar, but the `Chip` model consumed by `limel-chip-set` didn't expose it. Adds `progress` to `Chip` + forwards it (with a chip-set example).
- **`feat(file): support a per-file progress state`** — `FileInfo.progress` (0–100), a determinate bar on the file's chip, for measurable work like uploads or client-side resizing.
- **`feat(file): support a per-file invalid state`** — `FileInfo.invalid`, an error state on the file's chip, independent of the component-level `invalid` prop.

Per-file status lives on `FileInfo`, so it maps 1:1 onto the chip and is ready for a future multi-file `limel-file` — each file owns its own state, the pattern already used by `limel-chip-set`'s `Chip[]` and `limebb-document-chips`. Every field is optional and defaults to off, so this is fully backward compatible. Each feat carries its own documentation example.

## Design notes

- **Component vs. per-file.** `loading` exists both at the component level (the whole picker is busy — a spinner) and per file (`FileInfo.loading` — the chip's own indicator). A loading file bubbles up to the component's busy state (`aria-busy` + spinner); the two are otherwise independent. `progress` and per-file `invalid` are per-file only. There is intentionally **no** component-level `progress` — a cumulative/overall bar can be added later if/when the component supports multiple files.
- Per-file status only renders while a file is selected (there must be a chip to render it on).
- `loading` does not disable interactivity, matching other components' loading states.
- ARIA conveys *that* the component is busy (`aria-busy`), not *what* — announcing the specific activity ("Uploading…") is a natural fit for a future per-file status label.

## Summary by CodeRabbit

- **New Features**
  - Added `loading` support to the file component, including busy behavior with a spinner.
  - Added progress support (0–100) to file/chip rendering, enabling determinate progress display.
  - Added per-file `invalid` and `helperText` handling to improve messaging in chip UI.
- **Style & Accessibility**
  - Updated busy state semantics (e.g., `aria-busy`) and adjusted UI while busy.
- **Examples & Documentation**
  - Added new examples covering file loading, per-file loading, per-file progress, per-file invalidation, and chip-set progress.
  - Extended documentation example tags for the new loading/progress scenarios.

### DIFF
--- a/src/components/file/examples/file-loading.tsx
+++ b/src/components/file/examples/file-loading.tsx
@@ -1,0 +1,61 @@
+import { FileInfo } from '@limetech/lime-elements';
+import { Component, h, Host, State } from '@stencil/core';
+
+/**
+ * Loading
+ *
+ * Set the `loading` prop to `true` to put the whole component in an
+ * indeterminate busy state, rendering a spinner and setting `aria-busy` on the
+ * component. This is the component's own loading state, controlled entirely by
+ * the consumer: it is shown whether or not a file is selected, and stays until
+ * you set `loading` back to `false`. Use it for work that has no measurable
+ * percentage, such as reading a selected file's metadata.
+ *
+ * :::note
+ * Setting `loading` does _not_ disable the interactivity of the component. If
+ * it should be disabled meanwhile, set `disabled` separately.
+ * :::
+ */
+@Component({
+    tag: 'limel-example-file-loading',
+    shadow: true,
+})
+export class FileLoadingExample {
+    @State()
+    private value: FileInfo = {
+        filename: 'annual-report.pdf',
+        id: 1,
+    };
+
+    @State()
+    private loading = false;
+
+    public render() {
+        return (
+            <Host>
+                <limel-file
+                    label="Attach a file"
+                    value={this.value}
+                    loading={this.loading}
+                    onChange={this.handleChange}
+                />
+                <limel-example-controls>
+                    <limel-switch
+                        label="Loading"
+                        value={this.loading}
+                        onChange={this.setLoading}
+                    />
+                </limel-example-controls>
+            </Host>
+        );
+    }
+
+    private handleChange = (event: CustomEvent<FileInfo>) => {
+        this.value = event.detail;
+    };
+
+    private setLoading = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.loading = event.detail;
+    };
+}
--- a/src/components/file/examples/file-loading.tsx
+++ b/src/components/file/examples/file-loading.tsx
@@ -15,6 +15,9 @@ import { Component, h, Host, State } from '@stencil/core';
  * Setting `loading` does _not_ disable the interactivity of the component. If
  * it should be disabled meanwhile, set `disabled` separately.
  * :::
+ *
+ * To reflect a busy state on an individual file instead, set `loading` on that
+ * file — see the per-file loading example.
  */
 @Component({
     tag: 'limel-example-file-loading',
--- a/src/components/file/examples/file-per-file-loading.tsx
+++ b/src/components/file/examples/file-per-file-loading.tsx
@@ -1,0 +1,66 @@
+import { FileInfo } from '@limetech/lime-elements';
+import { Component, h, Host, State } from '@stencil/core';
+
+/**
+ * File loading
+ *
+ * A file can carry its own busy state via `loading` on its `FileInfo`, which
+ * renders an indeterminate indicator on that file's chip — for example while
+ * that specific file is being processed.
+ *
+ * Because a busy file means the component has work in progress, setting a
+ * file's `loading` — or its `progress` — also automatically puts the whole
+ * component into its loading state (the busy spinner shows and `aria-busy` is
+ * set), informing both users and assistive tech about the status.
+ */
+@Component({
+    tag: 'limel-example-file-per-file-loading',
+    shadow: true,
+})
+export class FilePerFileLoadingExample {
+    @State()
+    private file?: FileInfo = {
+        filename: 'annual-report.pdf',
+        id: 1,
+    };
+
+    @State()
+    private loading = false;
+
+    public render() {
+        const value: FileInfo | undefined = this.file && {
+            ...this.file,
+            loading: this.loading,
+        };
+
+        return (
+            <Host>
+                <limel-file
+                    label="Attach a file"
+                    value={value}
+                    onChange={this.handleChange}
+                />
+                <limel-example-controls
+                    style={{
+                        '--example-controls-column-layout': 'auto-fit',
+                    }}
+                >
+                    <limel-switch
+                        label="File loading"
+                        value={this.loading}
+                        onChange={this.setLoading}
+                    />
+                </limel-example-controls>
+            </Host>
+        );
+    }
+
+    private handleChange = (event: CustomEvent<FileInfo>) => {
+        this.file = event.detail;
+    };
+
+    private setLoading = (event: CustomEvent<boolean>) => {
+        event.stopPropagation();
+        this.loading = event.detail;
+    };
+}
--- a/src/components/file/examples/file-size-badge.tsx
+++ b/src/components/file/examples/file-size-badge.tsx
@@ -36,7 +36,11 @@ export class FileSizeBadgeExample {
                     onChange={this.handleChange}
                     value={value}
                 />
-                <limel-example-controls>
+                <limel-example-controls
+                    style={{
+                        '--example-controls-column-layout': 'auto-fit',
+                    }}
+                >
                     <limel-switch
                         value={this.includeSize}
                         label="Include file size"
--- a/src/components/file/file.tsx
+++ b/src/components/file/file.tsx
@@ -49,6 +49,7 @@ const DEFAULT_FILE_CHIP: Chip = {
  * @exampleComponent limel-example-file-basic
  * @exampleComponent limel-example-file-custom-icon
  * @exampleComponent limel-example-file-size-badge
+ * @exampleComponent limel-example-file-loading
  * @exampleComponent limel-example-file-menu-items
  * @exampleComponent limel-example-file-accepted-types
  * @exampleComponent limel-example-file-composite
@@ -104,6 +105,14 @@ export class File {
     public invalid = false;
 
     /**
+     * Set to `true` to put the component in the `loading` state, and render an
+     * indeterminate progress indicator. This does _not_ disable the
+     * interactivity of the component!
+     */
+    @Prop({ reflect: true })
+    public loading = false;
+
+    /**
      * The [accepted file types](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file#unique_file_type_specifiers)
      */
     @Prop({ reflect: true })
@@ -129,7 +138,7 @@ export class File {
 
     public render() {
         return (
-            <Host>
+            <Host aria-busy={this.loading ? 'true' : 'false'}>
                 <limel-file-dropzone
                     disabled={this.disabled || this.readonly || !!this.value}
                     accept={this.accept}
@@ -138,12 +147,21 @@ export class File {
                     {this.renderChipset()}
                 </limel-file-dropzone>
                 {this.renderDragAndDropTip()}
+                {this.renderSpinner()}
             </Host>
         );
     }
 
+    private renderSpinner() {
+        if (!this.loading) {
+            return;
+        }
+
+        return <limel-spinner />;
+    }
+
     private renderDragAndDropTip() {
-        if (this.value || this.disabled || this.readonly) {
+        if (this.value || this.disabled || this.readonly || this.loading) {
             return;
         }
 
--- a/src/components/file/file.tsx
+++ b/src/components/file/file.tsx
@@ -51,6 +51,7 @@ const DEFAULT_FILE_CHIP: Chip = {
  * @exampleComponent limel-example-file-size-badge
  * @exampleComponent limel-example-file-loading
  * @exampleComponent limel-example-file-per-file-loading
+ * @exampleComponent limel-example-file-per-file-progress
  * @exampleComponent limel-example-file-menu-items
  * @exampleComponent limel-example-file-accepted-types
  * @exampleComponent limel-example-file-composite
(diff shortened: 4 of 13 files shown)


src/components/file/examples/file-per-file-loading.tsx
```diff
@@ -0,0 +1,66 @@
+import { FileInfo } from '@limetech/lime-elements';
+import { Component, h, Host, State } from '@stencil/core';
+
+/**
+ * Per-file loading
```
Kiarokh: **Why is this independent of the component-level `loading` prop you wonder?** _(see the loading example)_

Because this is what a future multi-file `limel-file` would use to reflect the status of each individual file. Today, `limel-file` only allows consumers to select one single file. But we know that many have requested a file-picker component that allows selection of multiple files. In such a file picker, each file can have its own various states (e.g. loading, progress, invalid, etc…)

src/components/file/file.tsx
```diff
@@ -71,6 +75,13 @@ export class File {
     @Prop({ reflect: true })
     public label: string;
 
+    /**
+     * Optional helper text to display below the component. When the component
+     * is `invalid`, it is rendered as an error message.
```
Befkadu1: This description with the invalid behavior makes a general-purpose prop sound error-specific.

```suggestion
     * Optional helper text to display below the component. 
```

src/components/file/examples/file-loading.tsx
```diff
@@ -0,0 +1,61 @@
...
+                        value={this.loading}
+                        onChange={this.setLoading}
+                    />
+                </limel-example-controls>
+            </Host>
+        );
+    }
+
+    private handleChange = (event: CustomEvent<FileInfo>) => {
+        this.value = event.detail;
```
Befkadu1, line 57: Maybe is it good to set `this.loading = false` if the attached file is removed? It doesn't make sense that it shows loading icons while the file is already removed. I know that this example is only to show the loading example but might be good to have a realistic example

src/components/file/examples/file-size-badge.tsx
```diff
@@ -36,7 +36,11 @@ export class FileSizeBadgeExample {
                     onChange={this.handleChange}
                     value={value}
                 />
-                <limel-example-controls>
+                <limel-example-controls
```
Befkadu1, line 39: This change should not supposed to be in this commit

src/components/file/examples/file-loading.tsx
Kiarokh: I think the component might be loading due to some other reason. Maybe it doesn't have the file, but some background process is making it ready for the upload, or whatever... It's basically trying to get ready to go to the file upload party

src/components/file/examples/file-size-badge.tsx
Kiarokh: I know 😄 Just ignore it please

Befkadu1 (dismissed):

Kiarokh:
> 🤖 AI-generated review from 6 parallel agents, updated for the current revision. Treat as input, not a verdict — agents can be wrong or miss context.

## PR Summary

Adds loading / progress / invalid states to `limel-file` at two levels: a component-level `loading` prop (spinner + `aria-busy`) and per-file status on the shared `FileInfo` (`loading` / `progress` / `invalid`) rendered on the file's chip. A file that is `loading` or has `progress` also puts the whole component into its busy state; per-file `invalid` deliberately does not (validity stays the consumer's call). Also forwards `Chip.progress` through `limel-chip-set`, adds a `helperText` prop, and ships a documentation example per feature. 6 atomic commits; every new field/prop is optional and defaults off.

Strictly additive and better than `main` on every dimension — no existing API or behavior changes, and defaults-off means the rendered output is unchanged when the new props are unset. The design choices that might look surprising at first (per-file status on `FileInfo`; the `aria-busy`-without-`aria-live` gap) are explained in the PR description and tied to the planned multi-file + per-file `status` follow-up.

- **Blockers:** None.
- **Non-blocking but worth knowing:** one accessibility item, already acknowledged in the description and deferred to the follow-up — see Top Recommendations.

---

## Dimensions

Purely additive; nothing existing changes.

**What works well:** every new surface (`FileInfo.loading/progress/invalid`, `Chip.progress`, `limel-file` `loading`/`helperText`) is optional and defaulted; the API report changes are all additions, none removed or re-typed; new render paths are gated on the new state, so unset props reproduce `main`'s output exactly. Other `FileInfo` consumers (file-dropzone, file-input, profile-picture) only read known fields, so extra optional fields can't break them.

**Issues:** None.

### 2. Code Quality — GOOD ✅

Consistent with siblings; docs match behavior.

**What works well:** prop declarations, the `aria-busy` ternary, and `helperText` forwarding all match existing components; the "`helperText` renders as an error when invalid" JSDoc is accurate through the chip-set → helper-line chain; each feature ships a focused example, and the examples clearly document the two-tier busy model and that a per-file `invalid` does not auto-invalidate the field.

**Issues:** None.

<details>

- 🔵 No spec/e2e coverage for the new states — but `limel-file` has no spec at all today, so this is a pre-existing gap, not a regression.

</details>

### 3. Architecture — GOOD ✅

Coherent two-tier model, now explicitly motivated.

**What works well:** the data flow `limel-file → FileInfo → getChipArray() → chip-set → chip` forwards 1:1 onto the existing `Chip` model; putting per-file status on `FileInfo` is a deliberate, documented choice for the coming multi-file component (avoiding a one-to-one file↔picker coupling); and the bubbling rule is principled — the "busy" states (`loading`, `progress`) lift to the component, while validity stays a form-level decision the consumer owns, matching `limel-chip-set`.

**Issues:** None.

### 4. Security — SAFE ✅

No new injection surface.

**What works well:** `helperText` renders as escaped JSX text, never `innerHTML`; `aria-busy` is derived from booleans; the `progress` bar's CSS-custom-property sink is pre-existing on `limel-chip`, and this PR only forwards the already-typed number.

**Issues:** None.

<details>

- 🔵 `progress` isn't clamped to 0–100 (`chip.tsx:441-460`, forwarded via `file.tsx` `getChipArray`); pre-existing and not exploitable (a CSS custom-property value can't escape into other rules — worst case is a self-inflicted width glitch), just a robustness nice-to-have.

</details>

### 5. Observability — GOOD ✅

Real busy signal, now accurate for uploads too.

**What works well:** `aria-busy` on the host is a genuine signal that didn't exist on `main`, and it's now driven by `loading` **and** `progress`, so it's correct during determinate uploads as well; determinate per-file `progress` rides on the chip's already-accessible `role="progressbar"` + `aria-valuenow`.

**Issues:**
- 🟡 **[Medium]** Busy state announces *that* it's busy, not *what* (`file.tsx:144`, `renderSpinner`). See **Top Recommendation #1** — already acknowledged in the PR description and deferred to the follow-up.

<details>

- 🔵 Redundant (not conflicting) nested `aria-busy` when a per-file `loading` is set: both the host (`file.tsx:144`) and the chip's inner button (`chip.tsx:246`) get `aria-busy="true"`. Harmless.
- 🔵 The chip progress bar's accessible name is `aria-label="%"` (`chip.tsx:451`), which reads poorly ("percent, 40"); pre-existing chip behavior, now given a bit more exposure.

</details>

### 6. Performance — SAFE ✅

No new hot paths or leaks.

**What works well:** the spinner renders only while busy and unmounts declaratively (no timers/listeners); its animation is pure CSS on the compositor; progress updates change a single CSS custom property, not DOM structure; the new chip fields are trivial pass-throughs reusing the chip's existing progress rendering.

**Issues:** None.

<details>

- 🔵 The `isLoading` getter is recomputed on each of its ~3 calls per render (`file.tsx`, `isLoading` getter) — a couple of property reads, immeasurable.
- 🔵 `@Prop({ reflect: true }) loading` writes a `loading` attribute *and* the render sets `aria-busy` — two host attribute writes per toggle (`file.tsx:116,144`); low-frequency and inconsequential, matching the existing `invalid`/`disabled` pattern.

</details>

---

## Top Recommendations

1. **🟡 [Medium from Observability] — No spoken feedback when the busy state changes (already planned as a follow-up)**
   - **Where:** `src/components/file/file.tsx:144` (host `aria-busy`) and `renderSpinner`; `src/components/spinner/spinner.tsx` (the spinner has no role/label).
   - **What:** `aria-busy` is a passive state — screen readers don't announce load start/finish when it toggles, and the spinner carries no role or label of its own. A `role="status"` / `aria-live` region, or an accessible label, would give spoken feedback.
   - **Why:** assistive-tech users get no cue that work began or completed. This is explicitly acknowledged in the PR description and slated for the follow-up per-file `status` PR (which will also surface on the chip badge for sighted users), so it is **non-blocking** — captured here only so reviewers know it was checked.

Befkadu1 (approved):

src/components/file/file.tsx
```diff
@@ -132,12 +150,35 @@ export class File {
...
         );
     }
 
+    /**
+     * The component is busy for any reason: its own `loading`, or a file that
+     * is `loading` or has `progress` (including `0`). Completion is signalled
+     * by the consumer clearing these — not by `progress` reaching `100`, since
+     * a file at `100%` may still be finalizing (e.g. awaiting the server).
+     */
+    private get isBusy(): boolean {
```
Kiarokh, line 164: **why this is called `isBusy` and not `isLoading`?**

Because I got a feedback, 
<details>
  <summary>See the comment</summary>
  src/components/file/file.tsx:602-608 — isLoading's "busy" semantics disagree with the chip's own progress rendering at the value boundaries.
isLoading treats any defined progress (including 0) as busy: this.value?.progress !== undefined. But renderProgressBar() in src/components/chip/chip.tsx:442 uses a falsy check, if (!this.progress) return;, which was previously dead code for this component since chip-set.tsx never forwarded progress — this PR activates it. Result:
- progress: 0 (the exact value the shipped example file-per-file-progress.tsx starts at): the file shows a spinner + aria-busy="true", but the chip's own progress bar doesn't render at all.
- progress: 1–99: both the file-level indeterminate spinner and the chip's determinate bar render simultaneously — two contradictory busy indicators for the same operation.
- progress: 100: the spinner keeps spinning even though the value implies completion.
This was independently flagged by all four review angles — high confidence.
</details>


and the gist of it was: 
> Don't show the indeterminate spinner while the chip is already showing a determinate bar. 

I disagreed. Because if one (or several files in future version of this component) are in progress, for the parent component it means something is happening inside me. I'm busy for whatever reason. So I'll remain busy until the internal processes of my children are done. Then I'll show that I'm done.

Sanity-checking the matter with Claude, and this is the final comment form that discussion. Posting it here, in case anyone else wonders about these decissions:

You're right, and it's a better argument than the one I was making. I'll drop the recommendation.

The flaw in "don't show the spinner while a bar is showing" is that it treats the spinner and the bar as two answers to the same question. They're not — they're at different scopes:

- **Component spinner** answers "does this component have unsettled work?" — a boolean about the container.
- **Chip bar** answers "how far along is this file?" — a detail about one child.

The spinner isn't claiming the file's duration is unknown; it's claiming the component isn't settled. Both can be true at once without contradiction. So your model — "I'm busy for whatever reason, and I stay busy until my children's work is done, then I show done" — is coherent, and the three "boundaries" the reviewer flagged all resolve cleanly under it:

- 0 → component busy (spinner), child has nothing measurable to draw yet. Fine.
- 1–99 → component busy (spinner) + child detail (bar). Complementary, not contradictory.
- 100 → 100% is bytes transferred, not operation complete. With the upload flow you're still waiting on the server's 201. The component is legitimately still finalizing, so it stays busy until the consumer clears the state. The spinner only "spins forever" if a consumer forgets to clear it — which is a consumer bug, not ours.
The clincher is exactly the multi-file future you raised: with N files at 10/40/90/100%, the parent can't show a determinate bar — no single percentage is correct. An indeterminate "I'm busy" is the only right parent-level visual, with each chip carrying its own bar. So keeping the spinner during progress isn't just acceptable, it's the design that scales. My suggestion would've left multi-file with no parent busy affordance at all — a step backward.

This also lines up with what I already posted: the review's Observability section calls the "busy signal driven by loading and progress" a positive. I stand by that over this new finding. The one real kernel — isLoading uses progress !== undefined (so 0 counts) while the chip uses !progress (so 0 doesn't draw) — isn't an inconsistency to fix; those are two correct answers to two different questions ("is work attached?" vs. "is there a percentage to paint?").

So: no code change. The finding is a mismatch between the reviewer's mental model and your (deliberate, forward-compatible) one.

If you want to preempt the next person raising it, the only cheap things worth doing — all optional:

A one-line comment on the getter documenting the aggregate-busy decision ("stays busy while a file is loading or has progress, incl. 0; completion is consumer-signalled, not progress === 100").
Optionally rename isLoading → isBusy to match "busy for whatever reason" and stop the name implying "indeterminate only."

jgroth (approved):

PR #4156, opened 2026-07-02
fix(table): preserve page and scroll position on data refresh

devbymadde:
## What & why

In remote mode, `limel-table` snapped a paginated list back to page 1 and lost its scroll position whenever the rows were rebuilt — for example when a row was deleted or two objects were merged while the user was on a later page (e.g. page 3 of a list). The root cause is that `replaceData()` resets Tabulator's internal pagination to the first page and scrolls the row container to top-left, and re-entering the ajax pipeline then emitted `changePage`/`load` for page 1 back to the consumer.

This regressed after the remote-paginator refresh was added, and affected any consumer of `limel-table` in remote mode (the object list in lime-webclient being the visible one).

## Changes

All changes are scoped to remote mode.

- Guard the internal `changePage`/`load` emissions in `requestData` and `handleDataSorting` while a paginator refresh is in flight, using a depth counter so overlapping refreshes (`totalRows` and `pageSize` changing in the same tick) can't clear the guard early. The returned `last_page` still updates the paginator UI.
- After `replaceData()`, restore Tabulator to the controlled `page` (clamped to the new page count) so `getPage()` and the paginator UI stay in sync. When the current page no longer exists (the last row on the last page was removed), the clamp is emitted as a genuine `changePage`.
- Snapshot and restore the scroll position around the row rebuilds, on both the data-change path and the paginator-refresh path.
- Skip the paginator-refresh round-trip entirely when the page count is unchanged (the common case, e.g. removing one row of many), which avoids two redundant full row rebuilds per data change.

## Testing
1. Go to an explorer with many pages 
2. Go to a page > 1 
3. Scroll down 
4. Delete one item
5. You should stay on the same page, scrolled to the same position

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented unnecessary page-change and data-load events during table refreshes.
  * Preserved the current page when rows are deleted or data totals change.
  * Correctly clamped pagination to the last available page when needed.
  * Preserved table scroll position during data updates and pagination refreshes.
  * Improved handling of asynchronous data replacement and sorting updates.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/table/table.tsx
+++ b/src/components/table/table.tsx
@@ -267,6 +267,26 @@ export class Table {
     private shouldSort = false;
     private hasWarnedOnConflictingMovableAndSortable = false;
 
+    /**
+     * Depth of in-flight `refreshRemotePaginator` round-trips. While
+     * `isRefreshingPaginator` is true, `requestData`/`handleDataSorting` must
+     * not emit `changePage`/`load`, since the round-trip is not a user-initiated
+     * page change and the consumer's page state is authoritative. A counter
+     * rather than a boolean so overlapping refreshes (e.g. `totalRows` and
+     * `pageSize` changing in the same tick) don't clear the guard early.
+     */
+    private paginatorRefreshDepth = 0;
+
+    private get isRefreshingPaginator(): boolean {
+        return this.paginatorRefreshDepth > 0;
+    }
+
+    /**
+     * Page count reflected by the paginator UI at the last refresh. Used to skip
+     * the (row-rebuilding) refresh round-trip when the count hasn't changed.
+     */
+    private lastPaginatorPageCount: number | undefined;
+
     constructor() {
         this.handleDataSorting = this.handleDataSorting.bind(this);
         this.handlePageLoaded = this.handlePageLoaded.bind(this);
@@ -364,7 +384,10 @@ export class Table {
 
             if (shouldReplace) {
                 this.pool.releaseAll();
-                this.tabulator.replaceData(newData);
+                const restoreScroll = this.preserveScroll();
+                this.tabulator
+                    .replaceData(newData)
+                    .then(restoreScroll, () => {});
                 this.setSelection();
 
                 return;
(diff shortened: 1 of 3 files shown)


FredrikWallstrom:
@coderabbitai review

src/components/table/table.tsx
```diff
@@ -364,7 +384,10 @@ export class Table {
 
             if (shouldReplace) {
                 this.pool.releaseAll();
-                this.tabulator.replaceData(newData);
+                const restoreScroll = this.preserveScroll();
```
FredrikWallstrom, line 387: Tested this branch in the webclient and found a regression: after creating a new record, it used to show up highlighted right away — now it's nowhere to be seen. The highlight itself still works (`activeRow` just sets a CSS class), but nothing actually scrolls to the row. What made new records visible before was the very snap-to-top this PR removes: the new record sorts to the top of the list, and the scroll reset happened to reveal it. Now the position is kept, so the record sits highlighted somewhere off-screen — or on a different page if you weren't on page 1.

src/components/table/table.tsx
```diff
@@ -364,7 +384,10 @@ export class Table {
...
+                this.tabulator
+                    .replaceData(newData)
```
FredrikWallstrom, line 389: Another thing I ran into while testing: deleting a row can make the table go white for a moment before the scroll position is put back. See video:


https://github.com/user-attachments/assets/217a69eb-a2f4-435f-bb4f-9375d99c6d01

src/components/table/table.tsx
```diff
@@ -751,8 +784,28 @@ export class Table {
...
+        ) ?? null) as HTMLElement | null;
+    }
+
+    /**
+     * Snapshot the current scroll position and return a callback that restores
+     * it. Tabulator resets the row container to top-left whenever it rebuilds
+     * all row elements (e.g. `replaceData`), so wrap such operations with this
+     * and invoke the returned callback once the rebuild has settled.
+     */
+    private preserveScroll(): () => void {
```
FredrikWallstrom, line 798: Deleting a row while scrolled all the way down leaves the table a bit above the bottom instead of staying at the bottom. Expected behavior is to stay at the bottom of the table, see video:

https://github.com/user-attachments/assets/a5da5799-e873-4beb-a672-f2fdd598eba1

src/components/table/table.tsx
```diff
@@ -725,23 +748,33 @@ export class Table {
...
+            this.paginatorRefreshDepth--;
         }
 
-        if (scrollContainer) {
-            scrollContainer.scrollTop = scrollTop;
-            scrollContainer.scrollLeft = scrollLeft;
+        this.lastPaginatorPageCount = maxPage;
+        restoreScroll();
+
+        if (targetPage !== this.page) {
```
FredrikWallstrom, line 776: There's a follow-up problem when the clamp kicks in: `currentLoad` never gets updated during the refresh (the guard skips it), so it still says page 3. If the user later navigates back to page 3, `requestData` builds the same load object again, `isEqual` matches, and no `load` event is emitted — the consumer never fetches that page's data. Resetting `currentLoad` when emitting the clamp would avoid that.

src/components/table/table.tsx
```diff
@@ -725,23 +748,33 @@ export class Table {
...
         } catch {
             return;
+        } finally {
+            this.paginatorRefreshDepth--;
         }
 
-        if (scrollContainer) {
-            scrollContainer.scrollTop = scrollTop;
-            scrollContainer.scrollLeft = scrollLeft;
+        this.lastPaginatorPageCount = maxPage;
```
FredrikWallstrom, line 773: This stores the page count that was computed before the awaits, and that value can be wrong: when `totalRows` and `pageSize` change in the same tick, the watchers fire one prop at a time, so the first refresh runs with half-updated props. If that in-between count gets stored here, a later change that happens to land on the same number is skipped, and the paginator keeps showing the old count. Rare in practice, but recomputing `calculatePageCount()` after the awaits would avoid it.

FredrikWallstrom:
Edqvist flagged to me that the table view in product.lime has been acting up in general. Turns out product.lime runs the `useDeferredTotals` flag, which drives the remote paginator refresh, and that's what's behind it. I've filed a separate bug for that here: https://github.com/Lundalogik/crm-client/issues/1210

Mostly a heads-up, since this PR touches the same remote paginator refresh. Thinking about it more broadly, I'm starting to wonder whether that refresh mechanism is quite right as-is, might be worth an overhaul rather than more point fixes.

PR #4158, opened 2026-07-02
ci: cache npm and Sonar downloads with a split restore/save strategy

adrianschmidt:
## What this does

Makes CI runs faster and less flaky by caching downloads that were previously re-fetched on every run, and by making the caches survive failed runs.

**The wins:**

- **The 8 container test jobs now cache npm packages.** The `test`, `runtime-tests` (2 shards), `accessibility-tests` (4 shards) and `component-tests` jobs run inside the Playwright image and had no caching at all — every PR push downloaded every package from the registry 8 times.
- **The npm cache now survives a failed job.** Previously (via setup-node's built-in cache) the cache was only saved if the *whole job* succeeded. So a PR that changed `package-lock.json` and failed lint re-downloaded everything on every push until the job went green. Now the cache is saved as soon as `npm ci` succeeds.
- **A lock-file change no longer means a full re-download.** The built-in cache only restored on an exact lock-file match. The new setup falls back to the newest previous cache, which still contains every package the change didn't touch.
- **The Sonar scanner and analyzers are now cached.** We have a history of intermittent scanner-download failures (binaries served from IPs on GitHub's blocklist), so a cache hit removes a flake source, not just ~30 s of overhead. The cache is saved even when the scan or quality gate fails — those failures say nothing about the downloaded files.

## How

A new `install-dependencies` composite action wraps `npm ci` in the split cache pattern: `actions/cache/restore` → `npm ci` → `actions/cache/save`. Saving in a regular step right after the install (instead of a success-gated post-job step) is what makes the cache survive later failures. All jobs — host and container — now use it; `set-up-node` keeps handling only the Node version. The Sonar jobs use the same split pattern inline, with a save step that is best-effort (`continue-on-error`) so a cache problem can never block a PR check or a release.

## Notes for reviewers

- When several jobs miss on the same key simultaneously they race to save it; the losers log an "Unable to reserve cache" warning and continue. Expected, harmless.
- setup-node v6 auto-enables its built-in cache when `package.json` has a `packageManager` field (ours does), so it's now explicitly disabled with `package-manager-cache: false`.
- The Sonar cache key hashes the two workflow files, so it rotates whenever the pinned scan action (and with it the scanner version) is bumped by Dependabot.

## Summary by CodeRabbit

* **New Features**
  * Improved dependency setup across CI and release workflows for more consistent and reliable builds.
  * Added smarter cache handling to speed up repeated runs and reduce unnecessary downloads.
  * Enabled shared caching for code quality checks to improve performance.

* **Bug Fixes**
  * Reduced cache mismatches between different runner environments, helping avoid setup issues.
  * Made cache saving best-effort so cache-related problems are less likely to interrupt jobs.

### DIFF
--- a/.github/actions/install-dependencies/action.yml
+++ b/.github/actions/install-dependencies/action.yml
@@ -1,0 +1,60 @@
+name: 'Install npm dependencies'
+description: >-
+  Runs `npm ci` wrapped in a split cache strategy (actions/cache/restore and
+  actions/cache/save as separate steps). Compared to setup-node's built-in
+  `cache: npm`, this (1) saves the cache as soon as the install has succeeded,
+  so it survives a later job step failing — setup-node only saves in a
+  post-job step gated on `post-if: success()` — and (2) uses restore-keys, so
+  a package-lock.json change gets a partial restore instead of a full registry
+  download (npm's cache is content-addressed, so a stale cache still serves
+  every package the change didn't touch). Also usable in container jobs that
+  have no setup-node at all (the Playwright-image test jobs).
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Resolve npm cache location
+      id: npm-cache-dir
+      shell: bash
+      # npm keeps its cache under $HOME, which differs between host runners
+      # (/home/runner) and the Playwright container jobs (/github/home).
+      # actions/cache archives ~/.npm with absolute paths, so a cache saved by a
+      # host job restores into a directory the container's npm never reads (and
+      # vice versa) — silently defeating the container-caching win. Cache the
+      # resolved directory and fold $HOME into the key so host and container
+      # each keep their own effective cache.
+      run: |
+        dir="$(npm config get cache)"
+        home_tag="${HOME//\//_}"
+        # Echo to the step log too — this step exists to disambiguate the
+        # host/container home split, so make a wrong home visible right here.
+        echo "npm cache dir=$dir home-tag=$home_tag"
+        echo "dir=$dir" >> "$GITHUB_OUTPUT"
+        echo "home-tag=$home_tag" >> "$GITHUB_OUTPUT"
+    - name: Restore npm cache
+      id: npm-cache
+      uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+      with:
+        path: ${{ steps.npm-cache-dir.outputs.dir }}
+        key: npm-cache-${{ runner.os }}-${{ runner.arch }}-${{ steps.npm-cache-dir.outputs.home-tag }}-${{ hashFiles('package-lock.json') }}
+        restore-keys: |
+          npm-cache-${{ runner.os }}-${{ runner.arch }}-${{ steps.npm-cache-dir.outputs.home-tag }}-
+    - name: Install dependencies
+      shell: bash
+      run: npm ci
+    - name: Save npm cache
+      # Save right here — after a successful install but before the job's real
+      # work — so the cache lands even if a later step fails. Skipped on an
+      # exact key hit (a saved cache is immutable, re-saving is a no-op). On a
+      # restore-keys (partial) hit, cache-hit is 'false' and the topped-up
+      # cache is saved under the new key. When several jobs miss on the same
+      # key at once, they race to save it: the winner uploads, the losers log
+      # an "Unable to reserve cache" warning and move on — harmless.
+      # continue-on-error keeps any other save failure (tar/upload error) from
+      # failing an otherwise-green job — a cache save should never break a check.
+      if: steps.npm-cache.outputs.cache-hit != 'true'
+      continue-on-error: true
+      uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+      with:
+        path: ${{ steps.npm-cache-dir.outputs.dir }}
+        key: ${{ steps.npm-cache.outputs.cache-primary-key }}
--- a/.github/actions/set-up-node/action.yml
+++ b/.github/actions/set-up-node/action.yml
@@ -23,7 +23,13 @@ runs:
     - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
       with:
         node-version: "${{ steps.nvm.outputs.NVMRC }}"
-        cache: 'npm'
+        # npm caching is handled by the install-dependencies action instead,
+        # which saves the cache as soon as npm ci succeeds; setup-node's
+        # built-in cache only saves when the whole job succeeds (post-if:
+        # success()) and restores on exact lock-file match only. Explicitly
+        # disabled here because package.json has a `packageManager` field,
+        # which would otherwise auto-enable the built-in cache.
+        package-manager-cache: false
     - name: Verify Node.js setup
       shell: bash
       run: echo "✅ Node $(node -v), npm $(npm -v)"
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -68,7 +68,7 @@ jobs:
       with:
         ref: ${{ inputs.ref }}
     - uses: ./.github/actions/set-up-node
-    - run: npm ci
+    - uses: ./.github/actions/install-dependencies
     - run: git config --global user.email "93315277+lime-opensource@users.noreply.github.com"
     - run: git config --global user.name "Lime Technologies OSS"
     - name: Publish docs
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -15,7 +15,7 @@ jobs:
       with:
         persist-credentials: false
     - uses: ./.github/actions/set-up-node
-    - run: npm ci
+    - uses: ./.github/actions/install-dependencies
     - run: git config --global user.email "93315277+lime-opensource@users.noreply.github.com"
     - run: git config --global user.name "Lime Technologies OSS"
     - name: Remove PR docs
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -292,11 +292,50 @@ jobs:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
         with:
           fetch-depth: 0
+      # Cache the Sonar downloads: the scan action installs the scanner CLI
+      # into the runner tool cache, and the scanner fills ~/.sonar/cache with
+      # the JRE and analyzers at scan time. Both downloads have been flaky in
+      # the past (scanner binaries served from IPs on GitHub's blocklist), so
+      # a cache hit also removes a failure source, not just ~30s of overhead.
+      # The key hashes both workflow files, so it rotates whenever either file
+      # changes (including a scan-action bump, which changes the scanner CLI
+      # version); between rotations the restore-keys prefix serves the newest
+      # saved cache. The hashFiles list must stay identical to the one in
+      # release.yml, or the shared cache silently splits in two.
+      # These caches hold executable content (scanner CLI, JRE, analyzer JARs)
+      # restored without integrity checks; that is safe only because this job's
+      # `head.repo.full_name == github.repository` guard keeps forks — i.e.
+      # untrusted actors — off the write side.
+      - name: Restore Sonar caches
+        id: sonar-cache
+        uses: actions/cache/restore@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: |
+            ~/.sonar/cache
+            ${{ runner.tool_cache }}/sonar-scanner-cli
+          key: sonar-cache-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('.github/workflows/pr-checks.yml', '.github/workflows/release.yml') }}
+          restore-keys: |
+            sonar-cache-${{ runner.os }}-${{ runner.arch }}-
       - name: SonarCloud Scan
         uses: SonarSource/sonarqube-scan-action@713881670b6b3676cda39549040e2d88c70d582e # v8.2.0
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+      # Split save (rather than plain actions/cache, whose post-job save is
+      # gated on job success) so the downloads are kept even when the scan
+      # fails — scan failures (quality gate, SonarCloud 403 flakes) say
+      # nothing about the cached files. Best-effort: never fail the job over
+      # a cache save (e.g. paths missing because the scan died before
+      # downloading anything).
+      - name: Save Sonar caches
+        if: ${{ !cancelled() && steps.sonar-cache.outputs.cache-hit != 'true' }}
+        continue-on-error: true
+        uses: actions/cache/save@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
+        with:
+          path: |
+            ~/.sonar/cache
+            ${{ runner.tool_cache }}/sonar-scanner-cli
+          key: ${{ steps.sonar-cache.outputs.cache-primary-key }}
   autosquash:
     name: Block Autosquash Commits
     runs-on: ubuntu-latest
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -67,7 +67,7 @@ jobs:
         persist-credentials: false
     # No set-up-node: the image bundles a Node whose major is asserted against
     # .nvmrc by verify-playwright-container (below), so npm ci runs under it.
-    - run: npm ci
+    - uses: ./.github/actions/install-dependencies
     # Fail fast if the pinned image has drifted from the project (Node major vs
     # .nvmrc, and the image's Chromium build vs @playwright/test). See the action.
     - uses: ./.github/actions/verify-playwright-container
@@ -129,7 +129,7 @@ jobs:
         persist-credentials: false
     # No set-up-node: the image bundles a Node whose major is asserted against
     # .nvmrc by verify-playwright-container (below), so npm ci runs under it.
-    - run: npm ci
+    - uses: ./.github/actions/install-dependencies
     # Fail fast if the pinned image has drifted from the project (Node major vs
     # .nvmrc, and the image's Chromium build vs @playwright/test). See the action.
     - uses: ./.github/actions/verify-playwright-container
@@ -184,7 +184,7 @@ jobs:
         persist-credentials: false
     # No set-up-node: the image bundles a Node whose major is asserted against
     # .nvmrc by verify-playwright-container (below), so npm ci runs under it.
-    - run: npm ci
+    - uses: ./.github/actions/install-dependencies
     # Fail fast if the pinned image has drifted from the project (Node major vs
     # .nvmrc, and the image's Chromium build vs @playwright/test). See the action.
     - uses: ./.github/actions/verify-playwright-container
@@ -238,7 +238,7 @@ jobs:
         persist-credentials: false
     # No set-up-node: the image bundles a Node whose major is asserted against
     # .nvmrc by verify-playwright-container (below), so npm ci runs under it.
-    - run: npm ci
+    - uses: ./.github/actions/install-dependencies
     # Fail fast if the pinned image has drifted from the project (Node major vs
     # .nvmrc, and the image's Chromium build vs @playwright/test). See the action.
     - uses: ./.github/actions/verify-playwright-container
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
     - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
     - uses: ./.github/actions/set-up-node
-    - run: npm ci
+    - uses: ./.github/actions/install-dependencies
     - run: npm run lint
 
   build:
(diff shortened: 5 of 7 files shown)


adrianschmidt:
# Consolidated PR Review

*Multi-agent review (7 dimensions: backward compatibility, code quality, architecture, security, observability, performance, Lime platform), synthesized.*

## PR Summary

Adds a split `actions/cache/restore` → `npm ci` → `actions/cache/save` strategy via a new `install-dependencies` composite action (replacing setup-node's built-in success-gated cache and giving the 8 container test jobs caching for the first time), and caches the Sonar scanner CLI + JRE/analyzers in the `pr-checks.yml` and `release.yml` sonarcloud jobs with a best-effort save. 7 files changed, +115/−12, all in `.github/`.

The PR introduces no regression relative to `main` — worst case any cache misses and behavior degrades to today's plain download. It is blocked by nothing; the caveat is that one `[Medium]` finding (host/container `~/.npm` path mismatch) may silently prevent the PR's headline benefit from materializing for the container jobs, so it's worth verifying before or shortly after merge.

- **Blockers:** None — the PR is a net improvement over the target branch across all seven dimensions.
- **Non-blocking but worth addressing:** See Top Recommendations — notably the container-vs-host npm cache path question (Performance) and the cross-workflow Sonar cache key coupling (Architecture).

---

Additive caching plumbing: no public workflow contract (composite action inputs/outputs, `workflow_call` signatures, job names) changes, and every consumer of the changed pieces was migrated in lockstep.

- **[Low]** The `install-dependencies` save step lacks `continue-on-error` while the two Sonar save steps have it (`.github/actions/install-dependencies/action.yml:31-38` vs `pr-checks.yml:326-334`, `release.yml:57-66`). In practice `actions/cache/save@v6` only warns on the realistic failure modes (reserve race, tar/upload errors), so it won't actually fail a previously-passing job — but the author's own "cache saves should never fail a check" decision isn't encoded on this path. Worth adding `continue-on-error: true` for parity.
- **Positives:** Clean, consistent migration — every `npm ci` call site converted, fork-PR and `pull_request_target` guards untouched, no job names or interfaces altered.

## 2. Code Quality -- GOOD ✅

Expressions, output names (`cache-primary-key`, `cache-hit`), `hashFiles` usage, and `if:` conditions are all correct, and the comments' description of cache semantics matches actual `actions/cache` behavior — with one overstatement.

- **[Low]** The Sonar-cache comment overstates the key-rotation trigger (`pr-checks.yml:300-302`, mirrored in `release.yml:30-33`): the key hashes the *entire* two workflow files, so it rotates on *any* edit to either file (adding a job, bumping `actions/checkout`, editing a comment), not just a scan-action bump. Consequence is benign (`restore-keys` serves the newest cache; one redundant re-save), but the comment should say "whenever either workflow file changes" — or the key should hash something narrower, e.g. the pinned scan-action SHA. *(Also flagged by Architecture, Observability, and Performance.)*
- **[Low]** The Sonar cache key omits `runner.arch` while the npm key includes it (`pr-checks.yml:310` / `release.yml:41` vs `install-dependencies/action.yml:21`). The cached JRE and scanner binary are architecture-specific; harmless while all runners are x64, but adding `${{ runner.arch }}` makes it consistent and ARM-proof.
- **Positives:** Unusually thorough and accurate comments about non-obvious cache mechanics; the composite action removes real duplication across 8+ jobs.

## 3. Architecture -- GOOD ✅

Clean separation — `set-up-node` owns the Node version, `install-dependencies` owns install + cache — applied uniformly to every `npm ci` site, host and container.

- **[Medium]** The Sonar cache key duplicates a `hashFiles(...)` list across two files that must stay byte-identical for the shared cache to work (`pr-checks.yml:310`, `release.yml:41`). Nothing enforces the sync: a future edit to one list and not the other silently splits the cache — a quiet, gracefully-degrading failure invisible from either file alone. Add a cross-reference comment in each file ("must match release.yml / pr-checks.yml"), which also mitigates the drift risk Code Quality noted.
- **[Low]** The abstraction asymmetry (npm caching = composite action, Sonar caching = inline duplication) is *justified*: the Sonar restore/save must bracket the caller-owned scan step, which a composite action can't wrap. Flagged so a later reviewer doesn't "DRY it up" and break the restore/scan/save ordering — a keep-in-sync comment is the right fix, not extraction.
- **Positives:** Adding a new host or container job now follows one obvious pattern; host/container npm cache sharing is a nice emergent property (but see Performance on whether it holds in practice).

## 4. Security -- GOOD ✅

Cache-poisoning paths were traced against GitHub's cache scoping rules: PR-scoped writes can't reach `main`-scoped restores, `release.yml` and the `workflow_run`-triggered `publish-fork-pr-docs.yml` restore only `main`-scope caches, fork PRs never run the Sonar job (`head.repo.full_name == github.repository` guard), and `npm ci` independently verifies every package against `package-lock.json` integrity hashes. All new actions are pinned to full SHAs. No new secrets exposure.

- **[Low]** The Sonar caches restore *executable* content (scanner CLI, JRE, analyzer JARs) with no integrity verification on restore (`pr-checks.yml:303-332`, `release.yml:34-67`), unlike npm's content-addressed store. This is safe today only because the fork guard keeps untrusted actors off the write side. Worth a comment documenting that assumption so a future loosening of the fork guard doesn't silently turn the cache into a code-execution vector.
- **Positives:** SHA-pinned actions throughout; the `workflow_run` workflow checks out `main` and treats fork input as data-only; npm integrity checking backstops the npm cache independently of scope.

## 5. Observability -- GOOD ✅

Cache hit/miss is visible by default in the `actions/cache` log lines, the save-race warning is pre-documented in the code, and step names give good debugging context.

- **[Low]** The best-effort Sonar saves have no active degradation signal (`pr-checks.yml:324-332`, `release.yml:59-67`): with `continue-on-error: true`, a *persistently* failing save never fails a check, so the flake protection could silently never materialize — the only trace is a warning inside the step log. The not-blocking decision is documented and sound; a one-line job-summary note on save failure would close the gap. Acceptable to ship without it.
- **Positives:** The npm save step is *not* `continue-on-error`, so genuine npm-cache save failures stay visible; the "Unable to reserve cache" race is explained in the code where a maintainer will look.

## 6. Performance -- GOOD ✅

The strategy is sound and the dependency tree (1556 resolved packages, ~700 KB lock file) is large enough for the wins to be real. One genuine risk to the headline claim:

- **[Medium]** Host and container jobs share one cache key but resolve `~/.npm` to different homes (`install-dependencies/action.yml:19-21`): host runners have `HOME=/home/runner`, the Playwright-container jobs run as root with `HOME=/root`. The saved archive bakes in whichever path the save-race winner used; a job with the other home may restore files npm never looks at and re-download from the registry anyway — silently defeating the container-caching win (the PR's headline) for roughly half the jobs, while still paying restore/save cost. Not worse than target (which had no container cache), so not a blocker — but verify a container job's log shows npm actually *using* a restored cache, and if not, include the home path in the key or pin `npm config set cache` to a shared relative path.
- **[Low]** On a lock-file change, the losing jobs in the save race still build the full `~/.npm` tarball on the critical path before discarding it. Bounded (only on miss/partial-hit runs) and small next to multi-minute test shards; not worth changing.
- **[Low]** The in-step save puts tar+upload on the critical path for host jobs that previously saved post-job — the deliberate, documented trade-off that buys failure-survival; only bites on miss runs.
- **Positives:** 10 GB eviction budget is comfortably sufficient (one hot npm entry shared by all jobs, not per-job); exact-hit runs skip the save entirely; the 8-way save race wastes no upload volume (only the winner uploads).

Catalogue reachable (index via internal-docs MCP, `lime-issue-detector` SKILL.md via `gh api`); all 11 rule categories triaged against the changed files. **No findings** — the LPID catalogue targets Lime CRM platform-runtime and Stencil component-runtime hazards, and this PR changes only CI configuration. The two plausibly-relevant categories were explicitly checked: no hardcoded secrets (LPID-SEC-1/RH-1 — every token reference is a `secrets.*` binding), and the FE category doesn't reach `.github/`.

- **Positives:** Pinned SHAs, correct `cache-hit` gating, and documented rationale for disabling setup-node's built-in cache.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Code Quality | **GOOD** |
| Architecture | **GOOD** |
| Security | **GOOD** |
| Observability | **GOOD** |
| Performance | **GOOD** |
| **Merge Readiness** | **MERGE WITH CAVEATS** |

## Top Recommendations

1. **[Medium from Performance]** Verify the container jobs actually consume the shared npm cache: check a Playwright-container job log for a restore that `npm ci` then uses (no registry re-download). If the `/home/runner` vs `/root` home mismatch defeats it, add the home path (or an explicit `npm config set cache` location) to the design.
2. **[Medium from Architecture]** Add a cross-reference comment next to each Sonar `hashFiles(...)` key (`pr-checks.yml:310`, `release.yml:41`): "this list must stay identical to the one in <other file>" — the shared cache silently splits if they drift.
3. **[Low from Code Quality]** Fix the key-rotation comment (`pr-checks.yml:300-302`, `release.yml:30-33`) to say the key rotates on *any* edit to either workflow file — or narrow the key to the pinned scan-action version.
4. **[Low from Code Quality]** Add `${{ runner.arch }}` to the Sonar cache key for consistency with the npm key and ARM-runner future-proofing.
5. **[Low from Backward Compatibility]** Add `continue-on-error: true` to the `install-dependencies` save step for parity with the Sonar saves' "never fail a check over a cache save" policy.
6. **[Low from Security]** Document (in a comment) that the Sonar binary cache's safety depends on the fork guard (`head.repo.full_name == github.repository`) keeping untrusted actors off the write side.
7. **[Low from Observability]** Optionally emit a job-summary line when a Sonar cache save fails, so persistent save failures don't degrade silently.

adrianschmidt:
Addressed the review feedback. Two fixup commits pushed:

**`694e6d5a4`** — fixes the one finding that actually mattered. The container test jobs weren't really getting the npm cache: I checked a real run and the cache was being unpacked into the host's home directory (`/home/runner`), but inside the Playwright containers npm looks in a different home (`/github/home`), so it re-downloaded everything anyway. The fix keys the cache to the actual home, so host jobs and container jobs each keep their own working cache. Same commit also adds `continue-on-error` to the npm cache-save step, so a failed cache save can never fail an otherwise-green check (matching the Sonar cache-save steps).

**`d44841e4b`** — small correctness/clarity touches on the Sonar cache:
- Cross-reference comments in both workflow files noting the cache-key list must stay identical between them (if they drift, the shared cache silently splits).
- Corrected a comment that overstated when the key changes (it changes on *any* edit to either workflow file, not only a scanner-version bump).
- Added the runner architecture to the key (consistency with the npm key; future-proofs against ARM runners).
- Documented that the Sonar binary cache is safe only because forks can't write to it.

**Not changed:** the suggested job-summary line on a Sonar cache-save failure. The review itself flagged this as optional and fine to skip — a persistent save failure would still show as a warning in the step log, and wiring a summary around a `continue-on-error` step adds fiddly logic for little gain.

The "Block Autosquash Commits" check will go red because of the fixup commits — that's expected in this workflow and will resolve on squash.

adrianschmidt:
# Consolidated PR Review — Iteration 2 (after fixups `694e6d5a4`, `d44841e4b`)

*Multi-agent re-review (7 dimensions), focused on verifying the fixes from the previous review round.*

## PR Summary

Same scope as before — split restore/save npm caching via a new `install-dependencies` composite action plus Sonar scanner/analyzer caching in the two sonarcloud jobs — now with two fixup commits: `694e6d5a4` resolves the npm cache directory per-job (`npm config get cache`) and folds `$HOME` into the cache key so host (`/home/runner`) and container (`/github/home`) jobs each keep a working cache, and adds `continue-on-error` to the npm save; `d44841e4b` adds `runner.arch` to the Sonar keys, keep-in-sync cross-reference comments, the corrected key-rotation wording, and the fork-guard security comment.

Both `[Medium]` findings from the previous round are verified fixed, and no dimension found a regression relative to `main`. Every remaining finding is a `[Low]` nice-to-have or an explicitly documented trade-off. The PR is strictly better than the target branch. (The red "Block Autosquash Commits" check is the expected consequence of the fixup-commit workflow, to be resolved by squashing before merge.)

- **Blockers:** None.
- **Non-blocking but worth addressing:** Two small `[Low]` suggestions in Top Recommendations; both are optional.

---

## 1. Backward Compatibility -- SAFE ✅

The fixups were verified in every context the composite action is used (7 pr-checks jobs, build-docs, cleanup, release, fork-PR docs): `npm` is guaranteed on PATH before the resolve step in all of them, the fork-PR docs job gates the whole composite on one `if` (never half-runs), no repo `.npmrc` overrides the cache dir, and the release pipeline's pass/fail semantics are provably unchanged (Sonar restore is non-fatal on miss; both saves are best-effort).

- **[Low]** `install-dependencies` puts a resolve step + cache restore in front of `npm ci` on the release/build path; a hard failure there would now fail jobs where a bare `npm ci` previously ran. Probability is very low and this is the PR's deliberate design — not a blocker.
- **Positives:** The home-tag partition is exactly the right fix for the container-cache miss; all new save steps are best-effort, so no cache operation can break a green check or block a release.

## 2. Code Quality -- SAFE ✅

The fixup code was verified in detail: the resolve step's bash is correct (`$(…)` strips the newline; `npm config get cache` prints only the path to stdout), the two Sonar keys are **byte-identical** across `pr-checks.yml:316` and `release.yml:43` (same indent, arg order, spacing), and all reworded/added comments accurately describe actual `actions/cache` behavior.

- **[Low]** `home-tag` sanitizes only `/`, and an unset `$HOME` would produce an empty key segment — both non-issues on GitHub runners, where `$HOME` is always a clean absolute path.
- **[Low]** The Sonar key still rotates on any edit to either workflow file (more churn than the scanner version warrants), but the comment now says so accurately and `restore-keys` cushions each rotation — a documented design choice, not a defect.
- **Positives:** Empirically verified fixes; comments now match behavior precisely; keys and paths consistent across files and steps.

## 3. Architecture -- GOOD ✅

Per-home cache segregation was confirmed as the right call — `actions/cache` archives absolute paths (`tar -P`), and even a "workspace-relative" shared location wouldn't work since `$GITHUB_WORKSPACE` also differs between host and container. "Share one cache" was never cleanly achievable; the chosen design costs only a second cache entry per lock-hash.

- **[Low]** The key discriminator is `$HOME` while restore correctness is actually governed by the resolved cache dir — two derivations of the same fact that diverge only under an exotic `.npmrc` `cache=` override. Folding the sanitized resolved dir into the key instead would remove the double derivation. Speculative in this environment.
- **[Low]** The two-file Sonar key coupling is now documented (cross-reference comments) but not mechanically enforced; a restore/save composite-action pair could DRY the key into one place, at the cost of more indirection for a two-site case. Documentation over extraction is a reasonable trade-off — downgraded from the previous round's `[Medium]` now that it's documented.
- **Positives:** Correct root-cause fix verified against a real run; the npm pattern remains a single obvious drop-in applied uniformly across all 11 call sites.

## 4. Security -- SAFE ✅

The only fork-influenceable addition — the resolve step reading a checkout-controlled `.npmrc` — was analyzed for injection: a malicious fork `.npmrc` can only redirect cache path/key within the fork's *own* PR-scoped cache on its *own* ephemeral runner (which already executes fork lifecycle scripts during `npm ci`), and the step outputs are never interpolated into a shell — no path to code execution or cross-scope poisoning. The privileged `workflow_run` docs job checks out `main`, so no fork `.npmrc` reaches it. The previous `[Low]` (undocumented fork-guard assumption) is closed by the new comment.

- No findings.
- **Positives:** Fork guard now self-documents its security role; all actions SHA-pinned; fork-controllable inputs stay inside the attacker's own unprivileged scope.

## 5. Observability -- GOOD ✅

The fixups improve diagnosability: the fully-interpolated keys (including the new `home-tag` and `arch`) appear in the restore-step logs, and the updated comments give accurate debugging context.

- **[Low]** The "Resolve npm cache location" step (`install-dependencies/action.yml:16-28`) writes its outputs only to `$GITHUB_OUTPUT`, so its own log body is empty even though the step exists precisely to disambiguate the host/container home split. A one-line `echo` of `dir` and `home-tag` to stdout would make a wrong-home situation self-evident at the point of resolution. Enhancement, not a regression — the information is recoverable from the next step's log today.
- **Positives:** Step names unambiguous; interpolated cache keys surface in restore logs, making key/home mismatches debuggable from a run log alone.

## 6. Performance -- GOOD ✅

The fix was reasoned through end-to-end and confirmed to make the container cache actually work: all 8 container jobs share `HOME=/github/home`, so they now converge on one usable cache entry, and the home-tagged `restore-keys` prefix prevents cross-home fallback to an archive npm would never read. The doubled storage (2 entries per lock-hash) is comfortably within the 10 GB budget with graceful partial-restore degradation on eviction.

- **[Low]** On cache-miss/lock-change runs, the jobs that lose the save race still build a discarded tarball before the reservation fails. Bounded, documented as the expected race, inherent to the split-save design — noted, not actionable.
- **Positives:** The home-tag fix is precisely targeted and demonstrably revives the previously-dead container cache; the critical-path save remains a well-reasoned failure-survival trade-off.

Catalogue reachable and re-triaged against the iteration-2 diff. **No findings** — the change remains 100% CI YAML with no Lime runtime or Stencil component surface for any LPID rule to bind to, confirming the previous round's result.

- **Positives:** The change is confined to CI infrastructure and carries no Lime platform runtime risk.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Code Quality | **SAFE** |
| Architecture | **GOOD** |
| Security | **SAFE** |
| Observability | **GOOD** |
| Performance | **GOOD** |

## Top Recommendations

No `[Critical]`, `[High]`, or `[Medium]` findings remain. Two optional `[Low]` suggestions:

1. **[Low from Observability]** Add a one-line `echo "npm cache dir=$dir home-tag=$home-tag"` to the resolve step so the host/container disambiguation is visible in its own step log.
2. **[Low from Architecture]** Consider deriving the npm cache key's discriminator from the sanitized resolved cache dir rather than `$HOME`, removing the double derivation of the same fact. Speculative benefit; fine to skip.

The remaining `[Low]` notes (Sonar key churn on workflow edits, two-file key coupling, discarded-tar race cost, resolve-step-before-`npm ci` on the release path) are documented trade-offs requiring no action.

adrianschmidt:
Round 2 of feedback — one small fix, one declined.

**Fixed (`9028ff7a1`):** The "Resolve npm cache location" step now echoes the resolved cache directory and home tag to its own log. That step exists to tell host and container jobs apart, but it was writing only to the step's outputs, so its log body was empty — now a wrong-home situation is visible right where it's resolved.

**Not changed:** the suggestion to key the cache on the resolved directory instead of `$HOME`. The two only ever differ under an unusual npm config that this repo doesn't use, and switching would reset the cache one more time (a one-off cold run) for no real-world benefit. The review flagged it as speculative and fine to skip.

adrianschmidt:
# Consolidated PR Review — Iteration 3 (after fixup `9028ff7a1`)

*Multi-agent re-review (7 dimensions). The only change since the iteration-2 review is commit `9028ff7a1`, a 7-line refactor of the "Resolve npm cache location" step implementing the previous round's observability suggestion (the resolved cache dir and home-tag are now echoed to the step log).*

## PR Summary

Unchanged scope: split restore/save npm caching via the `install-dependencies` composite action (host + container jobs, per-home cache keys) and Sonar scanner/JRE/analyzer caching in the two sonarcloud jobs. Six commits on the branch: three original + three fixups.

**No findings in any dimension.** Every reviewer independently verified that `9028ff7a1` is a pure diagnostic refactor: the `dir`/`home-tag` outputs are byte-identical to before (no cold-cache miss introduced), keys/paths/save-conditions unchanged, correct bash quoting throughout, and the echoed values create no injection surface (the runner's `::set-output::`/`::save-state::` stdout commands are disabled, and the step runs with no secrets in the attacker-reachable context). The PR is strictly better than `main`.

- **Blockers:** None.
- **Non-blocking but worth addressing:** Nothing remains. The previously declined items (Sonar save-failure job summary; keying on resolved dir instead of `$HOME`) stay declined — no reviewer found a new argument to reopen them.

*(The red "Block Autosquash Commits" check is the expected consequence of the fixup-commit workflow; squash before merge as usual.)*

---

## 1. Backward Compatibility -- SAFE ✅
No significant issues found. Outputs, contracts, and all call sites verified unchanged; `bash -eo pipefail` fail-fast semantics preserved.
**Positives:** Every workflow that passed iteration 2 passes identically now.

## 2. Code Quality -- SAFE ✅
No significant issues found. Line-by-line bash verification: all expansions double-quoted, no word-splitting risk, emitted outputs byte-for-byte identical.
**Positives:** The refactor improves diagnosability at zero behavioral cost, with a comment explaining why the echo exists.

## 3. Architecture -- GOOD ✅
No significant issues found. No design property changed; the action remains a single self-contained caching primitive with an unchanged two-output contract.
**Positives:** Observability added at exactly the point where the host/container home ambiguity would bite.

## 4. Security -- SAFE ✅
No findings. The new stdout echo was analyzed for workflow-command/log injection: `::set-output::`/`::save-state::` are disabled in the current runner, remaining log commands only affect the attacker's own unprivileged, secret-free PR run; the `$GITHUB_OUTPUT` posture is unchanged from the iteration-2 SAFE analysis.
**Positives:** No new trust flow; all actions remain SHA-pinned.

## 5. Observability -- GOOD ✅
No significant issues found. The echo closes the exact gap flagged in iteration 2 — visible, correctly labeled (`dir` + `home-tag`), and at the right point (before restore/save, where the home split originates).
**Positives:** Diagnosability is now good end-to-end: resolve step shows the home decision; `actions/cache` logs hit/miss/save outcomes.

## 6. Performance -- GOOD ✅
No significant issues found. No added subprocess, keys/paths/conditions unchanged, existing caches still hit (no cold-start introduced).
**Positives:** The caching design's wins remain intact.

Catalogue reachable; re-triaged the delta. No findings — still zero Lime runtime surface, consistent with iterations 1 and 2.
**Positives:** CI-only change with no Lime platform runtime risk.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Code Quality | **SAFE** |
| Architecture | **GOOD** |
| Security | **SAFE** |
| Observability | **GOOD** |
| Performance | **GOOD** |

## Top Recommendations

None — no `[Critical]`, `[High]`, `[Medium]`, or actionable `[Low]` findings remain. The review/fix loop has converged: 8 of 9 findings across three rounds were fixed (2 Medium, 6 Low), and the 2 declined suggestions were reasoned declines the reviews endorsed as optional.

adrianschmidt:
Re-review came back clean across all seven dimensions with zero findings — verdict is Ready to Merge. No further changes needed; the two previously declined suggestions stay declined (no new argument to reopen them). The only remaining red check is Block Autosquash Commits, which is expected from the fixup-commit workflow and clears on squash.

PR #4160, opened 2026-07-07
File image resize

Befkadu1:
fix https://github.com/Lundalogik/lime-webclient/issues/7902

## Summary by CodeRabbit

* **New Features**
  * Added optional client-side image resizing to the file picker (dimensions, fit, output type, JPEG quality) with in-progress UI and responsive preview.
  * Added new image resize example(s), including a mixed best-effort scenario.
* **Bug Fixes**
  * Improved resizing feedback with an “optimizing” state, synchronized selection, and safer handling when users cancel mid-process.
  * If resizing fails, the original file is emitted unchanged.
* **Documentation**
  * Added “optimizing” progress text translations for supported languages.
* **Tests**
  * Added end-to-end and unit coverage for the image resize behavior.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/file/examples/file-resize-image.scss
+++ b/src/components/file/examples/file-resize-image.scss
@@ -1,0 +1,33 @@
+@use '../../../style/mixins';
+
+:host {
+    display: flex;
+    flex-direction: column;
+    gap: 1rem;
+}
+
+limel-example-controls {
+    --example-controls-max-columns-width: 10rem;
+    gap: 1rem;
+    padding-top: 0.5rem;
+    margin-top: 0;
+}
+
+h4 {
+    font-weight: 600;
+    margin: 0;
+}
+
+.preview {
+    min-height: 10rem;
+    border-radius: 0.5rem;
+    padding: 1rem;
+
+    text-align: center;
+    @include mixins.add-chessboard-background();
+
+    img {
+        box-shadow: var(--shadow-depth-16), var(--shadow-depth-64);
+        max-width: 100%;
+    }
+}
--- a/src/components/file/examples/file-resize-image.tsx
+++ b/src/components/file/examples/file-resize-image.tsx
@@ -1,0 +1,239 @@
+import {
+    FileInfo,
+    LimelSelectCustomEvent,
+    Option,
+    type ResizeOptions,
+} from '@limetech/lime-elements';
+import { Component, h, Host, State } from '@stencil/core';
+
+/**
+ * Client-side image resize
+ *
+ * When `resizeImage` is set, a selected image is downscaled and re-encoded on the
+ * user's device *before* the `change` event fires — reducing upload size and
+ * normalizing its dimensions and format. The emitted file's `filename`,
+ * `contentType` and `size` reflect the resized file, so its extension may
+ * change (for example `.png` → `.jpg`).
+ *
+ * The options passed to `resizeImage` are:
+ *
+ * - `width` / `height`: the target box, in pixels, to fit the image into.
+ *   Clear one to derive it from the other and keep the aspect ratio; clear
+ *   both to keep the source dimensions and only re-encode.
+ * - `fit`: only matters when both `width` and `height` are set. `cover` fills
+ *   the box and center-crops the overflow; `contain` fits the whole image
+ *   inside the box, letterboxing it. Omit it for the default (`cover`).
+ * - `type`: the output format. Omit it for the default (`image/jpeg`).
+ * - `quality`: JPEG compression, `0`–`1`. Ignored for `image/png`, which is
+ *   lossless. The control below is disabled when PNG is selected.
+ *
+ * Because resize re-encodes the image, do not enable it where the original
+ * bytes must be preserved — for example security scanning, checksums, or
+ * signatures.
+ *
+ * Resize completes before `change` fires; reflect any subsequent upload with
+ * the file's `loading`, `statusText` and `progress` fields.
+ *
+ * :::note
+ * `limel-file` is a general-purpose file picker, so pair `resizeImage` with an
+ * image-restricted `accept`, as this example does. Resize only touches
+ * decodable raster images; see the next example for what happens to everything
+ * else.
+ * :::
+ *
+ * :::tip
+ * If you are specifically building an image or avatar picker, look at
+ * [limel-profile-picture](#/component/limel-profile-picture/) instead. It
+ * shares the same resize options, but additionally renders a preview of the
+ * chosen image inline, and its examples explore the `fit` strategies and
+ * fallback behavior in more depth.
+ * :::
+ *
+ * @beta
+ */
+@Component({
+    tag: 'limel-example-file-resize-image',
+    shadow: true,
+    styleUrl: 'file-resize-image.scss',
+})
+export class FileResizeImageExample {
+    @State()
+    private value?: FileInfo;
+
+    @State()
+    private previewUrl = '';
+
+    @State()
+    // Every field is optional. When omitted, `fit` defaults to `cover`, `type`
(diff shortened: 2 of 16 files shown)


src/components/file/file.tsx
```diff
@@ -201,33 +245,79 @@ export class File {
...
+            this.isResizableImage(content)
+        ) {
+            // Show the selected file with a loading badge while it resizes.
+            this.resizingFile = {
+                ...file,
+                loading: true,
+                statusText: this.getTranslation('file.optimizing'),
+            };
+            try {
+                const processed = await resizeImage(content, this.resizeImage);
```
Befkadu1: **Naming collision (readability).** The imported util `resizeImage` and the `@Prop() resizeImage` share the same identifier, disambiguated only by `this.`. `await resizeImage(content, this.resizeImage)` reads ambiguously at a glance. Consider aliasing the import (e.g. `import { resizeImage as resizeImageFile }`) so the function and the prop are visually distinct.

src/components/file/file.tsx
```diff
@@ -201,33 +245,79 @@ export class File {
...
+            } catch {
+                // Best-effort: on any decode or encode failure emit the
+                // original file unchanged.
+            }
+
+            // Keep the result on the same chip until `value` catches up.
+            this.resizingFile = out;
+        }
+
+        this.change.emit(out);
```
Befkadu1, line 308: **Possible stale re-emit.** After the `await`, `resizingFile = out` / `change.emit(out)` run unconditionally, with no check that the selection is still current. During resize the dropzone and file-input are disabled, but the chip-set isn't (only gated by `this.disabled`) and a `loading` chip keeps `removable: true` — so a user can remove the chip mid-resize, which emits `null` and clears state, and then the in-flight resize completes and re-shows/re-emits the removed file. Narrow window (resize is usually fast), but capturing a token before the `await` and bailing if it changed would close it.

src/components/file/examples/file-resize-image.tsx
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        if (!this.value?.fileContent) {
+            return;
+        }
+
+        this.previewUrl = URL.createObjectURL(this.value.fileContent);
+    };
+
+    private getSelectedFit = (): Option => {
```
Befkadu1: **Nit.** `getSelectedFit` / `getSelectedType` are annotated `(): Option`, but `Array.find` can return `undefined`. Harmless here since a match always exists, but `Option | undefined` would be the accurate return type.

Befkadu1 (commented):
## Review notes

A few findings from reviewing the resize integration. The two on `file.tsx` are worth a look; the rest are minor. Inline comments below.

**One finding that isn't on a changed line, so noting it here:** `renderDragAndDropTip()` (`src/components/file/file.tsx:230`) still guards on `this.value`, while the rest of the render path moved to `displayedFile`. In the window after a resize succeeds (`resizingFile` set, `isBusy` false) but before the consumer commits `value`, the "Drag & drop your file here" tip can render alongside the resized chip. Switching that guard to `this.displayedFile` keeps it consistent. Only reproduces when the consumer commits `value` asynchronously.

src/components/file/examples/file-resize-image.scss
```diff
@@ -0,0 +1,34 @@
...
+    --example-controls-max-columns-width: 10rem;
+    gap: 1rem;
+    padding-top: 0.5rem;
+    margin-top: 0;
+}
+
+h4 {
+    font-weight: 600;
+    margin-bottom: 0.75rem;
+    margin: 0;
```
Befkadu1, line 18: Overridden property "margin-bottom" by shorthand "margin"

Befkadu1 (commented):
## PR Review: File image resize

**Intent**: Add optional client-side image resizing to `limel-file` (downscale + re-encode before `change` fires), with an "Optimizing…" progress state and graceful fallback for non-decodable files. Fixes lime-webclient#7902.
**Scope**: +599/−32 across 14 files (1 core file, 1 util, 2 examples, 8 translations, 1 API report).

### Findings

**Backward compatibility**: ✅ — no breaks introduced.
- `ResizeOptions.width`/`height` go required → optional (`image-resize.ts:113-124`). Widening an input object is compatible for existing callers, and the type is `@beta`. `resizeImage(file, options)` signature is unchanged.
- New `resizeImage` prop is additive/optional; `file.optimizing` added to all 8 locales; `etc/lime-elements.api.md` regenerated to match.
- One behavior change beyond the title's scope: `clearAllButton={false}` (`file.tsx:365`) removes the clear-all button for **every** `limel-file` consumer, not just resize users. It's a reasonable call for a single-file picker (the chip's own × remains), but confirm you intend it to land in this PR rather than a standalone one.

**Code quality**: ⚠️ — the feature's own tests were lost, and a comment now misdescribes the code.
- `file.tsx:180-182`: the `@Watch` comment says it keeps "the same chip mounted across the hand-off (see file.spec.tsx) so only its badge changes instead of the field flashing empty." But `file.spec.tsx` isn't in the PR (dropped during rebase), and the mount-stability change it documented is also gone — `renderChipset` still returns the bare chip-set when `value` is set (`file.tsx:378-380`), so the wrapper→bare swap on commit still remounts and can flash. The comment describes behavior the code no longer implements. Worth checking: restore the wrapper fix + spec, or correct the comment.
- No automated coverage for the new async logic (resize, the token-guard bail at `file.tsx:300`, removal-cancels-resize at `file.tsx:402`). This logic had real bugs during development; it's exactly what a unit test should pin.
- Minor: `updateNumericOption`'s `key` union includes `'quality'`, which never flows through it (`examples/file-resize-image.tsx`).

**Architecture**: ✅ — resize isolated in a dependency-free util; component holds only the transient `resizingFile` state + `displayedFile` derivation. Clean separation, no new cross-cutting mechanisms.

**Security**: ✅ — purely client-side canvas work; no injection surface, no `innerHTML`, no secrets. Object URLs are revoked in the example (`componentWillUnload` + on change). The re-encode caveat (original bytes destroyed → bad for checksums/scanning) is explicitly documented.

**Observability**: ✅ (mostly N/A for a UI lib) — the `catch {}` at `file.tsx:293` swallows resize failures silently, but that's the documented "best-effort → emit original" contract, and the examples show how consumers detect a skipped resize. `image-resize.ts` has a dev-only gated `console.debug` on the decode fallback — acceptable.

**Performance**: ⚠️ — worth knowing, not blocking.
- Resize runs on the main thread (`OffscreenCanvas` when available but no worker), so a large image briefly blocks the UI; mitigated by the "Optimizing…" badge and documented perf tips.
- Cancel (`file.tsx:402`) drops the *result* but can't abort the in-flight `createImageBitmap`/encode, so a cancelled large resize still burns CPU to completion. Bounded (dropzone is disabled during resize), so no unbounded concurrency.

### Merge readiness: CHANGES REQUESTED

No correctness, security, or compatibility **regressions** — functionally the branch is strictly better than `main` and the resize/cancel logic is now correct. This isn't blocked on a defect. The two things I'd want resolved first are quality, not safety: (1) the feature's unit tests (`file.spec.tsx`) were lost in the rebase, leaving the newly-fixed async logic uncovered; and (2) the `@Watch` comment references that missing file and claims a no-flash guarantee the code no longer delivers (the mount-stability fix was dropped too). Restore both — or drop the comment's claim and add coverage — and this is READY. If you'd rather ship now and follow up, none of the above will break consumers.

---
*🤖 Generated with [Claude Code](https://claude.com/claude-code)*

Kiarokh (approved):

Issue #4162, opened 2026-07-07
Language-aware components crash on a `language` value that has no translations (e.g. `nb`)


Kiarokh:
## Current behavior

`Translations.get` (`src/global/translations.ts`) resolves a translation with an unguarded double index:

```ts
const translation: string = allTranslations[language][key];
```

The public `Languages` type (`src/components/date-picker/date.types.ts`) lists **nine** languages — `da | de | en | fi | fr | nb | no | nl | sv` — but `allTranslations` maps only **eight** (`da, de, en, fi, fr, nl, no, sv`). There is no entry for `nb` (Norwegian Bokmål) and no `src/translations/nb.ts`.

So passing `language="nb"` — a value the TypeScript type explicitly accepts — makes `allTranslations['nb']` `undefined`, and the `[key]` access throws `TypeError: Cannot read properties of undefined`. Roughly 23 components expose a `language` prop and call `translate.get` during render, so this **crashes the component on render** rather than falling back. Any language string outside the map (reachable from loosely-typed / plain-JS consumers) fails the same way.

Steps to reproduce:
1. Render any language-aware component with `language="nb"`, e.g. `<limel-date-picker language="nb">` or `<limel-slider language="nb">`.
2. The component throws `TypeError` on render.

## Expected behavior

An unsupported or unmapped `language` should degrade gracefully — fall back to `en` (and then to the raw key, as `get` already does for a missing key) — instead of throwing. The `nb`/`no` mismatch should also be reconciled: either map `nb` (alias it to the existing `no` translations, or ship an `nb.ts`), or drop `nb` from the `Languages` type so it stops advertising unsupported support.

Suggested defensive fix (covers every unmapped value, not just `nb`):

```ts
const translations = allTranslations[language] ?? allTranslations.en;
const translation: string = translations[key];
```

## Environment

- **lime-elements** version: 39.x (current)
- Framework used: any — framework-agnostic; reproduces in plain Stencil/HTML
- Logs: `TypeError: Cannot read properties of undefined (reading '<key>')`

---

### Side note: default language

All ~23 language-aware components default `language` to `'en'`. It's worth discussing whether the default should instead follow the browser's language (`navigator.language`). Trade-offs:

- **For:** a better out-of-the-box experience for non-English users, without every consumer wiring up i18n.
[...]

PR #4173, opened 2026-07-08
Slider value unset

### LINKED ISSUE #4161: limel-slider: no way to unset / clear the value
## New feature motivation

Every other input component in Lime Elements can be empty (unset): `limel-input-field`, `limel-select`, `limel-checkbox`, and so on all let the user represent "no value chosen yet", and most of them let the user clear a value they have entered.

`limel-slider` is the exception. It is built on a native `<input type="range">`, which **always** resolves to a value. As a result:

- **There is no way for a user to unset or cancel their choice.** 
  Once the thumb is anywhere on the track, there is no affordance to clear it or return to a "nothing selected" state. This is terrible for "un-required" questions in a form or a survey for example.
- **"Unset" is indistinguishable from the minimum.** 
  The component falls back to `valuemin` when no value is provided, so an untouched slider looks exactly like one a user deliberately dragged to its minimum. There is no way to tell the two apart, visually or programmatically.
- **It is inconsistent with the rest of the library.** 
 In a form that mixes a slider with other fields, everything else can be empty but the slider cannot.
- **`required` / `invalid` are effectively meaningless on a slider.** 
  Since the slider can never be empty, an "empty but required" state cannot be represented, so required-validation on a slider has nothing to validate against.
- **Accessibility gap.** 
  
https://github.com/user-attachments/assets/5c116a0d-e8e0-48da-9ae1-f6e54cd8ede3

Because the value can only ever be a number, assistive technology always announces a concrete value, even when the user has made no choice — there is no "not set" state to announce.

This matters for real use-cases such as an optional **Priority**, **Probability**, or **Score** field, where "no value yet" is a legitimate and meaningful state the user must be able to reach — both initially and by clearing a previous choice.

## New feature description

- `limel-slider` should support an **unset** state that is visually distinct from the minimum (e.g. no active fill, a neutral thumb position, and a placeholder instead of a number in the value indicator).
- Users should be able to **clear** the value through a dedicated affordance (e.g. a clear button, and/or the keyboard).
- The unset state should be conveyed to assistive technology (e.g. via `aria-valuetext`), and the clear affordance should have an accessible, context-aware label.
- A slider should be able to be **initialized** unset, and consumers should be able to return it to the unset state programmatically.

https://github.com/user-attachments/assets/1fecafb3-52fa-4fc3-b360-321828c247ec

## New feature implementation

The component already treats "not a finite number" as its fallback path, so an unset state can build on that: track an internal "empty" flag separate from the `value`, suppress the fill and center the thumb when empty, show a placeholder (`?`) in the indicator, and add a clear affordance.

The main open design question is **how to signal "cleared" back to consumers** without a breaking change to the `change` event's current `number` payload (e.g. emitting `NaN` versus widening the payload to `number | null`). This trade-off is worth discussing on the accompanying PR.

Befkadu1:
fix https://github.com/Lundalogik/lime-elements/issues/4161

>[!note]
>see the parent issue https://github.com/Lundalogik/lime-elements/issues/4161 for more details and screen recordings before and after this change.

## Summary by CodeRabbit

* **New Features**
  * Added unset states and clear buttons for sliders.
  * Added required-state controls and an unset-state example.
  * Added localized labels, including Norwegian Bokmål support.
* **Bug Fixes**
  * Improved focus, accessibility, value handling, and translation fallbacks.
  * Updated progress and image-resizing examples for cleared values.
* **Documentation**
  * Added guidance for clearing and programmatically unsetting values.
* **Tests**
  * Expanded coverage for unset states, clearing, accessibility, and interactions.
* **Style**
  * Refined slider thumb, track, unset-state, and clear-button styling.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/slider/examples/slider-unset.tsx
+++ b/src/components/slider/examples/slider-unset.tsx
@@ -1,0 +1,50 @@
+import { Component, h, Host, State } from '@stencil/core';
+
+/**
+ * Unsetting the value
+ *
+ * This slider is initialized *unset*, which means its `value` is `null`.
+ * Therefore the thumb rests in the middle, and the value indicator shows a
+ * left-right arrow (`↔`) instead of a number.
+ * Assistive technologies announce the value as "Value not set".
+ *
+ * As soon as the user drags the thumb, presses anywhere on the track, or
+ * nudges the thumb with the arrow keys, the slider becomes set and the
+ * trailing **clear** button becomes active.
+ * Pressing it unsets the slider again, emitting `null` on the `change`
+ * event — so a handler must accept `number | null`.
+ *
+ * A `required` slider does not offer the clear button — a required value
+ * cannot be unset — but it can still start unset to prompt a first choice.
+ *
+ * To unset the slider programmatically, set its `value` to `null`. Any other
+ * value that is not a finite number — `undefined`, or `NaN` — works too.
+ */
+@Component({
+    tag: 'limel-example-slider-unset',
+    shadow: true,
+})
+export class SliderUnsetExample {
+    @State()
+    private value: number | null = null;
+
+    public render() {
+        return (
+            <Host>
+                <limel-slider
+                    label="Priority"
+                    value={this.value}
+                    valuemin={1}
+                    valuemax={5}
+                    step={1}
+                    onChange={this.handleChange}
+                />
+                <limel-example-value value={this.value} />
+            </Host>
+        );
+    }
+
+    private readonly handleChange = (event: CustomEvent<number | null>) => {
+        this.value = event.detail;
+    };
+}
--- a/src/components/slider/slider.tsx
+++ b/src/components/slider/slider.tsx
@@ -10,6 +10,8 @@ import {
 } from '@stencil/core';
 import { getPercentageClass } from './get-percentage-class';
 import { createRandomString } from '../../util/random-string';
+import translate from '../../global/translations';
+import { Languages } from '../date-picker/date.types';
 
 const DEFAULT_FACTOR = 1;
 const DEFAULT_MAX_VALUE = 100;
@@ -17,7 +19,20 @@ const DEFAULT_MIN_VALUE = 0;
 const MAX_VISIBLE_STEP_DOTS = 20;
 
 /**
+ * Whether the slider holds a value at all. `Number.isFinite` is typed
+ * `(number: unknown) => boolean`, so it tests the right thing but narrows
+ * nothing; this does, which lets the unset check double as proof that what
+ * remains is a number the arithmetic can use. `null`, `undefined` and `NaN`
+ * all mean unset.
+ * @param value - the slider's `value` prop, which may not be a number.
+ */
+const isSetValue = (value: unknown): value is number => {
+    return Number.isFinite(value);
+};
+
+/**
  * @exampleComponent limel-example-slider-basic
+ * @exampleComponent limel-example-slider-unset
  * @exampleComponent limel-example-slider-multiplier
  * @exampleComponent limel-example-slider-multiplier-percentage-colors
  * @exampleComponent limel-example-slider-unit
@@ -69,6 +84,9 @@ export class Slider {
 
     /**
      * Set to `true` to indicate that the slider is required.
+     * A required slider does not offer the clear button, since a required
+     * value cannot be unset. It can still be initialized unset (with a
+     * non-finite `value`) to prompt the user to make a choice.
      */
     @Prop({ reflect: true })
     public required = false;
@@ -94,10 +112,19 @@ export class Slider {
     public unit: string = '';
 
     /**
-     * The value of the input
+     * Defines the language for translations of the accessible labels.
      */
     @Prop({ reflect: true })
-    public value: number;
+    public language: Languages = 'en';
+
+    /**
+     * The value of the input. Set it to `null` to leave the slider unset,
+     * which is also what the `change` event emits once the value is cleared.
+     * Any other value that is not a finite number — `undefined`, or `NaN` —
+     * unsets the slider too.
+     */
+    @Prop({ reflect: true })
+    public value: number | null;
 
     /**
      * The maximum value allowed
@@ -118,28 +145,55 @@ export class Slider {
     public step: number;
 
     /**
-     * Emitted when the value has been changed
+     * Emitted when the value has been changed.
+     * Emits `null` when the value has been cleared and the slider becomes
+     * unset, so handlers must account for a value that is not a number.
      */
     @Event()
-    private change: EventEmitter<number>;
+    private change: EventEmitter<number | null>;
 
     @State()
-    private percentageClass: string;
+    private percentageClass: string | undefined;
 
     @State()
     private displayValue: number;
 
+    /**
+     * `true` while the slider has no value set. Kept separate from the `value`
+     * prop so that the unset state survives the brief window during dragging
+     * where the value has not yet been emitted back to the consumer.
+     */
+    @State()
+    private isUnset: boolean;
+
     private labelId: string;
     private helperTextId: string;
+    private readonly clearButtonId: string;
+    private inputElement?: HTMLInputElement;
+
+    /**
+     * `true` while the value on display is one the consumer has been told
+     * about — either it came from the `value` prop, or it has been emitted on
+     * `change`. It is `false` from the moment the slider becomes unset until a
+     * value is committed again.
+     *
+     * The native input always holds a value, so while unset it rests at the
+     * midpoint. A press that lands on that midpoint, or a drag that returns to
+     * it before releasing, leaves the input's value untouched and fires neither
+     * `input` nor `change`. Without this flag such an interaction is swallowed:
+     * the slider either stays unset, or shows a number the consumer never
+     * received.
+     */
+    private valueIsCommitted = false;
 
     public constructor() {
         this.labelId = createRandomString();
         this.helperTextId = createRandomString();
+        this.clearButtonId = createRandomString();
     }
 
     public componentWillLoad() {
-        this.displayValue = this.multiplyByFactor(this.getValue());
-        this.setPercentageClass(this.getValue());
+        this.syncStateFromValue();
     }
 
     public render() {
@@ -165,7 +219,7 @@ export class Slider {
                     invalid={this.invalid}
                     disabled={this.disabled}
                     readonly={this.readonly}
-                    hasValue={!!this.value}
+                    hasValue={!this.isUnset}
                     hasFloatingLabel={true}
                 >
                     <div slot="content">
@@ -184,8 +238,20 @@ export class Slider {
                                         ? this.helperTextId
                                         : undefined
                                 }
+                                aria-valuetext={
+                                    this.isUnset
+                                        ? translate.get(
+                                              'value-not-set',
+                                              this.language
+                                          )
+                                        : undefined
+                                }
+                                ref={(el?: HTMLInputElement) => {
+                                    this.inputElement = el;
+                                }}
                                 onInput={this.handleInput}
                                 onChange={this.handleChange}
+                                onClick={this.handleClick}
                                 {...inputProps}
                             />
                             <div class="track">
@@ -195,7 +261,9 @@ export class Slider {
                             <div class="thumb">
                                 <div class="knob" />
                                 <div class="indicator" aria-hidden="true">
-                                    {this.displayValue}
+                                    {this.isUnset
+                                        ? '\u2194\uFE0E'
+                                        : this.displayValue}
                                 </div>
                             </div>
                         </div>
@@ -211,6 +279,7 @@ export class Slider {
                         </div>
                     </div>
                 </limel-notched-outline>
+                {this.renderClearButton()}
                 {this.renderHelperLine()}
             </Host>
         );
@@ -218,9 +287,30 @@ export class Slider {
 
     @Watch('value')
     protected watchValue() {
-        this.displayValue = this.multiplyByFactor(this.getValue());
-        this.setPercentageClass(this.getValue());
+        this.syncStateFromValue();
     }
+
+    private readonly syncStateFromValue = () => {
+        const value = this.value;
+
+        if (!isSetValue(value)) {
+            this.enterUnsetState();
+
+            return;
+        }
+
+        this.isUnset = false;
+        this.valueIsCommitted = true;
+        this.displayValue = this.multiplyByFactor(value);
+        this.setPercentageClass(value);
+    };
+
+    private readonly enterUnsetState = () => {
+        this.isUnset = true;
+        this.valueIsCommitted = false;
+        this.displayValue = this.getRestingDisplayValue();
+        this.percentageClass = undefined;
+    };
 
     private renderStepDots = (min: number, max: number) => {
         if (!this.step) {
@@ -250,10 +340,50 @@ export class Slider {
         );
     };
 
+    /**
+     * A required slider must hold a value, so it offers no way to clear it,
+     * and a readonly one cannot be edited at all. The styling needs the same
+     * answer to reserve room for the button, so both read it from here rather
+     * than each deriving it — the stylesheet cannot see a parsed prop, only
+     * the attribute string, and `required="false"` is a present attribute.
+     */
+    private readonly hasClearButton = (): boolean => {
+        return !this.readonly && !this.required;
+    };
+
+    private readonly renderClearButton = () => {
+        if (!this.hasClearButton()) {
+            return;
+        }
+
+        const label = this.label
+            ? translate.get('clear-value-of', this.language, {
+                  label: this.label,
+              })
+            : translate.get('clear-value', this.language);
+
+        return (
+            <button
+                type="button"
+                id={this.clearButtonId}
+                class="clear-button"
+                aria-label={label}
+                onClick={this.handleClear}
+                disabled={this.isUnset || this.disabled}
+            >
+                <svg viewBox="0 0 32 32" aria-hidden="true" focusable="false">
+                    <path d="M7.219 5.781L5.78 7.22 14.563 16 5.78 24.781 7.22 26.22 16 17.437l8.781 8.782 1.438-1.438L17.437 16l8.782-8.781L24.78 5.78 16 14.563z" />
+                </svg>
+                <limel-tooltip elementId={this.clearButtonId} label={label} />
+            </button>
+        );
+    };
+
     private handleInput = (event: Event) => {
         event.stopPropagation();
         const input = event.target as HTMLInputElement;
         const value = Number(input.value);
+        this.isUnset = false;
         this.displayValue = value;
         this.setPercentageClass(value / this.factor);
     };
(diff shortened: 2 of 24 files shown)


Kiarokh:
> 🤖 AI-generated review from 6 parallel agents, **re-run after the review-driven fixes were applied to this branch**. Treat as input, not a verdict — agents can be wrong or miss context.

# Consolidated Review — `limel-slider`: allow the value to be unset

## Summary

Gives `limel-slider` an unset/clearable state. When the value is non-finite (`NaN`/`null`/`undefined`) the thumb rests centered, the indicator shows `?`, and assistive tech announces "Value not set" via `aria-valuetext`; the state is exposed as `is-unset` on the host. A trailing clear `<button>` (inline-SVG icon, context-aware `aria-label` + tooltip) is offered on optional, editable sliders and emits `NaN` on `change`, then focuses the slider. Adds a `language` prop, three generic translation keys across 8 locales, an example, and a behavioral spec.

No blocking defects — the branch is clean (build green, 10 behavioral specs passing). The two caveats below are **decisions/verifications, not bugs**.

- **Blockers:** None.
- **Caveats worth settling before/at merge:**
  1. Confirm the intended contract that *optional* sliders are clearable-by-default and emit `NaN` on clear (Top Recommendation #1 — the `NaN`-vs-`null` question, tracked in #4161).
  2. Verify `NaN` handling in lime-crm form save/validation (Top Recommendation #2).

### Already addressed on this branch during the review
Several findings from the run were fixed while reviewing, and are reflected below as resolved:
- **Midpoint anchoring** so both arrow-key directions are live after clearing (no dead keys, no thumb jump).
- **Focus moves to the slider on clear** — focus no longer falls to `<body>`, and AT announces the now-unset state.
- `required`/`readonly` gate the clear button (so it isn't forced on required fields); internal naming unified to `isUnset`/`.is-unset`; `is-unset` exposed on the host; `enterUnsetState()` extracted to keep the unset assignments in one place; the example reports the real value (`null`); 10 behavioral specs added.
- **Guarded `translate.get`** so an unmapped `language` falls back to English instead of throwing — fixing the pre-existing **#4162** crash that the composite example's runtime test surfaced; `nb` is also aliased to the Norwegian translations.

---

## Dimensions

### 1. Backward Compatibility — ACCEPTABLE ⚠️

The public API is additive (only a new `language` prop; `change` stays `EventEmitter<number>`, `value` stays `number`), but the feature changes default behavior for existing sliders.

**What works well:** no removed/retyped API; `required`/`readonly` gating limits the new surface; a `value` of `0` is now handled correctly (readonly `0` shows `0`, not the em-dash).

**Issues:**
- 🟡 **[Medium]** Clear button, reserved padding, and unset-instead-of-min rendering are now default-on for every optional editable slider, with no opt-out except `required`/`readonly`; clearing emits `NaN` on `change` (`slider.tsx` `renderClearButton`/`handleClear`, `form/widgets/slider.ts` forwards `event.detail`). Intended behavior — see **Top Recommendation #1**.
- 🟡 **[Medium]** That `NaN` reaches rjsf form state through the slider widget. See **Top Recommendation #2**.

<details>

- 🔵 `hasValue` for `value === 0` flipped `false`→`true` (`hasValue={!this.isUnset}`); the only visible effect is a readonly `0` now rendering the value instead of the `–` placeholder — a net fix, but an edge behavior change.

</details>

### 2. Architecture — GOOD ✅

Sound state model and interaction. The Medium (focus-on-clear stranding keyboard users) and the empty-state duplication Low from the run were **both fixed on this branch**.

**What works well:** single-source unset state via `enterUnsetState()`; `NaN` is fully contained (never reaches arithmetic — `getFraction`/`multiplyByFactor`/`setPercentageClass` only see finite values); focus-on-clear runs synchronously before the button self-disables (no race); midpoint anchoring keeps both arrow directions live; `is-unset` on the host lets the whole component (and consumers) style off it.

**Issues:** None.

<details>

- 🔵 `@Prop({ reflect: true })` on `value` can surface a cosmetic `value="NaN"` attribute on the host when cleared. Harmless (nothing reads the attribute), just untidy.

</details>

### 3. Code Quality — GOOD ✅

The new spec is genuinely behavioral (state class, indicator text, aria, event payload, focus, gating, contextual label) — no assertion that would pass if the feature broke. The `isUnset`/`is-unset` naming drift flagged in the run is now unified.

**What works well:** behavioral tests; consistent "unset" vocabulary; translations follow the `{ label }` merge-code convention.

**Issues:** None.

<details>
<summary>Minor nits (4)</summary>

- 🔵 `value: number` is documented to accept `null`/`undefined`, but only `NaN` typechecks for a strict-TS consumer (`value?: number` would at least make `undefined` honest).
- 🔵 Test gap: the contextual-label test only asserts English — one non-English case (e.g. `language: 'sv'`) would exercise the translation plumbing end-to-end.
- 🔵 Test gap: the clear button's `|| this.disabled` branch is untested (only the unset path is covered).
- 🔵 Narrow uncontrolled-use edge: after a button-clear, `value` stays finite locally while `isUnset` is set; if a consumer then re-assigns the identical value, `@Watch('value')` won't fire. Inherent to the deliberate `NaN`-emit design and very narrow.

</details>

### 4. Security — SAFE ✅

**What works well:** the consumer-controlled `label` reaches only text/attribute sinks — traced through `translate.get` (function replacer, so no `$`-pattern expansion; regex runs on the static template, so no ReDoS) into `aria-label` and the tooltip's escaped text node. The inline clear-icon SVG is a static path with no interpolation. `?`/`aria-valuetext` carry no untrusted input.

**Issues:** None.

### 5. Observability — GOOD ✅

Appropriate for a UI component; no swallowed errors, no stray `console.*`, and the unset contract is documented (JSDoc, example, reflected attribute, host class, `aria-valuetext`).

**Issues:** None.

<details>

- 🔵 The cleared-state announcement is timing-dependent: `handleClear` focuses the input synchronously, but `aria-valuetext="Value not set"` is applied on Stencil's async re-render. Most screen readers re-read after commit, but deferring the focus to after render would make it deterministic.
- 🔵 The focus test asserts `.focus()` was called, not that the unset state is actually announced.

</details>

### 6. Performance — GOOD ✅

No regression — the tooltip's heavy machinery (Popper, `ResizeObserver`, timers) stays lazy until hover.

**What works well:** `syncStateFromValue` runs only on load / `@Watch('value')` (not per drag frame); `setPercentageClass` early-returns while unset; the mounted tooltip adds no drag-time cost (its props are inequality-guarded).

**Issues:** None.

<details>
<summary>Minor nits (3)</summary>

- 🔵 The inline clear-icon SVG is re-patched (`setAttribute`) each render, vs the old zero-cost CSS background — negligible (browsers no-op identical values).
- 🔵 Per-render `ref` closure allocation on the range input.
- 🔵 `translate.get()` recomputes the same label string on every render (incl. per drag frame); could be memoized. Microseconds.

</details>

---

## Top Recommendations

1. **🟡 [Medium — Backward Compatibility] — Confirm the "optional = clearable, emits `NaN`" contract**
   - **Where:** `slider.tsx` `renderClearButton`/`handleClear`; forwarded via `form/widgets/slider.ts`.
   - **What:** Every optional (non-`required`, non-`readonly`) slider now shows a clear button and, when cleared, emits `NaN` on `change` — the event type is deliberately kept `EventEmitter<number>` (rather than `number | null`) to avoid a breaking type change. Decide whether that's the contract we want, or whether a `clearable` opt-in / migration note is warranted.
   - **Why:** It's the intended feature and arguably more correct than the old "silently rests at min", but it changes existing optional sliders' appearance and lets `NaN` flow to consumers with no opt-out. This is the core `NaN`-vs-`null` discussion (issue #4161).

2. **🟡 [Medium — Backward Compatibility] — Verify `NaN` handling in lime-crm form save/validation**
   - **Where:** `form/widgets/slider.ts` → `props.onChange(NaN)`; downstream validation/serialization.
   - **What:** Confirm a cleared optional slider behaves acceptably downstream — `NaN` serializes to `null` via `JSON.stringify`, and may hit AJV `type: "number"`.
   - **Why:** On `main` a slider could never emit a non-finite value into form state; this path is newly reachable. Mostly moot for required fields (no clear button), but worth a concrete check before shipping to form consumers.

---

_Net: clean and clearly better than `main`. The headline finding from the run (focus-on-clear stranding keyboard users) is fixed via midpoint anchoring, naming/dedup nits are resolved, tests are in place, and security/perf are clean. What remains is one design decision to ratify and one downstream check._

src/components/slider/slider.tsx
```diff
@@ -94,7 +100,14 @@ export class Slider {
...
+     */
+    @Prop({ reflect: true })
+    public language: Languages = 'en';
+
+    /**
+     * The value of the input. When the value is `NaN`, `null`, `undefined`,
+     * or otherwise not a finite number, the slider is considered unset.
      */
     @Prop({ reflect: true })
     public value: number;
```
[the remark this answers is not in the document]
Kiarokh: Addressed 👍 — clarified the JSDoc: `NaN` is the type-safe way to unset programmatically, and `null`/`undefined` still work at runtime. Kept `value: number` deliberately for now — widening the type is part of the broader API-shape discussion.

src/components/slider/slider.tsx
```diff
@@ -129,17 +143,27 @@ export class Slider {
...
     public constructor() {
         this.labelId = createRandomString();
         this.helperTextId = createRandomString();
+        this.clearButtonId = createRandomString();
     }
 
     public componentWillLoad() {
-        this.displayValue = this.multiplyByFactor(this.getValue());
-        this.setPercentageClass(this.getValue());
+        this.syncStateFromValue();
```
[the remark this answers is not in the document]
Kiarokh: Addressed 👍 — `percentageClass` is now typed `string | undefined`.

adrianschmidt:
I haven't thought deeply about it, but my gut says it's better to take the "honest" route, and make this a breaking change.

The typing in itself only matters for typescript, so the compiled javascript won't stop working. So for the _type itself_, other packages won't have to update anything until they update their dev-dependency on lime-elements to a version including the change.

So, the question is, do those existing packages already handle `null` the same way as they already handle `NaN`? If they do, I definitely think we should go with `null`. The way to test this is to install some other package that uses a slider along with a version of  lime-elements with the `null` version of the change, but not install that version of lime-elements in that other package. That package should be built with its own old dependency on lime-elements, and should only "meet" the new version at runtime inside the web client. Ping me if you're unsure how I mean!

src/components/slider/slider.tsx
```diff
@@ -294,6 +399,21 @@ export class Slider {
         return value;
     };
 
+    /**
+     * The display value the thumb rests at while unset: the midpoint of the
+     * range, so both arrow-key directions are live (unlike anchoring at the
+     * minimum). Aligned to the step — native range inputs default to a step of
+     * 1 — so it matches a real input value and the first key press doesn't jump.
+     */
+    private readonly getRestingDisplayValue = (): number => {
```
Befkadu1, line 474: **issue (blocking):** Clicking the middle of an unset slider leaves it unset.

In the new example (`valuemin={1} valuemax={5} step={1}`) the resting value is 3, so the native input already holds 3. A click or tap anywhere that maps to 3 — the middle quarter of the track — does not change the input's value, and a range input fires neither `input` nor `change` when the value is unchanged. `isUnset` stays true, nothing is committed, and the click is silently swallowed.

Dragging works, because the value passes through other steps on the way. Tap-to-set in the centre does not. The spec suite dispatches a synthetic `input` event, so it cannot catch this.

A `pointerdown` (or `click`) handler on the input that commits the resting value while unset would close the gap.

src/components/slider/slider.tsx
```diff
@@ -250,10 +307,40 @@ export class Slider {
...
+                <limel-tooltip elementId={this.clearButtonId} label={label} />
+            </button>
+        );
+    };
+
     private handleInput = (event: Event) => {
         event.stopPropagation();
         const input = event.target as HTMLInputElement;
         const value = Number(input.value);
+        this.isUnset = false;
```
Befkadu1, line 386: **issue (blocking):** Leaving the unset state here is not reversible when the value is never committed.

Start unset, drag away from the resting value, then drag back to it before releasing. `input` fires during the drag so `isUnset` flips to false, but the native `change` never fires because the released value equals the value at drag start. The indicator now shows a number, `value` is still `NaN`, and `watchValue` cannot fire because the prop never changed. The component and the consumer stay out of sync until some unrelated prop change.

Before this PR there was no internal flag and the display was always derived from `value`, so the divergence could not stick.

src/components/slider/slider.tsx
```diff
@@ -118,28 +133,39 @@ export class Slider {
     public step: number;
 
     /**
-     * Emitted when the value has been changed
+     * Emitted when the value has been changed.
+     * Emits `NaN` when the value has been cleared and the slider becomes unset.
```
Befkadu1: **question (blocking):** Should the clear affordance be opt-in rather than on by default?

Every existing slider that is neither `required` nor `readonly` now grows a clear button and can emit `NaN` on an event still typed `EventEmitter<number>`, so TypeScript warns nobody.

Two of this repo's own examples show what that does downstream. `src/components/file/examples/file-resize-image.tsx:198` computes `Math.max(0, Math.min(1, event.detail / 100))`, which is `NaN` for a cleared slider, and hands it to `resizeImage.quality`. `src/components/chip/examples/chip-progress.tsx:64` assigns the detail straight to `progress`, so three chips render `progress={NaN}`. Consumer code has the same shape.

The commits are all typed `fix(slider): …`, so this releases as a patch.

src/components/slider/examples/slider-unset.tsx
```diff
@@ -0,0 +1,47 @@
+import { Component, h, Host, State } from '@stencil/core';
+
+/**
+ * Unsetting the value
+ *
+ * This slider is initialized *unset*, which means its `value` is `NaN`.
+ * Therefore the thumb rests in the middle, and the value indicator shows a `?`.
```
Befkadu1: **todo (if-minor):** The indicator renders `↔`, not `?`.

`slider.tsx:236` renders `'\u2194\uFE0E'`. The same `?` claim is repeated in the comment at `slider.scss:179`. Both texts are published.

src/global/translations.ts
```diff
@@ -14,6 +14,7 @@ const allTranslations = {
     fi: fi,
     fr: fr,
     no: no,
+    nb: no, // Norwegian Bokmål shares the Norwegian (`no`) translations
```
Befkadu1, line 17: **suggestion (if-minor):** Add a spec for `translate.get` alongside this fix.

`nb` is in the `Languages` type, so before this change `allTranslations['nb']` was `undefined` and `undefined[key]` threw a `TypeError` — any component rendered with `language="nb"` crashed. That is a real fix across roughly thirty components, and `translate.get` has no test anywhere in `src`. Two cases would pin it: `get('save', 'nb')` returns the Norwegian string, and an unmapped language falls back to English instead of throwing.

src/components/slider/slider.tsx
```diff
@@ -309,6 +429,12 @@ export class Slider {
     };
 
     private setPercentageClass = (value: number) => {
+        if (this.isUnset) {
+            this.percentageClass = undefined;
```
Befkadu1: **suggestion (non-blocking):** This guard cannot be reached.

Both callers set `this.isUnset = false` before calling in — `syncStateFromValue` at line 271 and `handleInput` at line 343 — and `enterUnsetState` assigns `percentageClass` directly rather than routing through here.

Befkadu1 (commented):
Tapping the middle of an unset slider does nothing — the native input already rests there, so no `input` event fires and the slider stays unset.

Kiarokh:
@Befkadu1 tnx for the great review and all the fixups

src/components/slider/slider.tsx
```diff
@@ -250,28 +326,118 @@ export class Slider {
...
 
+    private readonly handleClear = (event: MouseEvent) => {
+        event.stopPropagation();
+        this.enterUnsetState();
+        this.change.emit(null);
+
+        // Move focus to the slider itself so keyboard users can immediately
+        // set a new value, and so assistive tech announces the now-unset state
+        // instead of focus falling to the body when the button self-disables.
+        this.inputElement?.focus();
```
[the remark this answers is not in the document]
Befkadu1: The premise is correct, but the suggested change would trade this for a worse problem, so I'm leaving the synchronous `focus()` in place.

I probed the actual ordering in the spec harness:

```
before click:            aria-valuetext = null
immediately after click: aria-valuetext = null            <- when .focus() runs
after waitForChanges:    aria-valuetext = "Value not set"
clear button [disabled]: false (sync)  ->  true (after render)
```

So yes, the attribute lags the focus call. But the same probe shows why the focus is synchronous: the clear button is *still enabled* at that instant. That's deliberate — `fix(slider): focus the slider after clearing its value` exists precisely to stop focus falling to `<body>` when the button self-disables a frame later.

Deferring to `requestAnimationFrame`/`componentDidRender` puts the focus call *after* the button becomes disabled. A focused element that becomes disabled hands focus to `<body>`, so focus would round-trip button -> body -> input, firing a `focusout`/`focusin` pair and giving assistive tech a document change to announce before it ever reaches the slider. That's a regression, not a fix.

On impact: `aria-valuetext` changing on an already-focused slider is the same mechanism that announces every drag, so screen readers do pick the new value up. The residue is a possible "3" then "Value not set" double announcement — imperfect, but no information is lost.

Also worth flagging: the proposed regression test wouldn't cover the claim. Asserting `aria-valuetext` on the focused input *after* `waitForChanges()` already passes on current code and says nothing about announcement ordering; there's no way to assert what a screen reader utters from this harness.

If we want it airtight without touching focus order, the option is to set the attribute imperatively immediately before focusing, so the render that follows writes the identical value:

```ts
this.inputElement?.setAttribute(
    'aria-valuetext',
    translate.get('value-not-set', this.language)
);
this.inputElement?.focus();
```

Happy to do that if reviewers prefer it; the cost is one imperative write duplicating render output.

Kiarokh (approved):
Approved from my side, what do you think @adrianschmidt ?

adrianschmidt (dismissed):
The BREAKING CHANGE bit is not in the commit message footer, so semantic-release will not detect it.

adrianschmidt:
The commit message footer is only the bit after the last blank line, so in this message:

image

… the footer is only `Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>`

Note that semantic-release includes _everything_ after the `BREAKING CHANGE:`, so the `Co-Authored-By` line should be moved before the `BREAKING CHANGE:` line, with no blank line between them. Like this:

```
fix(slider)!: emit null instead of NaN when the value is cleared

`NaN` typechecks as a `number`, so the compiler stayed silent while every
optional slider gained the ability to hand its consumers a value that
poisons arithmetic. `NaN` also has no JSON representation, so anything
persisting a cleared slider round-tripped as `null` anyway and the
in-memory value disagreed with the stored one.

Clearing now emits `null`, and `value` accepts `number | null` so the
emitted value can be handed straight back to the slider. Any non-finite
value, `NaN` and `undefined` included, still unsets it, so existing
callers keep working at runtime.

The library's own consumers are updated with it: the rjsf widget forwards
`null` into form state, the progress examples fall back to 0, and the
image resize example omits `quality` so the browser's native encoding
quality is used instead of an imposed one.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
BREAKING CHANGE: `limel-slider`'s `change` event emits `null` when the
value is cleared, widening its payload from `number` to `number | null`,
and the `value` prop accepts `number | null` for the same reason.
Handlers that assume a number must guard for `null`.
```

> [!IMPORTANT]
>  Make sure to change both commits. Changing only one will be enough for semantic-release to make the major version bump, but the BREAKING-message for the unchanged commit will be missing from the changelog.

Kiarokh:
Thanks @adrianschmidt 
Fixed in both commits. The `Co-Authored-By` trailers now sit directly above `BREAKING CHANGE:` with no blank line between them, so the breaking note is the last block in each message and semantic-release picks up the whole note and nothing more.

- `fix(slider)!: allow the value to be unset`: `Refs #4161`, blank line, then both `Co-Authored-By` lines immediately followed by `BREAKING CHANGE:`
- `fix(slider)!: emit null instead of NaN when the value is cleared`: `Co-Authored-By` immediately followed by `BREAKING CHANGE:`, exactly as you wrote it

The branch was rewritten by replaying the commits (amend + cherry-pick) and force-pushed with `--force-with-lease`. Only the two messages changed: the resulting tree is identical to the previous tip, and every other commit kept its original author, date and diff.

adrianschmidt:
Hmm… I just had the thought "what are the two breaking changes anyway?", and they are the same… or the descriptions are the same. But the commit message title of the second one says it's changing from `NaN` to `null`. Both breaking changes say they are changing to `null` though. I guess keeping a super tidy commit history becomes both harder and perhaps less important now that we all use AI to handle the coding parts anyway, but this kind of messiness still bothers me. I'll approve, but it would be really nice if the commits could be made a bit more atomic. 14 separate commits for something that is essentially a single new feature is kind of ridiculous. This PR alone would, in its current state, create _9_ changelog entries, with 2 breaking changes entries :persevere:

Kiarokh:
you're right. All commits after `fix(slider)!: allow the value to be unset` should be just squashed into the same commit. I'll fix this. Also, there is a typo in there 😅 `fix(slider)!:` Not sure how that ended up there. I think everything supposed to be a fixup

Kiarokh:
Fixed now. :zap And actually the API change was a `feat`. So I reworded even that.

Huh, Claude says:

> The `!` is the Conventional Commits marker for a breaking change — it's deliberate, not a typo.
> 
> It makes the break visible in the one-line header, so anyone scanning git log --oneline sees it without having to open the message and read down to the footer. Functionally it's redundant here: the repo's conventionalcommits preset (release.config.cjs:19) treats either signal as a major bump, and the `BREAKING CHANGE: `footer alone would already do it.

Kiarokh (approved):

adrianschmidt (approved):

Befkadu1:
@adrianschmidt Thank you for your support

PR #4176, opened 2026-07-09
test(text-editor): cover inlineImages validation and imagePasted suppression

john-traas:
Follow-up test coverage for the `inlineImages` config (#4150), complementing the unit tests that merged with the feature.

## What's covered

**`inlineImages.tagName` validation guard** (component-level, `text-editor.spec.tsx`)
- A built-in tag name like `img` is rejected with a console error and the editor still renders — the runtime guard that protects plain-JS consumers, since a built-in tag name would hijack the sanitizer whitelist rules for that tag.
- A valid custom element tag name is accepted silently.

**`imagePasted` suppression** (browser e2e, `text-editor.e2e.tsx`)
- When `inlineImages` is configured, pasting an image file runs the upload lifecycle and the legacy `imagePasted` event never fires. This mutual-exclusivity contract previously existed only as a code comment in `inserter.ts`.
- Positive control: without `inlineImages`, the same paste emits `imagePasted` — so the suppression assertion cannot pass vacuously.

Both suppression outcomes are synchronized on observable effects (the upload spy / the event itself), not sleeps. Mutation-checked: removing the suppressing `return` in `handlePastedImages` fails exactly the suppression test.

Out of scope: the paste→upload lifecycle tests (`allowPasteImages` gating, `imageUploadingChange` aggregation) are planned separately.

## Summary by CodeRabbit

* **New Features**
  * Text editors now support configurable inline image handling and uploads.
  * Image pasting supports uploader-based processing when configured, while preserving legacy event behavior otherwise.

* **Bug Fixes**
  * Improved validation for inline image element names.
  * Invalid built-in names now report an error while the editor continues to render.
  * Valid custom element names are accepted without errors.


src/components/text-editor/text-editor.spec.tsx
```diff
@@ -149,4 +152,44 @@ describe('limel-text-editor', () => {
...
+                .spyOn(console, 'error')
+                .mockImplementation(() => {});
+            const inlineImages: InlineImages = {
+                tagName: 'my-image',
+                getUrl: (id) => `/files/${id}`,
+            };
+
+            await createComponent({ inlineImages });
+
+            expect(consoleError).not.toHaveBeenCalledWith(invalidTagNameError);
```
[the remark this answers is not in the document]
john-traas: Agreed — tightened to `not.toHaveBeenCalled()`; the full spec suite confirms no unrelated console errors in this environment. ⚡ 69080024a17f6afe6eefe0665147ea929deaffaa

src/components/text-editor/text-editor.spec.tsx
```diff
@@ -149,4 +152,44 @@ describe('limel-text-editor', () => {
...
+            const consoleError = vi
+                .spyOn(console, 'error')
+                .mockImplementation(() => {});
+            // The type only allows hyphenated names; the cast exercises the
+            // runtime guard that protects plain-JS consumers.
+            const inlineImages = { tagName: 'img' } as unknown as InlineImages;
+
+            const { root } = await createComponent({ inlineImages });
+
+            expect(consoleError).toHaveBeenCalledWith(invalidTagNameError);
```
jgroth: **issue (if-minor):** The test covers the console error, not the "ignoring inlineImages config" half of the guard.

I removed `return undefined;` from `validateInlineImages` (`prosemirror-adapter.tsx:419`) and kept the `console.error`. All 13 tests in this file still pass. With that mutation the `img` config reaches the sanitizer whitelist, which is the failure the guard exists to prevent.

I could not find a cheap way to observe the drop from the spec level — `validatedInlineImages` is private and the config prop reaches the adapter either way. An assertion on the described outcome (a real `` keeps its `src`) may fit better as an e2e test next to the paste ones.

src/components/text-editor/text-editor.spec.tsx
```diff
@@ -149,4 +152,44 @@ describe('limel-text-editor', () => {
...
+                root.shadowRoot.querySelector('limel-prosemirror-adapter')
+            ).not.toBeFalsy();
+        });
+
+        test('a custom element tag name is accepted', async () => {
+            const consoleError = vi
+                .spyOn(console, 'error')
+                .mockImplementation(() => {});
+            const inlineImages: InlineImages = {
+                tagName: 'my-image',
```
jgroth: **suggestion (if-minor):** Add a malformed hyphenated name so the character pattern is covered too.

`isValidTagName` is `/^[a-z][a-z0-9._-]*$/.test(name) && name.includes('-')`. `img` fails only the second half, and `my-image` passes both, so nothing exercises the regex. I replaced the whole check with plain `name.includes('-')` and every test stayed green. A name such as `My-Image` or `1-x` has a hyphen but is still not usable as a custom element name.

src/components/text-editor/text-editor.e2e.tsx
```diff
@@ -347,4 +353,60 @@ describe('limel-text-editor', () => {
...
+                expect(imagePasted).toHaveBeenCalledTimes(1);
+            });
+        });
+
+        test('pasting an image never emits imagePasted when inline images are configured', async () => {
+            const upload = vi
+                .fn()
+                .mockResolvedValue('https://example.com/cat.png');
+            const { editable, imagePasted } = await createEditor({
+                inlineImages: { upload: upload },
```
jgroth: **issue (if-minor):** The suppression is only proven for a config that has `upload`.

`inserter.ts:344` states the other half of the contract: with `inlineImages` set but no `upload`, the paste is a no-op and the legacy event must still stay silent. I moved the `return` inside `if (inlineImages.upload)`, so a config without an uploader falls back to `imagePasted`, and all 20 e2e tests passed.

A third case with `{ tagName: 'my-image', getUrl }` and no `upload` covers it, and it needs no upload spy — only the absence of the event.

src/components/text-editor/text-editor.spec.tsx
```diff
@@ -149,4 +152,44 @@ describe('limel-text-editor', () => {
...
+                getUrl: (id) => `/files/${id}`,
+            };
+
+            await createComponent({ inlineImages });
+
+            expect(consoleError).not.toHaveBeenCalled();
```
jgroth: **suggestion (non-blocking):** Scope this to `invalidTagNameError` instead of the whole spy.

Any unrelated `console.error` during render — from an icon fetch, a Stencil warning, a sibling component — fails this test under a name that points at tag name validation. `expect(consoleError).not.toHaveBeenCalledWith(invalidTagNameError)` reuses the constant on line 157 and keeps the failure message honest. The name also promises more than the body checks: silence is the only thing asserted, not that the config took effect.

src/components/text-editor/text-editor.spec.tsx
jgroth: **suggestion (non-blocking):** This proves the error is logged, not that the config is dropped.

I deleted `return undefined;` from `validateInlineImages` (`prosemirror-adapter.tsx:419`), kept the `console.error`, and all 13 tests still passed.

The gap is small though. The sanitizer whitelist has its own hyphen guard in `buildContentWhitelist` (`utils/content-type-converter.ts:37`), already covered by `content-type-converter.spec.ts:40`. Dropping the config here only stops it reaching `buildEditorSchema` and `buildEditorPlugins`, and `validatedInlineImages` is private.

Fix: no cheap one from a spec. Fine to leave as is.

src/components/text-editor/text-editor.spec.tsx
jgroth: **suggestion (if-minor):** Half of `isValidTagName` is never tested.

The check is `/^[a-z][a-z0-9._-]*$/.test(name) && name.includes('-')`. `img` fails only the second half, and `my-image` passes both, so the regex never decides anything. I replaced the whole check with plain `name.includes('-')` and every test stayed green.

Fix: add one more rejected case with a name that has a hyphen but the regex still rejects, such as `My-Image` (uppercase) or `1-image` (leading digit).

src/components/text-editor/text-editor.e2e.tsx
jgroth: **issue (if-minor):** Suppression is only proven for a config that has `upload`.

`inserter.ts:344` states the other half: once `inlineImages` is set it owns the paste, so with no `upload` handler the paste is a no-op and `imagePasted` must still stay silent. I moved the `return` inside `if (inlineImages.upload)` and all 20 e2e tests passed, so nothing here guards that case.

Fix: add a third case with `{ tagName: 'my-image', getUrl }` and no `upload`, then assert `imagePasted` never fires. No upload spy needed.

src/components/text-editor/text-editor.spec.tsx
jgroth: **suggestion (non-blocking):** Assert on the specific error, not on the whole spy.

Any unrelated `console.error` during render fails this test, and the failure message will point at tag name validation.

Fix: `expect(consoleError).not.toHaveBeenCalledWith(invalidTagNameError)`, reusing the constant on line 157.

jgroth (approved):

PR #4178, opened 2026-07-09
fix(file): remove baked-in defaults from the image-resize util

### LINKED ISSUE #4177: image-resize util bakes in profile-picture defaults that leak to all consumers
## Summary

The shared `resizeImage` util (`src/util/image-resize.ts`) bakes in defaults that are only appropriate for one of its consumers (`limel-profile-picture`). Note that the util was invented together with the `limel-profile-picture`. This is why it's formed like this. 

Because the util is shared, those defaults silently apply to every consumer, which produces surprising results for anyone else — most notably `limel-file`.

## Background

`image-resize.ts` was originally written for `limel-profile-picture`. Avatars want a specific, opinionated output: fill-and-crop to a square, and a compact JPEG. So the util was given defaults tuned for that use case:

```ts
const {
    fit = 'cover',
    type = 'image/jpeg',
    quality = 0.85,
    ...
} = options;
```

Those defaults live in the wrong place. They are a *product* decision that belongs to the profile-picture component, not properties of "resize an image."

## The problem

When a second consumer (`limel-file`, driven by the new admin-configurable resize options) started using the same util, it inherited those defaults:

- **`type` defaults to `image/jpeg`.** Resizing a PNG while setting only a `width` silently converts it to JPEG and loses transparency — a PNG logo gets flattened. The caller never asked for a format change; it just wanted a smaller image.
- **`quality` defaults to `0.85`.** The caller is forced into a specific re-compression level even when it didn't ask to set quality at all.
- **`fit` defaults to `cover`.** Center-cropping (discarding pixels) is a reasonable default for avatars, but a surprising, lossy default for a generic file upload.

In every case the util is making an opinionated choice on the caller's behalf, and the caller can't express "just resize, don't change anything else."

## Proposed design

A shared, low-level helper should impose **no defaults of its own**. When an option is omitted, the util should preserve the source file's own property rather than substituting an opinionated value. Any consumer that wants a specific default should apply it itself.

Concretely:

- **`type` omitted** → keep the source file's format when the canvas can encode it (PNG stays PNG, otherwise JPEG).
- **`quality` omitted** → let the browser use its native encoding quality instead of imposing a number.
- **`fit` omitted** → scale the whole image to fit without cropping (no pixels discarded).

`limel-profile-picture` keeps its current behaviour, but applies its defaults (`cover` / `image/jpeg` / `0.85`) *in the component*, so avatars are unaffected while every other consumer gets predictable, non-destructive behaviour.

## Not a breaking change

The `resizeImage` util is marked `@beta`, so changing its default behaviour is not a breaking change under our stability guarantees.

Kiarokh:
Closes #4177
fix https://github.com/Lundalogik/lime-webclient/issues/7902

## What

The shared `resizeImage` util (`src/util/image-resize.ts`) was written for `limel-profile-picture` and baked in defaults tuned for avatars — `fit=cover`, `type=image/jpeg`, `quality=0.85`. Because the util is shared, those defaults silently applied to every consumer. The visible symptom: `limel-file` resizing a PNG with only a `width` set silently flattened it to JPEG and lost transparency.

This removes all baked-in defaults from the util. An omitted option now preserves the source file's own property; consumers that want a specific default apply it themselves.

## Changes

- **`image-resize.ts`** — no defaults imposed by the util:
  - `type` omitted → keep the source format when the canvas can encode it (PNG→PNG, otherwise JPEG), via a new `resolveOutputType` helper.
  - `quality` omitted → pass through to the browser's native encoding quality instead of `0.85`.
  - `fit` omitted → scale the whole image to fit without cropping instead of center-cropping.
  - Updated the type docs, module doc and usage example accordingly.
- **`profile-picture.tsx`** — applies its historical defaults (`cover` / `image/jpeg` / `0.85`) in the component, so avatars are unchanged. Callers can still override via `resize`.
- **`file.e2e.tsx`** — updated the round-trip test to assert the source PNG is preserved, added tests that a non-PNG source keeps its format and that an explicit `type` still forces the format, and fixed the mid-resize sentinel.
- **Examples** — corrected the "omit → image/jpeg / cover / 0.85" comments to describe the new source-preserving behaviour.

## Behaviour change for existing `limel-file` configs

Because the util no longer supplies defaults, a `limel-file` whose `resizeImage` omits an option now behaves differently from the previous release. Beyond the headline PNG fix:

- **`type` omitted** — a PNG stays a PNG instead of being flattened to JPEG (the bug fixed here). A lossless PNG can be larger than the old JPEG.
- **`fit` omitted** (only when both `width` and `height` are set) — scales the whole image to fit without cropping instead of center-cropping to `cover`.
- **`quality` omitted** — JPEG output uses the browser's native quality (~0.92 in Chrome) instead of `0.85`, so files are somewhat larger.

This is safe in practice: the `resizeImage` prop only shipped in 39.42.0 (the previous release) and has no real consumers yet — these behaviours were found while integrating the first one.

## Why it's safe

`resizeImage` is `@beta`, so changing its defaults is not a breaking change. `limel-profile-picture`'s output is unchanged because it now sets those values explicitly.

## Testing

- `file` + `profile-picture` spec: 83 passed
- `file` e2e (real canvas / `createImageBitmap`): 4 passed
- lint clean on all changed files

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated client-side image resizing: omitted options now preserve the source format when possible, scale without cropping, and use browser-native JPEG quality.
  * Profile picture resizing now applies consistent default output settings when callers omit resize `type`/`quality`.

* **Documentation**
  * Clarified resize option semantics to reflect actual omit behavior (no implied `cover`/default JPEG assumptions).

* **Tests**
  * Strengthened end-to-end coverage for format preservation, explicit JPEG re-encoding, and cancellation behavior.

### DIFF
--- a/src/components/file/examples/file-resize-image.tsx
+++ b/src/components/file/examples/file-resize-image.tsx
@@ -22,8 +22,10 @@ import { Component, h, Host, State } from '@stencil/core';
  *   both to keep the source dimensions and only re-encode.
  * - `fit`: only matters when both `width` and `height` are set. `cover` fills
  *   the box and center-crops the overflow; `contain` fits the whole image
- *   inside the box, letterboxing it. Omit it for the default (`cover`).
- * - `type`: the output format. Omit it for the default (`image/jpeg`).
+ *   inside the box, letterboxing it. Omit it to scale the whole image to fit
+ *   without cropping.
+ * - `type`: the output format. Omit it to keep the source format when the
+ *   canvas can encode it (PNG stays PNG; everything else becomes JPEG).
  * - `quality`: JPEG compression, `0`–`1`. Ignored for `image/png`, which is
  *   lossless. The control below is disabled when PNG is selected.
  *
@@ -64,9 +66,12 @@ export class FileResizeImageExample {
     private previewUrl = '';
 
     @State()
-    // Every field is optional. When omitted, `fit` defaults to `cover`, `type`
-    // to `image/jpeg` and `quality` to `0.85`; `width`/`height` fall back to the
-    // source dimensions (a missing one preserves the aspect ratio).
+    // Every field is optional, and the util imposes no defaults of its own:
+    // when omitted, `type` keeps the source format when the canvas can encode
+    // it (PNG stays PNG, otherwise JPEG); `fit` scales the whole image to fit
+    // without cropping; `quality` (JPEG only) uses the browser's native
+    // quality; `width`/`height` fall back to the source dimensions (a missing
+    // one preserves the aspect ratio).
     private options: ResizeOptions = {
         width: 800,
         height: 600,
--- a/src/components/file/examples/file-resize-mixed.tsx
+++ b/src/components/file/examples/file-resize-mixed.tsx
@@ -36,9 +36,11 @@ export class FileResizeMixedExample {
     @State()
     private notResized = false;
 
-    // Every field is optional. When omitted, `fit` defaults to `cover`, `type`
-    // to `image/jpeg` and `quality` to `0.85`; `width`/`height` fall back to the
-    // source dimensions (a missing one preserves the aspect ratio).
+    // This example pins every field. In particular it forces `type:
+    // 'image/jpeg'` so the "Uploaded as-is" detection below can tell a resized
+    // image (always re-encoded to `.jpg`) from one that was passed through
+    // untouched. For what each option does when omitted, see the
+    // `file-resize-image` example, which exposes controls for them.
     private options: ResizeOptions = {
         width: 1024,
         height: 1024,
--- a/src/components/file/file.e2e.tsx
+++ b/src/components/file/file.e2e.tsx
@@ -1,5 +1,6 @@
 import { render, h } from '@stencil/vitest';
 import { FileInfo } from '../../global/shared-types/file.types';
+import { ResizeOptions } from '../../util/image-resize';
 
 // Browser tests: a real canvas / `createImageBitmap` is available here, so the
 // actual client-side resize runs. These cover the two behaviors that cannot be
@@ -8,11 +9,12 @@ import { FileInfo } from '../../global/shared-types/file.types';
 
 const RESIZE_OPTIONS = { width: 50, height: 50 };
 
-// Draw a solid PNG of the given size so it is a genuinely decodable raster.
+// Draw a solid image of the given size so it is a genuinely decodable raster.
 const makeImageFile = async (
     filename: string,
     width: number,
-    height: number
+    height: number,
+    type: 'image/png' | 'image/jpeg' = 'image/png'
 ): Promise<File> => {
     const canvas = document.createElement('canvas');
     canvas.width = width;
@@ -22,10 +24,10 @@ const makeImageFile = async (
     context.fillRect(0, 0, width, height);
 
     const blob = await new Promise<Blob>((resolve) => {
-        canvas.toBlob((result) => resolve(result), 'image/png');
+        canvas.toBlob((result) => resolve(result), type);
     });
 
-    return new File([blob], filename, { type: 'image/png' });
+    return new File([blob], filename, { type });
 };
 
 const asFileInfo = (file: File): FileInfo => ({
@@ -55,18 +57,18 @@ const captureChanges = (root: HTMLElement): FileInfo[] => {
     return changes;
 };
 
-const setup = async () => {
+const setup = async (resizeImage: ResizeOptions = RESIZE_OPTIONS) => {
     const { root, waitForChanges } = await render(
         <limel-file label="Attach a file" />
     );
-    (root as any).resizeImage = RESIZE_OPTIONS;
+    (root as any).resizeImage = resizeImage;
     await waitForChanges();
 
     return { root, waitForChanges, changes: captureChanges(root) };
 };
 
 describe('limel-file (client-side resize)', () => {
-    it('resizes and re-encodes a selected image before emitting change', async () => {
+    it('resizes and keeps the source format when no type is set', async () => {
         const { root, changes } = await setup();
 
         selectFile(root, await makeImageFile('photo.png', 200, 120));
@@ -74,14 +76,50 @@ describe('limel-file (client-side resize)', () => {
         await vi.waitFor(() => expect(changes).toHaveLength(1));
 
         const emitted = changes[0];
-        // Re-encoded to the default JPEG output, with the extension updated.
-        expect(emitted.filename).toBe('photo.jpg');
-        expect(emitted.contentType).toBe('image/jpeg');
+        // No output type requested, so the source PNG is preserved rather than
+        // silently flattened to JPEG.
+        expect(emitted.filename).toBe('photo.png');
+        expect(emitted.contentType).toBe('image/png');
 
         // Actually downscaled to the requested box.
         const bitmap = await createImageBitmap(emitted.fileContent);
         expect(bitmap.width).toBe(50);
         expect(bitmap.height).toBe(50);
+    });
+
+    it('keeps a non-PNG source format when no type is set', async () => {
+        const { root, changes } = await setup();
+
+        selectFile(
+            root,
+            await makeImageFile('photo.jpg', 200, 120, 'image/jpeg')
+        );
+
+        await vi.waitFor(() => expect(changes).toHaveLength(1));
+
+        const emitted = changes[0];
+        // A JPEG source with no requested type stays JPEG — the branch every
+        // non-PNG upload hits — rather than being forced to another format.
+        expect(emitted.filename).toBe('photo.jpg');
+        expect(emitted.contentType).toBe('image/jpeg');
+    });
+
+    it('re-encodes to an explicitly requested output type', async () => {
+        const { root, changes } = await setup({
+            width: 50,
+            height: 50,
+            type: 'image/jpeg',
+        });
+
+        selectFile(root, await makeImageFile('photo.png', 200, 120));
+
+        await vi.waitFor(() => expect(changes).toHaveLength(1));
+
+        const emitted = changes[0];
+        // An explicit type wins, converting the PNG to JPEG and updating the
+        // extension.
+        expect(emitted.filename).toBe('photo.jpg');
+        expect(emitted.contentType).toBe('image/jpeg');
     });
 
     it('does not re-emit the file when the chip is removed mid-resize', async () => {
@@ -99,9 +137,11 @@ describe('limel-file (client-side resize)', () => {
         );
         await new Promise((resolve) => setTimeout(resolve, 150));
 
-        // The stale resized image must never be emitted after removal.
+        // The stale resized image must never be emitted after removal. The
+        // resized PNG keeps its source format, so that is what a leaked result
+        // would look like.
         expect(
-            changes.some((change) => change?.contentType === 'image/jpeg')
+            changes.some((change) => change?.contentType === 'image/png')
         ).toBe(false);
         // The last thing the consumer heard was the removal.
         expect(changes.at(-1)).toBeFalsy();
--- a/src/components/profile-picture/examples/profile-picture-basic.tsx
+++ b/src/components/profile-picture/examples/profile-picture-basic.tsx
@@ -1,4 +1,4 @@
-import { FileInfo } from '@limetech/lime-elements';
+import { FileInfo, type ResizeOptions } from '@limetech/lime-elements';
 import { Component, h, Host, State } from '@stencil/core';
 
 /**
@@ -7,6 +7,12 @@ import { Component, h, Host, State } from '@stencil/core';
  * This component can be both used as a placeholder for an avatar,
  * and in the same time act as an interactive element that enables
  * users to upload a new profile picture.
+ *
+ * This example enables client-side resizing with only `width` and `height`
+ * set on `resize`. Every other option is left to the component's
+ * avatar-oriented defaults: the file is emitted as an `image/jpeg` at
+ * `quality` `0.85`, cropped to the `imageFit` (`cover` here). Set `type`,
+ * `quality` or `fit` on `resize` to override any of them.
  *
  * :::note
  * You must add a proper `width` and `height` to the component.
@@ -21,12 +27,17 @@ export class ProfilePictureExample {
     @State()
     private value?: FileInfo | string = undefined;
 
+    // Only width and height are set; type, quality and fit fall back to the
+    // component's avatar defaults (image/jpeg, 0.85, and imageFit).
+    private resize: ResizeOptions = { width: 400, height: 400 };
+
     public render() {
         return (
             <Host>
                 <limel-profile-picture
                     label="Profile picture"
                     value={this.value}
+                    resize={this.resize}
                     onChange={this.handleChange}
                 />
                 <limel-example-value value={this.value} />
--- a/src/components/profile-picture/profile-picture.tsx
+++ b/src/components/profile-picture/profile-picture.tsx
@@ -134,6 +134,8 @@ export class ProfilePicture {
     /**
      * Optional client-side resize before emitting the file.
      * If provided, the selected image will be resized on the client device.
+     * Omitted options fall back to avatar defaults: `image/jpeg`, `quality`
+     * `0.85`, and the component's `imageFit`.
      * :::note
      * HEIC may not decode in all browsers; when decoding fails, the original
      * file will be emitted. See the examples for more info.
--- a/src/components/profile-picture/profile-picture.tsx
+++ b/src/components/profile-picture/profile-picture.tsx
@@ -19,6 +19,14 @@ import { resizeImage, ResizeOptions } from '../../util/image-resize';
 import { ImageTemplate } from '../../util/image.template';
 
 /**
+ * Avatar output defaults. These live in the component rather than in the shared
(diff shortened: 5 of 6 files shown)


Kiarokh:
> 🤖 AI-generated review from 6 parallel agents, **updated** after the follow-up commits on this branch. Treat as input, not a verdict — agents can be wrong or miss context.

# Consolidated PR Review (updated)

## PR Summary

Removes the avatar-tuned defaults (`fit=cover`, `type=image/jpeg`, `quality=0.85`) that the shared `resizeImage` util baked in, so an omitted option now preserves the source instead of silently rewriting it — most importantly, a PNG resized with only a `width` set stays a PNG instead of being flattened to JPEG. The defaults move into `limel-profile-picture`, whose output is unchanged. Since the initial review, the branch also gained follow-up commits addressing the review: a `refactor` (single `ENCODABLE_TYPES` map), a couple of `docs` commits, and fixups (named avatar constants, test-helper dedupe, a new test).

Strictly better than `main`: it fixes the PNG-flattening bug and cleanly relocates the product defaults to the component that owns them, with no regression in any dimension. `limel-profile-picture`'s output is provably byte-identical, and every item from the initial review is now resolved or reclassified.

- **Blockers:** None.
- **Resolved on the branch since the review:**
  1. ✅ The one substantive gap — `resolveOutputType`'s non-PNG→JPEG branch was untested — is now covered by a new e2e test.
  2. ✅ The "canvas encodes only PNG/JPEG" duplication is gone (single `ENCODABLE_TYPES` map).
  3. ✅ Avatar defaults extracted to named constants and documented on the `resize` prop + basic example.
  4. ✅ Test `setup()` deduped; misleading mixed-example and module-doc comments fixed.
  5. ✅ The `fit`/`quality` behavior change for `limel-file` is now spelled out in the PR description.
- **Deferred (pre-existing, not a regression from this PR):** the `ImageBitmap` cleanup is tracked in #4179.

---

## Dimensions

**What works well:** `@beta` is confirmed in both source and the committed API report, so changing defaults is sanctioned; `ResizeOptions`'s shape is unchanged; `limel-profile-picture` is byte-identical because `imageFit` is a non-optional prop defaulting to `cover`; and the `rename` change is actually more robust than `main` (handles a `null` rename).

**Issues:** None.

**Minor nits:**
- 🔵 ✅ **Resolved** — the `fit` (cover → scale-to-fit) and `quality` (0.85 → browser-native) changes for `limel-file` are now called out in the PR description's "Behaviour change for existing `limel-file` configs" section. Safe in practice: the `resizeImage` prop only shipped in 39.42.0 and has no real consumers yet.

### 2. Code Quality — GOOD ✅

**What works well:** `resolveOutputType` is correct and type-safe; `quality?` flows cleanly into both the `convertToBlob` and `toBlob` paths; the `options.rename ?? …` form is behavior-preserving and more robust than `main`.

**Issues:** None outstanding. The initial review's one 🟡 Medium — the non-PNG→JPEG branch of `resolveOutputType` was untested because the `makeImageFile` fixture only produced PNGs — is **✅ resolved**: `makeImageFile` now takes a `type` and a new e2e (`keeps a non-PNG source format when no type is set`) exercises the JPEG-source path.

**Minor nits (all resolved on the branch):**
- 🔵 ✅ The explicit-type test now reuses a parameterized `setup()` instead of re-implementing it.
- 🔵 ✅ The misleading "when omitted" narration on the fully-pinned mixed example is replaced with an explanation of why it pins `type` for its detection logic.
- 🔵 ✅ The module doc no longer frames the util around avatars.

### 3. Architecture — GOOD ✅

**What works well:** the defaults-move is coherent — the shared util imposes no product defaults and each consumer owns its own; the key correctness question (avatar `fit` never falls through to the no-crop path) holds because `imageFit` is non-optional and defaulted.

**Issues:** None.

**Minor nits:**
- 🔵 ✅ **Resolved** — the "canvas encodes only PNG/JPEG" fact, previously duplicated across the `type` union, `resolveOutputType`, and `renameWithType`, now derives from a single `ENCODABLE_TYPES` map, so adding a format is a one-line change.
- 🔵 ✅ **Resolved** — the avatar defaults are now `AVATAR_OUTPUT_TYPE` / `AVATAR_JPEG_QUALITY` constants.
- 🔵 Intended behavior change (not a defect): a `limel-profile-picture` with `imageFit="contain"` and a `resize` that omits `fit` now letterboxes to match its preview instead of center-cropping. Default avatars (`cover`) are unaffected; no test covers this path.
- 🔵 No WebP output path (`type` union is JPEG/PNG only) — pre-existing and a reasonable cross-browser floor; the new `ENCODABLE_TYPES` map makes adding one trivial if ever needed.

### 4. Security — SAFE ✅

**What works well:** the original file bytes are never emitted — every input is canvas-re-encoded, which strips EXIF/GPS for both PNG and JPEG outputs (no privacy regression); the untrusted `file.type` is only used in an equality check and never reaches a filename, blob type, or the DOM; content-type always matches the encoded bytes; and SVG is excluded by both consumers.

**Issues:** None.

**Minor nits (pre-existing, unchanged by this PR):**
- 🔵 `renameWithType` keeps path-like characters from the original `File.name` — a client-side label only, reachable solely with a self-crafted `File`.
- 🔵 An explicitly-requested out-of-contract `type` string would be returned verbatim; constrained by the TS union and sourced from trusted component config, and it matches `main`.

### 5. Observability — GOOD ✅

**What works well:** the new code paths are pure and non-throwing; `canvasToBlob`'s null-blob rejection still fires with a clear message despite `quality` becoming optional; and the "optimizing" badge correctly reaches a finished state on failure.

**Issues:** None introduced.

**Minor nits (pre-existing):**
- 🔵 `limel-file` swallows resize failures with an empty `catch {}` and no console signal.
- 🔵 `limel-profile-picture`'s catch is also silent (mitigated by its user-facing error popover).
- 🔵 The `isDev` guard on the decode-fallback `console.debug` can evaluate true in production browser bundles (negligible; no sensitive data logged).

### 6. Performance — GOOD ✅

**What works well:** no new per-pixel work or allocations of consequence; PNG + a passed/undefined `quality` is a safe no-op; resize runs once per file selection (never on render); object-URL hygiene is intact.

**Issues:** None outstanding. The initial review's 🟡 Medium — the decoded `ImageBitmap` is never `.close()`d — is **pre-existing** (not a regression from this PR) and is now tracked in #4179 rather than blocking merge.

**Minor nits:**
- 🔵 ✅ **Resolved (documented)** — omitting `type`/`quality` on `limel-file` yields larger outputs (a PNG stays lossless PNG; omitted quality uses browser-native ~0.92 JPEG). This is the intended fix, and it's now noted in the PR description.

---

## Resolutions (initial Top Recommendations)

1. **[was 🟡 Medium — Code Quality] Test the non-PNG→JPEG branch** — ✅ Done. `makeImageFile` takes a `type`; new e2e `keeps a non-PNG source format when no type is set` covers JPEG-source → JPEG.
2. **[was 🟡 Medium — Backward Compatibility] Document the fit/quality changes for `limel-file`** — ✅ Done and reclassified to a nit. Added a "Behaviour change for existing `limel-file` configs" section to the PR description. Low real-world impact: the prop shipped in 39.42.0 with no real consumers yet.
3. **[was 🟡 Medium — Performance] Release the `ImageBitmap` after drawing** — ↪️ Deferred as pre-existing. Filed as #4179; it predates this PR and isn't a regression, so it doesn't block merge.

adrianschmidt:
> 🤖 AI-generated review from 7 parallel agents (6 dimensions + a Lime-platform LPID pass). Treat as input, not a verdict — agents can be wrong or miss context.

# Consolidated PR Review

## PR Summary

This PR removes the avatar-tuned defaults (`fit=cover`, `type=image/jpeg`, `quality=0.85`) that the shared `resizeImage` util (`src/util/image-resize.ts`) baked in, so an omitted option now preserves the source rather than silently rewriting it — most importantly a PNG resized with only a `width` set stays a PNG instead of being flattened to JPEG. The three defaults move into `limel-profile-picture`, whose output stays identical. +179 / −45 across 6 files (util, profile-picture component + example, two file examples, and the file e2e suite). Closes #4177.

This PR is strictly better than `main`: it fixes the PNG-flattening / transparency-loss bug and cleanly relocates product defaults into the component that owns them. No regression was found in any of the seven dimensions — `limel-profile-picture`'s output is provably unchanged (verified against the committed `etc/lime-elements.api.md`), the public API shape is byte-stable, and the untrusted `file.type` is safely gated. All findings are non-blocking.

- **Blockers:** None.
- **Non-blocking but worth addressing:** one `[Medium]` test-coverage gap (the WebP/GIF/HEIC → JPEG fallback branch is untested and the "non-PNG branch is tested" claim is inaccurate) plus a handful of `[Low]` doc/dedup nits — see Top Recommendations.

---

## 1. Backward Compatibility — SAFE ✅

No public-API regression. Verified against source and the committed public API report (`etc/lime-elements.api.md`) that no exported shape changes.

- **[Low]** The `type`/`fit`/`quality` behavior change is real for any existing `limel-file` consumer that configured `resizeImage` (PNG now stays PNG; omitted quality → browser-native ~0.92). But it's fully documented in the PR description and code, the `resizeImage` prop is genuinely `@beta` (`file.tsx:145` + confirmed in `lime-elements.api.md:461`), and it only shipped in the immediately-preceding release (39.42.0). Documented `@beta` change → not breaking.
- **[Low]** The util itself is publicly exported (`resizeImage` / `ResizeOptions`); external callers omitting options get the new pass-through. Both are `@beta` in the API report — acceptable; noted for completeness.
- **Positives:** `ResizeOptions` shape is unchanged; `imageFit` is a non-optional prop defaulting to `cover`, so profile-picture never reaches the new no-crop path and its output (incl. `quality: 0` edge case) is preserved; `computeRects`/`canvasToBlob` are non-exported, so widening their params is internal-only. The `options.rename ?? …` form is also more robust than `main` (survives a `null` rename).

## 2. Code Quality — GOOD ✅

Clean, well-documented refactor; naming, cohesion, type-safety, and error handling are solid. One reachable untested branch.

- **[Medium]** The non-encodable → JPEG fallback of `resolveOutputType` (`src/util/image-resize.ts:493`) is untested. The three e2e tests cover PNG→PNG, JPEG→JPEG (both the *keep-source* branch), and explicit-type-wins — none feeds a source whose `file.type` is absent from `ENCODABLE_TYPES` (WebP/GIF/HEIC), which is the `? … : 'image/jpeg'` fallback. This path is reachable in production: `file.tsx`'s `isResizableImage` (`:313`) admits any `image/*` except SVG. Note the test comment "the branch every non-PNG upload hits" (`file.e2e.tsx:101`) is inaccurate — JPEG is encodable, so that test exercises keep-source, not the fallback. A WebP fixture is producible in Chromium via `canvas.toBlob(resolve, 'image/webp')`, so the gap is closeable. Code works → aspirational, does not block.
- **[Low]** `setup(resizeImage: ResizeOptions = …)` (`file.e2e.tsx:60`) — the param name shadows the `resizeImage` util concept; it's actually the prop value. Minor readability nit.
- **Positives:** `resolveOutputType`'s `hasOwnProperty`-gated cast is runtime-safe and the cast is necessary; optional `quality` flows correctly as `undefined` into both canvas APIs; avatar defaults are correctly re-applied so output is unchanged.

## 3. Architecture — GOOD ✅

The core move — shared util imposes no product defaults, each consumer owns its own — is coherently and correctly applied. Two-consumer model confirmed by grep (only `file.tsx` passes through, only `profile-picture.tsx` owns defaults).

- **[Low]** Doc/behavior mismatch (`src/util/image-resize.ts:18-25`, `:144-151`, `:423-424`). The docs present omitted `fit` ("scales the whole image to fit without cropping") as distinct from `contain` ("letterboxing if needed"), but `computeRects` (`:426-468`) branches only on `fit === 'cover'` — `'contain'` **and** `undefined` fall through to the identical letterbox branch. The behavior is sound (the intended fix over `main`'s crop-on-omit); only the documented three-way distinction is fictional. Collapse the doc language or make the behaviors actually differ.
- **[Low]** Format-knowledge duplication: `ResizeOptions.type: 'image/jpeg' | 'image/png'` (`:157`) is a hand-maintained duplicate of the `ENCODABLE_TYPES` keys, guarded only by a "keep the two in sync" comment. Since `EncodableImageType` is declared first, `type?: EncodableImageType` would make the map the single source of truth (inlines to the same union in the `.d.ts`, no new public symbol).
- **Positives:** Defaults land in the right layer (consumer owns policy, util owns mechanism); `ENCODABLE_TYPES` collapses the scattered "canvas encodes only PNG/JPEG" fact into one place; boundary fallbacks handled cleanly without threading opinions through helpers.

## 4. Security — SAFE ✅

No new exposure or injection path. The untrusted `file.type` can only leave `resolveOutputType` as one of two in-contract constants.

- No findings.
- **Positives:** The `Object.prototype.hasOwnProperty.call(ENCODABLE_TYPES, sourceType)` gate (`:488`) rejects prototype keys (`__proto__`, `constructor`, …), so an attacker-crafted `File.type` cannot leak through the `as EncodableImageType` cast; every sink (blob MIME, filename extension via `ENCODABLE_TYPES[type]`, emitted `contentType`) derives from the gated value; canvas re-encoding continues to strip EXIF/GPS on every path including the new PNG-passthrough; the chip filename is rendered as escaped text, not `innerHTML`. `renameWithType`'s path-char handling is byte-identical to `main` and only reachable with a self-crafted File.

## 5. Observability — GOOD ✅

No observability regression: no logging removed, no error path made more silent, no sensitive data added to logs.

- **[Low]** Both `resizeImage` call sites swallow rejections into an empty/comment-only `catch` (`file.tsx:293-296`, `profile-picture.tsx:435-438`) with no diagnostic signal — a silent resize failure just emits the original file. This is entirely pre-existing (the diff doesn't touch `file.tsx`; in `profile-picture.tsx` it only adds args inside the existing `try`) and is documented fallback intent. Not a regression; flagged only as the weakest point the PR sits adjacent to.
- **Positives:** `canvasToBlob`'s null-blob rejection (`Failed to create blob from canvas`) still fires despite `quality` becoming optional; the `isDev`-guarded decode-fallback `console.debug` is preserved; docs referencing removed defaults were corrected rather than left stale.

## 6. Performance — GOOD ✅

The decode/draw/encode hot path, canvas lifecycle, and object-URL hygiene are byte-for-byte unchanged from `main`; only O(1) option resolution was added ahead of the existing single pass.

- **[Low]** Documented output-size increase: omitted `type` keeps PNG lossless (can exceed the old JPEG) and omitted `quality` uses browser-native (~0.92) JPEG. Explicitly called out in the PR description and docs; consequences are exactly as described; `limel-profile-picture` re-imposes `image/jpeg`/`0.85` so avatar size is unchanged. Documented intent → Low.
- **[Low]** The decoded `ImageBitmap` from `loadSource` is drawn but never `.close()`d (`:198`/`:219`) — pre-existing, verified untouched by this diff, and tracked in #4179. Not introduced or worsened here.
- **Positives:** `resolveOutputType`'s one `hasOwnProperty` call per file-selection (never per render — confirmed at the `handleNewFiles` call site) is negligible; `ENCODABLE_TYPES` is a module constant, no per-call allocation.

## 7. Lime Platform Issues — SAFE ✅

The LPID catalogue was fetched successfully. The changed code is pure browser-side canvas processing (StencilJS/TS) with no new network, async, mutation, lifecycle, or DOM-sink surface, so no backend platform rules apply.

- No findings. Frontend rules (FE-1…FE-11) were each triaged against the diff and confirmed not to fire: no HTTP/fetch (FE-1–4, 11), no blocking lifecycle method (FE-5), profile-picture builds a **new** object via spread rather than mutating `@State`/props (FE-6/FE-9), no state update in render (FE-7), no listeners/timers added so no `disconnectedCallback` gap (FE-8), no `innerHTML`/`eval` (FE-10). The pre-existing `ImageBitmap` cleanup gap (#4179) is not an FE-8 case and is not worsened.
- **Positives:** The change keeps type resolution pure and allocation-free and pushes defaults to the consumer without adding any async, network, mutation, or lifecycle surface.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Code Quality | **GOOD** |
| Architecture | **GOOD** |
| Security | **SAFE** |
| Observability | **GOOD** |
| Performance | **GOOD** |

## Top Recommendations

1. **[Medium from Code Quality]** Add an e2e that feeds a WebP (or GIF) source with no `type` set and asserts the output is `image/jpeg` — this is the only untested branch of `resolveOutputType` and the most behavior-changing one, reachable in production via `limel-file`. Also correct the misleading `file.e2e.tsx:101` comment ("the branch every non-PNG upload hits") — the JPEG-source test hits keep-source, not the fallback.
2. **[Low from Architecture]** Fix the fictional fit doc distinction: omitted `fit` and `contain` are behaviorally identical in `computeRects` (only `cover` crops). Either collapse the doc language or make the behaviors differ.
3. **[Low from Architecture / Code Quality]** Make `ENCODABLE_TYPES` the single source of truth by typing `ResizeOptions.type?` as `EncodableImageType` (inlines to the same union, no new public symbol), retiring the "keep the two in sync" comment.
4. **[Low from Code Quality]** Rename the `setup(resizeImage: …)` parameter (`file.e2e.tsx:60`) to something like `resizeOptions` to avoid shadowing the util concept.

Non-blocking pre-existing items the PR merely sits near — the silent `catch` at both call sites (Observability) and the `ImageBitmap` cleanup gap (Performance, tracked in #4179) — are not introduced by this PR and don't gate merge.

adrianschmidt (approved):

PR #4180, opened 2026-07-09
fix(text-editor): size inline images given bare-pixel dimensions

FredrikWallstrom:
## What

Inline-image dimensions can be expressed as bare pixel counts (`166`, the HTML `` form) or as CSS lengths (`166px`, `50%`). `applyImageStyles` assigned the value straight to `style.width` / `style.height`, so a bare number was an invalid inline style and silently ignored — the image rendered at its intrinsic size (too big).

This normalizes bare pixel counts to `px`; values that already carry a unit (`166px`, `50%`) are left untouched.

Closes https://github.com/Lundalogik/crm-client/issues/1148

## Test

Added `applyImageStyles` unit tests: a bare pixel count gains a `px` unit; a value that already has a unit is left as-is.


john-traas (approved):
Good catch!

PR #4181, opened 2026-07-10
feat(text-editor): deprecate legacy image handling in favor of inlineImages

FredrikWallstrom:
The `inlineImages` prop now owns the whole inline-image lifecycle — paste, upload, storage format, and URL resolution — so the older event-based API it replaces is on the way out.

Marks these as deprecated, all pointing at `inlineImages`:

- `imagePasted` event and its `ImageInserter` interface (the manual paste-handling hook)
- `metadataChange` event and its `EditorMetadata` interface
- `imageRemoved` was already deprecated; its notice pointed at `metadataChange`, which is now also going away, so it's repointed at `inlineImages` too

No runtime behavior changes — this is JSDoc plus the regenerated API report.

## Summary by CodeRabbit

* **Documentation**
  * Updated developer-facing deprecation notes for legacy image and metadata-related editor events.
  * Marked older image/metadata interfaces as deprecated and clarified that inline image handling is managed via the `inlineImages` prop.
  * Clarified that deprecated metadata/link behavior has no direct replacement yet and is planned for future removal.
* **Chores**
  * Added maintenance guidance comments to support future event removal.

### DIFF
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -1025,7 +1025,7 @@ export interface DockMenu {
     };
 }
 
-// @alpha (undocumented)
+// @alpha @deprecated (undocumented)
 export interface EditorImage {
     fileInfoId: string;
     src: string;
@@ -1066,7 +1066,7 @@ export type EditorMenuTypes = (typeof EditorMenuTypes)[keyof typeof EditorMenuTy
 // @public (undocumented)
 export const editorMenuTypesArray: EditorMenuTypes[];
 
-// @alpha
+// @alpha @deprecated
 export interface EditorMetadata {
     images: EditorImage[];
     links: EditorLink[];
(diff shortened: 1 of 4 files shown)


etc/lime-elements.api.md
```diff
@@ -1282,7 +1282,7 @@ interface Image_2 {
 }
```
john-traas, line 1283: I could be wrong, but the new functionality was only about handling images. This interface also handles links. I'm not sure if we should deprecate it yet?

john-traas (dismissed):

etc/lime-elements.api.md
FredrikWallstrom: John and I talked this through. Honestly, EditorMetadata wasn't very well thought through when it first landed, and nothing at Lime Technologies actually consumes it. Since it's alpha, we're comfortable deprecating it now — and if we ever want it back, we can just un-deprecate it. We can design a proper interface for this later if a real need shows up.

On links specifically: I'm not sure consumers actually need to know when a link is pasted. If it turns out they do, a dedicated linkPasted event would probably be cleaner than folding it into metadata.

adrianschmidt (approved):

PR #4196, opened 2026-08-03
fix(markdown): keep tables visible when column widths live in colgroup

john-traas:
## Summary

- Allows `colgroup`/`col` through the markdown/HTML sanitizer so column-width metadata survives sanitization (`width`/`span` attributes are already covered by the schema's global attribute allowlist)
- Removes a zero `width` (any spelling: `0px`, `0`, `.0px`, `-0px`, …) from the inline style of `table` elements during style sanitization

## Why

Emails composed by pasting a Google Sheets table into Gmail produce markup where the `<table>` element carries `style="table-layout:fixed; width:0px"` and all real column widths live in `<colgroup><col width="...">`. The sanitizer stripped `colgroup`/`col` (not in the tag allowlist) while the style allowlist kept `width` — leaving an auto-layout table with `width: 0px` and no column definitions.

Measured in Chromium with the reported email's markup: that combination squeezes the table to min-content — every column crushed to its longest word, cells wrapping on every word (64×45px for a fragment that renders 352×22px in the original email). The content survives but the table is unreadable. Removing the zero width restores natural sizing; restoring `colgroup` renders the author's column widths faithfully. A zero width on a table can only ever harm rendering (auto layout already floors tables at min-content), so dropping it is safe for all senders.

## Test plan

- [x] New specs: `colgroup`/`col` preserved through `markdownToHTML`, including `width`/`span` attributes
- [x] New specs: zero widths in all spellings removed from `table` style; non-zero widths, `calc()` values, and non-table elements untouched
- [x] Regression spec with the spreadsheet-style markup shape from the reported email (zero-width fixed-layout table + colgroup widths)
- [x] Full markdown suite passes: `npx vitest run src/components/markdown --project spec` (164 tests)
- [x] Rendering verified in Chromium against the reported email's actual sanitized output: table renders 353×27px with faithful column widths (previously 64×45px min-content squeeze)

## How to test manually

1. Check out this branch and start the docs site: `npm start` → http://localhost:3333.
2. Go to **Components → Markdown** and scroll to the **HTML tables with column definitions** example.
3. **Expected:** both renders show a readable two-column table — a narrow right-aligned numbered column (59px) and a wide content column (292px), text on one line. Inspecting the table shows the `<colgroup>` preserved and no `width: 0px` in its inline style.
4. Unit-level check: `npm run test:spec -- src/components/markdown`.

## Summary by CodeRabbit

* **New Features**
  * Improved Markdown-rendered HTML tables with support for column groups, column widths, and column spans.
  * Added safer handling for invalid, excessive, or conflicting table column definitions.
  * Added an example demonstrating spreadsheet-style HTML tables with column widths.

* **Bug Fixes**
  * Prevented invalid table columns and unnecessary zero table widths from affecting rendered tables.

* **Documentation**
  * Added the HTML table example to the Markdown component examples.


Befkadu1 (commented):
## PR Review: fix(markdown): keep tables visible when column widths live in colgroup

**Intent**: Stop the sanitizer from destroying spreadsheet-pasted tables by allowing `colgroup`/`col` through, and drop the harmful `width: 0px` that Google Sheets puts on the `<table>` element itself.
**Scope**: +200/−7 across 6 files

### Findings

**Backward compatibility**: ✅ — no public API touched
  - `sanitizeStyleValue` gains an optional trailing param and has no callers outside `sanitize-style.ts` (`markdown-parser.ts` imports only `sanitizeStyle`). `markdownToHTML`/`sanitizeHTML` are internal, not package exports.
  - Output HTML changes for affected content, but only in the direction of the fix; both pipelines (`markdownToHTML` and `sanitizeHTML`) share `getWhiteList`, so they stay consistent.

**Code quality**: ✅ — thorough tests, and the diff even cleans up as it goes
  - Test coverage is genuinely good: zero-width spellings (`.0px`, `-0px`, `0e0px`, unitless), conservative keeps (`calc(0px)` → kept, since `parseFloat` returns `NaN`), non-table elements untouched, plus an integration regression test with the real Gmail/Sheets markup shape.
  - One leftover: the PR removed the stray empty JSDoc block above `sanitizeStyleValue` but its twin above `normalizeBackgroundColor` remains (`sanitize-style.ts:69-81`). Pre-existing, not a blocker — just a two-line deletion if you're touching the file anyway.

**Architecture**: ✅ — rule lives in the right place
  - Threading `tagName` into the style sanitizer is the minimal change; the table-specific predicate is isolated in `isCollapsedTableWidth` with a comment explaining why a zero table width can only harm rendering. No premature rules-engine abstraction for a single case — good call.

**Security**: ✅ — no new attack surface
  - `colgroup`/`col` are inert layout elements with no scripting capability, and their `width`/`span` attributes were already permitted by the default schema's `'*'` allowlist (verified in `hast-util-sanitize/lib/schema.js`) — this PR adds tags, not attributes. The style change only *removes* a value.

**Observability**: ✅ — N/A; no new failure paths. The existing `try/catch` with `console.error` in `sanitizeStyleValue` is unchanged and still covers the new filter.

**Performance**: ✅ — one extra array `.filter()` pass per styled element; negligible.

### Merge readiness: READY WITH NITS

The PR is strictly better than main: it fixes a real rendering bug with a well-documented root cause, the fix is conservative (unparseable widths like `calc()` are kept, only literal zeros on `table` elements are dropped), and all CI checks pass. The only nit is the leftover empty JSDoc block above `normalizeBackgroundColor` in the same file the PR already cleaned — worth deleting while there, but not worth blocking a merge.

One thing worth checking (not asserting as a problem): `min-width` is also in the CSS allowlist, so a `min-width: 0` on a table passes through — that's harmless (it's effectively the default), so scoping the rule to `width` alone looks intentional and correct.

---
🤖 Review generated with [Claude Code](https://claude.com/claude-code)

Befkadu1 (commented):
## Follow-up: deeper verification pass — revising verdict to CHANGES REQUESTED

After posting the review above, a deeper verification pass (live probes against `markdownToHTML`/`sanitizeHTML` on this branch, plus Chromium measurements) surfaced issues the first pass missed. Three are worth addressing before merge; repros included.

### 1. The zero-width removal un-hides intentionally hidden email tables

`<table style="width:0;max-height:0;overflow:hidden"><tr><td>preheader text</td></tr></table>` now sanitizes to `<table style="">…`. The CSS allowlist already strips `display`, `max-height`, and `overflow`, so `width: 0` was the **only** hiding mechanism that survived this sanitizer — and this PR removes it. Hidden spacer/preheader tables common in email HTML will now render as visible full-width blocks (`_tables.scss` forces `table { display: block }` unconditionally, including in `no-table-styles` mode). Worth scoping the removal to the actual Sheets signature — zero width *plus* a `colgroup` with real widths — rather than every zero-width table.

### 2. Whitelisting `colgroup`/`col` opens an unbounded `span` amplification

`<col span="2147483647">` passes through verbatim — `span` rides along via the `'*'` attribute allowlist. ~20KB of stacked `<col span="1000">` elements yields ~10⁶ column boxes, which hangs the renderer on untrusted content; pre-PR, every `col`/`colgroup` was stripped on this path. Suggest granting the attributes explicitly (`col: ['width', 'span']`, mirroring `sanitize-email-html.ts`) and capping `span`.

### 3. The two pipelines now diverge on orphan colgroup — one-line fix

`markdownToHTML('<colgroup><col width="9"></colgroup>')` returns the markup verbatim, while `sanitizeHTML` of the same input returns `''` — breaking the invariant documented in `markdown-parser.ts` itself that both paths behave identically. The default schema constrains the other table-internal tags via `ancestors`; `colgroup`/`col` need the same:

```ts
ancestors: { ...defaultSchema.ancestors, colgroup: ['table'], col: ['table', 'colgroup'] }
```

### 4. Smaller gaps (non-blocking)

- `width: calc(0px)` still collapses the table (`parseFloat` → `NaN` → kept), and `sanitize-style.spec.ts` pins that as *correct* with a green test.
- The regression spec asserts `not.toContain('width: 0px')` (with a space) against input written `width:0px` (no space), so it passes vacuously if the sanitize visitor ever stops running. `not.toMatch(/width:\s*0/)` would be load-bearing, and one test through the `sanitizeHTML` entry point would guard the both-paths invariant.
- The new example publicly documents `no-table-styles`, which `_tables.scss:8-9` explicitly records as internal-only with no public documentation.

### Design note (take or leave)

The underlying crush is caused by the allowlist stripping `table-layout`: under `table-layout: fixed`, `width: 0px` is harmless and the `col` widths render **exactly** (59/292 measured in Chromium — the email-viewer path proves this in-repo, since it preserves both). Deleting the width restores visibility but demotes the col widths to auto-layout hints, so unbreakable cell content still distorts the columns. An alternative: allow `table-layout` through, and/or implement the table fixup as a rehype plugin alongside `link-markdown-plugin`/`image-markdown-plugin` — at node level the rule can see the whole element (style + colgroup children) instead of threading `tagName` into `sanitizeStyleValue`, whose optional parameter fails open for any future caller that omits it.

### Revised verdict: CHANGES REQUESTED

Items 1–3. The core direction is right and the test discipline is good — the fixes are small scoping and schema tweaks, not a rethink.

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

Befkadu1 (approved):
Awesome ⚡

PR #4213, opened 2026-08-11
fix(form): allow clearing a field that has a schema default

jgroth:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed custom component fields so clearing them correctly leaves their value undefined, including after clearing another field.

* **Tests**
  * Added end-to-end coverage for clearing custom component fields.
  * Updated schema validation tests to use the shared additional-property marker.

* **Chores**
  * Updated RJSF development dependencies to version 6.7.1.
  * Grouped related dependency updates for easier maintenance.

Upgrades `@rjsf/core`, `@rjsf/utils` and `@rjsf/validator-ajv8` from 6.6.1 to 6.7.1.

On 6.6.1 a field with a schema `default` could not be cleared. Removing the value
restored the default at once, so an optional field could never be emptied. Custom
components suffered most: a picker rebuilt its chip from the restored value, which
left the user unable to pick anything else. The bug is
[rjsf #5125](https://github.com/rjsf-team/react-jsonschema-form/issues/5125), fixed
upstream in 6.7.0.

Two e2e tests cover it. Both fail on 6.6.1 and pass on 6.7.1.

Worth a look during review: `ADDITIONAL_PROPERTY_FLAG` changed from a string to a
`Symbol` in 6.7.0, which the release notes do not mention. `field-helpers.ts` reads
the imported constant and keeps working, so only the spec needed a change because it
hardcoded the old string. Code outside this repo that reads that flag as a literal
string will break. Say if you want the commit marked as breaking.

Dependabot never offered this upgrade because `@rjsf/core` peer-requires a matching
`@rjsf/utils` and cannot bump them one at a time. The second commit groups them so
future updates arrive on their own.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


Befkadu1 (approved):
I tested the PR in my PR [here](https://github.com/Lundalogik/lime-crm-components/pull/4573) and works as expected ⚡

adrianschmidt (approved):
Changes to `.github/dependabot.yml` approved 👍

PR #4223, opened 2026-08-13
Clear the picker's search query when an item is picked

devbymadde:
Picking an item in a `multiple` picker has kept the typed search query since 39.27.0, leaving the dropdown filtered by a query the user had already acted on. It was not popular, so it is removed rather than made configurable: a pick ends the search session, as it did before 39.27.0.

Removing a chip or clearing all now does the same. That previously reset only the picker's internal `textValue` and left the chip-set displaying what the user had typed, so the two disagreed about what was being searched for.

The refreshed search goes through the picker's existing debounce rather than running at once. Consumers commonly drop the picked item from `allItems` in response to the `change` event, and that arrives as a prop on their next render — searching immediately reads the list from before, so the item the user just picked would be offered again. The refresh is also skipped when the removal moved focus out of the input, or when the dropdown was already dismissed, so neither costs a round trip for a result nobody would see and Esc is not undone by the removal that follows it.

## Scope

This PR was much larger for a while. Reviewing it turned up four older bugs in the same code path — none of them regressions, most of them years old — and fixing them here made the change hard to review and hard to release on its own. They are filed separately instead, with the reproductions from this PR's review threads:

- #4268 — already picked items are not excluded from search results
- #4269 — a late search result reopens a dismissed dropdown
- #4270 — a failing searcher leaves the loading spinner up
- #4271 — removing a chip can duplicate one item and drop another

Working fixes for all of them are on `fix/picker-known-bugs`, tested and reviewed, for whoever picks them up. #4272 is closed — going through the debounce fixed it here.

## Backward compatibility

No API change: `etc/lime-elements.api.md` is identical to `main`, and no prop is added, removed or retyped.

One behavior change for the release notes: **the query is no longer kept after a pick**, with no opt-out. Apps that came to rely on the 39.27.0 behavior lose it.

## Verification

13 picker e2e tests, 31 including `limel-chip-set`, lint clean. Every behavior is covered by a test that was checked against a deliberately reverted implementation to confirm it fails without the fix.

Two earlier versions of this PR introduced regressions of their own, both caught in review and both now covered by a test: Esc followed by a Backspace removal reopened the dropdown, and the immediate refresh offered a just-picked item again for consumers whose `allItems` update lands on their next render. Each was verified against `main` with the same test to confirm it really was a regression rather than something inherited.

Not run: the `test:examples` suites, and no browser but Chromium. Worth testing by hand is **removing a chip with the mouse** — Chromium focuses the remove button and the suggestions are deliberately not refreshed, while Firefox and Safari do not focus buttons on click, so the behavior differs there.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved picker behavior when selecting, removing, or clearing items, including more reliable search input and suggestion updates.
  * Multi-select pickers now restore default suggestions while excluding selected items.
  * Prevented unnecessary searches and dropdown reopening after chip removal or dismissal.
  * Improved consistency between mouse and keyboard selection, including avoiding duplicate searches.
  * Pressing Esc now clears the query and immediately restores default suggestions.

* **Documentation**
  * Clarified guidance for managing search text when picker selections change.


src/components/picker/picker-helpers.ts
```diff
@@ -0,0 +1,38 @@
...
+ * The id that identifies a picked item across the picker/list boundary.
+ *
+ * For object values, the mandatory `id` of `PickerValue` is used. Plain
+ * `string`/`number` values are their own id. Items without a value have
+ * no id at all.
+ *
+ * @param item - the item to get an id for
+ * @returns the value id, or `undefined` if the item has no value
+ */
+export function getValueId(item: PickerItem): string | number | undefined {
```
Befkadu1: What do you think of passing the value instead of `PickerItem` ?

src/components/picker/picker-helpers.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+ * @param item - the item that was picked
+ * @returns a copy of the item, flagged as selected
+ */
+function asPicked(item: PickerItem): PickerItem {
+    return { ...item, selected: true };
+}
+
+describe('getValueId', () => {
+    it('returns a primitive value as-is', () => {
+        expect(getValueId({ text: 'Blueberry', value: 1 })).toBe(1);
```
Befkadu1:
```suggestion
        expect(getValueId(blueberry)).toBe(1);
```

src/components/picker/picker-helpers.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+import { PickerItem } from './picker-item.types';
+import { getValueId, hasPickableItems } from './picker-helpers';
+
+const blueberry: PickerItem<number> = { text: 'Blueberry', value: 1 };
+const header: ListSeparator = { separator: true, text: 'Results matching "b"' };
+
+describe('getValueId', () => {
+    it('returns a primitive value as-is', () => {
+        expect(getValueId({ text: 'Blueberry', value: 1 })).toBe(1);
+        expect(getValueId({ text: 'Blueberry', value: 'one' })).toBe('one');
```
Befkadu1: This test doesn't have much value

src/components/picker/picker-helpers.ts
```diff
@@ -36,3 +36,58 @@ export function hasPickableItems(
...
+ *   from them — so colliding ids misbehave in the same way elsewhere.
+ *
+ * @param items - the items to filter
+ * @param picked - the items that have already been picked. May contain
+ * nullish holes, since the picker derives it from `value`, which is
+ * consumer-owned and which `handleChange` can leave holes in.
+ * @returns the items that have not been picked, `ListSeparator` entries
+ * included
+ */
+export function withoutPickedItems(
```
Befkadu1: This function name is not clear what it does and would be great if we follow  verb-first as a prefix of the function. Maybe `excludePickedItems`?

src/components/picker/picker.tsx
```diff
@@ -180,6 +182,37 @@ export class Picker {
...
+     * so they don't have to retype it between picks.
+     *
+     * :::note
+     * Upgrading: from version 39.27.0 until this prop was introduced,
+     * the query was always kept when an item was picked, and there was no
+     * way to opt out. If your app came to rely on that, set this prop to
+     * `true` to keep it. Note that the query is now always cleared when
+     * the *chips* change — a chip is removed, or "Clear all" is pressed —
+     * regardless of this prop, since a chip change ends the search
+     * session.
```
Befkadu1: It looks this is a big doc. If we keep on adding big documentation for all Props, we will end up having a very big file.Maybe minor trimming?

src/components/picker/picker.tsx
```diff
@@ -611,18 +673,22 @@ export class Picker {
...
-        // If the search function resolves immediately,
-        // the loading spinner will not be shown.
-        clearTimeout(timeoutId);
-
-        this.handleSearchResult(query, result);
+        try {
+            const result = await searcher(this.textValue);
+            this.handleSearchResult(query, result, justPicked);
+        } catch {
+            this.handleSearchResult(query, [], justPicked);
```
Befkadu1: the failing searcher fix swallows the rejection entirely . Maybe better to have console.error(error) in the catch so a developer can still see their searcher threw?

src/components/picker/picker.tsx
```diff
@@ -180,6 +182,37 @@ export class Picker {
...
+     * the query was always kept when an item was picked, and there was no
+     * way to opt out. If your app came to rely on that, set this prop to
+     * `true` to keep it. Note that the query is now always cleared when
+     * the *chips* change — a chip is removed, or "Clear all" is pressed —
+     * regardless of this prop, since a chip change ends the search
+     * session.
+     * :::
+     */
+    @Prop()
+    public keepSearch = false;
```
Kiarokh: I feel this part of the PR is undoing a fix for a **problem** that I was trying to solve in an earlier PR https://github.com/Lundalogik/lime-elements/pull/4086 . So this brings the previous problematic behaviour back, but offers what we considered a better solution, under a new prop. Maybe it is good to add flexibility. But looking at the issue https://github.com/Lundalogik/crm-client/issues/1221 this PR is trying to solve, I feel uneasy. 

This is additionally itchy, because the docs say:

>  * :::note
>  * `keepSearch` only has an effect when `multiple` is `true`, since a
>  * single-pick picker is full — and its search session therefore over —
>  * as soon as an item is picked.
>  * :::

A prop that _sometimes_ has an effect (depends on another boolean prop), is always a signal for me to wonder if we are doing the right thing. The decision doesn't have to be wrong, but it may.

We have discussed this internally why this interaction design was considered better from my perspective. In summary, it doesn't force the user to retype the same thing. In many scenarios, the whole point with a multi select interaction is letting users pick many items containing the same word e.g. "lime.tech". 

But I can also understand that using this component as a "To" field (recipient field) in an email composer UI, can cause friction for some users. It is also unconventional design, compared to Gmail and Outlook.

Pressing <kbd>Esc</kbd> to clear the typed query exists, but it is a per-interaction manual step and poorly discoverable, so I understand why a config-level answer felt necessary

We all know props are cheap to add, but nearly impossible to remove. I'm wondering… instead of adding a new prop, could we: 

**1. Keep the search, but even easier to clear**
- Make the typed query `selected`, as soon as one suggestion is selected from the list? This means we keep it, to keep displaying the same search results, but users can either: 
  - just type a new query without pressing Esc or Backspace; 
  - or they can press the right arrow key, to keep typing more to what they have already typed
- Ensure that the typed query is still cleared, as soon as the focus has moved away from the input field (as it happens today)

**2. Add the new prop**
If we decide to keep the `keepSearch` prop, then instead of initiating it as `false`, we may be better to set it to `true` by default, and then set it to `false` in the email composer package, where the current interaction design hurts the most.

I think the name `clearSearchOnSelect` or `clearSearchOnPick` is more clear. (the of course we should initiate it accordingly based on our decission. Or rename it to `keepXYZ` instead of `clearXYZ`

*******
At this point, I have a hard time to decide, because we are catering for multiple scenarios demanding different UX. But I think we should try my first suggestion, before adding new props. 

What do you think?

p.s.
If we prototype my suggestion, the two things worth checking are how the selection behaves on touch, where the OS draws selection handles that read as user-initiated, and whether it survives focus leaving and returning, since a collapsed selection silently turns "type to replace" back into "type to append".

adrianschmidt: I agree. We should only add the prop if we genuinely see use cases where you definitely want to keep the search _and_ use cases where you definitely want to clear it. _And the difference has to be programmatically distinguishable._

We should not add it just because we're afraid to impose a certain behaviour. Lime Elements isn't meant to be a deeply configurable component library, it's meant to be a library that sets the standard for how things look, feel, and behave in the Lime product line.

Kiarokh: For the records @devbymadde and I discussed a bit more elsewhere. For the moment, it feels the best way to move forward is to revert the interaction design which was introduced here https://github.com/Lundalogik/lime-elements/pull/4223 and add no new props. 

If keeping the query in future is identified as a big-enough need, we can discuss how to fix it. By a boolean prop e.g. `clearSearchOnSelect`? Or by some more elaborate prop e.g. `seachInteraction: x | y | z = x`

Thanks for the thought exchange 🙏 and great work

src/components/picker/picker.tsx
devbymadde: Good point, I've shortened jsdocs in general

src/components/picker/picker.tsx
devbymadde: Fixed!

src/components/picker/picker.tsx
```diff
@@ -702,27 +770,34 @@ export class Picker {
...
         if (this.multiple) {
             const chips = event.detail as Chip[];
-            newValue = chips.map((chip) => {
-                return (this.value as PickerItem[]).find((item) => {
-                    const valueId = this.getValueId(item);
-
-                    return `${valueId}` === chip.id;
-                });
-            });
+            newValue = chips.map((chip) => chip.value as PickerItem);
```
john-traas: Removing a chip can still lose a pick that the consumer hasn't confirmed yet. The chips are always built from `value`, never from `emittedValue`, and this line rebuilds the emitted value from those chips. So if the user picks A and then removes chip B before the consumer echoes the new value back, the removal emits the old value minus B — and A is gone. It also overwrites `emittedValue`, so A is lost for good.

Suggestion: read the change as “these chips were removed” and apply that to `pickedItems()` — a pending pick has no chip yet, so keeping only the remaining chip ids would drop it.

Reproduced against this branch running in lime-webclient, with a consumer that echoes `value` after 800ms: value `[Anna, Bertil]`, picked Cesar (emitted `[Anna, Bertil, Cesar]`), X-clicked Bertil's chip 200ms later → the removal emitted `[Anna]`, and the settled value ended up `[Anna]` instead of `[Anna, Cesar]`.

src/components/picker/picker.tsx
```diff
@@ -702,27 +770,34 @@ export class Picker {
...
 
-        this.change.emit(newValue as PickerItem | PickerItem[]);
+        this.emitValue(newValue as PickerItem | PickerItem[]);
+
+        const wasEditingInput = await this.chipSet?.getEditMode();
+
+        if (this.multiple && wasEditingInput) {
+            this.resetSearchToDefault();
+        } else {
+            this.clearInputField();
```
john-traas, line 736: The two ways of removing a chip behave differently. Backspace keeps `emittedValue`; X-click and clear-all end up in `clearInputField`, which nulls it — right after this handler just emitted a value. So the removal is immediately forgotten: pick something before the consumer catches up, and the removed item comes back. I understand “session end defers to `value`” is deliberate, but here the session ends inside the same gesture that emitted.

Reproduced with a 1500ms-echo consumer: X-clicked Bertil (emitted `[Anna]`), picked Cesar before the echo landed → emitted `[Anna, Bertil, Cesar]` — the removal was undone. The suggestion list at pick time also still excluded the just-removed Bertil, the “just-removed item filtered straight back out” case from the PR description.

john-traas (commented):
Reviewed the branch and verified each scenario against it running in lime-webclient — details inline. A locally verified patch covering the correctness findings, plus a testing note, follow as separate comments.

src/components/picker/picker.tsx
```diff
@@ -296,38 +327,41 @@ export class Picker {
...
+                    {...props}
+                />
+                {this.renderDropdown()}
+            </Host>
+        );
     }
 
     @Watch('value')
     protected onChangeValue() {
+        this.emittedValue = null;
```
john-traas: This clears `emittedValue` on every value write, even when the write is the echo of an *older* emit. Example: pick A, pick B (emits `[A,B]`), then the consumer's echo of `[A]` arrives and clears the hold. The next pick builds on `value = [A]` and B is silently dropped — even though the consumer accepted everything. Suggestion: only clear when the incoming value actually matches the last emit (compare by value id).

Reproduced with a consumer that accepts every change 600ms later: picked Anna, then Bertil, then Cesar (each pick ~300ms apart) → the third emit was `[Anna, Cesar]`. Bertil was dropped by a consumer that never declined anything.

src/components/picker/searcher.types.ts
```diff
@@ -5,6 +5,13 @@ import { PickerItem } from './picker-item.types';
  * A search function that takes a search-string as an argument, and returns
  * a promise that will eventually be resolved with an array of `PickerItem`:s.
  *
+ * When the picker is `multiple`, items that are already picked are removed
```
john-traas: Fixing the exclusion changes behavior for custom searchers that return a fixed page (top-N, server paging). On main the exclusion never matched their results, so they were unaffected; now a page where everything is already picked filters down to nothing and shows a false “No results matching”. The doc tells searchers to account for picked items when paging, but a `Searcher` only receives the query — there's no way to do that. Suggestion: pass the picked ids to the searcher, or offer an opt-out.

Reproduced: a searcher returning a fixed top-2 page for “ann” (both hits already picked, 23 more matches beyond the page) → the dropdown showed “No results matching "ann"” with zero suggestions. We verified the pass-the-picked-items option locally (patch comment on the PR) — whether that or an opt-out fits better is your call.

src/components/picker/picker.tsx
```diff
@@ -793,42 +868,36 @@ export class Picker {
...
+     * being searched for.
+     *
+     * @param query - the query the result belongs to
+     * @param result - what the searcher returned
+     */
     private handleSearchResult(
         query: string,
         result: Array<PickerItem | ListSeparator>
     ) {
         if (query === this.textValue) {
```
john-traas, line 812: A slow `search('')` can reopen a dropdown the user closed. Nothing cancels the search when the session ends, and the guard here can't tell the difference — after closing, `textValue` is `''` and the search's query is `''` too. So: pick an item (which starts `search('')`), press Esc, and when the search resolves the dropdown pops back open. Suggestion: a simple session counter, bumped in `clearInputField` and checked here, would let stale results be dropped.

Reproduced on a real relation picker in lime-webclient (Deal field on a to-do, real API, 400ms network throttle to widen the window): Esc-Esc at t=3920ms; the first Esc fired `search('')` at t=3922, it resolved at t=4328 — after the second Esc had dismissed the dropdown — and the suggestions re-rendered at t=4402. Same end state observed without throttling.

src/components/picker/picker-helpers.ts
```diff
@@ -0,0 +1,86 @@
...
+            )
+    );
+
+    if (pickedIds.size === 0) {
+        return items;
+    }
+
+    return items.filter(
+        (item) => 'separator' in item || !pickedIds.has(getValueId(item))
+    );
```
john-traas: Exclusion and chip ids don't use the same equality. Exclusion is type-strict (`Set([1]).has('1')` is false), but `getChipId` stringifies. So if `value` has the item as `'1'` and `allItems` has it as `1`, it isn't excluded from the suggestions, the user can pick it again — and then both chips share id `'1'`, and removing one removes both (chip-set removes by id).

Reproduced: value `[{text: 'One', value: '1'}]` with allItems carrying `value: 1` → “One” was still offered, picking it produced two “One” chips, and X-clicking one emitted `[]` — both items removed. With the change below, only “Two” is offered.

```suggestion
    const pickedIds = new Set(
        picked
            .filter((item): item is PickerItem => !!item)
            .map(getValueId)
            .filter(
                (valueId): valueId is string | number =>
                    valueId !== undefined && valueId !== null
            )
            .map((valueId) => `${valueId}`)
    );

    if (pickedIds.size === 0) {
        return items;
    }

    return items.filter(
        (item) => 'separator' in item || !pickedIds.has(`${getValueId(item)}`)
    );
```

src/components/picker/picker.tsx
```diff
@@ -702,27 +770,34 @@ export class Picker {
...
-            });
+            newValue = chips.map((chip) => chip.value as PickerItem);
         }
 
-        this.change.emit(newValue as PickerItem | PickerItem[]);
+        this.emitValue(newValue as PickerItem | PickerItem[]);
+
+        const wasEditingInput = await this.chipSet?.getEditMode();
+
+        if (this.multiple && wasEditingInput) {
```
john-traas: Removing a chip can reopen a dropdown the user just closed. After Esc the dropdown is closed but focus stays in the input, so edit mode is still true — and a Backspace removal then fires `search('')` and the dropdown pops back open. Deciding based on whether a search session is actually open, rather than on focus, would keep it closed.

Reproduced in lime-webclient's admin Authentication Log user picker (a real `multiple` consumer, no simulation): pick a user, Esc (dropdown closes), Backspace twice (first selects the chip, second removes it) → the dropdown reopened with the full user list.

A locally verified fix for this is in the patch comment on the PR.

src/components/picker/picker.e2e.tsx
```diff
@@ -53,4 +281,451 @@ describe('limel-picker', () => {
...
+            await waitForSuggestions(waitForChanges);
+
+            const suggestions = getSuggestionTexts();
+            expect(suggestions.some((text) => text.includes('Cherry'))).toBe(
+                true
+            );
+            expect(suggestions.some((text) => text.includes('Blueberry'))).toBe(
+                true
+            );
+            expect(document.body.textContent).not.toContain(
```
john-traas: This assertion can never fail. The “Results matching” header lives inside `limel-list`'s shadow root, and `document.body.textContent` doesn't see into shadow DOM — so the check passes even when the header is still on screen. Reading the list's shadow DOM (like `getSuggestionTexts` does) would make it a real assertion.

Verified in a live browser: with the header visibly rendered, the list's shadow root contains “Results matching” while `document.body.textContent` does not.

src/components/picker/picker.e2e.tsx
```diff
@@ -1,22 +1,248 @@
...
+
+/**
+ * Flushes renders until `isSettled` holds, or the timeout runs out. The
+ * number of passes between "the searcher resolved" and "the suggestion
+ * text is in the DOM" varies with machine load, so it cannot be fixed.
+ *
+ * @param isSettled - what the caller is waiting for
+ * @param waitForChanges - flushes pending renders
+ */
+async function pollUntil(
```
john-traas, line 31: Three small robustness issues with the polling helpers:

1. `pollUntil` returns silently on timeout, so a test can sail past a wait that never succeeded and then fail later with a confusing diff. Throwing on timeout would point at the real problem.
2. The 500ms budget only leaves ~200ms over the 300ms search debounce — tight on a loaded CI runner.
3. `waitForSuggestions` just waits for “some items with text”, which the *old* list also satisfies. Right after a pick it can return while the stale list is still rendered. Waiting for something specific to the new result set would be safer.

src/components/picker/picker.e2e.tsx
```diff
@@ -53,4 +281,451 @@ describe('limel-picker', () => {
...
+                const suggestions = getSuggestionTexts();
+                expect(
+                    suggestions.some((text) => text.includes('Blueberry'))
+                ).toBe(true);
+                expect(
+                    suggestions.some((text) => text.includes('Blackberry'))
+                ).toBe(false);
+            });
+
+            it('keeps the first pick when the consumer echoes `value` late', async () => {
```
john-traas: This test only tests the new mechanism if the second pick happens within the 300ms echo window. On a slow runner the echo lands first, and the test passes through the ordinary path — even if `emittedValue` were completely broken. Triggering the echo manually after the second pick, instead of using a real timer, would make the ordering deterministic.

src/components/picker/picker.tsx
```diff
@@ -650,36 +714,40 @@ export class Picker {
...
      */
-    private handleListChange(event: LimelListCustomEvent<PickerItem>) {
+    private async handleListChange(event: LimelListCustomEvent<PickerItem>) {
         event.stopPropagation();
         if (!this.value || this.value !== event.detail) {
             let newValue: PickerItem | PickerItem[] = event.detail;
             if (this.multiple) {
-                const currentValue = (this.value as PickerItem[]) ?? [];
-                newValue = [...currentValue, event.detail];
+                newValue = [...this.pickedItems(), event.detail];
```
john-traas: There's no dedupe here, and the old list stays clickable until the reset search resolves. A fast double-click (or a slow searcher) can emit the same item twice, giving two chips that share an id and get removed together later. A value-id check before adding would be cheap insurance. (This window existed on main too — not a regression, just worth hardening while you're in here.)

src/components/picker/picker.tsx
```diff
@@ -793,42 +868,36 @@ export class Picker {
...
             if (this.multiple) {
-                const values = (this.value as PickerItem[]) ?? [];
-                nextItems = result.filter((item) => {
-                    if ('separator' in item) {
-                        return true;
-                    }
-
-                    return !values.includes(item);
-                });
+                nextItems = excludePickedItems(result, this.pickedItems());
```
john-traas: On the default-searcher path, picked items are filtered out twice: once in `defaultSearcher` (needed, so picked items don't eat the 20-item cap) and again here. Only applying this one when `this.searcher` is set would leave one exclusion per path.

src/components/picker/picker.tsx
```diff
@@ -650,36 +714,40 @@ export class Picker {
...
+     *
+     */
+    private updateSearchAfterPick() {
+        if (!this.multiple) {
+            this.clearInputField();
+
+            return;
+        }
+
+        this.resetSearchToDefault();
```
john-traas, line 683: Every pick and every while-editing removal now hits the searcher immediately (`resetSearchToDefault` calls `search('')` directly), skipping the debounce every other path uses. Rapid Backspace across five chips means five back-to-back `searcher('')` calls when only the last matters. Routing through `debouncedSearch('')` would coalesce them.

john-traas:
We reproduced the scenarios in the inline comments against this branch running in lime-webclient, then verified that the patch below fixes all of them (same protocols, plus a regression check that picking still clears the query and resets to the default suggestions). Sharing it as one unit rather than per-comment suggestions because the pieces interlock — applied separately they can leave the picker in an untested state. Take whatever is useful.

What it does:

- `handleChange` reads a chip-set change as “these chips were removed” and applies that to `pickedItems()`, so a pending pick survives a removal
- splits `clearInputField` into `endSearchSession` (keeps `emittedValue`) and `clearInputField` (forgets it) — the non-editing removal path uses the former, so a removal's own emit isn't forgotten in the same gesture
- the `value` watcher releases the hold only when the incoming value matches the last emit by value id, so the echo of an older emit doesn't drop later picks
- a `searchSession` counter, bumped whenever a session ends or resets, lets `handleSearchResult` drop a search that was in flight when the user dismissed the dropdown
- chip removal only resets the search when a session is actually open (spinner, typed query, or pickable items), so it can't reopen a dismissed dropdown
- `Searcher` receives the picked items as an optional second argument, so a paging searcher can keep its page from being emptied by the exclusion

<details><summary>Full diff (verified locally, not pushed)</summary>

```diff
diff --git a/src/components/picker/picker-helpers.ts b/src/components/picker/picker-helpers.ts
index 44d8db1ef..c4036e61a 100644
--- a/src/components/picker/picker-helpers.ts
+++ b/src/components/picker/picker-helpers.ts
@@ -74,6 +74,7 @@ export function excludePickedItems(
                 (valueId): valueId is string | number =>
                     valueId !== undefined && valueId !== null
             )
+            .map((valueId) => `${valueId}`)
     );
 
     if (pickedIds.size === 0) {
@@ -81,6 +82,6 @@ export function excludePickedItems(
     }
 
     return items.filter(
-        (item) => 'separator' in item || !pickedIds.has(getValueId(item))
+        (item) => 'separator' in item || !pickedIds.has(`${getValueId(item)}`)
     );
 }
diff --git a/src/components/picker/picker.tsx b/src/components/picker/picker.tsx
index f1c1e0b1a..d4981345c 100644
--- a/src/components/picker/picker.tsx
+++ b/src/components/picker/picker.tsx
@@ -29,6 +29,7 @@ import { getIconFillColor, getIconName } from '../icon/get-icon-props';
 import {
     excludePickedItems,
     getChipId,
+    getValueId,
     hasPickableItems,
 } from './picker-helpers';
 import { DebouncedFunc, debounce } from 'lodash-es';
@@ -264,6 +265,13 @@ export class Picker {
      */
     private emittedValue: PickerItem[] = null;
 
+    /**
+     * Incremented whenever a search session ends or resets, so a search
+     * that was in flight when that happened can be told from a current
+     * one and its result dropped.
+     */
+    private searchSession = 0;
+
     private debouncedSearch: DebouncedFunc<(query: string) => Promise<void>>;
     private chipSet: HTMLLimelChipSetElement;
     private portalId: string;
@@ -361,10 +369,33 @@ export class Picker {
 
     @Watch('value')
     protected onChangeValue() {
-        this.emittedValue = null;
+        if (this.valueReflectsEmitted()) {
+            this.emittedValue = null;
+        }
         this.chips = this.createChips(this.value);
     }
 
+    /**
+     * Whether `value` now holds the same items (by value id) as the last
+     * emitted value. The echo of an older emit does not release the hold.
+     *
+     * @returns `true` when there is no pending emit, or `value` matches it
+     */
+    private valueReflectsEmitted(): boolean {
+        if (this.emittedValue === null) {
+            return true;
+        }
+
+        const toIdKey = (value: PickerItem | PickerItem[]): string =>
+            (Array.isArray(value) ? value : [value])
+                .filter(Boolean)
+                .map((item) => `${getValueId(item)}`)
+                .sort()
+                .join('');
+
+        return toIdKey(this.value ?? []) === toIdKey(this.emittedValue);
+    }
+
     private renderDelimiter() {
         if (this.multiple) {
             return this.delimiter;
@@ -661,6 +692,7 @@ export class Picker {
     }
 
     private search = async (query: string) => {
+        const session = this.searchSession;
         const timeoutId = setTimeout(() => {
             this.loading = true;
         });
@@ -669,7 +701,7 @@ export class Picker {
         let result: Array<PickerItem | ListSeparator> = [];
 
         try {
-            result = await searcher(this.textValue);
+            result = await searcher(this.textValue, this.pickedItems());
         } catch (error) {
             console.error('limel-picker: the searcher failed.', error);
         } finally {
@@ -678,7 +710,7 @@ export class Picker {
             clearTimeout(timeoutId);
         }
 
-        this.handleSearchResult(query, result);
+        this.handleSearchResult(query, result, session);
     };
 
     private defaultSearcher: Searcher = async (
@@ -785,21 +817,50 @@ export class Picker {
 
         let newValue = null;
         if (this.multiple) {
-            const chips = event.detail as Chip[];
-            newValue = chips.map((chip) => chip.value as PickerItem);
+            // The rendered chips derive from `value`, which may lag behind
+            // `emittedValue`, so the removal is read as "these chips went
+            // away" and applied to the picked items — not taken verbatim.
+            const remaining = new Set(
+                (event.detail as Chip[]).map((chip) => `${chip.id}`)
+            );
+            const removed = new Set(
+                this.chips
+                    .map((chip) => `${chip.id}`)
+                    .filter((id) => !remaining.has(id))
+            );
+            newValue = this.pickedItems().filter(
+                (item) => !removed.has(getChipId(item))
+            );
         }
 
         this.emitValue(newValue as PickerItem | PickerItem[]);
 
         const wasEditingInput = await this.chipSet?.getEditMode();
 
-        if (this.multiple && wasEditingInput) {
+        if (this.multiple && wasEditingInput && this.hasOpenSearchSession()) {
             this.resetSearchToDefault();
         } else {
-            this.clearInputField();
+            // End the session but keep `emittedValue`: the emit above is
+            // still pending, and forgetting it here would let the next
+            // pick resurrect the removed item.
+            this.endSearchSession();
         }
     }
 
+    /**
+     * Whether a search session is showing anything the user has not
+     * dismissed: a spinner, a typed query, or pickable suggestions.
+     *
+     * @returns `true` if a search session is open
+     */
+    private hasOpenSearchSession(): boolean {
+        return (
+            this.loading ||
+            this.textValue !== '' ||
+            hasPickableItems(this.items ?? [])
+        );
+    }
+
     private handleInteract(event: LimelChipSetCustomEvent<Chip>) {
         event.stopPropagation();
         this.interact.emit(event.detail ? event.detail.value : event.detail);
@@ -877,8 +938,13 @@ export class Picker {
      */
     private handleSearchResult(
         query: string,
-        result: Array<PickerItem | ListSeparator>
+        result: Array<PickerItem | ListSeparator>,
+        session: number = this.searchSession
     ) {
+        if (session !== this.searchSession) {
+            return;
+        }
+
         if (query === this.textValue) {
             let nextItems = result;
             if (this.multiple) {
@@ -917,21 +983,31 @@ export class Picker {
      * search.
      */
     private clearTextValue() {
+        this.searchSession++;
         this.chipSet?.emptyInput();
         this.textValue = '';
         this.debouncedSearch.cancel();
     }
 
     /**
-     * Ends the search session, leaving the dropdown with nothing to show.
-     * Goes through `handleSearchResult` so `loading` is settled too.
+     * Ends the search session, leaving the dropdown with nothing to show,
+     * but keeps any pending emitted value. Goes through
+     * `handleSearchResult` so `loading` is settled too.
      */
-    private clearInputField() {
-        this.emittedValue = null;
+    private endSearchSession() {
         this.clearTextValue();
         this.handleSearchResult('', []);
     }
 
+    /**
+     * Ends the search session and defers to `value` again: a pending
+     * change that the consumer has not answered is forgotten.
+     */
+    private clearInputField() {
+        this.emittedValue = null;
+        this.endSearchSession();
+    }
+
     private resetSearchToDefault() {
         this.clearTextValue();
         this.search('');
diff --git a/src/components/picker/searcher.types.ts b/src/components/picker/searcher.types.ts
index e8b9f7d2b..62631ff98 100644
--- a/src/components/picker/searcher.types.ts
+++ b/src/components/picker/searcher.types.ts
@@ -14,9 +14,13 @@ import { PickerItem } from './picker-item.types';
  *
  * @param query - A search query. Typically what the user has written
  * in the input field of a limel-picker.
+ * @param pickedItems - The items the picker currently treats as picked.
+ * A paging searcher can use these to keep its page from being emptied
+ * by the exclusion.
  * @returns The search result.
  * @public
  */
 export type Searcher = (
-    query: string
+    query: string,
+    pickedItems?: readonly PickerItem[]
 ) => Promise<Array<PickerItem | ListSeparator>>;
```

</details>

john-traas:
While testing in lime-webclient, the service worker's cache may need to be cleared (unregister + clear `caches` in DevTools → Application), or a stale lime-elements bundle can end up being what's actually running — it survives cache-disabled reloads. `/webadmin/` is outside the SW scope, so it always loads fresh bundles. This cost us a false verification failure before we spotted it.

src/components/picker/picker.tsx
devbymadde: Good find - I've decided to try to keep this PR as small as possible now and not fix stuff that exist on main, so let's do this in another PR, another time :)

src/components/picker/picker.tsx
```diff
@@ -660,24 +671,29 @@ export class Picker {
...
-                this.items = this.hasPickableItems(remaining) ? remaining : [];
-            } else {
-                // Single-pick: the search session ends with the pick, so
-                // wipe the input. (In multi-pick we deliberately keep the
-                // typed query so the user can keep adding matches.)
-                this.items = [];
-                this.textValue = '';
-                this.chipSet?.emptyInput();
-            }
+            this.updateSearchAfterPick();
```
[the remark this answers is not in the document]
devbymadde: This is something that already exist on main and has been like this for several years. Filed as #4268 with the details. Out of scope here.

devbymadde:
Thanks for the thorough review and the reproductions @john-traas - they made it clear that I entered a rabbit hole with this PR and that it grew way too much compared to what the first intention was :D

In order not to block this anymore I've reverted the fixes to the pre-existing bugs so this PR only does what it says: don't keep the search term after selecting an item. The findings are filed as issues instead, with your reproductions:

- picker: already picked items are not excluded from search results #4268
- picker: a late search result reopens a dismissed dropdown #4269
- picker: a failing searcher leaves the loading spinner up #4270
- picker: removing a chip can duplicate one item and drop another #4271

Working fixes for #4268, #4270 and #4271 are on `fix/picker-known-bugs`, tested, for whoever picks them up. #4269 has no fix yet - your session counter suggestion is written up in the issue.

Two of your findings turned out to be regressions I'd introduced rather than pre-existing, so both are fixed here:

- Esc followed by a Backspace removal reopened the dropdown. The refresh was gated on edit mode, which stays true after Esc since focus remains in the input. It now checks whether the dropdown is actually showing something, so a dismissed one stays dismissed.
- The refreshed search ran immediately, which read `allItems` from before the consumer's re-render - so an item you'd just picked was offered again. It goes through the debounce now.

Verified both against main with the same test to be sure they really were regressions and not something I was inheriting.

That debounce also fixed #4272, so I'm closing it - five rapid removals made five `searcher('')` calls before and make one now. And I've made the test polling throw on timeout instead of returning silently, so a wait that never succeeded reports itself rather than showing up as whichever assertion fails next.

src/components/picker/picker.tsx
devbymadde: Same behaviour on main - will skip for now

src/components/picker/picker.tsx
devbymadde: Exists on main - will not fix in this PR

src/components/picker/picker.e2e.tsx
devbymadde: Fixed 1 &2, and 3 is mostly fixed now when `waitForSuggestions` is waiting for the debounce

john-traas (approved):

src/components/picker/picker.tsx
```diff
@@ -864,7 +876,7 @@ export class Picker {
 
     private resetSearchToDefault() {
         this.clearTextValue();
-        this.search('');
+        this.debouncedSearch('');
```
john-traas, line 891: In the escape handler the debounced search keeps the old results on screen for a while after Esc is pressed, about 300 ms. The text clears straight away but the "Results matching" header and the rows hang on until the debounced search lands. It's not a long delay, and the first time I hit it my network was acting up so it looked much longer than it was, but on main they go together.

Esc doesn't change the value, so there's nothing for it to wait for. Should we let the escape handler skip the debounce? The handler sits just below this hunk, so here it is spelled out:

```ts
    private handleEscape() {
        if (this.textValue === '') {
            this.clearInputField();

            return;
        }

        this.clearTextValue();
        this.search('');
    }
```

I tried this in limepkg-email's recipient picker: the old rows disappear at about 10 ms after Esc instead of 300, picking still clears and refreshes as intended, and the 13 picker e2e tests pass.

john-traas (changes requested):
One small thing on the Esc handler, see the inline comment.

john-traas (approved):

PR #4233, opened 2026-08-17
fix(text-editor): focus the link field when the link menu opens

### LINKED ISSUE #3181: Set focus in `link` instead of `text` in the link popover in limel-text-editor
## New feature motivation

Annoying to **always** select the link field since that is what you want to insert.

https://github.com/user-attachments/assets/ad59c76c-c9c0-49c4-95da-6ec9308a8fd6

john-traas:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved the link menu so the link field receives focus when opened.
  * Ensured the text cursor is correctly positioned in the link input for immediate editing.

* **Tests**
  * Added end-to-end coverage validating link-field focus and cursor placement.

Closes #3181

The link menu already moved focus to an input when it opened — it just picked
the *Text* field. Since the point of opening the menu is to enter a URL, the
`ref` now points at the *Link* field instead.

The menu element is torn down and rebuilt on every open (`renderLinkMenu()`
returns nothing while closed), so `componentDidLoad` — and the focus call —
runs each time the menu opens, not only the first.

The link field is focused, but its existing value is **not** selected. When
editing an existing link the URL still has to be cleared by hand. Worth
deciding after trying it out whether select-on-open is wanted too.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/text-editor/link-menu/editor-link-menu.e2e.tsx
+++ b/src/components/text-editor/link-menu/editor-link-menu.e2e.tsx
@@ -1,0 +1,44 @@
+import { render, h } from '@stencil/vitest';
+
+describe('limel-text-editor-link-menu', () => {
+    const link = {
+        text: 'Lime Technologies',
+        href: 'https://www.lime-technologies.com',
+    };
+
+    async function createMenu() {
+        const { root, waitForChanges } = await render(
+            <limel-text-editor-link-menu link={link} isOpen={true} />
+        );
+        await waitForChanges();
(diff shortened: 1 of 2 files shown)


devbymadde (approved):
Nice work! 🙌

PR #4257, opened 2026-08-27
Line up list labels when only some items have pictures

### LINKED ISSUE #4256: List items with only an icon are misaligned next to items with images
## Current behavior

In a list where some items have an `image` and others only have an `icon`, the icon-only items start their text further to the left than the items with a picture, so the labels no longer line up in a single column.

Image

An icon is narrower than an image, and it is additionally pulled `0.5rem` to the left. Measured in the browser, the label of an item with an image starts `3.5rem` from the item's left edge (`1rem` padding + `2rem` image + `0.5rem` gap), while an icon-only item's label starts at `2.25rem` with plain icons and at `2.75rem` with `badgeIcons` — a `1.25rem` and `0.75rem` shift respectively.

**Screenshot:** _to be attached._

1. Render a `limel-list` where at least one item has an `image` and at least one other item only has an `icon`. The "List with Pictures and Icons" example already does exactly this: its last item, "Befkadu Degefa", has an icon but no picture.
2. Compare where the labels start. The last row's text sits noticeably to the left of the rows above it.
3. Toggle the `badgeIcons` switch. The gap gets smaller, but the labels stay misaligned.

## Expected behavior

An item that has no picture, sitting among items that do, should reserve exactly the space a picture would have taken, so that every label starts in the same column. 

Image

In other words the lone icon should be as wide as the image, and badge icons and plain icons should behave identically in this respect — today they are misaligned by different amounts, which makes the bug look like two separate problems.

## Environment

- **lime-elements** version: 39.45.3
- Framework used: StencilJS — reproducible in the library's own documentation examples (`limel-example-list-pictures`)
- Logs: n/a

## Where the fix belongs

`limel-list` renders its rows through the private `limel-list-item` component, and all of the icon/image layout lives in that component's `list-item.scss`. The fix belongs there rather than in `list.scss`, so that `limel-menu-list` and direct uses of `limel-list-item` — which are affected in exactly the same way — are covered too. Since `limel-list-item` is declared with `shadow: false`, its stylesheet can reach the surrounding element and recognise the mixed case with `:has()`, without any change to the markup or the public API.

Related but not covered by this report: an item that has neither a picture nor an icon, among items that have pictures, is left misaligned in the same way.

related to https://github.com/Lundalogik/crm-client/issues/1235 and simplifies some of the decisions in that PR

Kiarokh:
Closes #4256.

In a list where some items have a picture and others only have an icon, the icon-only items started their label further to the left, so the labels sat in two different columns. Measured in the browser: a row with a picture starts its label 3.5rem from the item's edge (1rem padding + 2rem picture + 0.5rem gap), while an icon-only row started at 2.25rem with plain icons and 2.75rem with `badgeIcons`. Because the two variants were off by different amounts they read as two separate bugs, which is why the issue calls out that they should behave alike.

The fix gives a lone icon the picture's square whenever the list holds pictures, so the icon stands in for the missing picture and every label starts in the same column.

## Where it lives, and why

`limel-list` renders its rows through the private `limel-list-item`, and all of the icon/picture layout lives in that component's `list-item.scss` — so the rule goes there rather than in `list.scss`. That also covers `limel-menu-list` and any direct use of `limel-list-item`, which had the same defect. Since `limel-list-item` is declared `shadow: false`, its stylesheet can reach the surrounding element and recognise the mixed case on its own, with no markup change and no change to the public API.

I first wrote the detection in `list-renderer.tsx` instead, handing the size down through internal `--limel-` custom properties. It needed two properties rather than one (reserving the space is not enough — the icon also has to stop reaching outside the item's padding, or the labels still land half a rem apart), it left every other consumer of `limel-list-item` to opt in by itself, and it was the larger diff. We settled on the CSS-only version.

## Two things worth a reviewer's eye

**The `:has()` selector.** It reads "a lone icon, in a list that holds at least one picture":

```scss
:has(> limel-list-item > img) > limel-list-item:not(:has(> img)) > limel-icon
```

The first `:has()` is on the parent `<ul>`, not on a sibling item, so it is not sensitive to where in the list the pictures are. The `>` combinators keep it from reaching into a nested list.

**The clamp instead of `width`.** `limel-icon` sets its own width and height with `!important` from inside its shadow root. For important declarations the inner tree wins the cascade, so an outer `width` can never reach it — important or not. `min-*`/`max-*` are different properties and clamp the used value, which is why the rule is written that way.

An earlier draft used only `min-width`/`min-height`. That left `badgeIcons` with `iconSize="medium"` or `"large"` misaligned, since those badges are 2.5rem and 2.75rem — wider than the picture. Clamping from both sides covers every size; the trade-off is that an oversized badge is pulled down to the picture's 2rem in a mixed list, which seems right given it is standing in for one.

## Testing

16 e2e tests in `list.e2e.tsx` assert equal label offsets and an identical icon/picture box across all four `iconSize` values times both `badgeIcons` states. 15 of them fail without the SCSS change. Full e2e and spec suites are otherwise unchanged — the `form` and `text-editor/prosemirror-*` failures are pre-existing and reproduce on a clean `main`.

## Out of scope

An item with neither a picture nor an icon, sitting among items with pictures, is still misaligned. Noted in the issue.

## Summary by CodeRabbit

* **New Features**
  * List items now maintain consistent label alignment when combining images with icon-only items.
  * Icon-only items match the configured image footprint across different icon sizes and badge configurations.
  * Image sizing can be customized using the `--limel-list-item-image-size` property.
  * Updated examples demonstrate mixed image and icon list items.

* **Tests**
  * Added coverage for mixed-content lists across icon sizes and badge settings.

### DIFF
--- a/src/components/list-item/examples/list-item-pictures.tsx
+++ b/src/components/list-item/examples/list-item-pictures.tsx
@@ -4,8 +4,13 @@ import { Component, h, Host, State } from '@stencil/core';
  * List item with pictures
  *
  * This example demonstrates how to use images in list items.
- * The first item shows a picture only, while the second item
- * shows both a picture and an icon together.
+ * The first item shows a picture only, the second one shows both a picture
+ * and an icon together, and the third one shows an icon but no picture.
+ *
+ * Notice how the icon of the third item is given the same footprint as the
+ * pictures above it, so that all three labels start in the same column.
+ * This happens whenever an item without a picture sits among items that
+ * have one, and applies to badge icons and plain icons alike.
  */
 @Component({
     tag: 'limel-example-list-item-pictures',
@@ -41,6 +46,16 @@ export class ListItemPicturesExample {
                         image={{
                             src: 'https://lundalogik.github.io/lime-elements/2e86c284-d190-4c41-8da2-4de50103a0cd.png',
                             alt: 'A picture of Kiarokh Moattar, Product Designer at Lime Technologies',
+                        }}
+                    />
+                    <limel-list-item
+                        text="Befkadu Degefa"
+                        secondaryText="Engineer"
+                        badgeIcon={this.badgeIcon}
+                        icon={{
+                            name: 'bowler_hat',
+                            title: 'Bowler hat icon',
+                            color: 'rgb(var(--color-sky-default))',
                         }}
                     />
                 </ul>
--- a/src/components/list-item/list-item.scss
+++ b/src/components/list-item/list-item.scss
@@ -4,10 +4,13 @@
  * @prop --limel-list-item-menu-order: Defines the order of the menu, within the list item's flexbox. Defaults to `3`.
  * @prop --list-item-image-border-radius: Defines the border radius of the list item image. Defaults to `50%`.
  * @prop --list-item-image-object-fit: Defines the object-fit property of the list item image. Defaults to `cover`.
+ * @prop --limel-list-item-image-size: Defines the size of the list item image. Defaults to `2rem`.
  */
 
 @use '../../style/mixins';
 @forward '../checkbox/checkbox';
+
+$image-box-shadow: 0 0 0 1px rgb(var(--contrast-800), 0.5);
 
 *,
 *:before,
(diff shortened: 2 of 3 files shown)


LucyChyzhova (approved):
looks great!

PR #4262, opened 2026-09-01
Fix chart tooltip percentages and allow hiding them

jecrapo:
## Summary by CodeRabbit

- **New Features**
  - Added configurable tooltip percentage visibility for pie, doughnut, ring, and stacked-bar charts.
  - Added an interactive example for toggling tooltip percentages.
  - Improved whole-chart sizing and maximum-value handling.
  - Chart ranges now recalculate when the chart type changes.

- **Documentation**
  - Clarified how maximum values apply across chart types.

- **Tests**
  - Added coverage for sizing, rendering, tooltip values, percentage suppression, range handling, and chart type changes.

Two changes to the tooltip percentage of `limel-chart`.

**Fix.** For `stacked-bar`, `pie`, `doughnut`, and `ring`, an item's size and tooltip percentage are its share of `maxValue` when set. Otherwise, the whole is the combined scalar values and range extents. The chart does not round this whole to an axis step. `maxValue` is a scale, not a cap, so combined items can exceed 100%. A stacked bar used to divide by its largest item, so its percentages did not add up to 100, and a pie rounded its whole and left a gap. Charts with an axis show no percentage. `line` and `area` used to print one.

**Opt-out.** `displayItemPercentage`, default `true`. Set it to `false` to show the `text` and the value alone.

Part of Lundalogik/crm-insights-and-intelligence#354. Drawing a stacked bar chart with several bars in one element is Lundalogik/lime-elements#4263.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -247,6 +247,7 @@ export namespace Components {
         "accessibleValuesLabel"?: string;
         "axisIncrement"?: number;
         "displayAxisLabels": boolean;
+        "displayItemPercentage": boolean;
         "displayItemText": boolean;
         "displayItemValue": boolean;
         "items": ChartItem[];
@@ -1806,6 +1807,7 @@ export namespace JSX {
         "accessibleValuesLabel"?: string;
         "axisIncrement"?: number;
         "displayAxisLabels"?: boolean;
+        "displayItemPercentage"?: boolean;
         "displayItemText"?: boolean;
         "displayItemValue"?: boolean;
         "items": ChartItem[];
@@ -1838,6 +1840,8 @@ export namespace JSX {
         "axisIncrement": number;
         // (undocumented)
         "displayAxisLabels": boolean;
+        // (undocumented)
+        "displayItemPercentage": boolean;
         // (undocumented)
         "displayItemText": boolean;
         // (undocumented)
--- a/src/components/chart/chart.tsx
+++ b/src/components/chart/chart.tsx
@@ -7,6 +7,17 @@ import { ChartItem } from './chart.types';
 const PERCENT = 100;
 const DEFAULT_INCREMENT_SIZE = 10;
 
+/**
+ * Types whose items are parts of one whole. They draw no axis, and an
+ * item's size is its share of `maxValue`, or of the sum of the items.
+ */
+const WHOLE_TYPES: ReadonlySet<string> = new Set([
+    'stacked-bar',
+    'pie',
+    'doughnut',
+    'ring',
+]);
+
 interface AxisRange {
     minValue: number;
     maxValue: number;
@@ -139,15 +150,17 @@ export class Chart {
     public orientation?: 'landscape' | 'portrait' = 'landscape';
 
     /**
-     * Specifies the range that items' values could be in.
-     * This is used in calculation of the size of the items in the chart.
-     * When not provided, the sum of all values in the items will be considered as the range.
+     * The upper end of the range that items are drawn against.
+     * For `stacked-bar`, `pie`, `doughnut`, and `ring` charts it is the whole
+     * that each item is a share of, and defaults to the sum of the items.
+     * For the other types it defaults to the largest value among the items.
      */
     @Prop({ reflect: true })
     public maxValue?: number;
 
     /**
      * Specifies the increment for the axis lines.
+     * Has no effect on chart types without axis lines.
      */
     @Prop({ reflect: true })
     public axisIncrement?: number;
@@ -403,6 +416,13 @@ export class Chart {
             };
         }
 
+        if (this.isPartOfWhole()) {
+            return {
+                size: (this.getValueSpan(item) / totalRange) * PERCENT,
+                offset: 0,
+            };
+        }
+
         let startValue = 0;
         if (this.isRangeItem(item)) {
             startValue = this.getMinimumValue(item);
@@ -452,12 +472,7 @@ export class Chart {
             elementId: itemId,
         };
 
-        if (
-            this.type !== 'bar' &&
-            this.type !== 'dot' &&
-            this.type !== 'nps' &&
-            this.type !== 'scatter'
-        ) {
+        if (this.isPartOfWhole()) {
             tooltipProps.label = `${text} (${size.toFixed(PERCENT_DECIMAL)}%)`;
         }
 
@@ -471,25 +486,18 @@ export class Chart {
         );
     }
 
-    private calculateRange() {
+    private calculateRange(): AxisRange {
         if (this.range) {
             return this.range;
         }
 
+        if (this.isPartOfWhole()) {
+            return this.calculateWholeRange();
+        }
+
         const minRange = Math.min(0, ...this.items.map(this.getMinimumValue));
         const maxRange = Math.max(...this.items.map(this.getMaximumValue));
-        const totalSum = this.items.reduce(
-            (sum, item) => sum + this.getMaximumValue(item),
-            0
-        );
-
-        let finalMaxRange = this.maxValue ?? maxRange;
-        if (
-            (this.type === 'pie' || this.type === 'doughnut') &&
-            !this.maxValue
-        ) {
-            finalMaxRange = totalSum;
-        }
+        const finalMaxRange = this.maxValue ?? maxRange;
 
         if (!this.axisIncrement) {
             this.axisIncrement = this.calculateAxisIncrement(
@@ -509,6 +517,33 @@ export class Chart {
             totalRange: totalRange,
             increment: this.axisIncrement,
         };
+    }
+
+    /**
+     * The range of a chart whose items are parts of one whole: `maxValue`
+     * when set, otherwise the sum of the items. The whole is not rounded to
+     * an axis step, since these types draw no axis and an item's size has
+     * to be its exact share.
+     */
+    private calculateWholeRange(): AxisRange {
+        const sum = this.items.reduce(
+            (total, item) => total + this.getValueSpan(item),
+            0
+        );
+        const inferredWhole = sum > 0 ? sum : 1;
+        // An explicit zero also falls back to 1, so normalization stays finite.
+        const whole = (this.maxValue ?? inferredWhole) || 1;
+
+        return {
+            minValue: 0,
+            maxValue: whole,
+            totalRange: whole,
+            increment: whole,
+        };
+    }
+
+    private isPartOfWhole(): boolean {
+        return WHOLE_TYPES.has(this.type);
     }
 
     private calculateAxisIncrement(
@@ -587,6 +622,14 @@ export class Chart {
         return Array.isArray(value) ? Math.max(...value) : value;
     }
 
+    private getValueSpan(item: ChartItem): number {
+        const startValue = this.isRangeItem(item)
+            ? this.getMinimumValue(item)
+            : 0;
+
+        return this.getMaximumValue(item) - startValue;
+    }
+
     private isRangeItem(item: ChartItem): item is ChartItem<[number, number]> {
         return Array.isArray(item.value);
     }
@@ -594,6 +637,7 @@ export class Chart {
     @Watch('items')
     @Watch('axisIncrement')
     @Watch('maxValue')
+    @Watch('type')
     handleChange() {
         this.range = null;
         this.secondaryAxisRange = null;
--- a/src/components/chart/chart.tsx
+++ b/src/components/chart/chart.tsx
@@ -8,8 +8,8 @@ const PERCENT = 100;
 const DEFAULT_INCREMENT_SIZE = 10;
 
 /**
- * Types whose items are parts of one whole. They draw no axis, and an
- * item's size is its share of `maxValue`, or of the sum of the items.
+ * Types whose items are parts of one whole. They draw no axis, and an item's
+ * value or range extent is its share of `maxValue`, or of the combined items.
  */
 const WHOLE_TYPES: ReadonlySet<string> = new Set([
     'stacked-bar',
@@ -34,6 +34,7 @@ interface AxisRange {
  * @exampleComponent limel-example-chart-stacked-bar
  * @exampleComponent limel-example-chart-orientation
  * @exampleComponent limel-example-chart-max-value
+ * @exampleComponent limel-example-chart-display-item-percentage
  * @exampleComponent limel-example-chart-type-bar
  * @exampleComponent limel-example-chart-type-dot
  * @exampleComponent limel-example-chart-type-area
@@ -120,6 +121,15 @@ export class Chart {
     public displayItemValue = false;
 
     /**
+     * Set to `false` to hide the percentage in item tooltips.
+     * The percentage is an item's value or range extent as a share of the
+     * whole. The whole is `maxValue` when set and the combined items otherwise.
+     * Applies to `stacked-bar`, `pie`, `doughnut`, and `ring` charts.
+     */
+    @Prop({ reflect: true })
+    public displayItemPercentage = true;
+
+    /**
      * List of items in the chart,
      * each representing a data point.
      */
@@ -150,17 +160,21 @@ export class Chart {
     public orientation?: 'landscape' | 'portrait' = 'landscape';
 
     /**
-     * The upper end of the range that items are drawn against.
-     * For `stacked-bar`, `pie`, `doughnut`, and `ring` charts it is the whole
-     * that each item is a share of, and defaults to the sum of the items.
-     * For the other types it defaults to the largest value among the items.
+     * Sets the value used to scale item values.
+     * For `stacked-bar`, `pie`, `doughnut`, and `ring`, it is the denominator
+     * used as the whole. It defaults to the combined item values or range
+     * extents. The combined size can exceed `maxValue` and produce a total
+     * above 100%.
+     * For `bar`, `dot`, `area`, and `line`, it is the requested upper bound.
+     * The chart can round that bound up to the next `axisIncrement`.
+     * Has no effect on `nps` or `scatter`.
      */
     @Prop({ reflect: true })
     public maxValue?: number;
 
     /**
-     * Specifies the increment for the axis lines.
-     * Has no effect on chart types without axis lines.
+     * Sets the axis increment for `bar`, `dot`, `area`, and `line` charts.
+     * Has no effect on other chart types.
      */
     @Prop({ reflect: true })
     public axisIncrement?: number;
@@ -472,7 +486,7 @@ export class Chart {
             elementId: itemId,
         };
 
-        if (this.isPartOfWhole()) {
+        if (this.displayItemPercentage && this.isPartOfWhole()) {
             tooltipProps.label = `${text} (${size.toFixed(PERCENT_DECIMAL)}%)`;
         }
 
@@ -521,9 +535,10 @@ export class Chart {
 
     /**
      * The range of a chart whose items are parts of one whole: `maxValue`
-     * when set, otherwise the sum of the items. The whole is not rounded to
-     * an axis step, since these types draw no axis and an item's size has
-     * to be its exact share.
+     * when set, otherwise the combined item values or range extents. The value
+     * is a scale, not a cap, so the combined size can exceed it. The whole is
+     * not rounded to an axis step, since these types draw no axis and an item's
+     * size has to be its exact share.
      */
     private calculateWholeRange(): AxisRange {
         const sum = this.items.reduce(
--- a/src/components/chart/examples/chart-display-item-percentage.tsx
+++ b/src/components/chart/examples/chart-display-item-percentage.tsx
@@ -1,0 +1,43 @@
+import { Component, h, Host, State } from '@stencil/core';
+import { chartItems } from './chart-items-stack';
+
+/**
+ * Hiding the percentage in tooltips
+ *
+ * The tooltip of an item in a `stacked-bar`, `pie`, `doughnut`, or `ring`
+ * chart shows the item's share of the whole next to its `text`.
+ * Set `displayItemPercentage` to `false` to show the `text` and the value alone.
+ *
+ * @sourceFile chart-items-stack.ts
+ */
+@Component({
+    tag: 'limel-example-chart-display-item-percentage',
+    shadow: true,
+    styleUrl: 'chart-examples.scss',
+})
+export class ChartDisplayItemPercentageExample {
+    @State()
+    private displayItemPercentage = false;
+
+    public render() {
+        return (
+            <Host>
+                <limel-chart
+                    items={chartItems}
+                    displayItemPercentage={this.displayItemPercentage}
+                />
+                <limel-example-controls>
+                    <limel-switch
+                        label="displayItemPercentage"
+                        value={this.displayItemPercentage}
+                        onChange={this.handleChange}
+                    />
+                </limel-example-controls>
+            </Host>
+        );
+    }
+
+    private readonly handleChange = (event: CustomEvent<boolean>) => {
+        this.displayItemPercentage = event.detail;
+    };
+}
--- a/src/components/chart/chart.spec.tsx
+++ b/src/components/chart/chart.spec.tsx
@@ -7,7 +7,10 @@ const items: ChartItem[] = [
 ];
 
 type Props = Partial<
-    Pick<HTMLLimelChartElement, 'type' | 'items' | 'maxValue'>
+    Pick<
+        HTMLLimelChartElement,
+        'type' | 'items' | 'maxValue' | 'displayItemPercentage'
+    >
 >;
 
 async function setup(props: Props = {}) {
@@ -31,6 +34,7 @@ async function setup(props: Props = {}) {
 
     return {
         labels: tooltips.map((tooltip) => tooltip.label),
+        values: tooltips.map((tooltip) => tooltip.helperLabel),
         sizes,
         offsets: rows.map((row) =>
             Number(
@@ -135,6 +139,20 @@ describe('limel-chart', () => {
             const { sizes } = await setup({ items: negativeItems });
 
             expect(sizes.every((size) => size < 0)).toBe(true);
+        });
+    });
+
+    describe('displayItemPercentage', () => {
+        test('set to false, the tooltip shows the text alone', async () => {
+            const { labels } = await setup({ displayItemPercentage: false });
+
+            expect(labels).toEqual(['Applications', 'Photos']);
+        });
+
+        test('set to false, the tooltip still shows the value', async () => {
+            const { values } = await setup({ displayItemPercentage: false });
+
+            expect(values).toEqual(['25', '75']);
         });
     });
 
--- a/src/components/chart/chart.spec.tsx
+++ b/src/components/chart/chart.spec.tsx
@@ -1,0 +1,151 @@
+import { render, h } from '@stencil/vitest';
+import { ChartItem } from './chart.types';
+
+const items: ChartItem[] = [
+    { text: 'Applications', value: 25 },
+    { text: 'Photos', value: 75 },
+];
+
+type Props = Partial<
+    Pick<HTMLLimelChartElement, 'type' | 'items' | 'maxValue'>
+>;
+
+async function setup(props: Props = {}) {
+    const { root, waitForChanges } = await render(
+        <limel-chart items={items} {...props} />
+    );
+    await waitForChanges();
+
+    const shadow = root.shadowRoot;
+    const tooltips = [
+        ...shadow.querySelectorAll('limel-tooltip'),
+    ] as HTMLLimelTooltipElement[];
+    const rows = [...shadow.querySelectorAll('tr.item')];
+    const sizes = rows.map((row) =>
+        Number(
+            (row as HTMLElement).style.getPropertyValue(
+                '--limel-chart-item-size'
+            )
+        )
+    );
+
+    return {
+        labels: tooltips.map((tooltip) => tooltip.label),
+        sizes,
+        offsets: rows.map((row) =>
+            Number(
+                (row as HTMLElement).style.getPropertyValue(
+                    '--limel-chart-item-offset'
+                )
+            )
+        ),
+    };
+}
+
+describe('limel-chart', () => {
+    describe('items that are parts of a whole', () => {
+        test('a stacked bar sizes its items as shares of their sum', async () => {
+            const { labels, sizes } = await setup();
+
+            expect(sizes).toEqual([25, 75]);
+            expect(labels).toEqual([
+                'Applications (25.00%)',
+                'Photos (75.00%)',
+            ]);
+        });
+
+        test('maxValue is the whole, and is not rounded to an axis step', async () => {
+            const { labels, sizes } = await setup({ maxValue: 210 });
+
+            expect(sizes[0]).toBeCloseTo(11.9, 1);
+            expect(labels[0]).toBe('Applications (11.90%)');
+        });
+
+        test('a pie fills the whole circle', async () => {
+            const { sizes } = await setup({ type: 'pie' });
+
+            expect(sizes[0] + sizes[1]).toBe(100);
+        });
+
+        test('a ring shows each item as its share of maxValue', async () => {
+            const { labels, sizes } = await setup({
+                type: 'ring',
+                maxValue: 250,
+            });
+
+            expect(sizes).toEqual([10, 30]);
+            expect(labels[0]).toBe('Applications (10.00%)');
+        });
+
+        test('changing the type recalculates the range', async () => {
+            const { root, waitForChanges } = await render(
+                <limel-chart items={items} type="bar" />
+            );
+            await waitForChanges();
+
+            root.type = 'ring';
+            await waitForChanges();
+
+            const sizes = [...root.shadowRoot.querySelectorAll('tr.item')].map(
+                (row) =>
+                    Number(
+                        (row as HTMLElement).style.getPropertyValue(
+                            '--limel-chart-item-size'
+                        )
+                    )
+            );
+
+            expect(sizes).toEqual([25, 75]);
+        });
+
+        test('range items use their extents as shares', async () => {
+            const rangeItems: ChartItem[] = [
+                { text: 'Planning', value: [10, 40] },
+                { text: 'Delivery', value: [40, 100] },
+            ];
+            const { labels, sizes } = await setup({ items: rangeItems });
+
+            expect(sizes[0]).toBeCloseTo(100 / 3);
+            expect(sizes[1]).toBeCloseTo((100 * 2) / 3);
+            expect(labels).toEqual(['Planning (33.33%)', 'Delivery (66.67%)']);
+        });
+
+        test('ring range items start at zero', async () => {
+            const rangeItems: ChartItem[] = [
+                { text: 'Planning', value: [10, 40] },
+                { text: 'Delivery', value: [40, 100] },
+            ];
+            const { offsets, sizes } = await setup({
+                type: 'ring',
+                items: rangeItems,
+                maxValue: 90,
+            });
+
+            expect(offsets).toEqual([0, 0]);
+            expect(sizes[0]).toBeCloseTo(100 / 3);
+            expect(sizes[1]).toBeCloseTo((100 * 2) / 3);
+        });
+
+        test('negative items are not inverted into positive shares', async () => {
+            const negativeItems: ChartItem[] = [
+                { text: 'First', value: -5 },
+                { text: 'Second', value: -10 },
+                { text: 'Third', value: -15 },
+            ];
+            const { sizes } = await setup({ items: negativeItems });
+
+            expect(sizes.every((size) => size < 0)).toBe(true);
+        });
+    });
+
+    describe('items on an axis', () => {
+        test.each(['bar', 'dot', 'line', 'area'] as const)(
+            'a %s chart shows no percentage',
+            async (type) => {
+                const { labels } = await setup({ type });
+
+                expect(labels).toEqual(['Applications', 'Photos']);
+            }
+        );
+    });
+});
--- a/src/components/chart/examples/chart-max-value.tsx
+++ b/src/components/chart/examples/chart-max-value.tsx
@@ -4,23 +4,18 @@ import { chartItems } from './chart-items-stack';
 /**
  * Using the `maxValue` prop
  *
- * The `maxValue` prop defines the upper limit of the visual range for the chart.
- * It determines the maximum value represented on the axis and is used to
- * calculate the size of each item in the chart relative to this value.
+ * For `stacked-bar`, `pie`, `doughnut`, and `ring`, `maxValue` is the
+ * denominator used as the whole. An item with value `10` occupies 10% when
+ * `maxValue` is `100`. The combined item size can exceed `maxValue`, producing
+ * a total above 100%. Without `maxValue`, the chart uses the combined item
+ * values or range extents.
  *
- * For example, if `maxValue` is set to `100`, an item with a value of `10`
- * will occupy 10% of the chart, while an item with a value of `50` will occupy 50%.
- * If `maxValue` is set to `200`, an item with a value of `50` will occupy 25% of the chart.
+ * For `bar`, `dot`, `area`, and `line`, `maxValue` sets the requested upper
+ * bound. The chart can round that bound up to the next `axisIncrement`. The
+ * lower bound includes zero and negative item values, rounded down to the
+ * previous `axisIncrement`.
  *
- * If `maxValue` is not provided, the chart will calculate the maximum value based on
- * the actual data points, and the size of each item will be calculated relative to
- * the total range of the data.
- *
- * :::note
- * The `maxValue` only affects the upper limit of the chart's range.
- * The chart will always start from the smallest value present in the dataset,
- * which could be a negative number.
- * :::
+ * `maxValue` has no effect on `nps` or `scatter`.
  *
  * @sourceFile chart-items-stack.ts
  */


src/components/chart/chart.tsx
```diff
@@ -108,6 +108,16 @@ export class Chart {
...
 
+    /**
+     * Hides the percentage that a chart item's tooltip appends to its `text`.
+     * The percentage is the item's share of the chart's range, which is only
+     * meaningful when that range is a whole, such as a total or a capacity.
+     * Has no effect on the `bar`, `dot`, `nps`, and `scatter` types, whose
+     * tooltips never show a percentage.
+     */
+    @Prop({ reflect: true })
+    public hideItemPercentage = false;
```
Kiarokh: i kind of don't like the "negative" name. Here is why:
- Firstly, it's not a convention in this repo. For such prop names, we normally use `displayXYZ` or `showXYZ`. I think `display` is the more common pattern. In this component for instance, we have `displayItemValue` or `displayItemText`
- negative words give a harder DX. It's harder to think: "doNotDoThisThing = false`

Next, I haven't looked at the PR yet. But is there no other way to not get the percentage? We should try our best to avoid adding new props. Specially if the prop only affects some types or modes of the component. In the chart component this is hard though, since there are always exceptions. But in general, I wanna make sure that you have thought about it, and found no other way

adrianschmidt: @Kiarokh What is this percentage value actually meant to tell the user? The current way percentage values are calculated in stacked bar charts makes no sense to me. The value shows each block's size as a percentage of the biggest block:

image
image

Where is this used, and for what?

I would argue that we should remove the percentage value display by default, and if it's actually used for something somewhere, we can add an opt-in. But if we do add an opt-in, we need to add an example to the docs, that explains how it's meant to be used.

adrianschmidt (dismissed):
Just adding a request for changes so this doesn't get approved and merged until we've hashed things out 😊

src/components/chart/chart.tsx
jecrapo: Hey @Kiarokh 

~~Yes, this was kind of a tough decision.~~

~~Either I add the first "hide*" style property defaulting to `false` in order to hide the percentage...~~

~~...or~~

~~... I break another code standard in lime-elements and introduce the first ever prop defaulting to `true`.~~

I agree with Adrian above that the best thing would be to hide percentage by default and add an opt-in prop `displayPercentage`. Assuming any one actually needs the percentage.

But I didn't want to be destructive and change behavior without knowing what consumers might be out there.

Let's hash this out :D

jecrapo: For context:
See the screenshot in https://github.com/Lundalogik/lime-crm-components/pull/4652. Without this change in limel, each segment in a lane/bucket showed a percentage which was relative to the lane/bucket with the max height. So percentages didn't even sum to 100% in each lane/bucket. That became super confusing, and the reason this change is blocking for the feature in that PR.

Kiarokh: Looking at your comments, I think the problem are two then.
1. Percentages are wrong or meaningless
2. You as a consumer want to hide them, even if they were right.

Is my understanding correct?

In that case, I can first answer Adrian's question about percentages.

## 1. Tooltips percentages, why are they there
Looking again on the "Max Value" example in the documentations, I remember that the idea of displaying percentages in tooltips was to help end users understand what "8 hours" on a chart item means, in terms of percentages of the total, for instance in a pie, ring, bar, or stacked-bar chart. If sum of all items together is 80 hours, then 8 hours (one of the items) is 10% (of the total time). However, if the consumer has set a `maxValue` (e.g. Total Disk space), the 8 megabytes on one of the items should be calculated by that `maxValue` instead, so 8 megabytes of 128 megabytes, for instance. That's why we had the percentages in the tooltips, to give some better understanding. The percentages in the tooltips, would then align with the visual percentage of each item. The visual size = percentage size, would however be true only for pie/doughnut/stacked-bar/ring but Not for bar/dot/line/area. There the visual is deliberately height-on-the-axis (a bar chart scales to its tallest bar), while the tooltip is share-of-the-whole. With no `maxValue` the tallest bar is 100% tall but is not 100% of the total, and those two numbers should stay different. Also in those charts, we support negative values, which would make the concept of "percentage of the whole" really weird. 

## 2. Consumers want to hide the percentages
I'm not sure if this will be the case, if the percentages in the tooltips would be right. But if that's the case, then we should add a new prop and we better call it `displayItemPercentage = true` and let it be an opt-out for the consumers.

### Why are tooltip percentages wrong?
Claude's analysis 👇 
The tooltip prints `size` from `calculateSizeAndOffset` — the item's fraction of the **rounded visual axis range**, not of the whole. One number is being asked to do two different jobs, and three things go wrong.

**1. The denominator is the largest item, not the sum, for every type except `pie` and `doughnut`.** `calculateRange` substitutes the sum only for those two, so the documented `maxValue` fallback never happens for `stacked-bar`. In our own stacked-bar example the denominator is 80 (the Videos item) instead of 269, so the tooltips sum to **336.25%** and Videos reads `100.00%`. That is exactly Adrian's screenshot.

The reason nobody caught it is that the bar still *looks* right. `_stacked-bar.scss` sets `width: calc(var(--limel-chart-item-size) * 1%)` on flex items with the default `flex-shrink: 1`, so a 336% basis overflows and flexbox shrinks every item in proportion — landing on `value / Σvalues`, the correct share. Measured in Chromium: rendered widths 3.72% / 19.33% / 29.74% …, matching `value/sum` to two decimals, while the printed numbers sum to 336.25%. The visuals have been right by accident, and the tooltip was the only honest report of the calculation.

**2. `axisIncrement` rounds the denominator on types that have no axis.** `visualMaxValue = ceil(finalMaxRange / axisIncrement) * axisIncrement`, and the increment is derived from the item values, never from `maxValue`, so it rarely divides it. But `renderAxises()` returns early for `stacked-bar`, `pie`, `doughnut` and `ring` — those draw no axis lines at all, and get their denominator rounded anyway:

- The **ring example on the docs site** (`maxValue={10}`, an 8h item): increment `0.8`, denominator `10.4`, tooltip says `76.92%` where it should say `80.00%`.
- A **pie with no `maxValue`**: denominator 272 for a total of 269, so slices sum to 98.90% and the pie leaves a visible sliver of gap.
- **`maxValue={100}`** with items 10/20/30: increment 3, denominator 102, so a value of 10 reads `9.80%`.

John already found this one and cancels it in Lundalogik/lime-crm-components#4652 by setting `axisIncrement` equal to `maxValue`, which makes the `ceil` a no-op.
**3. `line` and `area` print a number that is neither.** The skip list is `bar`/`dot`/`nps`/`scatter`, and `line` and `area` were left off it — so they print `value / axisRange`, which is how far up the axis a point sits, not a share of anything.

### The fix

Split the two jobs `size` is doing:

- **geometry** — `(value − axisMin) / axisRange`, ceil-rounded so grid lines land on round numbers. Needed by the types that draw a grid: `bar`, `dot`, `line`, `area`, `scatter`.
- **share** — `value / (maxValue ?? Σvalues)`, unrounded. This is the tooltip number.

And show the share for exactly the four types whose choice already asserts that the items are parts of a whole: `stacked-bar`, `pie`, `doughnut`, `ring`. Those draw no grid, so for them geometry simply *becomes* the share, the two agree by construction, and the dependency on flex-shrink goes away.

That means the skip list grows rather than shrinks — `line` and `area` join it. And `bar`/`dot` stay out for a principled reason rather than by accident: our own bar examples use `chart-items-with-negative-values.ts`, which is temperatures (`11.5°, −15°, 38°, 0° …`). There is no total temperature, so Kiruna is not `−17.24%` of anything. Whether a bar chart's values sum to a meaningful quantity is a property of the unit, and only the consumer knows it.

If we later want the percentage on an all-positive bar chart, `displayItemPercentage` widens from `= true` to `?: boolean` — undefined derives from the type, `true` is the consumer asserting the values are summable, `false` hides it. Cheap to do on a `@beta` component, and not worth doing now, since `bar` has never shown a percentage at all.

**Unrelated, but in the same function:** the auto-derived increment is cached into the `axisIncrement` prop and never refreshed, so when `items` are replaced with data of a different scale, the grid and the denominator both stay stale.

etc/lime-elements.api.md
```diff
@@ -249,6 +249,7 @@ export namespace Components {
         "displayAxisLabels": boolean;
         "displayItemText": boolean;
         "displayItemValue": boolean;
+        "hideItemPercentage": boolean;
```
Kiarokh: Issue's description says:

> **Why hideItemPercentage rather than displayItemPercentage**
> The percentage is on by default, so the prop has to switch it off. Every boolean prop in lime-elements defaults to false, which is also how an HTML boolean attribute works, so the prop takes the shape of disabled instead of adding the library's first default-true boolean.

This is not a correct claim. There are many documented as "Set to false to […] ") e.g. `showCopyButton`, `sortableColumns`, `allowResize`, `clearAllButton`, `show3dEffect`.

jecrapo: You are right! That was a sloppy search on my end

src/components/chart/chart.tsx
jecrapo: Thanks @Kiarokh, the analysis matches what I see. You are also right about the default-true booleans, I will fix the description.

Some context on why the fix alone does not get my chart there, then three ways out. Your call, but a quick one helps me: this blocks the admin usage page for AI Assist GA.

**Context.** The usage page is twelve stacked bars side by side, one per month, and `limel-chart` draws one bar. I build the grid the way you described in [the 2024 review](https://github.com/Lundalogik/lime-elements/pull/3240#discussion_r1836319647): several elements with the same `maxValue`. That makes `maxValue` mean two things at once, the whole for the percentage and the shared scale for the grid. So even with the fix, my tooltips would read "share of the tallest month".

**Options**

1. **Fix only, no prop.** I land your fix as a `fix(chart)` PR. My grid drops `maxValue` and gets the shared scale from CSS heights, so each bar is its own whole and the tooltip shows a segment's share of its month. Smallest change to lime-elements.
2. **Fix plus `displayItemPercentage = true`.** Same as 1, plus the opt-out you named, for consumers who want a plain tooltip. A little more API, more freedom in how a chart reads. This is where I lean.
3. **Fix plus a real chart set.** The multi-bar case [Johan raised in 2024](https://github.com/Lundalogik/lime-elements/pull/3240#discussion_r1836227380). Several series in one component, one scale, one legend. The right thing long term, and a bigger PR. I am open to it if you want it now.

I can write the fix from your analysis either way. Which option do you prefer?

adrianschmidt: Wait, is this 12 separate instances of `limel-chart`? 😲

image

If that's the case, then limel-chart doesn't have a "stacked bar-chart" feature. A stacked bar-chart is what's shown above ☝️

This 👇 is just a single stacked bar, not a stacked bar-chart:

image

jecrapo: Yes, that's 12 separate `limel-chart` instances. To the long term fix is a much bigger PR in lime-elements (suggestion 3 above). Im willing to go down that road, but it requires reviewer availability since it will involve some design decision that might require quite a few iterations and time is kind of scarce now since we need to get things in the train for AI Assist GA.

One option is of course to use a third party graph in the Usage admin page for now, and swap to limel at a later stage, but I'd like to use lime-elements if possible

jecrapo: @Kiarokh @adrianschmidt Restructured the PR. The percentage fix from the analysis above is one commit, `displayItemPercentage = true` is a second, and a small docs example is the third. The description is updated. A stacked bar chart with several bars in one element is now #4263.

src/components/chart/examples/chart-display-item-percentage.tsx
Kiarokh, line 1: I think this doesn't need to be a separate `docs` commit. Since reverting the other commit (if it happens one day) will be hard to remember or track. Please squash it to the `feat` commit 🙏

src/components/chart/chart.tsx
Kiarokh: > **Options**
> 
> 1. **Fix only, no prop.** I land your fix as a `fix(chart)` PR. My grid drops `maxValue` and gets the shared scale from CSS heights, so each bar is its own whole and the tooltip shows a segment's share of its month. Smallest change to lime-elements.

This must be fixed anyways. So perfect with the fix 👍 

> 2. **Fix plus `displayItemPercentage = true`.** Same as 1, plus the opt-out you named, for consumers who want a plain tooltip. A little more API, more freedom in how a chart reads. This is where I lean.

I am fine with the opt-out feature. I can't however defend it as a must have. But if you as a consumer really feel it's not necessary or distractive for users, then we could offer it as a feature.

> 3. **Fix plus a real chart set.** The multi-bar case [Johan raised in 2024](https://github.com/Lundalogik/lime-elements/pull/3240#discussion_r1836227380). Several series in one component, one scale, one legend. The right thing long term, and a bigger PR. I am open to it if you want it now.

Chart and Char-set are a bit like Chip and Chip-set. One is fine on its own. But a set can be combined and presented in various ways. There are many types of charts that can be in a chart-set. But normally the ones which have X/Y axises. For example:
image

I don't think the stacked bar chart is necessarily one of them. It can be, but what you're doing is also completely fine. If we ever make a `limel-chart-set` component, we could offer a subset of `type`s from `lime-chart` in it. So consumers can choose. But keep in mind, `limel-chart` will have its own "Chart legend" feature at some point. And a chart-set's legend might be a bit complicated to visually produce. Anyhow… I'd keep those discussions away from this issue. Just wanted to say that your current consumption of limel-chart is fine in my opinion.

src/components/chart/examples/chart-display-item-percentage.tsx
jecrapo: Good point! And done :)

src/components/chart/chart.tsx
adrianschmidt: A stacked bar chart is not the same as a "limel-chart-set" with multiple of the separate "stacked bars" that is currently supported.

This is a stacked bar chart:
image

Note the X and Y axis.

adrianschmidt:
# Consolidated PR Review

_Automated 7-agent review (backward compatibility, code quality, architecture, security, observability, performance, Lime platform rules). Findings below were measured against a built `main`, not inferred from the diff._

## PR Summary

The `fix` commit makes `stacked-bar`, `pie`, `doughnut`, and `ring` size items and print tooltip percentages as shares of `maxValue` or the item sum, unrounded and without the `axisIncrement` ceiling, and stops `line` and `area` from printing a percentage. The `feat` commit adds `displayItemPercentage` (default `true`), a docs example, and a new `chart.spec.tsx` with 11 tests. Unblocks the AI usage chart in lime-crm-components #4652.

The PR is strictly better than `main`: every mainline case was measured and is now correct, the default stacked bar is a screen no-op, and no dimension found a regression. Two Medium follow-ups are worth landing before or right after merge, one of them a one-token hardening.

- **Blockers:** None.
- **Non-blocking but worth addressing:** see Top Recommendations 1–4 (non-positive whole guard, stale `chart-max-value` example prose, range-item sum, downstream ring consumers).

---

`limel-chart` is `@beta` (README states beta components may break without a major bump), so `fix`/`feat` without `!` is correct here. API report change is additive and CI-verified. Old and new `calculateRange` were reimplemented and diffed across all documented examples and every consumer found.

- **[Low]** Ring charts without `maxValue` change on screen. Old denominator was the ceil-rounded largest item, so the biggest ring always filled the circle; now it is the sum (items 10/5/3: 100/50/30% → 55.56/27.78/16.67%). `_ring.scss:42` feeds size straight into `conic-gradient`, so this is visible. `Lundalogik/limepkg-graph-view` renders user-chosen `type` including `ring`/`pie`/`doughnut`/`stacked-bar`, never passes `maxValue`, and passes `axisIncrement={1}` everywhere. Correct behavior and named in the commit body, but no downstream owner has been told.
- **[Low]** `axisIncrement` is now a silent no-op on the four whole types (documented in the new docblock). Whole types also no longer reflect an auto-derived `axis-increment` attribute; nothing selects on it.
- **[Low]** `line` and `area` lose their tooltip percentage with no opt-in path (`chart.tsx:478` gates on `isPartOfWhole()`). Settled intent per the PR thread.
- **Positives:** The change reaching the most consumers (default stacked bar) is provably a screen no-op, measured to within 0.002 points in both orientations. lime-crm-components #4652 no longer sets `maxValue`/`axisIncrement` at all and strictly improves.

## 2. Code Quality -- GOOD ✅

Spec runs and passes (11 tests).

- **[Medium]** The zero-guard at `chart.tsx:536-537`, `(this.maxValue ?? sum) || 1`, catches only exact `0`. A negative `maxValue` or negative sum passes through as a negative `totalRange`, and a sum that cancels to zero trips the guard: items `[+10, −10]` on the default `stacked-bar` print `1000.00%` / `−1000.00%` (main printed `50%` / `−50%`); `[−5, −10, −15]` renders the most negative item largest; `maxValue: 0` prints `2500%`/`7500%` (main: `NaN%`). Negatives were never a supported input for whole types (no negative-value styling exists for them) and the old output was nonsense too, so not a regression. Sibling guards `normalize` (line 598) and `calculateNiceStep` (line 566) both test for a non-positive span. Fix: `whole > 0 ? whole : 1`.
- **[Medium]** `calculateWholeRange` (`chart.tsx:532-535`) sums `getMaximumValue`, which for a `[start, end]` tuple is the end, not the extent. `chart.types.ts:18-21` documents tuples as ranges "for most types", and `calculateSizeAndOffset` honours them for all non-scatter types. Gantt-shaped items `[1,10] [10,40] [40,100]` on `stacked-bar` go from 9/30/60% to 6/20/40%. `_stacked-bar.scss` never reads the offset, so this combination was never a coherent timeline on main either; reach is narrow. Either sum extents or have whole types ignore start values.
- **[Medium]** Pre-existing, explicitly deferred in the thread: the auto-derived increment is cached into the public `axisIncrement` prop (`chart.tsx:505-509`) and never cleared. Measured: a `bar` with items 25/75 derives increment 8; replacing items with 2500/7500 keeps 8 and renders 939 `.axis-line` divs each holding a `limel-badge`. The PR improves this for whole types (they no longer write the prop or trip Stencil's immutable-prop warning). Deserves a follow-up issue.
- **[Low]** `WHOLE_TYPES` is `ReadonlySet<string>` (`chart.tsx:14`). `ReadonlySet<Chart['type']>` compiles and rejects a typo with TS2322.
- **[Low]** The new `axisIncrement` sentence "Has no effect on chart types without axis lines" (`chart.tsx:173`) invites a false converse: `scatter` draws axis lines and ignores `axisIncrement`.
- **[Low]** `maxValue` docblock (`chart.tsx:166`) says axis types default to "the largest value"; it is the largest value rounded up to the next axis step (bar 25/75 → 80).
- **[Low]** Two of three new comments lead with a what-summary: `chart.tsx:11-12` second sentence and `526-527` restate the code. `528-529` (why no `ceil`) is the keeper; `536` is a real why with a too-narrow premise.
- **[Low]** Example opens in the non-default state: `chart-display-item-percentage.tsx:20` starts `false` while the prop defaults `true`. Sibling opt-out examples (`text-editor-allow-resize.tsx:39`, `code-editor-copy.tsx:20`, `card-3d-effect.tsx:14`) start at the component default.
- **[Low]** Test fixture sums to exactly 100, so size assertions are also satisfied by raw pass-through (they do fail on old code). A 30/90 fixture separates the cases. Coverage gaps: `doughnut` (only untested whole type), all-zero branch, cancelling values, `maxValue` below sum. The type-change test duplicates `setup`'s extraction (`chart.spec.tsx:82-89`).
- **Positives:** Measurably correct on every mainline case (`stacked-bar` 31.25/93.75 → 25/75, `pie` 24.04/72.12 → 25/75, `ring` 9.77/29.30 → 10/30). One predicate now drives both sizing and tooltips. The exclusion half is tested, not just inclusion.

## 3. Architecture -- GOOD ✅

Two ad-hoc special cases (pie/doughnut sum substitution, four-type negated skip list) collapse into one predicate plus one pure calculator. The `@Watch('type')` closes a real cache-staleness gap.

- **[Low]** When `maxValue` is below the item sum, `_stacked-bar.scss:35`'s flex-shrink dependency silently re-engages: with `chart-items-stack` and `maxValue=100`, Videos' tooltip says 80% while it renders at 29.74%. Same class of divergence main had, so not a regression, but the new comment at `chart.tsx:528` ("has to be its exact share") and the `maxValue` docblock overstate the guarantee. Narrow the comments or normalise when the sum exceeds the whole.
- **[Low]** `AxisRange.increment` is filled with the whole for axis-less types. Nothing reads it there, and a zero would make the loop at `chart.tsx:288` never terminate, so it is the safe filler, but that reasoning is unwritten. Make it optional or say why.
- **[Low]** Three type lists exist (`WHOLE_TYPES`, the axis whitelist at `chart.tsx:259`, ad-hoc checks at 327/373) and `nps` is in none. The `type` union is inlined in the `@Prop` rather than exported from `chart.types.ts`, so none is compiler-checked.
- **Positives:** Pie conic-gradient math is now satisfied rather than broken: slices close the circle instead of leaving a sliver. The stated extension path (widening the prop to `?: boolean`) is a one-operator change at `chart.tsx:478`.

## 4. Security -- SAFE ✅

No findings. Tooltip `label` and `helperLabel` render as text vnodes (`tooltip-content.tsx:55,67`, no `innerHTML` anywhere in chart or tooltip). The CSS custom property is set via `style.setProperty`, which rejects declaration break-outs; every stylesheet consumer keeps it inside `calc()`. Measured in Chromium against eight payloads including `;`, `}`, `!important`, and `url()`.

- **Positives:** The value flowing into the tooltip is arithmetic output, not interpolated text, and the rendering path is byte-identical to main.

## 5. Observability -- GOOD ✅

- **[Medium]** The dedicated `maxValue` docs example still describes the old rule. `examples/chart-max-value.tsx:15-23` promises the chart "will always start from the smallest value present in the dataset, which could be a negative number" and frames the default as "the maximum value based on the actual data points". `calculateWholeRange` hard-codes `minValue: 0` and uses the sum. The prop docblock was rewritten; the example page it points at was not.
- **[Low]** No dev-time signal for `maxValue <= 0`. The ring example binds an unconstrained numeric input to `maxValue`, so a docs visitor can type 0 and get `2500%`. The repo has precedent: `warnOnConflictingMovableAndSortable` at `table.tsx:480-491`. Closing the Code Quality Medium makes this moot.
- **[Low]** `axisIncrement` becomes a silent no-op on whole types. Documented decision.
- **Positives:** Eliminates literal `NaN%` on every degenerate input constructed. The `@exampleComponent` tag auto-enrols the new example in the console-error gate and axe scan.

## 6. Performance -- SAFE ✅

Watcher graph traced against Stencil runtime source and confirmed at runtime.

- **[Low]** Pre-existing, bounded double computation when switching into an axis type: `chart.tsx:506` writes `axisIncrement`, which re-enters `handleChange` and computes a throwaway range. Depth-bounded at 2, still one render. The new `@Watch('type')` does not widen it: the whole-range branch returns before the write.
- **Positives:** Zero added renders (type changes already scheduled one). Item passes drop from 4 to 1 on whole types and 4 to 3 on axis types, and the whole path drops a spread `Math.max` that would blow the stack on huge lists. Exact-share sizing removes the flex-shrink overflow step. Examples and spec are excluded from the published bundle (`tsconfig.dist.json`).

No LPID rule fires: no network calls, `range`/`secondaryAxisRange` are plain private fields rather than `@State`, `calculateRange()` is unreachable from `render()`, and the tooltip label reaches no `innerHTML` sink.

- **Positives:** The whole-range path is cleanly separated from the axis path, so axis-free types no longer inherit rounding they never render.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Backward Compatibility | **GOOD** ✅ |
| Code Quality | **GOOD** ✅ |
| Architecture | **GOOD** ✅ |
| Security | **SAFE** ✅ |
| Observability | **GOOD** ✅ |
| Performance | **SAFE** ✅ |
| Lime Platform Issues | **SAFE** ✅ |
| **Merge Readiness** | **MERGE WITH CAVEATS** ⚠️ |

## Top Recommendations

1. **[Medium from Code Quality]** Replace `(this.maxValue ?? sum) || 1` at `chart.tsx:537` with a positivity clamp such as `whole > 0 ? whole : 1`, matching the sibling guards in `normalize` and `calculateNiceStep`. Add spec cases for `[+10, −10]`, all-zero items, and `maxValue: 0`. Widen the comment on line 536 to say non-positive, not zero.
2. **[Medium from Observability]** Rewrite the prose in `examples/chart-max-value.tsx:15-23` to match the new rule: whole types start at 0 and default to the sum; axis types start at the floored minimum and default to the rounded largest value.
3. **[Medium from Code Quality]** In `calculateWholeRange`, either sum extents (`getMaximumValue − getMinimumValue` for range items) or document that whole types treat a tuple's start as 0. Add one spec case with a `[start, end]` item on `stacked-bar`.
4. **[Medium from Code Quality]** File a follow-up issue for the pre-existing stale `axisIncrement` cache (939 axis lines after an items rescale); hoisting the derived increment into a private field is the fix, out of scope here.
5. **[Low from Backward Compatibility]** Post a heads-up to the `limepkg-graph-view` owners that ring/pie/doughnut without `maxValue` will look different after they bump, and that their `axisIncrement={1}` is now ignored for those types.
6. **[Low from Code Quality]** Type `WHOLE_TYPES` as `ReadonlySet<Chart['type']>`, or export the `type` union from `chart.types.ts` and type both `WHOLE_TYPES` and the axis whitelist against it.
7. **[Low from Code Quality]** Start the example's `displayItemPercentage` state at `true` so the page opens at the component default, as the sibling opt-out examples do.
8. **[Low from Code Quality]** Fix the two docblock inaccuracies: `axisIncrement` also has no effect on `scatter`; axis types default `maxValue` to the largest value rounded up to the next step.
9. **[Low from Architecture]** Narrow the `chart.tsx:528` comment to say the exact-share guarantee holds when `maxValue` is at least the sum, or normalise when the sum exceeds the whole.
10. **[Low from Code Quality]** Trim the what-summary sentences at `chart.tsx:11-12` and `526-527`; change the spec fixture to values not summing to 100 and add a `doughnut` case.

adrianschmidt (dismissed):

jecrapo:
Thanks for the thorough review, Adrian. I pushed two unsquashed fixups for the points that needed changes:

- Range-valued items in whole-based charts now use their extent (`end - start`) for both the item size and the inferred whole. A regression test covers tuple values.
- The `maxValue` and `axisIncrement` documentation now distinguishes whole-based charts from axis charts. It also states that `nps` and `scatter` ignore these props where applicable.

I left the non-positive input handling unchanged because the suggested clamp would not correct zero or cancelling totals, and negative values do not have defined part-of-whole semantics. The stale automatic `axisIncrement` after item changes predates this PR, so I kept it out of scope. The graph-view sizing change is intended: the tiles now occupy their shares of the full aggregated whole.

Could you take another look?

adrianschmidt:
# Consolidated PR Review (re-review)

_Automated 7-agent re-review of head `bde8d9ea7`, after the two fixups that followed the 2026-09-03 09:40 review. Findings were measured against the built worktree, not inferred from the diff._

## PR Summary

The `fix` commit makes `stacked-bar`, `pie`, `doughnut`, and `ring` size items and print tooltip percentages as shares of `maxValue` or the item sum, unrounded, and stops `line`/`area` printing a percentage. The `feat` commit adds `displayItemPercentage` (default `true`), a docs example, and `chart.spec.tsx` (12 tests, passing). The two fixups sum range items by extent via `getValueSpan` and rewrite the `maxValue`/`axisIncrement` prose. 5 files, +261/−38.

## Merge Readiness -- BLOCKED ❌

The PR is better than `main` on every mainline case, and the fixups resolved both prior Mediums they targeted. But the extent fixup left `calculateSizeAndOffset` computing offsets on an absolute scale while the whole is now a sum of extents, and `ring` is the one whole type whose stylesheet reads the offset. That combination rendered coherently on `main` and does not now.

- **Blockers:**
  1. **[High from Architecture]** `ring` with `[start, end]` items draws a truncated or missing arc. One-branch fix plus one spec case.
- **Non-blocking but worth addressing:** Top Recommendations 2–5 (negative inferred sum, random vdom keys follow-up, overclaiming prose, `console.warn`).

---

`limel-chart` is `@beta`; the API report change is purely additive; the reflected default-`true` boolean has five in-repo precedents. Every `limel-chart` consumer in the org was enumerated: the only one on a caret range that auto-adopts 40.x (`lime-crm-components`) renders `bar` and is untouched. `limepkg-graph-view` pins `^38.43.0`, so the ring resizing the prior review flagged arrives only on a deliberate major bump. Prior Lows on `axisIncrement` no-op and `line`/`area` are documented intent, informational.

- **[Low]** The extent fixup is a second behavioral shift for `pie`/`doughnut` with tuple items on top of the announced one: `[10,40]`/`[40,100]` go from 21.43/42.86% (35.7% gap) on `main` to 33.33/66.67%. Correct and consistent with how sizes were always computed, but not named in the PR description.
- **[Low]** A single-item `stacked-bar` or `ring` without `maxValue` now fills 100% where `main` left a gap (value 73: 91.25% → 100%). No consumer found that hits it.
- **[Low]** No spec covers a tuple item together with an explicit `maxValue`, which is exactly the path the second fixup changed.
- **Positives:** The only auto-adopting consumer is unaffected, and the `solution-dev-platform` ring gauge becomes more accurate (73% against target 100 instead of 70.19%).

## 2. Code Quality -- GOOD ✅

Spec run from the worktree: 12 passed. Prior range-extent Medium is RESOLVED and structurally so: tooltip and rendered size derive from one extent definition. Prior docblock Lows are RESOLVED. The declined zero-guard is two-thirds sound.

- **[Medium]** The author's decline of the positivity clamp is right about zero, cancelling totals, and negative `maxValue` (the clamp would turn invisible negative widths into 2500% overflow). It does not cover a **negative inferred sum**: items `[−5, −10, −15]` on `stacked-bar` render 16.67/33.33/50%, a plausible-looking inverted bar, where `main` rendered a visibly empty one. The narrow fix `this.maxValue ?? (sum > 0 ? sum : 1)` at `chart.tsx:539` restores `main` for that row and leaves all other degenerate rows unchanged. Undefined input with no consumer today, so non-blocking.
- **[Low]** `getValueSpan` (`chart.tsx:629-635`) duplicates the start-value logic inline in `calculateSizeAndOffset` (`431-434`). Routing the latter through the former is also the natural place for the Section 3 fix.
- **[Low]** Guard comment at `chart.tsx:538` says "Every item is zero" but the guard also fires on cancelling totals and `maxValue: 0`, both of which the author's own reply names.
- **[Low]** `calculateWholeRange` docblock (`528-529`) opens with a what-summary and repeats the `WHOLE_TYPES` docblock; keep only the why-no-ceil sentence. The `WHOLE_TYPES` docblock itself earns its place (disagreeing with the prior review).
- **[Low]** `WHOLE_TYPES` is `ReadonlySet<string>` (`chart.tsx:14`); `Set<Chart['type']>` builds, and a typo then fails the build instead of silently dropping a type from the whole branch.
- **[Low]** Coverage: `doughnut` untested; no degenerate-whole case; 25/75 fixture also satisfied by raw pass-through; type-change test re-implements `setup`'s extraction.
- **[Low]** `chart-display-item-percentage.tsx:20` opens at `false` while the prop defaults `true`; sibling opt-out examples start at the component default.
- **Positives:** Degenerate input got measurably better (all-zero prints `0.00%` where `main` printed `NaN%`), and the rewritten docblocks are accurate including the `nps`/`scatter` exclusions, verified against the stylesheets.

## 3. Architecture -- ACCEPTABLE ⚠️

One predicate plus one pure calculator replaces two ad-hoc special cases, and `@Watch('type')` closes a real staleness gap. The extent fixup, though, left two scales in one function.

- **[High]** `calculateSizeAndOffset` (`chart.tsx:436-444`) computes size as `(end − start) / totalRange` but offset as `(start − minValue) / totalRange`, an absolute position. `calculateWholeRange` sets `totalRange` to the sum of extents and `minValue` to 0, so for whole types the offset is on a different scale than the size. `_ring.scss:42-54` draws each ring's arc to `offset + size`. Measured on the PR head: `ring` with `[10,40]`/`[40,100]` gives the second item offset 44.44 + size 66.67, arc stop at 111%, truncated at the full turn; disjoint `[0,10]`/`[20,30]` gives offset 100 + size 50, nothing drawn. A negative start (`[−20,20]`) gives offset −25, so the arc ends at 25% while the tooltip says 50%. `main` rendered both cases closing at exactly 100%. `chart.types.ts:18-21` documents tuples as ranges "for most types", and the PR's own second fixup and spec test make tuples on whole types a supported semantic, so this is a new defect on a supported input. `pie`/`doughnut` are safe (offset overwritten with cumulative at `chart.tsx:329-332`); `stacked-bar` never reads the offset. Fix: for whole types, offset is 0 and size is `getValueSpan(item) / totalRange`. Rated Medium and Low by the three agents that found it because reach is narrow; re-derived to High because narrow reach does not change that it is a new rendering bug versus `main`.
- **[Low]** Prose overclaims when `maxValue` is below the item sum: `chart.tsx:162-169` ("defines the whole"), `chart-max-value.tsx:7-8` ("contains all items"), and the comment at `530-531` ("exact share"). Measured: items 25/75, `maxValue: 50` yields a tooltip of `150.00%` while flex-shrink squeezes the segment. Same class as `main` (133.93%), not a regression. One clause saying the whole is a scale, not a cap, fixes the prose.
- **[Low]** `AxisRange.increment` is filled with the whole for axis-less types against a field documented as "Step between adjacent axis lines"; verified unreachable since `renderAxises` whitelists axis types only.
- **[Low]** Three type lists (`WHOLE_TYPES`, axis whitelist at `261`, ad-hoc checks) with `nps` in none, and the `type` union inlined in the `@Prop` rather than exported. Path to #4263 is open; `calculateSizeAndOffset` reading `this.range` implicitly is the one function a per-series refactor must change, and this PR does not add that coupling.
- **Positives:** The whole-range branch is cleanly separated from the axis branch, which is the shape a multi-series follow-up needs.

## 4. Security -- SAFE ✅

No findings. Tooltip `label`/`helperLabel` render as text vnodes (`tooltip-content.tsx:55,67`); no `innerHTML`-family sink in chart, tooltip, or portal. `--limel-chart-item-size` goes through Stencil's `style.setProperty` path (CSSOM, not string interpolation). `toFixed(2)` cannot throw on non-finite input. New example is auto-enrolled in the console-error gate and cannot trip the immutable-prop warning.

- **Positives:** Exposure moves the right way: two types stop emitting a derived number, none starts.

## 5. Observability -- GOOD ✅

Prior Medium (stale `chart-max-value` prose) is RESOLVED; every claim in the rewrite was executed and holds, including axis rounding in both directions and the `nps`/`scatter` exclusions.

- **[Low]** No dev-time signal for a non-positive whole, and `examples/chart-type-ring.tsx:43-48` binds an unconstrained numeric input to `maxValue` (typing `0` prints `1000.00%`). The author's clamp decline is sound, which makes a `console.warn` in `calculateWholeRange` the only remaining remedy; precedent is `warnOnConflictingMovableAndSortable` in `table.tsx`.
- **[Low]** Informational: switching `pie` → `bar` at runtime now emits Stencil's immutable-`@Prop` warning for `axisIncrement` where `main` emitted none (it kept the stale range). Pre-existing root cause, explicitly out of scope; whole types at load now emit nothing, a net reduction in console noise.
- **Positives:** The rewritten prose is accurate on every executable claim.

## 6. Performance -- GOOD ✅

`calculateRange` is never on the render path; the whole path does strictly less work (one `reduce`, no n-ary spread). Real dist build confirmed examples and spec are excluded; chart entry grows 1484 bytes unminified.

- **[Medium]** Pre-existing, untouched: `chart.tsx:324` calls `createRandomString()` per item inside `renderItems` and uses it as `key`, so every render destroys and recreates every `<tr>` and `<limel-tooltip>`. Verified: no node survives a `maxValue` or `displayItemPercentage` change. Dominant GC source in the component and multiplied by instance count on the #4263 page. Follow-up issue; not this PR.
- **[Low]** Switching from a whole type into an axis type now renders twice (measured 2 renders / 2 `calculateRange` vs 1 / 0 without `@Watch('type')`), because the `axisIncrement` write at `507-511` re-enters `handleChange` and Stencil's snapshotted flags let `scheduleUpdate` run again. Bounded, avoided entirely when `axisIncrement` is passed, and the correct trade: without the watch, whole → `scatter` throws on a null `secondaryAxisRange`. The prior review's "zero added renders" positive was wrong.
- **Positives:** Whole types no longer write the `axisIncrement` prop at load, saving a re-entrant `handleChange`, a throwaway range, and an attribute reflection on first render.

Catalogue fetched via the engineering-docs MCP (index plus LPID-FE-6, FE-7, FE-9, FE-10). No rule fires: `range`/`secondaryAxisRange` are plain private fields, `calculateRange()` is unreachable from `render()`, the tooltip label reaches a text vnode, and the whole-derived `increment` can never reach the axis-line loop because `renderAxises` returns early for whole types.

- **Positives:** Splitting the whole path from the axis path keeps a non-positive `increment` structurally out of the axis-line loop.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Backward Compatibility | **GOOD** ✅ |
| Code Quality | **GOOD** ✅ |
| Architecture | **ACCEPTABLE** ⚠️ |
| Security | **SAFE** ✅ |
| Observability | **GOOD** ✅ |
| Performance | **GOOD** ✅ |
| Lime Platform Issues | **SAFE** ✅ |
| **Merge Readiness** | **BLOCKED** ❌ |

## Top Recommendations

1. **[High from Architecture]** In `calculateSizeAndOffset`, branch on `isPartOfWhole()`: return `offset: 0` and `size: getValueSpan(item) / totalRange * PERCENT`. Add a `ring` spec case with `[10,40]`/`[40,100]` asserting `offset + size <= 100` for every item, and one with a negative-start tuple asserting a non-negative offset. This also removes the `getValueSpan`/`calculateSizeAndOffset` duplication.
2. **[Medium from Code Quality]** Change `chart.tsx:539` to `this.maxValue ?? (sum > 0 ? sum : 1)` and widen the comment to name cancelling and negative sums. Add a spec case for `[−5, −10, −15]`.
3. **[Medium from Performance]** File a follow-up issue for the per-render `createRandomString()` keys at `chart.tsx:324`; derive the id from the item index or hoist it into a field. Bundle it with the stale-`axisIncrement` follow-up the author already agreed belongs outside this PR.
4. **[Low from Architecture]** Add one clause to the `maxValue` docblock, `chart-max-value.tsx`, and the comment at `chart.tsx:530` stating that a sum above the whole overflows rather than being contained.
5. **[Low from Observability]** `console.warn` in `calculateWholeRange` when the whole is not positive, and give the ring example's `maxValue` input a minimum like the sibling `chart-axis-increment.tsx` does.
6. **[Low from Code Quality]** Type `WHOLE_TYPES` as `ReadonlySet<Chart['type']>`; trim the what-summary at `chart.tsx:528-529`; add `doughnut`, a tuple-plus-`maxValue`, and a non-100-sum fixture to the spec; reuse `setup`'s extraction in the type-change test.
7. **[Low from Code Quality]** Start the example's `displayItemPercentage` state at `true`.
8. **[Low from Backward Compatibility]** Mention in the PR description that tuple items on `pie`/`doughnut` also change (extent instead of end value), since it is a second behavioral shift.

jecrapo:
Thanks, Adrian. I addressed the ring regression, the negative inferred total, and the documentation findings in two unsquashed fixups:

- Whole-based charts now calculate size from `getValueSpan()` and use an offset of zero. This fixes the truncated ring arcs and removes the duplicate start-value calculation. A ring test covers tuple values with an explicit `maxValue` and checks both sizes and offsets.
- A non-positive inferred total now falls back to `1`, so all-negative items do not turn into positive shares. The existing fallback for an explicit `maxValue` of zero remains unchanged.
- The component docs, example, and PR description now call `maxValue` a scale rather than a cap. They also state that tuple items contribute their range extents and that combined items can exceed 100%.

The focused chart spec now has 14 passing tests, and lint passes. I left the pre-existing random keys and stale automatic `axisIncrement` for separate work. I also left out the warning, type-list cleanup, broad test expansion, and example-state change because they do not fix this regression.

Could you take another look?

adrianschmidt (approved):
# Consolidated PR Review (round 3)

_Automated 7-agent re-review of head `9c208145b`, after the two fixups that followed the 2026-09-03 13:25 review. Findings were measured against built worktrees at head and at `origin/main`, not inferred from the diff. Note: the Performance section was written by the coordinating session after the performance sub-agent was cut off by a usage limit; it is based on that agent's partial bundle measurements and a direct read of the code._

## PR Summary

The `fix` commit makes `stacked-bar`, `pie`, `doughnut`, and `ring` size items and print tooltip percentages as shares of `maxValue` or the summed item extents, unrounded, and stops `line`/`area` printing a percentage. The `feat` commit adds `displayItemPercentage` (default `true`), a docs example, and `chart.spec.tsx` (14 tests). The two newest fixups make whole types return `offset: 0` with an extent-based size, fall back to `1` for a non-positive inferred sum, and rewrite the `maxValue` prose. 5 files, +308/−38.

The round-2 blocker is fixed and verified by four agents independently: no `ring` arc stop exceeds the full turn for any tuple input, and the fix also repairs a pre-existing `main` defect (a tuple item plus a `maxValue` below the tuple end overran the circle on `main`). No dimension found a regression. The only remaining `[Medium]` is pre-existing and explicitly deferred.

- **Blockers:** None.
- **Non-blocking but worth addressing:** Top Recommendations 1–6. All are small; none should gate merge.

---

`limel-chart` is `@beta` and `README.md:137` allows breaking changes without a major bump, so `fix`/`feat` without `!` is correct. Every org consumer was enumerated: the only consumers on caret ranges that auto-adopt 40.x (`lime-crm-components` `^40.0.3`, `lime-forms`) render `bar`, which is untouched. `limepkg-graph-view`, `solution-dev-platform`, `limepkg-sales-demo-global`, and `solution-demo-dispatch` render whole types but are pinned at `^38`/`^39`.

Measured, main vs head:

| Input | main | head |
|---|---|---|
| ring `[10,40]` `[40,100]`, maxValue 90 | stops 44.44 / **111.11** | 33.33 / 66.67 |
| ring `[0,10]` `[20,30]` | 33.33 / 100 | 50 / 50 |
| pie `[10,40]` `[40,100]` | closes at 64.29 | closes at 100 |
| doughnut 25 / 75 | closes at 96.15 | closes at 100 |
| stacked-bar, single item 73 | 91.25 | 100 |
| stacked-bar, all zero | `NaN%` | `0.00%` |
| default stacked-bar host | reflects `axis-increment="8"` | reflects none |

- **[Low]** Ring tuple items moved from positional to extent-share placement (`chart.tsx:433-438`): on `main` an arc ran to the item's absolute end, at head every arc starts at the ring's zero and spans the extent. Correct for concentric gauges and the reason the truncation is gone, but a second consumer-visible shift for `ring` beyond the sizing change the description names.
- **[Low]** `chart.types.ts:16-22` is stale for whole types: the `ChartItem.value` docblock says a tuple is "a value range `[start, end]` (such as a Gantt-style bar)", implying the start positions the item. For the four whole types only `end − start` is used (`chart.tsx:640-646`).
- **[Low]** No downstream owner told. `limepkg-graph-view` renders all four whole types with `axisIncrement={1}` and no `maxValue`; its rings, pies, and stacked bars resize and its `axisIncrement` goes inert the day it bumps past `^38.43.0`. A message, not a code change.
- **Positives:** The blocker fix also closes ring arcs that overran the full turn on `main`, and every consumer that auto-adopts the release renders a type this PR does not touch.

## 2. Code Quality -- GOOD ✅

Spec: 14 passed in an isolated worktree. `eslint` on the four changed files: clean. Non-vacuity check: with `origin/main`'s math swapped in (only the new prop declaration retained), **10 of 14 tests fail** with the exact old numbers (31.25/93.75, 11.57, 96.15, 9.77/29.30), so the suite genuinely pins the fix. The still-open CodeRabbit thread at `chart.tsx:545` was written against `bde8d9ea7`; the current early return is exactly the extent-based stop it asked for.

- **[Medium]** Pre-existing, deferred with sound reasoning: `calculateRange` writes the auto-derived step onto the public `axisIncrement` prop at `chart.tsx:516-520` and never clears it, so a `bar` chart with items 25/75 keeps `axisIncrement = 8` after a switch to another type or an items rescale (reproduced at 939 axis lines). The PR does not worsen it; whole types now return before the write. Worth the follow-up issue both prior rounds recommended.
- **[Low]** `chart.tsx:548`: the `sum > 0` guard fixes the sign of the inferred whole, not its magnitude. Items `10`/`-9` give sizes 1000/−900 and a `1000.00%` tooltip where `main` gave 52.63/−47.37. Covered by the author's stated position that negative values have no part-of-whole semantics. Informational.
- **[Low]** `chart.spec.tsx:133-142` asserts only that all-negative sizes are below zero; actual values are −500/−1000/−1500 against `main`'s −50/−100/−150, so the test pins sign but not scale.
- **[Low]** The "scale, not a cap" overflow contract is stated in three places (`chart.tsx:166-167`, `chart.tsx:539`, `chart-max-value.tsx:9-11`) and no test pins it. Items 25/75 with `maxValue: 50` give 50/150 and `150.00%`; a future clamp would contradict the docs and pass the suite.
- **[Low]** `chart.tsx:536-542`: the `calculateWholeRange` docblock opens with a sentence restating the code below it; the other three sentences earn their place.
- **[Low]** `doughnut` is the only whole type with no test; it shares `pie`'s branch (measured 25/75, offsets 0/25), so peripheral.
- **Positives:** One `isPartOfWhole()` branch drives sizing, offsets, and tooltips from a single extent definition, and every rewritten docblock sentence survived execution, including the `nps`/`scatter` claims that needed the stylesheets to verify.

## 3. Architecture -- GOOD ✅

The prior blocker is **RESOLVED**. `_ring.scss:42-54` draws one conic gradient whose only stop is `offset + size`; offset is now always zero, so the stop equals the size. Offset zero is coherent for all four whole types, checked per stylesheet: `_stacked-bar.scss` reads only size (lines 35, 52); pie/doughnut overwrite the offset with the cumulative total at `chart.tsx:331-334` and close at exactly 100; the two other stylesheets reading `offset + size` are scoped to axis types. The four axis types, `nps`, and `scatter` reproduce `main` exactly across 61 measured cases. `@Watch('type')` is complete for what it owns (`range` and `secondaryAxisRange`) and is load-bearing: switching a ring to `scatter` without it would dereference a null `secondaryAxisRange`.

- **[Low]** `chart.tsx:433-438` duplicates a size formula the fall-through already computes: with `minValue` pinned to 0, `normalizedEnd − normalizedStart` reduces to `getValueSpan / totalRange × 100`, so only `offset` differs. Collapsing the branch to vary only the start value was verified byte-identical across all 61 probes and 14 tests. Two copies of the size expression is how the round-2 unit mismatch became possible.
- **[Low]** `chart.tsx:14`: `WHOLE_TYPES` is `ReadonlySet<string>`. Retyping to `ReadonlySet<Chart['type']>` builds clean, still type-checks `.has(this.type)`, and rejects a planted typo with a named error. Today a typo there silently drops a type into the axis branch.
- **[Low]** `_ring.scss:46-52` still adds `var(--limel-chart-item-offset) * 1%` to a term the component guarantees is zero. Drop it, or say why it stays.
- **[Low]** `chart.tsx:331`: pie/doughnut cumulative stacking is the one whole-type behavior not derived from `WHOLE_TYPES`; a fifth stacking type means editing an inline comparison inside `render`.
- **[Low]** `chart.tsx:556`: `AxisRange.increment` is filled with the whole, against a field documented at line 25 as the axis step. Verified unreachable for whole types; making the field optional would let the type say so.
- **[Low]** Pre-existing: with `maxValue` below the item sum, tooltip and rendered width disagree (25/75 with `maxValue: 50` print 50%/150% while flex-shrink squeezes segments back to proportional). `main` did the same at 44.64%/133.93%.
- **Positives:** One predicate now drives range selection, sizing, and tooltip gating, and the spec pins the exact invariant that broke (`offsets` equals `[0, 0]` at `chart.spec.tsx:174`).

## 4. Security -- SAFE ✅

Pure arithmetic and one boolean prop on a shadow-DOM component; no network, auth, storage, or secret handling. Verified in Chromium: item `text` containing markup renders as a literal text node with the percentage appended (no `innerHTML` anywhere in chart, tooltip, or portal); CSS custom properties go through `style.setProperty`, which rejects a top-level semicolon; `NaN`/`Infinity`/negative sizes compute to zero or clamped widths and print as ordinary text. The percentage exposes nothing not already in the visible value cell.

- **[Low]** Pre-existing and untouched: `item.color` is interpolated raw into `--limel-chart-item-color` (`chart.tsx:374`, `:404`) and substituted into `background` in `_stacked-bar.scss:17`; a bare `url(...)` there does trigger a fetch. The API contract is a CSS color and the trust boundary is the consumer's. Noted for completeness only.
- **Positives:** The new number reaching the DOM is arithmetic output, and its rendering sinks are byte-identical to the target branch.

## 5. Observability -- GOOD ✅

Prior `[Medium]` on stale `chart-max-value` prose is **RESOLVED**; all six executable claims in the rewrite hold, including the non-obvious `nps` one (`maxValue` does change `--limel-chart-item-size` on `nps`, but `_nps.scss:72-75` reads only `--limel-chart-item-value`, so nothing moves). The `console.warn` decline is sound: of 30 warn/error sites in non-example component code, none validates a numeric prop, and `limel-slider` accepts any `valuemin`/`valuemax`/`step` silently. Round 2's console-noise claim was backwards: on `main` every whole-type chart emitted one Stencil immutable-`@Prop` warning at load; at head they emit none, and only the runtime whole-to-axis switch gains one.

- **[Low]** `chart.tsx:543-558`: a degenerate whole now looks right and reports wrongly where `main` failed visibly. `maxValue={0}` gives sizes 2500/7500 and a `2500.00%` tooltip, but flex-shrink restores the correct proportions, so the only symptom is a tooltip off by 100×. A `NaN` item value sinks the sum to the fallback of 1 and a valid item of 10 reads `1000.00%`. Cuts both ways; documented at line 549 and defended by the author. Informational.
- **[Low]** `examples/chart-type-ring.tsx:44-47` binds an unconstrained `type="number"` input straight to `maxValue`, so a docs visitor typing 0 sees four-figure percentages with no signal. The sibling `chart-axis-increment` example constrains its input; a `min` is the cheap remedy.
- **[Low]** `examples/chart-max-value.tsx:28` uses `maxValue={512}` against items summing to 269, so the page never demonstrates the newly documented overflow-above-100% clause.
- **[Low]** `examples/chart-display-item-percentage.tsx:20` opens at `false` while the prop defaults to `true`. Declined twice on scope, still open.
- **Positives:** Dev console noise drops on the library's default type, the new `display-item-percentage` attribute reflects correctly in both directions, and the new example mounts with zero console errors in the smoke gate.

## 6. Performance -- GOOD ✅

No new render work per item: `calculateSizeAndOffset` reads the cached `this.range`, `getValueSpan` is O(1), and `calculateWholeRange` is one O(N) pass per invalidation, run from `componentWillLoad` and the watch handler, never from `render`. `@Watch('type')` only nulls two cached fields and recomputes; `type` is a `@Prop`, so a render was already scheduled. Examples are excluded from the dist build by `tsconfig.dist.json` (`**/examples/**`), so the new example adds nothing to consumers.

- **[Low]** Switching from a whole type to an axis type now runs the watch handler, which assigns `axisIncrement` at `chart.tsx:517` (pre-existing write) and re-enters `handleChange` once; the `!this.axisIncrement` guard stops it after one wasted range recompute. Not a loop, and the opposite direction improves: a chart loading on a whole type never writes the prop at all.
- **[Low]** The `limel-chart` dist entry chunk grows by roughly 1.8 KB unminified (48.7 KB to 50.5 KB, from the performance agent's partial measurement), from the new methods and docblocks. Informational.
- **Positives:** Whole types skip the axis-increment derivation entirely, so the library's default type does strictly less work at load than on `main`.

Catalogue fetched (index via `gh api`, rule pages FE-6, FE-7, FE-9 via the engineering-docs MCP). The only frontend category is FE-1 through FE-11 and none matches anything this PR introduces. The state-lifecycle rules were the group worth checking: `calculateRange` writes a reflected prop at `chart.tsx:516-520`, but FE-7 fires only when reachable from `render()` or a post-render hook, and the only callers are `componentWillLoad` and the watch handler, identical to `origin/main`. No findings.

- **Positives:** The property write the state-lifecycle rules probe for is verifiably unchanged from the target branch.

---

## Overall Verdicts

| Dimension | Verdict |
|---|---|
| Code Quality | **GOOD** |
| Architecture | **GOOD** |
| Security | **SAFE** |
| Observability | **GOOD** |
| Performance | **GOOD** |

## Top Recommendations

1. **[Medium from Code Quality]** Open a follow-up issue for the pre-existing stale auto-derived `axisIncrement` at `chart.tsx:516-520`: the derived step is written onto the public prop and never cleared, so an items rescale or type switch keeps the old step. Out of scope here, as the author said, but it has now been deferred three times without a tracking issue.
2. **[Low from Architecture]** Retype `WHOLE_TYPES` at `chart.tsx:14` to `ReadonlySet<Chart['type']>`. One token, verified to build clean and to catch a typo that today silently routes a type into the axis branch.
3. **[Low from Architecture]** Collapse the early return at `chart.tsx:433-438` so the whole branch only changes the start value and the size expression exists once; verified byte-identical output. Then either drop the always-zero offset term in `_ring.scss:46-52` or leave a one-line why.
4. **[Low from Code Quality]** Add one spec case pinning the documented overflow contract: items 25/75 with `maxValue: 50` yield sizes 50/150 and label `Photos (150.00%)`.
5. **[Low from Backward Compatibility]** Add a clause to the `ChartItem.value` docblock at `chart.types.ts:16-22` saying that for `stacked-bar`, `pie`, `doughnut`, and `ring` a tuple contributes its extent and is drawn from zero, and message the `limepkg-graph-view` owners that their rings, pies, and stacked bars resize and `axisIncrement` goes inert when they bump past `^38`.
6. **[Low from Observability]** Add `min` to the `maxValue` input in `examples/chart-type-ring.tsx:44-47`, and lower `maxValue` in `examples/chart-max-value.tsx:28` below the item sum (269) so the page demonstrates the overflow clause it documents.

jecrapo:
Thanks @adrianschmidt! Merging this now

I filed an issue for the `axisIncrement` bug identified in your review here: https://github.com/Lundalogik/lime-elements/issues/4275

PR #4276, opened 2026-09-04
feat(button-group): support disabling individual buttons

### LINKED ISSUE #4274: button-group: disable one option not the whole component
## New feature motivation

Currently `limel-button-group` can only disable the entire group via the
`disabled` prop. Consumers have no way to make a single option unavailable without disabling every button in the group.

Add an optional `disabled` field to the `Button` type so individual items
can be disabled independently, while the rest of the group stays
interactive.

Image

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/4274

## Summary by CodeRabbit

* **New Features**
  * Button groups now support disabling individual buttons while keeping other options active.
  * Disabled buttons are visibly styled, expose their disabled state for accessibility, and cannot be selected.
  * Individual disabled states work alongside the group-level disabled setting.
  * Disabled buttons remain inactive even when change events are triggered programmatically.

* **Documentation**
  * Added an example demonstrating a button group with an individually disabled option.

* **Tests**
  * Added end-to-end coverage for disabled button behavior and state changes.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/etc/lime-elements.api.md
+++ b/etc/lime-elements.api.md
@@ -58,6 +58,7 @@ export interface BreadcrumbsItem {
 // @public (undocumented)
 export interface Button {
     badge?: number | string;
+    disabled?: boolean;
     icon?: string;
     id: string;
     selected?: boolean;
--- a/src/components/button-group/button-group.e2e.tsx
+++ b/src/components/button-group/button-group.e2e.tsx
@@ -1,4 +1,5 @@
 import { render, h } from '@stencil/vitest';
+import { Button } from '../button/button.types';
 
 describe('limel-button-group', () => {
     describe('basic button group', () => {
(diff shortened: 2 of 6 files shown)


Kiarokh:
## Why the disabled button keeps the native `disabled` attribute

A review of this branch raised an accessibility gap worth recording, because the fix we chose is deliberately partial.

**The finding.** A per-item disabled button signals its state only through the native `disabled` attribute on the underlying `input[type="radio"]`. Inside a native radio group, arrow-key navigation *skips* disabled radios, so a keyboard or screen-reader user in focus mode never lands on the disabled option and is never told that it exists but is unavailable — from their point of view it simply isn't there. The ARIA Authoring Practices pattern for a disabled option in a composite widget is the opposite: keep the option focusable, mark it `aria-disabled`, and reject activation in JavaScript, precisely so it stays discoverable. This library already does it that way in `limel-list-item`, `limel-chip` and `limel-menu-list`.

**What we changed.** We added `aria-disabled` alongside the native attribute (matching `select.template.tsx`, which sets both on the same element), and added a guard in `onChange` that rejects a change for a disabled button.

**What we deliberately did not change.** We kept the native `disabled` attribute rather than switching to the APG pattern, for four reasons:

1. **The platform does the hard part for free.** The native attribute suppresses label activation, `.click()`, focus, tab order and arrow-key selection. Hand-rolling all of that in JavaScript is more code and more ways to get it wrong, and the security-relevant property — that a disabled option cannot produce a `change` event — becomes something we maintain rather than something the browser guarantees.
2. **All of the disabled styling is keyed on `:has(:disabled)`.** The dimming, the `not-allowed` cursor, and the suppression of the hover lift and pressed animation all match on the disabled input. Removing the native attribute would make every one of those selectors stop matching, silently reverting the styling work in this PR.
3. **`input.disabled` is an observable part of the rendered DOM** and is asserted in the tests here. Consumers may reasonably key off it too, so dropping it is a wider change than it looks.
4. **The state is not actually invisible.** In browse mode (virtual cursor) a screen reader still announces the button as unavailable. What is lost is focus-mode navigation specifically — a real gap, but a narrower one than "no signal at all".

**So, honestly:** `aria-disabled` here is belt-and-braces. It is redundant next to a native `disabled` attribute, which already makes assistive technology report the control as unavailable, and it does **not** restore arrow-key discoverability. The `onChange` guard is likewise defence-in-depth rather than a fix for anything reachable in normal use — it matters because Stencil diffs vnode against vnode and never against the live DOM, so if the `disabled` attribute is ever removed from the DOM it is never re-asserted on later renders, and a hand-dispatched `change` event is not suppressed on a disabled control either. Both cases are now covered by tests.

If focus-mode discoverability of disabled options becomes a requirement, the change is well understood but not small: drop the native attribute, keep the radio focusable, rely on the JS guard for activation, and re-key the stylesheet off a class or attribute instead of `:has(:disabled)`. That deserves its own issue, and ideally a decision across the components that model per-item disabling, rather than being settled for `limel-button-group` alone.

Kiarokh (approved):

PR #4280, opened 2026-09-07
Add an example for the info tile's link property

Kiarokh:
@coderabbitai summary

## Why

The `link` property of `limel-info-tile` had no example of its own. Nothing in the docs showed what supplying a link does, what leaving it out does, or how to handle the common case where a single page application wants to route a click itself instead of letting the browser navigate.

The new example has three tiles:

1. **An external link.** `target="_blank"`, a `title` that becomes the tooltip, and `rel="noopener noreferrer"` filled in automatically by `getRel`.
2. **An in-app link with an intercepted click.** The `link` points at the real destination so modifier-clicking still opens it in a new tab, while a plain click is `preventDefault`ed for the application's own router — the same pattern as `limel-example-shortcut-with-click-handler`.
3. **No link at all.** Hovering this one next to the first two is the whole point of the example: it stays flat and still, while they elevate and tilt.

## Note on the split

This PR originally carried three separate concerns. Two have been split out so they can be reviewed and merged on their own:

- **#4285** — four `perf` commits fixing the `mousemove` listener leak (#4282). No behavioural change. Can merge whenever.
- **#4286** — `fix(info-tile): drop the 3d hover effect when the tile is not a link`. The look-and-feel and clickability change.

Their commits are deliberately **still on this branch**, unchanged and at the same SHAs, so this PR stands up on its own in the meantime. As they merge, this branch gets rebased and those commits drop out by patch-id, leaving only `docs(info-tile): add an example for the link property`.

**#4286 needs to merge before this one.** The example's prose describes the post-fix behaviour — it tells the reader the third tile stays "flat and completely still" and "is not reachable with the keyboard". Landing the docs first would mean shipping text that describes behaviour `main` does not have yet.

## Things worth a reviewer's attention on the example itself

**The tiles set text and icon colours, not just backgrounds.** They use soft `-lighter` backgrounds from the palette, and each also sets `--info-tile-text-color` and `--info-tile-icon-color` to the matching `-darker` swatch. Setting the background alone is not enough: the default `--info-tile-text-color` resolves to `--contrast-1100`, which flips to a *light* grey in dark mode and lands at roughly **1.1:1** over a pastel. Same pairing as `limel-example-callout-styles`. Verified by reading the computed backgrounds off the rendered anchors in both light and dark mode.

**An earlier revision of this example fetched live weather from Open-Meteo**, so the tile's `loading` state was real rather than simulated. That was dropped in favour of a static reading, to keep the docs site free of a third-party runtime dependency. Mentioning it only so the idea is on record rather than lost.

## Verified

- `npm run docs:rebuild` clean; `npm run lint` clean.
- Example accessibility suite: all `info-tile` examples pass, no new axe violations, no baseline entries added.
- Example runtime suite green.
- Screenshots checked in light and dark mode.

## Review:
- [x] Commits have the correct *type* for the changes made — every change to an `examples/*.tsx` file is `docs`, including a new one
- [x] Commits with *breaking changes* are marked as such (none in the documentation commit)

### Browsers tested:
Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/card/card.tsx
+++ b/src/components/card/card.tsx
@@ -127,17 +127,19 @@ export class Card {
 
     private handleMouseEnter: () => void;
     private handleMouseLeave: () => void;
+    private cleanup3dHoverEffect: () => void;
     private markdownResizeObserver?: ResizeObserver;
 
     public componentWillLoad() {
-        const { handleMouseEnter, handleMouseLeave } = getMouseEventHandlers(
-            this.host
-        );
+        const { handleMouseEnter, handleMouseLeave, cleanup } =
+            getMouseEventHandlers(this.host);
         this.handleMouseEnter = handleMouseEnter;
         this.handleMouseLeave = handleMouseLeave;
+        this.cleanup3dHoverEffect = cleanup;
     }
 
     public disconnectedCallback() {
+        this.cleanup3dHoverEffect();
         this.markdownResizeObserver?.disconnect();
         this.markdownElement?.removeEventListener(
             'scroll',
--- a/src/components/info-tile/examples/info-tile-link.scss
+++ b/src/components/info-tile/examples/info-tile-link.scss
@@ -1,0 +1,30 @@
+:host(limel-example-info-tile-link) {
+    display: grid;
+    grid-template-columns: repeat(auto-fit, minmax(10rem, 1fr));
+    gap: 1rem;
+}
+
+limel-info-tile {
+    height: 8rem;
+
+    // Each tile also sets its text and icon color, not just its background.
+    // The default text color resolves to `--contrast-1100`, which turns
+    // light in dark mode and would sit at roughly 1.1:1 over a pastel.
+    &.forecast {
+        --info-tile-background-color: rgb(var(--color-glaucous-lighter));
+        --info-tile-text-color: rgb(var(--color-glaucous-darker));
+        --info-tile-icon-color: rgb(var(--color-glaucous-darker));
+    }
+
+    &.deals {
+        --info-tile-background-color: rgb(var(--color-green-lighter));
+        --info-tile-text-color: rgb(var(--color-green-darker));
+        --info-tile-icon-color: rgb(var(--color-green-darker));
+    }
+
+    &.office {
+        --info-tile-background-color: rgb(var(--color-violet-lighter));
+        --info-tile-text-color: rgb(var(--color-violet-darker));
+        --info-tile-icon-color: rgb(var(--color-violet-darker));
+    }
+}
--- a/src/components/info-tile/examples/info-tile-link.tsx
+++ b/src/components/info-tile/examples/info-tile-link.tsx
@@ -1,0 +1,101 @@
+import { Component, h, Host } from '@stencil/core';
+import { Link } from '@limetech/lime-elements';
+
+/**
+ * Turning the tile into a link
+ *
+ * An info tile usually summarizes something the user will want to look
+ * into. Supply a `link` and the tile's anchor gets an `href`: it can then
+ * be opened in a new tab from the context menu, and assistive technologies
+ * announce it as a link rather than as a piece of text.
+ *
+ * Supplying a `link` is also what gives the tile its interactive styling:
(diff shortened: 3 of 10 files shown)


Kiarokh:
@coderabbitai full review please

Kiarokh:
Fixed in 1387b8470 (pushed as a fixup, per the review-feedback convention).

`fetch` has no timeout of its own, so a request that stalled without failing would never settle — and because `loadWeather` clears the tile's `loading` state in a `finally`, that state would never be cleared either, leaving the tile spinning for as long as the page stayed open. `fetchWeather` now passes `signal: AbortSignal.timeout(TIMEOUT)` with an 8s bound, which rejects the request and lands it in the same failure path as any other error, so the existing `catch`/`finally` handles it unchanged. The signal covers the body read as well, so a stalled `response.json()` is bounded too.

Verified in a browser against a request held open indefinitely: the tile settles 8237ms after the request is issued, with `loading` cleared, the label at "Forecast unavailable" and the `no_connection` icon — and the live path still resolves normally with the signal attached.

Three related things I fixed in the same pass while reviewing the file:

- **A malformed 200 would have rendered `NaN °C`.** `Math.round(current.temperature_2m)` yields `NaN` rather than throwing when the field is missing, so the tile would have displayed it. `fetchWeather` now rejects with a `TypeError` unless the temperature is finite.
- **`@public` on the `Weather` interface.** That is an API-Extractor release tag claiming membership of the library's supported public API, which an example helper is not — and no other helper under any `examples/` directory uses it. Removed.
- **`encodeURIComponent` on the city in the forecast URL.** The helper documents `CITY` as the constant you change, and a two-word city would have produced a malformed query string.

`npm run lint`, the example accessibility suite (8/8 `info-tile`, no new axe violations) and the example runtime suite (462/462) are all green on the fixup.

paulinewahle:
The unlinked tile would do better without the hover effect. Even with the cursor change, it still invokes a strong clue of click-ability. Really nice example though!

Kiarokh:
> The unlinked tile would do better without the hover effect. Even with the cursor change, it still invokes a strong clue of click-ability. Really nice example though!

agreed. This was mostly for testing this visual style. But we could actually make the non-interactive info-tile look "non-interactive" too. Want me to do it in the same PR @paulinewahle ?

Kiarokh:
The four `perf` commits here are now also open on their own as #4285, so they can be reviewed and merged separately.

They are deliberately left on this branch too, unchanged and at the same SHAs, so this PR still stands up on its own until #4285 lands. Once it does, this branch gets rebased onto `main` and the four drop out by patch-id, leaving just:

- `fix(info-tile): drop the 3d hover effect when the tile is not a link`
- `docs(info-tile): add an example for the link property`

Reviewing #4285 first is the easier order — it has no behavioural change in it at all, and this PR's `fix` commit is what actually changes what consumers see.

Kiarokh:
Updating the earlier note: this PR's work is now split across three, so each piece can be reviewed and merged on its own.

- **#4285** — the four `perf` commits (the `mousemove` listener leak, #4282). No behavioural change.
- **#4286** — `fix(info-tile): drop the 3d hover effect when the tile is not a link`. The look-and-feel and clickability change.
- **this PR** — the documentation example, once the two above land.

All six commits are deliberately still here, unchanged and at the same SHAs, so this PR stands on its own in the meantime. As #4285 and #4286 merge, this branch gets rebased and their commits drop out by patch-id, leaving just `docs(info-tile): add an example for the link property`.

**One ordering constraint worth knowing:** #4286 should merge before the documentation example does. The example's prose describes the post-fix behaviour — it tells the reader to hover the third tile and see it "stay flat and completely still", and that it "is not reachable with the keyboard". If the docs commit landed on its own first, that text would be describing behaviour `main` did not have yet. #4285 has no such constraint and can go in whenever.

PR #4285, opened 2026-09-08
Stop the 3D tilt hover effect leaking a mousemove listener

### LINKED ISSUE #4282: 3d-tilt-hover-effect: document mousemove listener leaks when the element is removed while hovered
## Current behavior

`getMouseEventHandlers` in the 3D tilt hover effect adds a `document`-level `mousemove` listener on `mouseenter` and removes it only on `mouseleave`. If the element is removed from the DOM while it is being hovered, `mouseleave` never fires and the listener is never removed.

The surviving listener closes over the `DOMRect` captured at `mouseenter` and over the now-detached element, and keeps running `tiltFollowingTheCursor` on every mouse move for the rest of the page's life — computing a `sqrt` and a `log` and writing two CSS custom properties to a node that is no longer in the document, while keeping that node and its bounds reachable. Each occurrence leaks one listener plus the element it retains.

All three consumers are affected: `limel-info-tile` and `limel-shortcut` have no `disconnectedCallback` at all, and `limel-card`'s (`src/components/card/card.tsx`, around line 140) only disconnects its markdown `ResizeObserver` and scroll listener.

1. Render a `limel-card` with `show3dEffect`, a `limel-shortcut`, or a `limel-info-tile`.
2. Hover the element, and while the pointer is still over it, remove it from the DOM. In a single page application, a click that routes away from the view holding the element does exactly this — which is the common case, since these components are usually clickable links.
3. `document` still carries the `mousemove` listener, and moving the pointer keeps invoking it against the detached element.

Relevant source: `getMouseEventHandlers` in `src/util/3d-tilt-hover-effect.ts` (around lines 149-166) — it returns only `handleMouseEnter` and `handleMouseLeave`, so a consumer has no handle to clean up with even if it wanted to.

## Expected behavior

The listener should be removed when the element is disconnected. Have `getMouseEventHandlers` also return a `cleanup()` that unconditionally calls `document.removeEventListener('mousemove', tiltCallback)`, then call it from a `disconnectedCallback` in each of the three consuming components.

## Environment

- **lime-elements** version: 40.1.0
- Framework used: any — most visible in a single page application, where views mount and unmount repeatedly
- Logs: none, the leak is silent

<sub>Found while reviewing #4280, which only touches this by adding three more tiles to the info-tile docs page.</sub>

Kiarokh:
## Summary by CodeRabbit

- **Bug Fixes**
  - 3D hover effects now stop correctly when cards, information tiles, or shortcuts are removed.
  - Prevented lingering hover behavior after interactive elements leave the page.
  - Improved reliability when repeatedly entering and leaving interactive elements.

- **Documentation**
  - Added guidance for cleaning up 3D hover effects.

- **Tests**
  - Added coverage for hover movement, glow updates, repeated interactions, mouse-leave behavior, and cleanup scenarios.

fix: https://github.com/Lundalogik/lime-elements/issues/4282

## Why

Closes #4282.

`getMouseEventHandlers` in the 3D tilt hover effect added a `document` level `mousemove` listener on `mouseenter` and removed it only on `mouseleave`. An element removed from the DOM while the pointer is still over it never gets a `mouseleave`, so the listener survived it — and `getMouseEventHandlers` returned no handle to that listener, so a consumer had no way to remove it even if it wanted to.

The surviving listener keeps running `tiltFollowingTheCursor` on every mouse move for the rest of the page's life: a `sqrt`, a `log` and two custom property writes against a node that is no longer in the document, while keeping that node and its captured `DOMRect` reachable. Each occurrence leaks one listener plus the element it retains.

In a single page application this is the ordinary case rather than an edge case, because these components are usually links: a click routes away from the view holding them while they are still hovered. Every such navigation leaked another listener and another retained element.

## The commits

| | |
|---|---|
| `fix: stop the 3d tilt hover effect stranding a mousemove listener` | The shared util. No scope, because it is not specific to any one component and has no tag of its own. |
| `fix(card): …` | Adds the `cleanup()` call to the `disconnectedCallback` the card already had, plus a regression test in the new `card.e2e.tsx`. |
| `fix(shortcut): …` | The component had no `disconnectedCallback` at all. Regression test added to `shortcut.e2e.tsx`. |
| `fix(info-tile): …` | Same, and the last of the three consumers still leaking, so this one carries `Closes #4282`. Regression test added to `info-tile.e2e.tsx`. |

The first commit deliberately lands a `cleanup()` that nothing calls yet — it is preparatory, so that each consumer's wiring is its own reviewable commit with its own correct scope and its own changelog line. It is not inert, though: it also stops `handleMouseEnter` stranding a listener it is about to replace, which on its own closes a second leak. Two `mouseenter` events with no `mouseleave` between them used to leave the first listener attached with nothing referencing it, so not even a later `mouseleave` could remove it.

## Things worth a reviewer's attention

**The new spec is a real regression test, not a test of the code as written.** I confirmed both of its key cases **fail** against the unfixed util and pass against the fixed one, by neutering `cleanup` to a no-op and removing the `handleMouseEnter` guard, then running the spec:

```
× when the element is removed while still hovered > stops tilting once cleaned up
× when `mouseenter` fires twice without a `mouseleave` > leaves no callback behind that `cleanup` cannot reach
```

It asserts behaviour rather than spying on `removeEventListener`: it dispatches a real `mousemove` and checks that the element's `--limel-3d-hover-effect-rotate3d` is no longer being written.

**`cleanup()` is the same teardown as `handleMouseLeave`** — it drops the listener *and* clears the two custom properties. It originally left the element alone, on the reasoning that there is no point styling a node being thrown away. Review pointed out that this holds for a removal but not for a DOM *move*, which disconnects the component just the same: the element stays in the document with a stale `rotate3d` still set, and `src/style/mixins.scss` keeps applying it under `&:hover`, so a moved tile sat frozen at its last angle. `removeProperty` on a detached node is free, so one teardown now covers both cases.

**No public API change.** `getMouseEventHandlers` is internal — absent from `src/index.ts` and from `etc/lime-elements.api.md` — so adding `cleanup` to its return type is not a `feat`, and `etc/lime-elements.api.md` needs no update.

**Type choice: `fix`, not `perf`.** These started as `perf` and were retyped during review. `commits-and-prs.md` defines `perf` as a change that *only* improves performance without changing behaviour, and these do change behaviour: the `handleMouseEnter` guard changes what a second `mouseenter` does, and `cleanup()` now resets a moved element. The same doc's carve-out points the same way — "if the performance was previously so bad it could be considered a bug, and that's been fixed, `fix` is a good fit" — and an unbounded listener retaining a detached node is a bug, filed on the bug template as #4282. Both types release as a patch bump, so the version is unaffected.

## Verified

- `npm run lint` clean and the spec suite green (**1203**) at **each of the four commits individually**, not just the tip — checked by walking each SHA, so the branch bisects cleanly.
- The e2e count climbs **16 → 17 → 18 → 19** across the four commits, i.e. each component's regression test lands in the same commit as the fix it guards.
- All four commit messages pass `commitlint`.
- The review fixups were autosquashed content-neutrally: the resulting tree hash is byte-identical to the pre-squash tip, and each of the four intermediate trees matches the preview that was verified commit-by-commit.
- Full e2e suite: 1082 passed, 8 skipped. The one failure is in `src/examples-smoke.e2e.ts`, which is **flaky and unrelated** — it failed on `limel-example-chip-readonly-border` in one run at the tip, on `limel-example-dock-notification` at the base commit `81a8678c3` (before this PR), and passed twice in a row at the tip. Nothing here touches chips or dock.
- #4280 carries copies of these four commits. **They will no longer drop out by patch-id** — the review fixups changed all four patches — so drop them explicitly when rebasing that branch.

## Review round 1

A [6-agent review](https://github.com/Lundalogik/lime-elements/pull/4285#issuecomment-5586995125) found no regression against `main` and no blockers. Four of its six recommendations were applied as fixups, since autosquashed into the four commits:

1. `cleanup3dHoverEffect?.()` with an optional field in all three components, and card's call moved *after* its pre-existing teardown so a throw cannot skip it. Stencil defers `componentWillLoad` to a microtask, so `disconnectedCallback` can run before the field is assigned.
2. `cleanup()` now returns the element to rest (see above).
3. A browser regression test per component, each in that component's own commit.
4. `perf` → `fix` (see above).

Two were left open on purpose, both pre-existing and both better as their own PR: a `@Watch('show3dEffect')` teardown for the case where the prop flips off mid-hover, and the larger question of moving the `mousemove` listener from `document` onto the element, which would make the leak structurally impossible but changes the effect in two edge cases. `limel-popover` has the same leak shape today and deserves its own issue.

Both CodeRabbit comments were valid and are fixed — the spec no longer leaves its own listener attached, and the doc snippet now type-checks. I did not add CodeRabbit's `!` to the `querySelector` call: this repo sets no `strict`/`strictNullChecks`, so `HTMLElement | null` assigns fine and the assertion would only teach a habit the codebase does not need. The `Element` → `HTMLElement` narrowing was the real error.

One correction to my own earlier claim: the util's doc comment is **not** published to the docs site. `src/index.ts` re-exports only `./components` and `./interface`, so the kompendium target never reaches `src/util/3d-tilt-hover-effect.ts`. The warning still reaches a contributor on editor hover, which is the audience that matters, but `src/style-utilities.md` would be the discoverable home for it.

## Review:
- [x] Commits with *breaking changes* are marked as such (none — no API change, and the behaviour changes are the leak fix itself)

### Browsers tested:
Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [x] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS

### DIFF
--- a/src/components/card/card.e2e.tsx
+++ b/src/components/card/card.e2e.tsx
@@ -1,0 +1,33 @@
+import { render, h } from '@stencil/vitest';
+
+const TILT = '--limel-3d-hover-effect-rotate3d';
+
+const moveMouse = () => {
+    document.dispatchEvent(
+        new MouseEvent('mousemove', { clientX: 20, clientY: 30 })
+    );
+};
+
+describe('limel-card', () => {
+    describe('when the card is removed while it is still hovered', () => {
+        // `mouseleave` never fires in this case, so the 3d tilt effect's
+        // `document` level `mousemove` listener has to be removed by the
+        // card's `disconnectedCallback` instead.
+        it('stops tilting', async () => {
+            const { root, waitForChanges, unmount } = await render(
+                <limel-card></limel-card>
+            );
+            await waitForChanges();
+
+            root.dispatchEvent(new MouseEvent('mouseenter'));
+            moveMouse();
+
+            expect(root.style.getPropertyValue(TILT)).not.toBe('');
+
+            unmount();
+            moveMouse();
+
+            expect(root.style.getPropertyValue(TILT)).toBe('');
+        });
+    });
+});
--- a/src/components/card/card.tsx
+++ b/src/components/card/card.tsx
@@ -127,14 +127,15 @@ export class Card {
 
     private handleMouseEnter: () => void;
     private handleMouseLeave: () => void;
+    private cleanup3dHoverEffect?: () => void;
     private markdownResizeObserver?: ResizeObserver;
 
     public componentWillLoad() {
-        const { handleMouseEnter, handleMouseLeave } = getMouseEventHandlers(
-            this.host
-        );
+        const { handleMouseEnter, handleMouseLeave, cleanup } =
+            getMouseEventHandlers(this.host);
         this.handleMouseEnter = handleMouseEnter;
         this.handleMouseLeave = handleMouseLeave;
+        this.cleanup3dHoverEffect = cleanup;
     }
 
     public disconnectedCallback() {
@@ -143,6 +144,7 @@ export class Card {
             'scroll',
             this.checkIfScrollable
         );
+        this.cleanup3dHoverEffect?.();
     }
 
     public componentDidLoad() {
--- a/src/components/info-tile/info-tile.e2e.tsx
+++ b/src/components/info-tile/info-tile.e2e.tsx
@@ -1,5 +1,13 @@
 import { render, h } from '@stencil/vitest';
 
+const TILT = '--limel-3d-hover-effect-rotate3d';
+
+const moveMouse = () => {
+    document.dispatchEvent(
+        new MouseEvent('mousemove', { clientX: 20, clientY: 30 })
+    );
+};
+
 describe('limel-info-tile', () => {
     describe('smoke test', () => {
         it('displays the correct value', async () => {
@@ -24,6 +32,28 @@ describe('limel-info-tile', () => {
             const label = root.shadowRoot!.querySelector('.label');
             expect(label).not.toBeNull();
             expect(label.textContent).toEqual('Test label');
+        });
+    });
+
+    describe('when the tile is removed while it is still hovered', () => {
+        // `mouseleave` never fires in this case, so the 3d tilt effect's
+        // `document` level `mousemove` listener has to be removed by the
+        // tile's `disconnectedCallback` instead.
+        it('stops tilting', async () => {
+            const { root, waitForChanges, unmount } = await render(
+                <limel-info-tile value="Test value"></limel-info-tile>
+            );
+            await waitForChanges();
+
+            root.dispatchEvent(new MouseEvent('mouseenter'));
+            moveMouse();
+
+            expect(root.style.getPropertyValue(TILT)).not.toBe('');
+
+            unmount();
+            moveMouse();
+
+            expect(root.style.getPropertyValue(TILT)).toBe('');
         });
     });
 });
--- a/src/components/info-tile/info-tile.tsx
+++ b/src/components/info-tile/info-tile.tsx
@@ -123,14 +123,19 @@ export class InfoTile {
 
     private handleMouseEnter: () => void;
     private handleMouseLeave: () => void;
+    private cleanup3dHoverEffect?: () => void;
 
     public componentWillLoad() {
-        const { handleMouseEnter, handleMouseLeave } = getMouseEventHandlers(
-            this.host
-        );
+        const { handleMouseEnter, handleMouseLeave, cleanup } =
+            getMouseEventHandlers(this.host);
         this.handleMouseEnter = handleMouseEnter;
         this.handleMouseLeave = handleMouseLeave;
+        this.cleanup3dHoverEffect = cleanup;
         this.updateHasPrimarySlotContent();
+    }
+
+    public disconnectedCallback() {
+        this.cleanup3dHoverEffect?.();
     }
 
     public render() {
--- a/src/components/shortcut/shortcut.e2e.tsx
+++ b/src/components/shortcut/shortcut.e2e.tsx
@@ -1,5 +1,13 @@
 import { render, h } from '@stencil/vitest';
 
+const TILT = '--limel-3d-hover-effect-rotate3d';
+
+const moveMouse = () => {
+    document.dispatchEvent(
+        new MouseEvent('mousemove', { clientX: 20, clientY: 30 })
+    );
+};
+
 describe('limel-shortcut', () => {
     describe('with a label', () => {
         let root: HTMLElement;
@@ -186,6 +194,28 @@ describe('limel-shortcut', () => {
                 const badge = root.shadowRoot!.querySelector('limel-badge');
                 expect(badge.getAttribute('label')).toEqual('3');
             });
+        });
+    });
+
+    describe('when the shortcut is removed while it is still hovered', () => {
+        // `mouseleave` never fires in this case, so the 3d tilt effect's
+        // `document` level `mousemove` listener has to be removed by the
+        // shortcut's `disconnectedCallback` instead.
+        it('stops tilting', async () => {
+            const { root, waitForChanges, unmount } = await render(
+                <limel-shortcut label="iSpiffy"></limel-shortcut>
+            );
+            await waitForChanges();
+
+            root.dispatchEvent(new MouseEvent('mouseenter'));
+            moveMouse();
+
+            expect(root.style.getPropertyValue(TILT)).not.toBe('');
+
+            unmount();
+            moveMouse();
+
+            expect(root.style.getPropertyValue(TILT)).toBe('');
         });
     });
 });
--- a/src/components/shortcut/shortcut.tsx
+++ b/src/components/shortcut/shortcut.tsx
@@ -61,13 +61,18 @@ export class Shortcut {
 
     private handleMouseEnter: () => void;
     private handleMouseLeave: () => void;
+    private cleanup3dHoverEffect?: () => void;
 
     public componentWillLoad() {
-        const { handleMouseEnter, handleMouseLeave } = getMouseEventHandlers(
-            this.host
-        );
+        const { handleMouseEnter, handleMouseLeave, cleanup } =
+            getMouseEventHandlers(this.host);
         this.handleMouseEnter = handleMouseEnter;
         this.handleMouseLeave = handleMouseLeave;
+        this.cleanup3dHoverEffect = cleanup;
+    }
+
+    public disconnectedCallback() {
+        this.cleanup3dHoverEffect?.();
     }
 
     public render() {
--- a/src/util/3d-tilt-hover-effect.spec.ts
+++ b/src/util/3d-tilt-hover-effect.spec.ts
@@ -1,0 +1,94 @@
+import { getMouseEventHandlers } from './3d-tilt-hover-effect';
+
+const TILT = '--limel-3d-hover-effect-rotate3d';
+const GLOW = '--limel-3d-hover-effect-glow-position';
+
+const moveMouse = () => {
+    document.dispatchEvent(
+        new MouseEvent('mousemove', { clientX: 20, clientY: 30 })
+    );
+};
+
+describe('getMouseEventHandlers', () => {
+    let element: HTMLElement;
+
+    beforeEach(() => {
+        element = document.createElement('div');
+        document.body.append(element);
+    });
+
+    afterEach(() => {
+        element.remove();
+    });
+
+    it('tilts the element while it is hovered', () => {
+        const { handleMouseEnter, cleanup } = getMouseEventHandlers(element);
+
+        handleMouseEnter();
+        moveMouse();
+
+        expect(element.style.getPropertyValue(TILT)).not.toBe('');
+        expect(element.style.getPropertyValue(GLOW)).not.toBe('');
+
+        cleanup();
+    });
+
+    it('stops tilting once the pointer leaves', () => {
+        const { handleMouseEnter, handleMouseLeave } =
+            getMouseEventHandlers(element);
+
+        handleMouseEnter();
+        handleMouseLeave();
+        moveMouse();
+
+        expect(element.style.getPropertyValue(TILT)).toBe('');
+        expect(element.style.getPropertyValue(GLOW)).toBe('');
+    });
+
+    it('returns the element to rest when cleaned up', () => {
+        const { handleMouseEnter, cleanup } = getMouseEventHandlers(element);
+
+        handleMouseEnter();
+        moveMouse();
+        cleanup();
+
+        expect(element.style.getPropertyValue(TILT)).toBe('');
+        expect(element.style.getPropertyValue(GLOW)).toBe('');
+    });
+
+    describe('when the element is removed while still hovered', () => {
+        // `mouseleave` never fires in this case, so without `cleanup` the
+        // listener stays on `document` for the rest of the page's life.
+        it('stops tilting once cleaned up', () => {
+            const { handleMouseEnter, cleanup } =
+                getMouseEventHandlers(element);
+
+            handleMouseEnter();
+            element.remove();
+            cleanup();
+            moveMouse();
+
+            expect(element.style.getPropertyValue(TILT)).toBe('');
+        });
+    });
+
+    describe('when `mouseenter` fires twice without a `mouseleave`', () => {
+        it('leaves no callback behind that `cleanup` cannot reach', () => {
+            const { handleMouseEnter, cleanup } =
+                getMouseEventHandlers(element);
+
+            handleMouseEnter();
+            handleMouseEnter();
+            cleanup();
+            moveMouse();
+
+            expect(element.style.getPropertyValue(TILT)).toBe('');
+        });
+    });
+
+    it('does not throw when cleaned up before ever being hovered', () => {
+        const { cleanup } = getMouseEventHandlers(element);
+
+        expect(() => cleanup()).not.toThrow();
+    });
+});
--- a/src/util/3d-tilt-hover-effect.ts
+++ b/src/util/3d-tilt-hover-effect.ts
@@ -50,7 +50,8 @@
  * 3. **Initialize in your component**:
  *
  * Use `getMouseEventHandlers()` to attach the required mouse event listeners
- * to the "interactive element" (`the-3d-element`). For example:
+ * to the "interactive element" (`the-3d-element`), and call the `cleanup()` it
+ * returns when the component is disconnected. For example:
  *
  * ```tsx
  * @Element()
@@ -58,15 +59,29 @@
  *
  * private handleMouseEnter: () => void;
  * private handleMouseLeave: () => void;
+ * private cleanup3dHoverEffect?: () => void;
  *
  * public componentWillLoad() {
- *     const { handleMouseEnter, handleMouseLeave } = getMouseEventHandlers(
- *         this.host.querySelector('.the-3d-element'),
- *     );
+ *     const { handleMouseEnter, handleMouseLeave, cleanup } =
+ *         getMouseEventHandlers(
+ *             this.host.querySelector<HTMLElement>('.the-3d-element'),
+ *         );
  *     this.handleMouseEnter = handleMouseEnter;
  *     this.handleMouseLeave = handleMouseLeave;
+ *     this.cleanup3dHoverEffect = cleanup;
+ * }
+ *
+ * public disconnectedCallback() {
+ *     this.cleanup3dHoverEffect?.();
  * }
  * ```
+ *
+ * Call it optionally: `disconnectedCallback` can run before
+ * `componentWillLoad` has assigned the field.
+ *
+ * Skipping the `disconnectedCallback` leaks a `document` level `mousemove`
+ * listener every time the element is removed while it is being hovered, and
+ * leaves a moved element stuck at the angle it was last tilted to.
  *
  * 4. **Attach event handlers in your render method**:
  *
@@ -148,20 +163,47 @@ export const tiltFollowingTheCursor =
 export const getMouseEventHandlers = (element: HTMLElement) => {
     let tiltCallback: (e: MouseEvent) => void;
 
-    const handleMouseEnter = () => {
-        const bounds = element.getBoundingClientRect();
-        tiltCallback = tiltFollowingTheCursor(bounds, element);
-        document.addEventListener('mousemove', tiltCallback);
-    };
-
-    const handleMouseLeave = () => {
+    /**
+     * Stops following the cursor and returns the element to rest.
+     *
+     * Safe to call when tracking never started, and safe to call on an
+     * element that has already been detached.
+     */
+    const stopTracking = () => {
         document.removeEventListener('mousemove', tiltCallback);
         element.style.removeProperty('--limel-3d-hover-effect-rotate3d');
         element.style.removeProperty('--limel-3d-hover-effect-glow-position');
     };
 
+    const handleMouseEnter = () => {
+        // A second `mouseenter` without an intervening `mouseleave` would
+        // otherwise strand the previous callback on `document` with nothing
+        // left holding a reference to remove it with.
+        stopTracking();
+
+        const bounds = element.getBoundingClientRect();
+        tiltCallback = tiltFollowingTheCursor(bounds, element);
+        document.addEventListener('mousemove', tiltCallback);
+    };
+
     return {
         handleMouseEnter: handleMouseEnter,
-        handleMouseLeave: handleMouseLeave,
+        handleMouseLeave: stopTracking,
+
+        /**
+         * The same teardown as `handleMouseLeave`, for the case where no
+         * `mouseleave` is coming.
+         *
+         * `mouseleave` never fires for an element that is removed from the
+         * DOM while the pointer is still over it, which in a single page
+         * application is the ordinary case: a click routes away from the
+         * view holding the element. It does not fire when the element is
+         * merely moved within the DOM either, and a move disconnects the
+         * component just like a removal does. Call this from
+         * `disconnectedCallback` so the listener does not outlive the
+         * element it tilts, and so a moved element does not stay stuck at
+         * the angle it was last tilted to.
+         */
+        cleanup: stopTracking,
     };
 };


Kiarokh:
> 🤖 AI-generated review from 6 parallel agents, re-issued after acting on it. Treat as input, not a verdict — agents can be wrong or miss context.

This replaces my earlier read of this branch. Four of the six recommendations have been applied to the branch as fixup commits, and the four original commits have been retyped `perf` → `fix`. What follows is the state **after** those changes, so a reviewer can read it against the branch as it stands rather than diffing two reviews.

## What changed since the first pass

| | |
|---|---|
| `fixup! fix: stop the 3d tilt hover effect stranding a mousemove listener` | `cleanup()` now also returns the element to rest; doc snippet corrected; two spec cases added/strengthened. |
| `fixup! fix(card): …` | `cleanup3dHoverEffect?.()`, moved after the pre-existing teardown; new `card.e2e.tsx`. |
| `fixup! fix(shortcut): …` | `cleanup3dHoverEffect?.()`; new regression case in `shortcut.e2e.tsx`. |
| `fixup! fix(info-tile): …` | `cleanup3dHoverEffect?.()`; new regression case in `info-tile.e2e.tsx`. |

The four base commits were also **retyped `perf` → `fix`**, and the first one's subject reworded (details in Recommendation 4 below).

**These fixups have since been autosquashed**, so the branch is now the four commits below and the **Block Autosquash Commits** check passes. The squash was content-neutral: the resulting tree hash is byte-identical to the pre-squash tip, and each of the four intermediate trees matches the preview that was verified commit-by-commit.

```
b13a1a1b7  fix(info-tile): remove the 3d tilt listener when the tile is disconnected
e140f824b  fix(shortcut): remove the 3d tilt listener when the shortcut is disconnected
8c3ad4df0  fix(card): remove the 3d tilt listener when the card is disconnected
82df7ed08  fix: stop the 3d tilt hover effect stranding a mousemove listener
```

Still no regression against `main`. The four items that were worth folding in are in; the two that remain are pre-existing and belong in their own PRs.

- **Blockers:** None.
- **Still open:** Recommendations 5 and 6 (both pre-existing, both follow-ups) and the `[Low]` nits listed per dimension.

### Verification of the branch as it now stands

- **Every commit green individually, after autosquash** — walked all four squashed commits: `npm run lint` clean, `npm run test:spec` 1203 passed at each, and the e2e files present at that commit passing. The e2e count climbs **16 → 17 → 18 → 19** across the four commits, i.e. each component's regression test lands in the same commit as the fix it guards. The branch bisects.
- **The four new tests are real regression tests.** Neutering the fixes — dropping the three `disconnectedCallback` calls and reverting `stopTracking` to removing only the listener — makes exactly these four fail and nothing else:
  ```
  × getMouseEventHandlers > returns the element to rest when cleaned up
  × limel-card       > when the card is removed while it is still hovered > stops tilting
  × limel-shortcut   > when the shortcut is removed while it is still hovered > stops tilting
  × limel-info-tile  > when the tile is removed while it is still hovered > stops tilting
  ```
- **Commit messages** — all four pass `commitlint`.
- **Full e2e suite** — 1082 passed / 8 skipped, with one failure in `src/examples-smoke.e2e.ts`. That file is **flaky and unrelated**: it failed on `limel-example-chip-readonly-border` in one run at the tip, on `limel-example-dock-notification` at the base commit `81a8678c3` (i.e. before this PR), and passed twice in a row at the tip. Nothing in this branch touches chips or dock.

### Two CodeRabbit comments — both addressed

Both were valid and both matched `[Low]` nits from the first pass, so they went into the util fixup:

- **The first spec case left its listener attached.** Fixed by calling `cleanup()` at the end of `it('tilts the element while it is hovered')`. A spec about listener removal should not itself leak one.
- **The doc snippet would not type-check** — `querySelector('.the-3d-element')` yields `Element`, not the `HTMLElement` the signature wants. Fixed as `querySelector<HTMLElement>('.the-3d-element')`. I did **not** add CodeRabbit's `!`: this repo sets no `strict`/`strictNullChecks` anywhere in `tsconfig*.json`, so `HTMLElement | null` assigns to `HTMLElement` without complaint and the non-null assertion would only teach a habit the codebase does not need. The `Element` → `HTMLElement` half was the actual error.

---

## Dimensions

Anything dealt with since the first pass is labelled **✅ Fixed after the first round of review**, with the finding itself struck through; anything still outstanding is labelled **⬜ OPEN**.

The "no public API change" claim verifies end to end; the only consumer-observable changes are the three new `disconnectedCallback` hooks, and both of their rough edges are now sanded off.

**What works well:** `src/index.ts` exports only `./components` and `./interface`, and `src/interface.ts` re-exports just two utils — not this one — so `getMouseEventHandlers` never reaches the `.d.ts` rollup or the API report, `etc/lime-elements.api.md` correctly needs no update, and `api:verify` will not fail; adding a third key to a returned object literal cannot break the three destructuring call sites; and the reconnect path is sound, since Stencil does not re-run `componentWillLoad` on re-attach but both the closure and the host's DOM listeners survive a move, so the effect re-arms on the next hover with no `connectedCallback` wiring needed.

**Issues:**
- **✅ Fixed after the first round of review** — ~~🟡 [Medium] `cleanup3dHoverEffect` could be undefined when `disconnectedCallback` ran.~~ See **Top Recommendation 1**.
- **✅ Fixed after the first round of review** — ~~🟡 [Medium] Reparenting a hovered element froze the tilt.~~ See **Top Recommendation 2**.
- **✅ Fixed after the first round of review** — ~~🟡 [Medium] `perf` was the wrong commit type.~~ See **Top Recommendation 4**.

**Minor nits:**
- 🔵 **The dedup plan for `#4280` no longer works.** The reword alone was patch-id neutral (verified: identical `git patch-id --stable` before and after, since patch-id ignores commit messages), but the autosquash folded new content in, so all four commits now have new patches. `#4280`'s copies will not drop out on rebase any more — drop them explicitly there instead of relying on it.
- 🔵 `Refs #4282` on the first three commits and `Closes #4282` on the last is the right split, so the earlier nit about the issue link was wrong — `Refs` already cross-references on every commit. Only `Closes` reaches the changelog, which is the correct outcome.

### 2. Code Quality — GOOD ✅

The fix is small and correctly scoped, the util spec is a genuine regression test, and the wiring now has coverage of its own.

**What works well:** The util change is minimal, and the `stopTracking()` guard in `handleMouseEnter` closes a second real hole (a repeated `mouseenter` stranding an unreachable closure) rather than only the reported one. The spec asserts observable behaviour instead of spying on `removeEventListener`, and its cases were independently confirmed to fail against the unfixed util — `removeEventListener` matches on function identity and each `tiltFollowingTheCursor(bounds, element)` returns a fresh closure, so without the guard `cleanup()` removes only the second registration and the first still writes the tilt property on the next `mousemove`. File naming, plain-DOM style and the absence of `vitest` imports all match the existing `src/util/*.spec.ts` conventions.

**Issues:**
- **✅ Fixed after the first round of review** — ~~🟡 [Medium] Unguarded cleanup call.~~ See **Top Recommendation 1**.
- **✅ Fixed after the first round of review** — ~~🟡 [Medium] The three `disconnectedCallback` wirings had no test.~~ See **Top Recommendation 3**.

**Minor nits:**
- **✅ Fixed after the first round of review** — ~~🔵 The first spec case leaked its own listener.~~ CodeRabbit raised the same point.
- **✅ Fixed after the first round of review** — ~~🔵 The doc snippet would not type-check.~~ CodeRabbit raised the same point.
- **✅ Fixed after the first round of review** — ~~🔵 `stops tilting once the pointer leaves` checked only the tilt property.~~ It now asserts the glow property too, which matters more than before since `handleMouseLeave` and `cleanup` are the same teardown.
- 🔵 `does not throw when cleaned up before ever being hovered` (`src/util/3d-tilt-hover-effect.spec.ts`) would pass even if `cleanup` were a no-op, since `removeEventListener` with an `undefined` callback is a legal no-op in both the DOM spec and mock-doc. Cheap and it documents intent, but it is not a regression guard like the others. Left as-is.
- 🔵 The mock-doc environment returns an all-zero `getBoundingClientRect()`, so in the spec `scalingFactor` is 0 and the tilt property is the string `Infinity, -Infinity, 0, Infinitydeg`; `not.toBe('')` accepts it. The case therefore proves the listener was attached but none of the tilt math. The three new browser tests do run against real geometry, so this is now much less of a gap. Stubbing a non-zero rect in `beforeEach` would still sharpen the spec.
- 🔵 Nothing pins that `cleanup` is re-armable. Stencil does not re-run `componentWillLoad` on re-connect, so a reconnected card depends on `cleanup()` being a reversible "stop tracking" rather than a permanent teardown; it is, but nulling `tiltCallback` later would silently kill the effect on reconnect with no test failing. One case covers it: `handleMouseEnter(); cleanup(); handleMouseEnter(); moveMouse();` then assert the tilt property is non-empty.
- 🔵 The same wiring is copy-pasted into three components. Storing one field (`private hover3dEffect: ReturnType<typeof getMouseEventHandlers>;`) would collapse the destructuring. Acceptable as-is; **Top Recommendation 6** would remove the need for it entirely.

### 3. Architecture — GOOD ✅

A cleanup handle is the only teardown shape a plain function utility can offer in Stencil, and the two teardown paths have now been collapsed into one.

**What works well:** `cleanup` is idempotent and reusable — after disconnect/reconnect the closure still works because `element` is the host and never changes, and `handleMouseEnter` recomputes `bounds` and clears any prior callback before attaching a new one, so `componentWillLoad` (run once) is a coherent place to build the handlers even though `disconnectedCallback` runs on every disconnect. The re-entrancy guard closes a stranding path the linked issue does not mention, and one `limel-card` can actually produce by toggling `show3dEffect` mid-hover. The four-way commit split gives each consumer its own scope and changelog line while keeping every commit green and bisect-safe, and the util spec sits at the level the repo tests its other utilities (`link-helper.spec.ts`, `hotkeys.spec.ts`).

**Issues:**
- **✅ Fixed after the first round of review** — ~~🟡 [Medium] `handleMouseLeave` and `cleanup` had different semantics.~~ They are now one function, so the API has a single teardown contract. See **Top Recommendation 2**.
- **⬜ OPEN** — **🟡 [Medium]** The design still depends on every consumer remembering to call `cleanup()`, with a `document`-level listener as the thing forgotten (`src/util/3d-tilt-hover-effect.ts`). See **Top Recommendation 6** — left open deliberately.
- **⬜ OPEN** — **🟡 [Medium]** `limel-card` can still strand a listener when `show3dEffect` flips off mid-hover (`src/components/card/card.tsx`). Pre-existing; this PR narrows the window but does not close it. See **Top Recommendation 5**.

**Minor nits:**
- 🔵 `limel-card`'s 3D-effect activation condition is expressed twice and the copies disagree: the SCSS activates `the-3d-element` for `:host([clickable])` **or** `:host([show-3d-effect])` (`card.scss`), while the TSX attaches the handlers only for `show3dEffect` (`card.tsx`). A `clickable` card with `show3dEffect: false` gets a `:hover` rule containing `rotate3d(var(--limel-3d-hover-effect-rotate3d))` for a property nothing sets, which invalidates the whole `transform` declaration and drops the hover scale with it. Pre-existing and unrelated — but see the next nit.
- 🔵 Rebase-time check for `info-tile`: on this branch the tile tilts unconditionally in both layers (`info-tile.tsx`, `info-tile.scss`), so the wiring here is coherent as it stands. Once `main`'s `fix(info-tile): drop the 3d hover effect when the tile is not a link` is rebased in, the tile takes on card's shape — the activation condition becomes duplicated across TSX and SCSS, and `link` becoming undefined mid-hover opens the same window as `show3dEffect`. Worth re-checking after that rebase.
- 🔵 `limel-popover` has the same defect this PR fixes, and no `disconnectedCallback` at all (`src/components/popover/popover.tsx`): `setupGlobalHandlers` adds `click` (capture) and `keydown` on `document` when `open` becomes true and removes them on the false transition, so a popover removed while still open leaks both plus a reference to the detached host. `date-picker`, `menu`, `menu-surface` and `editor-link-menu` all have teardown and look fine. Worth its own issue; the pattern established here is what fixes it.
- 🔵 On the alternatives: `AbortController` does not help — `{ signal }` still needs somebody to call `abort()` from `disconnectedCallback`, so it is the same contract with more machinery; a `MutationObserver` inside the util would have to watch a document-wide subtree, trading a leaked `mousemove` for a permanently-running observer; and `pointerleave`/`pointercancel` do not fire on node removal either. Only the element-level listener in **Top Recommendation 6** removes the contract.

### 4. Security — SAFE ✅

Nothing here touches untrusted data, and the CSS custom properties the effect writes are numeric-only and land in `transform` and gradient-position slots that cannot load resources or escape their declaration.

**What works well:** The values written via `element.style.setProperty` in `tiltFollowingTheCursor` come entirely from `MouseEvent.clientX/clientY` and `DOMRect` numbers, which WebIDL coerces to numbers even for a synthetic event, so no caller-supplied string can reach a custom property; their only two consumers — `rotate3d(var(--limel-3d-hover-effect-rotate3d))` in `src/style/mixins.scss` and `radial-gradient(circle at var(--limel-3d-hover-effect-glow-position, 50% -20%), …)` in `src/components/3d-hover-effect-glow/3d-hover-effect-glow.scss` — use them as a transform argument and a gradient position, with no `url()` and no `background` shorthand, so the worst case (a degenerate rect yielding `NaN`/`Infinity`) is a declaration the browser drops. The PR is a net improvement on resource exhaustion: it removes an unbounded accumulation of `document` listeners, a genuine slow-degradation vector in a long-lived single-page application. `addEventListener` and `removeEventListener` are called with matching options, so removal matches registration; no secrets, URLs or PII in the new tests and no network access; no `package.json` or lockfile changes.

**Issues:** None.

**Minor nits:**
- **✅ Fixed after the first round of review** — ~~🔵 In `card.tsx` the new call sat ahead of the pre-existing teardown, so a throw would skip it.~~ It now runs last and is optional-chained. See **Top Recommendation 1**.
- 🔵 `link.href` goes straight into `<a href>` with no protocol check in `info-tile.tsx` and `shortcut.tsx`, so a `javascript:` URL from untrusted data would execute on click. `getHref` exists in `src/util/link-helper.ts` but is only used by `chip-set` and `input-field`, so the pass-through looks like the deliberate library-wide convention. Pre-existing, untouched here, noted for context only.

### 5. Observability — GOOD ✅

No logging added and none needed — the repo deliberately has no logging framework. The diagnostic artifacts are the doc comment and the tests, and the test half is now materially stronger.

**What works well:** The repo's only `console.*` calls are deprecation notices plus one `console.error` in `code-editor.tsx` and one `console.debug` in `image-resize.ts`, with no logger util or telemetry in `src/`, so adding none is right. The `cleanup` property carries its own TSDoc on the returned object, which reaches an implementer on editor hover, and the inline comment in `handleMouseEnter` explains the double-`mouseenter` case at the point of the code that handles it.

**Issues:**
- **✅ Fixed after the first round of review** — ~~🟡 [Medium] The three component wirings had no coverage — the same silence that made #4282 hard to find.~~ See **Top Recommendation 3**.
- **✅ Fixed after the first round of review** — ~~🟡 [Medium] The unguarded call would have surfaced as a console error naming the 3D effect for an unrelated problem, and would have skipped the teardown after it.~~ See **Top Recommendation 1**.

**Minor nits:**
- 🔵 **Correction to the PR description:** the new doc-comment warning is *not* rendered into the docs site. `src/index.ts` re-exports only `./components` and `./interface`, so the kompendium target (`typeRoot: './src/index.ts'`) never reaches this file; it is absent from `etc/lime-elements.api.md` and from `guides.ts`; the context7 `docs-readme` target covers components only; and `src/style-utilities.md`, which documents the publicly consumable mixins, never mentions `parent-of-the-3d-element` or `the-3d-element`. In substance this is fine — the audience is an internal contributor who opens the file, and the TSDoc reaches them on hover — but the description should be corrected, and `src/style-utilities.md` is where a reader would actually look.
- 🔵 The spec's cleanup cases still infer "the listener is gone" from a custom property staying empty, so a future defensive change (`tiltFollowingTheCursor` returning early when `!element.isConnected`) would keep them green while the listener still leaked. The three new browser tests have the same shape. The repo has precedent for the direct assertion at `src/components/table/row-drag-manager.spec.ts`, which checks `removeEventListener` was called with the right arguments.
- 🔵 `tilts the element while it is hovered` still promises a working tilt in its name while only verifying that something was written (see the zero-rect nit under Code Quality). The other test names read well as diagnostics.

### 6. Performance — GOOD ✅

The leak is closed on every path traced, the per-instance cost is negligible, and the one place where this PR made teardown ordering worse has been corrected.

**What works well:** The leak analysis in the description holds up — on `main` the `mousemove` closure captured both `element` and its `DOMRect`, so `document`'s listener list kept the host, its shadow root and the component instance reachable forever, with no handle to remove it — and the fix reaches all three (and only three) call sites, with every disconnect path now removing the listener. `tiltFollowingTheCursor` reads no layout during the move (the rect is captured once at `mouseenter`), so there is no read/write interleaving and no layout thrashing, and the new code adds nothing to that hot path. Bundle impact is a few bytes: no new imports, and `**/*.spec.*` / `**/*.e2e.*` are excluded from the shipped build.

**Issues:**
- **✅ Fixed after the first round of review** — ~~🟡 [Medium] The unguarded call, placed first in card's `disconnectedCallback`, could throw and skip cleanup that on `main` could not be skipped.~~ See **Top Recommendation 1**.
- **⬜ OPEN** — **🟡 [Medium]** The listener still lives on `document`, so every hovered instance is called for mouse movement anywhere on the page, and correctness depends on each consumer honouring the contract. See **Top Recommendation 6**.
- **⬜ OPEN** — **🟡 [Medium]** `limel-card` can still leave the listener running when `show3dEffect` flips off mid-hover. See **Top Recommendation 5**.

**Minor nits:**
- 🔵 `stopTracking` does not release `tiltCallback`, so after cleanup the closure still points at the last callback and its captured `DOMRect`. Nothing outside the component graph holds it, so it is collected with the element and is not a leak — but `tiltCallback = undefined` after the `removeEventListener` would drop the rect immediately and make it obvious the handle has been released.
- 🔵 `bounds` is captured once per hover and goes stale: scrolling or resizing during a hover leaves the rect wrong, so every subsequent move does the same work to produce a wrong rotation and a glow offset that does not track the cursor. Reading the rect once is the right performance call; the missing piece is invalidation on `scroll`/`resize` while hovered. Pre-existing.
- 🔵 Per-move writes are not batched: each event builds two template strings and does two `setProperty` calls on a host that also declares `container-type: size` and feeds a `mix-blend-mode: plus-lighter` gradient child, so each write invalidates style for the shadow subtree and repaints the glow. Chrome and Safari already coalesce `mousemove` to about one event per frame, so a `requestAnimationFrame` wrapper would buy little and add a handle that itself needs cancelling. Pre-existing; worth knowing, not worth changing here.
- 🔵 `limel-card` allocates handlers for cards that never tilt: `getMouseEventHandlers` runs unconditionally in `componentWillLoad`, but `render` attaches the handlers only when `show3dEffect` is true, so a card with the effect off pays a few closures and an object for nothing. Tens of bytes per instance, not measurable even for a hundred cards. This does **not** apply to `info-tile` or `shortcut`: on this branch `info-tile.scss` applies `the-3d-element` to every `<a>` and gates only `the-3d-element--clickable` behind `.is-clickable`, and `shortcut.scss` applies both unconditionally, so every tile and shortcut genuinely needs the handlers.

---

## Top Recommendations

1. **✅ Fixed after the first round of review** — 🟡 [Medium from Code Quality, Backward Compatibility, Observability, Performance] — ~~Call the new cleanup with optional chaining, and put it last in card's teardown~~
   *Introduced by this PR. **Applied**, squashed into the three component commits.*
   - **Where:** `src/components/card/card.tsx`, `src/components/info-tile/info-tile.tsx`, `src/components/shortcut/shortcut.tsx`
   - **What was done:** the field is now `private cleanup3dHoverEffect?: () => void;` and the call `this.cleanup3dHoverEffect?.();` in all three. In `card.tsx` it also moved to the end of `disconnectedCallback`, after `markdownResizeObserver?.disconnect()` and the scroll-listener removal. The doc snippet in the util was updated to teach the same shape.
   - **Why:** the field was declared non-optional but assigned in `componentWillLoad`, which Stencil defers — `scheduleUpdate(hostRef, true)` ends in `queueMicrotask(() => dispatchHooks(...))` while the instance already exists — so an element connected and disconnected in the same task reached `disconnectedCallback` with the field still `undefined` and the call threw `TypeError`. The agents disagreed on reachability in the shipped lazy `dist` build (one traced a `$onReadyPromise$` deferral that would prevent it; three traced paths that would not, including the `isWaitingForChildren` early return and the case where the element is a child of another Stencil component, which is how these three are normally used), but all agreed it is deterministic for a consumer who rebuilds `dist/collection` as `dist-custom-elements`, where the proxy calls the user method with no initialization guard and no `try/catch`. The ordering was a separate point: the new call was the first statement in card, so a throw would have skipped two teardown steps that on `main` could not be skipped, because every statement there was already optional-chained.

2. **✅ Fixed after the first round of review** — 🟡 [Medium from Backward Compatibility, Architecture, Performance] — ~~Have `cleanup` return the element to rest, not just drop the listener~~
   *Introduced by this PR. **Applied**, squashed into the shared-util commit.*
   - **Where:** `src/util/3d-tilt-hover-effect.ts`
   - **What was done:** the two `element.style.removeProperty(...)` calls moved into `stopTracking`, so there is now a single teardown that both `handleMouseLeave` and `cleanup` point at. A spec case (`returns the element to rest when cleaned up`) pins it, and the `mouseleave` case now asserts the glow property too.
   - **Why:** the original rationale — "no point styling a node being thrown away, and the element may already be detached" — is right for removal but does not cover a DOM *move*. The repo documents the constraint itself in `src/components/portal/portal.tsx`: moving a node fires `disconnectedCallback` then `connectedCallback`, so teardown on disconnect needs matching setup on connect. A hovered element that was reparented (portal relocation, masonry re-layout, row drag-and-drop, a consumer's `appendChild`) lost its `mousemove` listener while keeping `--limel-3d-hover-effect-rotate3d` set — and no `mouseleave` fires on a move any more than on a removal, so nothing cleared it and nothing re-attached. `src/style/mixins.scss` applies `rotate3d(var(--limel-3d-hover-effect-rotate3d))` under `&:hover`, so the element sat visibly tilted at a stale angle until the pointer left or re-entered. `removeProperty` on a detached node is safe and cheap, so nothing is lost in the removal case.

3. **✅ Fixed after the first round of review** — 🟡 [Medium from Observability, Code Quality] — ~~Cover the three `disconnectedCallback` wirings with a test~~
   *Introduced by this PR. **Applied** — one browser test per component, each in that component's own commit.*
   - **Where:** new `src/components/card/card.e2e.tsx`; new cases in `src/components/shortcut/shortcut.e2e.tsx` and `src/components/info-tile/info-tile.e2e.tsx`
   - **What was done:** each test renders the component, dispatches `mouseenter` on the host, moves the mouse and asserts the tilt property was written, then `unmount()`s (which removes the container, disconnecting the host) and moves the mouse again, asserting the property stays empty. These run in the `e2e` project — real Chromium via Playwright — so unlike the mock-doc spec they exercise real geometry. Each landed in the same commit as the fix it guards, which is why the e2e count goes 16 → 17 → 18 → 19 across the four commits.
   - **Why:** the spec exercised the util only. Deleting `disconnectedCallback` from any of the three components left the whole suite green — the contract was tested, the three uses of it were not, and the use is the part that gets forgotten. Confirmed the new tests fail when the wiring is removed. The broader idea from the first pass — a teardown step in the blanket `examples-smoke.e2e.ts` / `example-tests/runtime.spec.ts` harnesses, which are mount-only today and would then cover every future consumer automatically — is still a good follow-up, and deliberately not done here.

4. **✅ Fixed after the first round of review** — 🟡 [Medium from Backward Compatibility, Architecture] — ~~Retype the commits `perf` → `fix`, and reword the first subject~~
   *Introduced by this PR. **Applied** by rewording the four base commits.*
   - **Where:** all four commit messages
   - **What was done:** all four are now `fix`, and the first reads `fix: stop the 3d tilt hover effect stranding a mousemove listener` rather than naming the internal utility and the word "removable". Its body was also corrected: it used to say `cleanup()` "removes the listener without touching the element", which Recommendation 2 makes untrue. All four pass `commitlint`. The reword itself left every patch-id byte-identical; the later autosquash is what changed them (see the Backward Compatibility nit about `#4280`).
   - **Why:** `src/commits-and-prs.md` defines `perf` as "a code change that **only** improves performance, without changing behavior", and says that "if the performance was previously so bad it could be considered a bug, and that's been fixed, `fix` is a good fit". Both halves point the same way here. An unbounded `document` listener that retains a detached node is a bug — it has an issue number, filed on the bug template. And these commits do not *only* improve performance: the `handleMouseEnter` re-entrancy guard changes what a second `mouseenter` does, and Recommendation 2 changes what a reparented element looks like. That is behaviour, so `perf` does not fit on its own terms. The same doc's rule that a `fix`/`perf` subject "should focus on the change for the consumer" is what the first subject was failing; both types release as a patch bump, so nothing about the version changes.

5. **⬜ OPEN** — 🟡 [Medium from Performance, Architecture] — Tear down the listener when `show3dEffect` flips off mid-hover
   *Pre-existing. Recommended: Better as a follow-up PR — it needs a `@Watch` on a prop this PR does not touch, and the leak's lifetime is already cut from "forever" to "until the next hover or disconnect".*
   - **Where:** `src/components/card/card.tsx`
   - **What:** add a `@Watch('show3dEffect')` that runs the teardown when the value becomes `false`.
   - **Why:** the host's `onMouseEnter`/`onMouseLeave` are `undefined` whenever `show3dEffect` is false, so setting the prop to `false` while the pointer is inside the card removes the only handler that would have detached the `mousemove` listener — the same class of bug as #4282, a listener whose removal trigger has been taken away. The listener then keeps writing both custom properties on every mouse move for a card that no longer renders the glow. This PR bounds it in two ways (the next `mouseenter` clears it, and disconnect clears it), so the branch is strictly better than `main`, but the case is not closed. The same window will open for `limel-info-tile` on `link` once `main`'s "not a link" change is rebased in.

6. **⬜ OPEN** — 🟡 [Medium from Performance, Architecture] — Consider attaching `mousemove` to the element instead of `document`
   *Pre-existing. Recommended: Better as a follow-up PR — it changes the effect in two edge cases that need a design decision, and this branch's approach works.*
   - **Where:** `src/util/3d-tilt-hover-effect.ts`
   - **What:** attach and remove the `mousemove` listener on `element` rather than `document`. Listeners on a removed node need no cleanup, so `cleanup()` disappears — and with it the `disconnectedCallback` in all three components and the three duplicated wiring blocks. `mousemove` bubbles and events from shadow descendants retarget to the host, so the event stream is nearly identical: the pointer is over the host whenever the effect is active, by definition.
   - **Why:** the current approach adds a contract that all three consumers and every future one must honour, enforced only by a doc comment that (see the Observability nit) is not published anywhere a consumer reads. Element-level listeners would make the leak structurally impossible instead, and would stop each hovered instance from being called for mouse movement in unrelated parts of the page. Two behaviours would change, and are the reason to think first: the moment the hover transform scales and rotates the element out from under a moving cursor, and any non-descendant overlay sitting on top of the tile — in both, a `document` listener keeps tracking where an element listener would stall. A closely related follow-up: `limel-popover` has the same leak shape today, and whichever pattern wins here is what fixes it.

adrianschmidt (approved):
LGTM

PR #4289, opened 2026-09-10
feat(form): let a select option carry an icon and a secondary text

TommyLindh2:
## Summary by CodeRabbit

- **New Features**
  - Select and multi-select options can display descriptions and icons defined in the form schema.
  - Options marked as read-only remain visible but cannot be selected.
  - Selected options retain their icons in the field display.
  - On mobile, rich options use the enhanced dropdown, while text-only options use the native selector.
  - Mobile typeahead supports matching options with icons.

- **Documentation**
  - Updated form example guidance for option descriptions, icons, and read-only behavior.

## Why

A `limel-form` select is built from a `oneOf` (or, for a multi select, `items.anyOf`), and each alternative already controls more than its own text: `readOnly: true` on an alternative renders that option as disabled. That is genuinely useful — but it was the *only* piece of `Option` a schema could reach. A form could grey out an option without being able to say why, and `limel-select` has supported icons and secondary text on options for a long time.

Since we already let an alternative say "this one cannot be picked", letting it also carry the metadata that explains that seems like the natural next step.

## What

`createOption` in `src/components/form/widgets/select.ts` is the single place every option — single and multi — passes through. It now fills in two more fields from the alternative's schema:

| `Option` field | Comes from | Note |
| --- | --- | --- |
| `secondaryText` | `description` | Plain JSON Schema; was ignored on an alternative until now |
| `icon` | `lime.icon` | Takes an `IconName` or the `Icon` interface, so it can be colored |
| `disabled` | `readOnly` | Unchanged |

`lime.icon` is a new key on `LimeSchemaOptions`.

```ts
{
    type: 'string',
    title: 'Priority',
    oneOf: [
        {
            const: 'high',
            title: 'High Priority',
            description: 'Handle it today',
            lime: { icon: { name: 'notification_alert', color: 'rgb(var(--color-yellow-darker))' } },
        },
        {
            const: 'critical',
            title: 'Critical Priority',
            description: 'Can only be set by the support team',
            readOnly: true,
            lime: { icon: { name: 'error', color: 'rgb(var(--color-red-default))' } },
        },
    ],
}
```

From the [docs example All built-in field types](https://lundalogik.github.io/lime-elements/versions/PR-4289/#/component/limel-form/#all-built-in-field-types)

Screenshot 2026-09-10 at 10 06 05

Because the options are mapped *before* the selected value is looked up among them, the icon follows the selected option onto the trigger, and not only into the dropdown list.

### Multi selects too

rjsf's `ArrayAsMultiSelect` builds its choices with `optionsList(itemsSchema, …)` and renders the same widget with `multiple: true`, so alternatives under `items.anyOf` / `items.oneOf` carry their sub-schema exactly like a single-value `oneOf`. Both paths go through the one conversion, and `limel-select` already renders option icons inline in the trigger for multiple values.

A plain `enum` is unchanged: rjsf gives those choices no sub-schema to read from, so they still get only a text and a value.

`primaryComponent` is deliberately left out — it takes a `ListComponent`, which needs its own story rather than a schema key.

## Heads-up for reviewers

Mapping `description` to `secondaryText` is a **visible change for anyone already setting `description` on an alternative**, where it currently renders nothing. I grepped `lime-crm-components` for that shape and found no occurrences, so I believe the blast radius is nil, but it is worth a second pair of eyes.

## Tests

Three new tests in `form.e2e.tsx`: a `oneOf` mapping to options, an `anyOf` array mapping to multi options, and one asserting the icon survives onto `select.value` so the trigger can render it.

The example (`limel-example-builtin-field-types-form`) now shows icons, secondary texts, and one `readOnly` option on its select and multi-select fields, so the behavior is visible in the docs.

`npm run lint` and `npm run build` are clean. `form.e2e.tsx` runs 61 tests with 59 passing; the two failures (`keeps a cleared field empty when the schema declares a default` and `… when another field is cleared afterwards`) reproduce identically on a clean `main` — 58 tests, same two red — so they are pre-existing and unrelated.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


etc/lime-elements.api.md
adrianschmidt, line 1: (Making this an inline comment, so please can reply in a thread 🙂)

This seems like a very nice addition! I just want to flag that I _think_ our `limel-select` component currently switches to using the native `<select>` element when on a mobile device, and that doesn't have support for showing icons. So I would encourage you to have a look at that, just to verify if that's indeed the case or not (we may have changed that already, or I may simply be wrong, and that was never the case).

> [!IMPORTANT]
> This can probably _not_ be checked by using your desktop browser's dev tools' mobile simulation mode. But you can just load the published docs on your phone and check.

If we do indeed use the native component for mobile browsers, I think it's worth making a conscious choice on whether we keep doing that, and options simply don't show icons and secondary text in the selection list (selected options should still show up the same in the component, apart from the list, I think?), or if we change so that we no longer use the native component, and, in that case, do we just stop using the native component in any case, or do we make that decision on the fly, depending on whether any options have an icon or secondary text defined?

cc @Kiarokh @LucyChyzhova @paulinewahle

src/components/form/widgets/select.ts
```diff
@@ -62,15 +63,34 @@ export class Select extends React.Component {
...
+ * `schema` is the schema of the corresponding alternative in a `oneOf` or
+ * an `anyOf`, and is only present for such schemas. A plain `enum` has no
+ * sub schema to read from, so those choices only carry a label and a value.
+ */
+interface EnumOption {
     label: string;
     value: string;
-    schema?: Record<string, unknown>;
-}): Option {
+    schema?: FormSchema;
```
adrianschmidt, line 76: I haven't analysed this with an AI, so take it with healthy scepticism, but my gut feeling is that it seems a bit weird to have the possibility to pass in a whole `FormSchema` to a single option in a `select`, just to be able to specify an icon and a secondary text.

Again, I haven't let an AI dig through and analyse, so I may simply be misunderstanding what `FormSchema` actually is, or what component this change applies to. Or both are exactly what I think and this is actually exactly what we want to do, and my gut feeling is just wrong. But I would like someone, AI or human, to consider if we are simply reusing a way too big type, because it happens to have some overlap.

src/components/form/form.types.ts
```diff
@@ -214,6 +215,29 @@ export interface LimeSchemaOptions {
...
+     *     oneOf: [
+     *         {
+     *             const: 'archived',
+     *             title: 'Archived',
+     *             lime: { icon: { name: 'archive', color: 'grey' } },
+     *         },
+     *     ],
+     * };
+     * ```
+     */
```
adrianschmidt, line 238: This comment seems to presuppose that this type is being used to pass options to an option belonging to a select component, but this is a general type that can be used on any schema field.

It also references behaviour in a component that may well change that behaviour, without this comment ever being updated.

These massive comments tend to just cause lots of review churn in future PRs, as they become stale. I'd prefer this to have a short and concise comment describing what we intend this field to "mean" when a value is defined, like the docblocks for the other properties in this type.

etc/lime-elements.api.md
TommyLindh2: A very valid point, I tried on my iPhone and got this result:

| Single option field <br> Native picker without icon/description | Multi option field <br> Custom picker with icon/description |
|--------|--------|
| image | image |

src/components/form/widgets/select.ts
TommyLindh2: Looking at the `FormSchema` types the `oneOf` and `anyOf` are FormSchemas, so I think it makes sense that is what we use here.

etc/lime-elements.api.md
adrianschmidt: Personally, I feel like that's a perfectly OK UX. The case where it would break down on mobile is if someone creates a list of options that _only_ have an icon, and no text. The best developer experience would clearly be if we detect that kind of case (perhaps detecting if there is any option that has an icon and no text), and then not use the native mode.

Kiarokh: I think limel-select has too many non-natively-supported features, and we want to keep adding to this list of features. For example, consumers have requested a "searcher" in other threads. 

So a good design for me is that the component automatically switches to the "native" select, only for single selection dropdowns, with list of options that purely use `text` in their `options`. For any other feature, we need to switch to our custom UI. Consumers shouldn't bother deciding anything

src/components/form/form.types.ts
TommyLindh2: I agree, I have made the comment much shorter and more non-specific 👍 

cf6f17df3744d08bfa030712892c37174a0dd127

etc/lime-elements.api.md
TommyLindh2: I totally agree with you Kia, should that be addressed in this PR or is that another PR.
What do you think? @adrianschmidt @Kiarokh

adrianschmidt: It belongs here imo, since this PR introduces the thing we would be switching on 👍

(It ought to be a pretty small addition as well. As Kia said, and as seen in your screenshots above, we already have similar switching.)

TommyLindh2: We've already supported icons and secondary text from before - So it's not something new in the select component.
I've only added support to pass it from a FormSchema in a limel-form.

But I agree that it's still a small fix in here

TommyLindh2: Here's the change of rendering native select: 9ad3f3b4df8e275a0482800e07dc6a41ce178f25

So now we render our own drop-down when any options have a primary component, secondary text or an icon

TommyLindh2: After the change this is on mobile:

image

Kiarokh: looks great now 😍

src/components/select/select.tsx
```diff
@@ -740,16 +741,21 @@ export class Select {
     }
 
     private shouldRenderNative(): boolean {
-        return (
-            this.isMobileDevice &&
-            !this.multiple &&
-            !this.hasPrimaryComponentMemo
-        );
+        return this.isMobileDevice && !this.multiple && !this.hasRichOptions;
```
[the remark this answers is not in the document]
TommyLindh2: Valid point, this have probably been a problem for primaryComponent from before.

Fixed - 011432b1b717a649fa326cc72aec080a8d5db199

Kiarokh:
> 🤖 AI-generated review from 6 parallel agents. Treat as input, not a verdict — agents can be wrong or miss context.
>
> *Generated by Claude Opus 5.*

# Consolidated PR Review

## PR Summary

Lets an alternative in a `oneOf` / `anyOf` carry two more pieces of `Option`: its `description` becomes the option's `secondaryText`, and a new `lime.icon` key becomes the option's `icon`. A second commit widens `limel-select`'s mobile check so that any option carrying an icon or a secondary text keeps the custom dropdown instead of falling back to a native `<select>` that would silently drop them. 9 files, +284 / −34, with three new form tests and a restructured mobile-dropdown suite.

The feature itself is clean and lands in exactly one place, and the mobile change fixes a real silent failure. But the second commit breaks an assumption two other methods still rely on: `setMenuFocus()` and `handleTypeaheadKey()` both return early on `isMobileDevice`, believing mobile always means a native `<select>`. That is no longer true, so a phone user with icon-bearing options now gets a dropdown nothing ever focuses. The fix is the same two-line substitution the PR already makes at `select.tsx:487`.

- **Blockers:** None that open a new defect class — `main` already has this focus gap for options with a `primaryComponent`. But **Top Recommendation 1** widens it from a rare case to a common one for a two-line cost, and should land before merge.

---

## Dimensions

### 1. Backward Compatibility — FAIR ⚠️

The published type surface is purely additive, but a large group of existing consumers is routed onto a mobile path whose focus management was never finished.

**What works well:** one added optional property in `etc/lime-elements.api.md`, nothing removed or narrowed; `IconName` degrades to `string` unless a consumer augments `IconNameRegistry`, so no existing schema stops compiling; the `description` blast radius was independently re-checked across `lime-crm-components`, `lime-crm-building-blocks` and `lime-webclient` and came back empty, matching the author's own grep.

**Issues:**
- **🔴 [High]** Mobile focus and typeahead still assume a native dropdown (`src/components/select/select.tsx:294`, `:600`). Options with an icon now render the custom dropdown on a phone, where focus is never moved into the list and typeahead is skipped. See **Top Recommendation 1** below for the fix.
- **🟡 [Medium]** An auto-select `change` event silently stops firing on mobile (`src/components/select/select.tsx:486-489`). Correct behaviour, but no test pins it and no changelog line announces it. See **Top Recommendation 6**.
- **🟡 [Medium]** The second commit is typed `feat` but is a `fix` by the repo's own rule (`src/commits-and-prs.md:104-135`). See **Top Recommendation 7**.
- **🟡 [Medium]** The `description` remapping — the most consumer-visible change here — will not appear in the changelog at all (`commit.hbs`). See **Top Recommendation 8**.
- **🟡 [Medium]** `lime.icon` is documented as if it applied anywhere, and collides with the existing `lime.layout.icon` (`src/components/form/form.types.ts:218-221`). See **Top Recommendation 3**.

**Minor nits:**
- 🔵 Every emitted `Option` now always carries `secondaryText` and `icon` keys, taking `Object.keys().length` from 3 to 5 — traced through `JSON.stringify`, `toEqual` and the widget's value conversion, with no consumer-visible break (`src/components/form/widgets/select.ts:88-94`).

### 2. Code Quality — GOOD ✅

A focused change with real tests and no regressions; the rough edges are naming, an overstated comment, and a hand-rolled type.

**What works well:** the change is concentrated in `createOption`, genuinely the single funnel both single and multi selects pass through, so one three-line change covers both; the `renderOnMobile` helper in `select.e2e.tsx` removes four copies of the same boilerplate; and the `optionsWithPrimary` fixture is still referenced by the `with a primary component` block, so the restructuring left no dead code behind.

**Issues:**
- **🟡 [Medium]** The new `computeHasRichOptions` comment claims a completeness the code does not have — separators are excluded before the check runs (`src/components/select/select.tsx:747-760`). See **Top Recommendation 5**.
- **🟡 [Medium]** `EnumOption` hand-rolls a type `@rjsf/utils` already exports as `EnumOptionsType<S>` (`src/components/form/widgets/select.ts:66-75`). See **Top Recommendation 9**.
- **🟡 [Medium]** `description` is repurposed as `secondaryText` with no opt-in, unlike `lime.icon` (`src/components/form/widgets/select.ts:90`). See **Top Recommendation 4**.
- **🟡 [Medium]** The new `lime.icon` key is documented in one vague line (`src/components/form/form.types.ts:218-221`). See **Top Recommendation 3**.

**Minor nits:**
- 🔵 `hasRichOptions` / `computeHasRichOptions` name a value judgement rather than the criterion; `optionsNeedCustomDropdown` would carry the rule, and the dropped `Memo` suffix was the only hint the field is cached (`src/components/select/select.tsx:140`, `:744`).
- 🔵 "keeps the icon on the selected option, so the trigger can render it" never touches the trigger — `findValue` returns the same object the previous test already checked, so the assertion is close to tautological (`src/components/form/form.e2e.tsx:186-204`).
- 🔵 The multi-select trigger goes through a different path (`renderOptionWithIcon`, `select.template.tsx:383-407`) and has no coverage.
- 🔵 The `secondaryText: undefined` / `icon: undefined` entries in the new `toEqual` objects assert nothing — `toEqual` ignores undefined-valued keys.
- 🔵 No test pins the "a plain `enum` is unaffected" claim; one `expect(select.options).toEqual([...])` against `enumSchema` would lock it in (`src/components/form/form.e2e.tsx:82-90`).
- 🔵 `describe('choosing the dropdown on a mobile device')` is nested inside `describe('limel-select (menu)')` but its first test asserts the *native* dropdown, so a native failure gets attributed to the menu suite (`src/components/select/select.e2e.tsx:472`).

### 3. Architecture — GOOD ✅

The layering is right — one conversion point in the form widget, and the "can a native `<select>` show this?" decision moved into `limel-select` where it belongs.

**What works well:** the single `createOption` funnel keeps the two rjsf code paths from drifting; no form knowledge leaks into `limel-select` and no select rendering knowledge leaks into the widget; typing `EnumOption.schema` as `FormSchema` rather than `Record<string, unknown>` is what makes the whole mapping type-safe; and having `getFirstNativeAutoSelectOption()` guard on `shouldRenderNative()` fixes a latent bug on `main`.

**Issues:**
- **🟡 [Medium]** `lime.icon` is option-scoped but lives in the field-scoped `LimeSchemaOptions`, beside an existing field icon (`src/components/form/form.types.ts:218-221`). See **Top Recommendation 3**.
- **🟡 [Medium]** `description` → `secondaryText` is unconditional with no way for a schema to opt out (`src/components/form/widgets/select.ts:91`). See **Top Recommendation 4**.
- **🟡 [Medium]** Separators still slip past the new "native cannot render this" rule (`src/components/select/select.tsx:747-760`). See **Top Recommendation 5**.

**Minor nits:**
- 🔵 The list of "not natively renderable" fields in `computeHasRichOptions` is a second copy of what `renderOption` supports, with nothing enforcing they stay in sync (`select.tsx:753-760` vs `select.template.tsx:253-268`).
- 🔵 Each further `Option` field pulled from the schema costs one key in `LimeSchemaOptions`, one line in `createOption`, and possibly one in `computeHasRichOptions` — linear rather than free, though acceptable at this size.

### 4. Security — GOOD ✅

Both new schema-to-DOM paths are safe as rendered, and everything the PR newly exposes was already reachable from a schema through wider channels.

**What works well:** `secondaryText` is interpolated as a JSX child and a reflected attribute only — no `innerHTML`, no markdown, no concatenated `title` or `aria-label` on that path; icon colours reach the DOM through Stencil's style accessor (`setProperty()`), so a crafted colour cannot escape the style context, and neither `color` nor `background-color` accepts `url()`, which closes the CSS exfiltration angle.

**Issues:**
- **🟡 [Medium]** A schema can name an arbitrary same-origin URL that is fetched and injected with `innerHTML` (`src/components/form/widgets/select.ts:91` → `src/components/icon/icon.tsx:99`). See **Top Recommendation 10**.

**Minor nits:**
- 🔵 The new example presents `readOnly` as access control ("Can only be set by the support team"), but it only sets `disabled` — ajv treats `readOnly` as an annotation, so the value still validates if submitted (`src/components/form/examples/builtin-field-types-schema.ts`).
- 🔵 `limel-select` rebuilds the icon as `{name, color}` in `createMenuItems`, dropping `Icon.title` and `Icon.backgroundColor` before they reach `list-item`'s `aria-label` and background style — incidental rather than deliberate, since `LimeSchemaOptions.icon` is typed as the full `Icon` interface.

### 5. Observability — GOOD ✅

The PR removes a real silent failure and does not introduce one, but the new schema key is silently ignored everywhere except one position.

**What works well:** the mobile change means the configuration a consumer wrote is actually rendered rather than quietly dropped; the rename plus new JSDoc makes the intent legible; and the form tests pin the full mapping including the `undefined` cases, so a regression in `createOption` fails a test rather than losing an icon.

**Issues:**
- **🟡 [Medium]** `lime.icon` type-checks anywhere and does nothing almost everywhere, with no warning — while the form already warns for the equivalent mistake in `fields/schema-field.ts:49-64` (`src/components/form/widgets/select.ts:85-95`). See **Top Recommendation 3**.
- **🟡 [Medium]** A misspelled icon name renders nothing and logs an unhandled rejection that never names the icon (`src/components/icon/icon.tsx:70-78`). See **Top Recommendation 11**.

**Minor nits:**
- 🔵 `triggerIconColorWarning` fires only from `componentDidLoad`, so a deprecated `iconColor` arriving in a later options array is never reported — the existing `@Watch('options')` could carry it (`src/components/select/select.tsx:209-213`).
- 🔵 Checked and clear: the schema-driven icon path cannot provoke that warning, since `createOption` sets `icon` only and `LimeSchemaOptions.icon` has no route to `Option.iconColor`.

### 6. Performance — FAIR ⚠️

The added per-option work is genuinely negligible and the bundle cost is zero, but the mobile switch is triggered by a keyword existing schemas already use.

**What works well:** the new checks in `computeHasRichOptions` are plain property reads with no allocation, and they short-circuit earlier than the version on `main`, which almost always scanned the whole list to return `false`; `getIconName` costs zero bundle bytes, since `select.template.tsx` already imports from that module in the same chunk; and the `shouldRenderNative()` guard removes a spurious `change` event and the render cascade behind it.

**Issues:**
- **🔴 [High]** Mobile forms lose the native picker without opting in, and the replacement is built at page load with no windowing (`src/components/form/widgets/select.ts:91` + `src/components/select/select.tsx:753-760`). See **Top Recommendation 2**.
- **🟡 [Medium]** Each icon-bearing option re-parses its SVG, because the CacheStorage icon cache memoizes nothing (`src/global/icon-cache/cache-storage-icon-cache.ts:21-33`). See **Top Recommendation 12**.
- **🟡 [Medium]** Option identity churn from the form widget forces a full re-render and an O(N) MDCList rebuild, now on mobile too (`src/components/form/widgets/select.ts:20-21`). See **Top Recommendation 13**.

**Minor nits:**
- 🔵 `hasRichOptions` goes stale on in-place mutation of the options array, since Stencil fires `@Watch` only on reference change — the same hole `main` had, but more now depends on it (`src/components/select/select.tsx:148-151`).
- 🔵 Every `render()` reads `--dropdown-z-index` via `getComputedStyle`, forcing a style recalculation on the same path the PR now routes mobile traffic through (`src/components/select/select.tsx:250-252`).

---

## Top Recommendations

1. **🔴 [High from Backward Compatibility]** — Mobile focus and typeahead still assume a native dropdown
   *Introduced by this PR (widens a gap `main` already has for `primaryComponent`). Recommended: Fix in this PR.*
   - **Where:** `src/components/select/select.tsx:294` (`setMenuFocus`) and `:600` (`handleTypeaheadKey`)
   - **What:** Both guards bail out on `this.isMobileDevice`. Gate them on `!this.shouldRenderNative()` instead — the exact substitution the PR already makes at line 487 — and fix the comment at line 596, which states the now-false assumption outright ("The native dropdown on mobile devices does its own typeahead").
   - **Why:** A consumer who has shipped icons on a `limel-select` for years goes, on Android and iOS, from the OS picker with platform focus and screen-reader handling to a list rendered into a portal at the end of `document.body` that nothing ever focuses; `limel-menu-surface` and `limel-portal` have no focus handling of their own to compensate.

2. **🔴 [High from Performance]** — The custom dropdown that replaces the native picker is built at page load and never windowed
   *Introduced by this PR. Recommended: Fix in this PR — at minimum decide deliberately whether `secondaryText` alone should disqualify the native path.*
   - **Where:** `src/components/form/widgets/select.ts:91` combined with `src/components/select/select.tsx:753-760`; mount path at `src/components/select/select.template.tsx:198-229`
   - **What:** Either stop letting `secondaryText` alone force the custom dropdown, or defer mounting `limel-list` in `MenuDropdown` until the dropdown has been opened once.
   - **Why:** The custom path is not lazy — `limel-menu-surface` renders its slot content regardless of `open`, so the whole list is created at page load even if the user never taps the select. Each option costs one `limel-list-item` (two `crypto.randomUUID()` calls in its constructor, ~8 reflected attributes) plus a `limel-icon` with its own shadow root. For a 200-option select that is roughly 400 custom-element upgrades at load against 200 plain `<option>` nodes drawn by the OS before. For a typical 5-to-10-option form select the absolute cost is trivial; the concern is that nothing bounds N and nothing asks the consumer first.

3. **🟡 [Medium from Backward Compatibility, Architecture, Code Quality and Observability]** — `lime.icon` is documented in one vague line, scoped ambiguously, and collides with `lime.layout.icon`
   *Introduced by this PR. Recommended: Fix in this PR — it is published API surface, so it is the one part of this change that is expensive to correct later.*
   - **Where:** `src/components/form/form.types.ts:218-221`, read at `src/components/form/widgets/select.ts:92`; collides with `RowLayoutOptions.icon` at `form.types.ts:313-317`, read at `src/components/form/row/row.tsx:73`
   - **What:** The docstring is just "Displays an icon." Say that the key applies to an alternative inside a `oneOf` / `anyOf` rendered as a select option and has no effect on a field schema, and cross-reference `lime.layout.icon`. Consider nesting it as `lime.option.icon` so the namespace carries the level. Separately, `Select.render()` already reads the parent schema at `select.ts:30`, so a `console.warn` there when `props.schema.lime?.icon` is set would cover both the parent-field and the plain-`enum` mistakes in a few lines.
   - **Why:** Every other key in `LimeSchemaOptions` describes the field the node renders, and `lime.layout.icon` already means "the field's icon" with a different type (`string`, not `IconName | Icon`). `examples/row-layout-schema.ts:35-41` already contains a field using `lime.layout.icon` that also has a `oneOf` — exactly the schema where an author picks the wrong `icon` and gets no icon, no type error, and no console output.

4. **🟡 [Medium from Architecture and Code Quality]** — `description` becomes visible UI text with no opt-out
   *Introduced by this PR. Recommended: Fix in this PR — the decision is one line, and reversing it after release is a second behaviour change on the same key.*
   - **Where:** `src/components/form/widgets/select.ts:90-91`
   - **What:** `lime.icon` is an explicit opt-in key; `description` is a standard JSON Schema keyword existing schemas may already carry on alternatives purely as documentation. Either source the secondary text from `lime.*` as the icon is, or accept the mapping and say so explicitly in the release notes.
   - **Why:** Affected schemas start rendering that text under every option and — through `computeHasRichOptions` — also flip from the native mobile picker to the custom menu, with no way to suppress either short of deleting the `description`. Worth noting the risk was independently confirmed to be small: no `description` on a `const` alternative was found in `lime-crm-components`, `lime-crm-building-blocks` or `lime-webclient`, and rjsf 6's `MultiSchemaField` builds its options with no `schema` key, so the object-alternative picker is untouched.

5. **🟡 [Medium from Architecture and Code Quality]** — Separators slip past the rule the PR just wrote
   *Pre-existing, but the PR's new comment claims to cover it. Recommended: Fix in this PR if the fix is the comment; better as a follow-up PR if it is `<optgroup>` support — that is a rendering feature, not a one-liner.*
   - **Where:** `src/components/select/select.tsx:747-760` against `src/components/select/select.template.tsx:231-234`
   - **What:** The new comment says "Any option that carries more than that forces the custom dropdown", but the check runs over `getOptionsExcludingSeparators()`, and `NativeDropdown` filters separators out too. Either narrow the comment to say it checks per-option fields only, or add a separator check to the condition.
   - **Why:** A `ListSeparator` carries a `text` label the native `<select>` silently drops — exactly the failure mode this PR exists to fix — so `limel-example-select-with-separators` still loses its grouping on a phone, and a reader who trusts the comment will assume otherwise.

6. **🟡 [Medium from Backward Compatibility]** — An auto-select `change` event silently stops firing on mobile, with no test
   *Introduced by this PR. Recommended: Fix in this PR — add the test; the behaviour change itself is correct.*
   - **Where:** `src/components/select/select.tsx:486-489`
   - **What:** Add a test rendering with `data-native` plus an icon-bearing option, asserting no auto-select `change` is emitted.
   - **Why:** On `main`, a mobile single select emitted the first option as a `change` the moment the user focused it; now any select with an icon or secondary text does not. The whole `describe('limel-select (native)')` auto-select suite uses text-only options, so nothing pins the new behaviour, and neither commit message mentions it.

7. **🟡 [Medium from Backward Compatibility]** — The second commit is typed `feat` but is a `fix`
   *Introduced by this PR. Recommended: Fix in this PR — the fixup commits already force an autosquash, so the retype is free.*
   - **Where:** commit `feat(select): use the menu dropdown on mobile for options a native one cannot show`, against `src/commits-and-prs.md:104-135`
   - **What:** Retype it as `fix(select)`.
   - **Why:** The repo's own rule is "can a consumer write a line of code today that they could not write yesterday?" They cannot — `icon` and `secondaryText` were always writable, just silently ignored on mobile, and the doc lists "a prop being ignored" as the textbook `fix`. It decides a minor versus a patch bump, and whether the entry lands under "Features" or "Bug Fixes", which is where a consumer wondering why their mobile select changed will look.

8. **🟡 [Medium from Backward Compatibility]** — The most consumer-visible change in the PR will not reach the changelog
   *Introduced by this PR. Recommended: Fix in this PR.*
   - **Where:** commit `feat(form): let a select option carry an icon and a secondary text`; template at `commit.hbs`
   - **What:** `commit.hbs` renders only the commit subject. Either name the `description` mapping in the subject, or split it into its own commit so it earns its own changelog line.
   - **Why:** The heads-up about `description` suddenly rendering exists only in the PR description, which no upgrading consumer reads — and `src/commits-and-prs.md:34` states exactly this rule about changes hiding inside a commit.

9. **🟡 [Medium from Code Quality]** — `EnumOption` hand-rolls a type rjsf already exports
   *Introduced by this PR. Recommended: Fix in this PR — it is a one-line swap that keeps the useful JSDoc.*
   - **Where:** `src/components/form/widgets/select.ts:20`, `66-75`
   - **What:** Use `type EnumOption = EnumOptionsType<FormSchema>`. `@rjsf/utils` exports it with exactly these three fields (`node_modules/@rjsf/utils/lib/types.d.ts:867`), and every other file in `src/components/form/` imports its rjsf types rather than redeclaring them.
   - **Why:** A duplicated structural type drifts silently on the next rjsf upgrade. Two knock-on details: the hand-rolled `value: string` is narrower than rjsf's `value: any`, so a numeric `enum` handing a `number` into `Option.value` — which `option.types.ts:25-29` explicitly forbids — type-checks as fine; and since `enumOptions` is optional in rjsf's type, the `as EnumOption[]` assertion hides that `.map` would throw if the widget is forced onto a non-enum field via `ui:widget: 'select'`. A `?? []` would cost nothing. Both match what `as any[]` already did, so neither is new.

10. **🟡 [Medium from Security]** — A schema can name an arbitrary same-origin URL that is fetched and injected with `innerHTML`
    *Introduced by this PR as a new entry point; the root cause is pre-existing and untouched. Recommended: Better as a follow-up PR — the fix belongs in `limel-icon` or `getIconName`, which this PR does not otherwise modify.*
    - **Where:** `src/components/form/widgets/select.ts:91` → `src/components/list-item/list-item.tsx:216-248` → `src/global/icon-cache/cache-storage-icon-cache.ts:72-79` → `src/components/icon/icon.tsx:99`
    - **What:** The icon name is interpolated unvalidated into `` `${iconPath}assets/icons/${name}.svg` ``, the response is checked only for "parses as SVG", then assigned to `container.innerHTML`. Reject anything that is not `[a-z0-9_-]+` before it reaches `limel-icon`.
    - **Why:** A schema value of `"../../attachments/9f3c1/payload.svg?"` resolves to an arbitrary same-origin path — the trailing `?` pushes the appended `.svg` into the query string — so on a host serving user-uploaded files from the app origin, the fetched SVG is injected via `innerHTML`, where `<animate onbegin=…>` executes. It cannot reach a foreign origin. This is not a new class of exposure: `row.tsx:27-31` already feeds `lime.layout.icon` straight into `limel-icon`, and `fields/schema-field.ts:50-80` lets a schema render any registered custom element with any props.

11. **🟡 [Medium from Observability]** — A misspelled icon name renders nothing and logs a rejection that never names the icon
    *Pre-existing in `limel-icon`, but the exposure is widened by this PR. Recommended: Better as a follow-up PR — the fix is in the icon cache modules, which this PR does not touch.*
    - **Where:** `src/components/icon/icon.tsx:70-78`; `src/global/icon-cache/in-memory-icon-cache.ts:30-55`; `src/global/icon-cache/cache-storage-icon-cache.ts:33-42`
    - **What:** Add a `catch` in `loadIcon` that warns with the name and the resolved `assets/icons/<name>.svg` path.
    - **Why:** `loadIcon` awaits `loadSvg` with no `catch`, so a 404 surfaces as a bare `Uncaught (in promise) Error: Invalid SVG` with no icon name, the outer promise never resolves, and the `limel-icon` stays an empty container — and the cache-storage variant memoizes the rejected promise, making it sticky for the page's lifetime. `IconName` degrades to plain `string` unless the consuming app augments `IconNameRegistry`, so `lime: { icon: 'organisation' }` against the example's `organization` compiles fine.

12. **🟡 [Medium from Performance]** — Each icon-bearing option re-parses its SVG, because the CacheStorage icon cache memoizes nothing
    *Pre-existing, amplified by this PR. Recommended: Better as a follow-up PR — the fix is in an icon-cache module unrelated to what this PR touches.*
    - **Where:** `src/global/icon-cache/cache-storage-icon-cache.ts:21-33`, `:45-58`
    - **What:** Memoize the resolved string by name, as `InMemoryIconCache` already does at `in-memory-icon-cache.ts:19-25`.
    - **Why:** `get()` does a `cache.match()`, a `response.text()`, a `replaceAll` over the whole SVG string and a full `DOMParser` parse on every call, and `caches` is the branch taken in any secure context. Before this PR a mobile select with icons instantiated zero `limel-icon`s; now it instantiates one per option.

13. **🟡 [Medium from Performance]** — Option identity churn forces a full re-render and an O(N) list rebuild, now on mobile too
    *Pre-existing, extended to the mobile path by this PR. Recommended: Better as a follow-up PR — the fix is a caching change in the widget with its own correctness questions.*
    - **Where:** `src/components/form/widgets/select.ts:20-21`; `src/components/select/select.template.tsx:199`; `src/components/list/list.tsx:151-178`
    - **What:** Cache the mapped array against the `enumOptions` reference so `options` is not reassigned when nothing changed.
    - **Why:** `enumOptions.map(createOption)` builds a fresh array of fresh objects on every React render, firing all three `@Watch('options')` handlers and making `limel-list` tear down and reconstruct `MDCList` in a `setTimeout`. Typing in an unrelated field re-renders every widget, so a rich mobile select now pays an O(N) relayout per keystroke where it previously re-rendered a cheap native `<select>`. `createOption` adding two keys is not itself a cost — the object shape is more uniform now.

etc/lime-elements.api.md
Kiarokh: AI reviews added. Wanna fix a few of them before merging? :)

TommyLindh2:
Thanks @Kiarokh! I went through all 13 and picked the one I think genuinely belongs in this PR: **Top Recommendation 1**. Fixed in de4c00c2c.

### What was wrong

My second commit made "mobile" stop meaning "native `<select>`", but two methods still equated the two:

- `setMenuFocus()` (`select.tsx:294`) bailed out on `this.isMobileDevice`, so a phone rendering the custom dropdown got a list that nothing ever focused.
- `handleTypeaheadKey()` (`select.tsx:600`) did the same, and its comment stated the now-false premise outright.

Both now gate on `shouldRenderNative()` instead — the exact substitution the PR already makes at line 487. They're coupled, so fixing one without the other would have done nothing: typeahead stores a `pendingTypeaheadIndex` that only `setMenuFocus` consumes.

Two tests in the `typeahead` block pin it, using icon-bearing options plus `data-native`: one asserts the dropdown opens on typing, one asserts the match actually receives focus. Both fail if I revert `select.tsx` and keep the tests. The existing `leaves typeahead to the native dropdown on mobile` test stays green, since its options are text-only.

Note this also closes the same gap for multi-selects on mobile, which have always used the custom dropdown and never had focus moved into it.

### Why not the others

- **2 (perf)** — speculative. It's about a hypothetical 200-option form select; nothing in the codebase suggests that shape, and the fix it proposes (lazy-mounting `limel-list`) is a `limel-menu-surface` change well outside this PR.
- **4 (`description` opt-out)** — this is the deliberate design decision the PR is built on, already discussed above with @adrianschmidt and confirmed to have an empty blast radius across `lime-crm-components`, `lime-crm-building-blocks` and `lime-webclient`.
- **3, 5, 9** — reasonable polish, but each is a design discussion (naming `lime.option.icon`, `<optgroup>` support for separators, swapping in rjsf's `EnumOptionsType`) rather than a defect. Happy to take any of them if you'd like them here.
- **10–13** — all pre-existing, in `limel-icon` and the icon-cache modules this PR doesn't touch. Worth their own issues.
- **6, 7, 8** — changelog/commit-type bookkeeping. I'm not convinced by 7: `fix` vs `feat` is arguable, since the mobile commit does make a previously-unavailable combination work.

Let me know if you disagree on any of those and I'll pick them up.

TommyLindh2:
@Kiarokh now I've addressed the AIs comments and it's up-to-date with main and PR checks are green

Kiarokh:
> 🤖 AI-generated review from 6 parallel agents at commit `27bccf649`. Treat as input, not a verdict — agents can be wrong or miss context. Nothing in the PR has been changed by this review.

# Consolidated PR Review

## PR Summary

Review tier: Full — 376 changed lines across 9 files after noise filtering. Six reviewers ran: four on Opus (Backward Compatibility, Code Quality, Observability, Performance) and two on Fable (Architecture, Security), with a Fable coordinator.

The PR lets an alternative in a form schema's `oneOf` / `anyOf` carry two more pieces of a select option: its `description` becomes the option's `secondaryText`, and a new `lime.icon` key becomes its `icon`. A second commit makes `limel-select` keep its custom dropdown on mobile whenever any option has an icon, a secondary text or a primary component, since a native `<select>` cannot show those, and routes focus and typeahead through the same decision. 2 commits, 9 files, +337/−39 lines.

Points the author already answered in the earlier review round — the `lime.icon` naming and docstring, the `description` mapping, `EnumOption` versus rjsf's `EnumOptionsType`, separators and the new comment, the `feat` / `fix` typing, the changelog line and the auto-select test — have been dropped rather than repeated. What follows is what is left after that.

The PR is strictly better than `main`: the mapping is additive, the mobile change turns a silent drop into visible output, and the focus and typeahead gap flagged in the earlier review is closed in `de4c00c2c`, with two tests that fail without it. No High or Medium findings remain — the earlier round's items were either fixed or answered by the author with reasons, and what is left is a short list of minor nits under each dimension.

- **Blockers:** None
- **Non-blocking but worth addressing:** nothing that rises to a Top Recommendation; see the **Minor nits** under each dimension.

---

## Dimensions

Short summaries. No issue in this round points to a Top Recommendation.

The published surface is purely additive, and the one mobile behaviour change that reaches consumers was settled with the author in the previous round.

**What works well:** one optional property added in `etc/lime-elements.api.md`, nothing removed or narrowed; `setMenuFocus` and `handleTypeaheadKey` now both gate on `shouldRenderNative()`, so the High from the earlier review is genuinely closed; and both dropdown paths emit the original `Option` object, so consumers comparing by identity see no change when a select switches from native to menu.

**Issues:** None.

**Minor nits:**
- 🔵 A required mobile select loses its empty placeholder option once any option is rich, because `createMenuItems` filters it out while `NativeDropdown` rendered it — matches desktop behaviour, so not a regression (`src/components/select/select.template.tsx:350-368` vs `:238-256`).

### 2. Code Quality — GOOD ✅

A focused change with real tests; what is left is assertion polish in the new test blocks.

**What works well:** the whole mapping sits in `createOption`, the single point both single and multi selects pass through; the `renderOnMobile` helper collapses four copies of render boilerplate into a readable table of cases; and the two new mobile typeahead tests fail without the fix, so they pin real behaviour.

**Issues:** None.

**Minor nits:**
- 🔵 The four "falls back to the menu dropdown" cases only assert the native `<select>` is absent, so they would also pass if nothing rendered; asserting the custom trigger is present would close that (`src/components/select/select.e2e.tsx:472-543`).
- 🔵 No test asserts `select.options` for the plain-`enum` path the PR says is unaffected, though `enumSchema` already exists as a fixture (`src/components/form/form.e2e.tsx:126-157`, `form.test-schemas.ts:24-32`).
- 🔵 `secondaryText: undefined` and `icon: undefined` in the expected objects assert nothing, since `toEqual` ignores undefined-valued keys (`src/components/form/form.e2e.tsx:134-157`, `:168-184`).

### 3. Architecture — GOOD ✅

The layering is right — one conversion point in the widget, and the "can native show this?" decision inside `limel-select`.

**What works well:** `createOption` serves both the `oneOf` path and the `items.anyOf` path, so the two rjsf routes cannot drift; no form knowledge leaks into the select and no rendering knowledge leaks into the widget; and `shouldRenderNative()` is now the single gate for all four mobile-sensitive call sites, replacing three `isMobileDevice` checks that had drifted apart.

**Issues:** None.

**Minor nits:**
- 🔵 What a native `<select>` can render is defined twice — in `computeHasRichOptions` and in the template's `renderOption` — with nothing tying the lists together, so the next `Option` field has to be mirrored by hand (`src/components/select/select.tsx:755-762`, `select.template.tsx:260-275`).
- 🔵 The key is typed `IconName | Icon`, but every rendering path in `limel-select` rebuilds the icon as `{ name, color }`, so `Icon.title` and `Icon.backgroundColor` written in a schema are silently dropped — same as `Option.icon` today, so it follows convention (`src/components/select/select.template.tsx:303-349`, `:390-412`, `:416-441`).

### 4. Security — GOOD ✅

Both new schema-to-DOM paths are safe as rendered, and the PR opens no new class of exposure.

**What works well:** `description` reaches the DOM only as a JSX text child, never through `innerHTML`, a `title` attribute or a concatenated `aria-label`; icon colours go through Stencil style objects that set single CSS properties and cannot load external resources; and the pre-existing icon-name-to-URL path was already reachable through `lime.layout.icon` and `lime.component`, and the author has deferred it to its own issue.

**Issues:** None.

**Minor nits:**
- 🔵 The example labels a `readOnly: true` option "Can only be set by the support team", but `readOnly` only disables the row; the validator has no `readOnly` handling, so the value still validates if it arrives programmatically, and enforcement has to live server-side (`src/components/form/examples/builtin-field-types-schema.ts:113-121`).

### 5. Observability — GOOD ✅

The PR removes a silent failure on mobile: an icon or secondary text a consumer configured is rendered instead of dropped.

**What works well:** the rename to `hasRichOptions` plus the corrected comment in `handleTypeaheadKey` no longer asserts the now-false "mobile means native"; the form tests fail rather than lose an icon if `createOption` regresses; and no schema content or user input reaches the console.

**Issues:** None.

**Minor nits:**
- 🔵 `triggerIconColorWarning` runs only from `componentDidLoad`, so a deprecated `iconColor` arriving in a later `options` array is never reported — the `@Watch('options')` the PR adds sits right beside where it could run (`src/components/select/select.tsx:209-212`, `select.template.tsx:463-471`).

### 6. Performance — GOOD ✅

No regressions: the new check has the same cost profile as the one it replaces and short-circuits sooner.

**What works well:** `computeHasRichOptions` adds only property reads and stops at the first rich option, where the old predicate almost always scanned the whole list to return `false`; `getIconName` costs no bundle bytes since the template already imports it from the same chunk; and guarding `getFirstNativeAutoSelectOption` on `shouldRenderNative()` removes a spurious `change` emit and the re-render it triggered.

**Issues:** None.

---

## Top Recommendations

No significant issues found. Everything this round would otherwise have listed was either fixed in `de4c00c2c` or answered by the author in the previous round; the remaining observations are the Minor nits under each dimension.

Kiarokh (approved):

PR #4291, opened 2026-09-10
fix(select): show helper text while the dropdown is open

### LINKED ISSUE #4290: limel-select: show helper text before picking an item
## Current behavior

`limel-select`'s helper text was documented to show while the field is focused, but opening the dropdown never revealed it. Fixes three compounding issues: the CSS selectors that toggled its visibility were dead code left over from an old refactor, the dropdown's floating position never accounted for the helper text's height so it drew over it even once visible, and Popper.js positioned the dropdown before the helper text's open animation had run. 

1. …

## Environment

- **lime-elements** version: 
- Framework used: 
- Logs:

LucyChyzhova:
fix: https://github.com/Lundalogik/lime-elements/issues/4290

https://github.com/user-attachments/assets/a783bdea-8449-4c86-9c8a-4a17029518d6

Now matches `limel-picker`'s behavior: opening the dropdown shows the helper text and the dropdown drops below it.

As a side effect, `limel-portal` now also re-positions its content while its anchor is resizing (not just while its own content is), which benefits every other portal-anchored dropdown (menu, date-picker, dialog, tooltip).

## Summary by CodeRabbit

- **Bug Fixes**
  - Select dropdowns now remain correctly positioned when the field or helper text changes size, including during animations.
  - Dropdowns align with the complete select field and avoid sizing issues from the surrounding component.
  - Helper text styling now updates consistently with the select’s open state.
  - Native select menus now correctly collapse helper text when closed and expand it when opened.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


Kiarokh:
> 🤖 AI-generated review from 6 parallel agents. Treat as input, not a verdict — agents can be wrong or miss context.

*Generated by Claude Opus 5.*

# Consolidated PR Review

## PR Summary

Makes `limel-select`'s helper text appear while the dropdown is open (#4290), and stops the dropdown drawing over it. The branch has been rebuilt into 7 commits across `portal.tsx`, `select.tsx`, `select.template.tsx`, `_helper-text.scss` and `select.e2e.tsx`. Three of those commits fix pre-existing `limel-portal` problems that the main fix would otherwise have amplified — they are sequenced first, so the resize callback is made cheap and leak-free *before* it starts firing once per animation frame.

Strictly better than `main` in every dimension reviewed. No agent found a `[High]` regression, and backward compatibility came back SAFE with nothing above `[Low]`. What remains is test coverage and naming, plus two pre-existing architectural splits that are better addressed on their own.

- **Blockers:** None.

Two claims were checked by measurement in real Chromium rather than accepted from an agent, because both would have changed the verdict:

- **Refuted.** An agent argued the new wrapper adds a baseline gap (`limel-notched-outline` does compute to `inline-flex`), shifting *every* select's dropdown a few px lower. Measured: field, outline and trigger bottoms are all 40, with and without `helperText`. There is no gap.
- **Confirmed.** In a flex row with a 200px sibling, host bottom = 200 while field bottom = 40. The wrapper is genuinely required — anchoring to the host would be off by 160px, reintroducing #3858.

A third was measured after an agent flagged it, and the code was changed as a result: Blink serializes CSS lengths to six significant digits, so writing `421.328125px` reads back `421.328px`. An earlier version of the skip-unchanged-write guard compared against `container.style` and therefore never fired for `width`. It now tracks what was written. The same probe showed `-16px` is rejected by CSSOM outright, so a negative max height was being rewritten on every callback; it is now clamped at zero.

---

## Dimensions

Nothing public is removed or changed, and no `BREAKING CHANGE` footer is warranted.

**What works well:** no `@Prop`/`@Event`/`@Method`/exported type moves, and `etc/lime-elements.api.md` correctly needs no update; no `::part`/`exportparts` exists anywhere in the repo and every select selector uses descendant combinators, so the new wrapper breaks nothing; the pre-existing `:focus-within` behaviour is preserved, because the new rule sets the custom property *directly* on `limel-helper-line` and only under `--open`, while the mixin sets it as an inherited value; invalid-while-closed still expands, since `:host(limel-helper-line.invalid)` is unopposed; the observer lifecycle is strictly better than `main`, which built a fresh `ResizeObserver` on every `init()` and leaked the previous one.

**Issues:** None.

**Minor nits:**
- 🔵 `limel-select` now reflows on every open/close when `helperText` is set, growing the host by ~1rem and pushing surrounding content down. That is the point of the fix, but it is visible in dense forms and grids — worth a line in the release notes.
- 🔵 A `required` empty select keeps its helper text open and red from the first dropdown close onward. Consistent with what `chip-set` and `input-field` already do — select was the outlier — and documented in the commit body.
- 🔵 Still no `@Watch('anchor')`: a swap while open leaves both Popper and the observation on the old element until the next open. No consumer hits this today.

### 2. Code Quality — GOOD ✅

Clearly better than `main` on every axis; what remains is naming and a test gap.

**What works well:** commit messages state the mechanism, the rejected alternative and the blast radius of touching a `@private` component, and scoping to `select` is what `src/commits-and-prs.md:173-184` requires; dropping `.limel-select--invalid + limel-helper-line` is verifiably safe because `helper-line.scss:27-33` sets both vars directly on the element, beating the inherited `0fr`; `setContainerWidth`'s early return is behaviourally identical to the old nested `if`; the new tests match the file's existing helper-plus-why-comment style, and `expect(field).not.toBeNull()` forecloses a `null === null` pass.

**Issues:**
- **🟡 [Medium]** Nothing pins `observe(this.anchorElement)` (`src/components/portal/portal.tsx:349`). See **Top Recommendation 1**.
- **🟡 [Medium]** Nothing tests the validity change (`src/components/select/select.template.tsx:91`). See **Top Recommendation 2**.
- **🟡 [Medium]** `setContainerStyle` and `setContainerStyles` are adjacent, differ by one letter, and do unrelated jobs (`src/components/portal/portal.tsx:300-314`). See **Top Recommendation 3**.

**Minor nits:**
- 🔵 The anchoring test sits in `describe('helper text')` but asserts portal anchoring; its own comment explains it in terms of #3858.
- 🔵 `helperLineRows` reads `--limel-h-l-grid-template-rows`, an internal of the `@private` `limel-helper-line`. Sound for dodging the 0.46s race and it fails loudly on a rename, but worth knowing it would not catch the line being hidden some other way.
- 🔵 `is expanded while the dropdown is open` depends on `.click()` *not* moving focus — a real user click would satisfy the mixin's `:focus-within` rule and make it pass even with the fix reverted. Worth a line in the test so nobody "improves" it into a tautology.
- 🔵 `getContentWidth` can still yield the string `"nullpx"`. Reachability is identical to `main`, and the short-circuit means it is computed less often.
- 🔵 `limel-select--focused` on the trigger is still named for focus while meaning open; the PR fixed the naming only in the scss it had to touch.

### 3. Architecture — GOOD ✅

The shape is right: the three moving parts compose into one mechanism rather than three patches.

**What works well:** tying the observer's lifetime to the Popper instance holds up under a full lifecycle walk — observer connected ⟺ `popperInstance` non-null on every path, with no window where observation is live without a popper or disconnected while open; using `disconnect()` rather than `unobserve(anchorElement)` means a consumer that swapped `anchor` between opens cannot strand an observation; the anchor choice and the observation are causally linked, not independent; validity now has one owner; and the modifier-class-over-`:has()` choice matches the repo's own documented precedent in `style/internal/boolean-input.scss:115-135`.

**Issues:**
- **🟡 [Medium]** The declarative anchor path is one render behind, and `updatePortalAnchor()` is what hides it (`src/components/select/select.template.tsx:209-215`). See **Top Recommendation 4**.
- **🟡 [Medium]** The portal positions against the anchor but measures `this.host` (`src/components/portal/portal.tsx:264-283`, `:410-428`). See **Top Recommendation 5**.
- **🟡 [Medium]** The "open" signal is trapped in the select's shadow root, so `limel-text-editor` can reuse none of it (`src/components/select/partial-styles/_helper-text.scss:4`). See **Top Recommendation 6**.

**Minor nits:**
- 🔵 `limel-picker` has byte-for-byte the same `:host` rule that caused #3858 and also passes no anchor — the wrapper-as-anchor shape here is directly transplantable. Worth an issue.
- 🔵 Two classes now describe one state: `limel-select__field--open` and `limel-select--focused`.
- 🔵 `select.tsx:275` does a `querySelector` on every render; a `ref` would be cheaper, though it would not remove the one-render lag.
- 🔵 Pre-existing dead code the neighbourhood makes visible: `limel-select--with-helper-text` is matched by no selector anywhere, and `initialize()`'s `.mdc-floating-label` lookup never matches, so `MDCFloatingLabel`/`MDCSelectHelperText` are never constructed.
- 🔵 `animateHideAndCleanup()`'s 300ms timeout is never tracked or cleared, so a fast close→open→close lets the first timeout tear down mid-second-animation.

### 4. Security — SAFE ✅

No significant issues found.

**What works well:** no unsafe DOM APIs in any touched component, and every `querySelector` in the diff uses a static literal; `setContainerStyle` takes a compile-time union with two literal call sites and values derived from `getBoundingClientRect()`, so there is no property-injection surface; overlay integrity holds — every path that shows the container creates the popper first, and `animateHideAndCleanup` re-checks `visible` before tearing down, so `disconnect()` never outlives a visible dropdown; the anchor change drops only the trigger and the notched outline from the `closest('.limel-portal--parent')` walk, neither of which carries that class.

Both `[Low]` notes from the first pass are resolved rather than relabelled: the consumer-class concern is moot now that the `:has()` rule is gone (a consumer class would have to land on a `div` *ancestor* of the helper line, and `renderListComponent` requires a hyphenated tag), and the observer retention concern is closed by the lifecycle rework.

**Issues:** None.

**Minor nits:**
- 🔵 `setContainerStyles()` iterates consumer-supplied `containerStyle` keys with no allowlist. Pre-existing, not exploitable as used, and a string value makes `__proto__` assignment inert — it would only matter if a consumer ever bound `containerStyle` to untrusted data.

### 5. Observability — GOOD ✅

Every silent failure mode from the first pass is now pinned by a test or a real CSS rule.

**What works well:** `.limel-select__field` is referenced from three places and every edge between them now has a test, so a rename fails CI instead of silently mis-positioning the dropdown; the tests do not pass for the wrong reason — `is collapsed` doubles as a canary that the stylesheet is applied at all, and the expanded case cannot be satisfied by `:focus-within` or by the `.invalid` rule; `:has()` is genuinely gone from the select; and a dev-mode warning from `updatePortalAnchor()` would now be *off*-norm, since every `console.warn` in `src/` is a consumer-facing notice and that method legitimately no-ops in native/mobile mode.

**Issues:**
- **🟡 [Medium]** The one line delivering the third leg of the fix is untested. See **Top Recommendation 1**.

**Minor nits:**
- 🔵 The `ResizeObserver` loop mitigation is sound by inspection but unexercised — `runtime.spec.ts` never opens a dropdown, and the e2e tests return long before the 0.46s transition produces frames. If it regresses, the only signal is the browser's un-attributed loop error, which names neither component.
- 🔵 `setContainerStyles()` still writes the `containerStyle` prop unguarded on every callback. Harmless today because select's only entry is `z-index`, but nothing in the code says why it is exempt.

### 6. Performance — GOOD ✅

The observer change is a real improvement and its extra cost is bounded.

**What works well:** removing the recursive `getContentWidth` walk from the hot path matters more than the second rect read — `main` ran it on every callback for every `inheritParentWidth` consumer; the lifecycle walk finds no path where observation outlives the popper; the `disconnect()`+re-`observe()` churn per open is negligible since the observer object is reused; rounding `maxHeight` changes no consumer-visible dimension (±0.5px against 16px of slack); and the PR incidentally fixes a latent leak on `main`, where setting `visible` before `componentDidLoad` ran `init()` twice and orphaned an observer.

**Issues:**
- **🟡 [Medium]** The new anchor observation lands on all seven portal consumers, not just select (`src/components/portal/portal.tsx:349`). See **Top Recommendation 7**.

**Minor nits:**
- 🔵 `popperInstance.update()` is now the dominant per-frame cost (~56 calls per open, since a `maxHeight` write resizes the observed container and produces a second delivery in the same frame). Each runs `flip` and `preventOverflow`, both walking clipping ancestors with uncached rect reads. Deliberate, user-initiated and bounded; the cheap levers if it ever shows in a profile are a static `boundary` or suppressing `flip` during the animation.
- 🔵 `showContainer()` does not null-check `this.container` and is called from an uncancelled `requestAnimationFrame`, so a disconnect inside that one-frame window throws. Pre-existing.

---

## Top Recommendations

1. **🟡 [Medium from Observability + Code Quality]** — The line delivering the third leg of the fix is untested
   *Introduced by this PR. Recommended: Fix in this PR.*
   - **Where:** `src/components/portal/portal.tsx:349`
   - **What:** `this.observer.observe(this.anchorElement)` is what keeps Popper repositioning while the helper line animates open over 0.46s. Delete it and all three new e2e tests stay green — two read a CSS custom property, one reads `portal.anchor`, and none involves positioning.
   - **Why:** the failure is purely visual (the dropdown overlaps the helper line for the duration of the transition) with no console output and no failing test — exactly the shape of the original #4290 bug. It matters more than a single-component change because tooltip and the prosemirror action bar also pass explicit anchors that are now observed.

2. **🟡 [Medium from Code Quality]** — Nothing tests the validity change, the commit with the most consumer-visible effect
   *Introduced by this PR. Recommended: Fix in this PR.*
   - **Where:** `src/components/select/select.template.tsx:91`, `src/components/select/select.e2e.tsx`
   - **What:** two cheap additions — a `required` select whose dropdown has closed once asserts `limel-helper-line[invalid]` and `1fr`; an `invalid` select asserts the same while closed. The second pins the behaviour the deleted scss rule used to provide, which now comes from a different stylesheet entirely.
   - **Why:** the commit's own message calls the result "a layout change consumers do not opt into", and a grep for `invalid`/`required` across `select.e2e.tsx` returns nothing.

3. **🟡 [Medium from Code Quality]** — `setContainerStyle` and `setContainerStyles` are one letter apart and unrelated
   *Introduced by this PR. Recommended: Fix in this PR — it is a rename.*
   - **Where:** `src/components/portal/portal.tsx:300-304` and `:306-314`
   - **What:** the plural applies the `containerStyle` prop unguarded; the singular writes one guarded dimension. They sit adjacent and are both called from `styleContainer`. Rename one — `applyContainerStyleProp` or `setContainerDimension`.
   - **Why:** the plural reads like a batch version of the singular, which it is not. This is the kind of ambiguity that gets "fixed" in the wrong direction later.

4. **🟡 [Medium from Architecture]** — The declarative anchor is one render behind, and the imperative call is load-bearing for a sharper reason than the comment says
   *Pre-existing, retargeted by this PR. Recommended: Fix in this PR — reordering two JSX attributes.*
   - **Where:** `src/components/select/select.template.tsx:209-215`, `src/components/select/select.tsx:410-427`
   - **What:** `visible={props.isOpen}` is listed *before* `anchor={props.anchor}`. Stencil sets props in JSX key order and fires `@Watch` synchronously, so the first open reads `this.anchor` as the previous render left it — on the very first render, `this.host`. Put `anchor` above `visible`.
   - **Why:** only `componentDidLoad()` → `updatePortalAnchor()` currently prevents the first open from reproducing #3858, and the comment above `getAnchorElement` does not say so. The same latent ordering exists in `tooltip.tsx:128-137` and `prosemirror-adapter.tsx:380-386`.

5. **🟡 [Medium from Architecture + Code Quality]** — The portal positions against the anchor but measures the host
   *Pre-existing. Recommended: Better as a follow-up PR — unifying all three call sites is wider than this fix needs.*
   - **Where:** `src/components/portal/portal.tsx:264-283` and `:410-428` use `this.host`; `:321` and `:349` use `anchorElement`
   - **What:** unify on one reference element.
   - **Why:** for a select in a flex row with a 200px sibling the dropdown is placed at y≈40 but its `max-height` comes from `viewHeight - 200` — roughly 160px shorter than the space actually below it, so it scrolls earlier than it needs to. `limel-tooltip`, whose anchor can be anywhere on the page, is worse off. Not a regression — `main` had the identical split — but this PR touches these exact functions.

6. **🟡 [Medium from Architecture]** — The "open" signal is trapped in the select's shadow root
   *Introduced by this PR. Recommended: Better as a follow-up PR — it wants a shared-mixin design decision, and the wrapper is required as the anchor either way.*
   - **Where:** `src/components/select/partial-styles/_helper-text.scss:4`, `src/style/mixins.scss:373-398`
   - **What:** the shared `hide-helper-line-when-not-needed` mixin still expresses only "focused". A reflected host attribute would let it say `:host(#{$tag}:focus-within), :host(#{$tag}[open])` once.
   - **Why:** `limel-text-editor` has the identical `:focus-within`-vs-portal root cause — opening its link menu moves focus into a portal on `document.body` and collapses its helper line — and can reuse nothing from this fix. Worth an issue.

7. **🟡 [Medium from Performance]** — The anchor observation reaches six other components this PR does not exercise
   *Introduced by this PR. Recommended: Better as a follow-up PR — it wants portal test coverage, which does not exist yet.*
   - **Where:** `src/components/portal/portal.tsx:349`
   - **What:** `anchorElement` falls back to `this.host`. Menu, popover, date-picker, picker and input-field pass no anchor, so the new target is the `limel-portal` element — `position: absolute; top: 0; bottom: 0` inside a positioned host, so its rect tracks the consuming component's full height.
   - **Why:** any height change there — input-field's own helper-line animation, an auto-growing textarea, chips wrapping in picker — now drives a full restyle plus `popperInstance.update()` per frame while a dropdown is open, where `main` fired nothing. Arguably an improvement (the dropdown follows a growing field), but it is per-frame work added to six components with no portal test coverage: `src/components/portal/` contains only `contains.spec.ts`.

Kiarokh (approved):

Issue #4293, opened 2026-09-13
list, menu-list, card and info-tile have no language prop, so nested labels stay English


Kiarokh:
A batch of translation-forwarding bugs is being fixed in a separate PR — components that *have* a `language` prop but drop it before it reaches a child that translates its own labels. While auditing for those, a second class of the same problem turned up that cannot be fixed the same way.

These parents render a child that translates its own labels, but the parent has no `language` prop at all, so there is nothing to forward. The child therefore falls back to its `en` default and the label is permanently English no matter what language the consumer configured everywhere else. Fixing them means **adding a new public prop**, which makes each one a `feat` and a change to the public API rather than the one-line forward the other cases needed — hence this separate issue.

## Where

| Parent (no `language` prop) | Child | Label stuck in English |
| --- | --- | --- |
| `limel-list` — [list-renderer.tsx:148](https://github.com/Lundalogik/lime-elements/blob/0f35e1954c0df289a5b77f2c7f8e59f36da8441a/src/components/list/list-renderer.tsx#L148) | `limel-list-item` | `file-viewer.more-actions` → "More…" |
| `limel-menu-list` — [menu-list-renderer.tsx:133](https://github.com/Lundalogik/lime-elements/blob/0f35e1954c0df289a5b77f2c7f8e59f36da8441a/src/components/menu-list/menu-list-renderer.tsx#L133) | `limel-list-item` | `file-viewer.more-actions` → "More…" |
| `limel-card` — [card.tsx:358](https://github.com/Lundalogik/lime-elements/blob/0f35e1954c0df289a5b77f2c7f8e59f36da8441a/src/components/card/card.tsx#L358) | `limel-action-bar` | `action-bar` → "Action bar" |
| `limel-info-tile` — [info-tile.tsx:284](https://github.com/Lundalogik/lime-elements/blob/0f35e1954c0df289a5b77f2c7f8e59f36da8441a/src/components/info-tile/info-tile.tsx#L284) | `limel-linear-progress` | `loading` → "Loading…", `progress-bar` → "Progress bar" |

`limel-list-item` is `@private` and is reached only through `limel-list` and `limel-menu-list`, so both of those need the prop for its [action-menu label](https://github.com/Lundalogik/lime-elements/blob/0f35e1954c0df289a5b77f2c7f8e59f36da8441a/src/components/list-item/list-item.tsx#L347) to ever be translated.

## Impact

[...]

PR #4294, opened 2026-09-13
Translate the labels that were silently falling back to English

Kiarokh:
## Summary by CodeRabbit

- **New Features**
  - Improved localization across chips, loading indicators, code-diff actions, date pickers, and text-editor controls.
  - Added translated editor toolbar labels in supported languages.
  - Regional and case-variant language codes now resolve to the appropriate primary-language translation, with English fallback for unknown languages.

- **Bug Fixes**
  - Components now consistently use the configured language for localized labels, accessibility text, and date-picker formatting.

- **Tests**
  - Added coverage for language propagation, locale normalization, and translation fallback behavior.

## Why

Reported as *"some Norwegian consumers get the English translations"*, with a suspicion that the two Norwegian language codes were behind it. They are not: `nb` and `no` are already mapped — aliased in `translations.ts`, and converted in both directions in the date picker, which needs `no` for flatpickr and `nb` for moment — and all eight translation files have identical key sets. Two other things were at fault.

**1. A region subtag misses the lookup entirely.** `translate.get` matched the language verbatim, so only a bare primary subtag ever hit. A consumer passing a full BCP 47 tag — or deriving the language from `document.documentElement.lang`, which on a real page is usually `nb-NO` rather than `nb` — fell straight through to the English fallback, and *every* translated string in *every* component rendered in English. This is the one that best matches the original report, and it is worth checking what the affected consumers actually pass.

**2. `language` gets dropped on the way down the tree.** Six places either failed to forward it to a child that translates its own labels, or ignored it outright.

## What each commit fixes

| Commit | Problem | User-visible? |
| --- | --- | --- |
| `fix(translations)` | `nb-NO` and `SV-SE` both resolved to English | everywhere |
| `fix(date-picker)` | the `--today-label` tooltip was filled in by a `translate.get` call that never passed `language`, so it read "Today" in all eight languages despite every file defining `date-picker.today` | yes |
| `fix(text-editor)` (link menu) | `limel-text-editor-link-menu` rendered without `language`, so "Text", "Link", "Cancel" and "Save" stayed English | yes |
| `fix(chip-set)` | `getChipProps` omitted `language`, so every chip fell back to `en` — this is where `limel-file` loses the language it correctly forwards to the chip set | accessible name only |
| `fix(chip)` | the loading chip's `limel-linear-progress` got no `language` | accessible name only |
| `fix(text-editor)` (toolbar) | `accessibleLabel` was the hardcoded English string `"Toolbar"` | accessible name only |
| `fix(code-diff)` | the search bar's `limel-action-bar` got no `language` | accessible name only |
| `fix(date-picker)` (regional tags) | `getFlatpickrLang`, `getMomentLang` and `DateFormatter.getLanguage` compared the whole string, so `nb-NO` gave a Norwegian tooltip on an English calendar | yes |
| `test(translations)` | nothing checked that the eight translation files kept the same key set | n/a |

## Please pay attention to

- **The new `editor-menu.toolbar` translations are mine, not a translator's.** The `da`, `de`, `fi`, `fr`, `nl`, `no` and `sv` values would benefit from a native-speaker glance.
- ~~The subtag split accepts `_` as well as `-`~~ — it was over-defensive, and worse: blessing `no_NO` pointed consumers at a value `Intl.DateTimeFormat` rejects outright. Underscored tags are no longer accepted, same as on `main`.

## Testing

`eslint --max-warnings=0` and the full spec suite were run against **each commit individually**, not only at the tip — all seven are green on their own (1223 → 1225 tests as the new specs land), and `commitlint --from main --to HEAD` reports 0 problems. Both new component specs were confirmed to fail (`expected 'en' to be 'no'`) with their fix reverted, so they are real regression guards rather than tests that pass either way.

Not browser-tested by hand: every change here is a prop being passed on or a string being looked up, and the affected labels are mostly accessible names, which the spec environment asserts on directly.

## Related

Auditing for these turned up a second class of the same problem that needs a **new public prop** rather than a one-line forward, so it is out of scope here and tracked separately in #4293.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


Kiarokh:
follow up issue https://github.com/Lundalogik/lime-elements/issues/4293

Kiarokh:
*Generated by Claude Opus 5.*

> 🤖 AI-generated review from 6 parallel agents, re-issued after the findings it raised were addressed. Treat as input, not a verdict — agents can be wrong or miss context.

# Consolidated PR Review (updated)

The original review ran against `aeb6b7f6e`. Five of its recommendations have since been fixed on the branch, and the fixups have been autosquashed, so what follows is the updated picture against `68211d1b6`. The **Resolved** section records what changed; **Top Recommendations** now lists only what is still open.

## PR Summary

Fixes translation lookups that silently fell back to English: one change to the central `translate.get` resolution so a regional tag (`nb-NO`) matches on its primary subtag, six changes that forward an existing `language` prop to a child component or to a `translate.get` call that ignored it, and — added during review — the same normalization for the date picker's own flatpickr/moment locale plumbing plus a key-parity guard across the eight translation files. 9 commits, +178/−15 across 22 files, four new files (`src/util/language.ts` and three spec files).

The PR is strictly better than `main` in every dimension reviewed. The one genuine regression the original review found — a new crash path for a non-string `language` — is fixed, as is the mixed-language date picker the subtag change would otherwise have created. Each of the nine commits lints clean and passes the full spec suite on its own (1224 → 1235 tests as the new specs land), and the squashed tree is byte-identical to the verified one.

- **Blockers:** None.
- **Non-blocking but worth addressing:** four findings remain, all pre-existing and all recommended as follow-ups. See Top Recommendations at the end.

## Resolved since the original review

| Original finding | Fixed by | What changed |
| --- | --- | --- |
| Non-string `language` threw instead of falling back | a2bf14e9e | `getPrimarySubtag` now returns `''` for anything that is not a string. Spec covers `Intl.Locale`, `null`, a number and an object. |
| `nb-NO` gave a Norwegian tooltip on an English calendar | fa0c1b8e5 | `getFlatpickrLang`, `getMomentLang` and `DateFormatter.getLanguage` all reduce to the primary subtag, with a new `date-formatter.spec.ts`. `Intl.DateTimeFormat` deliberately keeps the full tag, so `en-GB` long dates stay British. |
| A test blessed `no_NO`, which `Intl` rejects outright | a2bf14e9e | `_` dropped as a separator entirely, rather than normalized downstream. An underscored tag is now simply unsupported, as on `main`. |
| Regional-tag support was undocumented and untyped | a2bf14e9e | TSDoc on `Languages` records that a full BCP 47 tag resolves at runtime. Widening the union is a `feat` and stays a follow-up. |
| Nothing checked that the eight translation files stayed in sync | 68211d1b6 | `it.each` parity test against `en`. Verified to catch drift: removing one key fails with `no defines exactly the keys English defines`. |

One structural note: `getPrimarySubtag` moved to `src/util/language.ts` so the date picker can share it without importing the translations module.

---

## Dimensions

No public API removed, changed, or added; correctly shipped as eight `fix` (PATCH) commits plus one `test`.

**What works well:** `src/components.d.ts` and the generated `readme.md` files are gitignored and `etc/lime-elements.api.md` is genuinely unaffected, so their absence from the diff is correct; `translate.get('editor-menu.toolbar', 'en')` returns `'Toolbar'`, byte-identical to the literal it replaced; all eight translation files verified key-identical — and now guarded by a test; `@private` components are correctly scoped to their public parents per `src/commits-and-prs.md`.

**Issues:** None.

**Minor nits:**
- 🔵 Accessible names and one visible tooltip change for localized consumers, shipped as PATCH. Nothing in this repo breaks — the axe baseline holds rule ids only, the one visual snapshot is unaffected, and no e2e asserts on these strings — but downstream suites selecting on accessible names could. Worth a release-note callout.
- 🔵 Issue #4293 is missing two components of the same class: `breadcrumbs.tsx:78` (`aria-label="Breadcrumb"`, no `language` prop) and `circular-progress.tsx:86` (`aria-label="%"`, no `language` prop). See **Top Recommendation 3**.

### 2. Code Quality — GOOD ✅

Small, correct, well-placed; the specs follow the repo's own established idiom rather than inventing one.

**What works well:** the exact-match lookup stays first, so the subtag path is purely additive and no previously-resolving tag changes meaning; `chip-set.tsx` places the forward in `getChipProps`, which both `renderChip` and `renderInputChip` share, so one line covers both paths and nothing can shadow it; `code-diff.tsx` correctly picks `translationLanguage` and not the syntax-highlighting `language` prop declared 350 lines above — an easy trap, avoided; both new component specs assert a non-default value, so they genuinely fail when the fix is reverted.

**Issues:**
- **🟡 [Medium]** Toolbar item labels are computed once in `componentWillLoad` while the new `accessibleLabel` is reactive (`prosemirror-adapter.tsx:301`). See **Top Recommendation 1**.

**Minor nits:**
- 🔵 The two `limel-action-bar` call sites in this PR get opposite strategies — the editor passes a translated `accessibleLabel` but no `language`; code-diff passes `language` but no label, so its search bar announces the generic "Action bar" despite the commit being titled *translate the accessible label of the search toolbar* (`prosemirror-adapter.tsx:364`, `code-diff.tsx:452`).
- 🔵 `getChipProps` returns an untyped object spread as `{...chipProps}`, so TypeScript performs no excess-property check — a typo like `languange:` would compile and silently do nothing (`chip-set.tsx:654`).
- 🔵 `chip-set`'s `language` JSDoc still says "For example, the clear all chips label", now an understatement (`chip-set.tsx:218`).
- 🔵 `type="input"` in the chip-set spec builds an `MDCTextField` for no benefit; both component specs assert the forwarded prop rather than the resulting label.
- 🔵 Four of the translation-forwarding commits ship untested; the flatpickr `--today-label` one is a one-line assertion on the container's `style` attribute.

### 3. Architecture — GOOD ✅

The one genuinely architectural change lands at the single correct chokepoint, and the review round improved on it.

**What works well:** `getTranslationsFor` is the funnel every localized string passes through, so a three-line change fixes all 22 calling components at once — including `limel-ai-avatar`, which defaults `language` to `document.documentElement.lang` and was therefore silently English on every `<html lang="nb-NO">` page, an in-repo bug fixed as a side effect. `editor-link-menu` had a `language` prop that was never fed by anyone, so the PR turns dead surface into live surface. Extracting `getPrimarySubtag` into `src/util/language.ts` resolved the original review's "two homes for language normalization" finding: the date picker now shares one helper instead of carrying its own whole-string comparisons.

**Issues:** None.

**Minor nits:**
- 🔵 Two remaining "has `language`, doesn't forward it" sites: `action-bar.tsx:200` → overflow menu (blocked by #4293) and `profile-picture.tsx:383` → callout (masked today because `heading` is always supplied).
- 🔵 Useful for the #4293 discussion: `globalConfig.defaultLocale = navigator.language` already exists at `config.ts:35`, and `navigator.language` and `document.documentElement.lang` are both full tags — so `getPrimarySubtag` is precisely the missing piece that makes a shared config default viable. This PR moves *toward* that answer rather than away from it.
- 🔵 `Languages` lives under `date-picker` while 32 files import it; this PR adds no new importers, so it neither reinforces nor relieves that.
- 🔵 `limel-chip` is now internally inconsistent in this dimension: `renderSpinner` localizes correctly, `renderProgressBar` hand-rolls `aria-label="%"` (`chip.tsx:451`).

### 4. Security — SAFE ✅

No new security issue introduced; the three interesting sinks were each verified rather than eyeballed.

**What works well:** CSS injection is not reachable — `--today-label` contains a hyphen, so Stencil routes it through `style.setProperty` rather than `cssText`, and `setProperty` invalidates the whole declaration on an unbalanced quote instead of escaping into sibling rules; all eight `date-picker.today` values were checked and none contains `"`, `;`, `\` or a newline. `accessibleLabel` lands on `aria-label` via the vdom attribute path, no `innerHTML`. The prototype-shaped lookup was modelled empirically on both branches against every `Object.prototype` key — every input returns the correct English string on both, because the pre-existing per-key fallback absorbs it.

**Issues:** None.

**Minor nits:**
- 🔵 Both lookups still walk the prototype chain. Harmless today and verified so, but a latent trap: the day someone adds a key called `name`, `length` or `constructor` to `en.ts`, `translate.get(key, 'CONSTRUCTOR')` starts returning `"Object"` or `1`, and the `1` case throws at `.replaceAll`. A one-expression guard (`Object.hasOwn`, or a null-prototype object) would close it (`translations.ts:50`).

### 5. Observability — GOOD ✅

A net improvement — it converts a family of silent English fallbacks into correct behaviour, adds tests that fail legibly, and the review round added the drift guard that was the biggest remaining gap.

**What works well:** the new tests pin the input shapes that previously failed silently, with comments recording why each exists; both component specs guard with `expect(el).not.toBeFalsy()` before dereferencing and assert a *non-default* value, so they cannot silently pass the way `text-editor.spec.tsx:55` can (it asserts `.language === 'en'`, which is also the child's own default). On whether to add logging: **no** — the repo's ~33 `console.warn` calls are all for developer misconfiguration, none for runtime fallbacks, and a warn-on-unknown-language would fire every render.

**Issues:**
- **🟡 [Medium]** An unknown language makes the date picker render **Swedish**, not English (`date-formatter.ts:1-8`). See **Top Recommendation 2**.
- **🟡 [Medium]** Three forwarding gaps remain where the parent exposes no `language` prop at all (`card.tsx:358`, `info-tile.tsx:284`, `list-renderer.tsx:148`). See **Top Recommendation 3**.

**Minor nits:**
- 🔵 The `try`/`catch` at `callout.helpers.ts:41` was belt-and-braces against the old `nb` `TypeError`. Now that `getPrimarySubtag` cannot throw, it is unreachable while still able to swallow a genuine future error.
- 🔵 `limel-code-diff` and `limel-code-editor` overload the name `language` to mean *syntax highlighting*, and it is untyped (`string`) — so a consumer following the library-wide convention gets no type error, no warning, no highlighting and no translation. This PR's premise makes that collision more likely to be hit (`code-diff.tsx:102`).
- 🔵 A custom ESLint rule to catch the next missed forward is not warranted — the repo consumes a shared `@limetech/eslint-config`, so a local rule would need an upstream change. The proportionate guards are the parity test (now present) and a note in the authoring guide.

### 6. Performance — SAFE ✅

No regression on any path a real consumer exercises.

**What works well:** the fast path is genuinely untouched — `getTranslationsFor` short-circuits, so for every language code the library declares, `getPrimarySubtag` never runs; measured end-to-end, `translate.get(…, 'sv')` is 50 ns on `main` vs 53 ns here, within noise. Stencil compares props by value and `language` is a stable string, so the new props trigger **zero** extra re-renders. `'Toolbar'.replaceAll(REGEX, fn)` with no merge code returns the identical string reference, so the pre-existing `replaceAll` allocates nothing for parameterless keys. Bundle grows by 342 bytes raw, **129 bytes gzipped**.

**Issues:**
- **🟡 [Medium]** `getSearchActions()` rebuilds its array on every render, forcing an action-bar re-render per keystroke (`code-diff.tsx:495`). See **Top Recommendation 4**.

**Minor nits:**
- 🔵 Dropping `_` from the separator split also removed the regex: the helper now uses `split('-')`, which is the faster form measured (regex split was 38 ns/op vs 16 ns/op for the non-regex path). Worth recording that **hoisting a regex to module scope makes `String.split` slower**, not faster — V8 has a fast path for an inline literal that a hoisted `RegExp` defeats.
- 🔵 Recommend explicitly **against** memoizing the resolution: a `Map` would add ~15–25 ns to the 99% hit path to speed up the 1% case, keyed by attacker-controllable attribute strings.
- 🔵 `updateActiveActionBarItems` deep-clones 12 menu items per selection change; the new `translate.get` is one call on top of 12 where `cloneDeep` dominates by ~100× (`prosemirror-adapter.tsx:451`).

---

## Top Recommendations

All four remaining findings are pre-existing and none blocks merge.

1. **🟡 [Medium from Code Quality]** — Toolbar item labels don't react to a language change, while the new accessible label does
   *Pre-existing, newly surfaced by this PR. Recommended: Better as a follow-up PR — no `@Watch('language')` exists anywhere in the repo, so this is a library-wide pattern decision rather than a local fix.*
   - **Where:** `src/components/text-editor/prosemirror-adapter/prosemirror-adapter.tsx:301-303`, `:364-372`
   - **What:** `@Watch('language') private handleLanguageChange() { this.getActionBarItems(); }`
   - **Why:** `getActionBarItems()` runs only in `componentWillLoad`. The new `accessibleLabel` sits in `renderToolbar()` and *does* re-evaluate on a language change, while the button labels beside it stay in the old language. Previously both were frozen, so the inconsistency is new even though the root defect is not.

2. **🟡 [Medium from Observability]** — An unknown language makes the date picker render Swedish, not English
   *Pre-existing. Recommended: Better as a follow-up PR — it is a separate correctness question about moment's global state, and the subtag normalization already narrows how often it is reachable.*
   - **Where:** `src/components/date-picker/date-formatter.ts:1-8`, `src/components/date-picker/pickers/picker.ts:108-112`
   - **What:** Add an explicit `moment.locale('en')` after the locale imports.
   - **Why:** `moment/locale/*.js` sets the *global* moment locale as a side effect of loading, and the import block ends with `import 'moment/locale/sv'`. Reproduced with the repo's exact import order: `moment.locale()` is `'sv'` afterwards, so any unrecognised tag formats dates in Swedish. The `'no' → 'nb'` special case exists precisely to dodge this for one value. A silently *wrong* language is worse than a silent English fallback — and it is the most severe surviving instance of the class this PR exists to fix.

3. **🟡 [Medium from Observability, Backward Compatibility]** — Forwarding gaps remain where the parent has no `language` prop at all
   *Pre-existing. Recommended: Better as a follow-up PR — already tracked in #4293; this is a request to widen that issue's scope.*
   - **Where:** `card.tsx:358`, `info-tile.tsx:284`, `list-renderer.tsx:148`, `menu-list-renderer.tsx:133`, plus `breadcrumbs.tsx:78` and `circular-progress.tsx:86`, which are **not** currently listed in #4293.
   - **What:** Add breadcrumbs and circular-progress to #4293 so the follow-up `feat` covers the whole class.
   - **Why:** These are worse than the ones this PR fixed, because the parent exposes no `language` prop — a consumer cannot work around them at all. Worth noting that `list-renderer`'s `{...attributes}` spread only ever carries `tabindex`, so it is not a hidden escape hatch.

4. **🟡 [Medium from Performance]** — `getSearchActions()` rebuilds its array on every render, forcing an action-bar re-render per keystroke
   *Pre-existing. Recommended: Better as a follow-up PR — unrelated to what this PR changes on that line.*
   - **Where:** `src/components/code-diff/code-diff.tsx:453`, `:495-518`
   - **What:** Cache the array keyed on `totalSearchMatches === 0` plus `translationLanguage`.
   - **Why:** A fresh array of fresh objects means the `actions` prop reference always differs, so `limel-action-bar` always re-renders — and since `searchTerm` is `@State`, that fires on every keystroke alongside a full rebuild of every diff line. The new `language` prop is a stable string and adds nothing; this is simply the real cost sitting on the line the PR edits.

src/components/date-picker/date.types.ts
```diff
@@ -11,6 +11,11 @@ export type DateType =
     | 'year';
 
 /**
+ * The languages the components can be translated into.
+ *
+ * At runtime a full BCP 47 tag, such as `nb-NO`, also resolves: it is matched
+ * on its primary subtag. Cast to this type when passing one on.
+ *
```
devbymadde, line 18: This comment causes the build to fail because it counts as an API change (going from undocumented), so either run `npm run api:update` and check in the api.md file, or just remove this comment.

devbymadde (commented):
Looks good, ready to approve once this comment below is resolved :)

src/components/date-picker/date.types.ts
Kiarokh: AH thanks. I didn't see this before. Will fix ;)

Kiarokh: fixup is pushed

devbymadde (approved):

PR #4295, opened 2026-09-16
fix(picker): stop the dropdown resizing itself inside its own resize observer

FredrikWallstrom:
A picker dropdown is laid out at its content's width first, then the portal's `ResizeObserver` callback writes the anchor's width onto that same container — resizing an observed element from inside its own notification. The browser reports `ResizeObserver loop completed with undelivered notifications`, which the web client logs as an uncaught error and forwards to Sentry.

Skipping writes that change nothing doesn't cover it: that first width is genuinely new, so there is nothing to skip. Restyling on the next frame keeps the write out of the delivery that asked for it.

Verified by serving this build into the web client in place of the CDN's 40.2.5 and driving a column filter's coworker picker: 1 error before, 0 after.

Fixes Lundalogik/crm-client#1269

## Review:

### Browsers tested:
macOS:
- [x] Chrome
- [x] Firefox
- [x] Safari


Befkadu1 (approved):
Awesome ⚡

Issue #4298, opened 2026-09-16
Pagination: jump to a page from the ellipsis


Kiarokh:
*Follow-up to #4296. The design is mine; the scoping problem described below was found in a review session with Claude Opus 5.*

## New feature motivation

`limel-pagination` renders a fixed window of page numbers, so a large set is mostly gaps: `‹ 1 ··· 246 [247] 248 ··· 492 ›`. Prev and next move one step, and the first and last pages are always rendered as numbers, so the edges are one click away — but everything in between is not reachable in a single action. Getting to page 300 of 492 means walking there.

In the first version the `···` is an inert gap marker. This issue makes it the jump control.

## New feature description

The `···` becomes a button. Clicking it opens a small popover holding a numeric field and a `Go` submit. There is no always-visible "go to page" field: at rest the control stays prev · numbers · next and nothing else, which is the whole reason the jump lives behind the gap marker.

- Focus moves into the field on open.
- **Enter** commits, **Esc** closes and returns focus to the ellipsis, outside mousedown dismisses.
- Non-numeric characters are stripped as typed; out-of-range values clamp rather than error.
- The ellipsis carries `aria-expanded`, and its label is simply `Go to page`.

**Both ellipses open the same, unscoped control.** This is the part worth recording, because the obvious design is wrong: an earlier draft scoped each gap to the range it hides, with a header reading `PAGES 2–245`. The field still accepted any number and clamped to the whole set, so the two ellipses were functionally identical while being presented as two different controls — and the header was describing a restriction that did not exist.

Making the scoping real is worse than dropping it. If the left ellipsis accepted only `2–244`, a user who wants page 300 and clicks the nearer of two identical-looking buttons gets silently bounced to 244. Nobody can see which gap "owns" which range, so there is no way to pick correctly. Unscoped is the only honest option, and it has the side benefit that the label stops being the unwieldy `Jump to a page between 2 and 245`.

## Open question: is a field enough?

A bare number input assumes the user knows the number they want. That covers "it was on page 12", but not "get me roughly two thirds of the way in".

Options considered:

1. **Field only.** Smallest, and the two jobs people actually have are arguably covered already — *I know the number* by the field, *take me to an end* by the first and last pages, which are always rendered.
2. **Field plus buttons for pages in the hidden range.** The hidden range can be 243 pages wide, so any subset is an arbitrary choice the component is not qualified to make, and the full set needs its own pagination inside a popover, which is absurd.
[...]

PR #4299, opened 2026-09-16
Pagination component

### LINKED ISSUE #4296: Add a `limel-pagination` component
*The design (variants, states, interaction, a11y annotations) is mine. The API corrections under "New feature implementation" came out of a review session with Claude Opus 5 against the current `lime-elements` source and the CRM web client's existing pagination paths.*

## New feature motivation

`lime-elements` has no pagination primitive. The only component that paginates is `limel-table`, and it does so through Tabulator rather than through anything reusable — so every other surface that outgrows its first batch of rows ends in a "Load more" button instead.

In the Lime CRM web client that means list views, related-object tables on object cards, activity feeds and kanban columns all solve paging locally or not at all. The cost shows up in two ways: a user who wants row 400 has to click and scroll their way there with no option to jump, and — worse — leaving the view and coming back resets it to the first batch, so they press "Load more" repeatedly just to return to where they already were.

A shared page control fixes both, and gives consumers a single place to put position state (typically a URL query param) instead of re-deriving it.

Umbrella issue, with the surface-by-surface rollout: Lundalogik/crm-client#1320

## New feature description

`limel-pagination` reports where the user is in a result set and moves them somewhere else. It does **not** fetch data, own rows-per-page, touch the URL, or scroll anything. It emits one event and the consumer decides.

### Props

| Prop | Type | Default | Notes |
|---|---|---|---|
| `page` | `number` | `1` | Current page, 1-based. Controlled if set, uncontrolled if omitted. |
| `pageSize` | `number` | `25` | Rows per page. Read-only to the component. |
| `totalItems` | `number \| null` | `null` | Row count; the page count is derived from it and `pageSize`. `null` means *not known yet* — see the implementation notes, and #4297 for sources that can never know. |
| `loading` | `boolean` | `false` | Disables interaction, keeps layout, sets `aria-busy`. |
| `language` | `Languages` | `'en'` | Drives both the labels and the number formatting. |

There is no prop for the `21–40 of 248` range and no inline readout: the range lives in each page button's tooltip, as `helperLabel`. That keeps the control to one row of targets and nothing else, and leaves the horizontal space for the pages themselves. If a surface turns out to need the range on permanent display, an inline variant of it can be added later without changing anything here.

### Event

`changePage` → `{ page: number, pageSize: number, offset: number, limit: number }`

`offset` and `limit` are on the payload so consumers never redo the arithmetic — this is exactly the shape the CRM's load options already take. Emitted only on an actual change; clicking the current page emits nothing. **Name collision to resolve before implementing — see below.**

### Visual form

The first version renders one form and takes no prop to choose it. Two more are designed and wanted, but they are a visual axis over an identical model, so they arrive together with the prop that selects them rather than ahead of it — a `variant` prop with one legal value is worse than no prop at all.

- **`numbers`** — **the v1 form.** `‹ 1 ··· 245 246 [247] 248 ··· 492 ›`. One sibling each side, 5 page slots; first and last page always present as numbers. Full-page list views and any container wider than ~480px.
- **`counter`** — `‹ 3 / 13 ›`. Three fixed targets, no truncation logic to reason about. Mobile and very narrow containers; the touch-primary variant, and the one where the tooltips are least useful since touch has no hover.
- **`dots`** — `· · – · · ·`. Two siblings each side, 7 slots, current page as a stretched bar rather than a larger dot. Hover or focus any dot for its number, through the same `limel-tooltip` the `numbers` variant uses; the dots are real buttons, so it is a jump control, not an indicator. Narrow cards, side panels, feeds.

Later, below ~480px of **container** width (container query, not viewport) `numbers` will fall back to `counter`. The same table is narrow on a card and wide on a list page, and the consumer should not have to care which.

Caveat to document for `dots`: the dots show position within the *window*, not the whole set, so at 100 pages they are not a proportional scrubber. The tooltip on each dot carries the absolute truth.

### The gap marker

Where the window skips pages, a `···` marks the gap. In this version it is **inert** — a rendered marker, `aria-hidden`, not a button and not a tab stop. Turning it into a jump-to-page control is #4298; it is the least settled part of the design and the only place that would introduce a text field and focus management, so it is not what the first version should be arguing about.

Everything reachable without it: prev, next, the pages in the window, and the first and last pages, which are always rendered as numbers.

### States

| State | Behaviour |
|---|---|
| Many pages | Fixed slot count, square min-width and tabular numerals, so 9 → 10 does not reflow. |
| Count pending | `totalItems: null` while rows are already rendered. The last known page count is held if there is one; otherwise the navigation renders inert rather than absent. Never mistaken for empty or single-page. |
| Loading | Controls disabled, geometry unchanged, one teal 8px dot pulses. No overlay, no collapse — the row keeps its height so the table below does not jump. Root gets `aria-busy="true"`. |
| Empty, or a single page | Renders `‹ 1 ›` with both arrows dead, and keeps its height. It does **not** hide itself: a component that removes itself takes the decision away from the consumer rather than leaving it with them, and it does so at the worst moment — when a filter has just narrowed the results — shifting whatever sits below it. A consumer who would rather show nothing renders nothing. Saying *why* a list is empty is still theirs, since they have the filter context and this does not. |
| Fewer pages than the window | No gaps; degrades to exactly what is needed. |

Clamping: a `page` beyond the last clamps instead of rendering blank; if `totalItems` shrinks under a filter while the user sits on page 9, clamp and emit once; a `pageSize` change re-derives the page count on the same render.

### Structure

A flat `nav` of buttons. No list, no exposed parts.

```html
<nav aria-label="Pagination">
  <button class="arrow" aria-label="Previous page" aria-disabled="true"><span aria-hidden="true"></span></button>
  <button id="a_x-1" class="page" aria-current="page">1</button>
  <button id="a_x-2" class="page">2</button>
  <span class="gap" aria-hidden="true">…</span>
  <button id="a_x-9" class="page">9</button>
  <button class="arrow" aria-label="Next page"><span aria-hidden="true"></span></button>
</nav>
```

**No `ul`/`li` wrapper.** A list would announce "list, 5 items" — but 5 is the *window* size, not the page count, and it counts the ellipses too. For a 492-page set that is confidently wrong, which is worse than silent. Everything that carries real meaning is already there: the region is labelled, every button has a full label, and `aria-current` marks position.

**And not `role="listitem"` on the buttons either.** `role` replaces the implicit role rather than adding to it, so `<button role="listitem">` stops being exposed as a button at all, and a `listitem` outside an owning `list` means nothing. Real `li` elements are the only way to have both — which is moot given the point above.

The arrows hold an empty `<span aria-hidden="true">` for the triangle, drawn in CSS the way `limel-table`'s Tabulator paginator already draws its own. No icon component: `lime-icons8` is not distributed to consumers.

### Accessibility

- `<nav aria-label="Pagination">` around prev / pages / next, `aria-current="page"` on the current page.
- **Plain tab stops — every button is individually tabbable.** No roving tabindex, no arrow keys, no Home/End.
- Every target has a full label: `Go to page 7`, `Page 7, current page`, `Previous page`, `Next page`.
- Every page button is described by a `limel-tooltip`: `label` is `Page 3` and `helperLabel` is that page's range, `41–60 of 248`. The tooltip wires `aria-describedby` onto the button itself, and renders through a portal so it is not clipped by an ancestor's `overflow: hidden`.
- A page change is announced by a **visually hidden `aria-live="polite"` region** carrying `101–120 of 248`. A tooltip is `aria-describedby` and `aria-hidden` when closed, so it can never do this job, and with no inline readout the live region is the only thing that reports the move. It is not optional.
- Prev/next at the ends use `aria-disabled` plus a no-op click guard rather than the native `disabled` attribute, so they stay focusable and findable.
- Prev/next are **44 × 44px** targets wrapping a 36px visual, with negative margins absorbing the extra so row height stays 36px.
- `prefers-reduced-motion: reduce` drops the transforms, keeping the colour and shadow changes.

#### Why not roving tabindex

Roving belongs to **composite widgets**, where the children are options within a single control — tabs, radio groups, menus, listboxes, grids. A pagination component is a **navigation region**: each button is an independent destination. The library already splits on exactly this line, with `limel-tab-bar` roving and `limel-breadcrumbs` not.

Three things decide it:

- **Discoverability.** Under roving a keyboard user tabs in, lands on the current page, and nothing on screen says "now press the arrow keys". The reasonable conclusion is that the other pages are not keyboard-reachable at all.
- **Convention.** Every paginated site a user has already met makes each page individually tabbable. Breaking that prior costs more than the elegance buys.
- **Home/End are not ours to take.** They scroll the page today and people rely on it.

What this gives up is real and worth recording: roving would make the component one tab stop, and returning to it would land focus on the current page. Plain tab stops mean up to seven stops and landing on the prev arrow. That is an acceptable price at this window size, at the end of a table the user is usually leaving anyway — and the decision is easy to reverse in the additive direction, whereas removing roving after people have learned it is not.

Screen reader users are unaffected either way: the button rotor lists every button regardless of `tabindex`.
### Deliberately not props

`totalPages` (derivable from `totalItems` + `pageSize`; a third number is only a way to disagree with yourself) · rows-per-page picker (belongs next to the table's other settings, or to a user preference) · `siblingCount` (window size is a constant per variant, and picking a variant is already the statement about available room) · `hideOnSinglePage` (a single page always hides the nav) · `density` (no other `lime-elements` component takes one, and a smaller pagination component would forfeit the 44px touch minimum) · `unknownTotal` (absence of a count already means this) · first/last buttons (page 1 and the last page are always rendered as numbers, so the jumps exist as labels that say where they go).

## New feature implementation

Four things to settle before writing code. The first is an API decision that affects existing consumers.

### 1. `changePage` collides with `limel-table`

`limel-table` already declares `@Event() changePage: EventEmitter<number>`. Shipping a second `changePage` with an object payload gives us two components in one library emitting the same event name with incompatible `detail` types — and consumers that host both (the CRM's table view does) get a real footgun from event bubbling.

Options: widen the table's payload to the same object (breaking), give the pagination component a different name, or namespace one of them. I do not think this should be decided unilaterally in the pagination component PR — flagging it for a maintainer call.

### 2. Rows and totals do not arrive together

A common real case is a consumer that fetches the rows and the count as two separate requests and renders as soon as the rows land — a deliberate optimisation, since the `COUNT` is the expensive half. So the component will routinely hold rows and a page number while `totalItems` is still `null`.

It must not treat that as an empty or single-page result. Zeroing the count mid-load collapses the control to one page and snaps it to page one, and re-deriving geometry from a missing count thrashes layout on every filter change. `null` means *not known yet*: hold the last known page count if there is one, and otherwise render the navigation inert rather than absent, so the row does not appear and disappear between requests.

This is also why sources that can *never* be counted are deliberately not in scope here — `null` would then carry two incompatible meanings. That is #4297, and the discriminator is its first design problem, not this issue's.

### 3. Labels and number formatting go through the translations module

`lime-elements` has its own translations (`src/translations/{da,de,en,fi,fr,nl,no,sv}.ts`) with components taking `@Prop() language` and reading through a private `getTranslation(key)`. `Previous page`, `Next page`, `Go to page { page }`, `Page { page }, current page`, `Jump to a page between { from } and { to }`, and the range string `{ from }–{ to } of { total }` all belong there, not in props.

The same applies to the numbers. The prototype formats with a hardcoded `sv-SE` (`4 921–4 940 of 9 840`); grouping separators are not a property of the component, they are a property of the tenant. Derive the `Intl.NumberFormat` locale from `language`.

There are no label props to override any of this. The component owns no message that a consumer would need to reword — the empty and loading *states* are the consumer's to explain, and everything the pagination component says is about pages, which it knows better than its consumer does.

### 4. Smaller points

- **When the variants land, `variant` vs `type` needs a decision.** Only `limel-ai-avatar` uses `variant` today; `type` is the established name on chip, chip-set, callout, list, list-item, input-field, date-picker and chart. I lean `variant`, keeping `layout` free for alignment (left / center / split) later — but it is a deviation, not the convention, and should be a conscious one. Not a v1 problem, since v1 has no such prop.
- **`pageSize` default.** `25` rather than `20` — nothing in the CRM uses 20, and 25 is the smaller of the batch sizes actually in use.

### Scope and build order

**The first version is the `numbers` form, for sources that can count, and nothing else.** Everything held back is held back for the same reason: it widens the API before anything has used the narrow one. A `variant` prop with a single legal value, or a second count-mode with no consumer yet, are both easier to add later than to change once published.

In scope for v1:

1. Model + `changePage` (window maths, clamping, controlled/uncontrolled).
2. ARIA, labels and the visually hidden live region.
3. Page-button tooltips (`limel-tooltip`) carrying the page number and its range.
4. Loading, empty and single-page states.

Follow-up, separately reviewable:

- `counter` and `dots`, with the `variant` prop that selects them, and the container-query fallback.
- Jump to a page from the ellipsis — #4298.
- Sources with no total count — #4297.

An interactive design prototype (all three forms and every state, with the jump popover and keyboard behaviour working) exists and I will attach it to this issue.

LucyChyzhova:
fix https://github.com/Lundalogik/lime-elements/issues/4296

Report where the user is in a set of results, and move them somewhere else in it. The component loads nothing: it emits the page the user asked for, along with the offset and limit that page needs, and the consumer decides what that means.

Landing it whole rather than in parts, since a new component is not useful to review without the examples and tests that show what it does.

This first version covers the numbers form for sources that can count. Held back deliberately, each because it would widen the API before anything has used the narrow one: the counter and dots variants along with the `variant` property that selects them, jump-to-page from the ellipsis, and support for sources that can never report a total.

The page count is derived from `totalItems` and `pageSize`; there is no `totalPages` property, because two numbers already determine the third and a third way to say it is only a way to disagree with yourself. Rows per page is read-only here — it belongs beside whatever else a consumer lets people configure, not inside a navigation control.

The first and last pages are always rendered, so both ends of the set are one click away without separate first and last buttons, which would have to be inferred from a glyph rather than read as a number.

Plain tab stops rather than a roving tabindex. Roving belongs to composite widgets, where the children are options within one control; this is a navigation region whose buttons are independent destinations, and the library already splits on that line with limel-tab-bar roving and limel-breadcrumbs not. Under roving, nothing on screen tells a keyboard user to press the arrow keys, so the reasonable conclusion is that the other pages cannot be reached at all.

The range each page holds is carried by its tooltip rather than by a readout beside the navigation, which keeps the control to one row of targets. Tooltip ids are per slot rather than per page number, because limel-tooltip resolves its owner element once when it connects and never looks again — a page number moves between slots as the user navigates, a slot does not.

## Summary by CodeRabbit

* **New Features**
  * Added a pagination component with page navigation, loading states, page-size selection, item counts, and accessible status announcements.
  * Added support for localized pagination labels in Danish, Dutch, English, Finnish, French, German, Norwegian, and Swedish.
  * Added page-range handling with condensed navigation for large result sets.
  * Added examples demonstrating pagination behavior, localization, loading, page sizes, and changing totals.
* **Tests**
  * Added comprehensive coverage for navigation, validation, accessibility, localization, and edge cases.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


Kiarokh:
> 🤖 AI-generated review from 6 parallel agents at commit `85b9dd587`. Treat as input, not a verdict — agents can be wrong or miss context.
>
> 🔄 Updated review. The agents reviewed `12a9a13`; everything below has since been addressed and the branch squashed and force-pushed to `85b9dd587`. ✅ Fixed = resolved since the review ran. The fixes were verified by the test suite and CI, not by a second agent pass.

# Consolidated PR Review

## PR Summary

Review tier: Full — 2502 changed lines across 23 files after noise filtering, nothing filtered out, no security-sensitive paths; six reviewers ran (Backward Compatibility, Code Quality, Observability and Performance on `opus`; Architecture and Security on `fable`). The PR adds a new `limel-pagination` component: a controlled page-number strip that emits `goToPage` with `page` and `offset` and never moves on its own, clamps to the last page when the set shrinks and reports that as `reason: 'clamped'`, holds its shape while `totalItems` is `null`, and announces page changes through a live region. It ships with seven examples, a pure window-layout helper, spec coverage for both, nine translation keys in eight languages, and a `@beta` tag on the whole surface. 1 commit, 23 files.

The review found no blockers: the PR is purely additive, touches only `src/interface.ts` and the translation files outside its own directory, and nothing reviewed was a regression. All three Medium findings sat in new code the PR owns and have been fixed, along with five of the eight nits.

- **Blockers:** None
- **Still open:** three nits, deliberately left. Reasons in the following comment.

One change in this SHA did not come from these agents: `render` now returns a `<Host>` wrapper rather than an array of two keyed elements, from a CodeRabbit suggestion. Verified against the code first — the two `key` props were on a static two-element array and did nothing, and `<Host>` is the more common shape in this library (45 of 93 component files).

---

## Dimensions

Nothing existing is removed, renamed or re-typed; the only two new flat exports, `GoToPageEvent` and `GoToPageReason`, collide with nothing in this repo or its consumers.

**What works well:** `@beta` is applied consistently to the component, both exported types and the API report, so widening `GoToPageReason` later is explicitly allowed; all eight translation files gain the same nine keys with no duplicates; the event carries `offset` so consumers never re-derive it.

**Issues:** None.

### 2. Code Quality — GOOD ✅ *(was NEEDS ATTENTION)*

**Issues:**
- **✅ Fixed** Validating getters mutated state and logged during render. See **Top Recommendation 2**.
- **✅ Fixed** Loading UI and ARIA attributes had no test. See **Top Recommendation 3**.

**What works well:** Each spec case says what would break if it were wrong; the numeric edges that bite (`NaN`, `Infinity`, `2 ** 53`, prop assignment order within one tick) are covered directly; every fallback is warned about and documented where it is applied.

**Minor nits:**
- ✅ Fixed: `rangeIdentity` and `rangeLabel` computed the same `from`/capped `to` with different "no count" sentinels, and `emitGoToPage` computed the offset a third time. All three now go through one `offsetOf`/`itemsOn` pair.
- ✅ Fixed: the page-size example's title ran into its first paragraph with no blank line, so it would not have rendered as a heading at all; its second paragraph also restated the first.
- ✅ Fixed: the new `export *` split the contiguous picker block in `src/interface.ts`.

### 3. Architecture — GOOD ✅ *(was NEEDS ATTENTION)*

**Issues:**
- **✅ Fixed** Validated inputs were re-derived everywhere instead of settled in `view`. See **Top Recommendation 2**.

**What works well:** The component owns no navigation state and `reason: 'clamped'` separates corrections from user navigation; settling in `componentWillRender` rather than per-prop watchers makes the result independent of assignment order; `getPageSlots` is a pure, separately tested function.

**Minor nits:**
- ✅ Fixed: `GoToPageEvent` carried both `pageSize` and `limit`, always set from the same variable. `limit` is gone.
- ✅ Fixed (in part): `lastKnownTotalItems` no longer carries `@State()`, which did nothing there. `liveMessage` keeps it — see the following comment.
- 🔵 Still open: the `gap` variant of `PageSlot` carries `from`/`to` that `renderSlot` ignores.

### 4. Security — GOOD ✅

Every string reaches the DOM through JSX text nodes and attributes, and all numeric props are gated by `Number.isSafeInteger` before arithmetic or allocation.

**What works well:** `getPageSlots` caps the rendered window at seven slots so a huge `totalItems` cannot allocate an unbounded array; a malformed `language` is caught around `Intl.NumberFormat`; no new dependencies, network or storage access, and warnings carry only the offending value.

**Issues:** None.

### 5. Observability — GOOD ✅ *(was NEEDS ATTENTION)*

**Issues:**
- **✅ Fixed** An unusable `language` was swallowed without any warning. See **Top Recommendation 1**.

**What works well:** The `page`, `pageSize` and `totalItems` warnings name the prop, the value and the fallback with the tag prefix, matching the convention in ai-avatar, table and flex-container; `warnOnce` keys on prop plus value and clears on a good value, both halves covered by tests; the `clamped` reason tells consumers in-band when the component is not showing what they asked for.

**Minor nits:**
- ✅ Fixed: the `totalItems` warning said "Treating the count as not yet known" while the component actually kept the last count it had. It now says so.

### 6. Performance — GOOD ✅

Fixed-size slot allocation, once-bound handlers and a cached `Intl.NumberFormat` keep per-render cost flat regardless of set size.

**What works well:** `getPageSlots` allocates seven slots whether there are ten pages or a million; the component holds no timers, observers or document listeners, so nothing leaks on removal.

**Minor nits:**
- 🔵 Still open: `renderArrowTooltip` returns nothing while `loading` is set, so both arrow tooltips are torn down when a fetch starts and rebuilt when it ends.

---

## Top Recommendations

1. **✅ Fixed** — Warn when `language` cannot be used, like the other props do
   <details>

   *Introduced by this PR · small fix. Was recommended: belongs in this PR.*
   - **Where:** `src/components/pagination/pagination.tsx` (`createNumberFormat`)
   - **What:** In the `catch`, warn once through `warnOnce` with the bad value and the `en` fallback. Also compare `resolvedOptions().locale` against the requested tag, because a well-formed but unsupported tag such as `zz` does not throw and instead resolves to the viewer's machine locale.
   - **Why:** `en_US`, `sv_SE` and `''` all throw and are easy for a host to produce from a backend locale string or an empty `document.documentElement.lang`; in the `zz` case the control renders English labels beside machine-locale digit grouping, which is exactly the mix the comment in this `catch` says it exists to prevent.

   </details>

   Fixed by comparing the resolved primary subtag against the requested one through the repo's own `getPrimarySubtag`, so both the throwing and the silently-substituted cases warn. Two tests, one of them for `zz` specifically.

2. **✅ Fixed** — Settle the validated `totalItems` and `pageSize` into `view` once per render
   <details>

   *Introduced by this PR · medium fix. Was recommended: belongs in this PR.*
   - **Where:** `src/components/pagination/pagination.tsx` (`View`, `componentWillRender`, `knownTotalItems`, `usablePageSize`, `rangeLabel`)
   - **What:** Widen `View` to `{ page, pageCount, pageSize, totalItems }`, run the two validators exactly once in `componentWillRender`, and have the readers use `this.view` instead of the getters.
   - **Why:** Reading `this.knownTotalItems` or `this.usablePageSize` re-ran validation, mutated `warnedAbout` and could call `console.warn`, and a single render did that roughly a dozen times through seven tooltips; a call site like `this.rangeLabel(page)` gave no hint it might write to the console.

   </details>

   Fixed as described. The settle step absorbed `rememberTotalItems` and the `pageCount` getter, both of which are now gone.

3. **✅ Fixed** — Add tests for the loading UI and the ARIA attributes
   <details>

   *Introduced by this PR · medium fix. Was recommended: belongs in this PR.*
   - **Where:** `src/components/pagination/pagination.spec.tsx`, covering `renderSpinner`, `renderTooltip`, `renderArrowTooltip`
   - **What:** Assert a `limel-spinner` inside the current page button, `aria-busy="true"` on the `nav` and `aria-disabled="true"` on the buttons; add one test that a page button renders a nested `limel-tooltip` whose `elementId` matches the button's `id`.
   - **Why:** The `aria-disabled`-instead-of-`disabled` choice is defended in a comment as what keeps an end-of-set arrow focusable for screen-reader users, but swapping it for the real `disabled` attribute, removing the spinner or dropping the tooltips entirely would all have left every test green, because clicks are guarded separately in `selectPage` and `step`.

   </details>

   Fixed with four tests, including the negative case — that `aria-busy`, `aria-disabled` and the spinner are *absent* once the page lands — and one asserting every button's accessible name.

---

Verified after the fixes: 1320 spec tests (81 on pagination), 16 example tests, eslint clean, and `api:verify` passing on a regenerated report.

Kiarokh:
*Generated by Claude Opus 5.*

Three nits from the review above are deliberately left in. Recording why, so the next person reading the review does not have to work out whether they were missed.

### `liveMessage` keeps its `@State()` — I disagree with the finding

The review grouped `liveMessage` with `lastKnownTotalItems`: both are only written inside `componentWillRender`, where the setter cannot schedule an update because one is already in flight, so the decorator has no effect and misleads about data flow.

That is right about `lastKnownTotalItems`, and it has been dropped there. It is pure bookkeeping — `render` never reads it, only the settle step does.

`liveMessage` is a different case, and the two should not have been grouped. It **is** read by `render`. That it currently only ever gets written during an already-scheduled render is incidental to the field, not a property of it: the day someone sets it from a click handler or a `@Watch`, `@State()` is what makes the live region update, and without it the announcement would silently stop working. `@State()` is the honest annotation for "this affects render", so it stays.

### `PageSlot`'s `gap` keeps its `from` and `to`

Correct that `renderSlot` ignores them today and only the util tests read them. Two reasons to keep them.

They are what jump-to-page needs (#4298): a gap has to know which pages it hides before it can offer them. And `PageSlot` is internal — deliberately not routed through `src/interface.ts`, unlike `GoToPageEvent` — so nothing public is being widened by carrying two numbers nobody reads yet. Deleting and re-adding them costs more than leaving them, and the type reads better for describing what a gap actually stands for.

This one is genuinely arguable against the PR's own "no API before a consumer" principle. The principle is about the **public** surface; these are not on it.

### The arrow tooltips still unmount while `loading`

Correct, and it is the largest of the three: both arrow tooltips — each a `limel-tooltip`, a `limel-portal` and a `limel-tooltip-content` — are torn down when a fetch starts and rebuilt when it ends, so roughly twelve elements connect and disconnect per page change.

Left out because it is not a tidy-up. `renderArrowTooltip` returns nothing when the arrow is disabled, and `isArrowDisabled` is true whenever `loading` is. Gating the tooltip on end-of-set alone would keep it mounted, but it also means a temporarily inert arrow goes on describing where it would take you. I think that is the better behaviour — it answers "where does this go?" while you wait rather than going blank — but it is a decision about what the control says, not a refactor, and it deserves to be made deliberately rather than folded into a performance fix.

Tracked in #4304 with the rest of the arrow rendering, alongside the related point that `isArrowDisabled` is computed twice per arrow.

### For completeness

Everything else the review raised is fixed in `85b9dd587`: all three Medium findings and five of the eight nits. Two nits turned out to be worth more than their severity suggested — the page-size example's heading ran into its first paragraph with no blank line, so it would not have rendered as a heading at all on the docs site, and `GoToPageEvent` carried both `pageSize` and `limit` always set from the same variable, which is the same "two names for one value" shape this PR rejects for `totalPages`. `limit` is gone; consumers read `pageSize`.

Kiarokh (approved):

PR #4305, opened 2026-09-17
fix(table): defer sorting when the table is not yet created

Befkadu1:
fix: https://github.com/Lundalogik/crm-client/issues/1323

`updateSorting` was the only `@Watch` in `table.tsx` dereferencing
`this.tabulator` without checking it exists. `pageChanged`, `activeRowChanged`,
`updateData`, `updateColumns`, `updateAggregates` and `updateSortableColumns`
all guard; this one did not, so a `sorting` change arriving before the
Tabulator instance is built throws:

```
TypeError: Cannot read properties of undefined (reading 'setSort')
    at updateSorting (table.tsx:512)
```

- Guards on `!this.tabulator || !this.initialized` and defers through
  `shouldSort`, the file's existing channel — `handleRenderComplete` applies it
  via `setSort(getInitialSorting())`. A bare `return` would silently drop the
  sort instead.
- Four spec tests added. Two of them fail on `main` with the exact production
  error, so they are real regression tests rather than restatements.
- `npx vitest run src/components/table --project spec` → 91 passed.
- Known gap, deliberately left alone: `getInitialSorting()` prefers
  `currentSorting`, which `init()` never clears when it resets `initialized`, so
  after a rebuild the deferred sort re-applies the old sorters. Pre-existing —
  the two older `shouldSort` call sites route through the same path — and not
  what this fixes, which is the first build, where `currentSorting` is
  undefined. Filing separately.

Seen in the webclient's deal explorer: `table-view` renders `sorting` as
`{columns[0], 'ASC'}` while the layout request is in flight, then swaps in the
stored order. When that stored order differs — a saved `DESC` sort — the prop
changes and reaches the unguarded line. A stored `ASC` sort matches the
interim value, so `isEqual` returns one line earlier and hides the bug, which
is why it looks user-specific.


FredrikWallstrom (approved):

src/components/table/table.tsx
```diff
@@ -503,6 +503,12 @@ export class Table {
...
     ) {
+        if (!this.tabulator || !this.initialized) {
+            this.shouldSort = true;
+
+            return;
+        }
+
         const newSorting = this.getColumnSorter(newValue);
         const oldSorting = this.getColumnSorter(oldValue);
         if (isEqual(newSorting, oldSorting)) {
```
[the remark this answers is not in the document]
Befkadu1: `shouldSort` is a boolean and `getInitialSorting()` reads `this.sorting` at apply time, so the latest prop always wins. There's no pending value to go stale and nothing lost across multiple pre-init updates.

Gating `handleRenderComplete` on `initialized` would regress it: the `tableBuilt` handler only re-renders `if (this.data?.length)`, so on an empty table the sort would never apply at all.

There is a real issue nearby though — `getInitialSorting()` prefers `currentSorting`, which `init()` never clears when it resets `initialized`, so after a rebuild the deferred sort re-applies the old sorters. Pre-existing, and not what this PR fixes (first build, `currentSorting` is undefined). Filing separately.

FredrikWallstrom (approved):

src/components/table/table.tsx
```diff
@@ -503,6 +503,12 @@ export class Table {
         newValue: ColumnSorter[],
         oldValue: ColumnSorter[]
     ) {
+        if (!this.tabulator || !this.initialized) {
+            this.shouldSort = true;
```
Befkadu1: **question (blocking):** Is `currentSorting` really undefined on the first build in remote mode?

`requestData` (line 934) sets it from Tabulator's `initialSort` while the table is building, and `initialized` only flips in the `tableBuilt` handler. If the ajax call lands first, `handleRenderComplete` applies the interim sorters through `getInitialSorting()` and the new prop is dropped — no crash, but the saved `DESC` sort still does not show. Recording the prop in the defer branch closes that window:

```suggestion
            this.currentSorting = newValue;
            this.shouldSort = true;
```

src/components/table/table.spec.ts
```diff
@@ -382,3 +384,58 @@ describe('limel-table has-aggregation detection', () => {
...
+        (component as any).tabulator = { setSort: vi.fn() };
+        (component as any).initialized = false;
+
+        (component as any).updateSorting(sorting, []);
+
+        expect((component as any).tabulator.setSort).not.toHaveBeenCalled();
+        expect((component as any).shouldSort).toBe(true);
+    });
+
+    it('applies the deferred sort once the table has rendered', () => {
```
Befkadu1: **suggestion (if-minor):** The deferred-sort test leaves `currentSorting` undefined, which is the one case where `getInitialSorting()` cannot pick the wrong value.

```diff
 it('applies the deferred sort once the table has rendered', () => {
     (component as any).tabulator = { setSort: vi.fn() };
+    (component as any).currentSorting = [
+        { column: { field: 'name' }, direction: 'DESC' },
+    ];
     (component as any).sorting = sorting;
```

Copy-paste — the assertion on `dir: 'asc'` then covers the precedence.

PR #4307, opened 2026-09-18
Popover listeners and aria

LucyChyzhova:
Two defects in how `limel-popover` treats the trigger element it is handed, and the same accessibility fix in `limel-menu`, which carries a copy of the same code. No new behaviour; nothing here is specific to the component that turned them up.

## The document click listener was never removed

The listener is added with `{ capture: true }` and was removed without it. `removeEventListener` only matches a listener registered with the same capture flag, so the removal never matched and the listener stayed on `document` for the life of the page.

Nothing misbehaved, because the handler checks `open` before acting — but every popover that had ever been opened kept a capturing `document` listener, and that listener holds its host, so a popover removed from the page could not be collected.

The popover also had no `disconnectedCallback`, so one removed while still open went on listening. Setup now runs on connect and teardown on disconnect, and both use the same capture flag.

## `aria-expanded` disappeared instead of saying `false`

`setTriggerAttributes` wrote the attribute when open and **removed** it when closed. A closed trigger therefore read as an ordinary button: nothing announced that there was anything to open, and closing it announced nothing at all. It is now written in both states, which is what the attribute is for.

## Why `limel-menu` is in here too

`limel-menu` has its own copy of the same four lines. Fixing only the popover would have left the two writing different DOM onto a consumer-supplied trigger — a closed popover trigger reading `aria-expanded="false"` while a closed menu trigger had no attribute at all — so a consumer with one selector or one test covering both would have had to handle two shapes. The menu keeps the remove-when-false path for `disabled`, because there the presence of the attribute is what disables the element.

## What changes for consumers

The one visible change is that closed triggers now carry `aria-expanded="false"` where they previously carried nothing. I checked every `aria-expanded` selector in lime-elements, lime-crm-components and lime-crm-building-blocks: they are all either `[aria-expanded='true']`, `[aria-expanded]:not([aria-expanded='false'])`, or a `:not([aria-expanded='true'])` negation. None of them matches a bare `[aria-expanded]`, so nothing restyles.

## Testing

The full suite passes (2457 tests). These three commits add no tests of their own, so the two behaviours are not pinned: I can add a `popover.e2e.tsx` asserting that a closed trigger reads `aria-expanded="false"` and that removing an open popover leaves no capturing listener on `document`, if you would rather have that before this merges.

---

## Summary by CodeRabbit

* **Accessibility**
  * Trigger elements now consistently expose `aria-expanded`, `aria-haspopup`, `aria-controls`, and appropriate roles.
  * Menu-managed disabled states are updated without overriding disabled settings applied directly to triggers.

* **Bug Fixes**
  * Menus and popovers no longer respond to keyboard or outside-click interactions after being removed from the page.
  * Event listeners are properly cleaned up when components are disconnected.

* **Tests**
  * Added coverage for ARIA state changes, disabled attributes, hotkeys, outside clicks, and Escape-key behavior.

## Review:

### Browsers tested:

Windows:
- [ ] Chrome
- [ ] Edge
- [ ] Firefox

Linux:
- [ ] Chrome
- [ ] Firefox

macOS:
- [ ] Chrome
- [ ] Firefox
- [ ] Safari

Mobile:
- [ ] Chrome on Android
- [ ] iOS


Kiarokh:
> 🤖 AI-generated review from 7 parallel agents at commit `65d602d71`. Treat as input, not a verdict — agents can be wrong or miss context. Nothing in the PR has been changed by this review.
>
> 🔄 Updated review. The previous review was at `0a96cd8a9`. ✅ Fixed = resolved since then; 🆕 New = flagged for the first time.

# Consolidated PR Review

## PR Summary

Review tier: Lite — 216 changed lines across 4 files after noise filtering. The PR fixes two defects in how `limel-popover` treats the trigger it is handed, and applies the same accessibility fix to `limel-menu`, which carries a copy of the code: `aria-expanded` is now written as `"false"` when closed instead of being removed, and the document-level `click` listener is now removed with the same `capture: true` flag it was added with, from a new `disconnectedCallback`. A follow-up commit stops the menu stripping a `disabled` attribute the consumer set on their own trigger. 4 commits, 4 files, +180/−36. The branch was rebased since the last review, so everything was re-reviewed at the new head rather than as a pure delta.

The PR is clearly better than `main`: it closes a document listener leak that also swallowed clicks page-wide, makes the trigger's collapsed state readable to screen readers, and stops the menu re-enabling a button the consumer disabled. Both items from the previous review are fixed. Two Medium items remain, both cheap and both inside code the PR already changes — one of them was logged last time as a minor nit and is carried as Medium now that a reviewer verified what actually happens when it triggers.

- **Blockers:** None

---

## Dimensions

Short summaries. Each issue points to a numbered Top Recommendation below.

No breaking changes; the one observable DOM difference is the stated point of the PR.

**What works well:** No prop, event, method or slot signature changed, and the `componentWillLoad` → `connectedCallback` move is internal. A reviewer independently re-ran the check the PR description claims: every `aria-expanded` selector in this repo already guards on `[aria-expanded='true']` or excludes `'false'`, and no component reads that attribute back off a trigger, so nothing restyles when closed triggers gain `aria-expanded="false"`. The `disabled` change only widens correctness — a consumer's own attribute used to be stripped on every render and now survives.

**Issues:** None.

### 2. Code Quality — NEEDS ATTENTION ⚠️

The new tests reproduce both original bugs, but the branch added by the newest commit is itself untested.

**Issues:**
- **🆕 🟡 [Medium]** No test exercises the path where the menu removes a `disabled` attribute it added (`src/components/menu/menu.tsx:872-886`). Nothing flips `disabled` from `true` back to `false`, so a regression would leave a trigger permanently disabled and the suite would still pass. See **Top Recommendation 2** below.

**What works well:** The new tests in `popover.e2e.tsx` and `menu.e2e.tsx` directly reproduce both bugs the PR set out to fix and would fail against the pre-fix code; `setupGlobalHandlers` / `teardownGlobalHandlers` replaces the old per-branch add/remove duplication with one add path and one shared remove path, so the capture flag is written once; inline comments are one line each and explain genuinely non-obvious behaviour rather than restating the code.

**Minor nits:**
- 🔵 The four trigger attributes are still hand-written in both components and could live in a shared util next to `src/util/focus-trigger-element.ts`; the two copies have diverged further, since the menu's now also owns `disabled` and calls a second helper (`src/components/menu/menu.tsx:853-863`, `src/components/popover/popover.tsx:226-232`). The comment-wording drift half of this nit is fixed — both read the same now.
- 🆕 New: The disconnect regression test only dispatches a click, so a future change that drops the `keydown` removal from `teardownGlobalHandlers` would still pass (`src/components/popover/popover.e2e.tsx:100-117`).

### 3. Architecture — NEEDS ATTENTION ⚠️

The listener lifecycle is now a matched pair, but setup still keys purely off `open` and ignores whether the host is in the document.

**Issues:**
- **🟡 [Medium]** `setupGlobalHandlers` attaches the document listeners on any `open` change without checking `this.host.isConnected` (`src/components/popover/popover.tsx:125-136`). Setting `open = true` on an already-removed popover re-arms the capturing click listener with nothing left to take it off. See **Top Recommendation 1** below.

**What works well:** Setup on connect and teardown on disconnect is what `limel-portal` asks of components whose nodes get moved, so a popover nested in portal-relocated content re-arms itself; add and remove now go through one matched pair of helpers, so the capture flag cannot drift apart again; fixing the menu's copy in the same change keeps both components writing the same DOM onto a consumer-supplied trigger.

**Minor nits:**
- 🆕 New: `triggersDisabledByMenu` records ownership only at the moment the menu adds `disabled`, so if the consumer also disables that trigger afterwards and `menu.disabled` goes back to `false`, the menu removes the attribute anyway (`src/components/menu/menu.tsx:872-886`).
- 🆕 New: The trigger gets `aria-controls="<portalId>"` before `limel-portal` has ever created an element with that id, so a never-opened trigger points at nothing (`src/components/popover/popover.tsx:231`, `src/components/menu/menu.tsx:855`).

### 4. Security — GOOD ✅

No new exposure; the change removes a real one.

**What works well:** On `main`, a popover removed from the page while open left a capturing `click` listener on `document` calling `preventDefault()` and `stopPropagation()` on every click for the life of the page — that is now gone. All attribute writes use fixed ARIA names plus the component's own random portal id, so nothing consumer- or user-controlled reaches an attribute name or a DOM sink, and the id is never treated as a secret.

**Issues:** None.

### 5. Observability — GOOD ✅

Nothing to log here, and the new comments capture the reasoning that would otherwise have to be rediscovered.

**What works well:** The comments on `teardownGlobalHandlers` and `setTriggerDisabled` record why the capture flag must be repeated on removal and why only a menu-added `disabled` may be taken away; no existing error handling is removed or weakened.

**Issues:** None.

### 6. Performance — GOOD ✅

A small net improvement.

**What works well:** The capture-phase click listener now actually detaches instead of piling up on `document` once per popover ever opened and pinning each removed host in memory, `disconnectedCallback` releases the listeners when an open popover is removed, and the new `triggersDisabledByMenu` bookkeeping uses a `WeakSet` so it cannot retain a detached trigger.

**Issues:** None.

---

## Top Recommendations

These are suggestions for the author. Nothing has been applied.

1. **🟡 [Medium from Architecture]** — Only attach the document listeners while the popover is in the document
   *Pre-existing · small fix · belongs in this PR — the PR rewrites this exact function and adds the disconnect half of the pair.*
   - **Where:** `src/components/popover/popover.tsx:125-136` (`setupGlobalHandlers`), reached from `watchOpen` at `popover.tsx:98-106`
   - **What:** Tear down and return early when `!this.host.isConnected`, so the listeners exist only while both `open` and connectedness hold — the same guard `limel-portal` already uses at `src/components/portal/portal.tsx:150`.
   - **Why:** Stencil fires `@Watch` callbacks on detached instances (the `isWatchReady` flag is never cleared on disconnect), so a consumer setting `open = true` on a popover they have already removed puts the capturing click listener back with no later `disconnectedCallback` to take it off; since `portalContains` is false for everything once the host is detached, every click on the page is then cancelled and a `close` event emitted. This was logged as a minor nit last time and is carried as Medium now that the consequence has been verified.

2. **🆕 New — 🟡 [Medium from Code Quality]** — Test the path where the menu gives a trigger its `disabled` attribute back
   *Introduced by this PR · small fix · belongs in this PR.*
   - **Where:** `src/components/menu/menu.tsx:872-886` (`setTriggerDisabled`)
   - **What:** Add a test that renders `<limel-menu disabled={true}>`, calls `setProps({ disabled: false })`, and asserts the trigger's `disabled` attribute is gone.
   - **Why:** That removal branch is the whole reason the `triggersDisabledByMenu` `WeakSet` exists, and no existing test flips `disabled` back to `false`; if it regresses, a trigger the menu disabled stays disabled forever and the suite still passes.

3. **✅ Fixed** — Stop clearing a `disabled` attribute the menu did not set
   <details>

   *Pre-existing · small fix · belongs in this PR — the PR rewrites this exact function.*
   - **Where:** `src/components/menu/menu.tsx:852-868` (`setTriggerAttributes`), reached from `componentDidRender` at `menu.tsx:234`
   - **What:** Only write or clear `disabled` when the menu is the one that set it — for example track whether this component stamped it, or express the menu's own state with `aria-disabled` — instead of calling `removeAttribute('disabled')` unconditionally whenever the menu is enabled.
   - **Why:** A consumer writing `<limel-button disabled slot="trigger">` inside an enabled `limel-menu` has the attribute stripped on every render, and because `limel-button`'s `disabled` prop is declared `@Prop({ reflect: true })` (`button.tsx:69-70`), that sets the prop back to `false` and silently re-enables their button.

   </details>

4. **✅ Fixed** — Add tests for the two behaviours this PR fixes
   <details>

   *Introduced by this PR · small fix · belongs in this PR.*
   - **Where:** `src/components/menu/menu.e2e.tsx:41-52` (`sets the correct aria attributes on the trigger`) and `src/components/popover/popover.e2e.tsx`
   - **What:** Assert in the menu test that `aria-expanded` is `"false"` when closed and `"true"` when open, and add a popover case that opens, closes and reopens the component, confirming the close handler fires exactly once.
   - **Why:** Neither fix is covered today, and the capture-flag mismatch in particular is the kind of mistake that reappears silently — `removeEventListener` simply does nothing when the options do not match.

   </details>

Kiarokh:
> 🤖 AI-generated review from 7 parallel agents at commit `b106da78d`. Treat as input, not a verdict — agents can be wrong or miss context. Nothing in the PR has been changed by this review.
>
> 🔄 Updated review. The previous review was at `4b0a3d12d`. ✅ Fixed = resolved since then; 🆕 New = flagged for the first time.

# Consolidated PR Review

## PR Summary

Review tier: Lite — 84 changed lines across 2 files after noise filtering, reviewed as a delta on top of `4b0a3d12d`. The PR fixes two defects in how `limel-popover` treats the trigger it is handed — `aria-expanded` is now written as `"false"` when closed instead of being removed, and the document listeners are taken down with the same capture flag they went up with — and applies the same fixes to `limel-menu`, which carries a copy of the code. The full PR is 5 commits, 4 files, +275/−60; this round covers the newest commit, which gives the menu the popover's setup/teardown pair plus an `isConnected` guard, and one test for it (+60/−24 across `menu.tsx` and `menu.e2e.tsx`).

The PR is clearly better than `main`, and the one Medium carried into this round from Architecture is now fixed: the menu no longer arms a capturing `document` keydown listener on an instance that has left the page, and a test pins it. What remains is a single missing test, unchanged since the last review — the guard that stops the menu re-enabling a button the consumer disabled is still only exercised in the cases where it does nothing.

- **Blockers:** None

---

## Dimensions

Short summaries. Each issue points to a numbered Top Recommendation below.

Nothing in the delta touches the public surface; the one observable DOM difference is still the stated point of the PR.

**What works well:** No prop, event, method or slot changed — `setupGlobalHandlers` and `teardownGlobalHandlers` are private, and for a normally connected menu the new path behaves exactly as the old one did. A reviewer independently re-ran the check the PR description claims: every `aria-expanded` selector in this repo guards on `[aria-expanded='true']` or excludes `'false'`, so nothing restyles when closed triggers gain the attribute. The only behaviour change is that a menu whose `open` watcher fires after removal no longer arms a listener it cannot take down — a leak fix, not something a caller could depend on.

**Issues:** None.

**Minor nits:**
- 🔵 A closed trigger's `aria-expanded` is now the string `"false"`, which is truthy, so any JS testing it with `getAttribute` or `hasAttribute` would flip to always reporting "expanded" (`src/components/popover/popover.tsx:234`, `src/components/menu/menu.tsx:862`). No such read exists in lime-elements, lime-crm-components or lime-crm-building-blocks — the PR description's audit covered CSS selectors only, so the residual risk is external integrators.

### 2. Code Quality — NEEDS ATTENTION ⚠️

The new listener test targets exactly the bug the commit fixes, but the `disabled` guard from the previous commit is still untested in the one case it exists for.

**Issues:**
- **🟡 [Medium]** No test puts the menu and its trigger in conflict over `disabled` (`src/components/menu/menu.e2e.tsx:105-171`). Each existing test has a single, unambiguous owner, so deleting the `hasAttribute` guard would keep the suite green. See **Top Recommendation 1** below.

**What works well:** The duplicated add/remove-listener logic that lived in `connectedCallback`, `disconnectedCallback` and `openWatcher` is collapsed into one named pair (`menu.tsx:291-329`), so the capture flag is written once and the two sides cannot drift; the comment about Stencil firing `@Watch` after disconnect is adapted to the menu's actual consequence rather than copy-pasted from the popover's wording; the new test would fail against the pre-fix code.

**Minor nits:**
- 🔵 The four trigger attributes are still hand-written in both components and could live in a shared util next to `src/util/focus-trigger-element.ts`, which both already use (`src/components/menu/menu.tsx:855-864`, `src/components/popover/popover.tsx:230-237`).
- 🔵 The popover's reconnection-guard test only dispatches a click, so a regression that re-arms just the `keydown` listener on a detached, reopened popover would go uncaught (`src/components/popover/popover.e2e.tsx:133-155`).
- 🆕 New: The menu has no test for plain teardown — removing an open menu with no later `open` change — although the popover carries that case as its own test (`src/components/menu/menu.e2e.tsx:580-612` versus `src/components/popover/popover.e2e.tsx:111-131`).

### 3. Architecture — GOOD ✅

The menu now has the popover's shape: one reconcile function behind connect, disconnect and the `open` watcher, with the same connectedness guard.

**What works well:** Funnelling all three entry points through one function means the arm and disarm sides cannot drift apart, which is the class of bug this PR is fixing; the `isConnected` half of the guard is now pinned by a test rather than only described in a comment; the shape matches what `limel-portal` asks of components whose nodes it moves, so a relocated menu re-arms instead of going deaf.

**Issues:**
- **✅** The menu's listener setup now checks the host is still in the page — fixed. See **Top Recommendation 2**.

**Minor nits:**
- 🔵 `triggersDisabledByMenu` records ownership only at the moment the menu adds `disabled` and never re-checks, so if the consumer also disables that trigger afterwards and `menu.disabled` goes back to `false`, the menu removes the attribute anyway (`src/components/menu/menu.tsx:874-888`).
- 🔵 The trigger gets `aria-controls="<portalId>"` on every render, while `limel-portal` only creates the element carrying that id once it becomes visible, so a never-opened trigger points at an id that is not in the document (`src/components/popover/popover.tsx:235`, `src/components/menu/menu.tsx:857`).
- 🆕 New: The new guard stops the listener being armed on a detached menu, but the line below it still calls `setFocus()` when `open` flips to true on that same detached instance, leaving a stray timeout and `IntersectionObserver` on nodes outside the document (`src/components/menu/menu.tsx:299-308`).
- 🆕 New: The reconcile-against-`open`-and-`isConnected` pattern now exists as two near-identical copies with different spellings of the capture flag, and only the popover's carries the comment explaining why the flag must be repeated on removal (`src/components/menu/menu.tsx:310-329`, `src/components/popover/popover.tsx:125-152`).

### 4. Security — GOOD ✅

No new exposure; the delta removes one.

**What works well:** The `isConnected` guard stops an open-but-removed menu keeping a capture-phase `keydown` listener on `document`, so a detached component can no longer see or swallow keystrokes anywhere on the page, and teardown passes the same capture flag as setup so the removal actually matches.

**Issues:** None.

### 5. Observability — GOOD ✅

Nothing to log here, and the new comment captures reasoning that would otherwise have to be rediscovered.

**What works well:** `setupGlobalHandlers` explains exactly why a detached instance must not attach a listener, which is the context a future debugger needs for this class of leak, and the new lifecycle fails safe — any state that does not warrant a listener tears one down rather than leaving it attached.

**Issues:** None.

### 6. Performance — GOOD ✅

A small net improvement, with no new work on a hot path.

**What works well:** A menu removed while closed and flipped to `open` afterwards no longer re-attaches a `document` keydown listener that would hold the detached host alive; `handleDocumentKeyDown` stays a single stable reference, so repeated add/remove calls cannot accumulate duplicates; setup and teardown run only on connect, disconnect and `open` changes, not per render.

**Issues:** None.

---

## Top Recommendations

These are suggestions for the author. Nothing has been applied.

1. **🟡 [Medium from Code Quality]** — Test the case where the menu and the consumer both want the trigger disabled
   *Introduced by this PR · small fix · belongs in this PR.*
   - **Where:** `src/components/menu/menu.tsx:874-888` (`setTriggerDisabled`), tests at `src/components/menu/menu.e2e.tsx:105-171`
   - **What:** Add a test that renders `<limel-menu disabled={true}>` around `<button slot="trigger" disabled={true}>`, calls `setProps({ disabled: false })` on the menu, and asserts the trigger still has `disabled`.
   - **Why:** The existing tests each have one unambiguous owner, so removing the `if (!element.hasAttribute('disabled'))` guard — the menu then claiming every disabled trigger as its own — passes all of them, and the consumer's button is silently re-enabled the first time the menu toggles back to enabled. That is exactly the bug commit `4b0a3d12d` was written to prevent.

2. **✅ Fixed** — Give the menu the same connection check the popover just got
   <details>

   *Pre-existing · small fix · belongs in this PR — the PR touches this file and adds exactly this guard to the popover's twin.*
   - **Where:** `src/components/menu/menu.tsx:309-326` (`openWatcher`)
   - **What:** Only add the document `keydown` listener when `this.host.isConnected` as well as `open` — ideally by giving the menu the popover's shape, one setup/teardown pair called from `connectedCallback`, `disconnectedCallback` and the watcher, instead of adding and removing inline.
   - **Why:** Stencil keeps firing `@Watch` on a detached instance, so a consumer setting `open = true` on a menu they have already removed puts a capturing `document` listener back with no `disconnectedCallback` left to take it off; it holds the menu instance alive, and a hotkey press then runs selection on a component that is no longer in the page.

   </details>

Kiarokh (approved):

